PR #56, opened 2014-10-01
Implementation of login page design

Carreau:
Looks like that for now

capture d ecran 2014-10-01 a 16 44 46

Which not yet exactly what @awantulok  designed on https://github.com/jupyter/design/pull/4

Feel free to continue/use the PR this is mostly a test.

### DIFF
--- a/share/jupyter/hub/static/less/login.less
+++ b/share/jupyter/hub/static/less/login.less
@@ -1,4 +1,5 @@
-@login-primary: #ff9000;
+@login-primary: #f8A52D;
+@login-secondary: #E25C24;
 
 #login-main {
     display: table;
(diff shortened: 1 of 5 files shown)


Carreau:
6 month to rebase, but done.

Carreau:
And move the error login message.

capture d ecran 2015-03-03 a 16 53 24

minrk:
While this is probably an improvement for the login page, the effects of changing `brand-primary` to orange everywhere are not good:

screen shot 2015-03-04 at 14 20 09

screen shot 2015-03-04 at 14 20 02

I don't think orange is a good primary action color. Maybe special case the login button?

Carreau:
capture d ecran 2015-03-06 a 12 25 42


Done, login screen stays, centered now also. Color closer to logo. Apply only to login page.

Carreau:
We can also have transparent logo in the back (just to try how it looks)

capture d ecran 2015-03-06 a 12 38 59

share/jupyter/hub/static/less/login.less
```diff
@@ -27,7 +29,65 @@
         margin-top: -1px;
     }
     
-    button[type=submit] {
+    input[type=submit] {
         margin-top: 16px;
+        background-color: @login-primary;
+        background: linear-gradient(to bottom,  @login-primary 0%, @login-secondary 100%); /* W3C */
```
minrk: can you get rid of the gradients? They don't fit in well with the otherwise flat colors.

minrk: Actually, I think the gradient on the header is okay. But definitely no gradient on the button.

Issue #84, opened 2014-10-28
An active Jupyterhub session locks Postgres tables from writes.


ssanderson:
Noticed while testing permissions issues on #83.  It looks like maintaining a single open session for the entire lifetime of the App results in transactions never being closed, which causes Postgres to write-lock any tables being used by Jupyterhub.  From reading the SQLAlchemy docs, it looks like the expected pattern for webapps is to create a `Session` class at global scope, but create and tear down instances of the session on each request.

meeseeksmachine:
This issue has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/confusion-of-the-db-instance/3878/4

PR #85, opened 2014-10-29
[WIP] Scope database sessions to API requests.

ssanderson:
WIP fix for #84.  The basic idea is to lazily create session objects for each HTTP Request, cleaning them up during the request's `finish` method, with some extra wrinkles thrown in because some of the Hub's management routines also make database queries, and these have less obvious natural scopes than HTTP Requests.

There are a few issues with this strategy that I'm not sure how to solve:
1. When a database object is queried in a new session, it loses any Python-only attributes that have been tacked onto it.  This means, in particular, that the spawner column on User doesn't persist between requests (the lone remaining test failure is caused by this).  I'm not sure whether it's important to persist the Python-level spawner.  If it is important, I'm not sure whether it could be straightforwardly persisted in the database.
2. Making this work with the Hub's management routines is pretty awkward.  In particular, there's a lot of manual moving of the Proxy between database sessions, and there's at least one location where I'm nesting calls to `new_db_session`.  I could probably deal with the latter by making `new_db_session` idempotent.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -4,6 +4,7 @@
 # Copyright (c) IPython Development Team.
 # Distributed under the terms of the Modified BSD License.
 
+from contextlib import contextmanager, closing
 import io
 import logging
 import os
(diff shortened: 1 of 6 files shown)


meeseeksmachine:

https://discourse.jupyter.org/t/confusion-of-the-db-instance/3878/3

PR #96, opened 2014-11-19
DEV: Add DictionaryAuthenticator.

ssanderson:
Adds an authenticator that accepts a pre-defined dictionary from user -> hashed password.  This is useful for defining user lists that aren't tied directly to unix users.


Carreau:
this does not merge cleanly anymore, and seem to have stalled. 
I'm going to close, but fell free to reopen.

PR #126, opened 2015-01-14
DEV: Make login a base-class default on Authenticator.get_handlers.

ssanderson:
Rather than a module-level default.  This prevents `/login` from sticking around when another Authenticator replaces the login handler.


ssanderson:
https://github.com/jupyter/jupyterhub/issues/125

minrk:
Make sense

PR #132, opened 2015-01-28
add JupyterHub.admin_access

minrk:
optionally allow admin users to login to user servers by visiting a special admin-only URL that sets the relevant cookie

- disabled by default
- an 'access server' button is added to the admin panel, which sets the necessary cookie to log in to a given user server, one at a time
- each admin access is logged at warning-level, logging to admin user and the target user

ping @jhamrick

### DIFF
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -8,7 +8,7 @@ import json
 from tornado import gen, web
 
 from .. import orm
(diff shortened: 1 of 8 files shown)


jhamrick:
This looks super awesome! I will test it out on the deployment of JupyterHub I have in my lab and report back (I may not be able to get to this in the next few days, though, as I have a big paper deadline this weekend).

PR #142, opened 2015-01-29
admin page UI

### LINKED ISSUE #141: Admin UI suggestions
As I've been using the admin UI a bit for my class, I just had a few things that I thought would be nice to have/change in it:

* move the "add user" button to the top -- it is annoying with lots of users to have to scroll all the way down to find it
* add some simple statistics at the top of the page: number of running servers, number of currently active users, etc.
* it would be awesome to be able to sort users by the different columns -- admin, alphabetically, by last seen
* currently, if you shut down a user's server, it causes the page to refresh which often jumps you up to the top (or just somewhere else). It would be nice if the update could be done in the background without actually reloading the page.

Obviously, none of these are urgent, but I think they would make the admin experience a little easier. I can open separate issues for them if so desired.

minrk:
closes #141

- move user button to the top
- allow sorting columns
- add user, running count to header
- don't reload page on server stop/start from admin panel

screen shot 2015-01-29 at 14 27 26

### DIFF
--- a/jupyterhub/apihandlers/base.py
+++ b/jupyterhub/apihandlers/base.py
@@ -42,7 +42,7 @@ class APIHandler(BaseHandler):
             reason = getattr(exception, 'reason', '')
             if reason:
                 status_message = reason
-        
+        self.set_header('Content-Type', 'application/json')
         self.write(json.dumps({
             'status': status_code,
             'message': message or status_message,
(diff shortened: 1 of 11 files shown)


ssanderson:
@minrk this looks pretty slick.  How hard would it be to add additional metadata to this layout?  The thing I have in mind in particular cgroup memory accounting for docker containers.

minrk:
It shouldn't be too hard. I can make a few things blocks, which should make doing that with a custom template easier.

minrk:
@ssanderson with jinja blocks, I can give you an extension point for another column or two easily enough. They way things are done here, though, I can't make them sortable that easily. I *suspect* if you have significant information beyond the basic stuff here, especially if it touches other databases or other tables in the same database, that making your own admin tools that talk to JupyterHub via its REST API will be the more maintainable way to go.

ssanderson:
Fair enough

PR #143, opened 2015-01-30
DEV: Allow Authenticator.add_user to not be a future.

ssanderson:
In particular, this fixes an error that occurs if a user subclasses the base Authenticator class and doesn't override add_user.


minrk:
Definitely right, thanks.

PR #144, opened 2015-02-02
Logo links to Hub home instead of user URL

ckald:
Currently the Jupyter logo link brings us to the users' dashboard and there is no way to get to the initial screen (containing "Stop my server"/"My server"/"Admin" buttons)

### DIFF
--- a/share/jupyter/hub/templates/page.html
+++ b/share/jupyter/hub/templates/page.html
@@ -89,6 +89,7 @@
     <span id="login_widget">
       {% if user %}
         <a id="logout" class="btn navbar-btn btn-default pull-right" href="{{logout_url}}">Logout</a>
+        <a id="hub-home" class="btn navbar-btn btn-default pull-right" href="{{base_url}}">Hub</a>
       {% else %}
         <a id="login" class="btn navbar-btn btn-default pull-right" href="{{login_url}}">Login</a>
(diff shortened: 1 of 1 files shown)


minrk:
Can you change it so the logo sends you to the Hub home, instead of adding the new link?

ckald:
I thought this is an intended behaviour. If I will change this, then it will require 2 clicks to get to the user home screen

minrk:
>  If I will change this, then it will require 2 clicks to get to the user home screen

The only page normally served to a user is the page with the link to 'my server', so I don't think that's an issue.

ckald:
Okay, now I see the problem: the logo was already pointing to the {{base_url}}, but in that context this is not the hub page, but the users home page

minrk:
Thanks. There's some confusing in base_url and prefix in JupyterHub, since they have slightly different meanings from IPython. This is actually the original intended behavior.

minrk:
Oops, now that I've merged this, the original behavior was actually correct. While this results in no real change in behavior, this patch increases the number of redirects that result.

ckald:
I'm totally confused :) Will try to do this later again and more thoroughly

minrk:
Don't worry about it, the behavior in master is correct, now.

PR #148, opened 2015-02-07
add proxy API

minrk:
- GET fetches proxy table (duplicate of proxy's API endpoint, but allows scripts to only have one API key)
- POST prods routing table sync (useful when proxy restarts)
- PATCH allows updating the proxy API location, auth token

This should facilitate the CHP instance being managed outside the Hub. In master, running the proxy outside the Hub allows restarting the Hub independently of the Proxy, but not restarting the Proxy independently of the Hub. The main useful addition here is that a restarted proxy can `POST /hub/api/proxy` to cause the Hub to update the proxy's routing table. If, for some reason, the proxy has to start at a new URL, a PATCH request can be used to update that information.

For sanity checking, the Hub ensures the routing table makes sense on every last_activity update (five minutes).

A test is added that exercised restarting the Proxy while the Hub is running, both at the same URL via POST and at a new location via PATCH.

/cc @jhamrick

### DIFF
--- a/jupyterhub/apihandlers/__init__.py
+++ b/jupyterhub/apihandlers/__init__.py
@@ -1,9 +1,10 @@
 from .base import *
 from .auth import *
+from .proxy import *
 from .users import *
 
-from . import auth, users
+from . import auth, proxy, users
 
 default_handlers = []
-for mod in (auth, users):
+for mod in (auth, proxy, users):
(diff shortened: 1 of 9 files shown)


rgbkrk:
Oooh, nice.

minrk:
The test shows a relatively opaque example of starting the two separately. I'll try to put together a clearer one that's an actual example.

PR #151, opened 2015-02-23
DEV: Add support for logging to a file.

ssanderson:
Adds an `extra_log_file` configurable to `JupyterHub`, which, if set, creates a `FileHandler` and adds it to the application logger.

Also adds support for explicitly configuring more exotic loggers via `extra_log_handler`.

I originally started trying to make this change in the core `IPython` application class, but it looks like it's tricky to do correctly there because the logging machinery is set up much earlier in the lifetime of the object.  It's easier to do here because `init_logging` is called after all the rest of the configuration has happened.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -393,7 +393,8 @@ class JupyterHub(Application):
             datefmt=self.log_datefmt,
         )
         for handler in self.extra_log_handlers:
-            handler.setFormatter(_formatter)
+            if handler.formatter is None:
+                handler.setFormatter(_formatter)
             self.log.addHandler(handler)
 
         # hook up tornado 3's loggers to our app handlers
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -72,6 +72,7 @@ aliases = {
     'ip': 'JupyterHub.ip',
     'port': 'JupyterHub.port',
     'pid-file': 'JupyterHub.pid_file',
+    'log-file': 'JupyterHub.extra_log_file',
 }
 token_aliases = {}
 token_aliases.update(common_aliases)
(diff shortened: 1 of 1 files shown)


ssanderson:
I'm fairly confident the failure here isn't mine?

minrk:
Definitely not. There's a race condition in some of the returncode checks, which can be caused when subprocesses are slow to start.

jupyterhub/app.py
```diff
@@ -365,12 +366,33 @@ def _log_format_default(self):
         """override default log format to include time"""
         return "%(color)s[%(levelname)1.1s %(asctime)s.%(msecs).03d %(name)s %(module)s:%(lineno)d]%(end_color)s %(message)s"
 
+    extra_log_file = Unicode("", config=True)
+    extra_log_handler = Instance(
+        klass=logging.Handler,
+        allow_none=True,
+        config=True,
```
minrk: the handler doesn't need config=True, does it? Do you plan on setting the Handler itself via config?

jupyterhub/app.py
```diff
@@ -365,12 +366,33 @@ def _log_format_default(self):
         """override default log format to include time"""
         return "%(color)s[%(levelname)1.1s %(asctime)s.%(msecs).03d %(name)s %(module)s:%(lineno)d]%(end_color)s %(message)s"
 
+    extra_log_file = Unicode("", config=True)
```
minrk: can you add a `help="...` string? There should be one for every config=True trait, since it's what populates the `--help` output.

jupyterhub/app.py
ssanderson: I could imagine wanting to set, e.g., a `RotatingFileHandler` or `SysLogHandler`.  I figured the common case of "log to a file instead of stderr" was useful enough to support from the command line, and if people want a more exotic setup they can do it via config.

minrk: I'm trying to think if having the instance as a config value is more or less useful than calling:

```python
logging.getLogger('JupyterHub').add_handler(my_handler)
```

Especially since registering directly with the logger doesn't make it mutually exclusive with log-to-file. What do you think?

ssanderson: Would your proposed call to `getLogger` still go in the jupyterhub_config file?  The main downside I can see to that would be that it'd break if the application gets renamed (unlikely, but still feels brittle), and having it as a configurable makes it more self-documenting.

ssanderson: We could also remove the mutually-exclusive part by making it an `extra_log_handlers` list, and making `extra_log_file` just create a `FileHandler` and append it to the list.

ssanderson:
@minrk I replied to your comments on the configurable Handler, but they're stuck under an outdated diff from adding a help string.

minrk:
Makes sense. I think extra_log_handlers as a list is the best choice, then.

jupyterhub/app.py
```diff
@@ -365,12 +366,36 @@ def _log_format_default(self):
...
+            self.extra_log_handlers.append(
+                logging.FileHandler(self.extra_log_file)
+            )
+
+        _formatter = self._log_formatter_cls(
+            fmt=self.log_format,
+            datefmt=self.log_datefmt,
+        )
+        for handler in self.extra_log_handlers:
+            handler.setFormatter(_formatter)
```
minrk: should we always be setting the log formatter, or trusting that they are already configured with their own formatter?

ssanderson: Probably makes sense to make this only happen if `handler.formatter` is None.

ssanderson: Updated

minrk:
Thanks.

PR #160, opened 2015-03-02
make spawner IP configurable

minrk:
local spawners were hardcoded as 'localhost'

mainly to enable localhost->127.0.0.1 config in pathological cases


jupyterhub/spawner.py
```diff
@@ -41,6 +41,9 @@ class Spawner(LoggingConfigurable):
     user = Any()
     hub = Any()
     api_token = Unicode()
+    ip = Unicode('localhost', config=True,
+
```
minrk: need to add help string here

PR #171, opened 2015-03-05
DEV: Make the timeout used by wait_for_http_server configurable.

ssanderson:
On slow or heavily-populated hardware it can take more than 10 seconds for HTTP servers to come up after successful spawn.  This makes the timeout used in wait_for_http_server configurable.


minrk:
Makes sense, thanks.

PR #176, opened 2015-03-05
cleanup after SIGTERM

### LINKED ISSUE #174: `cleanup` doesn't fire when process is killed by SIGTERM
This is documented behavior for the `atexit` module, but since SIGTERM is meant to be used for graceful shutdown, I'd expect the hub to exit gracefully on receipt of a TERM.  I assume fixing this is just a matter of registering a SIGTERM handler, but I'm not sure if there are complications with the tornado eventloop.

minrk:
closes #174

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -8,6 +8,7 @@ import atexit
 import binascii
 import logging
 import os
+import signal
 import socket
 import sys
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


jupyterhub/app.py
```diff
@@ -864,12 +865,13 @@ def cleanup(self):
...
-                    self.proxy_process.terminate()
-                except Exception as e:
-                    self.log.error("Failed to terminate proxy process: %s", e)
+                if self.proxy_process.poll() is None:
+                    try:
+                        self.proxy_process.terminate()
+                    except Exception as e:
+                        self.log.error("Failed to terminate proxy process: %s", e)
             else:
                 self.log.info("I didn't start the proxy, I can't clean it up")
```
minrk: the changes here are because `I didn't start the proxy` log message was firing because SIGTERM shutdown the proxy as well, and proxy_process.poll() was returning the exit code.

ssanderson:
Sweet!  Sidenote: [Gunicorn](http://docs.gunicorn.org/en/latest/signals.html?highlight=signals) has some nice features for controlling a running server with signals.  The process incrementing and decrementing isn't especially useful in this context, but reloading configuration with SIGUSR would be pretty slick.

PR #182, opened 2015-03-07
Better Theme II

Carreau:
On top of my other PR, minimal UI improvement.

Some labels do not "wrap" on the next line on narrow screen anymore. `access server` was below `stop server`

Header is same size as notebook. 

capture d ecran 2015-03-06 a 17 45 09

### DIFF
--- a/share/jupyter/hub/static/less/page.less
+++ b/share/jupyter/hub/static/less/page.less
@@ -1,13 +1,32 @@
 .jpy-logo {
-  height: 40px;
-  margin: 8px;
+  height: 35px;
+  margin: 2px;
 }
 
-div#header {
-  border-bottom: 1px solid #ccc;
+#header {
+  border-bottom: 1px solid #e7e7e7;
+  height: 40px;
+
+
+}
+.navbar {
+  min-height:20px;
 }
 
 .hidden {
   display: none;
 }
 
+
+.dropdown.navbar-btn{
+    padding:0 5px 0 0;
+}
+
+#login_widget{
+
+    & .navbar-btn.btn-sm {
+      margin-top: 5px;
+      margin-bottom: 5px;
+    }
+
+}
--- a/share/jupyter/hub/static/less/page.less
+++ b/share/jupyter/hub/static/less/page.less
@@ -6,17 +6,11 @@
 #header {
   border-bottom: 1px solid #e7e7e7;
   height: 40px;
-
-
-}
-.navbar {
-  min-height:20px;
 }
 
 .hidden {
   display: none;
 }
-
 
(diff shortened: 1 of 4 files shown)


share/jupyter/hub/templates/page.html
```diff
@@ -88,9 +89,21 @@ <h4 class="modal-title" id="{{key}}-label">{{title}}</h4>
...
     <span id="login_widget">
       {% if user %}
-        <a id="logout" class="btn navbar-btn btn-default pull-right" href="{{logout_url}}">Logout</a>
+        <a id="logout" class="navbar-btn btn-sm btn btn-default pull-right" href="{{logout_url}}"> <i class="fa fa-sign-out"></i> Logout</a>
+        <div class="dropdown navbar-btn btn-sm pull-right">
+          <button class="btn-sm btn btn-default dropdown-toggle" type="button" id="dropdownMenu1" data-toggle="dropdown" aria-expanded="true">
+            <i class="fa fa-cogs"></i>
+            <span class="caret"></span>
+          </button>
+              <ul class="dropdown-menu" role="menu" >
```
minrk: This dropdown shouldn't exist if the user isn't admin. I'm not sure it should ever exist, actually.

minrk:
edit/delete have been crowded together now.

before:
screen shot 2015-03-06 at 18 37 06

after:

screen shot 2015-03-06 at 18 35 54

I think every difference there is better in the before than the after.

Carreau:
Edit/delete crushed together true also on master. 

capture d ecran 2015-03-06 a 19 02 08

Sorry for not seeing it.

I'll remove the dropdown if you dont' like it. I though it would be useful once we have more options later.

minrk:
>  I though it would be useful once we have more options later.

It may be, but I'd wait until there's more than one page. Plus, cogs indicate that it should bring up a settings panel.

minrk:
Want to rebase, now that #56 is merged?

share/jupyter/hub/static/less/page.less
```diff
@@ -1,12 +1,36 @@
...
-div#header {
-  border-bottom: 1px solid #ccc;
+#header {
+  border-bottom: 1px solid #e7e7e7;
+  height: 40px;
+
+
+}
+.navbar {
+  min-height:20px;
```
minrk: definitely don't want to set a global min-height on navbar. If navbar should change height, use variables.

Carreau:
Rebased.

Carreau:
(and remove this min-height for navbar thing)

minrk:
Perfect, thanks

PR #183, opened 2015-03-07
add docs on custom authenticators and spawners

### LINKED ISSUE #175: wiki as docs not really discoverable/navigable
I think it's fine for some things to stay there (e.g. getting sudo to work), but I'm going to move reference materials (e.g. writing a custom spawner) to markdown files in a docs dir.

minrk:
it makes sense to have some in-repo tree-navigable docs. The Wiki is appropriate for other things like indices of custom authenticators and spawners, but not the basics of how it works.

closes #175

### DIFF
--- a/docs/howitworks.md
+++ b/docs/howitworks.md
@@ -1,0 +1,75 @@
+# How JupyterHub works
+
+JupyterHub is a multi-user server that manages and proxies multiple instances of the single-user <del>IPython</del> Jupyter notebook server.
+
+There are three basic processes involved:
+
+- multi-user Hub (Python/Tornado)
+- configurable http proxy (nodejs)
+- multiple single-user IPython notebook servers (Python/IPython/Tornado)
+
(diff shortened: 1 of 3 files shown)


docs/howitworks.md
```diff
@@ -0,0 +1,75 @@
...
+The Spawner represents an abstract interface to a process,
+and needs to be able to take three actions:
+
+1. start the process
+2. poll whether the process is still running
+3. stop the process
+
+[More info on custom Spawners](spawners.md).
+
+[An example using Docker](https://github.com/jupyter/dockerspawner).
```
jhamrick: Maybe it would be nice to try to keep a list of custom spawners and authenticators? Perhaps this would be better suited to a wiki, but then we could at least link to it here.

minrk: Yup, the list [already exists](https://github.com/jupyter/jupyterhub/wiki/Spawners), I'll add links.

PR #196, opened 2015-03-17
DEV: Periodically clear single-user server cookie cache.

ssanderson:
Default is every 5 minutes.


ssanderson:
This is the easy half of https://github.com/jupyter/jupyterhub/issues/154.

minrk:
Thanks for the reminder, I did the other part a bit ago, but forgot to open the PR. Tornado already has builtin cookie expiry, so all we need to do is enable it.

PR #197, opened 2015-03-17
expire login cookies

### LINKED ISSUE #154: Automatically expire JupyterHub and single-user server cookies at configurable intervals
Right now, once a user has logged in to their single-user server, they're considered validated forever unless they log out.  In particular, even if their hub token is revoked, the single-user server never learns about this fact because it happily accepts their existing single-user token.

I think the obvious fix for this is to set PeriodicCallbacks that purge token caches.  It might also make sense to set TTL values for the tornado cookies themselves.

minrk:
via tornado's cookie max_age_days mechanism

default expiry is two weeks

closes #154

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -37,7 +37,7 @@ if V(IPython.__version__) < V('3.0'):
 
 from IPython.utils.traitlets import (
     Unicode, Integer, Dict, TraitError, List, Bool, Any,
-    Type, Set, Instance, Bytes,
+    Type, Set, Instance, Bytes, Float,
 )
 from IPython.config import Application, catch_config_error
 
@@ -183,6 +183,11 @@ class JupyterHub(Application):
     pid_file = Unicode('', config=True,
         help="""File to write PID
         Useful for daemonizing jupyterhub.
+        """
+    )
+    cookie_max_age_days = Float(14, config=True,
+        help="""Number of days for a login cookie to be valid.
+        Default is two weeks.
         """
     )
     last_activity_interval = Integer(300, config=True,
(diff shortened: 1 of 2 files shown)


minrk:
@ssanderson I think this should be sufficient for cookie expiry, right? Is there anything more that's needed? The units are days, but it accepts a float, so you can give it an expiry of an hour or a few minutes, if you want.

ssanderson:
@minrk, just to make sure I understand this correctly, this effectively puts a timeout on the *encrypted* version of the cookie that's provided to the user's browser, but the unencrypted token stored in the local sqlite database never changes?  I think eventually it'd be safer to expire the unencrypted tokens as well, but this seems like a reasonable (and probably sufficient for most use cases) first step.

minrk:
That's exactly right. Tornado's encrypted cookies always include a timestamp, so `get_secure_cookie` returns None if `now() - timestamp > max_age`.

I also think we should add expiry to the tokens themselves, but that requires adding database information, which is more work, and not necessary for basic login expiry.

minrk:
I also especially want to try adding the db token expiry, because it should be a relatively easy addition, and thus a good starting point for adding alembic integration.

ssanderson:
Gotcha.  I think this is reasonable as a standalone change.  I'm also happy to help with the alembic stuff.  It's not especially designed with public projects in mind, and I ended up having to add some bootleg-ish templating logic to make it work in pgcontents.

sangramga:
@minrk So in Jupyterhub version 0.8.1. The cookie expires after max_age_days but not the Authorization token. Am I right?
Or there has been a pull request for Auth token expiry also?

minrk:
Tokens still do not support expiry.

PR #200, opened 2015-03-18
DEV: Allow setting a whitelist_group on LocalAuthenticator.

ssanderson:
Any user in the group is considered in the whitelist.

One issue that arises as a result of this PR is that it seems reasonable to have an empty whitelist of users if you're specifying a whitelist group, but that's currently not possible because the hub requires you to have an admin user and places them into the whitelist.

### DIFF
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -3,6 +3,7 @@
 # Copyright (c) IPython Development Team.
 # Distributed under the terms of the Modified BSD License.
 
+from grp import getgrnam
 import pwd
 from subprocess import check_call, check_output, CalledProcessError
 
@@ -39,7 +40,14 @@ class Authenticator(LoggingConfigurable):
         It must return the username on successful authentication,
         and return None on failed authentication.
         """
-    
+
+    def check_whitelist(self, user):
+        """
+        Return True if the whitelist is empty or user is in the whitelist.
+        """
+        # Parens aren't necessary here, but they make this easier to parse.
+        return (not self.whitelist) or (user in self.whitelist)
+
     def add_user(self, user):
         """Add a new user
         
@@ -56,8 +64,7 @@ class Authenticator(LoggingConfigurable):
         
         Removes the user from the whitelist.
         """
-        if user.name in self.whitelist:
-            self.whitelist.remove(user.name)
+        self.whitelist.discard(user.name)
     
     def login_url(self, base_url):
         """Override to register a custom login handler"""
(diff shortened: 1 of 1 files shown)


ssanderson:
@minrk @YannBrrd I think this gets at most of https://github.com/jupyter/jupyterhub/issues/198

minrk:
Empty whitelist is meant to be allowed, indicating that any user that successfully authenticates is permitted. Adding admin users to the whitelist should only happen if it is already non-empty. Can you verify that it's actually populating the whitelist?

jupyterhub/auth.py
```diff
@@ -87,7 +94,28 @@ class LocalAuthenticator(Authenticator):
...
-    
+
+    whitelist_group = Unicode(
+        config=True,
+        help="Automatically whitelist anyone in this group.",
+    )
+
+    def check_whitelist(self, username):
+        return (
+            super().check_whitelist(username) or
```
minrk: In the case where the user whitelist is empty, `check_whitelist_group` will never be called.

For simplicity, I might require whitelisting by name or group to be mutually exclusive. What do you think?

ssanderson:
@minrk sorry, I misunderstood what was happening in my initial test. I originally tried to test this by setting the whitelist to be empty, but, as you point out, that signifies "anyone is allowed".  I then tried to test it by setting the whitelist to be `{"root"}`, but then my user gets added as an admin and consequently added to the whitelist.

Ignoring the slightly confusion above, it seems like there should be a way of saying "don't allow any users by username: only allow them by group".  That's not currently possible because the empty set is interpreted as "allow anyone".  My inclination is to say that `None` should be used as "allow anyone", and  empty set should signify "no one is whitelisted".

jupyterhub/auth.py
ssanderson: > In the case where the user whitelist is empty, check_whitelist_group will never be called.

This is the behavior I intended, though that doesn't necessarily mean it's correct.

I'm trying to think if there's ever a case where it's useful to be able to say "allow people in this group, plus these random other users"...I could imagine a world where you had something like an "admin" group on a server, and you wanted to allow access to all the admins, plus a few exceptions who shouldn't have admin permissions, but should still be able to access the hub.

minrk:
@ssanderson if the group whitelist and name whitelist were mutually exclusive (I think they should be), then there shouldn't be a problem. You would *either* check by group *or* check by name.

minrk:
Something like:

```python
if self.group_whitelist:
    return self.check_group_whitelist()
else:
    return self.check_whitelist()
```

jupyterhub/auth.py
minrk: If the group whitelist is a set as well, I don't think there's anything you can do with names that you can't do with groups. Of course, a special-purpose 'jupyterhub' group would also cover it.

ssanderson:
@minrk I updated this to make group and user whitelists mutually-exclusive.  Not sure what the deal is with the Python 3.3 failure on Travis.

jupyterhub/auth.py
```diff
@@ -87,7 +94,34 @@ class LocalAuthenticator(Authenticator):
...
+        if self.group_whitelist:
+            return self.check_group_whitelist(username)
+        else:
+            return super().check_whitelist(username)
+
+    def check_group_whitelist(self, username):
+        if not self.group_whitelist:
+            return False
+        try:
+            group = getgrnam(self.group_whitelist)
```
minrk: If self.group_whitelist is a set now, doesn't this need to be a loop?

minrk:
Don't worry about the test failure. There's a sporadic race condition that seems to cause it to fail once in a while. I haven't been able to track it down.

jupyterhub/auth.py
ssanderson: Oops.  Yep. Will fix.

minrk:
Thanks

Issue #133, opened 2015-01-28
terminado seems to ignore shell


dsblank:
Moving item from terminado to here:

https://github.com/takluyver/terminado/issues/2

See issue for details.

minrk:
Thanks

dsblank:
Did you mean to close? If so, what is resolution... user error? :) Could well be.

minrk:
Arg. Read the description backwards, I thought you were moving from here to there.

minrk:
@dsblank you are using the sudospawner, right?

dsblank:
Yes:
```
jupyterhub --JupyterHub.spawner=sudospawner.SudoSpawner --JupyterHub.ip=165.106.10.83 --JupyterHub.port=80  --JupyterHub.data_files_path='/home/dblank/jupyter.brynmawr'
```

jdavidheiser:
I made a simple change to spawner.py that seems to fix this, and pass on the SHELL environment variable.  Adding an export SHELL=bash before launching jupyterhub now works to give us bash in the terminado shells launched by our jupyter sessions.

minrk:
closed by #204

PR #204, opened 2015-03-20
spawner.py update for default shell

jdavidheiser:
Update spawner.py to do a better job keeping the shell variable intact for terminals launched from within the notebook.

Exporting the SHELL environment variable when launching jupyterhub should now use that shell for terminado terminals.  I set this in my /etc/init.d/jupyter script.  This should fix issue #133

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -81,7 +81,6 @@ class Spawner(LoggingConfigurable):
         'VIRTUAL_ENV',
         'LANG',
         'LC_ALL',
-        'SHELL'
     ], config=True,
         help="Whitelist of environment variables for the subprocess to inherit"
     )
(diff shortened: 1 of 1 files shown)


jupyterhub/spawner.py
```diff
@@ -81,6 +81,7 @@ class Spawner(LoggingConfigurable):
         'VIRTUAL_ENV',
         'LANG',
         'LC_ALL',
+        'SHELL'
```
minrk: The SHELL of the server terminal (often sh) user shouldn't set the SHELL of the single-user servers.

minrk:
I think the right fix is to set the shell for the user with `pwd.getpwnam(User.name).pw_shell` in LocalProcessSpawner.

jdavidheiser:
I was mostly trying to address the issue brought up in #133, by supporting the "set shell variable then launch jupyterhub" use, but I agree that using the user-specific shell is much cleaner.  I'll try to test the change later this afternoon and submit an updated PR.

jdavidheiser:
the update as suggested worked as expected, and user specified shells are respected.

minrk:
Thanks, merged.

PR #205, opened 2015-03-22
Update page flow based on dev meeting

minrk:
From the [dev meeting notes](https://ipython.hackpad.com/Cf6eyqe5Yqe#:h=JupyterHub),
some changes to the page flow:

- [x] remove logout page, redirect to / on logout
- [x] / should show login form, if there is a form
- [x] / should show login *link* otherwise
- [x] successful login should send user to their own server, if running
- [x] successful login should start their server and redirect, if not? (currently sends to control-panel - this may be better if the server is failing to start, otherwise the user can get stuck).
- [x] link to control panel in single-user header
- <del>move '/home/' to '/controlpanel' (I haven't moved the user page. When more Hub-specific content is added, I'm not sure if we will add a separate `/home` and `/controlpanel` for users. I'm inclined to leave it as `/home` for now.)</del>
- [x] Add `Control Panel` link to single-user header
- [x] jupyter logo on at least the landing page, but maybe not the control/admin pages.

Custom Authenticators that redirect to other services must set `Authenticator.login_service` to a non-empty string in order to trigger the `'Log in with [Service]'` button, instead of the login form.

The control-panel link in the header is accomplished through some monkeypatching instead of a custom template. The logic is the same as a custom `page.html` template, but doing it this way keeps the single-user server as a single standalone script, rather than needing to add a `page.html` template somewhere on the single-user filesystem. I might be able to simplify this a little bit.

/cc @ellisonbg @jhamrick


jupyterhub/singleuser.py
```diff
@@ -134,6 +134,45 @@ def _confirm_exit(self):
         # disable the exit confirmation for background notebook processes
         ioloop.IOLoop.instance().stop()
     
+    def initialize(self, argv=None):
```
ellisonbg: I think we eventually want to define a simple extension API that can pass jupyterhub template content to the single user notebook template so this monkey patch isn't required.

ellisonbg:
One issue:

* Go to the notebook dashboard.
* Click on the "Control Panel" button.
* Click on the "My Server" button.

Each time you do this you get another "Control Panel" button on the single user header:

screen shot 2015-03-30 at 9 01 50 pm

ellisonbg:
But the basic flow is much better and works as expected.

ellisonbg:
* I am fine using the jupyter logo on all pages. I don't think users will understand the two different parts of the web app and why they have different logos and names.
* I like the idea of renaming `home` to `controlpanel` but I am fine leaving for now for the reasons you mention.
* Not sure what the right initial page is. I think we will have fewer problems if we only go to the notebook dashboard if the 1p server is already running. If the 1p server is not running I think the controlpanel is best.

I haven't tested this with a non-form based login.

Very nice work!

minrk:
I fixed the template loading so it uses a regular template from a string, without any weird method monkeypatching. It depends on ipython/ipython#8215 for that, though. I could make it work in 3.0 by overriding the login_widget block, but that doesn't seem right.

ellisonbg:
OK once #8215 is merged I will retest this.

On Tue, Mar 31, 2015 at 2:29 PM, Min RK <notifications@github.com> wrote:

> I fixed the template loading so it uses a regular template from a string,
> without any weird method monkeypatching. It depends on
> ipython/ipython#8215 <https://github.com/ipython/ipython/pull/8215> for
> that, though. I could make it work in 3.0 by overriding the login_widget
> block, but that doesn't seem right.
>
> —
> <https://github.com/jupyter/jupyterhub/pull/205#issuecomment-88255922>.
>



-- 
Brian E. Granger
Cal Poly State University, San Luis Obispo
@ellisonbg on Twitter and GitHub
bgranger@calpoly.edu and ellisonbg@gmail.com

minrk:
Okay, I think this should be good to go, now.

PR #207, opened 2015-03-23
use `cookie_name/cookie_value` URL for cookie checking API

minrk:
instead of putting the value in the request body, which is verboten for GET requests

/cc @jdfreder

### DIFF
--- a/jupyterhub/singleuser.py
+++ b/jupyterhub/singleuser.py
@@ -48,10 +48,9 @@ class JupyterHubLoginHandler(LoginHandler):
         hub_api_url = self.settings['hub_api_url']
         hub_api_key = self.settings['hub_api_key']
         r = requests.get(url_path_join(
-            hub_api_url, "authorizations/cookie", cookie_name,
+            hub_api_url, "authorizations/cookie", cookie_name, encrypted_cookie,
         ),
             headers = {'Authorization' : 'token %s' % hub_api_key},
-            data=encrypted_cookie,
         )
         if r.status_code == 404:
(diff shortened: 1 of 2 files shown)


jdfreder:
Thanks @minrk , this looks good to me.  @jhamrick , should I target this new API in the HubAuth PR for nbgrader?

minrk:
I deprecated the body version, just in case.

jupyterhub/singleuser.py
```diff
@@ -48,10 +48,9 @@ def verify_token(self, cookie_name, encrypted_cookie):
         hub_api_url = self.settings['hub_api_url']
         hub_api_key = self.settings['hub_api_key']
         r = requests.get(url_path_join(
-            hub_api_url, "authorizations/cookie", cookie_name,
```
ssanderson: Is the text of encrypted_cookie guaranteed to be url-safe and/or does requests automatically escape it intelligently?

minrk: it's definitely not url-safe, requests does the escaping.

minrk:
This follows the pattern of GitHub, though their tokens aren't cookies, and are url-safe. The other fix is to leave the behavior, and make it a POST request instead of GET, with the cookie value still in the body.

jupyterhub/singleuser.py
minrk: I decided to escape it explicitly. I'm not sure exactly how requests decides what escaping to do, but I'd rather be safe.

PR #209, opened 2015-03-23
reply with full user model in auth handlers

minrk:
Mainly this adds the `admin` key,
but I don't see a reason to respond with different subsets of the user model in different circumstances.

based on #207

### DIFF
--- a/jupyterhub/apihandlers/auth.py
+++ b/jupyterhub/apihandlers/auth.py
@@ -17,9 +17,7 @@ class TokenAPIHandler(APIHandler):
         orm_token = orm.APIToken.find(self.db, token)
         if orm_token is None:
             raise web.HTTPError(404)
-        self.write(json.dumps({
-            'user' : orm_token.user.name,
-        }))
+        self.write(json.dumps(self.user_model(orm_token.user)))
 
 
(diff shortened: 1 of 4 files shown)


jupyterhub/apihandlers/auth.py
```diff
@@ -17,9 +17,7 @@ def get(self, token):
         orm_token = orm.APIToken.find(self.db, token)
         if orm_token is None:
             raise web.HTTPError(404)
-        self.write(json.dumps({
-            'user' : orm_token.user.name,
-        }))
+        self.write(json.dumps(self.user_model(orm_token.user)))
```
cemoody: fyi,
Hey @minrk, my jupyterhub multi-user build started breaking on this PR. Logging in and PAM auth works, but I get 500 internal server errors. Turning on debug options indicates that there's a json parsing error -- my guess it's the fault of this line or somewhere near it?

Going back to 3fd80f9f3a87f000171b91b78a5ed4da015b86af fixes this problem, so it's isolated to this PR.

minrk: Can you share the error message? Probably a simple fix.

PR #211, opened 2015-03-24
Make the hub bind to address specified in hub_ip

peterruppel:
Bug: the hub binds to all interfaces and ignores the address given in 'hub_ip'. The 'address' parameter was missing in the call http_server.listen().


minrk:
Thanks!

PR #220, opened 2015-04-07
add some test coverage

minrk:
- fix bug in group-whitelist (it didn't work at all)
- exercise basic HTML pages
- various testing changes, fixes found along the way

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,4 @@
+[run]
+omit =
+    jupyterhub/tests/*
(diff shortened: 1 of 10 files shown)


ssanderson:
> fix bug in group-whitelist (it didn't work at all)
:/ my bad...

minrk:
@ssanderson no problem, this is why we have (or should have had) tests. It was just one or two more replacements from when we decided to switch from a string to a set of strings that were missed.

PR #226, opened 2015-04-08
don't update last_activity on shutdown

### LINKED ISSUE #224: Admin Panel - Last seen column includes shutdown activity
The "Last Seen" column includes shutdown activity. I'm assuming this is an artifact of the proxy itself, but is a bit misleading. After shutting down several servers from users that had not been back in days, I noticed they all came up as visiting recently.

minrk:
closes #224


rgbkrk:
This makes sense.

PR #230, opened 2015-04-14
base64 encode username in cookie_name to accommodate special chars in em...

ksolan:
Fixes https://github.com/jupyter/jupyterhub/issues/218. 

Could *potentially* cause upgrade problems if existing usernames collide with base-64 representations of new usernames. E.g., if a user named "c2FtcGxldXNlcg" already exists in the database with cookie name "jupyter-hub-token-c2FtcGxldXNlcg", and a new user named "sampleuser" comes along, (base-64 encoded as "c2FtcGxldXNlcg"), the cookie names would collide. Maybe we should add a prefix (e.g., "jupyter-hub-token-b64-"), but there's no way to ensure that it won't collide with existing usernames in some case.

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -4,7 +4,6 @@
 # Distributed under the terms of the Modified BSD License.
 
 import re
-import base64
 from datetime import datetime, timedelta
 from http.client import responses
 
@@ -142,10 +141,7 @@ class BaseHandler(RequestHandler):
         """Get ORM User for username"""
         user = self.find_user(username)
         if user is None:
-            user = orm.User(
-                name=username, 
-                b64_name=base64.b64encode(bytes(username, "utf8")).decode("utf8").strip("=")
-            )
+            user = orm.User(name=username)
             self.db.add(user)
             self.db.commit()
         return user
@@ -400,7 +396,6 @@ class UserSpawnHandler(BaseHandler):
             self.set_login_cookie(current_user)
             without_prefix = self.request.path[len(self.hub.server.base_url):]
             target = url_path_join(self.base_url, without_prefix)
-            self.log.warn(target)
             self.redirect(target)
         else:
             # not logged in to the right user,
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -4,6 +4,7 @@
 # Distributed under the terms of the Modified BSD License.
 
 import re
+import base64
 from datetime import datetime, timedelta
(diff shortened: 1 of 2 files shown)


minrk:
Even if it collides with crazy usernames, that's not a huge problem. Changing the cookie name does mean that upgrading across this change requires rebuilding every single-user server, which isn't typically necessary on a relaunch of JupyterHub. cc @jhamrick, @rgbkrk for thoughts on changes that require single-user servers to relaunch?

rgbkrk:
That's a pretty far off collision. Seems fine on that aspect.

What does this mean for logged in users? Do they just have to re login?

minrk:
I'd need to check a few things to be sure. It would probably be prudent to reset the cookie secret so that all users are forced to re-login, so that the cookies don't get confused. It *should* set the new cookie when launching the server.

ksolan:
There are issues with '@' in urls as well...though it's a legal character (http://stackoverflow.com/questions/1547899/which-characters-make-a-url-invalid), some clients/browsers escape it as '%40' in requests. The proxy, however, only recognizes the route with the '@'. 

As a quick fix, I switched to base-64 usernames in urls as well. I added a 'b64_name' column to the User model for anywhere else that doesn't accept special chars (for example, I know Docker doesn't accept special chars in container names). 

Needless to say this is a breaking change. It works well for my use case, but I'm not sure if/how it should be incorporated into the main project.

minrk:
We definitely won't b64 the username in the URLs. There may be a fix to make either here or in CHP or both to properly handle the routing of names that need escaping. Normally, unicode URLs are not a problem, so it's either a problem in the proxy not properly handling the escape, or in the proxy failing to escape the URL when creating the route. Need to think carefully about that one, but b64-encoding the name is definitely not right.

jhamrick:
> thoughts on changes that require single-user servers to relaunch?

Fine by me. I probably won't be updating the version of JupyterHub again at this point in the semester since there's only a few weeks left of classes.

ksolan:
Alright, I'm gonna close this pull request and look into the URL escaping issue more carefully. I'm thinking it probably needs to be addressed in CHP. The only way I could come up with to fix it in JupyterHub is to register two different routes (one with '@' and one with '%40'), which isn't a good solution. There's likely a cleaner way to fix it in the proxy directly.

minrk:
I started working on it in CHP, but didn't have time to think carefully through the various cases of requests from both directions and JSON content vs URLs. I agree that it probably needs at least some change in both CHP and JupyterHub.

PR #231, opened 2015-04-16
Make cookie secure if used over https

Carreau:


jupyterhub/handlers/base.py
```diff
@@ -165,7 +166,9 @@ def set_hub_cookie(self, user):
         self.set_secure_cookie(
             self.hub.server.cookie_name,
             user.cookie_id,
-            path=self.hub.server.base_url)
+            path=self.hub.server.base_url,
+            secure=(self.request.protocol == 'https')
```
ssanderson: What happens if this is set to true when running over unencrypted HTTP?

minrk: It effectively negates setting the cookie at all. The browser will receive the cookie, see that it should only be sent over https, and then refuse to send it back to the server on any future request.

minrk:
apparently setting `secure=False` effectively sets `secure=True` on tornado < 4.2 (4.1 is latest stable). So need to do a `**kwargs` approach:

```python
kwargs = {}
if self.request.protocol == 'https':
    kwargs['secure'] = True
self.set_secure_cookie(
    ...,
    **kwargs
)
```

PR #236, opened 2015-04-20
DEV: Python2 compat in singleuser.py

ssanderson:
Adds Python 2 backwards-compat for singleuser.py only.

Fixes https://github.com/jupyter/jupyterhub/issues/235.

### DIFF
--- a/jupyterhub/singleuser.py
+++ b/jupyterhub/singleuser.py
@@ -5,7 +5,11 @@
 # Distributed under the terms of the Modified BSD License.
 
 import os
-from urllib.parse import quote
+try:
+    from urllib.parse import quote
+except ImportError:
+    # PY2 Compat
+    from urllib import quote
 
 import requests
 from jinja2 import ChoiceLoader, FunctionLoader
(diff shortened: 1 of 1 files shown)


jupyterhub/singleuser.py
```diff
@@ -177,7 +181,7 @@ def start(self):
                 self._clear_cookie_cache,
                 self.cookie_cache_lifetime * 1e3,
             ).start()
-        super().start()
+        super(SingleUserNotebookApp, self).start()
```
minrk: Can you add a py2 compat comment here, so we can more easily revert it when Python 2 support is dropped?

ssanderson: There are multiple calls to old-style super already in this file.  Do you want to mark all of them?  When the time comes to officially drop support, it'll probably be easier to just grep for `super(` than it will be to grep for PY2 comments.

minrk: Fair point. Leave it alone, then.

minrk:
I'm okay with this for now, though I'm probably going to stick with Python 2 support for single-user servers as undocumented and not officially supported, and strongly recommended against.

If the single-user starts to require more than this single file, or imports elsewhere in JupyterHub are needed, Python 2 support for single-user servers will likely be dropped.

PR #240, opened 2015-04-23
DEV: Allow configuration of default headers.

ssanderson:
Applies Content-Security-Policy: frame-ancestors 'self' by default.

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -71,6 +71,17 @@ class BaseHandler(RequestHandler):
         self.db.rollback()
         super(BaseHandler, self).finish(*args, **kwargs)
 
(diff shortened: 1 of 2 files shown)


ssanderson:
ping @minrk @rgbkrk.  This is mostly copied from the jupyter_notebook implementation of the same functionality.

minrk:
Makes sense. Looks like the test is failing due to the semicolon?

ssanderson:
> Looks like the test is failing due to the semicolon?

Gah.  Fixed.

rgbkrk:
Thanks, this is certainly good to have.

Would it make sense to expose a URI for logging CSP violations, as is done with the handler in the notebook server? CSP violations should be made known to an admin if they want to know.

PR #241, opened 2015-04-24
Get url_path_join from jupyter_notebook

toobaz:
Since the commit https://github.com/ipython/ipython/commit/e65db1aa9317b21d8c9e8a51ff04f24cda96e843 url_path_join must be retrieved from jupyter_notebook (right?)


minrk:
Since that package hasn't settled yet, I'm going to hold off a bit on breaking JupyterHub with current IPython.

toobaz:
... that's the purpose of the "except"! Am I missing anything?!

minrk:
No, that's fine. I must have misread it before.

PR #250, opened 2015-05-06
create multiple users in admin panel

### LINKED ISSUE #225: Admin panel - Add ability to import multiple users at once
It would be nice to add a list of users as a bulk list of newline delimited users.

minrk:
usernames separated by lines

adds POST /users endpoint for multi-user creation with a single request, so we don't need to make N API requests.

closes #225

### DIFF
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -30,13 +30,19 @@ class UserListAPIHandler(APIHandler):
         admin = data.get('admin', False)
         self._check_user_model(data)
         
+        to_create = []
         for name in usernames:
             user = self.find_user(name)
             if user is not None:
-                raise web.HTTPError(400, "User %s already exists" % name)
+                self.log.warn("User %s already exists" % name)
+            else:
+                to_create.append(name)
+        
+        if not to_create:
+            raise web.HTTPError(400, "All %i users already exist" % len(usernames))
         
         created = []
-        for name in usernames:
+        for name in to_create:
             user = self.user_from_username(name)
             if admin:
                 user.admin = True
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -27,8 +27,10 @@ class UserListAPIHandler(APIHandler):
             raise web.HTTPError(400, "Must specify at least one user to create")
         
         usernames = data.pop('usernames')
-        admin = data.get('admin', False)
         self._check_user_model(data)
+        # admin is set for all users
+        # to create admin and non-admin users requires at least two API requests
+        admin = data.get('admin', False)
         
         to_create = []
         for name in usernames:
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -18,6 +18,43 @@ class UserListAPIHandler(APIHandler):
         users = self.db.query(orm.User)
         data = [ self.user_model(u) for u in users ]
         self.write(json.dumps(data))
+    
+    @admin_only
+    @gen.coroutine
+    def post(self):
+        data = self.get_json_body()
+        print(data)
+        if not data or not isinstance(data, dict) or not data.get('usernames'):
+            raise web.HTTPError(400, "Must specify at least one user to create")
+        
+        usernames = data.pop('usernames')
+        self._check_user_model(data)
+        
+        for name in usernames:
+            user = self.find_user(name)
+            if user is not None:
+                raise web.HTTPError(400, "User %s already exists" % name)
+        
+        created = []
+        for name in usernames:
+            user = self.user_from_username(name)
+            if data:
+                self._check_user_model(data)
+                if 'admin' in data:
+                    user.admin = data['admin']
+                    self.db.commit()
+            try:
+                yield gen.maybe_future(self.authenticator.add_user(user))
+            except Exception:
+                self.log.error("Failed to create user: %s" % name, exc_info=True)
+                self.db.delete(user)
+                self.db.commit()
+                raise web.HTTPError(400, "Failed to create user: %s" % name)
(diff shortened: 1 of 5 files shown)


jupyterhub/apihandlers/users.py
```diff
@@ -18,6 +18,43 @@ def get(self):
         users = self.db.query(orm.User)
         data = [ self.user_model(u) for u in users ]
         self.write(json.dumps(data))
+    
+    @admin_only
+    @gen.coroutine
+    def post(self):
+        data = self.get_json_body()
+        print(data)
```
fperez: was this meant to be for debugging temporarily? I'm going to guess real messages go to logging?

jupyterhub/apihandlers/users.py
```diff
@@ -18,6 +18,43 @@ def get(self):
...
+        if not data or not isinstance(data, dict) or not data.get('usernames'):
+            raise web.HTTPError(400, "Must specify at least one user to create")
+        
+        usernames = data.pop('usernames')
+        self._check_user_model(data)
+        
+        for name in usernames:
+            user = self.find_user(name)
+            if user is not None:
```
fperez: As an alternative, it could collect all already existing users and provide a warning at the end about those users not having been re-created... Just a thought.

jupyterhub/apihandlers/users.py
```diff
@@ -18,6 +18,43 @@ def get(self):
...
+                raise web.HTTPError(400, "User %s already exists" % name)
+        
+        created = []
+        for name in usernames:
+            user = self.user_from_username(name)
+            if data:
```
fperez: Why do the `if data` check on every pass of the loop, if `data` doesn't change?

minrk: leftover from copying. It doesn't make sense here.

jupyterhub/apihandlers/users.py
minrk: I thought about that, and I figured failing early for a bad request, rather than partial success made a little more sense.

fperez: well, the use case I had in mind was someone maintaining a plaintext list of their users (say a class). It's a lot easier to copy/paste the whole file each time you want to make sure everyone has an account and you may have added users, than to manually track your new people in a separate list. For example if you made the list from an alphabetically sorted spreadsheet, having to manually prune it of existing names would be more annoying...

jupyterhub/apihandlers/users.py
```diff
@@ -18,6 +18,41 @@ def get(self):
...
+        
+        for name in usernames:
+            user = self.find_user(name)
+            if user is not None:
+                raise web.HTTPError(400, "User %s already exists" % name)
+        
+        created = []
+        for name in usernames:
+            user = self.user_from_username(name)
+            if admin:
```
fperez: Mmh, admin was pulled globally *once* above, but now it will be applied to all users? Either I'm misunderstanding the logic, or it's wrong...

minrk: admin is a single field. You cannot bulk-create admin and non-admin users.

jupyterhub/apihandlers/users.py
minrk: That makes sense. I'll just prune and warn, then, instead of failing.

minrk: done. Adding a user to a list should result in just adding that user and skipping over those that already exist. A request that creates no users will still be considered a failure.

fperez:
BTW, Travis is complaining. Is it meaningful?

jupyterhub/apihandlers/users.py
```diff
@@ -18,6 +18,47 @@ def get(self):
...
+    
+    @admin_only
+    @gen.coroutine
+    def post(self):
+        data = self.get_json_body()
+        if not data or not isinstance(data, dict) or not data.get('usernames'):
+            raise web.HTTPError(400, "Must specify at least one user to create")
+        
+        usernames = data.pop('usernames')
+        admin = data.get('admin', False)
```
fperez: Ah, if admin is global, I'd add here your comment in the code itself (the one about no bulk creation of mixed admin/non-admin users).  Could be useful for anyone reading the code.

jupyterhub/apihandlers/users.py
```diff
@@ -18,6 +18,47 @@ def get(self):
...
+        
+        if not to_create:
+            raise web.HTTPError(400, "All %i users already exist" % len(usernames))
+        
+        created = []
+        for name in to_create:
+            user = self.user_from_username(name)
+            if admin:
+                user.admin = True
+                self.db.commit()
```
fperez: Just for my own curiosity, is there an ORM such that modfying the `user` object and making a commit operation on `self.db` still propagates the `admin` flag into the db?

minrk: I'm not sure I understand the question. Modifying the object and sending the changes to the database are two separate actions. You don't want every change to hit the database immediately because that can get super slow.

fperez: I was just confused that the `.commit()` call was made without arguments, so I wasn't sure when the change to `user.admin` would propagate, that's all.

fperez:
Otherwise looks good, and Travis is happy now. Merging, thanks!!

jupyterhub/apihandlers/users.py
minrk: Ah, yes. The User object is attached to the session, so any changes get added to the pending transaction. `db.commit()` applies those changes to the database.

rgbkrk:
:smile:

jupyterhub/apihandlers/users.py
fperez: Got it, thanks!

PR #255, opened 2015-05-28
Include query arguments in GET requests to the notebook server (once it spawns)

TimShawver:
The problem: When a GET request includes query arguments, and also triggers a notebook server spawn, the query arguments don't get passed through to the notebook server once it's spawned.

Using self.request.uri instead of self.request.path in a few places fixes this problem.


TimShawver:
@minrk not sure why the build seg-faulted.  Could you run it again? @ssanderson

minrk:
@TimShawver it's a problem in the test runner. I poked Travis and it's all set now.

PR #258, opened 2015-06-10
For a standard installation, 'pip install' has to be run with sudo

nthiery:
I needed to do this to install jupyterhub on my machine; that was the correct thing to do, right?

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,11 +64,11 @@ As usual start with cloning the code:
 Then you can install the Python package by doing:
 
     pip install -r requirements.txt
-    pip install .
+    sudo pip install .
 
 If you plan to run notebook servers locally, you may also need to install the IPython notebook:
 
-    pip install "ipython[notebook]"
+    sudo pip install "ipython[notebook]"
 
 
 This will fetch client-side javascript dependencies and compile CSS,


minrk:
That's not generally true. For virtualenvs, anaconda/conda, or OS X with brew, that isn't the case. I would prefer to leave sudo off, since it's a lot worse to sudo when you don't have to than leave it off and have to run the command again with sudo.

nthiery:
On Wed, Jun 10, 2015 at 10:08:02AM -0700, Min RK wrote:
>    That's not generally true. For virtualenvs, anaconda/conda, or OS X
>    with brew, that isn't the case. I would prefer to leave sudo off, since
>    it's a lot worse to sudo when you don't have to than leave it off and
>    have to run the command again with sudo.

Ok. Then, the sudo should probably be removed from the pip calls
earlier in the README for consistency?

And maybe adding a note could be added specifying that, depending on
the context, jupyterhub may need to be installed as root?

Cheers,
				Nicolas
--
Nicolas M. Thiéry "Isil" <nthiery@users.sf.net>
http://Nicolas.Thiery.name/

minrk:
@nthiery yup, those were added in a recent PR, and I missed them. Actually both of those entire install lines for jsonschema or terminado should be removed entirely.

I would do that - remove sudo from any copy/paste-able examples, and add a note that

> sudo may be needed for `pip install`, depending on filesystem permissions.

nthiery:
On Wed, Jun 10, 2015 at 03:10:30PM -0700, Min RK wrote:
>    [1]@nthiery yup, those were added in a recent PR, and I missed them.
>    Actually both of those entire install lines for jsonschema or terminado
>    should be removed entirely.
> 
>    I would do that - remove sudo from any copy/paste-able examples, and
>    add a note that
> 
>      sudo may be needed for pip install, depending on filesystem
>      permissions.

Sounds good to me. Do you want me to do it, or are you just going for it?

Cheers,
				Nicolas
--
Nicolas M. Thiéry "Isil" <nthiery@users.sf.net>
http://Nicolas.Thiery.name/

minrk:
If you want to make a PR, that would be great!

PR #259, opened 2015-06-10
Removed sudo from pip install instructions, ...

nthiery:
Attempt at implementing @minrk's suggestion

https://github.com/jupyter/jupyterhub/pull/258#issuecomment-110932499


minrk:
Thanks

PR #261, opened 2015-06-22
remove dependency on IPython

minrk:
- Standalone traitlets has been released, use it directly.
- Copy url_path_join from notebook


rgbkrk:
Nice!

PR #265, opened 2015-06-25
Add some docs clarifications

jhamrick:
1. If `lessc` isn't installed, it should be installed by running `npm install`. This is apparently done in the travis script and the dockerfile, but isn't actually in the installation docs (one thing I'm not sure on: should this be documented with `sudo`, or without `sudo`?)
2. I believe the hub service listens on port 8080 by default, but this isn't documented in the getting started doc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,9 +31,9 @@ You will need nodejs/npm, which you can get from your package manager:
 (The `nodejs-legacy` package installs the `node` executable,
 which is required for npm to work on Debian/Ubuntu at this point)
 
-Then install javascript dependencies:
+Then install javascript dependencies (note that the first `npm install` should *not* be run with sudo):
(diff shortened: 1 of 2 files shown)


minrk:
`npm install` should *not* be done with sudo, because it installs inside the repo. It *should* also be triggered implicitly by other distutils steps, so shouldn't actually be necessary, but apparently that's not the case.

jhamrick:
Ok, yeah, that makes sense. I'm pretty sure I have successfully installed jupyterhub in the past without needing to run `npm install` but we were having difficulties installing it today because lessc was missing.

minrk:
Thanks. I'm guessing that somewhere it's not behaving correctly when something fails, and just moves on.

PR #267, opened 2015-07-07
require next_url to be an absolute path

minrk:
- disallow relative path
- disallow full URL (cross-site)

### DIFF
--- a/jupyterhub/handlers/login.py
+++ b/jupyterhub/handlers/login.py
@@ -31,7 +31,10 @@ class LoginHandler(BaseHandler):
         )
     
     def get(self):
-        next_url = self.get_argument('next', False)
+        next_url = self.get_argument('next', '')
+        if not next_url.startswith('/'):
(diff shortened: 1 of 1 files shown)


jhamrick:
This seems to have broken redirects for nbgrader formgrade. If I navigate to `http://localhost:8000/hub/nbgrader/course123ABC`, JupyterHub will redirect me to the login form at `http://localhost:8000/hub/login?next=http://localhost:8000/hub/nbgrader/course123ABC`, but after logging in, because full URLs are disallowed, it will take me to the user's server rather than to the formgrader. I think probably what needs to happen is for the login redirection to use the absolute path rather than the full URL; i.e. so the login form is `http://localhost:8000/hub/login?next=/hub/nbgrader/course123ABC`.

jhamrick:
Aha, upon further investigation this isn't JupyterHub's fault as it's actually nbgrader that redirects to `http://localhost:8000/hub/login?next=http://localhost:8000/hub/nbgrader/course123ABC`. Nevermind!

PR #268, opened 2015-07-09
ipython/ipython is deprecated.

dietmarw:
The correct one is now jupyter/notebook


dietmarw:
Not really sure why the test is failing. Perhaps somebody can look into this.

minrk:
There are some intermittent problems with the tests. Nothing wrong here.

PR #276, opened 2015-07-23
Redirect unauthenticated root to login

josephtate:
I was running into problems when trying to build an authenticator that didn't have a login page since he root handler rendered the login.html directly. This redirects to the authenticator's "login_url" instead.

This patch also makes the "url" for a running user a full path to fix a redirect loop I was having.

### DIFF
--- a/jupyterhub/handlers/pages.py
+++ b/jupyterhub/handlers/pages.py
@@ -25,15 +25,16 @@ class RootHandler(BaseHandler):
         if user:
             if user.running:
                 url = url_path_join(self.hub.server.base_url, user.server.base_url)
-                self.log.warn("User is running: %s", url)
+                self.log.debug("User is running: %s", url)
             else:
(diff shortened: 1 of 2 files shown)


minrk:
Thanks

minrk:
@Xarthisius turns out this auto-redirect to login_url isn't the right thing to do in most cases (esp. oauth). I'm ensuring that the login page is rendered again in #288, because the auto-redirect is wrong for everything in OAuthenticator. What's your custom authenticator doing? If it's an external service instead of a form, the authenticator needs to define a `login_service` attribute, which will cause the login page to look like this:

screen shot 2015-08-19 at 12 41 44

Xarthisius:
Right now we're using vanilla `jupyterhub.auth.PAMAuthenticator` that authenticates via LDAP, but that's handled outside of jupterhub

PR #290, opened 2015-08-22
Unset all login cookies

jhamrick:
If jupyterhub is configured so that admin users can access other user's servers, then there is a bug where if the admin user requests access, then logs out, the cookie(s) for the other users are still set and so those servers can still be accessed. This PR fixes that by keeping a set of all users who the admin user has cookies for, and then unsetting those as well when they log out.


jhamrick:
Oh, just realized that I am not removing users from the set after I clear the cookie. I don't have time to fix it now, but I will do it tomorrow.

jhamrick:
Ok, done.

minrk:
Thanks!

PR #296, opened 2015-09-08
use pamela instead of simplepam

### LINKED ISSUE #295: PAM/AD auth doesn't create homedir on first login
So from what i can see in auth.py, the 'login' pam service is used to authenticate the user. For some reason I don't think the 'session' part of this service is honoured. I've tested this by using pamtester with the login service (see below) and the home directory is created.

pamtester login <username> open_session close_session

Any idea why this may be? I think it has something to do with the simplepam python module functionality but i'm not certain.

any help is appreciated. Cheers

minrk:
and open PAM sessions after successful auth

closes #295

### DIFF
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -8,7 +8,7 @@ import pwd
 from subprocess import check_call, check_output, CalledProcessError
 
 from tornado import gen
-import simplepam
+import pamela
 
 from traitlets.config import LoggingConfigurable
 from traitlets import Bool, Set, Unicode, Any
@@ -208,10 +208,11 @@ class PAMAuthenticator(LocalAuthenticator):
         username = data['username']
         if not self.check_whitelist(username):
             return
-        # simplepam wants bytes, not unicode
-        # see simplepam#3
-        busername = username.encode(self.encoding)
-        bpassword = data['password'].encode(self.encoding)
-        if simplepam.authenticate(busername, bpassword, service=self.service):
+        try:
+            pamela.authenticate(username, data['password'], service=self.service)
+            pamela.open_session(username, service=self.service)
+        except pamela.PAMError as e:
+            self.log.warn("PAM Authentication failed: %s", e)
+        else:
             return username
     
--- a/jupyterhub/tests/mocking.py
+++ b/jupyterhub/tests/mocking.py
@@ -18,16 +18,18 @@ from ..app import JupyterHub
 from ..auth import PAMAuthenticator
 from .. import orm
 
+from pamela import PAMError
+
 def mock_authenticate(username, password, service='login'):
-    # mimic simplepam's failure to handle unicode
-    if isinstance(username, str):
-        return False
-    if isinstance(password, str):
-        return False
-    
     # just use equality for testing
     if password == username:
         return True
+    else:
+        raise PAMError("Fake")
+
+
+def mock_open_session(username, service):
+    pass
 
 
 class MockSpawner(LocalProcessSpawner):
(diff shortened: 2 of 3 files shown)


mogthesprog:
Good job on Pamela by the way. When I was trawling the internet last week I got exactly the same impression. I'll be using Pamela in my own project when I get back to work :) 

I'm away this week with work, will get a chance to test Jupyter on Monday. Looking at the source though, combined with some testing i did last week I think this should work. Will pop back to let you know early next week.

Cheers,

Morgan

Adamb5150:
Sorry for the stupid question. I have the exact same issue. I have installed Pamela on the server. I was wondering how I would force Jupyterhub to use Pamela instead of simplepam? 

Thanks

minrk:
@Adamb5150 update JupyterHub to master, and it will do this.

Adamb5150:
Works perfectly. I can login via Active directory and new users automatically have their home directories created. Many thanks!

mogthesprog:
Hi,

Okay so i'm having a strange issue. Login and directory creation is fine. But if i log out, then i can't log back in... I'm still trying to work out why.

In the logout script, Do you close the PAM session in the similar manner that you opened it?

mogthesprog:
and out of curiosity, i'm guessing @Adamb5150, you don't see the same problem?

mogthesprog:
Okay so i've made some progress... The problem is related to a line in /etc/pam.d/login. The line is:

#session    required     pam_loginuid.so

If it's commented out then login works and directory creation works. I believe this is a container specific issue which is why Adam isn't experiencing the same thing. In this article 

http://www.linuxweblog.com/blogs/sandip/20090203/setloginuid-failed-opening-loginuid

It explains that this issue can be seen in Linux (i.e not in a container) when the kernel is compiled wihtout AUDIT enabled. I don't think this is the case form within a container due to the difference in namespaces.

I'm going to continue testing this today and tomorrow. I think you can close this now if you like. I'm pretty sure this is a container related issue. I think your code is sound.

Thanks for all your help.

minrk:
I think what I should really do is add hooks for the authenticator to run before spawner start and after spawner stop. This will provide a better place to call session_stop

Adamb5150:
Hi, Just to confirm that I am not using any sort of container, I can login and log out with no problems.

mogthesprog:
Do you call close_session at all at the moment? I had a quick nose at the source yesterday but I couldn't see anything.

I think the new reliability on the AUDIT kernel functionality is due to the way the authenticate function works at the moment. I'm going to take a look.when I get into work in 5 mins. 

I also experienced a really strange problem yesterday. I login on jupyterhub (with my bodged fix, above) and users only have the primary group. Have you seen this behaviour before?

mogthesprog:
Okay done a bit of digging regarding the groups problem. I'll open a new ticket though. As I'm not sure they're related...

mogthesprog:
So, I've done a bit of digging, I've got a feeling this is related to the authentication. I get the same problem with simplepam though. Basically, when I open  a jupyterhub terminal, I only have my primary group...

[morganj@cb0677f233db ~]$ groups  
Unix_Users  

However, my groups are available to me...

[morganj@cb0677f233db ~]$ groups
Unix_Users
[morganj@cb0677f233db ~]$ newgrp UNIX_AS_AN
[morganj@cb0677f233db ~]$ groups
UNIX_AS_AN             <------- (1)
[morganj@cb0677f233db ~]$ exit 
[morganj@cb0677f233db ~]$ newgrp UNIX_AS_AN
[morganj@cb0677f233db ~]$ newgrp Unix_Users
[morganj@cb0677f233db ~]$ groups 
Unix_Users UNIX_AS_AN     <---- (2)
[morganj@cb0677f233db ~]$ su adamho
Password: 
[adamho@cb0677f233db morganj]$ groups 
Unix_Users cm GROUP1 GROUP2 GROUP3 GROUP4 <----- (3)

Okay so at (1) we see I have one group. What's strange is that a newgrp command replaces my primary group. This isn't native linux behaviour. So something is a bit weird. Then if i issue a second newgrp command in subshell, the groups start appending. This is how newgrp is supposed to work (2).                    

And lastly, we see me using 'su' to change to another user (3).... And lo-and behold, We get all our groups back. I get the impression this is something to do with the terminal implementation in Jupyter.

PR #304, opened 2015-09-21
Remove implicit admin of launching user

minrk:
when no admin user is specified via config, which can have confusing results.

instead, warn about missing admins and point to config.

cc @claresloggett

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -628,10 +628,8 @@ class JupyterHub(Application):
         admin_users = self.authenticator.admin_users
         
         if not admin_users:
-            # add current user as admin if there aren't any others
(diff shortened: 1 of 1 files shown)


jhamrick:
:+1: I like this change, I've never really been happy with it automatically adding root as a Jupyter user when it's not a real user either.

Carreau:
Make sens to me too.

Do you want to wait to merge ?

claresloggett:
This change is good from my point of view.

As a newbie having an automatically created admin actually helped me in one
case. However it's hit-and-miss, and the user may not even realise this has
been done for them. A warning on launch that there are no admins would be
more useful.

On 22 September 2015 at 07:36, Matthias Bussonnier <notifications@github.com

minrk:
Thanks all! Merged.

PR #336, opened 2015-11-16
Added npm/node which fixes the Dockerfile

Fokko:
Modified the Dockerfile so it will include npm and node. Also it runs apt-get update to get all the updated packages for security reasons. This will fix the docker build.

Cheers.


jdfreder:
@Fokko looks like the email in the commit isn't associated with your GitHub account, do you want to fix that?

Fokko:
My bad, it should be associated with my account now.

PR #346, opened 2015-12-05
Do not reset user.state on every spawn start

### LINKED ISSUE #337: Pass OAuth access_token to the Spawner
Hi,

Is it possible to pass the access_token generated during the OAuth authentication to the notebooks?

Our goal would be to authenticate the users against our API and setup the notebooks such that they don't have to authenticate again when using the API from their scripts.

yuvipanda:
This allows the Authenticator to also store state (like tokens)
in the user.state object

Fixes #337

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -139,8 +139,10 @@ class Spawner(LoggingConfigurable):
         state: dict
              a JSONable dict of state
         """
-        state = {}
-        return state
+        if self.user.state:
+            return self.user.state
+        else:
+            return {}
     
     def clear_state(self):
         """clear any state that should be cleared when the process stops


yuvipanda:
hmm, so after fiddling around with the test failures, I have come to realize that this is actually a breaking change for almost all spawners that use get_state, since they now expect it to be empty. Not sure if fixing all of them is worth it.

Maybe we can add another state field that holds state for the authenticator?

Not sure what the right thing to do here is.

minrk:
`load_state` should be called on creation of the Spawner, so loading it again in `get_state` would be circular, so this definitely is not the right fix. If anything, I suspect what you meant to do here is to remove the calls to `clear_state`.

Instead, what I think you want to do is to override `clear_state` in your Spawner subclass, so that it doesn't clear the state you want to persist across Spawner instances. This should already work.

yuvipanda:
The problem with that is that that couples the Authenticator and the Spawner very tightly. I can couple MWOAuthenticator and KubeSpawner but not DockerSpawner...

yuvipanda:
Also my understanding of how the different state related functions work is:

# get_state is for the spawner to store state from its instance variables / what not into the db
# load_state is for reversing ^ - from db onto instance variables / what not

Currently, this cycle starts with a `{}`, so anything stored in state from outside spawners is erased the first time a spawner goes through a load_state / get_state cycle. This might very well be ok, but then we need another mechanism for storing state from the authenticators that persists across process restarts.

minrk:
> Currently, this cycle starts with a {}

Only at the very beginning, before the Spawner has been created for the first time. For example:

```python
class PersistentSpawner(Spawner):
    key = Unicode(uuid.uuid4().hex)

    def get_state(self):
        state = super().get_state()
        state['key'] = self.key
        return state
    
    def load_state(self, state):
        print("Loading state", state)
        if 'key' in state:
            self.key = state['key']
        super().load_state(state)
    
```

will store 'key' in state, and load it on subsequent starts. Only if `clear_state` explicitly deletes the key and `get_state` understands the cleared value will the value be deleted.

> I can couple MWOAuthenticator and KubeSpawner but not DockerSpawner...

That state is also useless to DockerSpawner unless you subclass it, so how would persisting it help if the Spawner doesn't understand it?

PR #347, opened 2015-12-08
Log Remote IP Address for Failed Authentication Attempts

dblockow:
Simple change to log the remote IP address for failed authentication attempts.

This is useful when combined with tools like [fail2ban](http://www.fail2ban.org/) that parse application logs for malicious behaviour.


minrk:
Thanks!

PR #354, opened 2015-12-15
IPv6 ready /etc/hosts file without IPv6 enabled causing localhost issue

zoltan-fedor:
This is to resolve the 'Network is Unreachable' error experienced by a few when JupyterHUB is running on default localhost, see https://github.com/jupyter/jupyterhub/issues/271

On most recent linux OS versions like CentOS 6, 7, Red Hat 6, 7, Oracle Linux 6, 7, etc, the hosts file (/etc/hosts) usually has a line to make the server IPv6-ready:
    ::1 localhost
even if the given server actually has no IPv6 permissioned. In such case the Python socket library when connecting to 'localhost' will try to connect via the IPv6 protocol - which will fail with the 'Network is Unreachable' error.

To solve this we capture this error and try to reconnect on 127.0.0.1 instead of localhost, alias forcing the user of the IPv4 protocol.

### DIFF
--- a/jupyterhub/orm.py
+++ b/jupyterhub/orm.py
@@ -116,7 +116,17 @@ class Server(Base):
         try:
             socket.create_connection((self.ip or 'localhost', self.port))
         except socket.error as e:
-            if e.errno == errno.ECONNREFUSED:
+            if e.errno == errno.ENETUNREACH:
+                try:
+                    socket.create_connection((self.ip or '127.0.0.1', self.port))
+                except socket.error as e:
+                    if e.errno == errno.ECONNREFUSED:
+                        return False
+                    else:
+                       	raise
+                else:
+                    return True
+            elif e.errno == errno.ECONNREFUSED:
                 return False
             else:
                 raise


minrk:
Does #353 fix this?

zoltan-fedor:
Unfortunately #353 did not solve it, probably because as I wrote here (https://github.com/jupyter/jupyterhub/issues/271#issuecomment-164775526) that code returns 'localhost' and not the '127.0.0.1' fallback, so your code is still trying to connect via 'localhost'.

minrk:
Sorry, I'm struggling to see what's different in this code from the test in #353. Can you see why this connection fails with `ENETUNREACH`, while the other succeeds?

zoltan-fedor:
The problem is with your error handling in the original code.

Your original code only handles `ECONNREFUSED`, while when you are making a connection in this scenario via `localhost` you will get `ENETUNREACH` and NOT `ECONNREFUSED`.
This is why your original code will raise that as an unhandled exception.

What you should be testing in your localhost() function is not whether or not you can make a valid connection with `localhost` when there is such a socket, but what type of error message you will get back when there is no such socket. If you get `ENETUNREACH`, then if you switch `localhost` to `127.0.0.1` you will get `ECONNREFUSED` - and your error handling of `ECONNREFUSED` will start working as-is.

This is why my above patch could actually be even more simple, just changing a single line from your original code in the `is_up()` function:
`if e.errno == errno.ECONNREFUSED:`

to

`if e.errno == errno.ECONNREFUSED or e.errno == errno.ENETUNREACH:`

and the rest could stay the same.

If you want, I can submit a patch for that.

I hope this makes it more clear now.

minrk:
What I don't understand is why the initial test doesn't raise ENETUNREACH, if later calls do. I don't see the reason for that. In general, we *should* be using `localhost`, not `127`, so switching that doesn't seem like a fix to me.

PR #355, opened 2015-12-17
Add Spawner form page

minrk:
cc @zonca, @ellisonbg

Spawners can define an options_form class attribute. This is an HTML snippet, inserted into a form page. If defined, spawning a server loads this form instead of spawning directly.

The form data is passed to the Spawner.options_from_form *classmethod*, in order to turn the form structure (always a dict of lists of strings) into the intended shape (handling numbers, single values, defaults, etc.).

With this, Spawners can expose arbitrary parameters for users to override,
such as resources, hosts, etc.

Specifying user-configurable parameters is completely orthogonal to the admin-configured values from `config=True` traits, since there's ~no relationship between what is configrable in general and what should be exposed to the user.

Possible future ideas:

- declarative form data instead of pasting raw HTML. There are roughly one million ways to turn a small spec into a webform. IPython widgets are one, but roughly none of the work there is reusable here. We could build a similar, simpler traits->form control, but we probably shouldn't. This would be a convenience, but it doesn't preclude using an arbitary form, so I think we can start with a fully arbitary form, and come up with a more convenient form-building spec at a later point.


ellisonbg:
I like this idea and think it is a good idea to keep it simple (as it is here) starting out.

I see why these form options have to be handled separately from our traitlets config stuff. I am fine with that but if you can clean it up that is also fine too.

I think for now, going with raw HTML makes the most sense - eventually if we have completely standardize UI controls/layout, we could explore the declarative form generation - but I wouldn't worry about it now.

Great work!

zonca:
great, thanks, I'll try to test and provide feedback next week! @minrk do you have a gist of a simple working example of this? I see you have unit tests but I don't understand much the mocking machinery.

minrk:
I moved Spawner initialization earlier, so that `options_form` can be a configurable, and accessed on the instance.

@zonca I added an example which allows specifying env and args for the single-user notebook server.

zonca:
Thanks @minrk, works great:

screenshot from 2015-12-23 15-20-14

Just a question, I noticed `formdata.get` for an input field returns a list like ["12:00:00"], is there a specific reason for that? it would be more convenient to just get a string back.

minrk:
> Just a question, I noticed formdata.get for an input field returns a list like ["12:00:00"], is there a specific reason for that? it would be more convenient to just get a string back.

Yes, since HTTP forms are an extremely primitive format where it's impossible to tell the difference between a single value and a list of length one. In order to be generally correct, tornado always returns a list of strings. This is precisely what the `options_from_form` method is for: to turn the HTTP form data into the right structure.

PR #363, opened 2015-12-30
Initial commit of Sphinx docs using md and rst as source files

willingc:
@minrk @ellisonbg Here's a sphinx project for docs that will support either Markdown or Restructured Text. 

The guts of the changes to support Markdown are in several lines of ``conf.py``:
```
# Jupyter modifies this to allow markdown and restructured text files
source_suffix = ['.rst', '.md']

source_parsers = {
  '.md': 'recommonmark.parser.CommonMarkParser',
}
```

It's still rough in corners like the fonts used with code snippets which is probably a function of the source parsing using recommonmark. Also adding in Min's autoversioning code snippet to conf.py and rtd/not_rtd guard also in conf.py.

Within the documentation the users could use rst or md. We should encourage markdown here but it wouldn't matter for the rtd docs. It would be nicer to use md for reading on GitHub.

P.S. Min, I did my best to get all the sphinx cruft in a ``docs-sphinx`` directory that was at the same tree level as ``docs`` so that the source directory would only contain the md files. Unfortunately, even with changing configuration parameters and trying symlinks, Sphinx really expects to be the root of all documentation. We could create a ``docs/source`` directory to hold all the md source but that moves the md into a subdirectory and I wasn't sure that you wanted that. Easy change if desired.


willingc:
A new Sphinx project, [sphinxcontrib-markdown](https://github.com/tk0miya/sphinxcontrib-markdown), appears to do a better job of markdown conversion so I've modified the PR to use this conversion project.

minrk:
To me, it looks like recommonmark, which is maintained by RTD itself, is probably the way to go. There's an important bug (rtfd/recommonmark#22) that has been fixed in master (CommonMark-py#37), but I think we can wait for that to get in. sphinxcontrib-markdown doesn't handle fenced code blocks, or other GFM features that we use and are part of CommonMark.

willingc:
@minrk Thanks for the review and feedback. Nice find on the root cause of the duplicate output which I was seeing when using the recommonmark based extension. I'm going to close this PR since I agree that with the fix to CommonMark that recommonmark will be a better solution. Thanks.

PR #371, opened 2016-01-06
delete users via UserDict API

minrk:
avoids reusing user IDs when user creation fails

Fixes the most important part of #369


willingc:
+1 to logic. I haven't run the code to test.

PR #372, opened 2016-01-08
drop support for single-user server from IPython 3.x

minrk:


jupyterhub/singleuser.py
```diff
@@ -17,34 +17,22 @@
 from tornado import ioloop
 from tornado.web import HTTPError
 
+try:
+    import notebook
+except ImportError:
+    raise ImportError("JupyterHub single-user server requires notebook >= 4.0")
```
Carreau: Can you `except .. as e` and `raise ImportError(...) from e` in case the import error is due to something else ? Should be safe as jupyterhub is py3 only.

Carreau:
+1

jupyterhub/singleuser.py
minrk: single-user server isn't py3-only, actually.

ssanderson:
@minrk curious if something prompted this, or is this just an "it's finally time" change?  (I ask b/c we're still on IPython 3.x on Quantopian while we're waiting for the frontend to settle down.)  cc @richafrank @TimShawver.

minrk:
I saw a bug related to traitlets changes, and was surprised that this was still allowed when I looked into it.

How rough would it be for you folks to use notebook 4.0? This bump is going to happen at some point.

minrk:
This will need rebase on #387

ssanderson:
@minrk realistically speaking it's unlikely that we're going to upgrade to notebook 4.0 in the next month or so.  If this is a change that needs to happen, then you shouldn't wait up on us (though it would be nice to have a tagged release for the last version that supports the old version of the notebook.)

minrk:
@ssanderson thanks. I'm working on a PR now that I think will be simpler if it requires the Jupyter base. I think it will probably make sense to cut 0.4 shortly as the last release supporting IPython 3, and then drop it for 0.5.

minrk:
@ssanderson I released and tagged 0.4, so we can go ahead with this for 0.5. Let me know if there's anything I can do to help you guys update your servers to 4.0 (note: doesn't have to include updating kernels).

PR #375, opened 2016-01-09
Handle file upload in spawner form

betatim:
Allow files to be uploaded in the spawner form. Files are added to `form_options` by adding `_file` to their key to allow separating them from other form items.

Unsure, but I don't think we need to do anything like `[ bs.decode('utf8') for bs in byte_list ]` for the files as they can contain all sorts of binary data.


minrk:
Right, the passthrough should be as generic is possible. The Spawners can do their own decoding, as appropriate.

betatim:
Thanks for the quick merge!

PR #377, opened 2016-01-12
add swagger spec for REST API

minrk:
See it [rendered here]( http://petstore.swagger.io/?url=https://raw.githubusercontent.com/minrk/jupyterhub/swagger-spec/docs/rest-api.yml#/default)

### DIFF
--- a/docs/rest-api.yml
+++ b/docs/rest-api.yml
@@ -1,0 +1,259 @@
+# see me at: http://petstore.swagger.io/?url=https://raw.githubusercontent.com/jupyter/jupyterhub/master/docs/rest-api.yml#/default
+swagger: '2.0'
+info:
+  title: JupyterHub
+  description: The REST API for JupyterHub
+  version: 0.4.0
+schemes:
+  - http
+securityDefinitions:
+  token:
+    type: apiKey
+    name: Authorization
+    in: header
+security:
+  - token: []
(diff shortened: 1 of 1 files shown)


willingc:
Nice! Travis is only failing on Python 3.3 which I don't think is necessary to support for docs.

betatim:
It is failing on a return code being different for a call to the API. Seems
odd that this only fails one on platform and I don't think the commit has
anything to do with it. Heisenbug?

minrk:
Yes, it's a Heisenbug. The test is specifically for special handling when something takes too long, so I'm guessing sometimes it actually finishes promptly. In any case, it's an intermittent bug in the tests, not the code.

minrk:
It'll be interesting if we can render the swagger HTML into the sphinx docs on RTD. I'll look into the feasibility of that.

willingc:
@minrk I was thinking the same thing. I will play around with it tomorrow too. When I looked this morning there were some swagger to markdown converters which might help. Though using it live via swagger is awesome and really what I would prefer as a developer.

rgbkrk:
Love that swagger.

willingc:
@rgbkrk Lol

rgbkrk:
No seriously, I adore navigable API docs.

Ok, I *was* making a joke.

PR #380, opened 2016-01-13
move install commands around a bit

minrk:
npm/less notes are only relevant for dev installs


willingc:
@minrk Thanks for making the changes. Happy to merge.

PR #381, opened 2016-01-13
add preliminary API docs

minrk:
documents relevant Authenticator, Spawner, User signatures for subclass authors.

I'd like to also try an autodoc plugin that catches configurable values.


docs/source/conf.py
```diff
@@ -35,7 +35,7 @@
 extensions = [
     'sphinx.ext.autodoc',
     'sphinx.ext.intersphinx',
-    'sphinx.ext.mathjax',
+    'sphinx.ext.napoleon',
```
willingc: @minrk Nice! This should allow nicer docstrings. Probably should uncomment and update Line 30 to '1.3'. Prior to '1.3' the extension was named differently.

willingc:
Lots of good stuff in this PR. I'll give it a more thorough review later today :-)

docs/source/conf.py
minrk: done, thanks.

minrk:
Merging now, we can iterate in future PRs.

willingc:
@minrk Sorry I should have pinged you earlier that the review was good. The API docs are a great addition and it looks good.

I think that we need to enable the Webservice for this repo for ReadTheDocs since the docs were not automatically triggered to rebuild after this merge. I did go into RTD and manually build so the latest docs are current. If you want to enable that would be great, I'm happy to do it but would need repo privileges to do so.

minrk:
Thanks! I turned on the webhook, so hopefully it should be good. You should have permission to do that too, now.

willingc:
@minrk Thanks. :sunny:  I found a few other little things related to sphinx/markdown and submitted a PR and opened another issue.

PR #389, opened 2016-01-22
Fix docstrings *ix -> Linux/UNIX to prevent Sphinx build warnings

willingc:
When using autodoc, docstrings that contain `*` in  a line cause a warning. The warning indicates that emphasis was started but not stopped. Since this only occurs in 3 places in one file, I recommend replacing `*ix` with `Linux/UNIX` to keep Sphinx from whining.

Looks like Atom also cleaned up stray spaces on blank lines as well.


minrk:
Merged without whitespace changes in 1a21e822b67277ded9abb3dff0c34524ace440ce

PR #393, opened 2016-01-22
Use relative html link instead of local md

### LINKED ISSUE #390: [Docs] Markdown links are not handled properly by Sphinx if referencing local doc
See http://jupyterhub.readthedocs.org/en/latest/getting-started.html#further-reading 

Custom authenticators and spawners links redirect to a "Not Found" page. This is due to Sphinx not handling markdown links to local files correctly. I will research if there is an automatic workaround. If not, we could link to the RTD rendered pages such as: http://jupyterhub.readthedocs.org/en/latest/authenticators.html

willingc:
Closes #390 

@minrk Great suggestion. Works for a local build so unless RTD does something odd this should work.

### DIFF
--- a/docs/source/getting-started.md
+++ b/docs/source/getting-started.md
@@ -398,8 +398,8 @@ jupyterhub -f /path/to/aboveconfig.py
 # Further reading
 
 - TODO: troubleshooting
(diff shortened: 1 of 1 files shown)


minrk:
Perfect, thanks!

PR #399, opened 2016-01-25
Add circleci badge with status only API token

willingc:
I saw that #395 was merged. I've added the badge to the README.


Carreau:
Failures were random. Passes now.

PR #402, opened 2016-01-28
Fixing the custom_html feature in the login form

mistercrunch:
The template clearly still points to `custom_html`, which is consistent with the class property. `custom_login_form` is not referenced anywhere in the template.


minrk:
Thanks!

PR #405, opened 2016-02-01
Update documentation link to source code

willingc:
Update link to Spawner source code. I've used an external link to the file on GitHub. 

The relative link does not redirect properly (even when corrected) to the GitHub hosted file. Read The Docs returns not found. Sphinx and RTD will be happier with the external link.

@minrk Nice work on 0.4.


minrk:
Thanks!

Issue #406, opened 2016-02-01
Add more detail about cookie secret permissions and ip setting for configurable-http-proxy


willingc:
Based on a conversation on today's gitter channel, I think it would be helpful to document:

1) That the recommended permissions for the "Cookie secret should be 600 (owner-only rw )"

2) configurable-http-proxy link to https://github.com/jupyter/configurable-http-proxy/blob/master/README.md

From the command line, ``configurable-http-proxy --ip=0.0.0.0`` will set the 
configurable-http-proxy to any interface to this host on this network. 

The "*" alias for all interfaces is not liked by CHP. All interfaces is the default " ".

Perhaps this could go at the end of https://jupyterhub.readthedocs.org/en/latest/getting-started.html#networking. Also perhaps a link to CHP in https://jupyterhub.readthedocs.org/en/latest/howitworks.html bullet list re: proxy

PR #407, opened 2016-02-01
Add doc details for #406

willingc:
Addresses #406 

- Adds link to http-configurable-proxy repo
- Adds recommended permission setting (600) for Cookie Secret file
- Rewording to, hopefully, improve info on configuring the Proxy and Hub's networking based on Gitter feedback

### DIFF
--- a/docs/source/getting-started.md
+++ b/docs/source/getting-started.md
@@ -14,11 +14,11 @@ See [the readme](https://github.com/jupyter/jupyterhub/blob/master/README.md) fo
 JupyterHub is a set of processes that together provide a multiuser Jupyter Notebook server.
 There are three main categories of processes run by the `jupyterhub` command line program:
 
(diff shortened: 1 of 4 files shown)


docs/source/getting-started.md
```diff
@@ -96,32 +96,38 @@ on the config system Jupyter uses.
 
 ## Networking
 
-In most situations you will want to change the main IP address and port of the Proxy.
-This address determines where JupyterHub is available to your users.
-The default is all network interfaces (`''`) on port 8000.
+### Configuring the Proxy's IP address and port
+Most use cases will require a change of the Proxy's main IP address and port.
```
minrk: While we are here, can we change this part? I don't think it's true that in most cases this will want to be changed. I imagine that text may be leftover from the default being localhost in early prototyping. Perhaps just strike the first sentence and leave the rest as-is?

willingc:
@minrk Also added a troubleshooting section to get rid of the TODO in the docs that was irritating me.

PR #411, opened 2016-02-03
use 127.0.0.1 instead of localhost

minrk:
localhost can cause some issues on badly behaved or misconfigured systems, and 127 seems simpler.


yuvipanda:
:+1:

yuvipanda:
(http://yuvi.in/blog/dns-servers-localhost-asynchronous-code/ for where I ran into this)

minrk:
Thanks for the background!

PR #413, opened 2016-02-03
Restore /login handler

minrk:
While a custom `login_url` can be used, e.g. for oauth, `/login` is still used for the "Login with ..." confirmation page.

### DIFF
--- a/jupyterhub/handlers/login.py
+++ b/jupyterhub/handlers/login.py
@@ -86,7 +86,9 @@ class LoginHandler(BaseHandler):
             self.finish(html)
 
 
-# Only logout is a default handler.
+# /login renders the login page or the "Login with..." link,
+# so it should always be registered.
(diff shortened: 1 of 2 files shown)


jupyterhub/handlers/login.py
```diff
@@ -88,5 +88,6 @@ def post(self):
 
 # Only logout is a default handler.
```
fperez: You should edit the comment to reflect the code below.

fperez:
Other than a trivial comment fix, looks good, and since it's hitting users as per the on-list report, I suggest a quick merge.

jupyterhub/handlers/login.py
minrk: Good catch, thanks.

minrk:
Thanks, will merge and release 0.4.1 shortly.

PR #416, opened 2016-02-04
Add more prominent message for https

### LINKED ISSUE #415: Sending plain password through HTTP
I was listening to Jupyter's port, of course after changing it to unused port. So, when I read packets I find the following:

> T 192.168.124.43:42452 -> 192.168.124.43:18541 [AP]
  POST /login?next=%2Ftree HTTP/1.1..Host: 192.168.124.43:18541..User-Agent: Mozilla/5.0 (X11; Debian; Linux x86_64; rv:44.0) Gecko/20100101 Firefox/44.0..Accept: text/html,application/x
  html+xml,application/xml;q=0.9,*/*;q=0.8..Accept-Language: en-US,en;q=0.5..Accept-Encoding: gzip, deflate..Referer: http://192.168.124.43:18541/login?next=%2Ftree..Connection: keep-ali
  ve..Content-Type: application/x-www-form-urlencoded..Content-Length: 22....password=qwer+1234

It seems Jupyter hash password after receiving it, which can be exploited. 
 
 I suggest that jupyter save plain password in a safe place. Then, each time Jupyter sends a random string to the user. After that, user hash his password with that salt then send his hashed string. Finally, Jupyter salts plain password then hash it and compare it. 
* A better solution is to use HTTPS where all contents are encrypted. This solution is illustrated nicely in the documentations.

willingc:
Closes #415.

Make existing https message more prominent and add a second message in Security section. Ping @minrk for review.

### DIFF
--- a/docs/source/getting-started.md
+++ b/docs/source/getting-started.md
@@ -22,6 +22,9 @@ There are three main categories of processes run by the `jupyterhub` command lin
 
 ## JupyterHub's default behavior
 
+**IMPORTANT:** In its default configuration, JupyterHub runs without SSL encryption (HTTPS).
+**You should not run JupyterHub without SSL encryption on a public network.**
+See [Security documentation](#Security) for how to configure JupyterHub to use SSL.
 
 To start JupyterHub in its default configuration, type the following at the command line:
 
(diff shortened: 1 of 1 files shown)


minrk:
Nice. And it might even be worth it to require explicit confirmation that it should be run without https.

willingc:
@minrk I'm going to merge this PR. When PR #417 lands, I will add docs for that too. I think explicit confirmation is wise.

PR #417, opened 2016-02-04
require confirmation for JupyterHub to run without SSL

minrk:
ensures folks deploying JupyterHub on HTTP have been told what's up.

related: #416

cc @ellisonbg: for those running behind nginx, if we do this, JupyterHub will refuse to start running without SSL unless `--no-ssl` is specified.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -87,6 +87,9 @@ flags = {
     'no-db': ({'JupyterHub': {'db_url': 'sqlite:///:memory:'}},
         "disable persisting state database to disk"
     ),
+    'no-ssl': ({'JupyterHub': {'confirm_no_ssl': True}},
+        "Allow JupyterHub to run without SSL (SSL termination should be happening elsewhere)."
+    ),
 }
 
 SECRET_BYTES = 2048 # the number of bytes to use when generating new secrets
@@ -208,7 +211,12 @@ class JupyterHub(Application):
 
     def _template_paths_default(self):
         return [os.path.join(self.data_files_path, 'templates')]
-
+    
+    confirm_no_ssl = Bool(False, config=True,
+        help="""Confirm that JupyterHub should be run without SSL.
+        This is **NOT RECOMMENDED** unless SSL termination is being handled by another layer.
+        """
+    )
     ssl_key = Unicode('', config=True,
         help="""Path to SSL key file for the public facing interface of the proxy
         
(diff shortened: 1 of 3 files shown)


jupyterhub/app.py
```diff
@@ -800,6 +808,18 @@ def start_proxy(self):
             cmd.extend(['--ssl-key', self.ssl_key])
         if self.ssl_cert:
             cmd.extend(['--ssl-cert', self.ssl_cert])
+        # Require SSL to be used or `--no-ssl` to confirm no SSL on 
+        if ' --ssl' not in ' '.join(cmd):
+            if self.confirm_no_ssl:
+                self.log.warning("Running JupyterHub without SSL."
+                    " There better be SSL termination happening somewhere else...")
```
willingc: This is insecure. <-- Add between the two log messages.

jupyterhub/app.py
```diff
@@ -208,7 +211,12 @@ class JupyterHub(Application):
 
     def _template_paths_default(self):
         return [os.path.join(self.data_files_path, 'templates')]
-
+    
+    confirm_no_ssl = Bool(False, config=True,
+        help="""Confirm that JupyterHub should be run without SSL.
+        This is **NOT RECOMMENDED** unless SSL termination is being handled by another layer.
```
willingc: @minrk Might be helpful for end users to know True results in running without SSL and False will run the hub with SSL.

willingc:
+1 on the code. Two suggested. minor edits for the help and log messages.

minrk:
I'd like to hear from some of @jhamrick, @ellisonbg, @rgbkrk on this one. Many deployments do SSL termination at nginx, which would be doing prudent SSL, but have to pass `--no-ssl` to get JupyterHub to run.

rgbkrk:
This forces people to run an explicit flag to acknowledge what they're doing. I'm in favor.

I'm definitely happy about the warning and suggestion text too.

jhamrick:
That sounds fine to me, as long as there's the option for running without SSL. I think this is a good change, actually, as I expect there may be people trying to run without SSL who probably should.

willingc:
@minrk When we update the docs for this, we could give your nginx example of a prudent use case where one may choose to run --no-ssl.

rgbkrk:
On a similar note, I just talked to @smashwilson about how we'd adapt the lets-nginx container to include our websockets configuration. We're putting together some workshop materials for just this.

yuvipanda:
:+1: for explicit param.

PR #421, opened 2016-02-08
Added label to dockerfile for referencing

Fokko:
Added a label to reference the container. This is used to query the container from the docker api.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,7 +46,7 @@ WORKDIR /srv/jupyterhub/
 
 EXPOSE 8000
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -46,5 +46,7 @@ WORKDIR /srv/jupyterhub/
 
 EXPOSE 8000
 
+LABEL com.jupyter.service="jupyterhub"
```
minrk: should be org.jupyter, presumably. Otherwise :+1:

Fokko: Check!

minrk:
Thanks!

PR #432, opened 2016-02-18
disable port_retries in single-user server

minrk:
Since Spawners won't notice that the server has started somewhere other than where it was asked to, it should fail outright.


rgbkrk:
Ah, perfect. That helps the swarm case when we have errors.

PR #433, opened 2016-02-18
allow disabling user configuration of single-user servers

minrk:
If admins want to enforce total control, this should make it easy to switch off user-owned configuration.


rgbkrk:
Still WIP? Needs a rebase.

minrk:
Yeah, I need to do some polishing on this one. The main missing thing is enabling top-level config to turn it on.

minrk:
Should be good to go.

PR #431, opened 2016-02-17
Update README.md

ObiWahn:
Hi,

please let me know what needs to be changed.

~ Jan

PR #434, opened 2016-02-19
Don't let the default include `--no-ssl`.

rgbkrk:
Follow up to #431.


rgbkrk:
The text could use further refinement to include the SSL args, I'm putting this in so at the very least the `--no-ssl` isn't staying in.

/cc @willingc @minrk

willingc:
Thanks @rgbkrk

PR #435, opened 2016-02-20
add debug logging for adding users with no running server

minrk:
in check_routes, it has been reported that users without a running server are attempted to be added.

So something is wrong, either in sqlalchemy or my understanding of what it does (likely the latter), because [this filter]([this line](https://github.com/jupyter/jupyterhub/blob/0.4.1/jupyterhub/orm.py#L230)) for users with a non-None server is returning at least one result whose server is None [here](https://github.com/jupyter/jupyterhub/blob/0.4.1/jupyterhub/orm.py#L233).


minrk:
cc @ryanlovett who reported the issue.

PR #436, opened 2016-02-22
allow running single-user servers on subdomains

minrk:
relies on CHP's host-based routing (a feature I didn't add!)

requires wildcard DNS and wildcard SSL for a proper setup

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,3 +15,7 @@ script:
     - py.test --cov jupyterhub jupyterhub/tests -v
 after_success:
     - codecov
+matrix:
+  include:
+    - python: 3.5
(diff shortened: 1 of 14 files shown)


rgbkrk:
This is so great.

willingc:
Very cool. Subdomains for class/lecture sections :+1:

minrk:
Okay, I think everything's working, now. I just need to rig up some way to run the tests with the subdomain flag on via xip.io, and it should be good to go.

minrk:
It works! Thanks to xip.io, it's even testing with (and without) subdomains on Travis.

rgbkrk:
The xip tricks are so handy here.

willingc:
Being able to test with xip is really nice.

Issue #345, opened 2015-12-02
Broken links on single user servers if "@" symbol in username


CedricFch:
If a username contains '@' symbol, some bugs appear in single-user servers because the '@' symbol is escaped as '%40' for API links.

I solved this issue except a few links (for example when stopping the server) by putting at line 118 of User.py:
base_url=url_path_join(base_url, 'user', self.escaped_name.replace("@",".")),

 but I don't know whether this is the best way to deal with this issue...

minrk:
This is a bug in the single-user notebook package, and should be fixed by the upcoming release of notebook-4.1.

yuvipanda:
I'm running into the same issue I believe - can you point out the commit responsible for fixing it so I can cherry-pick it for my production install? Thanks

minrk:
I think it's a few fixes, but the primary PR is jupyter/notebook#472. Can you check if it works using jupyterhub and notebook master?

yuvipanda:
I just tested this on notebook 4.1b1 and didn't fix this :(

minrk:
It's possible that the way it's fixed in the notebook requires another change on the Hub's side.  I'll look into it. I know I've used it.

minrk:
I just tested in a fresh env with

    $ pip install --pre notebook jupyterhub
    $ pip list
    appnope (0.1.0)
    decorator (4.0.6)
    gnureadline (6.3.3)
    ipykernel (4.2.1)
    ipython (4.0.1)
    ipython-genutils (0.1.0)
    Jinja2 (2.8)
    jsonschema (2.5.1)
    jupyter-client (4.1.1)
    jupyter-core (4.0.6)
    jupyterhub (0.3.0)
    MarkupSafe (0.23)
    mistune (0.7.1)
    nbconvert (4.1.0)
    nbformat (4.0.1)
    notebook (4.1.0b1)
    pamela (0.2.1)
    path.py (8.1.2)
    pexpect (4.0.1)
    pickleshare (0.5)
    pip (7.1.2)
    ptyprocess (0.5)
    Pygments (2.0.2)
    pyzmq (15.1.0)
    requests (2.8.1)
    setuptools (18.2)
    simplegeneric (0.8.1)
    SQLAlchemy (1.0.10)
    terminado (0.5)
    tornado (4.3)
    traitlets (4.1.0b1)
    wheel (0.24.0)

and it works with `@` in the username and URLs. What do you see as not working?

yuvipanda:
[...]

PR #440, opened 2016-02-24
Do not consider `@` character url-safe

anderbubble:
Usernames that have an `@`-separated domain component
break JupyterHub when the server expects to see query
strings that contain an `@`, when browsers and other
clients send `%40`.

### DIFF
--- a/jupyterhub/user.py
+++ b/jupyterhub/user.py
@@ -146,7 +146,7 @@ class User(HasTraits):
     @property
     def escaped_name(self):
         """My name, escaped for use in URLs, cookies, etc."""
-        return quote(self.name, safe='@')
+        return quote(self.name, safe='')
     
     @gen.coroutine
     def spawn(self, options=None):


rgbkrk:
I can't tell if this is the right choice for the default. What spawner are you using?

anderbubble:
I'm using the slurmspawner.

https://github.com/mkgilbert/slurmspawner

I'd be interested to know how the choice of spawner would affect this, though. Even my browser, when following an href with an `@` in it, changed it to `%40` for the request, which generated a 404, because JupyterHub was expecting an `@`.

minrk:
@anderbubble what versions of things? I've been using `@` in usernames for a while now, without issue. This was a bug in the single-user notebook fixed in 4.1, where it was *our javascript* that turned `@` to `%40`, not the browser.

The spawner shoudn't be relevant, except insofar as it can influence what version of the notebook server is in use.

anderbubble:
@minrk, I'm pretty sure we're on 4.0, which matches up with what you're saying. I'll try to get us migrated to 4.1 to test.

That said:

* It's strange to me to think that javascript was prompting the change to `%40`. For example, hovering over the "home" button predicted a link to a non-encoded url in the status bar; but actually clicking on it used a %-encoded url.

* I'd argue that both %-encoded and non-encoded URLs should be valid at the server, though; potentially with one redirecting to the other.

Assuming you're right, though, I do think you did the right thing to revert the merge. Like I said: I'll try to upgrade us to 4.1 and re-test.

anderbubble:
@minrk Just for the record, I was also getting these [redacted] errors (repeatedly) in my singleuser log:

    [W 2016-02-24 13:52:04.897 ${user}@${domain} log:47] 404 GET /user/${user}%40${domain}/api/terminals?_=1456343250471 ($ip) 2.30ms referer=https://${server}/user/${user}@${domain}/tree
    [D 2016-02-24 13:52:18.180 ${user}@${domain} jupyterhub-singleuser:183] Clearing cookie cache
    [D 2016-02-24 13:53:02.559 ${user}@${domain} jupyterhub-singleuser:113] No token cookie

minrk:
The escaping was a bug in our javascript, not javascript itself. We were calling `encodeURIComponent` on what should have been treated as an already escaped url. This was true whether `@` was in there or not. Escaping `@` to `%40` on the same versions would result in 404s on `%2540` because the same js would escape the `%`.

anderbubble:
@minrck We upgraded to 0.4.1 today and see the same behavior. I'm going to open an actual issue, since this PR, even if I apply the patch to our server, still breaks at kernel connection.

anderbubble:
Found the existing issue #345. I'll move my discussion there.

PR #441, opened 2016-02-25
Revert "Do not consider `@` character url-safe"

minrk:
Reverts jupyter/jupyterhub#440

@ is safe when using notebook ≥ 4.1

Not to merge, until resolving some discussion in #440 if there's more to it than outdated notebook versions.


rgbkrk:
Ok, cool. Thanks for resurrecting.

rgbkrk:
Based on #440, seems like we can merge and find out if the version bump helps them.

PR #442, opened 2016-02-25
avoid calling Spawner.poll during Spawner.start

### LINKED ISSUE #428: Spawner poll return value during spawn
This is a question about how `start_timeout` and `slow_spawn_timeout` work together.

In `BaseHandler.spawn_single_user()` `user.spawn()` is called which should complete with in `start_timeout` I think. On line 296 the `slow_spawn_timeout` is used, which is much shorter. If we reach the timeout then `user.spawner.poll()` is called and if it is not `None` an exception is raised.

My expectation would be that `user.spawner.poll()` returns something that isn't `None` while still spawning, as it isn't yet running. However this means you effectively only have `slow_spawn_timeout` seconds to be done spawning.

Should `poll()` already return `None` while still spawning?

minrk:
moves `spawn_pending` flag to only around start, not the HTTP wait.

Some Spawners may not know how to poll until start has finished (DockerSpawner). Let's not require that they do.

cc @betatim

https://github.com/jupyter/dockerspawner/pull/73

closes #428

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -296,16 +296,23 @@ class BaseHandler(RequestHandler):
             yield gen.with_timeout(timedelta(seconds=self.slow_spawn_timeout), f)
         except gen.TimeoutError:
             if user.spawn_pending:
+                # still in Spawner.start, which is taking a long time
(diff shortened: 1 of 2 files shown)


betatim:
Nice!

PR #443, opened 2016-02-25
catch exceptions in options_from_form

### LINKED ISSUE #439: Spawner custom form validation
Are there ideas for allowing form validation for spawners that have a custom form?

I was thinking of raising an exception in `options_from_form()` and moving the `try` up by one line in [SpawnHandler](https://github.com/jupyter/jupyterhub/blob/master/jupyterhub/handlers/pages.py#L97).

minrk:
Allows form validation to be implemented in options_from_form, as well as start.

closes #439


rgbkrk:
I wonder about granularity of the two types of exceptions that surface under this same `try`. This has the right intent, so it doesn't matter so much.

betatim:
:zap: as fast as lightning!

PR #445, opened 2016-03-01
Don't add users with spawn_pending to the proxy

### LINKED ISSUE #410: Proxy error on user spawn but not admin spawn.
In some cases, when our hub users click on My Server to start their server, the hub responds with an HTTP 503. When this happens, admins can click on "start server" for the user from the Admin page and "stop server" to stop. The user can then hit My Server without error.

When the user does see the 503, the routing table has an entry for them pointing at a spawned container that no longer exists. I don't know how relevant this is because it plays no part in the admin being able to start the server. The user's server spawned by the admin ends up with an updated entry in the routing table (new node/ip).

We're using [swarmspawner](https://github.com/data-8/jupyterhub-deploy/blob/master/roles/jupyterhub/files/swarmspawner.py).

minrk:
`check_routes` checks for missing routes for running users. This is meant for when the proxy has been relaunched outside the Hub.

If spawners are slow to start, it's possible for check_routes to fire in the middle of spawning, triggering addition of the user's server (which has no defined location yet) to the proxy before it's up. Further, if the spawning fails, the route will remain indefinitely (because it never should have been added in the first place), and the user will see 503 until their server is launched manually again.

Checking `spawn_pending` in user.running prevents this.

closes #410

### DIFF
--- a/jupyterhub/orm.py
+++ b/jupyterhub/orm.py
@@ -178,6 +178,9 @@ class Proxy(Base):
         self.log.info("Adding user %s to proxy %s => %s",
             user.name, user.proxy_path, user.server.host,
         )
+        if user.spawn_pending:
+            raise RuntimeError(
+                "User %s's spawn is pending, shouldn't be added to the proxy yet!", user.name)
         
         yield self.api_request(user.proxy_path,
             method='POST',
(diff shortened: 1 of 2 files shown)


rgbkrk:
Yay for fixing race conditions.

PR #452, opened 2016-03-04
remove redundant use_subdomains

minrk:
non-empty subdomain_host is enough

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -235,11 +235,18 @@ class JupyterHub(Application):
     )
     
     subdomain_host = Unicode('', config=True,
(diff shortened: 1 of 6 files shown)


willingc:
@minrk Is there ever a case where subdomain_host was set to something and then for whatever reason is later set to None? I'm trying to reason through the cases (None, ' ', something) for subdomain_host.

minrk:
No, there should only be two states:

- default (`''`)
- a string (subdomains are in use)

willingc:
Excellent @minrk. The code looked good. Thanks for clarifying.

willingc:
@minrk I'm going to go ahead and merge this since I think it will solve the Travis issue on #456

PR #460, opened 2016-03-07
Add lengths to all Unicode() columns

### LINKED ISSUE #447: Fails to start when using mysql as db backend
Ran into `VARCHAR requires a length on dialect mysql` when attempting to use it with MySQL. All the Unicode columns need a length field for MySQL to work.

yuvipanda:
- Otherwise does not work with MySQL
- Change JSONDict to be TEXT (Unbounded) rather than VARCHAR.
  This makes most sense, since you can't index these anyway.
- The 'ip' field in Server is set to 255, since that is the
  max allowed length of DNS entries.
- Most of the rest of the Unicodes have approximately high
  values that most people should not mostly run into
  (famous last words).

Fixes #447

### DIFF
--- a/jupyterhub/orm.py
+++ b/jupyterhub/orm.py
@@ -12,7 +12,7 @@ from tornado import gen
 from tornado.log import app_log
 from tornado.httpclient import HTTPRequest, AsyncHTTPClient
 
-from sqlalchemy.types import TypeDecorator, VARCHAR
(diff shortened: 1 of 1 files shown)


minrk:
Great! I just need to check that this doesn't trigger db-incompatibilities with existing databases, then I'll merge.

minrk:
Thanks! I think I'll try to cut 0.5 with the subdomain stuff today.

PR #463, opened 2016-03-07
Increase some test coverage

minrk:
Found a couple of small bugs along the way.


willingc:
Looks great. Thanks for adding more tests. Good catch to use mkdtemp.

PR #467, opened 2016-03-08
Call `add_user` more often

### LINKED ISSUE #466: Setup allowing any Github user access
Hi there,

Trying to use Jupyterhub for a talk I'm giving on Thursday. I've basically setup a server and want anyone who is in the room to be able to login to the server using their github credentials. Is this possible? It seems possible to automatically add system users but not github users. Having no whitelist allows any Github user to login but once they hit "*Start my Server*" they will get a 500 and I'll see the error `KeyError: 'getpwnam(): name not found: GITHUB_USERNAME'`

Here is my current config:

```
# Github Settings
c.GitHubOAuthenticator.client_id = os.environ['GITHUB_CLIENT_ID']
c.GitHubOAuthenticator.client_secret = os.environ['GITHUB_CLIENT_SECRET']
c.GitHubOAuthenticator.oauth_callback_url = os.environ['OAUTH_CALLBACK_URL']

# Authentication Setup
c.JupyterHub.authenticator_class = 'oauthenticator.LocalGitHubOAuthenticator'
c.LocalGitHubOAuthenticator.create_system_users = True

# Create empty whitelist and setup admin users
c.Authenticator.whitelist = whitelist = set()
c.Authenticator.admin_users = admin = set()
c.JupyterHub.admin_access = True
    
# Use userlist to determine administrators
with open(join(here, 'userlist')) as f:
    for line in f:
        if not line:
            continue
        parts = line.split()
        name = parts[0]
        if len(parts) > 1 and parts[1] == 'admin':
            admin.add(name)
```

I'd appreciate any feedback of how to approach achieving what I want.

minrk:
- Ensures add_user is called as part of startup *for all users*.
  This was previously only true for users not already in the db.
- Normalize usernames in whitelist and admin sets
- Call add_user on new users logged in when there is no whitelist.

closes #466

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -677,6 +677,7 @@ class JupyterHub(Application):
             self.authenticator.normalize_username(name)
             for name in self.authenticator.admin_users
         ]
+        self.authenticator.admin_users = set(admin_users) # force normalization
         for username in admin_users:
             if not self.authenticator.validate_username(username):
                 raise ValueError("username %r is not valid" % username)
@@ -704,6 +705,7 @@ class JupyterHub(Application):
             self.authenticator.normalize_username(name)
             for name in self.authenticator.whitelist
         ]
+        self.authenticator.whitelist = set(whitelist) # force normalization
         for username in whitelist:
             if not self.authenticator.validate_username(username):
                 raise ValueError("username %r is not valid" % username)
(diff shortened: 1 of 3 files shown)


minrk:
@ellisonbg I think this might solve some of the issues you have seen where sometimes the user-add action is triggered and sometimes it isn't.

Carreau:
I dont' see anything obviously wrong, but I'm not knowlegable enough in how the hub works.

PR #472, opened 2016-03-09
Use encoded URL when redirecting user notebooks

yuvipanda:
Otherwise it breaks for usernames that have url unsafe
characters.

### DIFF
--- a/jupyterhub/handlers/pages.py
+++ b/jupyterhub/handlers/pages.py
@@ -8,17 +8,17 @@ from tornado import web, gen
 from .. import orm
 from ..utils import admin_only, url_path_join
 from .base import BaseHandler
-from .login import LoginHandler
+from urllib.parse import quote
 
 
 class RootHandler(BaseHandler):
     """Render the Hub root page.
-    
+
     If logged in, redirects to:
-    
+
     - single-user server if running
     - hub home, otherwise
-    
+
     Otherwise, renders login page.
     """
     def get(self):
@@ -49,9 +49,9 @@ class HomeHandler(BaseHandler):
 
 class SpawnHandler(BaseHandler):
     """Handle spawning of single-user servers via form.
-    
+
     GET renders the form, POST handles form submission.
-    
+
     Only enabled when Spawner.options_form is defined.
     """
     def _render_form(self, message=''):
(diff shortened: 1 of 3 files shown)


jupyterhub/handlers/pages.py
```diff
@@ -75,9 +75,11 @@ def get(self):
             self.finish(self._render_form())
         else:
             # not running, no form. Trigger spawn.
-            url = url_path_join(self.base_url, 'user', user.name)
+            # Creating the URL manually since the server does not
+            # exist yet
+            url = url_path_join(self.base_url, 'user', quote(user.name))
```
minrk: This should be user.escaped_name

share/jupyter/hub/templates/home.html
```diff
@@ -10,7 +10,7 @@
       {% endif %}
       <a id="start" class="btn btn-lg btn-success"
       {% if user.running %}
-        href="{{base_url}}user/{{user.name}}/"
+        href="{{ user.server.base_url }}"
```
minrk: This should be `user.url`

minrk:
Thanks for digging into this! The only reason we don't use `user.url` everywhere is that it is retrieved from user.server, which is undefined unless the user is running. But there's no actual reason for this restriction (there may be, once we have multiple servers), so after this fix, I think I'll make it so `user.url` is always computable, and make sure we never build user URLs without it.

share/jupyter/hub/templates/home.html
yuvipanda: done!

On Wed, Mar 9, 2016 at 12:18 AM, Min RK <notifications@github.com> wrote:

> In share/jupyter/hub/templates/home.html
> <https://github.com/jupyter/jupyterhub/pull/472#discussion_r55484142>:
>
> > @@ -10,7 +10,7 @@
> >        {% endif %}
> >        <a id="start" class="btn btn-lg btn-success"
> >        {% if user.running %}
> > -        href="{{base_url}}user/{{user.name}}/"
> > +        href="{{ user.server.base_url }}"
>
> This should be user.url
>
> —
> <https://github.com/jupyter/jupyterhub/pull/472/files#r55484142>.
>



-- 
Yuvi Panda T
http://yuvi.in/blog

yuvipanda:
@minrk yw! And thanks for merging! It looks like the commit with the changes you suggested got lost somehow - maybe my network died before I pushed it, and hence the merge didn't have the changes you suggested. Want me to put that in as a different branch? But I see that you're changing it anyway with #474 so maybe it is moot :)

minrk:
I might have merged before you got to it. Sorry! I did the user.url change in #474, which should make this harder to mess up in the future.

PR #474, opened 2016-03-09
allow user.url to be accessed without the server running

minrk:
Reduces the number of different ways we need to build the same URLs.

follow-up on #472


Carreau:
Seem good to me.

minrk:
Thanks, @Carreau!

PR #477, opened 2016-03-11
Use latest version of Sphinx to fix RTD "Edit on GitHub"

willingc:
This PR fixes the issue found by @parente where RTD's "Edit on GitHub" was not correctly redirecting to markdown source and throwing a 404 instead. The cause was an underlying bug in Sphinx which has been fixed as 1.3.6.

I've updated our version of Sphinx to 1.3.6 or greater.


willingc:
Thanks @Carreau

PR #479, opened 2016-03-12
Make Spawner.env configurable

### LINKED ISSUE #476: Can not add additional items to environment of Spawner
```
c.Spawner.env.update({
    'wat': 'waaaaaaaaaaaat'
})
```

Seems to have no effect. Is this because env is just a `Dict` without `config=True` set in the Spawner base class?

minrk:
moves `_env_default` logic to `get_env`, so that `Spawner.env` can be safely configurable

closes #476

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -116,14 +116,9 @@ class Spawner(LoggingConfigurable):
     ], config=True,
         help="Whitelist of environment variables for the subprocess to inherit"
     )
-    env = Dict()
-    def _env_default(self):
-        env = {}
-        for key in self.env_keep:
-            if key in os.environ:
-                env[key] = os.environ[key]
-        env['JPY_API_TOKEN'] = self.api_token
-        return env
+    env = Dict(config=True,
+        help="Environment variables to load into the Spawner environment."
+    )
     
     cmd = Command(['jupyterhub-singleuser'], config=True,
         help="""The command used for starting notebooks."""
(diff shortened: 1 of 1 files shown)


yuvipanda:
I wonder if this requires downstream changes in other spawners (kubespawner overrides _env_default)?

ssanderson:
> I wonder if this requires downstream changes in other spawners (kubespawner overrides _env_default)?

Looks like a similar change would be needed in `Dockerspawner`:

```
    def _env_default(self):
        env = super(DockerSpawner, self)._env_default()
        env.update(dict(
            JPY_USER=self.user.name,
            JPY_COOKIE_NAME=self.user.server.cookie_name,
            JPY_BASE_URL=self.user.server.base_url,
            JPY_HUB_PREFIX=self.hub.server.base_url
        ))

        if self.hub_ip_connect:
           hub_api_url = self._public_hub_api_url()
        else:
           hub_api_url = self.hub.api_url
        env['JPY_HUB_API_URL'] = hub_api_url

        return env
```

jupyterhub/spawner.py
```diff
@@ -204,12 +199,19 @@ def clear_state(self):
...
+        """Return the environment dict to use for the Spawner.
+
+        This applies things like `env_keep`, anything defined in `Spawner.env`,
+        and adds the API token to the env.
+
         Use this to access the env in Spawner.start to allow extension in subclasses.
         """
-        return self.env.copy()
+        env = self.env.copy()
+        for key in self.env_keep:
```
ssanderson: It looks like this will overwrite what the user supplied if they configured an env_keep variable, or if they supplied JPY_API_TOKEN as part of their env.  Should we skip writing and/or warn and/or barf if we're overwriting user-configured values here?

minrk: I think reversing the priority should be fine - env_keep lowest, self.env highest, since this change makes `self.env` clearly and explicitly env-overrides.

minrk:
Perhaps a new traitlet for config-specified env overrides is a less disruptive change.

minrk:
Went with adding `Spawner.environment` configurable, and deprecating `Spawner.env` for backward-compatibility.

willingc:
@minrk Seems a reasonable approach. Merging. Thanks.

PR #480, opened 2016-03-14
Update the configuration section of docs

### LINKED ISSUE #458: change jupyterhub configuration params by command line
Was trying to change `c.Spawner.container_ip` jupyterhub param by `--Spawner.ip` command line argument, but seems that those are not the same. Would it be possible to make `c.Spawner.container_ip` controllable by jupyterhub invocation, so I could override without config file modifications?
(juptyerhub version 0.5.1)

I guess it is not the only parameter that is not accessible by command line args. 

PS am I right that configuration file params are of higher priority than command-line ones? (say, I cannot override `c.JupyterHub.authenticator_class` by `--JupyterHub.authenticator_class` command-line one?

willingc:
Closes #458

Reorders the configuration section to put configuration file before command-line arguments.

- Add subtitles for configuration file and command-line arguments

- Provide an example for using the command-line argument for configuration.

### DIFF
--- a/docs/source/getting-started.md
+++ b/docs/source/getting-started.md
@@ -64,23 +64,13 @@ The location of these files can be specified via configuration, discussed below.
 
 JupyterHub is configured in two ways:
 
-1. Command-line arguments
-2. Configuration files
+1. Configuration file
+2. Command-line arguments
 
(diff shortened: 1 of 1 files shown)


minrk:
Thanks!

PR #487, opened 2016-03-21
Fix 'failed login for None' message

minrk:
on failed login, get username from form data, not the guaranteed-None return value of authenticate

### DIFF
--- a/jupyterhub/handlers/login.py
+++ b/jupyterhub/handlers/login.py
@@ -78,7 +78,7 @@ class LoginHandler(BaseHandler):
             self.redirect(next_url)
             self.log.info("User logged in: %s", username)
         else:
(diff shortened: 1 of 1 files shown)


jupyterhub/handlers/login.py
```diff
@@ -78,7 +78,7 @@ def post(self):
             self.redirect(next_url)
             self.log.info("User logged in: %s", username)
         else:
-            self.log.debug("Failed login for %s", username)
+            self.log.debug("Failed login for %s", data.get('username', 'unkown user'))
```
willingc: Typo unknown

Otherwise, +1

minrk: Fixed, thanks!

Issue #486, opened 2016-03-20
Login -> Logout -> Login fails


vbraun:
I can login once but not twice. Jupyterhub 0.5.0 running under a single user with pam authentication
~~~~
$ ./jupyterhub --log-level=DEBUG
[D 2016-03-20 16:45:56.367 JupyterHub application:529] Looking for ./jupyterhub_config in None
[D 2016-03-20 16:45:56.367 JupyterHub application:549] Loaded config file: /home/vbraun-hub/JupyterHub/jupyterhub_config.py
[I 2016-03-20 16:45:56.369 JupyterHub app:558] Loading cookie_secret from /home/vbraun-hub/JupyterHub/jupyterhub_cookie_secret
[D 2016-03-20 16:45:56.369 JupyterHub app:623] Connecting to db: sqlite:///jupyterhub.sqlite
[W 2016-03-20 16:45:56.390 JupyterHub app:292] 
    Generating CONFIGPROXY_AUTH_TOKEN. Restarting the Hub will require restarting the proxy.
    Set CONFIGPROXY_AUTH_TOKEN env or JupyterHub.proxy_auth_token config to avoid this message.
    
[W 2016-03-20 16:45:56.393 JupyterHub app:685] No admin users, admin interface will be unavailable.
[W 2016-03-20 16:45:56.393 JupyterHub app:686] Add any administrative users to `c.Authenticator.admin_users` in config.
[I 2016-03-20 16:45:56.393 JupyterHub app:712] Not using whitelist. Any authenticated user will be allowed.
[D 2016-03-20 16:45:56.400 JupyterHub app:787] Loaded users: 
    vbraun-hub
[I 2016-03-20 16:45:56.405 JupyterHub app:1113] Hub API listening on http://127.0.0.1:8081/hub/
[I 2016-03-20 16:45:56.408 JupyterHub app:860] Starting proxy @ http://*:443/
[D 2016-03-20 16:45:56.408 JupyterHub app:861] Proxy cmd: ['configurable-http-proxy', '--ip', '', '--port', '443', '--api-ip', '127.0.0.1', '--api-port', '444', '--default-target', 'http://127.0.0.1:8081', '--ssl-key', 'letsencrypt/etc/live/vbraun.cc/privkey.pem', '--ssl-cert', 'letsencrypt/etc/live/vbraun.cc/cert.pem']
16:45:56.542 - info: [ConfigProxy] Proxying https://*:443 to http://127.0.0.1:8081
[...]

PR #490, opened 2016-03-23
Allow disabling PAM sessions

minrk:
it's often buggy and rarely necessary, so allow it to be disabled when it's causing problems.

It's still on by default for consistency, though maybe it shouldn't be.


willingc:
@minrk looks good. I'm going to merge.

In light of #486, it may be better to make this off by default with users actively opting in.

PR #495, opened 2016-03-29
Import warnings, used on line 215, not imported.

Carreau:


willingc:
Good catch.

Carreau:
Was caught by https://landscape.io/github/jupyter/jupyterhub

(still a few false positive)

willingc:
Nice! Landscape all the things @Carreau

PR #497, opened 2016-03-31
Allow environment config values to be callable

yuvipanda:
This allows deployments to configure environment variables
that need to be different for each user / container (such as
credentials for various services, etc).

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -220,10 +220,17 @@ class Spawner(LoggingConfigurable):
         for key in self.env_keep:
             if key in os.environ:
                 env[key] = os.environ[key]
-        
-        # config overrides
-        env.update(self.environment)
-        
+
+        # config overrides. If the value is a callable, it will be called with
+        # one parameter - the current spawner instance - and the return value
+        # will be assigned to the environment variable. This will be called at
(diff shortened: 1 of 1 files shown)


yuvipanda:
These aren't coroutines, so should be pretty quick (in my install, it's just doing a HMAC - see https://github.com/yuvipanda/paws/blob/master/hub/jupyterhub_config.py#L29)

minrk:
Something missing? This PR appears to be only whitespace changes.

yuvipanda:
That's really weird - my first commit (
https://github.com/jupyter/jupyterhub/pull/497/commits/705ff787153290bad55dc4f888812451cfdb27b4)
has the changes, but the second commit somehow seems to have undone them?
I'll check and redo.

On Thu, Mar 31, 2016 at 10:15 PM, Min RK <notifications@github.com> wrote:

> Something missing? This PR appears to be only whitespace changes.
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/jupyter/jupyterhub/pull/497#issuecomment-204252028>
>



-- 
Yuvi Panda T
http://yuvi.in/blog

yuvipanda:
@minrk fixed.

PR #515, opened 2016-04-14
Post handler for  requesting authorization tokens; authenticated via form.

dominicfollett:


minrk:
Great! Can you add a test that exercises this?

dominicfollett:
Sure thing!

minrk:
I don't see any problem with including this, just a bit of a challenge on my side of communicating that it will only work for a certain class of Authenticators. Thanks for the PR!

dominicfollett:
Could you explain how to run tests locally? I tried looking at the travisci, but to no avail. I just want to make sure they pass locally before pushing onto a PR.

PR #517, opened 2016-04-14
allow pre-loading API tokens from config

minrk:
This is the first small part of easing the pain of jupyterhub-associated services, so you shouldn't need to initialize JupyterHub before setting API tokens for services that need API access to the Hub.

cc @jhamrick, who I know has suffered from the need to generate tokens with `jupyterhub token`.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -352,6 +352,13 @@ class JupyterHub(Application):
     cookie_secret_file = Unicode('jupyterhub_cookie_secret',
         help="""File in which to store the cookie secret."""
     ).tag(config=True)
+    
+    api_tokens = Dict(Unicode(),
+        help="""Dict of token:username to be loaded into the database.
+        
+        Allows ahead-of-time generation of API tokens for use by services.
+        """
+    ).tag(config=True)
 
     authenticator_class = Type(PAMAuthenticator, Authenticator,
         help="""Class for authenticating users.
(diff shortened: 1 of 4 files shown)


jupyterhub/app.py
```diff
@@ -352,6 +352,13 @@ def _hub_prefix_changed(self, name, old, new):
     cookie_secret_file = Unicode('jupyterhub_cookie_secret',
         help="""File in which to store the cookie secret."""
     ).tag(config=True)
+    
+    api_tokens = Dict(Unicode(),
+        help="""Dict of token:username to be loaded into the database.
+        
+        Allows ahead-of-time generation of API tokens for use by services.
```
willingc: @minrk I'm not sure the following is an improvement but "ahead-of-time" and "use by services" together felt a little vague to me.

Allows administrator to add pre-generated user API tokens for use by services.

jupyterhub/app.py
```diff
@@ -794,6 +801,28 @@ def init_users(self):
...
+                raise ValueError("Token username %r is not valid" % username)
+            orm_token = orm.APIToken.find(db, token)
+            if orm_token is None:
+                user = orm.User.find(db, username)
+                if user is None:
+                    self.log.debug("Adding user %r to database", username)
+                    user = orm.User(name=username)
+                    db.add(user)
+                    db.commit()
+                self.log.info("Adding API token for %s", username)
```
willingc: If `orm.User` call succeeds and later `user.new_api_token` fails, are we in a state where the user has been committed to the database without a token?

willingc:
@minrk Reviewed. One question in review on database state. Otherwise, looks good.

jupyterhub/app.py
minrk: Thanks

jupyterhub/app.py
minrk: Yes. Updated to remove the user afterward.

minrk:
Comments should be addressed

willingc:
@minrk Looks good :+1: Thanks :D

PR #520, opened 2016-04-17
cookie_secret file must be base64

robnagler:
cookie_secret file is decoded by binascii.a2b_base64 so need to document it must be Base64. 

Added better doc for other values, and included description of "cookie_secret" parameter as well.

A couple of lines had trailing spaces, which my editor automatically deleted.

### DIFF
--- a/docs/source/getting-started.md
+++ b/docs/source/getting-started.md
@@ -97,7 +97,7 @@ for the full command line help.
 
 All configurable options are technically configurable on the command-line,
 even if some are really inconvenient to type. Just replace the desired option,
-c.Class.trait, with --Class.trait. For example, to configure 
+c.Class.trait, with --Class.trait. For example, to configure
 c.Spawner.notebook_dir = '~/assignments' from the command-line:
 
     jupyterhub --Spawner.notebook_dir='~/assignments'
@@ -190,7 +190,7 @@ Some cert files also contain the key, in which case only the cert is needed. It 
 these files be put in a secure location on your server, where they are not readable by regular
 users.
 
-Note: In certain cases, e.g. behind SSL termination in nginx, allowing no SSL 
+Note: In certain cases, e.g. behind SSL termination in nginx, allowing no SSL
 running on the hub may be desired. To run the Hub without SSL, you must opt
 in by configuring and confirming the `--no-ssl` option, added as of [version 0.5](./changelog.html).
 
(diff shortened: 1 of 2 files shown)


docs/source/getting-started.md
```diff
@@ -205,26 +205,36 @@ as follows:
 c.JupyterHub.cookie_secret_file = '/srv/jupyterhub/cookie_secret'
 ```
 
-The content of this file should be a long random string. An example would be to generate this
-file as:
+The content of this file should be a long random string encoded in MIME Base64. An example would be to generate thisfile as:
```
willingc: Typo. Missing space at end of line. ``this file``

docs/source/getting-started.md
```diff
@@ -205,26 +205,36 @@ as follows:
...
+openssl rand -base64 2048 > /srv/jupyterhub/cookie_secret
 ```
 
 In most deployments of JupyterHub, you should point this to a secure location on the file
 system, such as `/srv/jupyterhub/cookie_secret`. If the cookie secret file doesn't exist when
-the Hub starts, a new cookie secret is generated and stored in the file. The recommended
-permissions for the cookie secret file should be 600 (owner-only rw).
+the Hub starts, a new cookie secret is generated and stored in the file. The
+file must not be readable by group or other or the server won't start.
+The recommended -permissions for the cookie secret file should be 600 (owner-only rw).
```
willingc: Strike leading ``-`` before permissions.

docs/source/getting-started.md
```diff
@@ -205,26 +205,36 @@ as follows:
 c.JupyterHub.cookie_secret_file = '/srv/jupyterhub/cookie_secret'
 ```
 
-The content of this file should be a long random string. An example would be to generate this
-file as:
+The content of this file should be a long random string encoded in MIME Base64. An example would be to generate thisfile as:
 
 ```bash
-openssl rand -hex 1024 > /srv/jupyterhub/cookie_secret
+openssl rand -base64 2048 > /srv/jupyterhub/cookie_secret
```
willingc: @minrk Will you check the content changes in this PR please? If base64 is used, I think it would need to be preceded by ``import base64``.

willingc:
@robnagler Thanks for submitting this PR. I've made a few editing suggestions, and I have asked @minrk to review the technical changes as well.

docs/source/getting-started.md
Carreau: No, this is bash, not python, no need to import anything. 

Should we also check for whether the content of the file might be hex (set of chars in 0-9A-F?), in which case decoding it as base64 lower entropy. And if so print a warning ?

minrk:
👍 to the changes, pending @willingc's suggestions.

docs/source/getting-started.md
minrk: There's no need for that. If you generated a 1024b hex string, it's still valid base64, but much longer. It's still 1024b of randomness, just not the binary value you intended.

robnagler: >
> Should we also check for whether the content of the file might be hex (set
> of chars in 0-9A-F?), in which case decoding it as base64 lower entropy.
> And if so print a warning ?
>
I don't think you can do this now. A hex string is valid (although
unlikely) Base64 so you can't know the difference.

The big problem is that the file is overwritten when it cannot be parsed. I
think if the parser fails, the server should not start.

minrk: Feel free to add a strict error if the file cannot be parsed.

robnagler:
Done with #522 and #520.

minrk:
Thanks, @robnagler!

PR #521, opened 2016-04-18
Move docker onbuild step to jupyterhub-onbuild

### LINKED ISSUE #491: Tag Docker image as "onbuild" and create non-onbuild version
Building a container that inherits from the JupyterHub dockerfile fails if `jupyterhub_config.py` is not available in the top-level directory. That's because the Dockerfile includes an `ONBUILD` instruction that copies the file from that location. While I totally understand (and need!) the ability to provide custom configurations, requiring `jupyterhub_config.py` to be in a fixed location seems overly restrictive. I would prefer, for example, to be able to stick it in a `config/` directory.

The dockerfile best practices document [recommends](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#onbuild) that images including an `ONBUILD` instruction add `-onbuild` to the image tag so that they can be told apart from a non-ONBUILD version.

I understand that creating a separate, non-ONBUILD dockerfile could create a little confusion for users at first, but I think the complexity is minimal compared to the surprise when the current image fails (and total lack of understanding unless one goes and reads the parent Dockerfile). If there were two versions, users could inherit the Dockerfile in a few ways, depending on their preferences:

1. The current approach with the config file at the top level, just adding "-onbuild" to the image

    ```docker
    FROM jupyter/jupyterhub-onbuild
    ...
    ```
1. Explicitly copying the file from an arbitrary location to the expected location

    ```docker
    FROM jupyter/jupyterhub
    COPY /path/to/jupyterhub_config.py /srv/jupyterhub/jupyterhub_config.py
    ...
    ```
1. Explicitly copying the file from an arbitrary location to an arbitrary location

    ```docker
    FROM jupyter/jupyterhub
    COPY /path/to/jupyterhub_config.py /other/path/to/jupyterhub_config.py
    ...
    CMD ["jupyterhub", "-f", "/other/path/to/jupyterhub_config.py"]
    ```

minrk:
Removes onbuild from from jupyter/jupyterhub image, though it remains incomplete and will not run without a config file. It just no longer dictates how/when that config file is added.

closes #491


Dockerfile
```diff
@@ -36,12 +41,14 @@ ENV PATH=/opt/conda/bin:$PATH
...
 RUN npm install -g configurable-http-proxy && rm -rf ~/.npm
 
-ADD . /srv/jupyterhub
-WORKDIR /srv/jupyterhub/
+ADD . /src/jupyterhub
+WORKDIR /src/jupyterhub
 
 RUN python setup.py js && pip install . && \
-    rm -rf node_modules ~/.cache ~/.npm
+    rm -rf node_modules ~/.cache ~/.npm $PWD
```
willingc: I don't think ``node_modules`` is strictly necessary after adding ``$PWD``. I'm assuming you are wanting to clear out all cruft from the PWD. Any reason why? Container size?

willingc:
Circle CI is unhappy with the SHA1.

minrk:
Okay, I think this is all set, now.

willingc:
Looks good @minrk. One suggestion would be to edit the comments on the Dockerfile a bit to make it easier for users and hopefully decrease confusion/support questions.

```
# An incomplete base Docker image for running JupyterHub
#
# Add your configuration to create a complete derivative Docker image.
# 
# Include your configuration settings by starting with one of two options:
#
# Option 1:
#
# FROM jupyterhub/jupyterhub:latest
#
# And put your configuration file jupyterhub_config.py in /srv/jupyterhub/jupyterhub_config.py.
#
# Option 2:
#
# Or you can create your jupyterhub config and database on the host machine, and mount it with:
#
# docker run -v $PWD:/srv/jupyterhub -t jupyterhub/jupyterhub
#
# NOTE
# If you base on jupyterhub/jupyterhub-onbuild
# your jupyterhub_config.py will be added automatically
# from your docker directory.
```

minrk:
@willingc thanks, updated.

willingc:
Thank you @minrk!

minrk:
Thanks! Looks like my imaged upload didn't quite work. I'll work on that next.

PR #525, opened 2016-04-18
Bugfixing links in getting-started.md

kimchouard:
Changed extension: html --> md.

### DIFF
--- a/docs/source/getting-started.md
+++ b/docs/source/getting-started.md
@@ -419,9 +419,9 @@ jupyterhub -f /path/to/aboveconfig.py
 
 # Further reading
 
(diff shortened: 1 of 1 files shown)


minrk:
These are actually correct, and needed in the [rendered docs](https://jupyterhub.readthedocs.org/en/latest/getting-started.html).

minrk:
@willingc I don't suppose you have an idea for how to transform markdown links such that they work both pre- and post-render?

willingc:
@minrk Not yet. But I will continue to investigate. Thanks. 

@kimchouard If you find a solution, please let us know. We appreciate folks helping out and contributing improvements. :sunny:

PR #527, opened 2016-04-19
polish cookie-secret PR

minrk:
- fix a couple of typos
- use ValueError instead of assert to ensure error is raised even when Python optimizes-out asserts


willingc:
Thanks!

PR #528, opened 2016-04-19
exercise cookie secret loading in tests

### LINKED ISSUE #523: Simplified tests for configuration
It would be useful if there was an easy way to test config parsing for both conformance and deviance. Would like to be able to test the various modes for cookie_secret/_file.

minrk:
closes #523


willingc:
Tests :cake: :cookie: Thanks @minrk.

PR #531, opened 2016-04-20
[DOC] Note that extra_log_file only affects Hub's logs

minrk:
not single-user server logs, or anything else.


willingc:
:heart: for the user and future developers. Happy to merge.

PR #538, opened 2016-04-21
open_browser needs to be a Bool(config=True)

robnagler:
Fixes jupyterhub/dockerspawner#88. [docker-stacks/minimal-notebook sets open_browser](https://github.com/jupyter/docker-stacks/blob/3411997451309475812792cd7860e48fe2b4b4e9/minimal-notebook/jupyter_notebook_config.py#L13), and the same `jupyter_notebook_config.py` is used by jupyter/singleuser image.

### DIFF
--- a/scripts/jupyterhub-singleuser
+++ b/scripts/jupyterhub-singleuser
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 """Extend regular notebook server to be aware of multiuser things."""
 
 # Copyright (c) Jupyter Development Team.
@@ -47,7 +47,7 @@ class JupyterHubLoginHandler(LoginHandler):
     @staticmethod
     def login_available(settings):
         return True
-    
+
     @staticmethod
     def verify_token(self, cookie_name, encrypted_cookie):
         """method for token verification"""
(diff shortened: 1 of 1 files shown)


scripts/jupyterhub-singleuser
```diff
@@ -171,7 +171,7 @@ class SingleUserNotebookApp(NotebookApp):
     hub_api_url = Unicode(config=True)
     aliases = aliases
     flags = flags
-    open_browser = False
+    open_browser = Bool(False, config=True, help='Here for backwards compatible reasons. Leave as False')
```
willingc: @robnagler Perhaps on the help phrase: `help='Leave parameter set as False. Only change to maintain backwards compatibility.'

minrk:
I don't think this is the right fix. Instead of making it a Bool (it should *not* be configurable, because this is never the right thing to do), the `open_browser = False` config should be removed from the single-user image.

The warning is harmless, and just says that the config is being ignored, as it should be.

minrk:
Alternately, we could make it an enum with a single choice:

```python
open_browser = Enum([False], config=True, help="not really configurable, here to avoid warnings about ignored config")
```

robnagler:
+1 for Enum

willingc:
Agreed @robnagler. I would use Min's suggestion for Enum.

PR #542, opened 2016-04-22
Add troubleshooting documentation for 500 issue

### LINKED ISSUE #537: FAQ/troubleshooting item for 500 on single-user server
Recording troubleshooting info from gitter debugging with @robnagler today. This should probably go in docs somewhere, but I'm not sure where (ideas, @willingc?). I'm writing it down here before I forget.

If you get a 500 error after spawning your single-user server (on the URL `/user/you/...`), it is often because your single-user server cannot check your cookies with the Hub.

There are two likely reasons for this:

1. The single-user server cannot connect to the Hub's API (networking configuration problems)
2. The single-user server cannot *authenticate* its requests (invalid token)

## Symptoms:

The main symptom is a failure to load *any* page served by the single-user server, met with a 500 error. This is typically the first page at `/user/you` after logging in or clicking "Start my server". When a single-user server receives a request, it makes an API request to the Hub to check if the cookie corresponds to the right user. This request is logged. If everything is working, it will look like this:

```
200 GET /hub/api/authorizations/cookie/jupyter-hub-token-name/[secret] (@10.0.1.4) 6.10ms
```

You should expect one of these messages in the Hub log when you first visit your single-user server. If you don't see one at all, that may mean that your single-user server isn't connecting to your Hub.

If it's a token problem, you will see a 403 (forbidden) like this:

```
403 GET /hub/api/authorizations/cookie/jupyter-hub-token-name/[secret] (@10.0.1.4) 4.14ms
```

You can also check the logs of the single-user server, which may have more detailed information.

### No authorization request
 
If no API request is making it to the server, this is typically due to a network configuration issue. Often, this happens when the Hub is only listening on 127.0.0.1 (default) and the single-user servers are not on the same 'machine' (can be physically remote, or in a docker container or VM). The fix for this case is to make sure that `c.JupyterHub.hub_ip` is an address that all single-user servers can connect to, e.g.:

```python
c.JupyterHub.hub_ip = '10.0.0.1'
```

### 403 GET /hub/api/authorizations/cookie
If you are getting a 403 error, this is likely caused by the API token for the single-user server being invalid. The most common cause of this is resetting the JupyterHub database (either removing jupyterhub.sqlite or some other action), but leaving single-user servers running. This happens most frequently when using DockerSpawner, because the default behavior is to stop/start containers, rather than destroying and recreating them every time. This means that the same API token is used by the server for its whole life, until the container is rebuilt.

The fix for the docker case is to remove any docker containers seeing this issue (typicaly all containers created before a certain point in time):

    docker rm -f jupyter-name

After this, when you start your server next via JupyterHub, it will build a new container.   If this was the issue, you should see your server again.

willingc:
Closes #537.

### DIFF
--- a/docs/source/troubleshooting.md
+++ b/docs/source/troubleshooting.md
@@ -2,10 +2,96 @@
 
 This document is under active development.
 
-## Networking
+When troubleshooting, you may see unexpected behaviors or receive an error
+message. These two lists provide links to identifying the cause of the
+problem and how to resolve it.
 
-If JupyterHub proxy fails to start:
+## Behavior problems
+- JupyterHub proxy fails to start
+
+## Errors
+- 500 error after spawning a single-user server
+
+----
+
+## If JupyterHub proxy fails to start:
 
 - check if the JupyterHub IP configuration setting is
   ``c.JupyterHub.ip = '*'``; if it is, try ``c.JupyterHub.ip = ''``
 - Try starting with ``jupyterhub --ip=0.0.0.0``
+
+----
+
+## I am receiving a 500 error after spawning my single-user server (on the
+URL `/user/you/...`).
+
+This is often because your single-user server cannot check your cookies with
+the Hub.
+
+There are two likely reasons for this:
+
+1. The single-user server cannot connect to the Hub's API (networking
+   configuration problems)
+2. The single-user server cannot *authenticate* its requests (invalid token)
+
+### Symptoms:
+
+The main symptom is a failure to load *any* page served by the single-user
+server, met with a 500 error. This is typically the first page at `/user/you`
+after logging in or clicking "Start my server". When a single-user server
+receives a request, it makes an API request to the Hub to check if the cookie
+corresponds to the right user. This request is logged.
+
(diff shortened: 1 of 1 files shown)


docs/source/troubleshooting.md
```diff
@@ -2,10 +2,98 @@
...
+When troubleshooting, you may see unexpected behaviors or receive an error
+message. These two lists provide links to identifying the cause of the
+problem and how to resolve it.
 
-If JupyterHub proxy fails to start:
+## Behavior problems
+- [JupyterHub proxy fails to start](troubleshooting.md#jupyterhub-proxy-fails-to-start)
+
+## Errors
+- [500 error after spawning a single-user server](troubleshooting.md#500-error-after-spawning-my-single-user-server)
```
minrk: For in-document links, I think you can use just the hash `#500...`, and it should work within the page (both on GitHub and RTD).

willingc: Sounds good. I'll update it.

minrk:
Thanks!

PR #543, opened 2016-04-22
Allow jupyterhub-singleuser to run on python 2 install

robnagler:
scripts/jupyterhub-singleuser works with python 2 so the #! line doesn't need to refer to a specific python. In our environment, we don't have python 3 installed.


minrk:
While it is technically possible to run the server as Python 2, it's undocumented, and that support will probably not be around for too much longer (A Python 3 server works happily with a Python 2 kernel). So this PR is fine, but I would caution you not to rely on it if you can avoid it.

PR #549, opened 2016-04-26
Make statsd an optional dependency

minrk:
only import it if it's used

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -12,7 +12,6 @@ import signal
 import socket
 import sys
 import threading
-import statsd
 from datetime import datetime
 from distutils.version import LooseVersion as V
 from getpass import getuser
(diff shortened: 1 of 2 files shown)


willingc:
A simple question but I'm going to ask it since I'm not sure where it's happening: When and where does the statsd package get installed into the user's environment?

rgbkrk:
With this, the operator needs to install it directly (no option in setup.py currently).

/cc @yuvipanda

minrk:
It would be a separate step, along the lines of:

## Monitoring JupyterHub with statsd

[statsd](https://github.com/etsy/statsd) is a tool for collecting simple stats. With JupyterHub 0.6, you can send info about JupyterHub's operations to statsd

First, install the statsd Python client:

    pip3 install statsd

Second, tell JupyterHub to send stats to statsd

```python
# jupyterhub_config.py
c.JupyterHub.statsd_host = '127.0.0.1'
```

willingc:
@minrk Great explanation which I will add to the JupyterHub docs. :sunny:

yuvipanda:
+1 to @minrk's documentation :D

On Wed, Apr 27, 2016 at 3:28 AM, Carol Willing <notifications@github.com>
wrote:

> @minrk <https://github.com/minrk> Great explanation which I will add to
> the JupyterHub docs. [image: :sunny:]
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/jupyterhub/jupyterhub/pull/549#issuecomment-215042833>
>



-- 
Yuvi Panda T
http://yuvi.in/blog

PR #550, opened 2016-04-26
Fix docs typo for Spawner.disable_user_config

daradib:


willingc:
Thanks @daradib :cookie: This will help a future user.

PR #554, opened 2016-04-28
Add HubAuth for authenticating services with JupyterHub

minrk:
- HubAuth implements request to identify users with the Hub
- HubAuthenticated is a mixin for tornado handlers

todo:

- [x] test HubAuthenticated with real tornado handler


minrk:
@jhamrick I've finally got time to get back to finishing this up. Do you think it's useful to write a base HubAuth that just uses requests and doesn't depend on tornado, so you can use it in non-tornado applications? Or is tornado-only sufficiently useful?

jhamrick:
Hmm, good question. I'm fine with requiring tornado since nbgrader uses it, but it might be useful for other people writing services if it were in requests so they can write their service in whatever. Maybe since you've already started writing it in tornado, the best thing to do would be just write it in a way such that it's easy to swap out tornado for requests in the future if someone requests that?

Also, skimming through the code -- it might be nice if there were also a helper function for adding new routes to the proxy. I know that's not auth itself, but it is related to the overall goal which is making it easier for services to interface with JupyterHub.

minrk:
@jhamrick do you need to add more routes at runtime, or is it enough to declare the routes in config?

My plan for services like nbviewer is two pieces:

1. 'declare' the service in Hub config:
  - name
  - process start info (if started by the Hub)
  - API key
  - proxy route(s)
2. HubAuth (etc.) utility for implementing auth with JupyterHub

An alternate route is to do it *all* in the service extensions, which REST APIs for registering, routes, etc. But that means there's more work to do when writing a service, I think, especially if the service doesn't use tornado.

minrk:
Okay, I think this is complete for a first round. We've got:

- HubAuth, a class that implements the authentication request, which has some tornado code in it (tornado HTTPErrors raised), but that can be used outside tornado
- HubAuthenticated, a mixin for tornado handlers

Things I want to do, but not in this PR:

- use this class in jupyterhub-singleuser
- require a bit less boilerplate to integrate HubAuth, by setting up some environment variables and defaults

PR #448, opened 2016-03-02
Redirect requests to logged in user

daradib:
If a user, alice, requests /user/bob/notebooks/mynotebook.ipynb,
redirect her to /user/alice/notebooks/mynotebook.ipynb.
Currently, such requests get stuck in a redirect loop because
the request will be redirected to login page with a next parameter
that when followed is again redirected.

When notebook_dir is consistent across users, this will allow
users to share notebook URLs. Fixes #424.

PR #560, opened 2016-05-05
Process single-user server redirects

daradib:
Follow-up to #448.

If single-user notebook is running, it will redirect other
users to hub root with next argument, which was previously
ignored.

### DIFF
--- a/jupyterhub/handlers/pages.py
+++ b/jupyterhub/handlers/pages.py
@@ -16,6 +16,9 @@ from .base import BaseHandler
 class RootHandler(BaseHandler):
     """Render the Hub root page.
 
+    If next argument is passed by single-user server,
+    redirect to base_url + single-user page.
+
     If logged in, redirects to:
 
     - single-user server if running
@@ -24,6 +27,11 @@ class RootHandler(BaseHandler):
     Otherwise, renders login page.
     """
     def get(self):
+        next_url = self.get_argument('next', '')
+        if next_url and next_url.startswith('/user/'):
+            # Redirect /hub/?next=... requests from single-user servers.
+            url = url_path_join(self.hub.server.base_url, next_url.lstrip('/'))
+            self.redirect(url)
         user = self.get_current_user()
         if user:
             if user.running:


daradib:
@minrk If you think this is an acceptable solution to #448, I'll add a test.

jupyterhub/handlers/pages.py
```diff
@@ -24,6 +27,11 @@ class RootHandler(BaseHandler):
     Otherwise, renders login page.
     """
     def get(self):
+        next_url = self.get_argument('next', '')
+        if next_url and next_url.startswith('/user/'):
```
minrk: `startswith('/user/')` won't handle JupyterHub on a prefix. I think this needs to be `url_path_join(self.base_url, 'user') + '/'`

minrk:
I made a comment in-line about the URL prefix, but I think this is a fine approach, if you want to go ahead with the test.

akaihola:
We have the same issue, and I just verified that this patch fixes redirection.

My test setup is with dockerspawner and a custom authenticator which accepts all logins.

Without the patch, if user `receiver` opens the link `/user/sender/notebooks/foo.ipynb`, redirection goes to `/user/receiver/tree` if the Docker container for `sender` has been started. With the patch, this works correctly.

jupyterhub/handlers/pages.py
akaihola: @minrk, I tested replacing `startswith('/user/')` by `startswith(url_path_join(self.base_url, 'user') + '/')` in @daradib's patch, and it works correctly if I don't customize the base URL.

However, if I set:

    c.JupyterHub.base_url = '/base/'

and leave this as it as:

    c.JupyterHub.hub_prefix = '/hub/'

and log in as `receiver` and try to open the link `/base/user/sender/notebooks/foo.ipynb` while sender of the link is logged in, I'm redirected to `/base/hub/base/user/sender/notebooks/foo.ipynb` (which doesn't work).

I also tried setting

    c.JupyterHub.hub_prefix = '/base/hub/'

But it made no difference.

akaihola: This seems to work:
```
    def get(self):
        next_url = self.get_argument('next', '')
        if next_url and next_url.startswith(url_path_join(self.base_url, 'user') + '/'):
            # Redirect /hub/?next=... requests from single-user servers.
            url = url_path_join(self.hub.server.base_url, next_url[len(self.base_url):].lstrip('/'))
            self.redirect(url)
```

So on the second last line I strip the base URL from the next URL before tacking it onto the hub base URL.

daradib:
Taken care of by #613 thanks @minrk @willingc

PR #583, opened 2016-05-20
Minor typo fix

JeanLescut:
"npm install -g configurable-http-proxy" was written twice : in the dependecies section and in the Installation section

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,6 @@ Notes on the `pip` command used in the installation directions below:
 
 JupyterHub can be installed with pip, and the proxy with npm:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,7 +54,6 @@ Notes on the `pip` command used in the installation directions below:
 
 JupyterHub can be installed with pip, and the proxy with npm:
```
minrk: This line should be updated, since it refers to npm which is no longer used.

minrk:
Looks like another fix has been merged that removes the duplicate entry in another way. Thanks!

PR #584, opened 2016-05-24
Edit README based on Write The Docs tips

willingc:
Most of these changes are organization, wording, and structure to improve user clarity. In general, this PR reorganizes the existing content with a few changes in explanation which can be seen in the rich diff.


rgbkrk:
Looking good!

PR #585, opened 2016-05-25
Fix RTD links to PDF

willingc:
For some reason, RTD PDF files are served from readthedocs.org instead of the newer readthedocs.io


willingc:
Merging. Issue filed with readthedocs https://github.com/rtfd/readthedocs.org/issues/2224

PR #597, opened 2016-05-31
use HubAuth in single-user server

minrk:
This means running the singleuser server now requires the JupyterHub package, it is no longer a standalone script. This, in turn, means that single-user servers must be run with Python 3.

cc @ssanderson who had asked about if/when single-user servers would require Python 3.


rgbkrk:
I'm certainly in favor.

PR #601, opened 2016-06-03
[WIP][services] add Authenticators.api_only_admins

minrk:
Adds notion of api-only admin users. This is part of services, and enables
services to get an API token that doesn't correspond to a 'real' user.

API-only users:

- are always admins
- cannot have servers
- do not have to exist on the system or whatever service the Authenticator wraps.

Alternately, we could add the notion of dedicated *services*, distinguishing them from users.
I will investigate this prior to committing to api_only users, but it might be a somewhat more complicated database change.


minrk:
Closing in favor of #604

PR #608, opened 2016-06-12
Add a warning on login if page not over ssl.

Carreau:
The --no-ssl option in the backend make sens, but still too many
deployment are not over SSL because they underestimate / do not
understand the risks.

--- 
screen shot 2016-06-12 at 13 20 35

On secured connexions there might be a quick flash of the yellow box  before it get removed, the alternative is to hide it by default, and show it on non-https, but seem more complicated.

### DIFF
--- a/share/jupyter/hub/static/less/login.less
+++ b/share/jupyter/hub/static/less/login.less
@@ -1,6 +1,11 @@
 #login-main {
     display: table;
     height: 80vh;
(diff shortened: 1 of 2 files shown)


minrk:
I'll fix it to be hidden until shown. We don't want to flash a warning at people, which is very unpleasant.

PR #609, opened 2016-06-13
hide http warning until it's relevant

minrk:
avoids flash of invalid warning when everything is correct

finish #608


Carreau:
Thanks, I was unsure about hiding by default in case the JS at the end is never reached. 
I also searched if css could have rules that would depend on  protocol, but didn't found a thing.

PR #612, opened 2016-06-16
fix links to authenticator/spawner pages from howitworks

### LINKED ISSUE #611: Docs: Broken link
http://jupyterhub.readthedocs.io/en/latest/howitworks.html

`More info on custom Spawners.` at the end is broken.

minrk:
closes #611


willingc:
Thanks @minrk. Good catch @Carreau.

PR #613, opened 2016-06-16
Finish up cross-user redirects

minrk:
adds and tests fixes discussed in #560


willingc:
@minrk Finish up looks good.

Thanks @daradib @minrk for the work 👍

PR #627, opened 2016-07-01
allow running alembic with `python -m jupyterhub.dbutil`

minrk:
since we only have generated alembic.ini, present a command that generates one and uses it.

enables generating new revisions with:

    python -m jupyterhub.dbutil revision -m msg


willingc:
Looks good @minrk.

PR #634, opened 2016-07-08
cleanup servers, api tokens after spawner shutdown

minrk:
prevents growing table of unused servers and tokens

closes jupyterhub/dockerspawner#104

### DIFF
--- a/jupyterhub/orm.py
+++ b/jupyterhub/orm.py
@@ -304,7 +304,7 @@ class User(Base):
     id = Column(Integer, primary_key=True, autoincrement=True)
     name = Column(Unicode(1023))
     # should we allow multiple servers per user?
-    _server_id = Column(Integer, ForeignKey('servers.id'))
+    _server_id = Column(Integer, ForeignKey('servers.id', ondelete="SET NULL"))
     server = relationship(Server, primaryjoin=_server_id == Server.id)
     admin = Column(Boolean, default=False)
     last_activity = Column(DateTime, default=datetime.utcnow)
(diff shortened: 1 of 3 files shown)


brblord:
Using MySQL as the DB backend, this update triggers an intergrity error:

> (1451, 'Cannot delete or update a parent row: a foreign key constraint fails (`jupyterhub`.`users`, CONSTRAINT `users_ibfk_1` FOREIGN KEY (`_server_id`) REFERENCES `servers` (`id`))') [SQL: 'DELETE FROM servers WHERE servers.id = %s'] [parameters: (191,)

It would seem the foreign key needs to be set on the servers table, not the user table?

brblord:
I updated an existing installation, but the tables were originally built by sqlalchemy.  I've replaced the sqlite database with mysql, and I believe that is the cause.  In orm.py, a foreign key is specified in the User class on servers.id, so MySQL is rejecting the delete of the server when server_id is referenced by a user row.  MySQL foreign keys default on delete behavior to RESTRICT.  It appears that sqlite defaults on delete to NO ACTION.  I believe if the ondelete parameter is set to "SET NULL" in the User class in orm.py, the MySQL schema should accept the query.

minrk:
I added SET NULL, but I'm firing a bit blind without the ability to test it. Can you check if it works for you?

brblord:
Works great here, thank you for adding that so quickly.  I noticed you added the ondelete="SET NULL" on the user_id in APIToken as well.  Since in this scenario the user table is the parent, there won't be any integrity errors when deleting the tokens, and you may possibly want to set it to ondelete="CASCADE", as API Tokens should be removed if a user is deleted.

minrk:
Thanks, switched it to cascade.

PR #635, opened 2016-07-09
use traitlets 4.2 API in singleuser script

minrk:
fixes a compatibility issue with notebook master


willingc:
I restarted Travis and noticed that there was a segfault when running a test:
screen shot 2016-07-21 at 5 07 43 am

PR #639, opened 2016-07-13
Correct Spawner.start typo

ryanlovett:
As documented at https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/spawner.py#L103


minrk:
Thanks!

PR #650, opened 2016-07-22
return (ip, port) from Spawner.start

minrk:
removes the need for Spawners to set db state themselves in most cases

Should be backward-compatible with warnings.

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -46,6 +46,9 @@ class Spawner(LoggingConfigurable):
     ip = Unicode('127.0.0.1',
         help="The IP address (or hostname) the single-user server should listen on"
     ).tag(config=True)
+    port = Integer(0,
+        help="The port for single-user servers to listen on. New in version 0.7."
+    )
     start_timeout = Integer(60,
         help="""Timeout (in seconds) before giving up on the spawner.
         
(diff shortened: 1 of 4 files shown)


jupyterhub/spawner.py
```diff
@@ -46,6 +46,9 @@ class Spawner(LoggingConfigurable):
     ip = Unicode('127.0.0.1',
         help="The IP address (or hostname) the single-user server should listen on"
     ).tag(config=True)
+    port = Integer(0,
+        help="The port for single-user servers to listen on. New in 0.7."
```
willingc: Might be helpful for end user to see: ``New in version 0.7.``

jupyterhub/spawner.py
```diff
@@ -254,6 +256,13 @@ def get_args(self):
             ]
         if self.ip:
             args.append('--ip=%s' % self.ip)
+
+        if self.port:
+            args.append('--port=%i' % self.port)
+        elif self.user.server.port:
+            self.log.warning("Setting port from user.server is deprecated")
```
willingc: Do we want to specify in which version re: deprecation?

willingc:
@minrk I'm glad kicking travis helped last night. I fell asleep before it finished or I would have merged. 😉

minrk:
Thanks, @willingc!

Issue #643, opened 2016-07-15
Proxy redirects static|images|js|css to /hub/...


mpekalski:
I get a lot of 404, after redirects (302), and in the result the notebook does not work. The initial JupyterHub page works, I can also login using my system credentials. What I see in the log is basically

```
[I 2016-07-15 14:37:28.492 JupyterHub log:100] 302 GET /static/base/images/logo.png?v=7c4597ba713d804995e8f8dad448a397 (@127.0.0.1) 5.97ms
[W 2016-07-15 14:37:28.496 JupyterHub log:100] 404 GET /hub/static/base/images/logo.png?v=7c4597ba713d804995e8f8dad448a397 (@127.0.0.1) 2.50ms
14:37:28.528 - debug: [ConfigProxy] PROXY WEB /static/base/images/logo.png?v=7c4597ba713d804995e8f8dad448a397 to http://172.17.42.1:8081
```

When I go to directly the address (http://127.0.0.1:32778) the proxy should redirect me to everything works fine.
`14:40:48.503 - debug: [ConfigProxy] PROXY WEB /user/marcin to http://127.0.0.1:32778
`

When I go through JupyterHub (http://127.0.0.1:8000) I get a message

> 
Jupyter Notebook
404 : Not Found
You are requesting a page that does not exist!

The image I am using comes from
`docker pull jupyter/all-spark-notebook:latest`

On the host machine I have

```
marcin@vb:~$ npm --version
1.4.21
marcin@vb:~$ docker --version
Docker version 1.6.2, build 7c8fca2
marcin@vb:~$ jupyter --version
4.1.0
marcin@vb:~$ which -a jupyterhub
/usr/local/bin/jupyterhub
marcin@vb:~$ configurable-http-proxy --version
1.2.0
marcin@vb:~$ node --version
v0.10.25

```

and the docker image
```
jovyan@vb:~/work$ jupyter notebook --version
4.2.1
jovyan@vb:~/work$ jupyter --version
4.1.0
```

mpekalski:
I also tried different docker image, systemuser as described https://recordnotfound.com/dockerspawner-jupyterhub-119183, but I got exactly the same problem.

mpekalski:
I tried downgrading JupyterHub to 0.6.0 or 0.5.0 and it did not help.

mpekalski:
[...]

PR #651, opened 2016-07-22
more informative error if single-user server can't connect to Hub for auth

minrk:
error message points to hub_ip setting if Hub doesn't appear to be accessible at 127.0.0.1

cf #643


willingc:
Looks good to me. Thanks @minrk.

PR #655, opened 2016-07-24
Add links to REST API docs

willingc:


minrk:
@willingc I added a build to the docs in #656, so it will be included in RTD as well for the appropriate version, as well as a bit of info about how to use it.

willingc:
@minrk Excellent! Will review in a bit.

PR #656, opened 2016-07-25
Add REST API to docs

minrk:
Includes example API request and generated swagger docs, built on RTD.

Includes local build by bootprint-openapi, even though it's not as nice as petstore. Due the that, link to petstore as well.


willingc:
@minrk Love this! Happy to merge.

PR #662, opened 2016-07-25
Fix typo regarding user's interactions with PATH

mwmarkland:
Not a big deal, but I found it while reading the docs.


willingc:
@mwmarkland Thanks for taking the time to fix this. We appreciate it. :cookie:

Issue #665, opened 2016-07-28
Proxy does not stay open after exiting hub


vilhelmen:
Not sure if it's my setup (CentOS 7, JupyterHub 6.1), but when shutting down the hub (via SIGINT), the proxy is always terminated, regardless what the value of c.JupyterHub.cleanup_proxy is. However, if you shut the hub down through the web interface, the proxy will stay up if you uncheck the box for it.

I poked through the code and verified that cleanup_proxy was false in both cases, so I'm not sure what would cause the process to be killed. I did some light googling and it was suggested that the child process may be receiving the hub's SIGINT as well. Adding "preexec_fn=os.setpgrp" to the Popen call, as shown here http://stackoverflow.com/a/16928558/6521809 fixed the issue.

minrk:
@vilhelmen adding preexec sounds like the right thing to do. Do you want to send a PR?

vilhelmen:
Can do!

guoxiangyang12:
when i want to re-login after logout,there will be a mistake "[PAM Error 7] Authentication failure  ",why?

PR #667, opened 2016-07-28
Proxy will no longer recieve Hub's SIGINT

vilhelmen:
Fix for #665

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1026,7 +1026,7 @@ class JupyterHub(Application):
         self.log.info("Starting proxy @ %s", self.proxy.public_server.bind_url)
         self.log.debug("Proxy cmd: %s", cmd)
         try:
-            self.proxy_process = Popen(cmd, env=env)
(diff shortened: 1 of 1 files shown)


jupyterhub/app.py
```diff
@@ -1026,7 +1026,7 @@ def start_proxy(self):
         self.log.info("Starting proxy @ %s", self.proxy.public_server.bind_url)
         self.log.debug("Proxy cmd: %s", cmd)
         try:
-            self.proxy_process = Popen(cmd, env=env)
+            self.proxy_process = Popen(cmd, env=env, preexec_fn=os.setpgrp)
```
willingc: @vilhelmen Thanks for the PR! I agree preexec is a reasonable approach. 

@minrk, I'm wondering if `os.setpgip` may be a more general solution that `os.setpgrp`. The POSIX standard has marked `setpgrp` as obsolete http://pubs.opengroup.org/onlinepubs/9699919799/

minrk:
Good catch, thanks @willingc! @vilhelmen can you update it to use `start_new_session=True` instead?

minrk:
Thanks!

PR #675, opened 2016-08-01
Add /api/ and /api/info endpoints

minrk:
`/api/` is not authenticated, and just reports JupyterHub's version for now.

`/api/info` is admin-only, and reports more detailed info about Python, authenticators/spawners in use, etc.

Added missing parameters in docs for `/api/shutdown` as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ node_modules
 build
 dist
 docs/_build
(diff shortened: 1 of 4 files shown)


rgbkrk:
Awesome. Happy to see this kick off.

PR #689, opened 2016-08-15
log errors adding users already in db

### LINKED ISSUE #687: `KeyError("User %k does not not exist.")` when Unix user has been deleted.
JupyterHub version: 0.6.1
OS: Debian Jessie

We had followed this wiki to setup JupyterHub to run as a service on Debian Jessie, using the init.d script instead of a systemd service unit file: https://github.com/jupyterhub/jupyterhub/wiki/Run-jupyterhub-as-a-system-service

However, we are running into the exact same problem as this guy: https://serverfault.com/questions/794938/jupyterhub-error-on-start-keyerror-user-k-does-not-exist

His solution is to delete the jupyter.sqlite file.

In our case, we allow Unix account users to log into JupyterHub. But one of the users has left the company, so we removed his Unix account, and his username is showing up in the KeyError.

Oddly enough, this problem only happens if we try to start JupyterHub as a service. If we run JupyterHub at the terminal, the KeyError doesn't show up, and users can actually use JupyterHub.

Is it safe to just delete the jupyterhub.sqlite file while JupyterHub is not running?

minrk:
avoids removal of system users preventing Hub startup

closes #687

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -855,7 +855,13 @@ class JupyterHub(Application):
         # but changes to the whitelist can occur in the database,
         # and persist across sessions.
         for user in db.query(orm.User):
-            yield gen.maybe_future(self.authenticator.add_user(user))
+            try:
+                yield gen.maybe_future(self.authenticator.add_user(user))
+            except Exception:
+                # TODO: Review approach to synchronize whitelist with db
+                # known cause of the exception is a user who has already been removed from the system
+                # but the user still exists in the hub's user db
+                self.log.exception("Error adding user %r already in db", user.name)
         db.commit() # can add_user touch the db?
 
         # The whitelist set and the users in the db are now the same.


jupyterhub/app.py
```diff
@@ -855,7 +855,13 @@ def init_users(self):
         # but changes to the whitelist can occur in the database,
         # and persist across sessions.
         for user in db.query(orm.User):
-            yield gen.maybe_future(self.authenticator.add_user(user))
+            try:
+                yield gen.maybe_future(self.authenticator.add_user(user))
+            except Exception:
+                # TODO: known cause of this is system users that have been removed.
```
willingc: Perhaps a bit more clarity in the comment. Not sure if I have the details correct.

# TODO: Review approach to synchronize whitelist with db
# known cause of the exception is trying to add a user who has already been removed from the system
# but the user still exists in the hub's user db

minrk:
@willingc merged after updating the comment per your feedback. Thanks!

PR #699, opened 2016-08-20
Remove link to SwarmSpawner

jhamrick:
I have deprecated SwarmSpawner in my compmodels deployment, so the link to it here needs to be removed.


willingc:
Thanks @jhamrick

PR #705, opened 2016-08-27
WIP: implement services API

minrk:
cc @willingc @jhamrick @ellisonbg

- [x] make the basics work
- [x] update cull-idle example to run as a managed service
- [x] implement proxied endpoints
- [x] tests!
- [ ] example for formgrader service as external with proxied endpoint
- [x] API endpoint for services to retrieve their own info
- [x] work with subdomains

For now, ignore the implementation because it's mostly notes at this point. The main thing to look at is the services.md description of capabilities and how to specify services in config.

The gist:

- a service is something that talks to the Hub API, so it always needs an API token.
- a service can be managed (local process, started by GitHub) or 
- external: not managed by the Hub, only informing the Hub that it exists, up to you to start it, keep it running, etc.
- a service MAY have one HTTP(S) endpoint that will be proxied at `/services/service-name`
- managed services get environment variables that tell it about the context (The API token assigned, Hub URLs, etc.)
- managed services can generate a new API token on start
- external services must specify their API token in config

- cull-idle would logically be a managed script with no http endpoint, approximately:

```python
{
    'name': 'cull-idle',
    'command': '/path/to/cull-idle',
    'admin': True,
}
```

And that's it.

- formgrader would want an http endpoint. To run formgrader as an external service:

```python
{
    'name': 'formgrader',
    'api_token': 'added-to-db-on-start',
    'url': 'http://host:port',
}
```

where the Hub would assume that another system, such as docker/systemd/humans is keeping formgrader running.

- starting services is part of starting JupyterHub
- Services cannot be added at runtime
- Managed services will always be subprocesses of the Hub, if you want to run services in docker, etc. use external services. The Spawner abstractions will not be applied to managed services.
- All services will share the same cookie (no protection of services from each other)

### DIFF
--- a/docs/source/services.md
+++ b/docs/source/services.md
@@ -54,7 +54,7 @@ JUPYTERHUB_SERVICE_NAME: the name of the service ('cull-idle' above)
 JUPYTERHUB_API_TOKEN: API token assigned to the service
 JUPYTERHUB_API_URL: URL for the JupyterHub API (http://127.0.0.1:8080/hub/api)
 JUPYTERHUB_BASE_URL: Base URL of the Hub (https://mydomain[:port]/)
-JUPYTERHUB_SERVICE_PATH: Base path of this service (/service/cull-idle/)
+JUPYTERHUB_SERVICE_PREFIX: URL path prefix of this service (/services/cull-idle/)
 ```
 
 ## External services
--- a/docs/source/services.md
+++ b/docs/source/services.md
@@ -18,9 +18,11 @@ it will be added to the Hub's proxy at `/service/:service-name`.
 ## Managed services
 
 **Managed** services are services that the Hub starts and is responsible for.
-These can only be local subprocesses of the Hub.
+These can only be local subprocesses of the Hub,
+and the Hub will take care of starting these processes and restarting them if they stop.
+
 While there are similarities with notebook Spawners,
-there are no plans to support the same spawning abstractions as notebook
+there are no plans to support the same spawning abstractions as notebook.
 If you want to run these services in docker or other environments,
 you can register it as an external service below.
 
@@ -51,7 +53,7 @@ When the service starts, the Hub will pass the following environment variables:
 JUPYTERHUB_SERVICE_NAME: the name of the service ('cull-idle' above)
 JUPYTERHUB_API_TOKEN: API token assigned to the service
 JUPYTERHUB_API_URL: URL for the JupyterHub API (http://127.0.0.1:8080/hub/api)
-JUPYTERHUB_BASE_URL: Base URL of the Hub (https://mydomain/)
+JUPYTERHUB_BASE_URL: Base URL of the Hub (https://mydomain[:port]/)
 JUPYTERHUB_SERVICE_PATH: Base path of this service (/service/cull-idle/)
 ```
 
@@ -59,6 +61,8 @@ JUPYTERHUB_SERVICE_PATH: Base path of this service (/service/cull-idle/)
 
 You can use your own service management tools, such as docker or systemd, to manage JupyterHub services.
 These are not subprocesses of the Hub, and you must tell JupyterHub what API token the service is using to perform its API requests.
+Each service will need a unique API token because the Hub authenticates each API request,
+identifying the originating service or user.
 
 An example of an externally managed service with admin access and running its own web server:
 
--- a/docs/source/services.md
+++ b/docs/source/services.md
@@ -1,0 +1,84 @@
+# JupyterHub services
+
+JupyterHub 0.7 adds the notion of Services.
+A Service is a process that interacts with the Hub REST API.
+Services may perform actions such as shutting down user servers that have been idle for some time,
+or registering additional web servers that should also use the Hub's authentication
+and be served behind the Hub's proxy.
+
+There are two main characteristics of services:
+
+1. Is it **managed** by JupyterHub?
+2. Does it have a web server that should be added to the proxy?
+
+If a `command` is specified for launching the service, it will be started and managed by the Hub.
+If a `url` is specified for where the service runs its own webserver,
+it will be added to the Hub's proxy at `/service/:service-name`.
+
+## Managed services
+
+**Managed** services are services that the Hub starts and is responsible for.
+These can only be local subprocesses of the Hub.
+While there are similarities with notebook Spawners,
+there are no plans to support the same spawning abstractions as notebook
+If you want to run these services in docker or other environments,
+you can register it as an external service below.
+
+A managed service is characterized by the `command` specified for launching the service.
+
+
+```python
+c.JupyterHub.services = [
+    {
+        'name': 'cull-idle',
+        'admin': True,
+        'command': ['python', '/path/to/cull-idle.py', '--interval']
+    }
+]
+```
+
+In addition to `command`, managed services can take additional optional parameters,
+to describe the environment in which to start the process:
+
+- `env: dict` additional environment variables for the service.
+- `user: str` name of the user to run the server as if different from the Hub.
+    Requires Hub to be root.
+- `cwd: path` directory in which to run the service, if different from the Hub directory.
+
+When the service starts, the Hub will pass the following environment variables:
+
+```
+JUPYTERHUB_SERVICE_NAME: the name of the service ('cull-idle' above)
+JUPYTERHUB_API_TOKEN: API token assigned to the service
+JUPYTERHUB_API_URL: URL for the JupyterHub API (http://127.0.0.1:8080/hub/api)
+JUPYTERHUB_BASE_URL: Base URL of the Hub (https://mydomain/)
+JUPYTERHUB_SERVICE_PATH: Base path of this service (/service/cull-idle/)
+```
+
+## External services
+
+You can use your own service management tools, such as docker or systemd, to manage JupyterHub services.
+These are not subprocesses of the Hub, and you must tell JupyterHub what API token the service is using to perform its API requests.
+
+An example of an externally managed service with admin access and running its own web server:
+
+```python
+c.JupyterHub.services = [
+    {
+        'name': 'my-web-service',
+        'url': 'https://10.0.1.1:1984',
+        'api_token': 'super-secret',
+    }
+]
+```
+
+
+## Writing your own services
+
+TODO
+
+### Authenticating with the Hub
+
+TODO
+
+JupyterHub 0.7 introduces some utiltiies for you to use that allow you to use the Hub's authentication mechanism.
--- a/jupyterhub/apihandlers/services.py
+++ b/jupyterhub/apihandlers/services.py
@@ -1,0 +1,64 @@
+"""Service handlers
+
+Currently GET-only, no actions can be taken to modify services.
+"""
+
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+import json
+
+from tornado import web
+
+from .. import orm
+from ..utils import admin_only
+from .base import APIHandler
+
+def service_model(service):
+    """Produce the model for a service"""
+    return {
+        'name': service.name,
+        'admin': service.admin,
+        'url': service.url,
+        'prefix': service.server.base_url if service.server else '',
+        'command': service.command,
+        'pid': service.proc.pid if service.proc else 0,
+    }
+
+class ServiceListAPIHandler(APIHandler):
+    @admin_only
+    def get(self):
+        data = {name: service_model(service) for name, service in self.services.items()}
+        self.write(json.dumps(data))
+
+
+def admin_or_self(method):
+    """Decorator for restricting access to either the target service or admin"""
+    def m(self, name):
+        current = self.get_current_user()
+        if current is None:
+            raise web.HTTPError(403)
+        if not current.admin:
+            # not admin, maybe self
+            if not isinstance(current, orm.Service):
+                raise web.HTTPError(403)
+            if current.name != name:
+                raise web.HTTPError(403)
+        # raise 404 if not found
+        if name not in self.services:
+            raise web.HTTPError(404)
+        return method(self, name)
+    return m
+
+class ServiceAPIHandler(APIHandler):
+
+    @admin_or_self
+    def get(self, name):
+        service = self.services[name]
+        self.write(json.dumps(service_model(service)))
+
+
+default_handlers = [
+    (r"/api/services", ServiceListAPIHandler),
+    (r"/api/services/([^/]+)", ServiceAPIHandler),
+]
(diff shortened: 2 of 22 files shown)


willingc:
@minrk Great start and thanks for sharing.

When I get some time at the end of the weekend or Monday morning, I'll take a pass at editing the Services specification document and submit a PR against your branch.

As an FYI, I spent time earlier in the week editing docstrings for [`jupyterhub/services/auth.py`]( https://github.com/willingc/jupyterhub/blob/docstring-svcs/jupyterhub/services/auth.py) since they flow into the API docs. Not yet ready for a PR.

docs/source/services.md
```diff
@@ -0,0 +1,84 @@
...
+If a `command` is specified for launching the service, it will be started and managed by the Hub.
+If a `url` is specified for where the service runs its own webserver,
+it will be added to the Hub's proxy at `/service/:service-name`.
+
+## Managed services
+
+**Managed** services are services that the Hub starts and is responsible for.
+These can only be local subprocesses of the Hub.
+While there are similarities with notebook Spawners,
+there are no plans to support the same spawning abstractions as notebook
```
jhamrick: Grammar/missing a period?

docs/source/services.md
```diff
@@ -0,0 +1,84 @@
...
+    Requires Hub to be root.
+- `cwd: path` directory in which to run the service, if different from the Hub directory.
+
+When the service starts, the Hub will pass the following environment variables:
+
+```
+JUPYTERHUB_SERVICE_NAME: the name of the service ('cull-idle' above)
+JUPYTERHUB_API_TOKEN: API token assigned to the service
+JUPYTERHUB_API_URL: URL for the JupyterHub API (http://127.0.0.1:8080/hub/api)
+JUPYTERHUB_BASE_URL: Base URL of the Hub (https://mydomain/)
```
jhamrick: I assume this includes the port as well?

docs/source/services.md
```diff
@@ -0,0 +1,84 @@
...
+## Managed services
+
+**Managed** services are services that the Hub starts and is responsible for.
+These can only be local subprocesses of the Hub.
+While there are similarities with notebook Spawners,
+there are no plans to support the same spawning abstractions as notebook
+If you want to run these services in docker or other environments,
+you can register it as an external service below.
+
+A managed service is characterized by the `command` specified for launching the service.
```
jhamrick: Will managed services also be monitored and restarted if they exit?

docs/source/services.md
```diff
@@ -0,0 +1,84 @@
...
+JUPYTERHUB_API_TOKEN: API token assigned to the service
+JUPYTERHUB_API_URL: URL for the JupyterHub API (http://127.0.0.1:8080/hub/api)
+JUPYTERHUB_BASE_URL: Base URL of the Hub (https://mydomain/)
+JUPYTERHUB_SERVICE_PATH: Base path of this service (/service/cull-idle/)
+```
+
+## External services
+
+You can use your own service management tools, such as docker or systemd, to manage JupyterHub services.
+These are not subprocesses of the Hub, and you must tell JupyterHub what API token the service is using to perform its API requests.
```
jhamrick: Does each service need a unique API token? What would happen if two services are started with the same API token?

docs/source/services.md
```diff
@@ -0,0 +1,84 @@
...
+
+An example of an externally managed service with admin access and running its own web server:
+
+```python
+c.JupyterHub.services = [
```
jhamrick: For external services, it might be nice if there was a JupyterHub API endpoint they could hit that would tell them the information that JupyterHub knows about -- for example, it could include most of the environment variables that you pass to managed services, like:

* JUPYTERHUB_SERVICE_NAME
* JUPYTERHUB_SERVICE_PATH
* JUPYTERHUB_BASE_URL

Presumably this wouldn't need to include the API token or API url since those would be needed to make this call in the first place.

The reason why I think this might be useful is that then I don't need to hardcode in a bunch of configuration options for my services -- those options can be loaded dynamically from JupyterHub itself. For example, the formgrader has its own configuration options for specifying the base hub URL, which makes configuration a pain since you have to remember to set them both in the JupyterHub config as well as the nbgrader config.

jhamrick:
🎉 🍰 

This looks awesome! I left few questions/comments in the design doc -- mostly clarification things, though I had one suggestion for an API endpoint to include that services could query.

docs/source/services.md
minrk: Good call, I'll add an API endpoint where you can fetch this.

docs/source/services.md
minrk: Yes, each service needs its own token because the requests are authenticated based on the originating service/user. Services are peers to Users in this way - each request is performed 'by' a user or service, so each token maps to a single user or service.

docs/source/services.md
minrk: Yup, thanks.

docs/source/services.md
minrk: yes, ip:port, host:port, host should all work here.

docs/source/services.md
minrk: yes, I'll add a note to this effect

minrk:
Thanks for the feedback, @jhamrick! I've added a few todo items. I know I said at the dev meeting that I expected to have this ready to go for you by Fall semester, but it's been a much crazier summer of conferences and workshops than I expected, so sorry about that. I'm sprinting at EuroSciPy right now, which is the last such thing I have for a couple of months, so this can be my focus for the next while.

jupyterhub/orm.py
```diff
@@ -153,6 +153,35 @@ def api_request(self, path, method='GET', body=None, client=None):
...
+            body=dict(
+                target=service.server.host,
+                service=service.name,
+            ),
+            client=client,
+        )
+
+    @gen.coroutine
+    def delete_service(self, service, client=None):
+        """Remove a service's server to the proxy table."""
```
willingc: to -> from

jhamrick:
No worries, I've been super busy myself this summer so I'm not sure if I would have had time to incorporate the changes into nbgrader anyway. Once I'm back in Berkeley (tomorrow!) I should have more time to get it integrated, so you're timing is actually perfect 😄

docs/source/services.md
```diff
@@ -0,0 +1,84 @@
...
+If a `url` is specified for where the service runs its own webserver,
+it will be added to the Hub's proxy at `/service/:service-name`.
+
+## Managed services
+
+**Managed** services are services that the Hub starts and is responsible for.
+These can only be local subprocesses of the Hub.
+While there are similarities with notebook Spawners,
+there are no plans to support the same spawning abstractions as notebook
+If you want to run these services in docker or other environments,
```
ssanderson: The main use-case we've had at Q for something like services is for one-per-hub containers that do things like run [pgbouncer](https://pgbouncer.github.io/) or maintain data volumes that are shared with user notebook containers.

Am I right in thinking that these uses wouldn't be well-suited for services as they're outlined here?  It sounds like I basically want a managed service that isn't a simple local process (I could shell out to `docker run` I suppose, but then I'd have to keep the container open in interactive mode or the process would exit immediately), which is explicitly noted as being unsupported here.

minrk: I think you are right, but I'll clarify a few things to be sure.

There's two use cases I intend to cover here:

1. simple subprocesses run locally, managed by the Hub
2. externally managed services, where this code only helps with:
    1. setting up API tokens (e.g. cull-idle, HubAuth)
    2. setting up proxy forwarding for web services (e.g. formgrader)

If what you are after is keeping an external service up while the Hub is up, then probably adding it one level up at the docker-compose (or similar) level, rather than internal to JupyterHub, makes the most sense to me.

ssanderson: We've so far found it useful to have singleton spawning in the hub itself, since that lets it participate on the hub config system.  This makes it easy, for example, to run only sub-parts of our full hub setup (e.g. not using pgbouncer when running locally, or not creating and destroying heavyweight containers between development runs).

Currently what we're doing is overriding the hub's setup and teardown logic to first invoke our own setup code before continuing.  There are a couple things we're hooking this way, but the singleton container code looks like this:

```python
    @gen.coroutine
    def init_spawners(self):
        self.init_singleton_containers()
        yield super().init_spawners()

    def init_singleton_containers(self):
        """
        Initialize singleton containers and remove any single-user containers
        that are left over from the last run of the hub.
        """
        log = self.log
        log.info("Initializing singleton containers.")

        self.cleanup_single_user()

        # The client instantiated in each container doesn't survive getting
        # copied from the config file, so share our client among
        # the singleton containers.
        for container in self.singleton_containers:
            log.info("Starting [%s]" % container.name)
            instances = container.instances()
            # Remove any existing instances of this container before starting.
            if instances and self.clean_singletons:
                log.warn("Found instance of [%s]. Purging." % container.name)
                container.purge()
            elif not instances:
                container.run()

    def cleanup_containers(self):
        """
        Clean up singleton containers and any single-user containers that
        haven't been removed yet.
        """
        self.log.info("Cleaning up singleton containers.")
        if self.research_config.clean_singletons:
            for container in self.singleton_containers:
                self.log.info("Stopping singleton: [%s]" % container.name)
                container.purge(stop_first=True, remove_volumes=True)

        self.cleanup_single_user()
```

This has worked well enough for us so far, so I'm not super concerned that services won't support our use -case.  I do, however, think there's a useful class of services which are conceptually like local processes but require some other startup mechanism.  Another example of this might be the ability to set up new `PeriodicCallback`s to start and stop with the hub (we do this to make our hubs send heartbeats to an external monitoring system).

minrk: Would a generic extension mechanism be sufficient? For the most part, I would like to discourage people adding code that runs in the Hub process itself, but that doesn't mean I need to make it impossible.

I could start simple with two hooks:

1. setup, fires after all Hub initialization, before starting the main loop
2. teardown, fires during cleanup

Alternately, I could allow passing a Service object, in addition to the current dict spec, which would let you define your own start/stop behavior (pretty much equivalent to above). This might have the benefit of being more clearly not a general extension mechanism.

ssanderson: For generic setup and teardown hooks, do you think there's a benefit of having configurable extension points vs telling advanced users to subclass and override `JupyterHub.initialize` and `JupyterHub.cleanup`?  Flask, for example, [explicitly encourages](http://flask.pocoo.org/docs/0.11/becomingbig/#becoming-big) advanced users to override documented APIs on their main application class.  

I think the main downside to a subclass/extend strategy for us would be that you have to write your own entrypoint to use a subclass, since the default `jupyterhub` cli calls `JupyterHub.launch_instance` directly. (I'm running from my own entrypoint on Quantopian, which is easy enough if you know where to look.)

I'd be :+1: on supporting passing instances of `Service` to set up services that don't fall cleanly into the existing buckets of managed/non-managed.  If that were the API though, seems a little odd to also support passing dictionaries, since those are presumably getting turned into `Service` objects by the hub anyway.

What would you think about making `Service` an interface (consisting, say, of `start`, `stop`, and `close`), and defining two concrete implementations `LocalProcessService` and `ExternalService`, which would express the two currently-targeted cases, but would leave room for enterprising third-party developers to write, say, `LocalDockerService` or `PeriodicalCallbackService`?  This would also make it easier to write services as jupyterhub plugins, since you could just tell users to put something like the following in their jupyterhub config:
`
```
from my_cool_module import MyCoolService

c = get_config()
c.JupyterHub.services = MyCoolService(...)
```

minrk: >  do you think there's a benefit of having configurable extension points vs telling advanced users to subclass and override JupyterHub.initialize and JupyterHub.cleanup?

Not really, and subclass makes it a bit clearer who the appropriate audience for such customization is, which I like. I should document that, though, so it's not quite at the "if you know where to look" level.

>  If that were the API though, seems a little odd to also support passing dictionaries, since those are presumably getting turned into Service objects by the hub anyway.

I agree that it's weird. I wanted to do something a bit more declarative so there are fewer black boxes to deal with, and the capabilities are explicitly not another arbitrary scope-expansion point, but if I end up allowing an instance, I would probably make it *always* an instance as you suggest, instead of sometimes a dict and sometimes an instance.

A service is, by definition here, something that has API access to the Hub, so it will go in the DB and get an API token. That might be one point against using services as the extension point for things that don't actually want that.

minrk:
It works! @jhamrick I've got everything working for services short of the formgrader-as-external-service example.

@ssanderson I'm going to stop short of trying out the instance API in this PR, but I'll take a look at it next week, and will make another PR if it seems better than the dicts here.

willingc:
@minrk Awesome! Do you want to merge and then I can iterate the docs and docstrings to be end-user friendly in a separate PR (which is probably less confusing than doing it here)?

ssanderson:
Sounds reasonable to me.

jhamrick:
@minrk awesome. I would maybe say go ahead and merge this, and I will work on updating the formgrader (maybe tonight?). If I run into any issues we can iterate on master (unless you'd prefer to get that working first before merging).

jupyterhub/apihandlers/services.py
```diff
@@ -0,0 +1,64 @@
...
+class ServiceListAPIHandler(APIHandler):
+    @admin_only
+    def get(self):
+        data = {name: service_model(service) for name, service in self.services.items()}
+        self.write(json.dumps(data))
+
+
+def admin_or_self(method):
+    """Decorator for restricting access to either the target service or admin"""
+    def m(self, name):
```
willingc: Do we want a more descriptive name than 'm'?

jupyterhub/app.py
```diff
@@ -291,6 +292,15 @@ def _subdomain_host_changed(self, name, old, new):
             # if not specified, assume https: You have to be really explicit about HTTP!
             self.subdomain_host = 'https://' + new
 
+    domain = Unicode(
+        help="domain name without proto,port"
```
willingc: Are we set on using 'proto'? My first reaction was "is this a typo for proxy?" Since it's only in 3 or so places in the code base, could we change it to 'protocol'?

jupyterhub/orm.py
```diff
@@ -222,9 +267,22 @@ def check_routes(self, user_dict, routes=None):
                 # catch filter bug, either in sqlalchemy or my understanding of its behavior
```
willingc: Hmm... I wouldn't hold up the PR for this but it would be nice to catch why this is true occasionally.

willingc:
+1 on merging. I left some small comments on the PR. Once it is merged, I will flesh out the user docs and docstrings for the API.

jupyterhub/orm.py
minrk: This hasn't changed, it was introduced some time ago. I still don't understand how it was happening, but I've never seen it personally.

minrk:
@willingc thanks, I think I addressed your comments.

@jhamrick merging now, thanks!

willingc:
Thanks @minrk 🌅

PR #693, opened 2016-08-18
Documentation refresh

willingc:
The docs have evolved with the project over the past few months. 

Though still a WIP, these are some of the edits:
- Primarily reorganizing the flow of content to better reflect how users will install and plan configuration of JupyterHub
- Simplify section titles
- Correct spelling and links
- Edit grammar to be more active and precise when discussing hub components

Overall, trying to better reflect the flow of the PyData London talk in the docs.

PR #707, opened 2016-08-28
Update conda env and conf.py for clean build

willingc:
PR #693 added HTML theme options that are breaking build for docs. Remove this from conf.py.

Move sphinx to conda install in `environment.yml`


willingc:
Merging so the docs are up-to-date and building green again.

PR #711, opened 2016-08-29
Update changelog format

willingc:
- Add dates for releases.
- Add links to GitHub compare listing with the prior release


minrk:
Nice!

PR #717, opened 2016-09-01
HubAuth login_url changes:

minrk:
1. use `/hub/login` instead of full URL in HubAuth.login_url
2. allow absolute URLs if they are for our host (equivalent to absolute path)

cf https://github.com/jupyter/nbgrader/pull/538

cc @jhamrick

### DIFF
--- a/jupyterhub/handlers/login.py
+++ b/jupyterhub/handlers/login.py
@@ -40,10 +40,10 @@ class LoginHandler(BaseHandler):
     def get(self):
         self.statsd.incr('login.request')
         next_url = self.get_argument('next', '')
-        if next_url.startswith('%s://%s' % (self.request.protocol, self.request.host)):
+        if (next_url + '/').startswith('%s://%s/' % (self.request.protocol, self.request.host)):
             # treat absolute URLs for our host as absolute paths:
             next_url = urlparse(next_url).path
-        if not next_url.startswith('/'):
+        elif not next_url.startswith('/'):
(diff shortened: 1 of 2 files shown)


jupyterhub/handlers/login.py
```diff
@@ -38,8 +40,11 @@ def _render(self, login_error=None, username=None):
     def get(self):
         self.statsd.incr('login.request')
         next_url = self.get_argument('next', '')
+        if next_url.startswith('%s://%s' % (self.request.protocol, self.request.host)):
```
jhamrick: Maybe `if next_url.startswith('%s://%s/'` (with a slash on the end) to make sure people can't do e.g. `https://compmodels.tmpnb.org.com`?

jupyterhub/handlers/login.py
```diff
@@ -38,8 +40,11 @@ def _render(self, login_error=None, username=None):
...
+            # treat absolute URLs for our host as absolute paths:
+            next_url = urlparse(next_url).path
         if not next_url.startswith('/'):
```
jhamrick: `elif`?

jupyterhub/handlers/login.py
minrk: Good catch, done.

jupyterhub/handlers/login.py
minrk: This was on purpose, but now that I think about it you're right. Done here as well.

minrk:
Thanks for the review, @jhamrick!

PR #727, opened 2016-09-07
Edit Services document

willingc:
@minrk @jhamrick I've done some pretty significant editing to the Services document to be more user focused. I've tried not to change the meaning of Min's original spec but please do double-check that I haven't misinterpreted anything.

### DIFF
--- a/docs/source/services.md
+++ b/docs/source/services.md
@@ -21,8 +21,8 @@ Two main characteristics help define a Service:
 2. Does the *Service* have a web server that should be added to the proxy's 
    table?
 
-Currently, a Service may be either a 'Managed Service' which is managed by
-JupyterHub, or an 'External Service' which runs its own web server and
+Currently, a Service may be either a 'Hub-Managed Service' which is managed by
+JupyterHub, or an 'Externally-Managed Service' which runs its own web server and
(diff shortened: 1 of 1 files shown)


docs/source/services.md
```diff
@@ -1,70 +1,115 @@
...
+### Properties of a Service
+
+If a `command` is specified for launching the Service, the Service will be
+started and managed by the Hub.
+
+If a `url` is specified for where the Service runs its own web server,
+JupyterHub will add the Service to the Hub's proxy at 
+`/service/:service-name`.
+
+## Managed Service
```
willingc: Should we change this to 'Hub-managed Service'?

As well as change 'External Service' to 'Externally-managed Service' or 'Self-managed Service'? Or simply leave as 'External Service'?

minrk: Yes, I think Hub-Managed and Externally-Managed is clearer.

minrk:
The changes are great, thanks @willingc!

willingc:
Updated naming to 'Hub-Managed' and 'Externally-Managed'.

Corrected meaning of 'admin' in Line 55.

minrk:
Awesome, thanks @willingc! 🍰

PR #734, opened 2016-09-08
backward-combat for ip, port in LocalProcessSpawner subclasses

### LINKED ISSUE #726: Deprecation of old spawners doesn't have a fallback
The current code in `user.py` says:

    233                 # prior to 0.7, spawners had to store this info in user.server themselves.
    234                 # Handle < 0.7 behavior with a warning, assuming info was stored in db by the Spawner.
    235                 self.log.warning("DEPRECATION: Spawner.start should return (ip, port) in JupyterHub >= 0.7")

It's probably better to raise an exception instead
As it is, it gives the impression that the code can keep working as is, but in practice, the server ip and port are not set and the wait_up times out.

minrk:
Subclasses prior to 0.6 may assume return value of LocalProcessSpawner.start can be ignored instead of passing it through.

For these cases, keep setting ip/port in the deprecated way so that it still works with a warning, rather than failing with the wrong port.

cc @stefanoborini

closes #726

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -495,6 +495,16 @@ class LocalProcessSpawner(Spawner):
             raise
 
         self.pid = self.proc.pid
+
+        if self.__class__ is not LocalProcessSpawner:
+            # subclasses may not pass through return value of super().start,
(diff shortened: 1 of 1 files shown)


willingc:
LGTM.

PR #738, opened 2016-09-09
Add navigation to README

willingc:
The README page is getting a bit long. I was over on the zulip repo and saw this navigation for the top of the README which saves the user some scrolling if they know which section they want.

Let me know what you think.


minrk:
Nice!

PR #761, opened 2016-09-19
Update API docs

willingc:
- Update so first link in 'Using the JupyterHub API' redirects to the API spec
- Add links and updated text for swagger and its transfer to the OpenAPI Initiative
- Provide link to swagger rendering of API in the 'JupyterHub API Spec' doc.


minrk:
Nice!

PR #762, opened 2016-09-19
Edit descriptions in API spec for user clarity

willingc:
@minrk Assuming that it is okay to update descriptions without breaking any API promise with the users.

~~Will remove notesapi.txt as it is just notes to myself~~

`notesapi.md` has my notes and questions from reviewing `rest-api.yaml`.


docs/rest-api.yml
```diff
@@ -7,29 +7,29 @@ info:
...
   - application/json
 paths:
   /:
     get:
-      summary: See the version of JupyterHub itself
+      summary: Get JupyterHub version
       description: |
-        This endpoint is not authenticated,
-        so clients can identify the version of JupyterHub before setting up authentication.
+        This endpoint is not authenticated. Clients and user can identify
```
minrk: I would like to preserve the information that it is not authenticated *for the purpose* of this being the case (previously indicated by `so ...`), rather than separate statements.

docs/rest-api.yml
```diff
@@ -200,12 +200,12 @@ paths:
           type: string
       responses:
         '204':
-          description: The server has stopped
+          description: The user's notebook server has stopped
         '202':
-          description: The server has been asked to stop, but is taking a while
+          description: The user's notebook server has not stopped, but a stop request has been submitted
```
minrk: This has lost the information that 202 will only happen if it is taking a while to stop. Can you add that somewhere?

docs/rest-api.yml
```diff
@@ -200,12 +200,12 @@ paths:
...
   /users/{name}/admin-access:
     post:
-      summary: Grant an admin access to this user's server
+      summary: Grant admin access to this user's notebook server
```
minrk: This is specifically "an admin", as in "an administrator". It only gives one individual access to the server.

docs/rest-api.yml
```diff
@@ -214,7 +214,7 @@ paths:
           type: string
       responses:
         '200':
-          description: Sets a cookie granting the requesting admin access to the user's server
+          description: Sets a cookie granting admin access to the user's notebook server
```
minrk: Same here - "the requesting admin" is important information. Perhaps "requesting administrator"?

docs/rest-api.yml
```diff
@@ -387,13 +387,13 @@ paths:
         type: string
       responses:
         '200':
-          description: The user identified by the API token
+          description: The user is identified by the API token
```
minrk: The content of the response here is the user identified by the token. Adding 'is' no longer describes the content of the response.

docs/rest-api.yml
```diff
@@ -405,7 +405,7 @@ paths:
         type: string
       responses:
         '200':
-          description: The user identified by the cookie
+          description: The user is identified by the cookie
```
minrk: same as above, `is` decreases accuracy of the description.

willingc:
@minrk thanks for the review. I'll make the suggested changes. 👍

Issue #723, opened 2016-09-05
Add docs on upgrading JupyterHub


yuvipanda:
Especially when db fields are added / changed, upgrading requires manually. I think we use alembic for this, but we don't have any specific docs on how to do this.

We should have those docs!

minrk:
Indeed we should! Where I expect this error to occur (connecting to the db in the first place), there is [an informative error](https://github.com/jupyterhub/jupyterhub/blob/acb49adfea7d3923310e5bed02d55014632a654f/jupyterhub/app.py#L761) message pointing to:

    jupyterhub upgrade-db

I'm surprised by the recent debugging chat that an error occurred at such a later time.

It certainly should be added to the docs, probably including the fact that sqlite doesn't allow for all of the operations (ALTER TABLE being the key missing operation) that may be necessary for upgrading the db, so upgrade may only be available on postgres, etc.

willingc:
Great suggestion @yuvipanda 👍 

Some topics to cover:

- Upgrading JupyterHub reference deployments
- Upgrading JupyterHub when no database changes are being made (i.e. no new fields)
- Upgrading JupyterHub when adding a new db field (sqllite vs. postgres...)
- Informative message to user as mentioned by @minrk prompting the user to run `jupyterhub upgrade-db` after upgrading jupyterhub.

willingc:
Here's a pass at some ideas for upgrading doc. Possibly more questions than answers currently re: our support:

## Definitions
db type: sqlite, postgres, mysql, others???
db fields: individual fields in a database table

## Basic use cases
- [ ] Simple upgrade - no database fields changed; no db type change
- [ ] Field changes  - database fields added, deleted or changed by a new JHub release/functionality; no db type change

## More complex use cases
- [ ] Field changes - database fields added, deleted or changed by user; no db type change
- [ ] db type change (i.e. sqlite to postgres) - fields may or may not change; db type does change

## Process
- Remind admins to back up Hub databases
- Possibly remind admins to make sure single user notebook data is saved/backed up
- Admin notify users (if desired) of planned upgrade

**Multi-step upgrade vs all at once**??
- Migrate database type (Hub at same release level) user-initiated
- Add/change/delete fields (Hub at same release level) user-initiated
- Add/change/delete files (Hub initiated - in prep for JHub release upgrade)
- Upgrade Hub to 0.7 after

- Upgrade/changes to: CHP, authenticators, spawners

## Other topics
- limitations of sqlite upgrade
[...]

PR #767, opened 2016-09-20
Add document on upgrading JupyterHub and its db

willingc:
Addresses #723


minrk:
For upgrading the db, the process is:

0. backup, shutdown
1. upgrade JupyterHub
2. run `jupyterhub upgrade-db`

which will run alembic to upgrade the tables. It is a good idea for the system to be shutdown (the Hub must be), though running single-user servers can *probably* be fine across an upgrade.

One of the big caveats is that sqlite's lack of ALTER TABLE means that not all upgrades are possible if the db uses sqlite. So far, I think all upgrades do work because alembic can add columns without ALTER TABLE, but most downgrades do not.

willingc:
@minrk Thanks for the comments. I'll try to update today during the Grace Hopper keynotes. I've given you push access to my fork if you want to push any changes to it as we iterate this.

minrk:
Thanks! I probably won't before the weekend, because I'm heading to Linköping for [SBW](http://sbw2016.se) for Thursday and Friday. Have fun at Grace Hopper!

willingc:
@minrk Great! Thanks for the link to SBW. I'm sure that the workshop will go well. On another note, I met a Google SRE last night and she was really interested and happy that Kubernetes was used at the SciPy workshop for parallel processing.

docs/source/upgrading.md
```diff
@@ -0,0 +1,59 @@
...
+```
+
+## A note about JupyterHub databases
+
+The default database for JupyterHub is a sqlite database. sqlite is useful for
+testing, small deployments, and workshops.
+
+When running a long term deployment or a production system, we recommend using
+a database that supports ALTER_TABLE, such as Postgres or MySQL, as it will
+simplify future upgrades that may change database fields.
```
minrk: Relevant info to possibly go here or elsewhere:


Disadvantages of SQLite:

- upgrade-db may not work, and you may need to start with a fresh database
- downgrade-db **will not** work if you want to rollback to an earlier version, so backup the jupyterhub.sqlite file before upgrading

Losing the Hub database is often not a big deal. Information only in the Hub db includes:

- active login tokens (user cookies, service tokens)
- users added via GitHub UI, instead of config files
- info about running servers

If the following are true, you should be fine clearing the Hub database and starting over:

- users specified in config file
- user servers are stopped during upgrade
- don't mind causing users to login again after upgrade

minrk:
Great!

PR #776, opened 2016-09-26
remove stopped users from proxy on startup

### LINKED ISSUE #775: Proxy entries not cleaned up when single user servers are stopped
I'm running the proxy and the hub separately in installs (this allows me to restart them independently). This leads to some issues when the hub is restarted but the proxy isn't.

1. Hub calls user.stop() on all currently running users
2. This stops all the running singleuser services, but doesn't actually remove entries from the proxy!
3. Hub comes back, correctly determins there are no singleuser servers running. But the Proxy still thinks they are, and so this state is mismatched
4. Next time when you try to spawn something, the proxy proxies it back to where the singleuser server *used* to run, and fails with a 5xx error.
5. Restarting the proxy clears the proxy route table, bringing it back in sync with what the hub sees. This 'fixes' the issue

This doesn't manifest itself in the traditional configuration of the hub managing the proxy, since the proxy does usually get restarted with the hub. It also doesn't manifest when `c.JupyterHub.cleanup_servers = False`, since single user servers aren't stopped when the hub restarts. So it only manifests when you are running the proxy and hub separately with `cleanup_servers = True`

I think the real fix is to also cleanup the routes in the proxy when cleaning up single user servers on restart.

minrk:
We already added running users, but we didn't handle removing users from the proxy if the user's server was stopped (e.g. while the Hub was restarting).

closes #775

### DIFF
--- a/jupyterhub/user.py
+++ b/jupyterhub/user.py
@@ -161,9 +161,9 @@ class User(HasTraits):
     @property
     def proxy_path(self):
         if self.settings.get('subdomain_host'):
-            return url_path_join('/' + self.domain, self.server.base_url)
+            return url_path_join('/' + self.domain, self.base_url)
         else:
-            return self.server.base_url
+            return self.base_url
     
     @property
     def domain(self):
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1112,7 +1112,7 @@ class JupyterHub(Application):
         if self.proxy.public_server.is_up() or self.proxy.api_server.is_up():
             # check for *authenticated* access to the proxy (auth token can change)
             try:
-                yield self.proxy.get_routes()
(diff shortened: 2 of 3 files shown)


jupyterhub/user.py
```diff
@@ -161,9 +161,9 @@ def escaped_name(self):
     @property
     def proxy_path(self):
         if self.settings.get('subdomain_host'):
-            return url_path_join('/' + self.domain, self.server.base_url)
+            return url_path_join('/' + self.domain, self.base_url)
         else:
-            return self.server.base_url
+            return self.base_url
```
willingc: Just for clarification since it's in the code in various places: when would one use `self.server.base_url` vs `self.base_url`?

willingc:
One developer question for my own education. This looks good to merge.

PR #777, opened 2016-09-26
Work on service authentication

minrk:
- HubAuth loads defaults from service environment variables
- HubAuthenticated mixin populates default HubAuth
- Fix setting service cookie (including test)
- Simplify HubAuthenticated example, now that defaults work

This means that the *only* thing you need to authenticate a tornado-based service with the Hub is:

```python

class MyHandler(HubAuthenticated, RequestHandler):
    @web.authenticated
    def get(self):
        ...
```

for a Hub-managed service. An externally managed service may need to set a few variables,
or (perhaps simplest) set the same environment variables that a managed service would get:

    JUPYTERHUB_API_TOKEN
    JUPYTERHUB_API_URL

cc @jhamrick as this may simplify formgrader integrating HubAuth.

### DIFF
--- a/examples/service-whoami/README.md
+++ b/examples/service-whoami/README.md
@@ -5,7 +5,7 @@ Uses `jupyterhub.services.HubAuthenticated` to authenticate requests with the Hu
 ## Run
 
 1. Launch JupyterHub and the `whoami service` with `source launch.sh`.
-2. Visit http://127.0.0.1:8000/hub/whoami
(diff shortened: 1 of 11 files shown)


jhamrick:
Very cool, thanks! I'll see if I can play around with this and formgrader later today...

willingc:
@jhamrick I'm going to merge this and we can iterate on anything your find in a separate PR :smile:

jhamrick:
Ok, sounds good!

PR #778, opened 2016-09-26
cleanup managed services in shutdown

minrk:
don’t leave them running

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1311,6 +1311,13 @@ class JupyterHub(Application):
         """Shutdown our various subprocesses and cleanup runtime files."""
 
         futures = []
+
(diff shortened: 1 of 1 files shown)


jupyterhub/app.py
```diff
@@ -1311,6 +1311,13 @@ def cleanup(self):
         """Shutdown our various subprocesses and cleanup runtime files."""
```
willingc: Perhaps update here: """Shutdown our managed services, various subprocesses, and cleanup runtime files."""

willingc: Suggestion for Line 1330: Replace SUS with single user servers

 # clean up proxy while SUS are shutting down

willingc:
Code looks good. Two small suggestions for comments that may help others' understanding when your clones are busy :wink:

PR #779, opened 2016-09-26
Dockerfile: set debian CDN

minrk:
because the default httpredir fails with some regularity

based on info from http://deb.debian.org


willingc:
Nice!

PR #782, opened 2016-09-26
Updated Logos To Match New Brand Guidelines

spoorthyv:
I updated all the logos to match the new brand guidelines we created for Jupyter. Let me know if you have any issues/questions.


willingc:
Thanks @spoorthyv. These look good to me. I'll leave to @minrk to merge.

We should probably check the other repos under the jupyterhub org to see if they need any updated files.

spoorthyv:
Thanks for doing that check for me @willingc!

minrk:
Thanks!

PR #788, opened 2016-09-29
Update link to docs

willingc:
As pypi.org gets more usage vs the older site, the long_description will be more prominently displayed. Over time, we may want to expand the description. For now, let's just update the link to RTD to use the `.io` suffix.


minrk:
🍰

PR #789, opened 2016-09-30
Deprecate `--no-ssl`

minrk:
it's unnecessarily pedantic. Just warn instead.


willingc:
@minrk Looks good. I like the change. :sunny:

I'm going to go ahead and merge.

PR #793, opened 2016-10-04
Slim conf.py comments and options cruft

willingc:
@minrk I would like to start slimming down the `conf.py` files used in the docs across projects to increase maintainability across projects. Ideally, moving to a sensible defaults `conf.py` template for Sphinx.


docs/source/conf.py
```diff
@@ -297,105 +139,49 @@
...
     # build rest-api, since RTD doesn't run make
     from subprocess import check_call as sh
     sh(['make', 'rest-api'], cwd=docs)
-# otherwise, readthedocs.org uses their theme by default, so no need to specify it
 
-# Spell checking using sphinxcontrib-spelling
+# -- Spell checking -------------------------------------------------------
+
+try:
+    import sphinxcontrib.spelling
```
minrk: Do you want to make this a requirement, rather than optional?

willingc: For now let's leave it optional. Running spelling successfully depends on enchant which can be more annoying than helpful at times. I only run it occasionally.

willingc:
Thanks Min!

PR #799, opened 2016-10-06
Move README installation instructions to docs

willingc:
@minrk This adds the README's installation instructions to the Sphinx docs as a **Quickstart - Installation**. It's basically a replication of the README content with reflowed text in places. 

Normally, I wouldn't choose to repeat stuff. In reviewing the docs, it was a key piece that was missing from it. 

There's still some links that need to checked which I can do in this PR or iterate in another.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -38,6 +38,7 @@ Contents
 
 **User Guide**
 
+* :doc:`quickstart`
 * :doc:`getting-started`
 * :doc:`howitworks`
 * :doc:`websecurity`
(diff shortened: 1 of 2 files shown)


minrk:
@willingc do you want to replace the content in README with a `see quickstart...` link?

willingc:
@minrk I could go either way with that. I think the README appeals to some types of folks that just want to clone and go. Your call -- either is good with me.

For now, I would say the README is the most up-to-date definitive install instructions. What we may want to do (though I'm not sure now) is to refactor the entire user getting started section of the JHub docs.

minrk:
Yeah, makes sense. I like to keep duplication down, so maybe pare down what's in the README and link to the docs for more. That can be another PR, though.

PR #797, opened 2016-10-05
spawn_pending set too soon causing redirect loop

ianabc:
Attempts to fix #569 where redirects loop between /hub/user/USERNAME and /user/USERNAME while waiting for jupyterhub-singleuser to become available. I'm using dockerspawner.systemuserspawner along with a custom authenticator.

If you artificially set the spawn_timeout in base.py to something low (2s) it makes it easier to trigger the redirect behaviour. Testing against f8229c9fb6ab5a5135f67ea253e6833cebd761f7 and https://github.com/jupyterhub/dockerspawner/commit/e7e2af631b65bcaad4f205d2f481b2f5b6a4f722 my browser complains about a redirect loop without the patch. The patch moves the spawn_pending to after server.wait_up.

PR #802, opened 2016-10-07
add User.waiting_for_response

minrk:
There are two major async stages to spawning:

1. Spawner.start
2. waiting for Spawner to become responsive at its http endpoint

Previously, `.spawn_pending` only covered the `Spawner.start` window, which meant that weird things could happen if servers are slow to start up, because they are not added to the proxy until after they become responsive.

`.waiting_for_response` can only be True while `.spawn_pending` is also True.

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -325,10 +325,13 @@ class BaseHandler(RequestHandler):
         try:
             yield gen.with_timeout(timedelta(seconds=self.slow_spawn_timeout), f)
         except gen.TimeoutError:
-            if user.spawn_pending:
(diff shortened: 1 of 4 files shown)


minrk:
Merging, since this is needed to get the tests back to passing after #797

egorlitvinenko:
Hello, @minrk! How can know in which Jupyterhub version this PR was included?
I'm having the same "redirect loop" problem with JupyterHub 0.6.1 and DockerSpawner 0.5.0 installations now.

minrk:
@egorlitvinenko it will be in 0.7, which will be in beta this week.

PR #805, opened 2016-10-11
MNT: Add hub host and prefix to template vars in prep for JLab extension

danielballan:
See https://github.com/jupyter/jupyterlab/pull/1071

A PR with the actual extension that uses this change is forthcoming. It would be great to get this (and maybe also the extension itself, if it's ready) into the next tagged release.


minrk:
Is the Hub private URL really desirable here? That's a private host that's accessible to the notebook server, but generally not accessible to the *client* who is getting the javascript context.

minrk:
Ignore me, I was thinking of api_url. hub_host is exactly right here.

minrk:
I'll add some comments so that it's clearer what everything is, since I'm clearly forgetting.

danielballan:
No worries, thanks.

PR #813, opened 2016-10-13
Add error message when generate config path does not exist

### LINKED ISSUE #812: Generate config file with a path produces error
`jupyterhub --generate-config -f /etc/jupyterhub/jupyterhub_config.py` produces the following error.

screen shot 2016-10-12 at 22 34 56

Looking at the tests, it seems as if we are only testing the filename not the path+filename.

The docstring seems to indicate that the above command is valid:

screen shot 2016-10-12 at 22 42 11

willingc:
Check when writing config that path exists. Add error message when it does not.

Also, edit docstring.

Closes #812

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -180,7 +180,7 @@ class JupyterHub(Application):
 
     generate default config file:
 
-        jupyterhub --generate-config -f /etc/jupyterhub/jupyterhub.py
+        jupyterhub --generate-config -f /etc/jupyterhub/jupyterhub_config.py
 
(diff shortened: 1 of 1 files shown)


jupyterhub/app.py
```diff
@@ -1351,6 +1351,9 @@ def cleanup(self):
 
     def write_config_file(self):
         """Write our default config to a .py config file"""
+        if not os.path.exists(self.config_file):
```
minrk: It should check for the *directory*, not the file (`os.path.dirname(self.config_file)`). The file missing is fine (and likely).

PR #817, opened 2016-10-17
ENH: Add JupyterLab Hub extension.

danielballan:
This is still a WIP, but it's close.

### DIFF
--- a/jupyterlab_hub/README.md
+++ b/jupyterlab_hub/README.md
@@ -1,0 +1,36 @@
+# jupyterlab_hub
+
+JupyterLab extension intergating JupyterHub
+
+
+## Prerequisites
+
+* JupyterLab 0.3.0 or later
+
+## Installation
+
+To install using pip:
+
+```bash
+pip install jupyterlab_hub
+jupyter labextension install --py --sys-prefix jupyterlab_hub
+jupyter labextension enable --py --sys-prefix jupyterlab_hub
+```
+
+## Development
+
+For a development install (requires npm version 4 or later), do the following in the repository directory:
+
+```bash
+npm install
+pip install -e .
+jupyter labextension install --symlink --py --sys-prefix jupyterlab_hub
+jupyter labextension enable --py --sys-prefix jupyterlab_hub
+```
+
+To rebuild the extension bundle:
+
+```bash
+npm run build
+```
+
--- a/jupyterlab_hub/package.json
+++ b/jupyterlab_hub/package.json
@@ -1,0 +1,29 @@
+{
+  "private": true,
+  "name": "jupyterlab_hub",
+  "version": "0.1.0",
+  "description": "JupyterLab extension intergating JupyterHub",
+  "author": "Project Jupyter",
+  "main": "lib/plugin.js",
+  "keywords": [
+    "jupyter",
+    "jupyterlab"
+  ],
+  "scripts": {
+    "build:src": "tsc --project src && node scripts/copyfiles.js",
+    "build:extension": "node scripts/build_extension.js",
+    "build": "npm run build:src && npm run build:extension",
+    "clean": "rimraf lib && rimraf jupyterlab_hub/static",
+    "prepublish": "npm run build"
+  },
+  "dependencies": {
+    "jupyterlab": "^0.5.1",
+    "url-join": "^1.1.0"
+  },
+  "devDependencies": {
+    "fs-extra": "^0.30.0",
+    "rimraf": "^2.5.4",
+    "jupyterlab-extension-builder": "^0.6.2",
+    "typescript": "^2.0.3"
+  }
+}
--- a/jupyterlab_hub/scripts/build_extension.js
+++ b/jupyterlab_hub/scripts/build_extension.js
@@ -1,0 +1,7 @@
+var buildExtension = require('jupyterlab-extension-builder').buildExtension;
+
+buildExtension({
+        name: 'jupyterlab_hub',
+        entry: './lib/plugin.js',
+        outputDir: './jupyterlab_hub/static'
+});
--- a/jupyterlab_hub/src/plugin.ts
+++ b/jupyterlab_hub/src/plugin.ts
@@ -1,0 +1,91 @@
+// Copyright (c) Jupyter Development Team.
+// Distributed under the terms of the Modified BSD License.
+
+import {
+  Menu
+} from 'phosphor/lib/ui/menu';
+
+import {
+  getBaseUrl, getConfigOption
+} from '@jupyterlab/services/lib/utils';
+
+import {
+  ICommandPalette
+} from 'jupyterlab/lib/commandpalette';
+
+import {
+  IMainMenu
+} from 'jupyterlab/lib/mainmenu';
+
+import {
+  JupyterLab, JupyterLabPlugin
+} from 'jupyterlab/lib/application';
+
+import * as urljoin
+  from 'url-join';
+
+
+/**
+ * Activate the jupyterhub extension.
+ */
+function activateHubExtension(app: JupyterLab, palette: ICommandPalette, mainMenu: IMainMenu): void {
+
+  // This config is provided by JupyterHub to the single-user server app.
+  // The app passes in jinja template variables which populate lab.html.
+  let hubHost = getConfigOption('hubHost');
+  let hubPrefix = getConfigOption('hubPrefix');
+
+  if (!hubPrefix) {
+    console.log('No JupyterHub configuration found.');
+    return
+  }
+
+  console.log('JupyterHub configuration found: ' + hubHost + hubPrefix);
+
+  let { commands, keymap } = app;
+  let category = 'Hub';
+
+  let menu = new Menu({ commands, keymap });
+  menu.title.label = 'Hub';
+
+  let command: string;
+
+  // Add commands and menu itmes for each link.
+  command = 'hub:control-panel';
+  commands.addCommand(command, {
+    label: 'Control Panel',
+    execute: () => {
+      window.open(hubHost + urljoin(hubPrefix, 'home'), '_blank');
+    }
+  });
+  palette.addItem({command: command, category: "Hub"});
+  menu.addItem({ command });
+
+  command = 'hub:logout';
+  commands.addCommand(command, {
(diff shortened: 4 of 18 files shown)


danielballan:
@minrk I think I'm missing something in `jupyterlab_hub/jupyterlab_hub/singleuser.py` to correctly serve the static files from the labextension.

Running:

```
jupyterhub --Spawner.cmd="['jupyter-singlelabuser', '--debug']"
```

and browsing to `http://localhost:8000/user/dallan/lab`

logs:

```
404 GET /hub/labextension/jupyterlab_hub/jupyterlab_hub.bundle.js (dallan@::1) 2.86ms
```

jupyterlab_hub/typings/xterm/xterm.d.ts
```diff
@@ -0,0 +1,83 @@
+// Type definitions for term.js 0.0.7
```
minrk: Something seems wrong if we have to copy these typing files into this repo. Comments, @jasongrout?

jasongrout: Typescript 2.0 introduces a new system for propagating typing information about 3rd party packages that would eliminate this, once these typings are stored in the central DefinitelyTyped repository. We need to submit the typings that we created for 3rd party libraries that aren't in that repository, but in the meantime, they may need to be copied.

jupyterlab_hub/src/typings.d.ts
```diff
@@ -0,0 +1,4 @@
+/// <reference path="../typings/url-join/url-join.d.ts"/>
+
+// For a Thenable reference in jupyter-js-services
+type Thenable<T> = Promise<T>
```
jasongrout: This can go away soon, due to recent changes in jupyter-js-services. We'll update the cookie cutter soon.

jupyterlab_hub/typings/xterm/xterm.d.ts
jasongrout: However, if a typings file is not referenced in the `typings.d.ts` file, by our convention it's not used, so no need to copy it.

danielballan:
I lean toward a separate repo (at least for now) to keep the release cycles separate.

> I think that should not require a new entrypoint anywhere.

The second entrypoint seems gratuitous, but I'm not sure how else to work in the labextensions-related attributes: https://github.com/jupyterlab/jupyterlab/blob/master/jupyterlab/labapp.py#L32

 `jupyter-singleuser` is tied to `SingleUserNotebookApp`. How else can more configurable attributes be added?

minrk:
Thanks for working on this, @danielballan!

Perhaps @jasongrout can speak to how to enable lab extensions without adding a new entrypoint. I could be wrong about that, in which case the entrypoint you have makes perfect sense.

I made this repo: https://github.com/jupyterhub/jupyterhub-labextension if you want to push forward on a standalone package for this.

jasongrout:
> Perhaps @jasongrout can speak to how to enable lab extensions without adding a new entrypoint.

Enabling lab extensions is a configuration thing, exactly like enabling nbextensions. I don't think it really has anything to do with a new entry point.

danielballan:
@jasongrout Right, but the `LabApp` adds certain configurable traitlets that aren't available on the `SingleUserNotebookApp`. We need the union of the configurables on the `LabApp` and the configurables on the `SinlgerUserNotebookApp` (which is `NotebookApp` + hub-specific configurables). To that end, I made a subclass multiply inheriting from these two.

Now, we have to tell the hub's spawner class (which is itself configurable) what command to run to start a single-user server. That's where the new entry point comes in: instead of starting `SingleUserNotebookApp`, we need to start an app from our new subclass.

I'm absolutely open to finding a simpler way to do this, but I'm not seeing one at the moment.

jasongrout:
> the LabApp adds certain configurable traitlets that aren't available on the SingleUserNotebookApp.

Ah, right - it is only a bit of configuration data (e.g., in what directories to look for the lab configuration file), but the way we have our configuration system set up, that necessitates adapting the class.

minrk:
@jasongrout gotcha, so to use JupyterLab extensions, JupyterLab cannot itself be used as a serverextension. Is that intended? Do you want to get rid of the jupyterlab server extension if the two behaviors are diverging like this? Or try to keep the extension behavior consistent?

jasongrout:
> @jasongrout gotcha, so to use JupyterLab extensions, JupyterLab cannot itself be used as a serverextension. Is that intended? Do you want to get rid of the jupyterlab server extension if the two behaviors are diverging like this? Or try to keep the extension behavior consistent?

It would be great if we could reconcile the two (server extension and configurability requiring subclassing). We face the same problem with pulling the notebook out of the new Jupyter Server.

jasongrout:
Huh, using `jupyter notebook` and changing the url to `/lab` still works fine. I didn't expect it to work after the discussion above.

minrk:
>  I didn't expect it to work after the discussion above.

On looking at it now, I see that the serverextension does load lab extensions, because it skips the normal config loading, and uses JSONConfigManager (like frontend config) in a get_labextensions function. So there doesn't appear to be a need to use LabApp here.

jasongrout:
> On looking at it now, I see that the serverextension does load lab extensions, because it skips the normal config loading, and uses JSONConfigManager (like frontend config) in a get_labextensions function. So there doesn't appear to be a need to use LabApp here.

I'm not sure if this is what you're saying, but I see the server extension augmenting the normal web application at https://github.com/jupyterlab/jupyterlab/blob/master/jupyterlab/__init__.py#L136. This seems to override the app-specific configuration at https://github.com/jupyterlab/jupyterlab/blob/master/jupyterlab/labapp.py#L69, so I'm not sure the app is doing anything, really.

jasongrout:
CC @blink1073, who wrote the labextension stuff based on the nbextension mechanism.

jupyterlab_hub/scripts/build_extension.js
```diff
@@ -0,0 +1,7 @@
+var buildExtension = require('jupyterlab-extension-builder').buildExtension;
```
blink1073: @jupyterlab/extension-builder

jupyterlab_hub/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "prepublish": "npm run build"
+  },
+  "dependencies": {
+    "jupyterlab": "^0.5.1",
+    "url-join": "^1.1.0"
+  },
+  "devDependencies": {
+    "fs-extra": "^0.30.0",
+    "rimraf": "^2.5.4",
+    "jupyterlab-extension-builder": "^0.6.2",
```
blink1073: @jupyterlab/extension-builder

jupyterlab_hub/README.md
```diff
@@ -0,0 +1,36 @@
+# jupyterlab_hub
+
+JupyterLab extension intergating JupyterHub
+
+
+## Prerequisites
+
+* JupyterLab 0.3.0 or later
```
blink1073: 0.5.1

jupyterlab_hub/README.md
```diff
@@ -0,0 +1,36 @@
+# jupyterlab_hub
+
+JupyterLab extension intergating JupyterHub
```
blink1073: itegrating

jupyterlab_hub/src/plugin.ts
```diff
@@ -0,0 +1,91 @@
+// Copyright (c) Jupyter Development Team.
+// Distributed under the terms of the Modified BSD License.
+
+import {
+  Menu
+} from 'phosphor/lib/ui/menu';
+
+import {
+  getBaseUrl, getConfigOption
+} from '@jupyterlab/services/lib/utils';
```
blink1073: We're discouraging importing from fully qualified paths in `@jupyterlab`.  Import the utils namespace and access these functions as `utils.getBaseUrl`.

jupyterlab_hub/typings/xterm/xterm.d.ts
blink1073: None of these typings files should be required here (except the url-join).  If they are then we've accidentally leaked a third party API.

danielballan:
@minrk @jasongrout I see. Yes, I agree that renders the LabApp unnecessary.

@blink1073 Thanks for the careful look. Are we correct in concluding, above, that the server extension does all the necessary lab-specific configuration and that `LabApp` is not needed? JupyterLab starts successfully without `LabApp`, but I want to confirm that we are not missing anything about your future intent for the config-loading.

blink1073:
I'd say that the `LabApp` should not be required.  The `LabApp` is only intended for use with `jupyter lab`.  If the JupyterLab serverextension is enabled, then any NotebookApp should be able to reach `/lab` with the properly loaded extensions.

jasongrout:
One thing the Lab app does give us is a configurable search path for the lab extension config files. (Though I think it's getting overwritten...)

danielballan:
Right... `get_labextensions` is following the same default search path as a normal configurable, but it clobbers the ability to configure that path, say, with command line args. Are we OK with that?

jasongrout:
I think some of @SylvainCorlay's work on the Jupyter Server will change how we think about configuration of extensions - unless we have a need now for the configurability, I think we should address this as part of that larger refactoring.

danielballan:
That works for me. I'll address the line comments and continue this on the
new repo Min created.

willingc:
Hi @danielballan, Do you think we should keep this open or close?

danielballan:
I will rebase my work against the new repo Mom created.

danielballan:
Sigh. Mom -> Min. Thanks, iPhone. :- )

blink1073:
Nice.  Usually people only refer to their own mother as capital Mom ;).

willingc:
@danielballan Thanks for the 😆. For a brief moment, after 21 years with my kids, I thought you were referring to me as I've been called Mom millions of times. I do have a 🎃 costume idea now ;-)

willingc:
@danielballan Siri, please apologize to Dan for the iPhone's autocorrect. Siri, please also give us an Escape key on our new MacBook Pro.

danielballan:
Right?! Maybe we should capitalize on a coming market for Esc-mapped foot
pedals?

PR #818, opened 2016-10-18
don’t try to start unmanaged services

minrk:
they don't have anything to start


willingc:
Thanks @minrk. Good catch on the typo too.

PR #830, opened 2016-10-25
Flesh out custom services examples

minrk:
- add flask whoami service example, so we have one non-tornado case
- allow HubAuthenticated to check groups

cc @willingc for docs and @parente who's looking at testing out using the services auth.

closes #823

PR #832, opened 2016-10-26
Use warning instead of warn for logs

willingc:
A clean up PR. Following up on something that caught my eye in #830, I've  replaced the handful of remaining log messages that still used 'warn' instead of the more current 'warning'.


minrk:
Whoops! Thanks, @willingc!

PR #835, opened 2016-10-26
Edit model check to be consistent for user and group

### LINKED ISSUE #834: Remove or edit unreachable code block
@minrk It looks like `_check_user_model` has a few code lines that don't seem to be reachable. Should we remove the lines after the return statement? https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/apihandlers/base.py#L144

willingc:
Closes #834 

@minrk I think this makes these two functions more consistent. Let me know what needs additional changes. Thanks.


minrk:
Thanks!

PR #838, opened 2016-10-28
quotes around single-user CLI args

### LINKED ISSUE #836: Redirect loop after starting server
My JupyterHub installation uses GitHub authentication. It works well for a dozen users, however, for one user, myself, I enter a redirect loop when starting the server. It’s completely bizarre. The only difference between my user and the other users is that my username starts with a digit. Any ideas?

minrk:
avoids mishandling things such as integer-literals

closes #836

I was able to test locally and verify that it works.

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -298,15 +298,15 @@ class Spawner(LoggingConfigurable):
     def get_args(self):
         """Return the arguments to be passed after self.cmd"""
         args = [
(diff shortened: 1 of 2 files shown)


willingc:
Hey @minrk. Looks like two of the api tests are failing. They may need to be updated with this PR.

minrk:
Thanks for the catch, should be fixed, now.

0x00b1:
Thanks, @minrk!

PR #842, opened 2016-11-01
finish error message when generate-config path does not exist

### LINKED ISSUE #812: Generate config file with a path produces error

minrk:
finish up #813

- checks directory existence
- exit with message on stderr if not found

closes #812

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1351,9 +1351,11 @@ class JupyterHub(Application):
 
     def write_config_file(self):
         """Write our default config to a .py config file"""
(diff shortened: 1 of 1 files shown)


minrk:
Thanks!

PR #852, opened 2016-11-08
[WIP] Update Dockerfile

parente:
* Use nodejs, CHP from condaforge
* Bump the version of conda used


parente:
Test failure for py3.5 looks bogus. I don't have permissions to restart it.

PR #858, opened 2016-11-11
Reflow text and minor edits following PR #855

willingc:
There are a few edits in wording for reader clarity. Text is reflowed for future maintainability. A follow up PR to #855.


yuvipanda:
<3

Issue #843, opened 2016-11-01
Add more info on security to docs


willingc:
Per discussion on gitter, it would be helpful to add documentation and/or admonition to answer:

@parente's question:

>are the tokens in the jupyterhub db signed/hashed so that the true API tokens can't be recovered if someone gets their hands on the database?

@minrk's response:

> [here](https://github.com/jupyterhub/jupyterhub/blob/0.6.1/jupyterhub/utils.py#L151) for hashing and [here](https://github.com/jupyterhub/jupyterhub/blob/0.6.1/jupyterhub/orm.py#L335) for what goes in the db - the first 4 bytes are stored in cleartext as a prefix to limit the number of comparisons that must be computed

parente:
@willingc is it worth my sending a PR adding a small note about the DB contents to the Web Security in JupyterHub &rarr; Extra notes section or Getting Started &rarr; Security as they exist today? Or are you working on an overhaul that makes a PR moot?

willingc:
@parente If you want to open a PR that would be great, we'll be merging #865 in a minute.

parente:
On it.

PR #866, opened 2016-11-12
Add two short notes about db security

parente:
For #843.


parente:
I'll rebase and reflow.

willingc:
Thanks @parente :sunny:

PR #869, opened 2016-11-13
Edit Services doc content

willingc:
Primarily edits for user clarity and additional links to examples.

Added links to major sections to ease user navigation.

Reflowed some text to make easier for doc edits.

### DIFF
--- a/docs/source/services.md
+++ b/docs/source/services.md
@@ -1,60 +1,90 @@
 # Services
 
+With version 0.7, JupyterHub adds support for **Services**.
 
-With version 0.7, JupyterHub adds support for 'Services'.
+This section provides the following information about Services:
 
+- [Definition of a Service](services.html#definition-of-a-service)
+- [Properties of a Service](services.html#properties-of-a-service)
+- [Hub-Managed Services](services.html#hub-managed-services)
+- [Launching a Hub-Managed Service](services.html#launching-a-hub-managed-service)
+- [Externally-Managed Services](services.html#externally-managed-services)
+- [Writing your own Services](services.html#writing-your-own-services)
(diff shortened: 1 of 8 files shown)


docs/source/services.md
```diff
@@ -283,7 +320,7 @@ and taking note of the following process:
     where cookie-value is the url-encoded value of the `jupyterhub-services` cookie.
     This request must be authenticated with a Hub API token in the `Authorization` header.
     For example, with [requests][]:
```
parente: Missing link?

docs/source/services.md
```diff
@@ -301,10 +338,10 @@ and taking note of the following process:
...
 
    ```json
-   { 
+   {
      "name": "inara",
      "groups": ["serenity", "guild"],
-     
+
    }
    ```
```
parente: Worth adding a section at the bottom about the nbviewer implementation? It's not very instructive, but it is a an example of something you might conceivably want to secure behind Hub auth.

* https://github.com/jupyter/nbviewer#securing-the-notebook-viewer
* https://github.com/jupyter/nbviewer/blob/master/nbviewer/providers/base.py#L94

docs/source/services.md
parente: Ah. My bad. Missed it there.

docs/source/services.md
parente: Thank you for adding it! 🍰  + 🍰 

One clarification: nbviewer can be run either as a hub managed or externally managed service. The README for nbviewer does say it. Not sure if it's worth clarifying in the Hub docs here.

willingc: Will add both. Thanks!

PR #872, opened 2016-11-15
Change py.test to newer convention of pytest

willingc:
Older command `py.test` is not working on Sierra. Updating to newer convention for pytest.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,10 +183,10 @@ python3 setup.py js    # fetch updated client-side js
 python3 setup.py css   # recompile CSS from LESS sources
 ```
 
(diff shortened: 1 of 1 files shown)


willingc:
@minrk This Travis segfault looks very similar to what I saw this morning: https://travis-ci.org/jupyterhub/jupyterhub/jobs/176197146

minrk:
Interesting! I have only used `py.test` on Sierra so far.

willingc:
Hmm...They should both work looking at the pytest docs and release note. http://blog.pytest.org So not sure why it wasn't working with the older `py.test`. Going forward `pytest` will be the preferred command by the  pytest project.

PR #876, opened 2016-11-21
allow `check_whitelist` to be a coroutine

jbweston:
Some authenticators may have whitelist checking that requires
async operations, and this should be supported.

Notably [BitBucketOAuthenticator](https://github.com/jupyterhub/oauthenticator/blob/master/oauthenticator/bitbucket.py#L94) needs this in order for team whitelisting to function correctly
(currently team whitelisting does not actually work).


minrk:
Thanks, @jbweston!

PR #881, opened 2016-11-22
handle async check_whitelist in app

minrk:
follow-up to previous PR

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -920,6 +920,7 @@ class JupyterHub(Application):
         # From this point on, any user changes should be done simultaneously
         # to the whitelist set and user db, unless the whitelist is empty (all users allowed).
     
+    @gen.coroutine
     def init_groups(self):
         """Load predefined groups into the database"""
         db = self.db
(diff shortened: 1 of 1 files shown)


minrk:
follow-up to #876

jupyterhub/app.py
```diff
@@ -980,11 +982,12 @@ def _add_tokens(self, token_dict, kind):
...
+    
+    @gen.coroutine
     def init_api_tokens(self):
         """Load predefined API tokens (for services) into database"""
-        self._add_tokens(self.service_tokens, kind='service')
-        self._add_tokens(self.api_tokens, kind='user')
+        yield self._add_tokens(self.service_tokens, kind='service')
+        yield self._add_tokens(self.api_tokens, kind='user')
     
     def init_services(self):
```
willingc: @minrk Does this need `@gen.coroutine` too?

minrk: it doesn't, only those that call async methods need it.

willingc: Got it. Thanks.

PR #886, opened 2016-11-28
Expand inline documentation

yuvipanda:


yuvipanda:
@willingc @minrk this should be a pure doc update. PTAL when/if you have the time and let me know if I got any details wrong :)

yuvipanda:
I'll try to do the same for JupyterHub itself, and the base authenticator at some point.

jupyterhub/spawner.py
```diff
@@ -27,63 +27,102 @@
...
     ).tag(config=True)
-    port = Integer(0,
-        help="The port for single-user servers to listen on. New in version 0.7."
+
+    port = Integer(
+        0,
+        help="""
+        The port for single-user servers to listen on.
+
+        Defaults to `0`, which uses the default port (8888).
```
minrk: This isn't accurate. The default is a random port.

jupyterhub/spawner.py
```diff
@@ -92,6 +131,8 @@ class Spawner(LoggingConfigurable):
               <option value="A">The letter A</option>
               <option value="B">The letter B</option>
             </select>
+
+        The data from this form submission will be passed on to your spawner in `self.user_options`
```
minrk: worth specifying that they will pass through `self.options_from_form`

jupyterhub/spawner.py
```diff
@@ -119,50 +165,93 @@ def options_from_form(self, form_data):
...
+        with the name of the user.
+
+        Note that this does *not* prevent users from accessing files outside of this path! They
+        can do so with many other means.
         """
     ).tag(config=True)
 
-    default_url = Unicode('',
-        help="""The default URL for the single-user server. 
+    # FIXME: Provide example of such an URL!
```
minrk: An example url would be `/lab`

jupyterhub/spawner.py
```diff
@@ -119,50 +165,93 @@ def options_from_form(self, form_data):
...
+    default_url = Unicode(
+        '',
+        help="""
+        The URL the single-user server should start in.
 
-        Can be used in conjunction with --notebook-dir=/ to enable 
-        full filesystem traversal, while preserving user's homedir as
-        landing page for notebook
+        You can set `notebook_dir` to `/` and then construct a URL here to allow people to navigate
+        the whole filesystem from their notebook, but still start in their home directory.
```
minrk: The case described here would be:

```python
notebook_dir = '/'
default_url = '/home/{username}'
```

jupyterhub/spawner.py
```diff
@@ -107,9 +148,14 @@ def options_from_form(self, form_data):
         prior to `Spawner.start`.
         """
         return form_data
-    
-    user_options = Dict(help="This is where form-specified options ultimately end up.")
-    
+
+    user_options = Dict(
+        help="""
+        Dict of user specified options specific to this spawned instance of the single-user server.
```
willingc: Dict of user specified options for the user's spawned instance of a single-user server. These user options are usually provided by an `options_form` displayed to the user for setting options.

jupyterhub/spawner.py
```diff
@@ -119,50 +165,93 @@ def options_from_form(self, form_data):
...
-    cmd = Command(['jupyterhub-singleuser'],
-        help="""The command used for starting notebooks."""
+
+    # FIXME: Add info about shell expansion here.
+    cmd = Command(
+        ['jupyterhub-singleuser'],
+        help="""
+        The command used for starting the single-user server.
+
+        You can provide this as either a list or a string. Note that this should only contain
```
willingc: Provide either a string or a list containing the path to the startup script command. Extra arguments, other than this path, should be provided via...

jupyterhub/spawner.py
```diff
@@ -119,50 +165,93 @@ def options_from_form(self, form_data):
...
-        `~` will be expanded to the user's home directory
-        `{username}` will be expanded to the user's username
+
+    notebook_dir = Unicode(
+        '',
+        help="""
+        Path to the notebook directory for the single-user server.
+
+        The user will see a file listing of this directory when they start, and the Notebook
+        interface itself will not allow browsing outside the contents of this directory (and its
```
willingc: Path to the notebook directory for the single-user server.

The user sees a file listing of this directory when the notebook interface is started. The current interface does not easily allow browsing beyond the subdirectories in this directory's tree.

jupyterhub/spawner.py
willingc: The default URL for starting the single-user server.

Can be used in conjunction with `notebook_dir` to provide greater flexibility for filesystem navigation. For example, setting `notebook_dir = '/'` and `default_url = '/home/{username}` permits a user to view their home directory when starting the notebook interface and to navigate the entire filesystem (not just subdirectories of the home directory).

jupyterhub/spawner.py
```diff
@@ -174,12 +263,16 @@ def _deprecate_percent_u(self, proposal):
...
-        
-        This prevents any config in users' $HOME directories
-        from having an effect on their server.
+
+    disable_user_config = Bool(
+        False,
+        help="""
+        Disable per-user configuration of single-user servers.
+
+        This prevents any config in users' $HOME directories from having an effect on their server.
```
willingc: ... when starting their single-user server.  (@minrk, please verify)

I'm not sure that the sentence of circumventing belongs here. Thoughts?

minrk: verified - this is correct.

yuvipanda: In general, if something looks like it could be used for anything security related, I always try to put a note explaining the cases where it would *not* work right there.

yuvipanda:
Thanks for all the comments, @willingc! I'll try to incorporate them during my next flight.

jupyterhub/spawner.py
willingc: Perhaps, this would work :-)

```
Disable per-user configuration of single-user servers.

When starting the user's single-user server, any config file found in the user's $HOME directory will be ignored.
 
Note: a user could circumvent this if the user modifies their Python environment, such as when they have their own conda environments / virtualenvs / containers.
```

jupyterhub/spawner.py
```diff
@@ -27,63 +27,102 @@
...
     api_token = Unicode()
-    ip = Unicode('127.0.0.1',
-        help="The IP address (or hostname) the single-user server should listen on"
+
+    ip = Unicode(
+        '127.0.0.1',
+        help="""
+        The IP address (or hostname) the single-user server should listen on.
+
+        The JupyterHub proxy implementation should be able to send packets to this interface.
```
yuvipanda: Hmm, there's only one proxy running in our case, and I also want to explicitly specify which one it is (the one fronting JupyterHub) since there can be many number of them between the user and the instance (on wikimedia's install there's at least 3!).

jupyterhub/spawner.py
```diff
@@ -27,63 +27,102 @@
...
     _callbacks = List()
     _poll_callback = Any()
-    
+
     debug = Bool(False,
-        help="Enable debug-logging of the single-user server"
+        help="""
+        Enable debug-logging of the single-user server.
+
+        These logs will be sent to wherever the single-user server is configured to send its logs
```
yuvipanda: I've just gotten rid of that bit, since it was confusing either way.

yuvipanda:
I think I've modified it based on most of the comments here.

yuvipanda:
I've also marked this PR as 'allow edits from maintainers', so feel free to edit yourselves too! I'm dealing with some medical things that might keep me offline for a while, don't want to be a blocker :)

jupyterhub/spawner.py
willingc: Sounds good.

minrk:
@yuvipanda thanks! I made a few more tweaks and merged.

willingc:
@minrk Awesome.

@yuvipanda Thanks. A well earned 🍨 for you.

PR #887, opened 2016-11-28
Rename MemorySpecification to ByteSpecification

yuvipanda:
Since we might use this in other places (such as storage, which has already come up in KubeSpawner).


yuvipanda:
@minrk do you think we can get this in for 0.7, so we don't end up with a version out with the wrong class name?

PR #890, opened 2016-12-01
Expand ~ before quoting notebook_dir

### LINKED ISSUE #878: notebooks_dir ~ expands incorrectly

Configure jupyterhub 0.7.0b1 with:

```
c.SudoSpawner.sudospawner_path = "/some/where/bin/sudospawner"
c.SudoSpawner.sudo_args = ['-nH']
c.Spawner.notebook_dir = '~/notebooks'
```

Try to login. Notebook server startup logs:

```
[C 2016-11-21 12:32:15.936 SingleUserNotebookApp application:91] No such notebook dir: '/home/pparente/~/notebooks'
```

Path should be expanded properly.

**What actually happens**

Path is expanded but also gets the ~ part tacked back on.

0.7.0b1

I put a print in the jupyterhub-singleuser script and confirmed that it is receiving `--notebook-dir="~/notebooks"` as in `sys.argv`. So it appears the incorrect expansion is happening somewhere after that.

parente:
closes #878

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -394,7 +394,7 @@ class Spawner(LoggingConfigurable):
             args.append('--port=%i' % self.user.server.port)
 
         if self.notebook_dir:
-            notebook_dir = self.format_string(self.notebook_dir)
+            notebook_dir = os.path.expanduser(self.format_string(self.notebook_dir))
             args.append('--notebook-dir="%s"' % notebook_dir)
         if self.default_url:
             default_url = self.format_string(self.default_url)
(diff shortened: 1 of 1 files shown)


parente:
Don't merge quite yet. Testing a bit.

parente:
Nope. Expands too early for sudospawner.

parente:
Restoring the original, unquoted `--notebook_dir=%s` template in spawner.py fixes the ~ expansion bug. It's questionable whether this is trading one fix for other problems, however.

minrk:
I think this is a bug for the single-user server to fix (by calling `expanduser`), and not to change here. The reason to not change it here is that the quotes were added to deal with an issue in traitlets CLI parsing, to ensure the result is a string (i.e. adding quotes shouldn't change the value).

I also don't think the shell expansion will work for Spawners that launch without instantiating a shell (`shell=True` for Popen), because raw command-launching doesn't expand `~` (I think).

willingc:
Thanks @parente and @minrk :D

PR #893, opened 2016-12-01
call expanduser on singleuser notebook_dir

### LINKED ISSUE #878: notebooks_dir ~ expands incorrectly

minrk:
This copies validate_notebook_dir from notebook with one addition: calling expanduser.

closes #878

cc @parente does this work for you?

### DIFF
--- a/jupyterhub/singleuser.py
+++ b/jupyterhub/singleuser.py
@@ -22,6 +22,7 @@ from traitlets import (
     CUnicode,
     default,
     validate,
+    TraitError,
 )
 
 from notebook.notebookapp import (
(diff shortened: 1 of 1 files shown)


willingc:
@minrk Do we want to add a test for this?

Likely unrelated to this PR but running the tests with this PR applied to master caused a segfault as it had a few weeks ago. My gut instinct is that its related to mock but can't reproduce so just a heads up in case you see anything.

parente:
Thanks for continuing this. I'll give it a test right now.

parente:
Looks good here. :shipit:

PR #895, opened 2016-12-02
Update changelog for 0.7 final

minrk:
Should be ready to go. I'll release as soon as this is merged.

cc @willingc


willingc:
Looks good. Thanks @minrk!

PR #902, opened 2016-12-09
Don't warn about empty next_url

minrk:
empty next_url is fine


willingc:
Thanks @minrk. Reran the Travis job and now it's green 👍

PR #907, opened 2016-12-17
Allow users to have multiple servers (in db)

barrachri:
I added the UserServer model and corrected the User model.

Now you can access servers from user.servers and Users from server.users.

With user.server is equal to user.servers[0] otherwise is None.

I corrected the test_orm.py but probably there is something more to do.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1095,6 +1095,7 @@ class JupyterHub(Application):
                 # if user.server is defined.
                 log = self.log.warning if user.server else self.log.debug
                 log("%s not running.", user.name)
+                # remove all server or servers entry from db related to the user
                 for server in user.servers:
                     db.delete(server)
                 db.commit()
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -146,18 +146,18 @@ class NewToken(Application):
 
 class UpgradeDB(Application):
     """Upgrade the JupyterHub database schema."""
-    
+
     name = 'jupyterhub-upgrade-db'
     version = jupyterhub.__version__
     description = """Upgrade the JupyterHub database to the current schema.
-    
+
     Usage:
 
         jupyterhub upgrade-db
     """
     aliases = common_aliases
     classes = []
-    
+
     def _backup_db_file(self, db_file):
         """Backup a database file"""
         if not os.path.exists(db_file):
@@ -171,7 +171,7 @@ class UpgradeDB(Application):
             backup_db_file = '{}.{}.{}'.format(db_file, timestamp, i)
         if os.path.exists(backup_db_file):
             self.exit("backup db file already exists: %s" % backup_db_file)
-        
+
         self.log.info("Backing up %s => %s", db_file, backup_db_file)
         shutil.copy(db_file, backup_db_file)
 
@@ -222,12 +222,12 @@ class JupyterHub(Application):
         Authenticator,
         PAMAuthenticator,
     ])
-    
+
     load_groups = Dict(List(Unicode()),
         help="""Dict of 'group': ['usernames'] to load at startup.
-        
+
         This strictly *adds* groups and users to groups.
-        
+
         Loading one set of groups, then starting JupyterHub again with a different
         set will not remove users or groups from previous launches.
         That must be done through the API.
@@ -414,7 +414,7 @@ class JupyterHub(Application):
 
     api_tokens = Dict(Unicode(),
         help="""PENDING DEPRECATION: consider using service_tokens
-        
+
         Dict of token:username to be loaded into the database.
 
         Allows ahead-of-time generation of API tokens for use by externally managed services,
@@ -437,14 +437,14 @@ class JupyterHub(Application):
         Allows ahead-of-time generation of API tokens for use by externally managed services.
         """
     ).tag(config=True)
-    
+
     services = List(Dict(),
         help="""List of service specification dictionaries.
-        
+
         A service
-        
+
         For instance::
-        
+
             services = [
                 {
                     'name': 'cull_idle',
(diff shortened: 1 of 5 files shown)


minrk:
Great, thanks! Sorry for the slow response, I have had various fires to put out over the last week.

It looks like some [tests are failing](https://travis-ci.org/jupyterhub/jupyterhub/jobs/184718430#L564) because the server is cleared by setting `user.server = None`, when it now needs to clear the `user.servers` list.

> ...fix some pep8

In the future, could you please avoid this part? It is a safe assumption that if pep8 would make substantial changes to a file, that pep8 isn't followed in that file, and making those changes may not be desirable. pep8 is a good guideline, and a great reference whenever you are in doubt, but it's okay to use your own judgement, and we try not to be strict about these things.

jupyterhub/orm.py
```diff
@@ -355,13 +367,16 @@ class User(Base):
 
     A `state` column contains a JSON dict,
     used for restoring state of a Spawner.
+
+    `server` returns the first entry for the users' servers.
+    'servers' is a list that contains a reference to the ser's Servers.
```
minrk: 'user' missing 'u' here

jupyterhub/orm.py
```diff
@@ -68,6 +70,13 @@ class Server(Base):
     base_url = Column(Unicode(255), default='/')
     cookie_name = Column(Unicode(255), default='cookie')
 
+    # added to handle multi-server feature
+    last_activity = Column(DateTime, default=datetime.utcnow)
+
+    users = association_proxy("server_to_users", "user",
```
minrk: Does it make sense to have `server.users`? There should be always 0 or 1 here.

barrachri: Do you prefer to call this 'user' or to remove the relationship ?

jupyterhub/orm.py
barrachri: corrected. I'll wait to push this new PR

jupyterhub/orm.py
minrk: Yes, I think so.

barrachri: Sorry, which one ? :)

minrk: oops, sorry. I'd remove the relationship.

barrachri: Done :)

minrk:
@barrachri thanks, this is ready to go. I want to release 0.7.2 from master today, then I'll merge this one.

barrachri:
What do we need more for this ? API ? Docs ?

minrk:
We'll need to modify the REST API and think carefully about the redirect logic to handle multiple servers.

Via the REST API, users must be able to list servers, create new servers, and access their existing ones.

The UI needs to change as well - right now, we have a 'start my server' button, but we need something for listing servers, creating named servers, etc. (this will require the REST API changes).

jupyterhub/app.py
```diff
@@ -1095,7 +1095,9 @@ def user_stopped(user):
                 # if user.server is defined.
                 log = self.log.warning if user.server else self.log.debug
                 log("%s not running.", user.name)
-                user.server = None
+                for server in user.servers:
```
willingc: It may make sense to clarify that all servers will be removed when users have more than one server.

jupyterhub/orm.py
```diff
@@ -343,7 +351,7 @@ def find(cls, db, name):
 class User(Base):
     """The User table
 
-    Each user has a single server,
+    Each user can have more than a single server,
```
willingc: I think it may help future developers to be a bit more precise when using the word `servers` since it's a frequently used term in the JupyterHub world.

Here's one suggestion for revising the docstring:

```
Each user can have one or more single user notebook servers.

Each single user notebook server will have a unique token for authorization.
Therefore, a user with multiple notebook servers will have multiple tokens.

```

jupyterhub/orm.py
```diff
@@ -355,13 +363,16 @@ class User(Base):
 
     A `state` column contains a JSON dict,
     used for restoring state of a Spawner.
+
+    'server' returns the first entry for the users' servers.
+    'servers' is a list that contains a reference to the user's Servers.
```
willingc: Perhaps:

```
`servers` is a list that contains a reference for each of the user's single user notebook servers.
The method `server` returns the first entry in the user's `servers` list.
```

jupyterhub/orm.py
```diff
@@ -393,28 +414,57 @@ def __repr__(self):
...
         """Find a user by name.
-
         Returns None if not found.
         """
         return db.query(cls).filter(cls.name==name).first()
 
 
+class UserServer(Base):
+    """The UserServer table
+    Each user can have have more than one server,
```
willingc: This is a bit unclear to me as to what the table is storing.

My understanding is:
```
A table storing the One-To-Many relationship between a user and servers.

Each user may have one or more servers.

A server can have only one (1) user. This condition is maintained by UniqueConstraint.
```

jupyterhub/user.py
```diff
@@ -316,10 +316,9 @@ def stop(self):
             self.state = spawner.get_state()
             self.last_activity = datetime.utcnow()
             # cleanup server entry, API token from defunct server
```
willingc: `# Cleanup defunct servers - delete entry and API token for each server`

jupyterhub/user.py
```diff
@@ -316,10 +316,9 @@ def stop(self):
...
-            if self.server:
-                # cleanup server entry from db
-                self.db.delete(self.server)
-            self.server = None
+            for server in self.servers:
+                # cleanup servers entry from db
```
willingc: `# remove server entry from db

willingc:
@barrachri Nice work :-)

@minrk I've made some suggestions for clarifications on the docstrings. I'm aiming for greater clarity, but I may be misunderstanding the implementation.

Both of you, please correct me if my understanding is off. Thanks!

minrk:
Thanks!

willingc:
Thanks @barrachri. I'm happy to see this PR merge. Nice work :sunny:

PR #910, opened 2016-12-19
Avoid cleaning up API tokens for Spawners that will resume

### LINKED ISSUE #901: Cleaning up API tokens on stop isn't always safe
0.7 deletes a single-user server's API token when the server stops. For DockerSpawner with `remove_containers=False`, this isn't okay, because the server has the first API token baked in and it can't be changed. The symptom being that any time the single-user server resumes after the first spawn, it will fail to check cookies because its API token is no longer valid.

Spawners need a way to communicate that they are still using their API token on stop. One option is a flag on the Spawner to indicate that it's safe to clear the token. Another is to allow the Spawner to *set* its API token, and for the Hub to check this after start, to ensure that it's valid.

### LINKED ISSUE #905: Dockerspawner with Github Oauth breaks in 0.7.0 (works in 0.6.1)
I set dockerspawner with the oauth example provided in the repository, using the pre-built singleuser image (`docker pull jupyter/singleuser`), and everything works when using jupyterhub 0.6.1.  However, if I switch to 0.7.0, the following break:

* logging out - it just redirects back to `/hub/home#` without logging out
* shutting down a server, then trying to launch it again - can no longer connect to the docker image, with a 500 error: `Permission failure checking authorization, I may need to be restarted`

If I rm -f the docker container, it starts back up again fine, but has the same problem if I shut down my server or the jupyterhub server.

With debug mode logging, when logging out, this is the output:

```
[I 2016-12-16 16:34:58.647 JupyterHub login:19] User logged out: X@Y.com
[I 2016-12-16 16:34:58.653 JupyterHub log:100] 302 GET /hub/logout (X@Y.com@50.202.79.166) 5.50ms
[W 2016-12-16 16:34:58.734 JupyterHub pages:32] Disallowing redirect outside JupyterHub: ''
[I 2016-12-16 16:34:58.738 JupyterHub log:100] 302 GET /hub/ (@50.xxxxxx) 1.73ms
[I 2016-12-16 16:34:58.821 JupyterHub log:100] 302 GET /oauth_login (@50.xxxxxx) 1.95ms
[I 2016-12-16 16:34:58.899 JupyterHub google:35] redirect_uri: 'http://ec2-xxxxx:8000/hub/oauth_callback'
[I 2016-12-16 16:34:58.902 JupyterHub log:100] 302 GET /hub/oauth_login (@50.x)
```

minrk:
in which case the previous API token should be left alone.

closes #901
closes #905

Will require a PR to DockerSpawner to set `.will_resume = True` when remove_containers is False.

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -52,6 +52,14 @@ class Spawner(LoggingConfigurable):
     authenticator = Any()
     api_token = Unicode()
 
+    will_resume = Bool(False,
+        help="""Whether the Spawner will resume on next start
+
+        If a Spawner will resume instead of starting anew (e.g. resuming a Docker container),
+        API tokens in use when the Spawner stops will not be deleted.
+        """
+    )
+
     ip = Unicode('127.0.0.1',
         help="""
         The IP address (or hostname) the single-user server should listen on.
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -55,8 +55,11 @@ class Spawner(LoggingConfigurable):
     will_resume = Bool(False,
         help="""Whether the Spawner will resume on next start
 
-        If a Spawner will resume instead of starting anew (e.g. resuming a Docker container),
-        API tokens in use when the Spawner stops will not be deleted.
+
+        Default is False where each launch of the Spawner will be a new instance.
+        If True, an existing Spawner will resume instead of starting anew
+        (e.g. resuming a Docker container),
+        and API tokens in use when the Spawner stops will not be deleted.
         """
     )
 
--- a/jupyterhub/user.py
+++ b/jupyterhub/user.py
@@ -234,6 +234,12 @@ class User(HasTraits):
                 # prior to 0.7, spawners had to store this info in user.server themselves.
                 # Handle < 0.7 behavior with a warning, assuming info was stored in db by the Spawner.
                 self.log.warning("DEPRECATION: Spawner.start should return (ip, port) in JupyterHub >= 0.7")
+            if spawner.api_token != api_token:
+                # Spawner re-used an API token, discard the one we just created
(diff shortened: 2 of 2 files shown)


jupyterhub/spawner.py
```diff
@@ -52,6 +52,14 @@ class Spawner(LoggingConfigurable):
     authenticator = Any()
     api_token = Unicode()
 
+    will_resume = Bool(False,
+        help="""Whether the Spawner will resume on next start
+
```
willingc: Perhaps for clarity add:

`Default is False where the Spawner will start anew.`
`If True, an existing Spawner will resume...`

jupyterhub/user.py
```diff
@@ -234,6 +234,12 @@ def spawn(self, options=None):
                 # prior to 0.7, spawners had to store this info in user.server themselves.
                 # Handle < 0.7 behavior with a warning, assuming info was stored in db by the Spawner.
                 self.log.warning("DEPRECATION: Spawner.start should return (ip, port) in JupyterHub >= 0.7")
+            if spawner.api_token != api_token:
+                # Spawner re-used an API token, discard the one we just created
```
willingc: `... discard the just created token, api_token`

jupyterhub/user.py
```diff
@@ -313,10 +319,11 @@ def stop(self):
             if self.server:
                 # cleanup server entry from db
                 self.db.delete(self.server)
-            orm_token = orm.APIToken.find(self.db, api_token)
-            if orm_token:
-                self.db.delete(orm_token)
             self.server = None
+            if not spawner.will_resume:
+                orm_token = orm.APIToken.find(self.db, api_token)
```
willingc: `# find and remove api token from db`

minrk:
Thanks, @willingc! Comments should be addressed, I think.

willingc:
Looks great @minrk :cake:

PR #912, opened 2016-12-20
missing version.py file in jupyterhub

MAliNaqvi:
not having this line was preventing the automated build


minrk:
Thanks for the PR! Adding this line appears redundant with the line immediately below, which adds the same file in the same location. The docker builds are also tested on every PR and haven't been failing. Can you share some more information about the failure you are seeing and how you are building the image?

MAliNaqvi:
hmm... can't reproduce the error any more.

willingc:
@MAliNaqvi Thanks for the update and for taking the time to contribute the PR. Please keep sharing your feedback with us. :cookie:

PR #914, opened 2016-12-21
fix rest-api doc building

minrk:
- fix link to rest-api page
- update bootprint to 0.10, since 0.8 has stopped working

@willingc this should fix the RTD builds you saw yesterday

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -8,7 +8,7 @@
   "author": "",
   "license": "BSD-3-Clause",
   "devDependencies": {
-    "bootprint": "^0.8.5",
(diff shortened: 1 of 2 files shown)


willingc:
@minrk Thanks! You and your clones are on fire today :fire: I'm only working for an hour (technically) today, but will review/merge as I can over the next week :D

minrk:
Thanks. Have a good holiday!

willingc:
Thank you! All the best over the holidays. I'll be back in full force on Tuesday ;-) Intermittently until then. 🌲 ⛄️

Issue #879, opened 2016-11-22
Docker environment throws 500 because it cant find `notebook` module


zbyte64:
**How to reproduce the issue**
Run the jupyterhub docker image:

`docker run -P -d --name jupyterhub -v $PWD:/srv/jupyterhub -t jupyterhub/jupyterhub`

Create a user as stated in the readme:

```
docker exec -it jupyterhub bash
useradd student
passwd student
``` 

Go to the local jupyterhub in the browser and login.

**What you expected to happen**

Upon successful login we should see a jupyter notebook.

**What actually happens**

We get a 500 error:

```
500 : Internal Server Error
Spawner failed to start [status=1]
```

The logs show:

```
[I 2016-11-22 00:58:13.974 JupyterHub spawner:633] Spawning jupyterhub-singleuser '--user="student"' '--cookie-name="jupyter-hub-token-student"' '--base-url="/user/student"' '--hub-host=""' '--hub-prefix="/hub/"' '--hub-api-url="http://127.0.0.1:8081/hub/api"' '--ip="127.0.0.1"' --port=51639
Couldn't set CWD to /home/student ([Errno 2] No such file or directory: '/home/student')
Traceback (most recent call last):
  File "/opt/conda/lib/python3.5/site-packages/jupyterhub/singleuser.py", line 15, in <module>
    import notebook
ImportError: No module named 'notebook'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/conda/bin/jupyterhub-singleuser", line 3, in <module>
    from jupyterhub.singleuser import main
  File "/opt/conda/lib/python3.5/site-packages/jupyterhub/singleuser.py", line 17, in <module>
    raise ImportError("JupyterHub single-user server requires notebook >= 4.0")
ImportError: JupyterHub single-user server requires notebook >= 4.0
```

**Share what version of JupyterHub you are using**

`0.7.0.b1`

Running `jupyter troubleshoot` in the docker container results in `command not found`

minrk:
The `jupyterhub/jupyterhub` image only has the dependencies to run JupyterHub, not single-user servers. If you want to run single-user servers in the same container as the Hub, you will need to add the notebook package. Traditionally, the jupyterhub container is used in conjunction with a Spawner that puts single-user servers in other containers, such as DockerSpawner.

willingc:
Closing this issue as resolved after PR #915.

Issue #772, opened 2016-09-23
500 : Internal Server Error (Spawner failed to start)


olyerickson:
Immediately after user login (we're using PAM) we see the following:
```
500 : Internal Server Error
Spawner failed to start [status=1]
```

Based on the documentation, we don't know where even to begin to diagnose the problem. Looking for some basic, fine-grained advice...Thanks!

willingc:
Hi @olyerickson,

This troubleshooting FAQ section may help http://jupyterhub.readthedocs.io/en/latest/troubleshooting.html#error-after-spawning-my-single-user-server

If you could provide some of the information logged when starting JupyterHub with `--debug` that would be helpful. Thanks!

olyerickson:
Our sysadmin fixed by modifying the config file to point to shared places. Added a couple configuration sections to /etc/apache2/sites-enabled/000-default.conf for /jupyter, /hub and /user

willingc:
Great! Thanks for the update @olyerickson. Feel free to ping us with questions in the future, if needed.

jacknlliu:
I got this error from the [jupyterhub docker image](https://hub.docker.com/r/jupyterhub/jupyterhub/), and I followed the [instruction](https://github.com/jupyterhub/jupyterhub#docker) .

I want to know if anyone really tested this docker image usage with step by step.

minrk:
@jacknlliu The `jupyterhub/jupyterhub` image contains only the Hub itself, with no configuration. In general, one needs to make a derivative image, with at least a `jupyterhub_config.py` setting up an Authenticator and/or Spawner.

jacknlliu:
@minrk I followed the [configuration instruction](http://jupyterhub-tutorial.readthedocs.io/en/latest/timeline.html#configuring-jupyterhub), but the docker image still got the error, why not provide a basic default configuration to proof the docker image/jupyterhub could work just like [jupyter-notebook](https://github.com/jupyter/docker-stacks/tree/master/minimal-notebook)?

DrSpatula:
@olyerickson Could you be more specific about what configuration options were added to your Apache conf?

PR #915, opened 2016-12-21
Update README to clarify docker image contents

willingc:
Addresses #879 and #772 re: confusion about the docker image contents

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,11 +131,11 @@ Some examples, meant as illustration and testing of this concept:
 ----
 
 ## Docker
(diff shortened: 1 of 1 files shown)


parente:
Drive by comment: is it worth pointing to jupyterhub/jupyterhub-deploy-docker in this section somewhere?

willingc:
@parente Yes 👍

@minrk How would you feel about recommending that 1)  users try `jupyterhub/jupyterhub-deploy-docker` as a "batteries included" Docker deployment and 2) `jupyterhub/jupyterhub` Docker instance may be used as a baseline image that should be customized with auth/spawners/etc. since it does not include the notebooks. Perhaps offering the end-user the "batteries included" option first would better address questions like in #879 and #772.

minrk:
That sounds like an excellent idea.

Issue #889, opened 2016-11-29
Toree Integration causes issues with Ambari rack awareness scripts


rachmaninovquartet:
Hi,

I've setup jupyterhub in a Hortonworks cluster. Everything works fine, until I add the Apache Toree kernel. As soon as I start a notebook, with that, it prints to the logs something like this once per second:

```
16/11/29 16:24:20 WARN ScriptBasedMapping: Exception running /etc/hadoop/conf/topology_script.py some.ip.address
ExitCodeException exitCode=1:   File "/etc/hadoop/conf/topology_script.py", line 63
    print rack
             ^
SyntaxError: Missing parentheses in call to 'print'

	at `org.apache.hadoop.util.Shell.runCommand(Shell.java:576)`
```

It is calling the Ambari generated rack awareness scripts, which have /usr/bin/env python at the top of them. Using jupyterhub, this needs to be python 3. Unfortunately, the scripts only work with python 2.

This occurs in the latest jupyterhub, available via pip (11/29/16). I've asked the user groups of both Ambari and Toree for help, but neither have come up with anything actionable. 

Any potential work arounds?

Thanks,

Ian

mariusvniekerk:
So that is caused by your 

`/etc/hadoop/conf/topology_script.py` being run with the python environment used by jupyterhub (which is python3) and that script not being python 2/3 compatible.

If you modify script to use `print(rack)` instead you should be fine.

mariusvniekerk:
So the one workaround that you try that would not involve modifying the ambari generated files would be to have is to have the script you run to launch the toree kernel alter the `$PATH` so that a python2 kernel will be found by `/usr/bin/env python`.

You'll very likely run into the same issue if you have other python 3 kernels that get run, which activate a python3 conda environment/virtualenv.

rachmaninovquartet:
@mariusvniekerk Do you have any documentation, you would suggest looking into for this "to have the script you run to launch the toree kernel alter the $PATH so that a python2 kernel will be found by /usr/bin/env python"
That sounds like a strong solution.

beeva-gerardovazquez:
Hi,
I have found the same behaviour trying to run pyspark as kernel or loading it from a python2 kernel.
pyspark on console works fine

Regards

rachmaninovquartet:
I did find a fix, but will have to look up exactly what I changed. I
believe I used a custom file and updated an HDFS conf or two, in order to
keep it from getting overwritten. I'll try to look it up later today.

[...]

PR #916, opened 2016-12-21
Added Toree troubleshooting to docs

rachmaninovquartet:
#889 
@willingc as per our conversation, please verify the changes to the docs

### DIFF
--- a/docs/source/troubleshooting.md
+++ b/docs/source/troubleshooting.md
@@ -249,3 +249,26 @@ jupyter kernelspec list
 ```bash
 jupyterhub --debug
 ```
+
+## Toree integration with HDFS rack awareness script
+
+The Apache Toree kernel will an issue, when running with JupyterHub, if the standard HDFS 
+rack awareness script is used. This will materialize in the logs as a repeated WARN:
+
+```bash
+16/11/29 16:24:20 WARN ScriptBasedMapping: Exception running /etc/hadoop/conf/topology_script.py some.ip.address
+ExitCodeException exitCode=1:   File "/etc/hadoop/conf/topology_script.py", line 63
+    print rack
+             ^
+SyntaxError: Missing parentheses in call to 'print'
+
+    at `org.apache.hadoop.util.Shell.runCommand(Shell.java:576)`
+```
+
+In order to resolve this issue, there are two potential options.
+
+1. Update HDFS core-site.xml, so the parameter "net.topology.script.file.name" points to a custom 
+script (e.g. /etc/hadoop/conf/custom_topology_script.py). Copy the original script and change the first line point
+to a python two installation (e.g. /usr/bin/python).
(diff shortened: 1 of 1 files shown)


docs/source/troubleshooting.md
```diff
@@ -249,3 +249,26 @@ jupyter kernelspec list
...
+    at `org.apache.hadoop.util.Shell.runCommand(Shell.java:576)`
+```
+
+In order to resolve this issue, there are two potential options.
+
+1. Update HDFS core-site.xml, so the parameter "net.topology.script.file.name" points to a custom 
+script (e.g. /etc/hadoop/conf/custom_topology_script.py). Copy the original script and change the first line point
+to a python two installation (e.g. /usr/bin/python).
+2. In spark-env.sh add a Python 2 installation to your path (e.g. export PATH=/opt/anaconda2/bin:$PATH).
+
```
parente: A third option from @mariusvniekerk, perhaps worth mentioning, is to modify `/etc/hadoop/conf/topology_script.py` so that it reads `print(rack)` and becomes Python 2 and 3 compatible.

willingc:
@rachmaninovquartet Thanks for the PR and for sharing your knowledge with others. :sunny: 

Would you mind adding @parente's suggestion? If time is tight due to the holidays, I'm happy to merge as is. Just let me know what you prefer.

---

Unrelated to PR:

@minrk Would you mind looking at the Travis output https://travis-ci.org/jupyterhub/jupyterhub/jobs/185882424 ? This is another one of the test timeouts with the mocked proxy. Likely still a mocking issue on the test, but please see if anything else catches your eye.

docs/source/troubleshooting.md
rachmaninovquartet: @parente @willingc 
The problem with  @mariusvniekerk's suggestion, is that there are actually several changes required in the script and even then, when using Ambari, the scripts will be overwritten, whenever services are restarted. I considered it not a robust enough solution for my situation. If you think it should still be added, let me know.

mariusvniekerk: Feels like we should push this upstream to Ambari.

rachmaninovquartet: @mariusvniekerk I think they're counting on the fact that you can create a custom topology script, from that perspective, there isn't much reason to force changes to save in the default script. The real question for the Ambari people is why isn't python 3 supported, when python 2 will no longer be supported in 2020.

rachmaninovquartet:
@willingc I think it could be merged, check my comments above for details

willingc:
@rachmaninovquartet @mariusvniekerk @parente Thanks for the expanded discussion. I agree that the upstream improvements would be the best approach. 👍 

Merging the PR for the docs. Thanks @rachmaninovquartet :cookie:

willingc:
Happy holidays too!

rachmaninovquartet:
Thanks @willingc, Happy Holidays!

PR #922, opened 2017-01-06
Pass query params through with user-redirect

yuvipanda:
Currently query params are stripped when sent through user-redirect,
which causes problems in many places (including UC Berkeley's interact
service)


yuvipanda:
Don't merge yet, am actively testing this now.

yuvipanda:
Tested and works!

PR #933, opened 2017-01-13
Fixed help and documentation related to service tokens

DeepHorizons:
The documentation stated that the key `token` should be used to specify
the pregenerated token in `JupyterHub.services`. This is wrong as the key
should be `api_token`.

This changes the help on the trait, along with changing the module
docstring in `service.py`.


minrk:
Thanks!

PR #936, opened 2017-01-15
Allow services to request admin-access

barrachri:
Currently when you call the api with a service
`
http://127.0.0.1:8000/hub/api/users/<name>/admin-access`

you receive a 500 error because the service doesn't have a `other_user_cookies` defined inside the orm.

With this fix `current.other_user_cookies.add(name)` is only called for users and not for services.

I also added a little bit more info inside the doc about api_token as a property of a service.


minrk:
Thanks!

PR #938, opened 2017-01-16
Make JupyterHub an OAuth provider

### LINKED ISSUE #553: Make JupyterHub an OAuth provider
When discussing deploying services and things, we need to provide a mechanism for services to authenticate with the Hub.

nbgrader [has an implementation of this](https://github.com/jupyter/nbgrader/blob/v0.2.2/nbgrader/auth/hubauth.py) with the current JupyterHub scheme. We're planning to make an importable implementation in JupyterHub for use in tornado-based apps, which should cover many cases. However, if we just made JupyterHub an OAuth provider then services written in many different ways could more easily "authenticate with JupyterHub" using standard libraries for whatever toolkit they use.

I'm not planning to do this any time soon, but writing it down here, just in case.

minrk:
Making JupyterHub itself an OAuth provider should remove the need for passing cookies around.

closes #553

Services can use OAuth to login users with the Hub and set cookies themselves, instead of the Hub preemptively setting cookies on various prefixes, which will start to fall down as we add multiple servers per user, etc.

This should also get rid of some redirect-loop bugs caused by the cookie setup.

Still extremely WIP with hardcoded examples, etc., but getting there.

TODO:

- [x] tests (obvs)!
- [x] get it working at all
- [x] Figure out OAuth client config
- [x] <del>REST API for creating new OAuth clients?</del> #1056
- [x] provide OAuth client implementations in `services.auth`, and use it in singleuser.
- [ ] examples
- [x] deprecate shared cookie auth model
- [x] remove `user.other_user_cookies`
- [x] implement admin-access with OAuth
- [x] set OAuth environment variables for services
- [x] make sure it works when using subdomains


jupyterhub/apihandlers/users.py
```diff
@@ -12,6 +12,18 @@
...
 
+class SelfAPIHandler(APIHandler):
+    """Return the authenticated user's model
+    
+    Based on the authentication info. Acts as a 'whoami' for auth tokens.
+    """
+    @web.authenticated
+    def get(self):
+        user = self.get_current_user()
+        print(user)
```
minrk: remove me

barrachri:
Hi @minrk how can I help with this?

minrk:
@barrachri hi, thanks! This is almost done. I've got tests passing, both manual and the test suite now. I mainly need to do updates on examples and docs, and add a few explicit tests for the new stuff. If you'd like to have a go at updating an example or two (perhaps copy `service-whoami` to `service-whoami-oauth`), that would be great!

The services don't yet get OAuth client credentials, but that should be a small addition.

willingc:
@minrk Nice!

minrk:
Huzzah! Everything is green. I think this can go in, and updates to examples and docs can follow in subsequent PRs.

minrk:
I think there might be some room for refinement of environment variables, etc. because oauth needs more information (it needs to know the full public URL of both the service/single-user server and the Hub in the subdomain case), but that can happen in incremental PRs.

willingc:
Will review this today. Thanks @minrk :-)

jupyterhub/services/auth.py
```diff
@@ -333,6 +522,7 @@ def get(self):
     hub_services = None # set of allowed services
     hub_users = None # set of allowed users
     hub_groups = None # set of allowed groups
+    allow_admin = False # allow any admin user access
```
willingc: Maybe for comment: 
`allow any admin user access if True. Default is False.`

jupyterhub/services/auth.py
```diff
@@ -418,3 +614,33 @@ def get_current_user(self):
...
+
+    .. versionadded: 0.8
+    """
+    
+    @coroutine
+    def get(self):
+        code = self.get_argument("code", False)
+        if not code:
+            raise HTTPError(400, "oauth callback made without a token")
+        # TODO: make async (in a Thread?)
```
willingc: Assuming so things continue if the check takes a while.

jupyterhub/tests/mockservice.py
```diff
@@ -47,21 +49,37 @@ def get(self, path):
...
-    """Reply with the name of the user who made the request."""
+    """Reply with the name of the user who made the request.
+    
+    Uses deprecated cookie login
+    """
+    @web.authenticated
+    def get(self):
+        self.write(self.get_current_user())
+
+class OWhoAmIHandler(HubOAuthenticated, web.RequestHandler):
```
willingc: `OhWhoAmI` for those with amnesia...

jupyterhub/tests/test_api.py
```diff
@@ -722,16 +723,17 @@ def test_group_delete_users(app):
...
     db = app.db
     r = api_request(app, 'services')
     r.raise_for_status()
     assert r.status_code == 200
 
     services = r.json()
     assert services == {
-        'mock-service': {
-            'name': 'mock-service',
+        mockservice.name: {
```
willingc: Is this to handle tests where there are multiple services?

minrk: It helped debug a bit because I was having some state leak from one test or another. This was mainly to ensure that a service's name wasn't re-used from one test to the next.

jupyterhub/app.py
```diff
@@ -1065,6 +1068,16 @@ def init_services(self):
                     raise AttributeError("No such service field: %s" % key)
                 setattr(service, key, value)
 
+            if service.managed:
```
minrk: > This is the case where the OAuth is a service managed by the hub. 

OAuth isn't a service. The service would be something like nbviewer, formgrader, etc.

> Essentially JupyterHub acts as the OAuth provider and issues tokens.

Yes, JupyterHub itself is now an OAuth provider (always). Services may use OAuth to authenticate users.

This token block is unchanged, I just hand to move it up so that the API token is defined before the next block, which uses it.

jupyterhub/oauth/store.py
```diff
@@ -0,0 +1,315 @@
...
+        This method must be implemented to make use of unique access tokens.
+
+        :param client_id: Identifier of the client a token belongs to.
+        :param grant_type: The type of the grant that created the token
+        :param user_id: Identifier of the user a token belongs to.
+        :return: An instance of :class:`oauth2.datatype.AccessToken`.
+        :raises: :class:`oauth2.error.AccessTokenNotFound` if no data could be
+                 retrieved.
+        """
+        raise NotImplementedError("Unique tokens not implemented")
```
minrk: Yup, the AccessTokenStore API defines several methods that our subset of OAuth doesn't need or use, so I haven't implemented them in our implementation. I've now removed them instead of leaving the NotImplementedErrors.

minrk:
Thanks @willingc! I think the review should be addressed.

willingc:
🎉 :cake: 🍨

minrk:
Yay! I'll try to finish #864 next.

brospars:
Hi,
is there any documentation about this ?

I want to use Jupyterhub as an OAuth provider for Gitlab but I'm not sure how to configure it...

Thanks in advance

Issue #766, opened 2016-09-20
Question: Multiple singleuser servers


danielfrg:
Is there plans (or interest) to add support for multiple single-user servers per user? I kinda see it in the [roadmap project sharing](https://github.com/jupyter/roadmap#next--1-08) but not sure how is that planned on being implemented, from what I can guess is basically multiple users pushing to the same git repo that is managed by one service in JupyterHub.

Is there interest on refactoring the code to support this feature? Looking at the models in `orm.py` right now the one single-user server is a little hardcoded in the `User` class but I guess that this is something that can change.

Another solution is probably to have "admin" servers that multiple users can access based on some permissions.

Basically I am just asking for thoughts on these ideas and if there is interest on those features to be added upstream in case they get implemented.

willingc:
Thanks @danielfrg for your thoughts and questions about the roadmap. @minrk would be the best person to provide insights into planning for release 0.8+. :sunny:

micha-f:
We are also interested in this. In our case we'd like to use the docker spawner to spawn multiple containers per user - with different settings per notebook. We'd be happy to contribute if there is an agreement on whether and how to implement support for multiple servers per user.

minrk:
This is definitely planned. Changing server from a singleton to a list in the ORM is not very complicated. The challenge is more in the UI and REST API. We would need:

1. UI for listing and managing servers, rather than a single stop/start buttons.
2. Same thing for the admin page (a little more complex, since it's much more information)
3. last_activity needs to be added as a server-level data point
4. make some cookie/URL-related decisions, e.g.
    - single cookie per user
    - user servers at `/user/:name/:server-id`

And, with all of this, we want deployments to be able to choose to preserve the single-server-per-user notion, which dramatically simplifies the application for users (the Hub is just a login page to them, in this case). This mostly affects views and not the API, but it could reach into the API a little bit in a few places.

parente:
We're also interested in this for giving users the ability to run classic notebook and new jupyter lab separately. (Running both out of the same server instance has some sideeffects like clicking the Jupyter logo always returning the user to /lab even when they were in a classic notebook). My stopgap is to bring up a second jupyterhub instance, but a single hub instance that can manage multiple servers is clearly better in the long run.

I can help out some here.

barrachri:
Hello !

can we set a kind of roadmap ?

I plan to submit a PR decoupling the User and Server models inside orm.py

[...]

PR #942, opened 2017-01-18
Multiple singleuser servers

barrachri:
This PR is a first implementation of the multi singleuser servers paradigm.

For sure it needs to be reviewed

https://github.com/jupyterhub/jupyterhub/issues/766

### DIFF
--- a/docs/source/config-examples.md
+++ b/docs/source/config-examples.md
@@ -36,7 +36,7 @@ if not os.path.exists(ssl_dir):
 
 
 # Allows multiple single-server per user
-c.JupyterHub.allow_multiple_servers = False
+c.JupyterHub.allow_named_servers = True
 
 # https on :443
 c.JupyterHub.port = 443
--- a/docs/source/config-examples.md
+++ b/docs/source/config-examples.md
@@ -35,6 +35,9 @@ if not os.path.exists(ssl_dir):
     os.makedirs(ssl_dir)
 
 
+# Allows multiple single-server per user
+c.JupyterHub.allow_multiple_servers = False
+
 # https on :443
 c.JupyterHub.port = 443
 c.JupyterHub.ssl_key = pjoin(ssl_dir, 'ssl.key')
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -163,7 +163,7 @@ class UserAPIHandler(APIHandler):
 class UserServerAPIHandler(APIHandler):
     """Create and delete single-user servers
     
-    This handler should be used when c.JupyterHub.allow_multiple_servers = False
+    This handler should be used when c.JupyterHub.allow_named_servers = False
     """
     @gen.coroutine
     @admin_or_self
@@ -198,15 +198,17 @@ class UserServerAPIHandler(APIHandler):
         self.set_status(status)
 
 
-class UserCreateMultiServerAPIHandler(APIHandler):
-    """Create multi single-user server
+class UserCreateNamedServerAPIHandler(APIHandler):
+    """Create a named single-user server
     
-    This handler should be used when c.JupyterHub.allow_multiple_servers = True
+    This handler should be used when c.JupyterHub.allow_named_servers = True
     """
     @gen.coroutine
     @admin_or_self
     def post(self, name):
         user = self.find_user(name)
+        if user is None:
+            raise HTTPError(404, "No such user %r" % name)
         if user.running:
             # include notify, so that a server that died is noticed immediately
             state = yield user.spawner.poll_and_notify()
@@ -219,12 +221,12 @@ class UserCreateMultiServerAPIHandler(APIHandler):
         self.set_status(status)
 
 
-class UserDeleteMultiServerAPIHandler(APIHandler):
-    """Delete multi single-user server
+class UserDeleteNamedServerAPIHandler(APIHandler):
+    """Delete a named single-user server
     
     Expect a server_name inside the url /user/:user/servers/:server_name
     
-    This handler should be used when c.JupyterHub.allow_multiple_servers = True
+    This handler should be used when c.JupyterHub.allow_named_servers = True
     """
     @gen.coroutine
     @admin_or_self
@@ -273,7 +275,7 @@ default_handlers = [
     (r"/api/users", UserListAPIHandler),
     (r"/api/users/([^/]+)", UserAPIHandler),
     (r"/api/users/([^/]+)/server", UserServerAPIHandler),
-    (r"/api/users/([^/]+)/servers", UserCreateMultiServerAPIHandler),
-    (r"/api/users/([^/]+)/servers/([^/]+)", UserDeleteMultiServerAPIHandler),
+    (r"/api/users/([^/]+)/servers", UserCreateNamedServerAPIHandler),
+    (r"/api/users/([^/]+)/servers/([^/]+)", UserDeleteNamedServerAPIHandler),
     (r"/api/users/([^/]+)/admin-access", UserAdminAccessAPIHandler),
 ]
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -76,7 +76,7 @@ class UserListAPIHandler(APIHandler):
 
 def admin_or_self(method):
     """Decorator for restricting access to either the target user or admin"""
-    def m(self, name):
+    def m(self, name, *args, **kwargs):
         current = self.get_current_user()
         if current is None:
             raise web.HTTPError(403)
@@ -86,7 +86,7 @@ def admin_or_self(method):
         # raise 404 if not found
         if not self.find_user(name):
             raise web.HTTPError(404)
-        return method(self, name)
+        return method(self, name, args, kwargs)
     return m
 
 class UserAPIHandler(APIHandler):
@@ -160,10 +160,12 @@ class UserAPIHandler(APIHandler):
         self.write(json.dumps(self.user_model(user)))
 
 
-class UserServerAPIHandler(APIHandler):
+class UserCreateServerAPIHandler(APIHandler):
+
     @gen.coroutine
     @admin_or_self
     def post(self, name):
+            
         user = self.find_user(name)
         if user.running:
             # include notify, so that a server that died is noticed immediately
@@ -176,9 +178,12 @@ class UserServerAPIHandler(APIHandler):
         status = 202 if user.spawn_pending else 201
         self.set_status(status)
 
+
+class UserDeleteServerAPIHandler(APIHandler):
+
     @gen.coroutine
     @admin_or_self
-    def delete(self, name):
+    def delete(self, name, server_name):
         user = self.find_user(name)
         if user.stop_pending:
             self.set_status(202)
@@ -222,6 +227,7 @@ class UserAdminAccessAPIHandler(APIHandler):
 default_handlers = [
     (r"/api/users", UserListAPIHandler),
     (r"/api/users/([^/]+)", UserAPIHandler),
-    (r"/api/users/([^/]+)/server", UserServerAPIHandler),
+    (r"/api/users/([^/]+)/servers", UserCreateServerAPIHandler),
+    (r"/api/users/([^/]+)/servers/([^/]+)", UserDeleteServerAPIHandler),
     (r"/api/users/([^/]+)/admin-access", UserAdminAccessAPIHandler),
 ]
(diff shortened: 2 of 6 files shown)


jupyterhub/orm.py
```diff
@@ -64,6 +64,8 @@ class Server(Base):
     """
     __tablename__ = 'servers'
     id = Column(Integer, primary_key=True)
+    
+    name = Column(Unicode(32), unique=True) # this is a unique reference to the server, probably a UUID4
```
barrachri: So I think the main problem is to decide what we want to use as server/{name}

UUID, primary key (id), or something else ?

About the constraint you are right, but we don't have the name of the user inside the Server table so I think we cannot put like Unique(user_name, server_name).
But as you said we can mix both name together and put that field like  unique.

jupyterhub/user.py
```diff
@@ -196,11 +199,21 @@ def url(self):
...
+        
+        Because there could be more then one server per user
+        each server has to have a unique reference (UUID4)
+        
+        base_url is built using user's base url and adding /servers/{name}
+        where name is the server uuid urlsafed
+        """
         db = self.db
+        server_uuid = uuid4()
+        server_url = urlsafe_b64encode(server_uuid.bytes)
```
barrachri: Is uuid.uuid4().hex url safe ?

I am not so sure this is why I decided to use urlsafe.

jupyterhub/user.py
```diff
@@ -196,11 +199,21 @@ def url(self):
...
+        """Start the user's spawner
+        
+        Because there could be more then one server per user
+        each server has to have a unique reference (UUID4)
+        
+        base_url is built using user's base url and adding /servers/{name}
+        where name is the server uuid urlsafed
+        """
         db = self.db
+        server_uuid = uuid4()
```
barrachri: You mean using the server.id ?

jupyterhub/orm.py
minrk: I think we can probably leave the constraint out of the database, and enforce it only at the code level. `Server.name` should probably be human-readable, and doesn't need to be unique. I think it's fine for several Servers to have no name, for instance.

jupyterhub/user.py
minrk: I don't. I mean a counter per-user, e.g. along the lines of:

```python
def default_name(user):
    """Return the default name for a new server for a given user.

    Will be the first available integer string, e.g. '1' or '2'.
    """
    existing_names = { server.name for server in user.servers }
    # if there are 5 servers, count from 1 to 6
    for n in range(1, len(existing_names) + 2):
        name = str(n)
        if name not in existing_names:
            return name
    raise RuntimeError("It should be impossible to get here")
```

jupyterhub/user.py
minrk: hex is url-safe. But it's also unique on a scale we really don't need to deal with (users aren't going to have 2^10 servers). See the above comment about using simple counters.

jupyterhub/user.py
barrachri: ok

jupyterhub/user.py
barrachri: ok

jupyterhub/orm.py
barrachri: +1 on keeping db constraint inside the db.

Keep server_url unique (practically /user/{name}/server/{name}) I think would be enough.

jupyterhub/orm.py
```diff
@@ -64,10 +64,12 @@ class Server(Base):
     """
     __tablename__ = 'servers'
     id = Column(Integer, primary_key=True)
+    
+    name = Column(Unicode(32)) # must be unique between user's servers
     proto = Column(Unicode(15), default='http')
     ip = Column(Unicode(255), default='')  # could also be a DNS name
     port = Column(Integer, default=random_port)
-    base_url = Column(Unicode(255), default='/')
+    base_url = Column(Unicode(255), unique=True)
```
minrk: This column also doesn't need to be unique, and the default should be left here. `Server` is used to store any ip/port combos, so it's only the entirety of ip+port+base_url that would need to be unique, which is not important to enforce at the db level.

barrachri: done :)

jupyterhub/user.py
```diff
@@ -196,11 +199,23 @@ def url(self):
...
         db = self.db
+        if 'server_name' in options:
+            server_name = options['server_name']
+        else:
+            server_name = default_server_name(self)
         server = orm.Server(
+            name = server_name,
             cookie_name=self.cookie_name,
-            base_url=self.base_url,
+            base_url=url_path_join(self.base_url, 'server', server_name),
```
minrk: Just `self.base_url, server_name`, not 'server' in the middle.

jupyterhub/apihandlers/users.py
```diff
@@ -209,14 +218,16 @@ def post(self, name):
...
+        # cause service doesn't have 'other_user_cookies'
         if getattr(current, 'other_user_cookies', None) is not None:
             current.other_user_cookies.add(name)
 
 
 default_handlers = [
     (r"/api/users", UserListAPIHandler),
     (r"/api/users/([^/]+)", UserAPIHandler),
-    (r"/api/users/([^/]+)/server", UserServerAPIHandler),
+    (r"/api/users/([^/]+)/servers", UserCreateServerAPIHandler),
```
minrk: For future: we'll probably want to keep `/server` endpoints for the default server. That can be restored in a new PR, though.

We'll probably also want to keep the notion of a server with no name (`name=''`), with no prefix, since most of the time server-names won't be used.

barrachri: I understand the need for this but considering the Api is not very "consistent" ?

You would have different endpoints doing almost the same.

minrk: > I understand the need for this but considering the Api is not very "consistent" ?

Sorry, I don't quite know what you are asking here.

> You would have different endpoints doing almost the same.

Yes, that's okay. We may be able to handle it with a simple redirect. We just need to be mindful of users.


We need to think carefully about the single-server case, since that's going to continue to be the vast majority of deployments. It's important that people be able to use JupyterHub with:

- never having a server with a name
- only `/user/:name` URL ever proxied to a single server
- never allowing a second server to start

The simplest case is the most important, and we need to be very careful that added features don't make it more complex to use JupyterHub in this simple mode, which should likely remain the default.

barrachri: > Sorry, I don't quite know what you are asking here.
Nevermind :)

> never having a server with a name
I would think as "never have to specify a name", so would be /user/{name}/1 by default.
What do you think ?

> only /user/:name URL ever proxied to a single server
what do you mean with this ?

> never allowing a second server to start
Mmmm... this should be add to the config, like enable_multiserver = True or False

minrk: I mean we should still allow the case where the user's URL is `/user/:name/`, *not* `/user/:name/1/`.

> Mmmm... this should be add to the config, like enable_multiserver = True or False

Yes, I think we are going to need an `allow_multiple_servers` flag.

barrachri: > I mean we should still allow the case where the user's URL is /user/:name/, not /user/:name/1/.

How do you think to handle this ?

Because in the case you start with /user/{name} (first server) and then you add a new server, what would be the name ? /user/{name}/1  or 2 ?

It has more sense to me /user/:name/:server_name (maybe more consistent) but probably I don't see a lot of compatibility problems :)

minrk: It would only be allowed when multiple servers are not allowed, so there cannot be ambiguity between `/user/name` and `/user/name/1/`. So the condition is either:

- require exactly one server and that it has no name
- require multiple server and that they have names

barrachri: Thanks @minrk.

I'll send a new PR considering this.

barrachri:
any idea why it fails with Python 3.5 and with the subdomain ? :(

willingc:
Hi @barrachri, I kicked Travis on the two failing builds and they failed again. What's odd is that I would expect if 3.5 failed so would 3.4 or 3.6.

I wonder if you do an interactive rebase or apply all of your changes on top of current HEAD of master if the changes that you made will be easier to review at a glance. Happy to help walk you through how to clean up your branch if you are interested.

I'll give this a closer look in the morning too.

barrachri:
@willingc I made some order, sorry for the mess inside the commits 👍 

Back to the tests...now everything works apart from the nighty version and subdomain -_-

willingc:
Thanks @barrachri :sunny: No worries on the commits. I just felt it might be easier for all of us to look at and troubleshoot. Thanks.

I'll run the tests in the morning and pass along any insights that I may have. Keep up the good work :smile:

jupyterhub/apihandlers/users.py
```diff
@@ -193,6 +197,52 @@ def delete(self, name):
...
+
+class UserCreateMultiServerAPIHandler(APIHandler):
+    """Create multi single-user server
+    
+    This handler should be used when c.JupyterHub.allow_multiple_servers = True
+    """
+    @gen.coroutine
+    @admin_or_self
+    def post(self, name):
+        user = self.find_user(name)
```
willingc: What happens if a user is not found?

barrachri: I guess is not really possible.

I tried to mimic this: https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/apihandlers/users.py#L163

With the only thing that doesn't check if a server is already running, but for sure it needs more work this part.

willingc:
I reran Travis this morning and all are green now @barrachri :smile:

barrachri:
Random test results :D

willingc:
@barrachri Do you have permissions to restart your tests on Travis? If not, @minrk or I should see if we can give you access to do so.

barrachri:
@willingc I don't see any button like "restart" in Travis. I guess you have to give me the access.

jupyterhub/apihandlers/users.py
```diff
@@ -193,6 +197,52 @@ def delete(self, name):
...
+            if state is None:
+                raise web.HTTPError(400, "%s's server is already running" % name)
+
+        options = self.get_json_body()
+        yield self.spawn_single_user(user, options=options)
+        status = 202 if user.spawn_pending else 201
+        self.set_status(status)
+
+
+class UserDeleteMultiServerAPIHandler(APIHandler):
```
minrk: Let's called this and above `...NamedServer...` instead of `MultiServer`

docs/source/config-examples.md
```diff
@@ -35,6 +35,9 @@ if not os.path.exists(ssl_dir):
     os.makedirs(ssl_dir)
 
 
+# Allows multiple single-server per user
+c.JupyterHub.allow_multiple_servers = False
```
minrk: Probably want `= True` for this example instead of False

jupyterhub/apihandlers/users.py
minrk: @barrachri it was not possible before, but it is now due to admin services. This can add

```python
if user is None:
    raise HTTPError(404, "No such user %r" % name)
```

willingc:
@minrk I couldn't figure out how to add @barrachri to Travis to restart flaky tests without giving push access on this repo.

minrk:
@willingc I think that's the only way. PR authors who are not committers can only retrigger by pushing a new commit to the branch. I sometimes do this with an empty `--amend` on the last commit:

    git commit --amend
    git push mine -f

willingc:
Thanks @minrk. @barrachri, if you need us to kick Travis to rerun the tests, feel free to @ mention me.

jupyterhub/apihandlers/users.py
willingc: I wonder if it would be better to move away from `allow_multiple_servers` too and use `allow_named_servers` instead.

barrachri: what about `UserCreateMultiServerAPIHandler` ?
do you prefer to keep the this name or `NamedServer` ?

willingc: Personally, I prefer `NamedServer` over `MultiServer` so I would change `UserCreateMultiServerAPIHandler` to `UserCreateNamedServerAPIHandler`.

`NamedServer` will likely be more intuitive in future documentation. For example, I believe:

> You may connect to multiple Named Servers.

is more straightforward than:

> You may connect to multiple Single User Servers.

which can be a bit too much `multiple` and `single` strung together in a paragraph.

barrachri:
This is the 6th time that I run the tests....

minrk:
Thanks, @barrachri!

willingc:
Congrats @barrachri. Nice to see this get merged :cake: :cookie: 🍨

PR #944, opened 2017-01-19
check name collisions on rename

### LINKED ISSUE #941: API /user/{name} ambiguous
Both PATCH and POST method accept this json body:

{
  "name":  "nameOFtheuser",
  "admin": true
}

The hub will use the {name} inside the URL but if you try to call the endpoint with a PATCH method and different names (body and url) you will get 

 sqlalchemy.exc.InvalidRequestError: This Session's transaction has been rolled back due to a previous exception during flush.]

I think it would better to accept only one name and not both.

I propose to remove the name from the json body and return an error in case is provided.

https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/apihandlers/users.py#L152
https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/apihandlers/base.py#L141
https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/apihandlers/base.py#L112

barrachri:
It checks if the new name provided is unique, if yes it raises an error.

It doesn't raise an error in case names inside json and {name} are the same.

closes #941

### DIFF
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -151,8 +151,8 @@ class UserAPIHandler(APIHandler):
         data = self.get_json_body()
         self._check_user_model(data)
         # check if the new name is already taken
-        new_username = self.find_user(data['name']) if 'name' in data else None
-        if new_username is not None:
+        new_name = self.find_user(data['name']) if 'name' in data else None
+        if new_name is not None:
             # check if the provided name inside the json is the same of the url
             if name != data['name']:
                 raise web.HTTPError(400, "User %s already exists, username must be unique" % data['name'])
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -150,11 +150,9 @@ class UserAPIHandler(APIHandler):
             raise web.HTTPError(404)
         data = self.get_json_body()
         self._check_user_model(data)
-        # check if the new name is already taken
-        new_name = self.find_user(data['name']) if 'name' in data else None
-        if new_name is not None:
-            # check if the provided name inside the json is the same of the url
-            if name != data['name']:
+        if 'name' in data and data['name'] != name:
+            # check if the new name is already taken:
+            if self.find_user(data['name']):
                 raise web.HTTPError(400, "User %s already exists, username must be unique" % data['name'])
         for key, value in data.items():
             setattr(user, key, value)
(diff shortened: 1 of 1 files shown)


jupyterhub/apihandlers/users.py
```diff
@@ -150,6 +150,12 @@ def patch(self, name):
             raise web.HTTPError(404)
         data = self.get_json_body()
         self._check_user_model(data)
+        # check if the new name is already taken
+        new_name = self.find_user(data['name']) if 'name' in data else None
+        if new_name is not None:
```
minrk: `new_name` will be a `User` object, which is a bit confusing. How about:

```python
if 'name' in data:
    # check if the new name is already taken:
    if self.find_user(data['name']):
        raise ...
```

barrachri: but you have to put another if under if.self to check if name and data['name'] are the same.....

minrk: The presence of the `name` field in the request indicates that a name change is requested. It doesn't make sense to ask the user to change its name to the current value.

barrachri: but basically if you patch /hub/api/user/chris with this body

{
'name': 'chris',
'admin': true
}

you will get User %s already exists, username must be unique

minrk: How about this:

```python
if 'name' in data and data['name'] != name:
    # check if the new name is already taken:
    if self.find_user(data['name']):
        raise ...
```

barrachri: Perfect!

jupyterhub/apihandlers/users.py
```diff
@@ -150,6 +150,10 @@ def patch(self, name):
             raise web.HTTPError(404)
         data = self.get_json_body()
         self._check_user_model(data)
+        if 'name' in data and data['name'] != name:
+            # check if the new name is already taken:
```
willingc: Perhaps @barrachri:

`# check if the new name is already in db`

jupyterhub/apihandlers/users.py
```diff
@@ -209,7 +213,7 @@ def post(self, name):
             raise web.HTTPError(400, "%s's server is not running" % name)
         self.set_server_cookie(user)
         # a service can also ask for a user cookie
-        # this code prevent to raise an error
+        # this code prevents to raise an error
```
willingc: @minrk: For my own understanding. Is this just preventing an attribute error if `current.other_user_cookies` does not exist?

`# this code prevents raising an AttributeError`

barrachri: If a service ask for user cookie it prevents to raise an error cause service doesn't have `other_user_cookies`.

PR #949, opened 2017-01-20
Refactor test classes used to mock services

willingc:
- For classes that inherit from the abstract base class, `web.RequestHandler`, add methods that were previously not implemented.

- Remove unused imports

### DIFF
--- a/jupyterhub/tests/mockservice.py
+++ b/jupyterhub/tests/mockservice.py
@@ -1,7 +1,9 @@
-"""Mock service for testing
+"""Helpers for testing a Service
 
-basic HTTP Server that echos URLs back,
-and allow retrieval of sys.argv.
+Each handler functions as a basic HTTP Server that echos URLs back,
+and allows retrieval of sys.argv.
+
+These handlers are useful when testing a mock Service.
 """
 
 import json
@@ -15,20 +17,20 @@ from jupyterhub.services.auth import HubAuthenticated
 
 
 class EchoHandler(web.RequestHandler):
-    """Handler that mocks a request to the tornado webserver"""
+    """Reply to an HTTP request with the path of the request."""
     def get(self):
         self.write(self.request.path)
 
 
 class EnvHandler(web.RequestHandler):
-    """Handler that mocks a request using header and environment settings"""
+    """Reply to an HTTP request with the service's environment as JSON."""
     def get(self):
         self.set_header('Content-Type', 'application/json')
         self.write(json.dumps(dict(os.environ)))
 
 
 class APIHandler(web.RequestHandler):
-    """Handler that mocks a request using an API token"""
+    """Relay API requests to the Hub's API using the service's API token."""
     def get(self, path):
         api_token = os.environ['JUPYTERHUB_API_TOKEN']
         api_url = os.environ['JUPYTERHUB_API_URL']
@@ -41,7 +43,7 @@ class APIHandler(web.RequestHandler):
 
 
 class WhoAmIHandler(HubAuthenticated, web.RequestHandler):
-    """Handler that mocks a request with an authenticated user"""
+    """Reply with the name of the user who made the request."""
     @web.authenticated
     def get(self):
         self.write(self.get_current_user())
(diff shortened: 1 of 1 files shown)


minrk:
Cool! What prompted this? Does `data_received` get called somewhere?

willingc:
I've been thinking for a while about improving the test coverage and the speed of the test runs. I was playing around with some tools and metrics, i.e.[codescene](https://codescene.io/projects/451/jobs/757/results/code/hotspots/all) on my fork.

Before we get too deep in the weeds for sharing and its API, I wanted to clean up (formatting, understandability vs. changing the tests) the tests a bit first. Perhaps improve coverage on app.py and api.

PR #956 and #955 are a start to the test formatting/understandability updates.

willingc:
@minrk Looking back at the Tornado source, the subclassing only has to subclass one of the entrypoints from the abstract class. As we've done that with `get`, I think that `data_received` does not need to be added at this time.

jupyterhub/tests/mockservice.py
```diff
@@ -15,41 +15,33 @@
 
 
 class EchoHandler(web.RequestHandler):
-    def data_received(self, chunk):
-        pass
-
+    """Handler that mocks a request to the tornado webserver"""
```
minrk: This handler doesn't mock a request. More accurate:

> reply to HTTP requests with the path of the request

jupyterhub/tests/mockservice.py
```diff
@@ -15,41 +15,33 @@
...
     def get(self):
         self.write(self.request.path)
 
 
 class EnvHandler(web.RequestHandler):
-    def data_received(self, chunk):
-        pass
-
+    """Handler that mocks a request using header and environment settings"""
```
minrk: Also doesn't mock a request:

> Reply with the service's environment as JSON.

jupyterhub/tests/mockservice.py
```diff
@@ -15,41 +15,33 @@
...
     def get(self):
         self.set_header('Content-Type', 'application/json')
         self.write(json.dumps(dict(os.environ)))
 
 
 class APIHandler(web.RequestHandler):
-    def data_received(self, chunk):
-        pass
-
+    """Handler that mocks a request using an API token"""
```
minrk: > Relay API requests to the Hub's API using the service's API token.

jupyterhub/tests/mockservice.py
```diff
@@ -15,41 +15,33 @@
...
         r.raise_for_status()
         self.set_header('Content-Type', 'application/json')
         self.write(r.text)
 
 
 class WhoAmIHandler(HubAuthenticated, web.RequestHandler):
-    def data_received(self, chunk):
-        pass
-
+    """Handler that mocks a request with an authenticated user"""
```
minrk: > Reply with the name of the authenticated user

willingc:
Thanks @minrk for the clarifications. I've updated the file docstring to better clarify that the handlers. Please let me know if that needs a change.

jupyterhub/tests/mockservice.py
```diff
@@ -1,44 +1,49 @@
-"""Mock service for testing
+"""Helpers for testing a Service
```
minrk: This script *is* a mock service for testing, it does not provide helpers. It has the main function at the bottom. It is a complete application to be started by the `mockservice` fixtures in `conf.py`.

How about this docstring:

> Mock service for testing Service integration
>
> A JupyterHub service running a basic HTTP server.
> Used by the mockservice fixtures.
>
> Handlers allow:
> - echoing proxied URLs back
> - retrieving service's environment variables
> - testing service's API access to the Hub

minrk:
One last comment on the module docstring, then 👍

jupyterhub/tests/mockservice.py
willingc: Thanks @minrk for both the clarification and the suggested docstring. Understanding other folks' code is hard. Hoping this will help the next devs. Updating...

minrk: It certainly is. This is a great help, thanks!

PR #953, opened 2017-01-20
Rename test class to stub class to allow pytest collection

### LINKED ISSUE #952: Determine why codecov tool is incorrectly reporting coverage
The past few days, codecov tool is reporting coverage ~55% instead of the more recent ~85%. Will investigate if codecov tool has new settings causing this spurious drop.

willingc:
Fixes #952 

pytest throws a warning to indicate it will not collect test classes with an inherited `__init__` method. 

This PR renames a test class from TestSingleUserSpawner to StubSingleUserSpawner to satisfy pytest that the mocked class is not being tested but instead simply sets up the test case. [See reference re: fix by pytest core dev](http://hackebrot.github.io/pytest-tricks/show_pytest_warnings/).

Individual tests are renamed to reflect the class name change.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -13,4 +13,5 @@ exclude_lines =
 ignore_errors = True
 omit =
     jupyterhub/tests/*
+    jupyterhub/alembic/*
     */site-packages/*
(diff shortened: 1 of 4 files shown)


willingc:
Moving this back to WIP since latest two commits (which should only impact report output) uncovered some interesting socket errors in different test runs on different Python versions.

willingc:
Thanks @minrk. I'll keep an eye on the spurious errors.

PR #960, opened 2017-01-24
Added user_for_token method, moved r check into its own function

DeepHorizons:
This pull request adds the `user_for_token` method to the HubAuth class. This enables easy access to authenticate and get a user by their token.

The response check was moved into its own function to remove duplicating code.

### DIFF
--- a/jupyterhub/services/auth.py
+++ b/jupyterhub/services/auth.py
@@ -156,8 +156,9 @@ class HubAuth(Configurable):
             app_log.warning("Failed to check authorization: [%i] %s", r.status_code, r.reason)
             raise HTTPError(500, "Failed to check authorization")
         else:
+            data = r.json()
             app_log.debug("Received request from Hub user %s", data)
-            return r.json()
+            return data
 
     def user_for_cookie(self, encrypted_cookie, use_cache=True):
         """Ask the Hub to identify the user for a given cookie.
--- a/jupyterhub/services/auth.py
+++ b/jupyterhub/services/auth.py
@@ -138,6 +138,27 @@ class HubAuth(Configurable):
     def _cookie_cache(self):
         return _ExpiringDict(self.cookie_cache_max_age)
 
+    def _check_response_for_authorization(self, r):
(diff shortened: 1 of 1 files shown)


jupyterhub/services/auth.py
```diff
@@ -138,6 +138,27 @@ class HubAuth(Configurable):
...
+            app_log.error("I don't have permission to verify cookies, my auth token may have expired: [%i] %s", r.status_code, r.reason)
+            raise HTTPError(500, "Permission failure checking authorization, I may need a new token")
+        elif r.status_code >= 500:
+            app_log.error("Upstream failure verifying auth token: [%i] %s", r.status_code, r.reason)
+            raise HTTPError(502, "Failed to check authorization (upstream problem)")
+        elif r.status_code >= 400:
+            app_log.warning("Failed to check authorization: [%i] %s", r.status_code, r.reason)
+            raise HTTPError(500, "Failed to check authorization")
+        else:
+            app_log.debug("Received request from Hub user %s", data)
```
minrk: `data` here isn't declared. Perhaps:

```python
data = r.json()
app_log.debug(...)
return data
```

jupyterhub/services/auth.py
```diff
@@ -138,6 +138,27 @@ class HubAuth(Configurable):
...
+    def _check_response_for_authorization(self, r):
+        """Verify the response for authorizing a user
+
+        Raises an error if the response failed, otherwise returns the json
+        """
+        if r.status_code == 404:
+            app_log.warning("No Hub user identified for request")
+            data = None
+        elif r.status_code == 403:
+            app_log.error("I don't have permission to verify cookies, my auth token may have expired: [%i] %s", r.status_code, r.reason)
```
minrk: Update message to be more general:

"I don't have permission to check authorization with JupyterHub; my auth..."

DeepHorizons:
@minrk Fixed the message and the logging error.

minrk:
Great, thanks!

PR #961, opened 2017-01-25
Enable token authentication in HubAuth

### LINKED ISSUE #891: Support token auth in single-user servers
The default LoginHandler will support token auth in 4.3: https://github.com/jupyter/notebook/pull/1831 . Since JupyterHub replaces the LoginHandler, token auth won't work with JupyterHub without modifications. We should add support for the Authorization header in the single-user server so that JupyterHub tokens can be used to access single-user servers via the Authorization header.

This will also likely drive demand for creation of new API tokens via UI, but that can be a separate step.

cf https://github.com/jupyterhub/jupyterhub-deploy-docker/issues/30

minrk:
follow-up to #960
closes #891

- further consolidate some duplicate code asking the Hub to identify a user
- enable token authentication in `.get_user()`
- fully support notebook-4.3 API in single-user for token auth
- add missing `.get_login_url` that tornado needs for redirecting to the Hub for login

### DIFF
--- a/jupyterhub/services/auth.py
+++ b/jupyterhub/services/auth.py
@@ -312,6 +312,10 @@ class HubAuthenticated(object):
     def hub_auth(self, auth):
         self._hub_auth = auth
 
+    def get_login_url(self):
+        """Return the Hub's login URL"""
+        return self.hub_auth.login_url
+
     def check_hub_user(self, user_model):
         """Check whether Hub-authenticated user should be allowed.
 
--- a/jupyterhub/services/auth.py
+++ b/jupyterhub/services/auth.py
@@ -8,6 +8,7 @@ HubAuthenticated is a mixin class for tornado handlers that should authenticate 
 """
 
 import os
+import re
 import socket
 import time
 from urllib.parse import quote
@@ -238,6 +239,23 @@ class HubAuth(Configurable):
             use_cache=use_cache,
         )
     
+    auth_header_name = 'Authorization'
+    auth_header_pat = re.compile('token\s+(.+)', re.IGNORECASE)
+
+    def get_token(self, handler):
+        """Get the user token from a request
+
+        - in URL parameters: ?token=<token>
+        - in header: Authorization: token <token>
+        """
+
+        user_token = handler.get_argument('token', '')
+        if not user_token:
+            # get it from Authorization header
+            m = self.auth_header_pat.match(handler.request.headers.get(self.auth_header_name, ''))
+            if m:
+                user_token = m.group(1)
+        return user_token
 
     def get_user(self, handler):
         """Get the Hub user for a given tornado handler.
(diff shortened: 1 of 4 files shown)


jupyterhub/services/auth.py
```diff
@@ -138,14 +139,46 @@ class HubAuth(Configurable):
...
 
-    def _check_response_for_authorization(self, r):
-        """Verify the response for authorizing a user
+    def _check_hub_authorization(self, url, cache_key=None, use_cache=True):
+        """Identify a user with the Hub
+        
+        Args:
+            url (str): The API URL to check the Hub for authorization
+                       (e.g. http://127.0.0.1:8081/hub/api/authorizations/token/abc-def)
+            cache_key (str): The for checking the cache
```
willingc: missing word 'key'

jupyterhub/services/auth.py
```diff
@@ -138,14 +139,46 @@ class HubAuth(Configurable):
...
+            use_cache (bool): Specify use_cache=False to skip cached cookie values (default: True)
+
+        Returns:
+            user_model (dict): The user model, if a user is identified, None if authentication fails.
+        An HTTPError is raised if
```
willingc: stray line?

jupyterhub/services/auth.py
```diff
@@ -138,14 +139,46 @@ class HubAuth(Configurable):
...
+        Verify the response for authorizing a user
 
         Raises an error if the response failed, otherwise returns the json
         """
+        if use_cache:
+            if cache_key is None:
+                raise ValueError("cache_key is required when using cache")
+            cached = self.cookie_cache.get(cache_key)
+            if cached is not None:
+                return cached
```
willingc: Min, Clarifying that I understand what is going on here...

1. `cached` contains the cookie that had been previously cached
2.  What happens on the `return`? Does it continue to the try block or back to what called this private method?

jupyterhub/services/auth.py
```diff
@@ -138,14 +139,46 @@ class HubAuth(Configurable):
...
+                },
+            )
+        except requests.ConnectionError:
+            msg = "Failed to connect to Hub API at %r." % self.api_url
+            msg += "  Is the Hub accessible at this URL (from host: %s)?" % socket.gethostname()
+            if '127.0.0.1' in self.api_url:
+                msg += "  Make sure to set c.JupyterHub.hub_ip to an IP accessible to" + \
+                       " single-user servers if the servers are not on the same host as the Hub."
+            raise HTTPError(500, msg)
+
```
willingc: Very helpful error message :sunny:

jupyterhub/services/auth.py
minrk: `cached` is the cached result of this method, from calling this method before with the same argument. If a cached result is found, return it instead of running the method again. It does not continue after the `return`.

willingc: Okay. I think we're on the same page.

If using caching, this method will check if the cached key exists in the user's cookie. If the cached key exists in the user's cookie, the method returns the cached key to the calling method. Otherwise, try to authenticate the user with an api token and storing the key, if found, to use later.

If not using caching, identify an authorized user by the checking the api token.

willingc:
Thanks for the clarification Min. So for 0.8, we will not cache using cookies but cache in the hub's db, correct?

The latest commit really helps make things more understandable.

Still trying to figure out why there is an extra @ in the failing test url.

minrk:
> So for 0.8, we will not cache using cookies but cache in the hub's db, correct?

Sorry, I could swear I responded to this the other day. This is not a change in behavior, it's consolidating code that got duplicated now that there's two ways to ask the Hub about an authorization. Here's what is happening:

1. A request is made to the HubAuth-using server with a cookie or API token
2. HubAuth asks the Hub who the cookie or token identifies
3. HubAuth caches the Hub's *response* to that query (for a time), so it doesn't have to add a Hub API request to every handler.

PR #966, opened 2017-01-27
support service tokens in HubAuth

### LINKED ISSUE #954: HubAuth does not work with service tokens
It is currently not possible to authenticate a service using the HubAuth class.
When attempting to use a service token to authenticate a service via `/api/authorizations/token/{token}`,
 it returns a 404.

Take a service token and make a request to `/api/authorizations/token/{token}` using the token and the proper authorization header.

A User model is returned about the service, with a `name` and `admin` field.

**What actually happens**
The request returns with a 404

0.7.2

minrk:
- support service tokens in /api/authorizations
- add hub_services to HubAuth
- add 'kind' field to models, which is needed to easily distinguish between users and servers

closes #954

### DIFF
--- a/jupyterhub/services/auth.py
+++ b/jupyterhub/services/auth.py
@@ -12,6 +12,7 @@ import re
 import socket
 import time
 from urllib.parse import quote
+import warnings
 
 import requests
 
@@ -183,7 +184,6 @@ class HubAuth(Configurable):
             raise HTTPError(500, msg)
 
         data = None
-        print(r.status_code)
         if r.status_code == 404:
             app_log.warning("No Hub user identified for request")
         elif r.status_code == 403:
@@ -330,9 +330,19 @@ class HubAuthenticated(object):
                 ...
 
     """
+    hub_services = None # set of allowed services
     hub_users = None # set of allowed users
     hub_groups = None # set of allowed groups
     
+    @property
+    def allow_all(self):
+        """Property indicating that all successfully identified user
+        or service should be allowed.
+        """
+        return (self.hub_services is None
+            and self.hub_users is None
+            and self.hub_groups is None)
+
     # self.hub_auth must be a HubAuth instance.
     # If nothing specified, use default config,
     # which will be configured with defaults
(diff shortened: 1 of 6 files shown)


willingc:
@minrk Is this still WIP or ready for review?

minrk:
I had one thing I wanted to change before merging, and forgot to get back to it. I will finish tomorrow. Thanks!

jupyterhub/tests/test_services_auth.py
```diff
@@ -217,7 +219,8 @@ def finish_thread():
         assert auth.login_url in r.headers['Location']
 
 
-def test_service_cookie_auth(app, mockservice_url):
+def test_hubauth_cookie(app, mockservice_url):
```
willingc: Min, Do we want to start moving to pytest's parameterize to test specific names? It might make for easier additions to the tests going forward and simplify testing when trying to find a bug as we could more easily drop in individual test cases.

minrk: I'm not sure it would help in this particular test case, but in cases where it would, parameterize is a good idea.

minrk:
Changed my mind about what I wanted to change (it was renaming 'kind' to 'type', but now I like 'kind' better). Should be all set now.

jupyterhub/services/auth.py
```diff
@@ -330,9 +330,19 @@ def get(self):
                 ...
 
     """
+    hub_services = None # set of allowed services
```
willingc: Let's update the docstring here.

jupyterhub/services/auth.py
```diff
@@ -352,34 +362,45 @@ def get_login_url(self):
...
+    def check_hub_user(self, model):
+        """Check whether Hub-authenticated user or service should be allowed.
 
         Returns the input if the user should be allowed, None otherwise.
 
         Override if you want to check anything other than the username's presence in hub_users list.
 
         Args:
-            user_model (dict): the user model returned from :class:`HubAuth`
+            model (dict): the user or service model returned from :class:`HubAuth`
```
willingc: Is this user, group or service?

jupyterhub/tests/test_services_auth.py
willingc: For the simple case here, it doesn't matter. For some of the longer tests, it may make sense to parameterize the input test case since I think the test becomes more readable and more useful later on (with multiple services and hubshare adding complexity). The test can remain the same even if there are additional test input needed. For now, I think they are fine as is.

jupyterhub/services/auth.py
minrk: no, user or service. A request can't be identified as a group (at this point, at least), but it can be a *member* of a group (user).

willingc:
Nice work @minrk :sunny:

PR #971, opened 2017-02-08
Add page for requesting tokens

minrk:
Since this is a new page, I took the opportunity to cleanup the nav header a bit, to make it a proper navbar. This removes `admin` from the big button collection and adds it to the nav at the top, where the new `Token` page is added.

I don't imagine the Token page will last too long, as it will probably get folded into a user preferences / advanced page, once we have more than one page. But for now, it's the only action on that page.

New home page:

screen shot 2017-02-08 at 14 20 03

Token request page:

screen shot 2017-02-08 at 14 19 54


minrk:
cc @ellisonbg @jhamrick since this should be useful for admins requesting new tokens.

Things we could add:

- admins requesting tokens on behalf of others

willingc:
@minrk Nice! I'll give it a more thorough review in a bit.

jhamrick:
Awesome, this looks great!

I would indeed be useful too as you mentioned in https://github.com/jupyter/nbgrader/issues/659 if this token could get set automatically in the home directory or something when the server is launched. Maybe on the token request page, it could also save the token to the home directory (in addition to displaying it on the screen)? Or are there reasons I'm not thinking of why we wouldn't want that?

minrk:
> I would indeed be useful too as you mentioned in jupyter/nbgrader#659 if this token could get set automatically in the home directory or something when the server is launched.

It might be tricky to send the token to the single-user server from this page (server may not be running, etc.). There is however one important detail that might make an easier fix: The single-user server already gets a token needed for authenticating with the Hub. It's currently not passed to subprocesses where user code runs because I didn't want to leak tokens, but I realize that doesn't make a terrific amount of sense. #976 simply allows that token env to pass through to subprocesses, so anything you run in a shell in your single-user server should have access to it.

PR #974, opened 2017-02-08
Refactor travis config for more efficient builds

willingc:
- Remove travis_retry since travis does this in most of their builds now by default
- Remove sudo config since false is the default
- Add [fast_finish to matrix](https://docs.travis-ci.com/user/customizing-the-build/#Fast-Finishing)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,6 @@ python:
     - 3.5
     - 3.6
     - nightly
-    # PyPy versions
-    - pypy3
 
 # installing dependencies
 before_install:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,23 +1,29 @@
...
     - npm install -g configurable-http-proxy
     - git clone --quiet --depth 1 https://github.com/minrk/travis-wheels travis-wheels
 install:
     - pip install --pre -f travis-wheels/wheelhouse -r dev-requirements.txt .
+
+# running tests
 script:
-    - travis_retry py.test --cov=jupyterhub jupyterhub/tests -v
+    - if [[ $TRAVIS_PYTHON_VERSION == '3.6' || $TRAVIS_PYTHON_VERSION == 'nightly' ]]; then pytest --maxfail=2 --cov=jupyterhub jupyterhub/tests; fi # for Python 3.6 and above
+    - if [[ $TRAVIS_PYTHON_VERSION == '3.5' || $TRAVIS_PYTHON_VERSION == '3.4' || $TRAVIS_PYTHON_VERSION == '3.3' ]]; then py.test --maxfail=2 --cov=jupyterhub jupyterhub/tests; fi # for < 3.5
```
minrk: Why are these two lines different?

.travis.yml
```diff
@@ -1,23 +1,29 @@
...
 python:
-    - 3.7-dev
-    - 3.6
-    - 3.5
-    - 3.4
     - 3.3
+    - 3.4
+    - 3.5
+    - 3.6
+    - nightly
```
minrk: I think it's better to use the descending order, because nightly will often take the longest, so it should start first.

willingc: Good point :-)

.travis.yml
willingc: I pulled this from an older travis config on another project. It was to capture when py.test and pytest were being migrated to pytest command for small performance gains based on what Python version was in the Travis VM. Now that the migration is done in pytest, I think it would be better to have them as one. The performance gain was minimal anyway (one less pip install --upgrade pytest on the older pythons at the time).

willingc:
Thanks for the review @minrk. Will make the changes and push later today.

willingc:
@minrk Going to go ahead and merge this one.

minrk:
Thanks! Sorry I forgot to.

PR #976, opened 2017-02-08
Don't pop token env in single-user servers

minrk:
Users should be allowed to access their own tokens to talk to the Hub API.

I popped the token for vague 'security reasons', but users are allowed to allocate new tokens, so it doesn't make a huge amount of sense to deprive the user's kernels from using the API token set up for their single-user server.

cc @jhamrick this might be the quickest and easiest fix for your nbgrader needs.

### DIFF
--- a/jupyterhub/singleuser.py
+++ b/jupyterhub/singleuser.py
@@ -277,9 +277,9 @@ class SingleUserNotebookApp(NotebookApp):
         api_token = None
         if os.getenv('JPY_API_TOKEN'):
             # Deprecated env variable (as of 0.7.2)
-            api_token = os.environ.pop('JPY_API_TOKEN')
+            api_token = os.environ['JPY_API_TOKEN']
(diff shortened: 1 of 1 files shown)


jhamrick:
Awesome, this sounds like it would fit exactly what I need.

minrk:
If putting it in a file is still desirable, the easiest mechanism is probably to use a custom launch script:

```bash
#!/usr/bin/env bash
# save the token env to my token-needing file
echo "$JUPYTERHUB_API_TOKEN" > ~/.my_hub_token
# custom launch scripts always end the same way:
exec jupyterhub-singleuser $@
```

minrk:
(bonus:  that custom launch script will work for versions of JupyterHub without this patch, since it gets the env before the singleuser script has popped it)

willingc:
Thanks @minrk. Going to open an issue so we remember to document this and the other PR. :sunny:

Issue #851, opened 2016-11-08
Setup a pylintrc for CI checks


yuvipanda:
Setting a pylintrc that does error checks rather than style checks would be really useful. I've been doing it for kubespawner and it's really nice and helped me find a bunch of bugs.

Only style enforcement I might like is docstring enforcement :D

yuvipanda:
I'd like to have something like https://github.com/ipython/ipynb/blob/master/.pylintrc in here. It would require that we fix all the errors listed, which mostly means writing a ton of docstrings. I might take a stab at it.

manics:
We've got [pre-commit with multiple hooks](https://github.com/jupyterhub/jupyterhub/blob/c3ca924ba896bf2da40cab29e2a08784850a945b/.pre-commit-config.yaml).

PR #978, opened 2017-02-09
Add flake8 support

willingc:
@minrk This PR is the first step toward implementing #851 for docstring and style checks.

- Adds flake8 config for general spacing errors and comment style
- Ignores a number of style errors that would introduce changes in long lines and multiline commands
- Excludes all files other than the core jupyterhub source files found in `jupyterhub/jupyterhub`. Does not check tests, services, handlers, and other subdirectories
- Most of the changes are spacing changes. If you dislike any of the style changes, let me know and we can ignore those and I can revert in the PR.


jupyterhub/utils.py
```diff
@@ -115,24 +118,31 @@ def decorated(self, *args, **kwargs):
...
     (no cookies)
     """
     if self.get_current_user_token() is None:
         raise web.HTTPError(403)
 
+
 @auth_decorator
 def authenticated_403(self):
-    """like web.authenticated, but raise 403 instead of redirect to login"""
+    """Decorator for method to raise 403 error instead of redirect to login.
```
minrk: This isn't a sentence, so adding a period at the end isn't correct. In general, I would prefer *not* to add periods to the first lines of docstrings, since they are generally headings, not in-paragraph sentences.

jupyterhub/version.py
```diff
@@ -1,4 +1,4 @@
-"""jupyterhub version info"""
+"""JupyterHub version info."""
```
minrk: No period

.flake8
```diff
@@ -0,0 +1,31 @@
...
+# E201: whitespace after `(`
+# E202: whitespace before `)`
+# E402: module level input not at top of file
+# E501: line too long (82 > 79 characters)
+# E711: comparison to None should be `if cond is None:`
+# E731: do not assign a lambda expression, use a def
+# F401: module imported but unused
+# F811: redefinition of unused `name` from line `N`
+# F841: local variable assigned but never used
+ignore = E123, E124, E128, E201, E202, E402, E501, E711, E731, F401, F811, F841
```
minrk: I'd prefer these broader exclusions:

```
ignore = E,W,C
```

I don't think there are any items in these three categories that should be enforced.

.flake8
```diff
@@ -0,0 +1,31 @@
...
+
+exclude =
+    .cache,
+    .github,
+    docs,
+    examples,
+    jupyterhub/alembic*,
+    jupyterhub/apihandlers*,
+    jupyterhub/handlers*,
+    jupyterhub/services*,
```
minrk: Is there a reason handlers and services are excluded?

jupyterhub/utils.py
willingc: I could go either way on this. The periods are a convention in pep8 and numpy docstrings. I think it's fine without the period too. In general, I'm less concerned about the punctuation and more focused on having at short, descriptive docstrings :-)

minrk:
@willingc thanks!

PR #987, opened 2017-02-13
Make cookie_id be unique and not-null too

### LINKED ISSUE #985: cookie_id column in users table should be NOT NULL and UNIQUE

Accidentally delete the jupyterhub database, and when recreating it from logs make cookie_id for more than one user be empty or the same dummy value.

The db schema complains, telling me 'nuh-uh'.

**What actually happens**

All users who have the same cookie_id value are redirected to the notebook of the user with the lowest id among the users with same cookie_id, leaking private contents of notebook to others.

0.7.2

yuvipanda:
Fixes #985

### DIFF
--- a/jupyterhub/orm.py
+++ b/jupyterhub/orm.py
@@ -381,7 +381,7 @@ class User(Base):
     last_activity = Column(DateTime, default=datetime.utcnow)
 
     api_tokens = relationship("APIToken", backref="user")
-    cookie_id = Column(Unicode(1023), default=new_token)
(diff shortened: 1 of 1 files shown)


yuvipanda:
@minrk is there anything special needed to make this work with the migration script?

minrk:
I think new constraints will only be applied when recreating the table, which doesn't happen on reopen or upgrade. An alembic stage can be added to upgrade in-place, if it's important. I haven't done a new constraint in the upgrade before.

If you went through the process in #985 where the db is deleted and recreated, these constraints would be applied, though, so it's probably enough.

Issue #997, opened 2017-02-22
JUPYTERHUB_SERVICE_PREFIX problem


whitead:
**How to reproduce the issue**

Add the [shared notebook as a service managed configuration example](https://github.com/jupyterhub/jupyterhub/tree/master/examples/service-notebook) to your jupyter configuration with the service name as `formgrader-class101`. Visit https://hub-url/services/formgrader-class101 and you'll see that `custom.css` and `custom.js` give 404s.

**What you expected to happen**

To not get 404s.

**What actually happens**

You get 404s

**Share what version of JupyterHub you are using**

master

Running `jupyter troubleshoot` from the command line, if possible, and posting
its output would also be helpful.

```

$PATH:
	/home/awhite38/.local/bin
	/home/awhite38/bin
	/home/awhite38/.local/bin
	/usr/local/sbin
	/usr/local/bin
	/usr/sbin
	/usr/bin
	/sbin
	/bin
	/usr/games
	/usr/local/games
	/snap/bin

sys.path:
	/home/awhite38/.local/bin
	/home/awhite38/dockerspawner
	/usr/lib/python35.zip
	/usr/lib/python3.5
	/usr/lib/python3.5/plat-x86_64-linux-gnu
	/usr/lib/python3.5/lib-dynload
	/home/awhite38/.local/lib/python3.5/site-packages
	/usr/local/lib/python3.5/dist-packages
	/usr/lib/python3/dist-packages

sys.executable:
	/usr/bin/python3

sys.version:
	3.5.2 (default, Nov 17 2016, 17:05:23) 
	[GCC 5.4.0 20160609]

platform.platform():
	Linux-4.4.0-62-generic-x86_64-with-Ubuntu-16.04-xenial

which -a jupyter:
	/home/awhite38/.local/bin/jupyter
	/home/awhite38/.local/bin/jupyter
	/usr/local/bin/jupyter

```

**Workaround**

The issue is that the urls for `custom.js` and `custom.css`  are `https://hub-url/services/formgrader-class101custom/custom.css` and are missing a forward slash. The following change fixes this:

```diff
@@ -178,7 +178,8 @@ class SingleUserNotebookApp(NotebookApp):
     # defaults for some configurables that may come from service env variables:
     @default('base_url')
     def _base_url_default(self):
-        return os.environ.get('JUPYTERHUB_SERVICE_PREFIX') or '/'
[...]

PR #1003, opened 2017-02-27
Ensure that SERVICE_FIX has trailing slash

whitead:
Fixes Issue #997


whitead:
Not sure what's going on with that one test..? Random timeout?

minrk:
Thanks! It was indeed an intermittent issue, which happens with the tests sometimes.

PR #1025, opened 2017-03-17
correct self.allow_multiple_servers to self.allow_named_servers

barrachri:
We decided to use allow_named_servers instead of allow_multiple_servers.


barrachri:
It should work now.

PR #1027, opened 2017-03-17
Teach FormSpawner to handle query next=/path/to/landing

mgeplf:
* before, if /user-redirect/ was used, or if ?next=/path/
  and there was a FormSpawner, the query param would be lost


mgeplf:
I'm not sure that the test failures relate to changes I have made.  It seems that they are happening `hub.init_secrets()` due to encoding errors:

```
>                   secret = binascii.a2b_hex(text_secret)
E                   binascii.Error: Non-hexadecimal digit found
```

minrk:
Thanks!

PR #1040, opened 2017-03-24
Add rotating log info by @minrk to docs

### LINKED ISSUE #1038: Daily Log Files
Hi,

Is it possible to create daily jupyterhub log files. At the moment we direct the log >file when we start the hub.

Thank you

willingc:
Closes #1038.

I've seen this asked several times so let's include in the docs under the "how do I" FAQ section. Content is from @minrk's response on #1038.


minrk:
Thanks!

PR #1043, opened 2017-03-27
run subdomain test with localhost.jovyan.org

minrk:
instead of relying on xip.io, which seems to be flaky sometimes

Ultimately, it's the same thing: a wildcard domain that always resolves to 127.0.0.1,
but since we don't need xip's fancy domain-for-any-ip, we can use a static IP on a domain we control.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,4 +26,4 @@ matrix:
   fast_finish: true
   include:
     - python: 3.6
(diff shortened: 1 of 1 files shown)


willingc:
Going to merge this since #1044 should fix the test errors. Thanks @minrk!

PR #1044, opened 2017-03-27
match the whole content when checking if cookie secret is hex

minrk:
fixes spurious message with base64 cookie secrets that start with hex subset.

Also fix the fact that the tests were still using a base64 cookie secret, which was the source of the messages.


willingc:
Thanks @minrk

PR #1046, opened 2017-03-27
Add default behavior of whitelist to FAQ in docs

### LINKED ISSUE #753: What is the default behaviour when no user/group  lists are set ?
Hi,
How can we describe the default behaviour of JupyterHub when none of the lists: admin_users, white_list, and white_groupList is set ?

willingc:
Closes #753


minrk:
Thanks!

PR #1050, opened 2017-03-31
Clarify admin users and access in docs

willingc:
Based on @parente's question on gitter, I've gone ahead and clarified the `admin_users` and `admin_access` a bit more in the docs.

### DIFF
--- a/docs/source/getting-started.md
+++ b/docs/source/getting-started.md
@@ -411,19 +411,26 @@ started.
 
 ### Managing Hub administrators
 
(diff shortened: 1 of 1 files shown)


parente:
Beat me to it. I was going to add a blurb about how an admin token does not implicitly grant access to notebooks servers via the API  even when the admin_access setting is true

parente:
LGTM though. Let's get this in and we can send a PR against the API docs separately.

PR #1068, opened 2017-04-08
[DOC] gallery deployments, improve UCSD

zonca:
added post titles and added one more post.


willingc:
Thanks @zonca for the improvement and your continued great work with JupyterHub.

PR #1069, opened 2017-04-08
Prevent crashes due to deleted system users

### LINKED ISSUE #1060: Deleting OS users causes Jupyterhub server to fail to start
Jupyterhub does not properly check if a user exists when it attempts to start. If an OS user is deleted, Jupyterhub will crash until those users have been removed from Jupyterhub's database. 

* Install Jupyterhub according to https://github.com/jupyterhub/jupyterhub/wiki/Using-sudo-to-run-JupyterHub-without-root-privileges. 
* Create a test user (`useradd test`). 
* Log in, make a notebook or three with said test user. 
* Delete test user (`userdel -r test`).
* Restart jupyterhub.

Jupyterhub should start successfully.

**What actually happens**

Jupyterhub will now crash on start.

0.7.2

**HOW TO FIX**

Check for non-existent users on Jupyterhub initialization, delete them from Jupyterhub's `users` table.

```
sqlite3 jupyterhub.sqlite
DELETE FROM users WHERE name LIKE 'test%';
.quit
# jupyterhub now starts normally
```

**Logs (from a jupyterhub systemd service)**

Apr 06 14:00:07 some_host.ca sudo[8433]: jupyterhub : unknown user: testme
Apr 06 14:00:07 some_host.ca jupyterhub[8429]: sudo: unknown user: testme
Apr 06 14:00:07 some_host.ca jupyterhub[8429]: sudo: unable to initialize policy plugin
Apr 06 14:00:07 some_host.ca jupyterhub[8429]: [E 2017-04-06 14:00:07.756 JupyterHub app:1527]
Apr 06 14:00:07 some_host.ca jupyterhub[8429]: Traceback (most recent call last):
Apr 06 14:00:07 some_host.ca jupyterhub[8429]: File "/opt/python/anaconda3/lib/python3.6/site-packages/jupyterhub/app.py", line 1524, in launch_instance_async
Apr 06 14:00:07 some_host.ca jupyterhub[8429]: yield self.initialize(argv)
Apr 06 14:00:07 some_host.ca jupyterhub[8429]: File "/opt/python/anaconda3/lib/python3.6/types.py", line 184, in throw
Apr 06 14:00:07 some_host.ca jupyterhub[8429]: return self.__wrapped.throw(tp, *rest)
Apr 06 14:00:07 some_host.ca jupyterhub[8429]: File "/opt/python/anaconda3/lib/python3.6/site-packages/jupyterhub/app.py", line 1315, in initialize
Apr 06 14:00:07 some_host.ca jupyterhub[8429]: yield self.init_spawners()
Apr 06 14:00:07 some_host.ca jupyterhub[8429]: File "/opt/python/anaconda3/lib/python3.6/site-packages/jupyterhub/app.py", line 1087, in init_spawners
Apr 06 14:00:07 some_host.ca jupyterhub[8429]: status = yield spawner.poll()
Apr 06 14:00:07 some_host.ca jupyterhub[8429]: File "/opt/python/anaconda3/lib/python3.6/site-packages/jupyterhub/spawner.py", line 834, in poll
Apr 06 14:00:07 some_host.ca jupyterhub[8429]: alive = yield self._signal(0)
Apr 06 14:00:07 some_host.ca jupyterhub[8429]: File "/opt/python/anaconda3/lib/python3.6/site-packages/sudospawner/spawner.py", line 90, in _signal
Apr 06 14:00:07 some_host.ca jupyterhub[8429]: reply = yield self.do('kill', pid=self.pid, signal=sig)
Apr 06 14:00:07 some_host.ca jupyterhub[8429]: File "/opt/python/anaconda3/lib/python3.6/site-packages/sudospawner/spawner.py", line 69, in do
Apr 06 14:00:07 some_host.ca jupyterhub[8429]: data_str = data_str[data_str.index('{'):data_str.rindex('}')+1]
Apr 06 14:00:07 some_host.ca jupyterhub[8429]: ValueError: substring not found
Apr 06 14:00:07 some_host.ca systemd[1]: jupyterhub.service: main process exited, code=exited, status=1/FAILURE

minrk:
In three ways:

1. add opt-in `Authenticator.delete_invalid_users`, which will remove any users that fail to be added. Since it's opt-in, the config to enable it is logged when the situation is encountered and the option is not enabled.
2. only poll a server on startup if it could be running (User.server is defined). If a Spawner appears to be running and User.server is not defined, there's not enough information to recover anyway.
3. Catch and log exceptions in pre-flight Spawner.poll, so that Spawner bugs cannot prevent the app from launching.

Each of these individually would fix the specific crash in #1060, but all together, JupyterHub should be better protected from Spawners and the option to keep the userlist in sync should be useful.

closes #1060

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1151,7 +1151,14 @@ class JupyterHub(Application):
             self.users[orm_user.id] = user = User(orm_user, self.tornado_settings)
             self.log.debug("Loading state for %s from db", user.name)
             spawner = user.spawner
-            status = yield spawner.poll()
+            status = 0
+            if user.server:
+                try:
+                    status = yield spawner.poll()
+                except Exception:
+                    self.log.exception("Failed to poll Spawner for %s, assuming it is not running.", user.name)
+                    status = -1
+
             if status is None:
                 self.log.info("%s still running", user.name)
                 spawner.add_poll_callback(user_stopped, user)
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -954,11 +954,18 @@ class JupyterHub(Application):
             try:
                 yield gen.maybe_future(self.authenticator.add_user(user))
             except Exception:
-                # TODO: Review approach to synchronize whitelist with db
-                # known cause of the exception is a user who has already been removed from the system
-                # but the user still exists in the hub's user db
                 self.log.exception("Error adding user %r already in db", user.name)
-        db.commit()  # can add_user touch the db?
+                if self.authenticator.delete_invalid_users:
+                    self.log.warning("Deleting invalid user %r", user.name)
+                    db.delete(user)
+                else:
+                    self.log.warning(dedent("""
+                    You can set
+                        c.Authenticator.delete_invalid_users = True
+                    to automatically delete users that have been invalidated,
+                    e.g. by deleting them from the external system without notifying JupyterHub.
+                    """))
+        db.commit()
 
         # The whitelist set and the users in the db are now the same.
         # From this point on, any user changes should be done simultaneously
@@ -1156,7 +1163,7 @@ class JupyterHub(Application):
                 try:
                     status = yield spawner.poll()
                 except Exception:
-                    self.log.exception("Failed to poll Spawner for %s, assuming it is not running.", user.name)
+                    self.log.exception("Failed to poll spawner for %s, assuming the spawner is not running.", user.name)
                     status = -1
 
             if status is None:
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -956,14 +956,16 @@ class JupyterHub(Application):
             except Exception:
                 self.log.exception("Error adding user %r already in db", user.name)
                 if self.authenticator.delete_invalid_users:
-                    self.log.warning("Deleting invalid user %r", user.name)
+                    self.log.warning("Deleting invalid user %r from the Hub database", user.name)
                     db.delete(user)
                 else:
                     self.log.warning(dedent("""
                     You can set
                         c.Authenticator.delete_invalid_users = True
-                    to automatically delete users that have been invalidated,
-                    e.g. by deleting them from the external system without notifying JupyterHub.
+                    to automatically delete users from the Hub database that no longer pass
+                    Authenticator validation,
+                    such as when user accounts are deleted from the external system
+                    without notifying JupyterHub.
                     """))
         db.commit()
 
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -126,11 +126,19 @@ class Authenticator(LoggingConfigurable):
     ).tag(config=True)
 
     delete_invalid_users = Bool(False,
-        help="""Delete any invalid users from the database
+        help="""Delete any users from the database that do not pass validation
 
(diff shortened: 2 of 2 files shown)


jupyterhub/auth.py
```diff
@@ -125,6 +125,15 @@ def validate_username(self, username):
         """
     ).tag(config=True)
 
+    delete_invalid_users = Bool(False,
+        help="""Delete any invalid users from the database
```
willingc: Perhaps:

```
help="""Delete users from the hub database if users fail validation

When the JupyterHub server starts, a `validate_users` check will be performed
on the Hub's database. 

If set to True, any users that do not pass validation will be
deleted.

If set to False (default), invalid users remain in the Hub's database. It is the
default to avoid data loss due to config changes.
```

jupyterhub/app.py
```diff
@@ -1151,7 +1158,14 @@ def user_stopped(user):
             self.users[orm_user.id] = user = User(orm_user, self.tornado_settings)
             self.log.debug("Loading state for %s from db", user.name)
             spawner = user.spawner
-            status = yield spawner.poll()
+            status = 0
+            if user.server:
+                try:
+                    status = yield spawner.poll()
+                except Exception:
+                    self.log.exception("Failed to poll Spawner for %s, assuming it is not running.", user.name)
```
willingc: Just to clarify what is happening here and meaning of `it`:

"Failed to poll Spawner for %s, assuming Spawner is not running."

jupyterhub/app.py
```diff
@@ -954,11 +954,18 @@ def init_users(self):
             try:
                 yield gen.maybe_future(self.authenticator.add_user(user))
             except Exception:
-                # TODO: Review approach to synchronize whitelist with db
-                # known cause of the exception is a user who has already been removed from the system
-                # but the user still exists in the hub's user db
                 self.log.exception("Error adding user %r already in db", user.name)
-        db.commit()  # can add_user touch the db?
+                if self.authenticator.delete_invalid_users:
+                    self.log.warning("Deleting invalid user %r", user.name)
```
willingc: Maybe "Deleting invalid user %r from Hub database"

jupyterhub/app.py
```diff
@@ -954,11 +954,18 @@ def init_users(self):
...
+                    db.delete(user)
+                else:
+                    self.log.warning(dedent("""
+                    You can set
+                        c.Authenticator.delete_invalid_users = True
+                    to automatically delete users that have been invalidated,
```
willingc: ~~have been invalidated~~  do not pass Hub database validation.

Allows users to be deleted from the external system without notifying JupyterHub and its Hub database."

Not perfect wording; yet I think when it comes to validation and deletion it's better to err on the side of being precise/specific

jupyterhub/app.py
```diff
@@ -954,11 +954,18 @@ def init_users(self):
...
+                    e.g. by deleting them from the external system without notifying JupyterHub.
+                    """)
```
willingc: OK, this took me way too long to hunt down why the tests were failing. There should be a second paren to close the warning.

minrk:
Finally came back to this. Thanks for the review, I think they should be better, now.

willingc:
Thanks @minrk. Sorry this fell off my radar.

PR #1076, opened 2017-04-12
Updates Jupyter server config file names and locations 

philejmath:


willingc:
Thanks @philejmath :cake:

PR #1082, opened 2017-04-15
fix named_servers, change server.base_url

barrachri:
https://github.com/jupyterhub/jupyterhub/issues/766
Some fixes and improvements for named_serves

### DIFF
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -220,11 +220,11 @@ class UserCreateNamedServerAPIHandler(APIHandler):
         user = self.find_user(name)
         if user is None:
             raise HTTPError(404, "No such user %r" % name)
-        if user.running:
-            # include notify, so that a server that died is noticed immediately
-            state = yield user.spawner.poll_and_notify()
-            if state is None:
-                raise web.HTTPError(400, "%s's server is already running" % name)
+        #if user.running:
+        #    # include notify, so that a server that died is noticed immediately
+        #    state = yield user.spawner.poll_and_notify()
+        #    if state is None:
+        #        raise web.HTTPError(400, "%s's server is already running" % name)
 
         options = self.get_json_body()
         yield self.spawn_single_user(user, options=options)
@@ -246,8 +246,8 @@ class UserDeleteNamedServerAPIHandler(APIHandler):
         if user.stop_pending:
             self.set_status(202)
             return
-        if not user.running:
-            raise web.HTTPError(400, "%s's server is not running" % name)
+        #if not user.running:
+        #    raise web.HTTPError(400, "%s's server is not running" % name)
         # include notify, so that a server that died is noticed immediately
         status = yield user.spawner.poll_and_notify()
         if status is not None:
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -350,24 +350,10 @@ class Spawner(LoggingConfigurable):
         """
     ).tag(config=True)
 
-    _instances = {}
-    
     def __init__(self, **kwargs):
         super(Spawner, self).__init__(**kwargs)
         if self.user.state:
             self.load_state(self.user.state)
-    
-    def save_spawner(self, server_name):
-        Spawner._instances[server_name] = self
-        self.server_name = server_name
-
-    @classmethod
-    def get_spawners(cls):
-        return cls._instances
-    
-    @classmethod
-    def get_spawner(cls, server_name):
-        return cls._instances[server_name]
 
     def load_state(self, state):
         """Restore state of spawner from database.
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -350,10 +350,24 @@ class Spawner(LoggingConfigurable):
         """
     ).tag(config=True)
 
+    _instances = {}
+    
     def __init__(self, **kwargs):
         super(Spawner, self).__init__(**kwargs)
         if self.user.state:
             self.load_state(self.user.state)
+    
+    def save_spawner(self, server_name):
+        Spawner._instances[server_name] = self
+        self.server_name = server_name
+
+    @classmethod
+    def get_spawners(cls):
+        return cls._instances
+    
+    @classmethod
+    def get_spawner(cls, server_name):
+        return cls._instances[server_name]
 
     def load_state(self, state):
         """Restore state of spawner from database.
@@ -469,8 +483,8 @@ class Spawner(LoggingConfigurable):
             ns (dict): namespace for string formatting.
         """
         d = {'username': self.user.name}
-        if self.user.server:
-            d['base_url'] = self.server_base_url
+        if self.server:
+            d['base_url'] = self.server.base_url
         return d
 
     def format_string(self, s):
@@ -495,7 +509,7 @@ class Spawner(LoggingConfigurable):
         """
         args = [
             '--user="%s"' % self.user.name,
-            '--base-url="%s"' % self.server_base_url,
+            '--base-url="%s"' % self.server.base_url,
             '--hub-host="%s"' % self.hub.host,
             '--hub-prefix="%s"' % self.hub.server.base_url,
             '--hub-api-url="%s"' % self.hub.api_url,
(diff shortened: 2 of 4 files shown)


barrachri:
Any idea how to have `server_base_url` inside the test?

The problem with is that the spawner is called outside of the user context where `server_base_url` is set.

@minrk could you help me with these stuff?
1. How can I handle oauth with named servers?
2. Does `/users/{name}/admin-access` still work to grant access to a user notebook?

jupyterhub/user.py
```diff
@@ -235,8 +235,11 @@ def spawn(self, options=None):
         api_token = self.new_api_token()
         db.commit()
 
+        self.server_base_url = base_url
```
minrk: User.server_base_url probably shouldn't be defined. With multiple servers, this no longer makes sense.

jupyterhub/spawner.py
```diff
@@ -495,7 +495,7 @@ def get_args(self):
         """
         args = [
             '--user="%s"' % self.user.name,
-            '--base-url="%s"' % self.user.server.base_url,
```
minrk: The `server` object should probably be passed directly to the Spawner, rather than accessing it through `user.server`, which no longer resolves to the right instance.

jupyterhub/user.py
```diff
@@ -235,8 +235,11 @@ def spawn(self, options=None):
...
+
         spawner = self.spawner
-        # Passing server_name to the spawner
+        # Passing server_name and base_url to the spawner
+        spawner.server_base_url = base_url
```
minrk: Rather than passing a new `server_base_url`, let's pass the `server` itself.

jupyterhub/user.py
```diff
@@ -235,8 +235,11 @@ def spawn(self, options=None):
         api_token = self.new_api_token()
         db.commit()
 
+        self.server_base_url = base_url
+
         spawner = self.spawner
```
minrk: A user now needs multiple Spawners, not just one. There needs to be a separate Spawner per server.

jupyterhub/user.py
barrachri: base_url is used to set the base url inside the jupyter server, to have the same path proxy/server, with one single server is equal to `/user/{user}` but with named servers you need to change this to fit `/user/{user}/{server}.

I thought of doing this with server_base_url, or perhaps I miss something....

jupyterhub/user.py
barrachri: you mean that we need an instances of Spawner for each server?

jupyterhub/user.py
barrachri: the Orm object?

jupyterhub/spawner.py
barrachri: So base_url should be moved inside the server object? (though could be already inside the model, I have to check....)

jupyterhub/user.py
minrk: yes. A Spawner corresponds to a server instance, so we have to have a Spawner instance for every running server.

jupyterhub/user.py
minrk: yes

jupyterhub/spawner.py
minrk: server.base_url is already defined, as seen in this line. Since each Spawner now needs to have a different `server`, it should be `self.server` instead of `self.user.server` here.

jupyterhub/user.py
barrachri: How do you plan to track each instance?

jupyterhub/user.py
barrachri: ok

jupyterhub/user.py
minrk: I suppose a user.spawners dict keyed by server_name.

barrachri: What do you thin about this implementation?
```python

class Spawner():
    _instances = {}
    
    def __new__(cls, *args, **kw):
        instance = object.__new__(cls)
        cls._instances[args[0]] = instance
        return instance

    def __init__(self, name):
        self.name = name

    @classmethod
    def get_instances(cls):
      return cls._instances
    
    @classmethod
    def get_instance(cls, name):
      return cls._instances[name]
```

minrk: I don't think Spawner needs to be modified at all, except perhaps for how it accesses `self.server` instead of `self.user.server`. The User just needs to have a dict of Spawner instances instead of a singleton.

jupyterhub/spawner.py
```diff
@@ -350,10 +350,24 @@ def _deprecate_percent_u(self, proposal):
...
+        Spawner._instances[server_name] = self
+        self.server_name = server_name
+
+    @classmethod
+    def get_spawners(cls):
+        return cls._instances
+    
+    @classmethod
+    def get_spawner(cls, server_name):
+        return cls._instances[server_name]
```
barrachri: so you would rather prefer to move this implementation inside the user.py?

I was thinking that from a perspective of a service that would like to control the spawner then it would be hard to access Spawner's instances.

minrk: > so you would rather prefer to move this implementation inside the user.py?

Yes. A user has spawners, so `spawners` should be an attribute of the User object.

> I was thinking that from a perspective of a service that would like to control the spawner then it would be hard to access Spawner's instances.

There aren't services in memory that have access to the Spawner class, so they wouldn't be able to access the instances class attribute.

barrachri:
I moved instances inside user :)

jupyterhub/apihandlers/users.py
```diff
@@ -220,11 +220,11 @@ def post(self, name):
         user = self.find_user(name)
         if user is None:
             raise HTTPError(404, "No such user %r" % name)
-        if user.running:
-            # include notify, so that a server that died is noticed immediately
-            state = yield user.spawner.poll_and_notify()
-            if state is None:
-                raise web.HTTPError(400, "%s's server is already running" % name)
+        #if user.running:
```
minrk: This running check needs to stay. However, it needs to be updated to check for the specified named server, instead  of the user's default server.

jupyterhub/apihandlers/users.py
```diff
@@ -246,8 +246,8 @@ def delete(self, name, server_name):
         if user.stop_pending:
             self.set_status(202)
             return
-        if not user.running:
-            raise web.HTTPError(400, "%s's server is not running" % name)
+        #if not user.running:
```
minrk: same here

jupyterhub/user.py
```diff
@@ -132,6 +134,20 @@ def __init__(self, orm_user, settings, **kwargs):
             config=self.settings.get('config'),
         )
 
+    @property
+    def get_spawners(self):
+        return self._instances
```
minrk: I don't think we need to hide spawners behind a private attribute exposed as a public property. I think `user.spawners` can be the dict of Spawner instances

jupyterhub/user.py
```diff
@@ -132,6 +134,20 @@ def __init__(self, orm_user, settings, **kwargs):
...
+    
+    @property
+    def get_spawner(self, server_name):
```
minrk: property methods don't take arguments. Maybe this is meant to be a regular method? It doesn't appear to be used anywhere.

jupyterhub/user.py
```diff
@@ -236,7 +252,12 @@ def spawn(self, options=None):
         db.commit()
 
         spawner = self.spawner
-        # Passing server_name to the spawner
+
+        # Save spawner's instance inside self._instances
+        self.save_spawner(server_name)
```
minrk: Perhaps `self.spawners[server_name] = spawner`

jupyterhub/apihandlers/users.py
barrachri: @minrk  am I right that `user.running` is expected to be the only server running?
Same goes for `self.spawn_pending` and `self.stop_pending`

So we should change them in something like:

```python
self.spawners[server_name].running
self.spawners[server_name].spawn_pending
self.spawners[server_name].stop_pending
```

minrk: yes, all of these are going to need to move to the spawner, since singletons no longer make sense. We'll probably still want to keep the properties on the user, mapped to the always-present `self.spawners['']` 'default' Spawner.

minrk:
@barrachri thanks for getting this going! We finally landed a finished version of this PR in #1243.

PR #1092, opened 2017-04-19
[WIP] run tests with mysql, postgres

### LINKED ISSUE #1088: Jupyterhub 0.7.2 does not work with a MySQL backend

1. Setup a JupyterHub with mysql database backend, pointing to an empty database

2. JupyterHub creates the required tables in the database on start

**What actually happens**

It fails to start, with:

```
    sqlalchemy.exc.InternalError: (pymysql.err.InternalError) (1071, 'Specified key was too long; max key length is 767 bytes') [SQL: '\nCREATE TABLE groups (\n\tid INTEGER NOT NULL AUTO_INCREMENT, \n\tname VARCHAR(1023), \n\tPRIMARY KEY (id), \n\tUNIQUE (name)\n)\n\n']
```

I think the group name needs to be smaller.

minrk:
make sure everything works (spoiler: it doesn't)

will close #1088 when everything is green.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,10 @@ before_install:
   - |
     if [[ $JUPYTERHUB_TEST_DB_URL == mysql* ]]; then
       mysql -e 'CREATE DATABASE IF NOT EXISTS jupyterhub;'
+      pip install 'mysql-connector<2.2'
     elif [[ $JUPYTERHUB_TEST_DB_URL == postgresql* ]]; then
       psql -c 'create database jupyterhub;' -U postgres
+      pip install psycopg2
     fi
 install:
   - pip install --pre -r dev-requirements.txt .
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -5,12 +5,21 @@ python:
...
+  - mysql
+  - postgresql
 
 # installing dependencies
 before_install:
   - nvm install 6; nvm use 6
   - npm install
   - npm install -g configurable-http-proxy
+  - |
+    if [[ $JUPYTERHUB_TEST_DB_URL == mysql* ]]; then
```
willingc: Do we need to pip install mysqlclient and psycopg2?

minrk: yes, indeed. Thanks.

minrk:
I spent much of yesterday fighting with this one, and I don't want to burn too much energy on it because most of the issues have to do with #1096, which has no bearing on a production environment. I may wait for #1096 before coming back to this.

.travis.yml
yuvipanda: I would highly recommend using pymysql rather than mysqlclient

minrk: > I would highly recommend using pymysql rather than mysqlclient

I tried that first, but ran into latin-1 encoding issues in the unicode tests. It does seem preferable in general, though.

minrk:
Closing this PR because I won't come back to it for a while. At least until after #1096

Issue #1064, opened 2017-04-07
update docs for oauth


minrk:
- [x] update descriptions of how auth works
- [x] update service docs, pointing to new OAuth classes (existing docs / cookie-based auth still works)
- [ ] add new services.auth.OAuth classes to API docs

PR #1093, opened 2017-04-19
Add docs for OAuth

willingc:


docs/source/getting-started.md
```diff
@@ -258,17 +272,20 @@ allow running without SSL using the command `jupyterhub --no-ssl`. The
...
 
-Security is the most important aspect of configuring Jupyter. There are four main aspects of the
-security configuration:
+Security is the most important aspect of configuring Jupyter. These are the
+main elements of a security configuration:
 
 1. SSL encryption (to enable HTTPS)
-2. Cookie secret (a key for encrypting browser cookies)
+2. Cookie secret (a key for encrypting browser cookies) (Beginning in version
+   0.8, cookie secret is deprecated, though available for use.)
```
minrk: Cookie secret is not deprecated. It is only the shared-cookie behavior that is deprecated. The Hub itself still uses a cookie.

docs/source/getting-started.md
```diff
@@ -310,7 +327,7 @@ Note on **chain certificates**: If you are using a chain certificate, see also
 Note: In certain cases, e.g. **behind SSL termination in nginx**, allowing no SSL
 running on the hub may be desired.
 
-### Cookie secret
+### Cookie secret TODO Does this change at all with OAuth2?
```
minrk: For the most part, no. However, the sentence below:

> If this value changes for the Hub, all single-user servers must also be restarted.

Should be removed because it is no longer true (I think maybe it never was).

docs/source/getting-started.md
```diff
@@ -499,7 +532,12 @@ Each user may have one of these files in `$HOME/.jupyter/`.
 Jupyter also supports loading system-wide config files from `/etc/jupyter/`,
 which is the place to put configuration that you want to affect all of your users.
 
-## External services
+## External services (Change in version 0.8 for using OAuth2)
+
+Beginning with version 0.8, JupyterHub is also an OAuth2 provider. This allows
+JupyterHub to provide tokens for the Service to use, as well as allow the
+client to access and authenticate with the REST API on behalf of the
+Service. TODO How does this impact the next few paragraphs?
```
minrk: The only change for external services related to OAuth is environment variables. There is now a JUPYTERHUB_CLIENT_ID environment variable to set the JupyterHub OAuth client id of the service.

The rest should be unchanged. The only bit that OAuth affects is how services may identify the originating user of an incoming request, which isn't discussed in this section.

docs/source/quickstart.md
```diff
@@ -107,34 +108,32 @@ Generate a default config file:
...
-Spawn the server on ``10.0.1.2:443`` with **https**:
+For example, to start the jupyterhub on ``10.0.1.2:443`` with **https**:
 
     jupyterhub --ip 10.0.1.2 --port 443 --ssl-key my_ssl.key --ssl-cert my_ssl.cert
 
-The authentication and process spawning mechanisms can be replaced,
-which should allow plugging into a variety of authentication or process
-control environments. Some examples, meant as illustration and testing of this
-concept, are:
+The authenticator and spawner can be replaced from the default provided.
```
minrk: awkward phrasing "replaced from the default". Perhaps "The default authenticator...can be replaced"?

willingc:
Picking this back up for upcoming 0.8 release to address #1064.

willingc:
@minrk I'll add information about being an OAuth provider in another PR.

docs/rest-api.yml
```diff
@@ -435,21 +447,70 @@ paths:
...
+        '404':
+          description: A user is not found.
+  /oath2/authorize:
+    get:
+      summary: 'OAuth 2.0 authorize endpoint'
+      description: 'OAuth 2.0 authorize endpoint'
+      parameters:
+        - name: client_id
+          description: The client id
+          in: path
```
minrk: I think this (and all oauth args) should be `in: query`, not `in: path`

docs/rest-api.yml
```diff
@@ -435,21 +447,70 @@ paths:
...
+          type: string
+        - name: state
+          description: A state string
+          in: path
+          required: false
+          type: string
+        - name: redirect_uri
+          description: The redirect url
+          in: path
+          required: false
```
minrk: redirect_uri is required

docs/rest-api.yml
```diff
@@ -435,21 +447,70 @@ paths:
           description: The user identified by the cookie
           schema:
             $ref: '#/definitions/User'
-  /shutdown:
```
minrk: It looks like shutdown got deleted. Is that on purpose?

docs/rest-api.yml
```diff
@@ -435,21 +447,70 @@ paths:
...
+          description: A state string
+          in: path
+          required: false
+          type: string
+        - name: redirect_uri
+          description: The redirect url
+          in: path
+          required: false
+          type: string
+        - name: scope
```
minrk: There is no scope argument (for now)

docs/rest-api.yml
```diff
@@ -435,21 +447,70 @@ paths:
...
+      summary: 'OAuth 2.0 authorize endpoint'
+      description: 'OAuth 2.0 authorize endpoint'
+      parameters:
+        - name: client_id
+          description: The client id
+          in: path
+          required: true
+          type: string
+        - name: response_type
+          description: The response type (token | code)
```
minrk: response_type will always be 'code' (for now), no other value is supported yet.

docs/rest-api.yml
```diff
@@ -435,21 +447,70 @@ paths:
...
+          in: path
+          required: false
+          type: string
+        - name: scope
+          description: The requested scopes
+          in: path
+          required: false
+          type: string
+      responses:
+  /oath2/token:
```
minrk: s/oath/oauth/

docs/rest-api.yml
```diff
@@ -435,21 +447,70 @@ paths:
...
       parameters:
-        - name: proxy
-          in: body
-          type: boolean
-          description: Whether the proxy should be shutdown as well (default from Hub config)
-        - name: servers
-          in: body
-          type: boolean
-          description: Whether users's notebook servers should be shutdown as well (default from Hub config)
+        - name: client_id
```
minrk: /oauth2/token takes the following arguments:

- client_id
- client_secret
- grant_type (always 'authorization_code' for now)
- code
- redirect_url

All arguments are strings and required and `in: body`

Response is JSON, of the form:

```python
{ 'access_token': token}
```

docs/rest-api.yml
willingc: Nope, an error on my part.

willingc:
@minrk I think that I have incorporated all of your suggested changes.

minrk:
Thanks!

PR #1097, opened 2017-04-21
Don't give OAuth access tokens access to the REST API

minrk:
We don't want servers that identify users via OAuth to be able to take actions on behalf of those users (for now).

Eventually, we may want to enable this (e.g. services that can start user servers), which would require:

1. a full introduction of auth scopes to API tokens.
2. a permissions-confirmation page in the OAuth flow, which is currently absent because there is only one OAuth scope.

I don't want to introduce that complexity now, so I've just separated OAuth access tokens and enabled them them only on the `/authorizations/token` and `/user` (whoami) entrypoints.

### DIFF
--- a/jupyterhub/oauth/store.py
+++ b/jupyterhub/oauth/store.py
@@ -6,7 +6,7 @@ implements https://python-oauth2.readthedocs.io/en/latest/store.html
 import threading
 
 from oauth2.datatype import Client, AccessToken, AuthorizationCode
-from oauth2.error import AccessTokenNotFound, AuthCodeNotFound, ClientNotFoundError, UserNotAuthenticated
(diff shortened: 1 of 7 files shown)


jupyterhub/oauth/store.py
```diff
@@ -17,8 +17,7 @@
 from tornado.escape import url_escape
 
 from .. import orm
-from jupyterhub.orm import APIToken
-from ..utils import url_path_join, hash_token, compare_token
+from ..utils import url_path_join, hash_token, compare_token, new_token
```
willingc: @minrk Is `new_token` used?

minrk: Good catch. Removed.

willingc:
Thanks @minrk!

PR #1127, opened 2017-05-04
set sudo: false on Travis

minrk:


minrk:
From docs, it seems like this should be the travis default, but it doesn't appear to be. Perhaps it's only the default for new repos?

willingc:
It was the default for 2015+. I wonder if their recent beta of trusty and updates may have changed sudo to be explicitly set or it may default to "required".

PR #1131, opened 2017-05-08
clear poll callbacks list before using it

minrk:
avoids accumulating callbacks for repeated spawns


willingc:
Thanks @minrk

willingc:
Sorry, clicked the wrong button.

PR #1132, opened 2017-05-09
Correct stop docstring

minrk:
stop should always wait to finish, `now` indicates how forcefully it should be done.

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -539,10 +539,13 @@ class Spawner(LoggingConfigurable):
     def stop(self, now=False):
         """Stop the single-user server
 
-        If `now` is set to `False`, do not wait for the server to stop. Otherwise, wait for
-        the server to stop before returning.
+        If `now` is False (default), shutdown the server as gracefully as possible,
+        e.g. starting with SIGINT, then SIGTERM, then SIGKILL.
+        If `now` is True, terminate the server immediately.
 
-        Must be a Tornado coroutine.
+        The coroutine should return when the single-user server process is no longer running.
+
+        Must be a coroutine.
         """
         raise NotImplementedError("Override in subclass. Must be a Tornado gen.coroutine.")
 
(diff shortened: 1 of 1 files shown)


jupyterhub/spawner.py
```diff
@@ -539,8 +539,11 @@ def start(self):
     def stop(self, now=False):
```
yuvipanda: In the long run, 'force' might be a better name for this, since it is used for gentleness rather than time directly. Doesn't really matter tho.

jupyterhub/spawner.py
```diff
@@ -917,8 +920,11 @@ def _signal(self, sig):
...
         """Stop the single-user server process for the current user.
 
-        If `now` is set to True, do not wait for the process to die.
-        Otherwise, it'll wait.
+        If `now` is False (default), shutdown the server as gracefully as possible,
+        e.g. starting with SIGINT, then SIGTERM, then SIGKILL.
+        If `now` is True, terminate the server immediately.
+
+        The coroutine should return when the process is no longer running.
```
yuvipanda: Should say 'singleuser-server', not 'process' I think?

jupyterhub/spawner.py
```diff
@@ -539,8 +539,11 @@ def start(self):
     def stop(self, now=False):
         """Stop the single-user server
 
-        If `now` is set to `False`, do not wait for the server to stop. Otherwise, wait for
-        the server to stop before returning.
+        If `now` is False (default), shutdown the server as gracefully as possible,
+        e.g. starting with SIGINT, then SIGTERM, then SIGKILL.
+        If `now` is True, terminate the server immediately.
+
+        The coroutine should return when the process is no longer running.
```
yuvipanda: Should say 'singleuser-server', not 'process' I think?

jupyterhub/spawner.py
willingc: Perhaps: `single-user server process` for consistency with `_signal`

jupyterhub/spawner.py
willingc: Perhaps: `single-user server process` for consistency.

jupyterhub/spawner.py
yuvipanda: IMO we should get rid of all references to 'process' from the base abstract authenticators, since the thing being spawned might not be a process (could be a container, or VM, etc). Unrelated to this PR tho maybe.

PR #1146, opened 2017-05-21
log redirect targets

minrk:
on both Hub and singleuser

should help debugging redirect loops, etc.

### DIFF
--- a/jupyterhub/log.py
+++ b/jupyterhub/log.py
@@ -85,16 +85,38 @@ def log_request(handler):
     headers = _scrub_headers(request.headers)
 
     request_time = 1000.0 * handler.request.request_time()
-    user = handler.get_current_user()
+
+    try:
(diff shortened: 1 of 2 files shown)


jupyterhub/log.py
```diff
@@ -85,16 +85,38 @@ def log_request(handler):
...
+    msg = "{status} {method} {uri}{location} ({user}@{ip}) {request_time:.2f}ms"
     if status >= 500 and status != 502:
         log_method(json.dumps(headers, indent=2))
+    elif status in {301, 302}:
+        # log redirect targets
+        # FIXME: _headers is private, but there doesn't appear to be a public way
+        # to get headers from tornado
+        location = handler._headers.get('Location')
+        if location:
+            ns['location'] = ' → {}'.format(location)
```
willingc: Nice!

PR #1165, opened 2017-06-06
add hub_connect_ip

### LINKED ISSUE #1163: Any way to give a different hub IP to singleuser servers?
My current setup uses Jupyterhub 0.7.2 in a docker container, using a modified version of the batchspawner to start singleuser servers on a cluster. `JupyterHub.hub_ip` is set to `0.0.0.0`, but it binds to `127.0.0.1`.  This address is given to the singleuser server under the `--hub-api-url` argument. This isn't OK since the singleuser servers are running on different machines than the hub. I can't set the `hub_ip` to the public address because it is running in a docker container and can't bind to that address.

Is there any way from the config to tell the singleuser servers that it should connect to a different IP than the `hub_ip`? I'm currently doing a janky hack in the spawner to inject the correct address that obviously isn't the correct way to do it.

minrk:
allows specifying the connect ip/hostname for the Hub when it differs from hub_ip (the bind address).

Used when the Hub is not on the same host as the spawners and/or proxy (e.g. docker, kubernetes, etc.)

One default change: when `hub_ip` is 0.0.0.0 (all-interfaces), the default connect ip is `socket.gethostname()` instead of 127.0.0.1, which increases the likelihood that it will work by default, though container cases will likely still need to set this value.

This should mean that Spawner-specific config (e.g. `DockerSpawner.hub_connect_ip`) should no longer be necessary.

closes #1163

cc @yuvipanda

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -392,11 +392,26 @@ class JupyterHub(Application):
         )
 
     hub_port = Integer(8081,
-        help="The port for this process"
+        help="The port for the Hub process"
     ).tag(config=True)
     hub_ip = Unicode('127.0.0.1',
-        help="The ip for this process"
+        help="""The ip address for the Hub process to *bind* to.
+
+        See `hub_connect_ip` for cases where the bind and connect address should differ.
+        """
     ).tag(config=True)
+    hub_connect_ip = Unicode('',
+        help="""The ip or hostname for proxies and spawners to use
+        for connecting to the Hub.
+
+        Use when the bind address (`hub_ip`) is 0.0.0.0 or otherwise different
+        from the connect address.
+
+        Default: when `hub_ip` is 0.0.0.0, use `socket.gethostname()`, otherwise use `hub_ip`.
+
+        .. versionadded:: 0.8
+        """
+    )
     hub_prefix = URLPrefix('/hub/',
         help="The prefix for the hub server.  Always /base_url/hub/"
     )
@@ -836,7 +851,8 @@ class JupyterHub(Application):
             cookie_name='jupyter-hub-token',
             public_host=self.subdomain_host,
         )
-        print(self.hub)
+        if self.hub_connect_ip:
+            self.hub.connect_ip = self.hub_connect_ip
 
     @gen.coroutine
     def init_users(self):
--- a/jupyterhub/objects.py
+++ b/jupyterhub/objects.py
@@ -3,6 +3,7 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
+import socket
 from urllib.parse import urlparse
 
 from tornado import gen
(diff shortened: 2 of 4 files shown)


jupyterhub/app.py
```diff
@@ -836,7 +851,8 @@ def init_hub(self):
             cookie_name='jupyter-hub-token',
             public_host=self.subdomain_host,
         )
-        print(self.hub)
+        if self.hub_connect_ip:
+            self.hub.connect_ip = self.hub_connect_ip
```
DeepHorizons: This line is redundant

minrk: It is not. Note that `hub.connect_ip` is not the same as `hub_connect_ip`.

jupyterhub/app.py
```diff
@@ -392,11 +392,26 @@ def _deprecated_proxy_api(self, change):
...
+        """
     ).tag(config=True)
+    hub_connect_ip = Unicode('',
+        help="""The ip or hostname for proxies and spawners to use
+        for connecting to the Hub.
+
+        Use when the bind address (`hub_ip`) is 0.0.0.0 or otherwise different
+        from the connect address.
+
+        Default: when `hub_ip` is 0.0.0.0, use `socket.gethostname()`, otherwise use `hub_ip`.
```
yuvipanda: Why use socket.gethostname()?

jupyterhub/objects.py
```diff
@@ -26,10 +27,22 @@ class Server(HasTraits):
...
     base_url = Unicode('/')
     cookie_name = Unicode('')
+    
+    @property
+    def _connect_ip(self):
+        """Property for connect_ip"""
+        if self.connect_ip:
+            return self.connect_ip
+        elif self.ip in {'', '0.0.0.0'}:
+            # if listening on all interfaces, default to hostname
```
yuvipanda: Why not just actually listen on all interfaces?

yuvipanda: Binding to socketname has very different (and IMO unexpected) behavior from binding to 0.0.0.0.

```bash
python3 -m http.server --bind 0.0.0.0 8000
```

curl localhost:8000 works, curl courage:8000 works (courage is my hostname).

```
python3 -m http.server --bind courage 8000
```

curl localhost:8000 does *not* work, and curl courage:8000 does work. If we do this substitution by default you can never listen on localhost and public interfaces, even though you specified 0.0.0.0 which has that behavior - unintuitive and confusing.

If people want it to only listen on a public interface, they should be asked to explicitly specify that IMO.

minrk: Nothing here will bind to a hostname. I'll add some comments that clarify that `connect_ip` is only ever used for connecting, never binding. Nothing about binding is changed in this PR and users can still specify whatever they want for the bind address. This is only for "If bound on all interfaces, tell clients to connect via hostname unless another ip is specified for connections."

The reasoning behind switching from 127.0.0.1 to hostname when the bind address is all-interfaces:

The default for this case so far has been 127.0.0.1, which works if everything is local and doesn't work when clients are remote. If the bind address is explicitly changed to 0.0.0.0, it is probable that there are remote clients (otherwise, why choose it?). The result up to now is that that setting `ip = 0.0.0.0` *always requires* setting a different config value for the connect address explicitly as well (e.g. `DockerSpawner.hub_ip_connect`), depending on the Spawner. This default case is changed to `hostname`, which will work for remote and local clients in sensible networking conditions, and can always be overridden with `JupyterHub.hub_connect_ip`, just like DockerSpawner and KubeSpawner have been doing themselves.

yuvipanda: Aah, whoops. +1

PR #1167, opened 2017-06-07
pass more arguments as environment variables instead of CLI args

minrk:
env is often easier to deal with for Spawners

Now, only optional args are passed on the command-line and all required args come from environment variables.


jupyterhub/singleuser.py
```diff
@@ -233,13 +241,13 @@ def _validate_base_url(self, proposal):
     def _port_default(self):
         if os.environ.get('JUPYTERHUB_SERVICE_URL'):
             url = urlparse(os.environ['JUPYTERHUB_SERVICE_URL'])
-            return url.port
+            return url.port or 8888
```
willingc: @minrk Do you want this inside the `if` block or outside of it? (Same question for line 250.)

minrk: a bit of both, should be fixed now.

PR #1177, opened 2017-06-21
define routespec

### LINKED ISSUE #1175: Make sure pluggable proxy config supports user per-domain
We currently support this, and should continue to in a way that works across implementations!

minrk:
namedtuple separating host and path. host is optional and will be `''` unless host-based routing is  involved.

Since RouteSpecs are almost always just a path, everywhere that accepts a RouteSpec argument must also accept a str and wrap it in `RouteSpec(s)`, e.g. via `RouteSpec.as_routespec(spec)`, which is a no-op for RouteSpec objects, for convenience/clarity.

closes #1175

### DIFF
--- a/jupyterhub/apihandlers/proxy.py
+++ b/jupyterhub/apihandlers/proxy.py
@@ -9,8 +9,26 @@ from urllib.parse import urlparse
 from tornado import gen, web
 
 from .. import orm
+from ..proxy import RouteSpec
 from ..utils import admin_only
 from .base import APIHandler
+
+class _RouteSpecJSONEncoder(json.JSONEncoder):
+    """JSON encoder that handles routespecs"""
+    def encode(self, obj):
+        if isinstance(obj, RouteSpec):
+            obj = obj.host + obj.path
+        return super().encode(obj)
+
+    def iterencode(self, obj, _one_shot=True):
+        if isinstance(obj, dict):
+            obj = {
+                key.host + key.path if isinstance(key, RouteSpec) else key : value
+                for key, value in obj.items()
+            }
+        return super().iterencode(obj)
+
+_json = _RouteSpecJSONEncoder()
 
 class ProxyAPIHandler(APIHandler):
     
@@ -18,19 +36,22 @@ class ProxyAPIHandler(APIHandler):
     @gen.coroutine
     def get(self):
         """GET /api/proxy fetches the routing table
-        
+
         This is the same as fetching the routing table directly from the proxy,
         but without clients needing to maintain separate
         """
         routes = yield self.proxy.get_all_routes()
-        self.write(json.dumps(routes))
-    
+        self.write(_json.encode(routes))
+
     @admin_only
     @gen.coroutine
     def post(self):
-        """POST checks the proxy to ensure"""
+        """POST checks the proxy to ensure that it's up to date.
+
+        Can be used to jumpstart a newly launched proxy
+        without waiting for the check_routes interval.
+        """
         yield self.proxy.check_routes(self.users, self.services)
-        
     
     @admin_only
     @gen.coroutine
--- a/jupyterhub/apihandlers/proxy.py
+++ b/jupyterhub/apihandlers/proxy.py
@@ -9,26 +9,9 @@ from urllib.parse import urlparse
 from tornado import gen, web
 
 from .. import orm
-from ..proxy import RouteSpec
 from ..utils import admin_only
 from .base import APIHandler
 
-class _RouteSpecJSONEncoder(json.JSONEncoder):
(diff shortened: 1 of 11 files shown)


jupyterhub/proxy.py
```diff
@@ -26,6 +28,54 @@
...
+class RouteSpec(namedtuple('RouteSpec', ['path', 'host'])):
+    """Object wrapping a proxy route specification
+
+    Attributes:
+
+        path (str): The URL path prefix for this route. Required.
+        host (str): The hostname used for routing when host-based routing is enabled.
+
+    """
+    def __new__(cls, path, *, host=''):
```
yuvipanda: IMO host should be None (or '*') rather than '' when we don't care. Some proxy systems might have special meaning for ''.

jupyterhub/proxy.py
```diff
@@ -61,8 +112,8 @@ def add_route(self, routespec, target, data):
         """Add a route to the proxy.
 
         Args:
-            routespec (str): A specification for which this route will be matched.
-                Could be either a url_prefix or a fqdn.
+            routespec (RouteSpec or str): A specification for which this route will be matched.
```
yuvipanda: When will we ever pass in a string? IMO proxy implementations should never have to care about casting into RouteSpec.

minrk: I disagree. When approximately all cases are passing a trivial wrapper for a single string whose interpretation is unambiguous and trivial to handle, the API should handle it.

jupyterhub/proxy.py
minrk: How could `''` have a meaning as a host when it's not a valid hostname? If host is not specified, the Proxy implementation (the Python API here) should not send host info to the upstream implementation at all. It's a matter for JupyterHub-specific Python code, not the upstream implementation.

I have a slight preference to '' over None so that the type is stable, but None is okay, too. I don't think we should use '*' because `bool(host)` should be False when host is not specified.

jupyterhub/apihandlers/proxy.py
```diff
@@ -9,28 +9,49 @@
 from tornado import gen, web
 
 from .. import orm
+from ..proxy import RouteSpec
 from ..utils import admin_only
 from .base import APIHandler
 
+class _RouteSpecJSONEncoder(json.JSONEncoder):
```
minrk: This is the part I like the least - now that routes are no longer valid keys in a JSON dict, we need to cast them to strings. And since we are casting them to strings, maybe we should just be using strings anyway?

Since paths must always start with `/`, a routespec can already be unambiguous as a URL:

    'host/path/...'
    '/path/only/...'

yuvipanda: I'm actually ok with this idea!

yuvipanda: It would also address my comments about RouteSpec vs String.

jupyterhub/proxy.py
```diff
@@ -1,12 +1,21 @@
-"""API for JupyterHub's proxy."""
+"""API for JupyterHub's proxy.
+
+Route Specification:
+
+- A routespec is a URI (excluding scheme), e.g.
+  'host.name/path/' for host-based routing or '/path/' for default routing.
```
yuvipanda: It's probably host.name:port. Most code that does splits should handle this just fine, but we should explicitly mention it here I think.

jupyterhub/proxy.py
```diff
@@ -55,14 +65,32 @@ def stop(self):
...
+            return routespec
 
     @gen.coroutine
     def add_route(self, routespec, target, data):
         """Add a route to the proxy.
 
         Args:
-            routespec (str): A specification for which this route will be matched.
-                Could be either a url_prefix or a fqdn.
+            routespec (str): A URI (excluding scheme) for which this route will be matched,
```
yuvipanda: It's technically a URL prefix we're matching, rather than a URI. We're skipping query params and fragments too, for instance.

jupyterhub/proxy.py
```diff
@@ -55,14 +65,32 @@ def stop(self):
...
+                e.g. host.name/path/
             target (str): A URL that will be the target of this route.
```
yuvipanda: This will include scheme right?

jupyterhub/proxy.py
minrk: Thanks, clarified now, I think.

jupyterhub/proxy.py
minrk: yes, I've just updated the docstrings with more detail, which should cover this.

yuvipanda:
w00t!

PR #1178, opened 2017-06-21
Encrypt auth_state at rest

yuvipanda:
We might not end up using sqlalchemy-utils in the end - am scrutinizing its encryption implementation.

### DIFF
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -37,6 +37,25 @@ class Authenticator(LoggingConfigurable):
     """Base class for implementing an authentication provider for JupyterHub"""
 
     db = Any()
+    
+    enable_auth_state = Bool(False, config=True,
+        help="""Enable persisting auth_state (if available).
+
+        auth_state will be encrypted and stored in the Hub's database.
+        This can include things like authentication tokens, etc.
+        to be passed to Spawners as environment variables.
+
+        Encrypting auth_state requires the cryptography package.
+
+        Additionally, the JUPYTERHUB_CRYPTO_KEY envirionment variable must
+        contain one (or more, separated by ;) 32B encryption keys.
+        These can be either base64 or hex-encoded.
+
+        If encryption is unavailable, auth_state cannot be persisted.
+
+        New in JupyterHub 0.8
+        """,
+    )
 
     admin_users = Set(
         help="""
--- a/jupyterhub/crypto.py
+++ b/jupyterhub/crypto.py
@@ -1,0 +1,123 @@
+
+from concurrent.futures import ThreadPoolExecutor
+import json
+import os
+
+from traitlets.config import SingletonConfigurable, Config
+from traitlets import Any, Dict, Integer, List, default, validate
+
+try:
+    import privy
+except ImportError:
+    privy = None
+
+
+KEY_ENV = 'JUPYTERHUB_CRYPT_KEY'
+
+class EncryptionUnavailable(Exception):
+    pass
+
+class PrivyUnavailable(EncryptionUnavailable):
+    def __str__(self):
+        return "privy library is required for encryption"
+
+class NoEncryptionKeys(EncryptionUnavailable):
+    def __str__(self):
+        return "Encryption keys must be specified in %s env" % KEY_ENV
+
+class CryptKeeper(SingletonConfigurable):
+    """Encapsulate encryption configuration
+
+    Use via the encryption_config singleton below.
+    """
+
+    privy_kwargs = Dict({'server': True},
+        help="""Keyword arguments to pass to privy.hide.
+        
+        For example, to 
+        """
+    )
+
+    n_threads = Integer(max(os.cpu_count(), 1),
+        help="The number of threads to allocate for encryption",
+        config=True,
+    )
+
+    @default('config')
+    def _config_default(self):
+        # load application config by default
+        from .app import JupyterHub
+        if JupyterHub.initialized():
+            return JupyterHub.instance().config
+        else:
+            return Config()
+
+    executor = Any()
+    def _executor_default(self):
+        return ThreadPoolExecutor(self.n_threads)
+
+    keys = List(config=True)
+    def _keys_default(self):
+        if KEY_ENV not in os.environ:
+            return []
+        # key can be a ;-separated sequence for key rotation.
+        # First item in the list is used for encryption.
+        return [ k.encode('ascii') for k in os.environ[KEY_ENV].split(';') if k.strip() ]
+
+    @validate('keys')
+    def _ensure_bytes(self, proposal):
+        # cast str to bytes
+        return [ (k.encode('ascii') if isinstance(k, str) else k) for k in proposal.value ]
+
+    def check_available(self):
+        if privy is None:
+            raise PrivyUnavailable()
+        if not self.keys:
+            raise NoEncryptionKeys()
+
+    def _encrypt(self, data):
+        """Actually do the encryption. Runs in a background thread.
+        
+        data is serialized to bytes with pickle.
+        bytes are returned.
+        """
+        return privy.hide(json.dumps(data).encode('utf8'), self.keys[0], **self.privy_kwargs).encode('ascii')
+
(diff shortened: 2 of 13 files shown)


yuvipanda:
Actually *not* going to use sqlalchemy-utils, since they seem to be using parts of the encryption key as IV, which... is pretty inescure!

jupyterhub/orm.py
```diff
@@ -54,6 +56,38 @@ def process_result_value(self, value, dialect):
         return value
 
 
+class OptionalEncrypted(EncryptedType):
+    def __init__(self, type_in=None, key=None, engine=None, **kwargs):
+        try:
+            import cryptography
```
minrk: I think setting auth_state should probably raise rather than warn in the absence of encryption

yuvipanda: yeah, I agree. Should raise when auth_state is actually set tho - __init__ is called regardless.

minrk: Gotcha. Then maybe change the message to something like "auth_state will be unavailable". We'll probably also want to make sure there's a sensible error when auth_state is attempted to be set when it is not available.

One option is to do this at a slightly higher level, in a Python property on User and leave the column as raw bytes as far as sqlalchemy is concerned.

minrk:
#1180 should allow this to be testable

minrk:
@yuvipanda do you have a change to come back to this? Or should I pick it up?

yuvipanda:
Could you pick it up for a bit? I might have to focus on scaling right now
- there is a possibility I might have to support up to 15k users in 2
months...

minrk:
Okay, I think this is good to review. I went through a few revisions, but landed on this one:

- Data is encrypted using [fernet](https://github.com/fernet/spec), provided by the `cryptography` package
- Use `MultiFernet` to enable key rotation
- Users must specify 32B encryption keys (can be raw, hex, or base64-encoded)
- If keys are not specified, auth_state cannot be saved (login will fail if Authenticator provides auth_state)
- Keys are pulled from `JUPYTERHUB_CRYPTO_KEY` env and can be a `;`-separated sequence to enable key rotation.

Some implementation highlights:

- encryption is always passed to a thread
- the db / orm don't know anything about the encryption. They just know that `encrypted_auth_state` is a column of raw bytes

Things tried, but ultimately not chosen:

- Using sqlachemy-utils for its EncryptedType. Doing the encryption in the ORM seemed nice at first, but didn't turn out great. This would also preclude putting the encryption in the background, and ultimately didn't offer much that we could actually use.
- Using [privy](https://github.com/ofek/privy) to ensure that we do things securely. Makes secure storage of password-encrypted data super nice by choosing a good KDF and using Fernet encryption. All you need to specify is a thing to store and a password and it does a good job. However, we aren't using user-specified passwords so we can ask deployments to pick a proper 32B key, thus eliminating the need for a KDF, which is most of what privy accomplishes.

jupyterhub/auth.py
```diff
@@ -37,6 +37,25 @@ class Authenticator(LoggingConfigurable):
...
+    enable_auth_state = Bool(False, config=True,
+        help="""Enable persisting auth_state (if available).
+
+        auth_state will be encrypted and stored in the Hub's database.
+        This can include things like authentication tokens, etc.
+        to be passed to Spawners as environment variables.
+
+        Encrypting auth_state requires the cryptography package.
+
+        Additionally, the JUPYTERHUB_CRYPTO_KEY envirionment variable must
```
yuvipanda: Should we name this something more specific to auth_state? I suspect we might have other encryption keys in the future, and would not want to re-use the same key.

jupyterhub/auth.py
```diff
@@ -37,6 +37,25 @@ class Authenticator(LoggingConfigurable):
...
+        contain one (or more, separated by ;) 32B encryption keys.
+        These can be either base64 or hex-encoded.
```
yuvipanda: IMO we shouldn't allow choices in things like this - let's pick hex and stick with it?

jupyterhub/crypto.py
```diff
@@ -0,0 +1,161 @@
...
+
+    return key
+
+class CryptKeeper(SingletonConfigurable):
+    """Encapsulate encryption configuration
+
+    Use via the encryption_config singleton below.
+    """
+
+    n_threads = Integer(max(os.cpu_count(), 1), config=True,
```
yuvipanda: I've been thinking we should base almost all our threadpools on concurrent_spawn_limit - so we don't end up blocking waiting on any single thread. Then we can set the concurrent_spawn_limit to a reasonable default, and have it propogate forward. It lets admins also tweak one variable that is easy to understand, rather than many related ones. For example, if n_threaads is 1 we're effectively limiting the number of concurrent user logins to 1, regardless of any other setting - I don't think you'd ever want to limit login to 1 but spawn to 50, for example.

jupyterhub/auth.py
minrk: > I suspect we might have other encryption keys in the future, and would not want to re-use the same key.

I was thinking the opposite - if we add any other encrypted data into the database, we *would* use the same key, wouldn't we? Is there value in having different keys encrypting different columns of the database? If that's wrong, I'll definitely change the name back to include AUTH_STATE.

jupyterhub/auth.py
minrk: I allowed base64 because Fernet keys are *defined* as 32-byte url-safe-base64-encoded keys. So if we pick one, it would probably be b64 for consistency with Fernet.

jupyterhub/crypto.py
minrk: How many threads you should use depends on what the threads are doing. Since encryption is CPU-bound, it shouldn't be greater than n cpus, or we'll just thrash context switching as the actual CPU threads hop between the large number of threads we have asked it to run concurrently that it can't actually do. Other pools are waiting on I/O, so a larger number makes sense, but concurrent connection limits (for example) can also dictate an upper bound for those that could easily be lower than the number of concurrent spawns allowed. Each spawn goes through many stages, and one thread waiting on an HTTP response doesn't conflict with another spawn from encrypting its auth_state, so it's not nearly as simple as setting a single concurrency limit and applying it to all resources.

> For example, if n_threaads is 1 we're effectively limiting the number of concurrent user logins to 1

This isn't true. Limiting the number of users that can be in a *single stage* of login isn't the same as limiting the number of concurrent logins, especially when one value causes a user-facing error message, while the other causes a minor increase in login time. Pool size is setting a bound on the amount of CPU/memory resources a given task can be allowed to ask for at a given time. Encryption is designed to be CPU and memory intensive, so allowing 1000 encryption tasks to spawn at the same time doesn't make sense on an 8 core server.

jupyterhub/auth.py
yuvipanda: ah, hmm. And we have hex keys already. Makes sense!

jupyterhub/auth.py
yuvipanda: True - we can revisit later if we introduce more of these!
 
re: other keys, yes, depending on the sensitivity of stuff we wanna encrypt, we *should* allow users to use different keys that might be rotated at different intervals, for example. However, I can't think of anything else we might wanna encrypt, so this is good!

jupyterhub/crypto.py
yuvipanda: This makes sense! I was only considering the fact that encryption probably drops the GIL. We can always measure and adjust defaults later if necessary!

yuvipanda:
@minrk is right as usual :D I've merged this! <3

jupyterhub/auth.py
minrk: After thinking about it more, I think *you* are right :)

I have a tendency to try to allow every possible input, which tends to cause headaches down the road. After thinking about this more, I think you're right and we should only allow hex keys because our users aren't going to know or care what a Fernet key is and we already generate 32B hex secrets elsewhere in the docs, so there isn't a *good* reason for this to be different. Consistency with ourselves is probably better than consistency with the docs of an underlying implementation that our users won't read.

PR #1180, opened 2017-06-22
allow .authenticate to return a dict containing name and state

minrk:
Allows authenticators to set .auth_state from info in the initial authentication.

cc @yuvipanda this is the PR I forgot to open

### DIFF
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -171,32 +171,37 @@ class Authenticator(LoggingConfigurable):
     def get_authenticated_user(self, handler, data):
         """Authenticate the user who is attempting to log in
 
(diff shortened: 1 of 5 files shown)


jupyterhub/auth.py
```diff
@@ -210,13 +223,19 @@ def authenticate(self, handler, data):
 
         Checking the whitelist is handled separately by the caller.
 
+        .. versionchanged:: Allow `authenticate` to return a dict containing auth_state.
```
willingc: Do you wish to add the version 0.8 here?

minrk: I certainly do

PR #1181, opened 2017-06-23
APIHandlers now handle CORS

alexxxxx:
Signed-off-by: Alexandre Beche <alexandre.beche@gmail.com>

### DIFF
--- a/jupyterhub/apihandlers/base.py
+++ b/jupyterhub/apihandlers/base.py
@@ -31,6 +31,12 @@ class APIHandler(BaseHandler):
         if not referer:
             self.log.warning("Blocking API request with no referer")
             return False
+
+        # Checking CORS
+        allow_origin = self.settings.get('allow_origin', None)
+        self.log.warning(allow_origin)
+        if (allow_origin == '*') or (allow_origin and allow_origin in referer_path):
+            return True
         
         host_path = url_path_join(host, self.hub.base_url)
         referer_path = referer.split('://', 1)[-1]
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -339,6 +339,9 @@ class JupyterHub(Application):
     base_url = URLPrefix('/',
         help="The base URL of the entire application"
     ).tag(config=True)
+    allow_origin = Unicode(
+        help="Origin to allow executing Cross-origin resource sharing"
+    ).tag(config=True)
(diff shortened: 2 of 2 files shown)


jupyterhub/apihandlers/base.py
```diff
@@ -31,6 +31,12 @@ def check_referer(self):
         if not referer:
             self.log.warning("Blocking API request with no referer")
             return False
+
+        # Checking CORS
+        allow_origin = self.settings.get('allow_origin', None)
+        self.log.warning(allow_origin)
+        if (allow_origin == '*') or (allow_origin and allow_origin in referer_path):
```
minrk: referer_path isn't defined yet. Plus `in` is an insufficient check for referer_path, which can contain arbitrary strings. Referer will need to be parsed to compare hosts exactly.

For example:

```python
allow_origin = 'https://mydomain.com'
Referer = ''
```

alexxxxx: Thanks for your feedback.
To make the story short, I am trying to run JupyterHub behind a frontend. When doing so, clicking the butting "Stop my server" fails because of X-Origin.

alexxxxx:
Hello,

I'm not sure to properly understand the difference between "Behind" or
"Next to". Basically, in my case, an httpd server is terminating SSL and
forwarding all traffic to the "configurable-http-proxy" of JupyterHub.
Looking of the code, you are comparing the referrer with the host headers
and I don't think I can override any of those without breaking the standard
(and the host header will always be my frontend).

Would it be acceptable if in addition to the existing checks, I also
compare referer with either X-Forwaded-For or the Forwarded headers?

Thanks in advance,
Cheers,
Alex

minrk:
Terminating SSL in front is common with JupyterHub (e.g. with nginx), preserving the Host header seems to be the common practice there. With Apache, the [ProxyPreserveHost](https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#proxypreservehost) flag should do this.

We'll have to be very careful trusting X headers here, because we shouldn't trust such headers the client may have set.

willingc:
Hi @alexxxxx, I'm considering closing this PR. Would you like to pick it back up or prefer that we close it? Thanks for the effort too.

willingc:
Closing this PR as it's been inactive for several months. Thanks! :sunflower:

Issue #1183, opened 2017-06-23
Spawner default_url needs '/tree'


evanlinde:
The setting `c.Spawner.default_url` doesn't work like the existing help says it should.  This is in version 0.7.2.

The documentation doesn't mention '/tree'; you get this in the URL when leaving `c.Spawner.default_url` unset, but not when it's set. Without including '/tree' at the beginning of `default_url`, the URL user is directed to after logging in doesn't include '/tree' and the user sees a 404 error.

So it looks like `default_url` needs to be `'/tree/'` + path relative to `notebook_dir`.

i.e. 
if `notebook_dir` is `/` and you want to land in `/home/{username}`, then `default_url` needs to be `/tree/home/{username}`
if `notebook_dir` is `/networkshare` and you want to land in `/networkshare/users/{username}`, then `default_url` needs to be `/tree/users/{username}`

PR #1184, opened 2017-06-23
Add '/tree' to spawner default_url

evanlinde:
Proposed fix for issue #1183


minrk:
Thanks! If anything, this should be in documentation, rather than code. If we hardcode `/tree` in code here, that eliminates the possibility of other landing pages, such as a specific notebook or other page.

evanlinde:
OK, I see what you're talking about. Pull request should be revised now.

minrk:
Perfect, thanks!

PR #1187, opened 2017-06-26
fix db reconnect when user.db changes

minrk:
store id on outer User, rather than accessing orm_user.id, which seems to fail sometimes

this may fix the recent increase in intermittent test failures


minrk:
seems to have improved at least a little bit.

PR #1189, opened 2017-06-27
add default hub route via Proxy.add_route

### LINKED ISSUE #1186: Have the hub explicitly add a default route to the proxy that points back to the hub
The hub should add a default route to the proxy that points `/` back to itself. This makes the lives of proxy implementations much easier.

minrk:
instead of relying on default target when the proxy starts

also add filtering to CHP to avoid including any routes not associated with JupyterHub on a shared proxy.

closes #1186

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ install:
 
 # running tests
 script:
-  - pytest --maxfail=2 --cov=jupyterhub jupyterhub/tests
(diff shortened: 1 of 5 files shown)


yuvipanda:
@minrk the test failures here seem genuine.

minrk:
Got it, works now.

PR #1191, opened 2017-06-28
Fix in-document anchors

kroq-gar78:
Removes the explicit extension of the target if it's in the same
document. This allows these links to work in GitHub or in Read the Docs.

The link in "See [Security documentation]" in `getting-started.md` already omits the filename, showing that this should work (unless the Markdown file is sourced somewhere else).

### DIFF
--- a/docs/source/getting-started.md
+++ b/docs/source/getting-started.md
@@ -2,14 +2,14 @@
 
 This section contains getting started information on the following topics:
 
-- [Technical Overview](getting-started.html#technical-overview)
-- [Installation](getting-started.html#installation)
-- [Configuration](getting-started.html#configuration)
-- [Networking](getting-started.html#networking)
-- [Security](getting-started.html#security)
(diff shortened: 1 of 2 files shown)


minrk:
Should work, thanks!

willingc:
Thanks for the contribution @kroq-gar78, and welcome. :cake:

kroq-gar78:
Thank you @willingc !

There was a small thing I forgot to mention. In https://github.com/jupyterhub/jupyterhub/blob/7b96950a9cde27bf07b337ea96134fb1021b6fa4/docs/source/config-examples.md#L75 , the link `oauth-setup` is not defined anywhere, but the same link is defined in https://github.com/jupyterhub/jupyterhub/blob/1742065f77b9809d172a3887fc78ac9215399a96/docs/source/getting-started.md#L540 to be `https://github.com/jupyterhub/oauthenticator#setup`.

I didn't think it would be worth creating another pull request for this.

PR #1198, opened 2017-06-29
Don't fail when there's no 'data' for a route

yuvipanda:
Adding extra data is optional, so we shouldn't fail when
data is not present.


minrk:
What's returned from the API should have `data` defined, presumably. It's optional when creating a route because there's an available default of an empty dict. An empty dict should come out of `get_route` when there's no data, though.

yuvipanda:
In that case let's switch the default from `None` to `{}`. I think `{}` is more appropriate anyway.

PR #1202, opened 2017-07-04
Add missing config tag to `will_resume`

AFrancani:
Following jupyterhub/jupyterhub#910 we tried to add `c.Spawner.will_resume = True` to our settings, but JupyterHub doesn't recognize it, following line is printed in the logs: 
```
[W 2017-07-04 11:46:15.817 JupyterHub configurable:168] Config option `will_resume` not recognized by `DockerSpawner`.
```
We found out that this is due to the `will_resume` option not being exposed with the config tag.

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -63,7 +63,7 @@ class Spawner(LoggingConfigurable):
         (e.g. resuming a Docker container),
         and API tokens in use when the Spawner stops will not be deleted.
         """
-    )
+    ).tag(config=True)
 
     ip = Unicode('',
         help="""


minrk:
I don't think this should be configurable. Instead, it should be a property of the Spawner. Users shouldn't set it, Spawner authors should. See [DockerSpawner](https://github.com/jupyterhub/dockerspawner/blob/master/dockerspawner/dockerspawner.py#L535) for an example.

ak0ska:
Hello,

We did some experimenting  prior to submitting this patch and if you never push Stop on the admin interface (or invoke the [stop method](https://github.com/jupyterhub/dockerspawner/blob/master/dockerspawner/dockerspawner.py#L516) in any other way), the `will_resume` setting is never set properly.

Our problem is that when the containers are stopped outside of Jupyterhub, the api tokens get removed from the database while remain unchanged in the stopped container. Our understanding is that the `will_resume` switch is aimed to address this exact problem, but the only way to set it is to invoke `stop()`. So if you use `docker stop` directly on the containers, this setting is not set to `not self.remove_containers` but remains `False` [as initialized](https://github.com/HumanBrainProject/jupyterhub/blob/0.7.2/jupyterhub/spawner.py#L55)

If you don't want to expose this setting as configurable, would it be ok to set out outside of `stop()`, somewhere around [here](https://github.com/jupyterhub/dockerspawner/blob/master/dockerspawner/dockerspawner.py#L151)?

Thanks,
Akos

minrk:
@ak0ska indeed, that's a bug in DockerSpawner that it's not always set correctly. jupyterhub/dockerspawner#167 should fix this to always be set, rather than during `.stop()`.

PR #1208, opened 2017-07-13
Minor fixes to get setup.py to work on Windows

dhirschfeld:
I'm attempting to get JupyterHub up & running on Windows. This was the first hurdle I've come across...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -181,10 +181,9 @@ class Bower(BaseCommand):
         env = os.environ.copy()
         env['PATH'] = npm_path
         args = ['bower', 'install', '--allow-root', '--config.interactive=false']
-        if os.name == 'nt':
-            args = ['cmd', '/C'] + args
+        shell = True if os.name == 'nt' else False
         try:
-            check_call(args, cwd=here, env=env)
+            check_call(args, cwd=here, env=env, shell=shell)
         except OSError as e:
             print("Failed to run bower: %s" % e, file=sys.stderr)
             print("You can install js dependencies with `npm install`", file=sys.stderr)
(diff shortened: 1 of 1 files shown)


dhirschfeld:
Fix shamelessly stolen from jupyter-widgets/declarativewidgets#380

dhirschfeld:
I think the CI error is unrelated?

setup.py
```diff
@@ -179,13 +180,11 @@ def run(self):
         
         env = os.environ.copy()
         env['PATH'] = npm_path
-        
+        args = ['bower', 'install', '--allow-root', '--config.interactive=false']
+        if os.name == 'nt':
+            args = ['cmd', '/C'] + args
```
minrk: I think `shell=True` should do this, right?

setup.py
```diff
@@ -119,7 +119,8 @@ def get_package_data():
 from distutils.command.sdist import sdist
 
 
-npm_path = ':'.join([
+sep = ';' if os.name == 'nt' else ':'
```
minrk: This should be `os.pathsep`, I think

setup.py
dhirschfeld: That sounds plausible! I'll give it a go and report back...

setup.py
dhirschfeld: Yep. I incorrectly thought it was for forward/back slashes. Will change...

setup.py
dhirschfeld: That worked and is cleaner. New commit pushed...

dhirschfeld:
CI failure seems unrelated `MemoryError`

PR #1214, opened 2017-07-14
Tag hub_ip_connect as configurable

### LINKED ISSUE #1213: Hub should know which route actually points back to it in proxy.add_hub_route
Currently it is:

```
        return self.add_route('/', self.hub.host, {'hub': True})
```

which when running on k8s leads to the hub telling the proxy:

```
[I 2017-07-14 02:56:03.191 JupyterHub proxy:338] Adding default route for Hub: / => http://hub-deployment-1084503180-rzv5m:8081
```

But that's the wrong IP to connect to, since it's using a hostname rather than an IP address. In fact, the current IP address is also wrong, since it needs to be the hub service IP rather than the hub pod IP (since the hub pod can be restarted independently of the proxy pod).

I *think* hub_connect_ip is the right thing to use here?

yuvipanda:
This value is expected to be provided by the user.

Fixes #1213

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -412,7 +412,7 @@ class JupyterHub(Application):
 
         .. versionadded:: 0.8
         """
-    )
+    ).tag(config=True)
     hub_prefix = URLPrefix('/hub/',
         help="The prefix for the hub server.  Always /base_url/hub/"
     )


yuvipanda:
Tested and works!

minrk:
Whoops! Nice catch. Based on the title, do you think we should chnage it to `hub_ip_connect` instead of `hub_connect_ip`?

PR #1220, opened 2017-07-14
fix filtering of services with no endpoint

### LINKED ISSUE #1215: Error messages about managed cull-idle server
Lots of messages in log about 
```
[E 2017-07-14 03:18:49.501 JupyterHub proxy:319] Service <Service(name=cull-idle managed)> has no server, but wasn't filtered out.
```

Not sure why - the cull-idle service doesn't make an entry into the proxy...

minrk:
use `!=` to check for None, not `is not`. The weird thing appears to be that `is not` *sometimes* works? Seems like it should be never or always, but `!=` is the right thing to use, in any case.

closes #1215

### DIFF
--- a/jupyterhub/proxy.py
+++ b/jupyterhub/proxy.py
@@ -308,8 +308,7 @@ class Proxy(LoggingConfigurable):
         # check service routes
         service_routes = {r['data']['service']
                           for r in routes.values() if 'service' in r['data']}
-        for orm_service in db.query(Service).filter(
-                Service.server is not None):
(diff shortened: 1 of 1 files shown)


willingc:
Odd. With sqlalchemy's operator overloading, `!=` is the right thing to use. *sometimes* works is odd. Perhaps another coincidence with the way `filter` sets up a SQL query. 

Double checked that there are no other `is None` or `is not None` in `filter` in other parts of the code.

PR #1221, opened 2017-07-14
add User.proxy_pending

### LINKED ISSUE #1217: Tracking down 'too many redirects' errors at scale
I've been able to reliably reproduce the 'too many redirect' errors by simply hitting a JupyterHub with enough traffic. 

I've a rough script over at https://github.com/yuvipanda/jupyterhub-loadtest that can do this. 

If I make a simple JupyterHub, with the following config:

```
c.JupyterHub.authenticator_class = 'dummyauthenticator.DummyAuthenticator'
c.JupyterHub.spawner_class = 'simplespawner.SimpleLocalProcessSpawner'
```

and hit it with 50 simultaneous users I get at least a few too many redirect errors.

Going to use this to collect info and try to solve it!

minrk:
flag for waiting for the proxy to be updated

avoids User.running being True when the user's server has not yet been added to the proxy, causing potential redirect loops.

With this, I was able to run the [load test](https://github.com/minrk/jupyterhub-loadtest) with 1000 users, 100 concurrent and no failures.

200 concurrent users introduces a small number of failures where single-user servers fail to start because the Hub is insufficiently responsive when they check-in during startup. A relatively minor issue.

closes #1217

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -321,6 +321,7 @@ class BaseHandler(RequestHandler):
         tic = IOLoop.current().time()
 
         f = user.spawn(options)
+        user.proxy_pending = True
 
         @gen.coroutine
         def finish_user_spawn(f=None):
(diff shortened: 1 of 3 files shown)


minrk:
I went ahead and handled the timeout in single-user startup as well. There was another small race on `.stop_pending`, which was being unset before the last yield, which it shouldn't have been. With this, I was able to login 1k users with 200 concurrent with zero errors.

yuvipanda:
Yay! The python code is a little bogus (since it doesn't test actually starting a kernel) but *should* catch stuff with too many redirects.

I'm gonna load test this too now. EXCITING.

yuvipanda:
Got it to work on a 1000 user setup now. Creating a 100 node cluster now to see how far I can take it

yuvipanda:
2000 created, many failed due to errors (some with curl, some because it took more than 5 minutes to start them (not sure where this delay is from!)) but not a single redirect loop!

yuvipanda:
Going to run a 5000 one with higher timeout limits to see how it goes!

yuvipanda:
At 5000 (and even at 2000) most of the errors were timeouts talking to the proxy, because the output of get_all_routes() is just too big. But no redirect errors, so I'm going to merge this and call this a win! <3 ty @minrk

PR #1223, opened 2017-07-15
apply exponential backoff to all waits

minrk:
Waiting for servers to come up and shut down was polled at an even interval of 100ms. If things are slow and busy, this is a lot of waiting events (and outstanding http requests). Exponential backoff reduces the number of callbacks triggered by slow spawners.

Note that this only really affects Spawners where the time between `Spawner.start()` returning and the server becoming responsive is long. *Not* spawners where `Spawner.start` itself takes a long time.

This may improve the load a bit when there’s a bunch of outstanding spawns.

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -27,7 +27,7 @@ from traitlets import (
 )
 
 from .traitlets import Command, ByteSpecification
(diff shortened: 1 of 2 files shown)


willingc:
Thanks @minrk.

PR #1234, opened 2017-07-18
Add pre_spawn_hook

### LINKED ISSUE #1219: Enhancement: automatically create a directory for the user after successful login
We are using jupyterhub with LDAPAuthenticator and DockerSpawner. 

We would like to mount a host volume into the docker container, which is not a problem if the directory to be mounted as volume exists before running the container using docker. If the directory to be mounted as volume does not exists beforehand, the docker daemon will create it "on demand" - but then, that created directory will be chown'ed by root:root - the user jovyan inside of the container will not be able to write into her own directory /home/jovyan/work as the host directory is owned by root.

There are some workarounds using docker, especially to write an ENTRYPOINT which executes a shellscript to re-chown the directory to the user. However, it would be a lot of work to enhance all of the docker images we'd like to use just for this case. And yes, we do not know who will be loggin in into the jupyterhub in the first place, as everyone of a certain LDAP-group will be allowed to login. 

The idea is as follows:

* Some user logs in into jupyterhub using one of the authenticators, in our case, the LDAP Authenticator
* After a successful login, the jupyterhub process will create a directory (to be specified in the configuration) on the host, if not already exists. Then, the directory will not be automatically owned by root as in the case when docker will create the directory during the first volume-mount process. Therefor, one additional traitlet will be made available to specify which directory is to be created. The traitlet will allow the variable {username} to be substituted by the name of the user who has just logged in.
* the Dockerspawner will mount the volume /home/jovyan/work to the just created directory, which is an already available dockerspawner feature. 

In case this idea finds some community support, I'd be willing to implement the optional directory creation after succesful login. 

I'm happy if you share your thoughts on this. Thank you.

Regenschein

CRegenschein:
Based on the discussion in #1219 

Intermediate work with the own Bootstrap class has been deleted.
pre_spawn_hook has been added.

Thanks.

closes #1219

### DIFF
--- a/examples/bootstrap-script/README.md
+++ b/examples/bootstrap-script/README.md
@@ -1,0 +1,130 @@
+# Bootstrapping your users
+
+Before spawning a notebook to the user, it could be useful to 
+do some preparation work in a bootstrapping process.
+
+Common use cases are:
+
+*Providing writeable storage for LDAP users*
+
+Your Jupyterhub is configured to use the LDAPAuthenticator and DockerSpawer.
+
+* The user has no file directory on the host since your are using LDAP.
+* When a user has no directory and DockerSpawner wants to mount a volume,
+the spawner will use docker to create a directory.
+Since the docker daemon is running as root, the generated directory for the volume 
+mount will not be writeable by the `jovyan` user inside of the container. 
+For the directory to be useful to the user, the permissions on the directory 
+need to be modified for the user to have write access.
+
+*Prepopulating Content*
+
+Another use would be to copy initial content, such as tutorial files or reference
+ material, into the user's space when a notebook server is newly spawned.
+
+You can define your own bootstrap process by implementing a `pre_spawn_hook` on any spawner.
+The Spawner itself is passed as parameter to your hook and you can easily get the contextual information out of the spawning process. 
+
+If you implement a hook, make sure that it is *idempotent*. It will be executed every time 
+a notebook server is spawned to the user. That means you should somehow 
+ensure that things which should run only once are not running again and again.
+For example, before you create a directory, check if it exists.
+
+Bootstrapping examples:
+
+### Example #1 - Create a user directory
+
+Create a directory for the user, if none exists
+
+```python
+
+# in jupyterhub_config.py  
+import os
+def create_dir_hook(spawner):
+    username = spawner.user.name # get the username
+    volume_path = os.path.join('/volumes/jupyterhub', username)
+    if not os.path.exists(volume_path):
+        # create a directory with umask 0755 
+        # hub and container user must have the same UID to be writeable
+        # still readable by other users on the system
+        os.mkdir(volume_path, 0o755)
+        # now do whatever you think your user needs
+        # ...
+        pass
+
+# attach the hook function to the spawner
+c.Spawner.pre_spawn_hook = create_dir_hook
+```
+
+### Example #2 - Run a shell script 
+
+You can specify a plain ole' shell script (or any other executable) to be run 
+by the bootstrap process.
+
+For example, you can execute a shell script and as first parameter pass the name 
+of the user:
+
+```python
+
+# in jupyterhub_config.py    
+from subprocess import check_call
+import os
+def my_script_hook(spawner):
+    username = spawner.user.name # get the username
+    script = os.path.join(os.path.dirname(__file__), 'bootstrap.sh')
+    check_call([script, username])
+
+# attach the hook function to the spawner
+c.Spawner.pre_spawn_hook = my_script_hook
+
+```
+
+Here's an example on what you could do in your shell script. See also 
+`/examples/bootstrap-script/`
+
+```bash
+#!/bin/bash
+
+# Bootstrap example script
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+# - The first parameter for the Bootstrap Script is the USER.
+USER=$1
+if ["$USER" == ""]; then
+    exit 1
+fi
+# ----------------------------------------------------------------------------
+
+
+# This example script will do the following:
+# - create one directory for the user $USER in a BASE_DIRECTORY (see below)
+# - create a "tutorials" directory within and download and unzip 
+#   the PythonDataScienceHandbook from GitHub
+
+# Start the Bootstrap Process
+echo "bootstrap process running for user $USER ..."
+
+# Base Directory: All Directories for the user will be below this point
+BASE_DIRECTORY=/volumes/jupyterhub/
+
+# User Directory: That's the private directory for the user to be created, if none exists
+USER_DIRECTORY=$BASE_DIRECTORY/$USER
+
+if [ -d "$USER_DIRECTORY" ]; then
+    echo "...directory for user already exists. skipped"
+    exit 0 # all good. nothing to do.
+else
+    echo "...creating a directory for the user: $USER_DIRECTORY"
+    mkdir $USER_DIRECTORY
+
+    echo "...initial content loading for user ..."
+    mkdir $USER_DIRECTORY/tutorials
+    cd $USER_DIRECTORY/tutorials
+    wget https://github.com/jakevdp/PythonDataScienceHandbook/archive/master.zip
+    unzip -o master.zip
+    rm master.zip
+fi
+
+exit 0
+```
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -350,8 +350,6 @@ class Spawner(LoggingConfigurable):
     ).tag(config=True)
 
     pre_spawn_hook = Any(
-        None,
-        allow_none=True,
         help="""
         An optional hook function that you can implement to do some bootstrapping work before
         the spawner starts. For example, create a directory for your user or load initial content.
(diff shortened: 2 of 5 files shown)


CRegenschein:
For the moment, I have nothing to add from my point of view.

examples/bootstrap-script/README.md
```diff
@@ -0,0 +1,116 @@
+# Bootstrapping your users
+
+Before spawning a notebook to the user, it could be useful to 
+do some preparation work in a bootstrapping process.
+
+One common use case is at follows:
```
willingc: ~~at~~ as

examples/bootstrap-script/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+* You are using an LDAPAuthenticator and have no home directory for your user on the host
+* You are using a DockerSpawner and would like to mount a volume, but as no directory exists, docker will create one.
+However, as the docker daemon is running as root, the on-the-fly generated directory for the volume mount will not be
+writeable by user jovyan inside of the container
+
+Another use case could be that every newly spawned notebook should come with initial content that you'd like to
```
willingc: Perhaps:

*Prepopulating Content:* Another use would be to copy initial content, such as tutorial files or reference material, into the user's space when a notebook server is newly spawned.

examples/bootstrap-script/README.md
```diff
@@ -0,0 +1,116 @@
+# Bootstrapping your users
+
+Before spawning a notebook to the user, it could be useful to 
+do some preparation work in a bootstrapping process.
+
+One common use case is at follows:
+
+* You are using an LDAPAuthenticator and have no home directory for your user on the host
```
willingc: *Providing writeable storage for an LDAP user:*

Your JupyterHub is configured to use the LDAPAuthenticator and DockerSpawner.
- The user has no home directory on the host since you are using LDAP.
- When a user has no home directory and DockerSpawner wants to mount a volume,
the spawner will use docker to create a directory.
Since the docker daemon is running as root, the generated directory for the volume mount will not be writeable by the `jovyan` user inside of the container. For the directory to be useful to the user, the permissions on the directory need to be modified for the user to have write access.

examples/bootstrap-script/README.md
```diff
@@ -0,0 +1,116 @@
...
+* You are using a DockerSpawner and would like to mount a volume, but as no directory exists, docker will create one.
+However, as the docker daemon is running as root, the on-the-fly generated directory for the volume mount will not be
+writeable by user jovyan inside of the container
+
+Another use case could be that every newly spawned notebook should come with initial content that you'd like to 
+copy into the user's space, such as tutorials or reference material.
+
+You can define your own bootstrap process by implementing a pre_spawn_hook on any spawner.
```
willingc: Let's put `pre_spawn_hook` within backticks so it will render as code.

examples/bootstrap-script/README.md
```diff
@@ -0,0 +1,116 @@
...
+The Spawner itself is passed as parameter to your hook and you can easily get the contextual information out of the spawning process. 
+
+If you implement a hook, make sure that it is *idempotent*. It will be executed every time 
+a notebook server is spawned to the user. That means you should somehow 
+ensure that things which should run only once are not running again and again.
+For example, before you create a directory, check if it exists.
+
+See examples:
```
willingc: Bootstrapping examples:

examples/bootstrap-script/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+### Example #1 - Create a directory for the user
```
willingc: Create a user directory

examples/bootstrap-script/README.md
```diff
@@ -0,0 +1,116 @@
...
+    username = spawner.user.name # get the username
+    check_call(['./examples/bootstrap-script/bootstrap.sh', username])
+
+# attach the hook function to the spawner
+c.Spawner.pre_spawn_hook = my_script_hook
+
+```
+
+Here's an example on what you could do in your shell script. See also 
+/examples/bootstrap-script/
```
willingc: Backticks should go around `/examples/bootstrap-script`

examples/bootstrap-script/README.md
yuvipanda: I would like to *not* have LDAPAuthenticator be part of the example, since doing this is actually insecure and we recommend against it in the LDAPAuthenticator README: https://github.com/jupyterhub/ldapauthenticator#configuration-note-on-local-user-creation

CRegenschein:
Thanks a lot @willingc for your suggestions.

examples/bootstrap-script/README.md
CRegenschein: Hello @yuvipanda , I didn't want to suggest to create a local user. I deleted the term "home"-directory in the description. 

Altough one major drawback of the suggested solution is that a directory once created for one user will be there forever, even when the user is kicked out of the LDAP directory. 
Maybe we should anticipate this by providing another cleanup hook once the user is deleted in the jupyterhub database?

examples/bootstrap-script/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+*Providing writeable storage for LDAP users*
+
+Your Jupyterhub is configured to use the LDAPAuthenticator and DockerSpawer.
+
+* The user has no file directory on the host since your are using LDAP.
+* When a user has no directory and DockerSpawner wants to mount a volume,
+the spawner will use docker to create a directory.
+Since the docker daemon is running as root, the generated directory for the volume 
+mount will not be writeable by the ```jovyan``` user inside of the container.
```
minrk: Future-reference markdown note: You can use single back-ticks:

```markdown
like `this`
```

for inline blocks. You only need triple-backticks for indented code blocks.

examples/bootstrap-script/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+```python
+
+# in jupyter_config.py  
+import os
+def create_dir_hook(spawner):
+    username = spawner.user.name # get the username
+    volume_path = os.path.join('/volumes/jupyterhub', username)
+    if not os.path.exists(volume_path):
+        os.mkdir(volume_path, 0o755)
```
minrk: Maybe add a comment that `755` is world-readable and owner-writable, so it will only be writable if the user and Hub have the same UID, and will not be private at all if there are other users on the system.

examples/bootstrap-script/README.md
```diff
@@ -0,0 +1,125 @@
...
+For example, you can execute a shell script and as first parameter pass the name 
+of the user:
+
+```python
+
+# in jupyter_config.py    
+from subprocess import check_call
+def my_script_hook(spawner):
+    username = spawner.user.name # get the username
+    check_call(['./examples/bootstrap-script/bootstrap.sh', username])
```
minrk: You can use 

```python
os.path.join(os.path.dirname(__file__), 'bootstrap.sh')
```

to point to `bootstrap.sh` in the same directory as `jupyterhub_config.py`. This path as it is assumes that `jupyterhub` is run in the top-level JupyterHub repo root.

examples/bootstrap-script/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+You can specify a plain ole' shell script (or any other executable) to be run 
+by the bootstrap process.
+
+For example, you can execute a shell script and as first parameter pass the name 
+of the user:
+
+```python
+
+# in jupyter_config.py
```
minrk: jupyterhub_config.py

examples/bootstrap-script/README.md
```diff
@@ -0,0 +1,125 @@
...
+    username = spawner.user.name # get the username
+    check_call(['./examples/bootstrap-script/bootstrap.sh', username])
+
+# attach the hook function to the spawner
+c.Spawner.pre_spawn_hook = my_script_hook
+
+```
+
+Here's an example on what you could do in your shell script. See also 
+```/examples/bootstrap-script/```
```
minrk: You can use single backticks here.

jupyterhub/spawner.py
```diff
@@ -522,6 +543,17 @@ def get_args(self):
...
         return args
 
+    def run_pre_spawn_hook(self):
+        """
+        run a pre spawn hook function which you can define in your jupyter_config.py
+        :return: void
+        """
+        if (callable(self.pre_spawn_hook)):
+            try:
+                self.pre_spawn_hook(self)
```
minrk: Need a `return` here to properly wait for Futures.

jupyterhub/spawner.py
```diff
@@ -522,6 +543,17 @@ def get_args(self):
...
+            except Exception as e:
+                self.log.error("pre_spawn_hook failed with exception: %s", str(e))
```
minrk: If you use `self.log.exception`, it will log an error message and the traceback, so you wouldn't need `str(e)` here.

minrk: Come to think of it, perhaps letting this raise would be better than catching and logging. We probably don't want to continue spawning a server if the pre-spawn hook fails.

CRegenschein:
Thanks for your comments, I think I checked and fixed them all. 
I will now raise the exception from the hook.

minrk:
@CRegenschein Thanks for addressing the review! I squashed the commits and added a tiny bit of cleanup. Thanks for the PR!

CRegenschein:
Great. Thanks to all for helping during my first PR :-)

willingc:
Congrats @CRegenschein :tada: :cake: 😄

PR #1236, opened 2017-07-19
Break up long Getting Started doc page

willingc:
This PR puts the documentation in a format where it will be easier to simplify and deduplicate content.

The PR changes:
- Installation guide section only contains installation info now (no configuration)
- The very long Getting Started document is broken into smaller documents and is temporarily its own Getting started section (see TOC screenshot and Rendered docs)
- Using docker section is done in RestructuredText so that a note can be added at the top that recommends using the Zero to JupyterHub process.

The next steps in a separate PR would be to edit the Getting Started documents and the Configuration Reference documents to remove duplication and simplify for users.

[Rendered docs](http://test-jupyterhub.readthedocs.io/en/doc-intro/)

screen shot 2017-07-18 at 6 20 42 pm


minrk:
Sounds like a good idea. Thanks!

PR #1242, opened 2017-07-20
various singleuser header fixes

minrk:
Workaround occasional Content security policy bug:

- always set content security policy header, to workaround bug in notebook 5.0 (https://github.com/jupyter/notebook/pull/2671)

Fixes for version-checking, which had some mistakes:

- set x-jupyterhub-version on all requests, including simple redirects
- fix version comparison in _check_version (leftover `__version__`)
- even log version matches at debug-level (verifies that check happened)


willingc:
Thanks @minrk. @tkinz27, fyi if you wish to try out this version.

tkinz27:
I believe this fixes our problems. At least on first look things are starting to work.

PR #1243, opened 2017-07-20
Flesh out named-servers

minrk:
Finish up #1082

The goal of this PR is to have named servers fully working, short of UI. The REST API, proxying, etc. should all be set, though.

- move state, server relationship to new Spawner table
- `user.spawners` is a dict by name
- avoid `user.server`, instead `user.spawners['name'].server`
- fix named server resume from state at startup
- named servers are still disabled by default, behind `allow_named_servers` flag
- default behavior and behavior of server '' should remain unchanged.

I'd like to save UI until 0.9, but I think we can get this internal part settled, since it's been outstanding for so long, and managing merge conflicts is getting difficult because it touches everything.


jupyterhub/app.py
```diff
@@ -1112,8 +1116,9 @@ def _user_summary(user):
             parts = ['{0: >8}'.format(user.name)]
             if user.admin:
                 parts.append('admin')
-            if user.server:
-                parts.append('running at %s' % user.server)
+            for name, spawner in sorted(user.spawners.items(), key=itemgetter(0)):
+                if spawner.server:
+                    parts.append('%r running at %s' % (name, spawner.server))
```
willingc: Nitpicky. Why are we using %r (repr) in some places?

jupyterhub/spawner.py
```diff
@@ -43,11 +44,27 @@ class Spawner(LoggingConfigurable):
...
-
-    db = Any()
+    
+    # private attributes for tracking status
+    _spawn_pending = False
+    _stop_pending = False
+    _proxy_pending = False
+    _waiting_for_response = False
+
+    orm_spawner = Any()
```
willingc: Duplicate in 54 and 58

jupyterhub/user.py
```diff
@@ -107,11 +116,7 @@ def _orm_user_changed(self, change):
         else:
             self._user_id = None
 
-    spawner = None
-    spawn_pending = False
-    stop_pending = False
-    proxy_pending = False
-    waiting_for_response = False
+    spawners = None
```
willingc: Any reason not to move this statement up to line 92 to follow `settings = Dict()`? It's a bit hard to distinguish at quick glance if this is part of `_db_changed`. It's not.

jupyterhub/user.py
```diff
@@ -96,7 +104,8 @@ def _db_changed(self, change):
         db = change.new
         if self._user_id is not None:
             self.orm_user = db.query(orm.User).filter(orm.User.id == self._user_id).first()
-        self.spawner.db = change.new
+        for name, spawner in self.spawners.items():
```
willingc: Line 106: The query after new session gets the user from the db in 106 and sets the instance's orm_user to the result of the query. 

Line 107: Iterates and gets the name and spawner for each entry in the spawner dictionary for a user.

Line 108: While iterating, get the named spawner that corresponds to the user and name in the db dictionary of user's spawners. So `spawner.orm_spawner` is holding ???

willingc: These few lines were a bit confusing. Might be worth a comment.

jupyterhub/apihandlers/users.py
```diff
@@ -185,80 +185,77 @@ class UserServerAPIHandler(APIHandler):
...
-    """
     @gen.coroutine
     @admin_or_self
     def delete(self, name, server_name):
+        if not self.allow_named_servers:
+            raise web.HTTPError(400, "Named servers are not enabled.")
         user = self.find_user(name)
-        if user.stop_pending:
+        if user is None:
+            raise web.HTTPError(404, "No such user %r" % name)
```
minrk: Good catch. I know I'm not consistent. I think it's whether I feel like `'username'` quotes seem right at the time. If I have a rule, it would be when it's a referring to the user by name, no quotes, but referring to an unattached username, quotes. I could also use `"%s"`, which would probably be clearer. For example:

> willingc's server has started

and

> No user named "willingc"

I see that the two examples you highlighted here don't even follow that rule, so I'll fix them.

jupyterhub/user.py
minrk: Comment added about re-attaching things to the db session.

PR #1248, opened 2017-07-21
Remove unused import and add version to deprecations.

Carreau:


jupyterhub/app.py
```diff
@@ -464,7 +462,8 @@ def _cookie_secret_check(self, change):
 
     @observe('api_tokens')
     def _deprecate_api_tokens(self, change):
-        self.log.warning("JupyterHub.api_tokens is pending deprecation."
+        self.log.warning("JupyterHub.api_tokens is pending deprecations"
```
willingc: Strike the `s` at the end of deprecations. Should read `deprecation` :-)

jupyterhub/app.py
```diff
@@ -1163,8 +1162,7 @@ def init_oauth(self):
         self.oauth_provider = make_provider(
             self.session_factory,
             url_prefix=url_path_join(base_url, 'api/oauth2'),
-            login_url=url_path_join(base_url, 'login')
-,
+            login_url=url_path_join(base_url, 'login'),
```
willingc: I don't know that we technically need a comma here as `make_provider` only takes 3 arguments.

jupyterhub/app.py
Carreau: Oops, don't know why a s was added.

jupyterhub/app.py
Carreau: Good point.

Carreau:
> I'm assuming you added the versions to make things more user friendly :-)

And to be more confident in when we can delete code. I like to delete code :-)

willingc:
Thanks @Carreau 👍

PR #1249, opened 2017-07-21
Do not 500 if cannot authenticate.

Carreau:
self.authenticate can return None, in which case you can't subscript.
So move extracting data into the branch checking whether authenticate is
not `None`.

Now that extracting the username is inside the if branch, it can't be
used in the else one, so extract username from the request itself.

This can be easily reproduce with the default PAM login with a wrong
non existing/ wrong username.


minrk:
THanks!

PR #1250, opened 2017-07-22
Apply exponential backoff when we're redirecting to ourselves

yuvipanda:
Lots of custom proxy implementations that are distributed are
eventually consistent, and it might take upto a few seconds for
all the components to start redirecting properly. If we do
exponential backoff when doing these redirects, it gives the
proxies a lot of time to catch up. We also explicitly raise an
error if it's going on too long, instead of giving the user
juts a 'redirected too many times' error.

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -6,7 +6,7 @@
 import re
 from datetime import timedelta
 from http.client import responses
-from urllib.parse import urlparse
+from urllib.parse import urlparse, urlunparse, parse_qs, urlencode
 
 from jinja2 import TemplateNotFound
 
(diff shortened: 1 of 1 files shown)


willingc:
Hi @yuvipanda, Can the exponential backoff that @minrk added https://github.com/jupyterhub/jupyterhub/blob/1bafdf9130780af088d79a9d04a4edb6eb1f2a8d/jupyterhub/utils.py#L51 be used in this pr?

yuvipanda:
Possibly, yeah. I'm not even sure this is the right thing to do yet tho! I'm going to iterate on this first before switching it out - will add a WIP label.

tkinz27:
@yuvipanda after looking at this, I'm having trouble understanding when this handler is called. It doesn't look like it's called in the path of spawning a user notebook server, but I am just looking through code, not actually tracing it. 

My issue is that after my proxy's add_route returns, the proxy is configured but will not route to the target until after a health check succeeds (which by default it needs 5 health checks to succeed to be considered healthy and each health check is performed at a 30 second interval -- can be tuned for sure). For me I have added code into my add_route function to wait to return until the target is healthy. This has fixed my redirect loop issues for now though.

yuvipanda:
@tkinz27 yeah, this is used when we get a URL directed to us outside the prefix ('/hub'). So in a redirect loop of '/user/<something>' -> '/hub/user/<something>', this is what's called for the '/user/<something>'.

I'm not sure if we need this, or if this is the way it should be implemented. This could be a bad idea :D 

My problem is that I don't have an easy way of checking if the ingress route has been seen by the controllers - there could be any number of them, and it usually settles down quickly (but not always!). However, I believe I'm seeing race conditions elsewhere in JupyterHub that might be the cause of my problems (which I'm seeing only when I'm load testing!) - see https://github.com/jupyterhub/jupyterhub/pull/1251

tkinz27:
@yuvipanda so the sequence of events must go like this?
1) /login --> /user
2) /user --> /hub/user
3) /hub/user --> /hub/user/username
... spawner polling is going on here ...
4 /hub/user/username --> /user/username

For your problem of not having a good way to see if your proxy is working, you could try just making an http request through the public url/proxy. If the request succeeds you know the proxy has settled . Too many http request failures and you can return a failure to the user.

yuvipanda:
I'm now convinced we don't need this, so going to close! 

@tkinz27 I can't do that unfortunately - there might be many many pods that could be serving the http request, so just the fact that one request went through doesn't guarantee anything unfortunately.

yuvipanda:
Re-opening, since @minrk says this is something we could do!

willingc:
@minrk @yuvipanda Is this ready for review or are you still working on this?

minrk:
@willingc I think this should be good to go now.

yuvipanda:
Ya I stress tested this yesterday, worked good!

PR #1251, opened 2017-07-22
Don't add route if there's already a route addition pending

yuvipanda:
This race seems to happen when check_routes is called in the middle of
the route being added.

### DIFF
--- a/jupyterhub/proxy.py
+++ b/jupyterhub/proxy.py
@@ -304,6 +304,8 @@ class Proxy(LoggingConfigurable):
                     self.log.warning(
                         "Adding missing route for %s (%s)", user.name, user.server)
                     futures.append(self.add_user(user))
+            elif user.proxy_pending:
+                good_routes.add(user.proxy_spec)
 
         # check service routes
         service_routes = {r['data']['service']


yuvipanda:
This doesn't solve it, I still run into:

```
[I 2017-07-22 09:25:20.200 JupyterHub base:339] User j-benchmark-1-j5x9616 server took 5.647 seconds to start
[I 2017-07-22 09:25:20.208 JupyterHub proxy:232] Adding user j-benchmark-1-j5x9616 to proxy /user/j-benchmark-1-j5x9616/ => http://10.244.5.85:8888
[I 2017-07-22 09:25:20.267 JupyterHub proxy:120] Created endpoints/jupyter--2fuser-2fj-2dbenchmark-2d1-2dj5x9616-2f-route
[I 2017-07-22 09:25:20.565 JupyterHub proxy:120] Created service/jupyter--2fuser-2fj-2dbenchmark-2d1-2dj5x9616-2f-route
[I 2017-07-22 09:25:20.749 JupyterHub proxy:120] Created ingress/jupyter--2fuser-2fj-2dbenchmark-2d1-2dj5x9616-2f-route
[I 2017-07-22 09:25:20.749 JupyterHub proxy:177] waiting for ingress jupyter--2fuser-2fj-2dbenchmark-2d1-2dj5x9616-2f-route to show up!
[W 2017-07-22 09:25:20.812 JupyterHub proxy:329] Deleting stale route /user/j-benchmark-1-j5x9616/
[I 2017-07-22 09:25:21.977 JupyterHub proxy:177] waiting for ingress jupyter--2fuser-2fj-2dbenchmark-2d1-2dj5x9616-2f-route to show up!
``

And then it just waits forever, because I haven't implemented exponential backoff there yet - but more importantly because we're deleteing the route before proxy add has finished.

**Comment by yuvipanda** [comment]
With check_routes turned off, almost all of my errors are now of the form:

```
[E 2017-07-22 10:32:10.223 JupyterHub gen:914] Exception in Future <tornado.concurrent.Future object at 0x7ff7645afcf8> after timeout
      File "/usr/local/lib/python3.5/dist-packages/tornado/gen.py", line 910, in error_callback
        future.result()
      File "/usr/local/lib/python3.5/dist-packages/jupyterhub/user.py", line 359, in spawn
        raise e
      File "/usr/local/lib/python3.5/dist-packages/jupyterhub/user.py", line 335, in spawn
        resp = yield server.wait_up(http=True, timeout=spawner.http_timeout)
      File "/usr/local/lib/python3.5/dist-packages/jupyterhub/utils.py", line 111, in wait_for_http_server
        "Server at {url} didn't respond in {timeout} seconds".format(**locals())
    TimeoutError: Server at http://10.244.8.55:8888/user/j-benchmark-5-7xzls72/ didn't respond in 30 seconds

```

File: jupyterhub/proxy.py [hunk 1]
```diff
@@ -300,7 +301,7 @@ def check_routes(self, user_dict, service_dict, routes=None):
             user = user_dict[orm_user]
             if user.running:
```
minrk: I think this can be `if user.running or user.proxy_pending`

jupyterhub/proxy.py
```diff
@@ -300,7 +301,7 @@ def check_routes(self, user_dict, service_dict, routes=None):
...
                 good_routes.add(user.proxy_spec)
-                if user.name not in user_routes:
+                if user.name not in user_routes and not user.proxy_pending:
```
minrk: At the moment, this `not user_proxy_pending` will always be true because `user.running` is contingent on `user.proxy_pending` being False.

We need to make sure that nether `add_user` nor `delete_user` is called if `user.proxy_pending`

I think this should fix the mishandled case:

```python
if user.running:
    # unchanged
elif user.proxy_pending:
    good_routes.add(user.proxy_spec)
```

jupyterhub/proxy.py
yuvipanda: oh, hmm right. A proxy_pending is not a good route!

jupyterhub/proxy.py
yuvipanda: Isn't that equivalent to what I already had?

minrk: Almost. `user.running` will never be True if proxy_pending is True, so the proxy_spec won't get added to `good_routes`, and thus will get deleted as stale in the next stage.

PR #1257, opened 2017-07-24
Dirty database fixes

minrk:
- Only set attributes on orm_server if they changed.
  Setting things on orm_server set the dirty flag, even if they haven't changed,
  so avoid it. The wrappers were effectively calling `orm_server.ip = orm_server.ip`,
  due to how the trait notifiers are hooked up.
- Only call `db.rollback()` if there are dirty objects to rollback.
  And log what objects are dirty when we do.
  Note: These may not actually be modified (`db.is_modified` could check this).

### DIFF
--- a/jupyterhub/objects.py
+++ b/jupyterhub/objects.py
@@ -83,7 +83,8 @@ class Server(HasTraits):
     @observe('ip', 'proto', 'port', 'base_url', 'cookie_name')
     def _change(self, change):
         if self.orm_server:
-            setattr(self.orm_server, change.name, change.new)
+            if getattr(self.orm_server, change.name) != change.new:
+                setattr(self.orm_server, change.name, change.new)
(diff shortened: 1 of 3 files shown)


yuvipanda:
Marathon hack session victory! \o/

jupyterhub/objects.py
```diff
@@ -83,7 +83,8 @@ def _orm_server_changed(self, change):
     @observe('ip', 'proto', 'port', 'base_url', 'cookie_name')
     def _change(self, change):
         if self.orm_server:
-            setattr(self.orm_server, change.name, change.new)
+            if getattr(self.orm_server, change.name) != change.new:
```
yuvipanda: Can you add a comment here explaining why?

willingc:
NIce!

PR #1258, opened 2017-07-24
don't salt & re-hash generated UUID tokens

minrk:
UUIDs have enough entropy on their own, so use just a simple sha512 (sha256 ought to be enough, really).

ref: https://security.stackexchange.com/a/151262/155114

The same scheme is used, just with one round of hashing and zero bytes of salt.

If admins insert short API tokens, the original salt&hash is still applied.

cc @ssanderson for double-checking that this is actually sensible.

@yuvipanda has found that our token hashing is a significant contributor to the total CPU time spent in the Hub, and a limiting factor with thousands of users, so using a simple hash here could save some real time. On my test machine, hashing tokens drops from ~3-4ms to ~60-80µs.

### DIFF
--- a/jupyterhub/orm.py
+++ b/jupyterhub/orm.py
@@ -427,7 +427,7 @@ class OAuthAccessToken(Hashed, Base):
 
     # from Hashed
     hashed = Column(Unicode(64))
-    prefix = Column(Unicode(16))
+    prefix = Column(Unicode(16), index=True)
     
     def __repr__(self):
         return "<{cls}('{prefix}...', user='{user}'>".format(
--- a/jupyterhub/orm.py
+++ b/jupyterhub/orm.py
@@ -268,7 +268,19 @@ class Hashed(object):
     def token(self, token):
         """Store the hashed value and prefix for a token"""
         self.prefix = token[:self.prefix_length]
-        self.hashed = hash_token(token, rounds=self.rounds, salt=self.salt_bytes, algorithm=self.algorithm)
+        if len(token) >= 32:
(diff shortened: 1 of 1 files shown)


yuvipanda:
This improved performance a lot on my 2k user profile run :D 

I'm going to merge this now - it looks pretty sensible to me. @ssanderson feel free to revert!

ssanderson:
@minrk @yuvipanda this seems sensible to me.  I also ran it by @jikamens, who has more expertise in this area than I do.

My only possible worry here would be that choosing the hash/salt algorithm based solely on input length feels a bit brittle to me.  There's still the unlikely possibility that someone is persistently storing long but low-entropy api keys in the db.  I think the safest possible option would probably be to just keep the existing implementation, but make it possible to disable if you have larger performance needs (which I'd expect to be the exception rather than the rule).  That might be more conservative than is warranted though.

yuvipanda:
Hmm, I think the default should probably be still the faster one tho - ideally users would have to tweak less things just to get performance, and since 99% of the time we're only hash UUIDs this should be ok. And it was a really hot codepath.

We could possibly do something like https://github.com/dxa4481/truffleHog (but not depend on it!) for  keys that are manually inserted?

minrk:
@ssanderson thanks! I opened #1299 to apply the logic only to internally generated tokens, rather than checking their size, which seems better to me, as well. User-provided tokens go through the extra rounds, which is fine because user-provided tokens are not super common to use (only externally managed services, really) and there will never be a large number of them.

PR #1267, opened 2017-07-25
add login_user and get_next_url methods

minrk:
for easier re-use of login in custom handlers

Further, allow auto_login + no custom login handler to mean that auth info is already present in requests (e.g. REMOTE_USER header).
This allows Authenticators to set auto_login without registering a custom login handler.


yuvipanda:
next_url will be useful for binder too!

yuvipanda:
Since we currently set 'Notebook.base_url' instead, which is wrong.

minrk:
next_url's always been around. You can use it with any version of JupyterHub: `/hub/login?next=...`. This is just consolidating the implementation to be used in more places.

PR #1272, opened 2017-07-26
Move exponential backoff into a function

yuvipanda:
Also use the 'Full Jitter' jitter algorithm from
https://www.awsarchitectureblog.com/2015/03/backoff.html

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -26,7 +26,7 @@ from traitlets import (
 
 from .objects import Server
 from .traitlets import Command, ByteSpecification
-from .utils import random_port, url_path_join, DT_MIN, DT_MAX, DT_SCALE
+from .utils import random_port, url_path_join, exponential_backoff
 
 
 class Spawner(LoggingConfigurable):
@@ -666,21 +666,19 @@ class Spawner(LoggingConfigurable):
                 self.log.exception("Unhandled error in poll callback for %s", self)
         return status
 
-    death_interval = Float(DT_MIN)
-
     @gen.coroutine
     def wait_for_death(self, timeout=10):
         """Wait for the single-user server to die, up to timeout seconds"""
-        loop = IOLoop.current()
-        tic = loop.time()
-        dt = self.death_interval
-        while dt > 0:
+        @gen.coroutine
+        def _wait_for_death():
             status = yield self.poll()
-            if status is not None:
-                break
-            else:
-                yield gen.sleep(dt)
-            dt = min(dt * DT_SCALE, DT_MAX, timeout - (loop.time() - tic))
+            return status is not None
+
+        yield exponential_backoff(
+            _wait_for_death,
+            'Process did not die in {timeout} seconds'.format(timeout=timeout),
+            timeout=timeout
+        )
 
 
 def _try_setcwd(path):
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -666,6 +666,7 @@ class Spawner(LoggingConfigurable):
                 self.log.exception("Unhandled error in poll callback for %s", self)
         return status
 
+    death_interval = Float(0.1)
     @gen.coroutine
     def wait_for_death(self, timeout=10):
         """Wait for the single-user server to die, up to timeout seconds"""
@@ -673,12 +674,16 @@ class Spawner(LoggingConfigurable):
         def _wait_for_death():
             status = yield self.poll()
             return status is not None
-
-        yield exponential_backoff(
-            _wait_for_death,
-            'Process did not die in {timeout} seconds'.format(timeout=timeout),
-            timeout=timeout
-        )
+        try:
+            r = yield exponential_backoff(
+                _wait_for_death,
+                'Process did not die in {timeout} seconds'.format(timeout=timeout),
+                start_wait=self.death_interval,
+                timeout=timeout,
+            )
+            return r
+        except TimeoutError:
+            return False
 
 
 def _try_setcwd(path):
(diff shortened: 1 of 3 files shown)


jupyterhub/objects.py
```diff
@@ -115,12 +115,13 @@ def bind_url(self):
             return self.url.replace(self._connect_ip, self.ip or '*', 1)
         return self.url
 
+    @gen.coroutine
     def wait_up(self, timeout=10, http=False):
         """Wait for this server to come up"""
         if http:
-            return wait_for_http_server(self.url, timeout=timeout)
+            return (yield wait_for_http_server(self.url, timeout=timeout))
```
minrk: Why this change? It adds an additional async layer that shouldn't be needed if we return the underlying result directly.

jupyterhub/spawner.py
```diff
@@ -652,21 +652,19 @@ def poll_and_notify(self):
...
+        def _wait_for_death():
             status = yield self.poll()
-            if status is not None:
-                break
-            else:
-                yield gen.sleep(dt)
-            dt = min(dt * DT_SCALE, DT_MAX, timeout - (loop.time() - tic))
+            return status is not None
+
+        yield exponential_backoff(
```
minrk: I'd like to keep `start_wait=self.death_interval` so subclasses can slow this down if needed.

jupyterhub/utils.py
```diff
@@ -80,13 +114,13 @@ def wait_for_http_server(url, timeout=10):
...
-    loop = ioloop.IOLoop.current()
-    tic = loop.time()
     client = AsyncHTTPClient()
-    dt = DT_MIN
-    while dt > 0:
+    @gen.coroutine
+    def is_reachable():
         try:
             r = yield client.fetch(url, follow_redirects=False)
+            print('returning r!')
```
minrk: print statement

jupyterhub/utils.py
```diff
@@ -48,29 +49,62 @@ def can_connect(ip, port):
...
+            return ret
+        else:
+            yield gen.sleep(dt)
+        # Add some random jitter to improve performance
+        # This makes sure that we don't overload any single iteration
+        # of the tornado loop with too many things
+        # See https://www.awsarchitectureblog.com/2015/03/backoff.html
+        # for a good example of why and how this helps. We're using their
+        # full Jitter implementation equivalent.
+        dt = min(max_wait, random.uniform(0, dt * scale_factor))
```
minrk: `loop.time() - start_tic` should also be added here as an upper limit. We don't want the last call to take 5 seconds if we've already waited 8 out of 10. See the implementation in master:

```python
while dt > 0:
    ...
    dt = min(max_wait, random(...), timeout - (loop.time() - start_tic))
```

which will turn negative and exit the loop when `loop.time() - start_tic` exceeds timeout.

minrk: Another point: probably want to introduce jitter on the first `dt` before `gen.sleep`, right? Rather than only after the second iteration? I think the dt calculation also probably makes more sense before gen.sleep. Maybe something like this:

```python
if ret:
    return ret
deadline = timeout - (loop.time() - start_tic)
if deadline < 0:
    # timeout exceeded
    break
dt = min(max_wait, deadline, random.uniform(0, dt * scale_factor))
yield gen.sleep(dt)
```

jupyterhub/spawner.py
```diff
@@ -666,21 +666,24 @@ def poll_and_notify(self):
                 self.log.exception("Unhandled error in poll callback for %s", self)
         return status
 
-    death_interval = Float(DT_MIN)
-
+    death_interval = Float(0.1)
```
willingc: Perhaps add a brief comment on what the death_interval is. Assuming 0.1s and is used when adding random jitter.

jupyterhub/spawner.py
```diff
@@ -666,21 +666,24 @@ def poll_and_notify(self):
...
+        @gen.coroutine
+        def _wait_for_death():
             status = yield self.poll()
-            if status is not None:
-                break
-            else:
-                yield gen.sleep(dt)
-            dt = min(dt * DT_SCALE, DT_MAX, timeout - (loop.time() - tic))
+            return status is not None
+        try:
```
willingc: Add LF between 676 and 677.

jupyterhub/utils.py
```diff
@@ -48,29 +49,69 @@ def can_connect(ip, port):
...
+    deadline = loop.time() + timeout
+    # add some jitter to the deadline itself, so that we don't
+    # re-align a bunch of timing out calls once the deadline is reached.
+    if timeout_tolerance:
+        tol = timeout_tolerance * timeout
+        deadline = random.uniform(deadline - tol, deadline + tol)
+    scale = 1
+    while True:
+        ret = yield gen.maybe_future(pass_func(*args, **kwargs))
+        # Truthy!
```
willingc: Perhaps:
`# Return pass_func value when truthy`

jupyterhub/utils.py
```diff
@@ -48,29 +49,69 @@ def can_connect(ip, port):
...
+        pass_func,
+        fail_message,
+        start_wait=0.2,
+        scale_factor=2,
+        max_wait=5,
+        timeout=10,
+        timeout_tolerance=0.1,
+        *args, **kwargs):
+    """
+    Exponentially backoff until pass_func is true.
```
willingc: Going to edit the docstring

willingc:
@yuvipanda Thanks!

yuvipanda:
This looks good to me! tvym for the cleanups, @minrk and @willingc!

willingc:
And tyvm for the cherrypick. 🍨 :tea: 💤 for you

Issue #1270, opened 2017-07-25
Decouple Handlers from Hub


Carreau:
I was trying to write #1263 (HubDispatcher), not on top of Hub, but one of the tight coupling here is the request handler has a `hub` property and make a few uses of it. (`self.hub.base_url`, `self.hub.base_url`,`self.hub.cookie_name`, `self._set_user_cookie(user, self.hub.server`). 

As authenticator do inherit their handler from `BaseHandler`, then they are coupled to the existence of a `Hub` thing. 

If we could transform the need for a hub to separate parameters that would make reusing handler in another context – like when you do not have a Hub –  Easier.

yuvipanda:
A more general question is why do we have a hub object in the database? Am confused about what use it is meant for...

minrk:
> A more general question is why do we have a hub object in the database?

All info about the Hub is removed from the db in master. This lives only in config now. Hub is a simple Python object.

As for decoupling, since these are properties the absence of `.hub` should only be noticed if it's used, so I'm not sure. The hub object encapsulates information about the Hub, so it's probably good to have in any Handler that needs it, and Authenticator handlers need that information for cookies, redirects, etc.. I'm not sure what the benefit would be of requiring several `hub_` settings over requiring one `hub` setting.

What about adding:

    from jupyterhub.objects import Hub
    hub = jupyterhub.objects.Hub.from_url('http://1.2.3.4:8081/hub/')
 
Would that be sufficient?

Setting cookies requires full access to the database for the User object, which seems a higher bar than a small wrapper around url and cookie info.

Carreau:
> What about adding:
> 
> from jupyterhub.objects import Hub
> hub = jupyterhub.objects.Hub.from_url('http://1.2.3.4:8081/hub/')
> Would that be sufficient?

It probably would, I'll try. I mostly get thrown off by various level of indirection.

manics:
Is this still relevant to the current JupyterHub?

manics:
No response, closing

PR #1274, opened 2017-07-26
simplify Hub object a bit

minrk:
- remove use of deprecated Hub.server
- add deprecation warning to Hub.server property
- move cookie_name declaration to Hub, so it doesn't need to be passed to the constructor.
  It's not configurable, anyway.

Related to #1270

### DIFF
--- a/jupyterhub/objects.py
+++ b/jupyterhub/objects.py
@@ -5,8 +5,7 @@
 
 import socket
 from urllib.parse import urlparse
-
-from tornado import gen
+import warnings
 
 from traitlets import (
     HasTraits, Instance, Integer, Unicode,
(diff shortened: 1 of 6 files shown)


jupyterhub/objects.py
```diff
@@ -136,9 +135,14 @@ class Hub(Server):
     of the server base_url.
     """
 
+    cookie_name = 'jupyter-hub-token'
```
willingc: Perhaps rename to `hub-token` or `hub-cookie-token`?

willingc: The way it is named now makes me think that the jupyter notebook is involved.

minrk: It should be `jupyterhub`, since it's the cookie for jupyterhub, but we can't change it without logging everybody out on upgrade. I've only moved it so that it's the class default.

willingc: Ah...I see what you did. It was in `app.py` and is now in `objects.py`.

willingc:
Thanks @minrk.

PR #1275, opened 2017-07-26
add log statement when spawn starts

minrk:
for easier monitoring of spawn time

also improve consistency of named spawner logging


willingc:
Nice!

PR #1280, opened 2017-07-27
Add hub_connect_port 

### LINKED ISSUE #1279: Allow setting `hub_connect_port` too
Currently it looks like we can only set `hub_connect_ip` - not the port. If the hub is receiving traffic on a different port than it is listening on (because there's an nginx in front, for example) this doesn't work.

yuvipanda:
Fixes #1279


minrk:
Should we just call this a full `hub_connect_url` and be done with it?

PR #1281, opened 2017-07-27
eliminate test application thread

minrk:
Thanks to @willingc for finding pytest-tornado, we can run tests in the application thread.

To avoid blocking the IOLoop, instead of trading requests for tornado's AsyncHTTPClient, I added `utils.requests_async`, which just plops the request in a ThreadPoolExecutor. It's still a thread, but the important thing is that test code isn't run in a thread (no concurrent connections to the database). Only actual requests run in threads, which should be fine.

This should hopefully eliminate at least a few of the intermittent failures we have been seeing.

TODO:

- [x] remove some remaining run_sync tests in favor of `@gen_test`
- [x] remove some of the actual application code added to handle the test thread

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,9 @@ python:
   - 3.6
   - 3.5
   - 3.4
+env:
+  global:
+    - ASYNC_TEST_TIMEOUT=15
 
 # installing dependencies
 before_install:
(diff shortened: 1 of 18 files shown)


minrk:
woo! It seems to work. I'm sure there are still one or two potential race conditions in the tests (how could there not be?), but hopefully this will improve things. Since this touches ~all of the tests, I'd like to land it quickly to avoid conflicts.

willingc:
Thanks @minrk! Super happy to :ship: this PR.

PR #1285, opened 2017-07-27
Try enforcing methods overwrite at import time.

Carreau:
Currently Spawners need to overwrite start,stop,poll and need to make
them tornado.gen.coroutine, if this is not done, it will fail at
runtime, maybe not fail if you do not make a coroutine.

This replicate this check at class definition time, meaning that
potential errors will be caught way earlier. It also have not runtime
cost as the check is a class definition time (ie often import time).

This takes only effect on Python 3.6+ which introduce __init_subclass__,
we could do it with metaclasses, but that's might be too complicated.

If one want to create a class the avoid these restriction they can
overwrite __init_subclass__ and not call the super() method.

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -56,6 +56,18 @@ class Spawner(LoggingConfigurable):
     orm_spawner = Any()
     user = Any()
 
+    def __init_subclass__(cls, **kwargs):
+        super().__init_subclass__()
+
+        missing = []
+        for attr in ('start','stop', 'poll'):
+            if getattr(Spawner, attr) is getattr(cls, attr):
+                missing.append(attr)
+
+        if missing:
+            raise NotImplementedError("class `{}` needs to redefine the `start`,"
+                  "`stop` and `poll` methods. `{}` not redefined.".format(cls.__name__, '`, `'.join(missing)))
+
     @property
     def server(self):
         if self.orm_spawner and self.orm_spawner.server:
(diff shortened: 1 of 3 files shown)


jupyterhub/services/service.py
```diff
@@ -101,6 +103,7 @@ def user_env(self, env):
         else:
             return super().user_env(env)
 
+    @gen.coroutine
```
minrk: This shouldn't need to be added.

jupyterhub/spawner.py
```diff
@@ -56,6 +56,25 @@ class Spawner(LoggingConfigurable):
...
 
+    def __init_subclass__(cls, **kwargs):
+        super().__init_subclass__()
+
+        missing = []
+        not_coro = []
+        for attr in ('start','stop', 'poll'):
+            if getattr(Spawner,attr) is getattr(cls, attr):
+                missing.append(attr)
+            if not getattr(getattr(cls, attr), '__tornado_coroutine__', None):
```
minrk: Can we remove the coroutine check? I don't think it solves a problem and it introduces an artificially strict check. For instance, a method that returns a Future is perfectly valid. All that really matters is that the return value is awaitable.

jupyterhub/tests/mocking.py
```diff
@@ -74,17 +74,30 @@ def stop(self):
...
 class NeverSpawner(MockSpawner):
     """A spawner that will never start"""
-    
+
     @default('start_timeout')
     def _start_timeout_default(self):
         return 1
-    
+
+    @fake_gen
```
minrk: returning a Future should be valid. If it's not, the check is making an inappropriate assertion.

Carreau:
Remove all the things checking methods are coroutines.

minrk:
Thanks!

PR #1287, opened 2017-07-28
Allow Hub to listen on a unix-socket

### LINKED ISSUE #1283: Allow hub to be listening on a unix domain socket
If there's a frontend proxy (*not* CHP!) doing the http heavy lifting in front of tornado, there's performance to be gained from using a unix domain socket for tornado to listen on rather than a TCP socket - see https://stackoverflow.com/questions/14973942/performance-tcp-loopback-connection-vs-unix-domain-socket.

Right now since we ask for a bind 'IP' and 'port' we can't use a unix domain socket.

Carreau:
Add the hub_socket option to the JupyterHub class, which takes
precedence over the hub_ip and hub_port setting. It does not forward
this setting to the Hub class though, and a few log messages still say
the hub is listening on `http://:8000` that works fine when testing with
netcat:

```
$ nc -U /tmp/jhub.sock
GET /login HTTP/1.1

HTTP/1.1 302 Found
Server: TornadoServer/4.5.1
Content-Type: text/html; charset=UTF-8
Date: Fri, 28 Jul 2017 02:05:36 GMT
X-Jupyterhub-Version: 0.8.0.dev
Content-Security-Policy: frame-ancestors 'self'; report-uri /hub/security/csp-report
Location: /hub/login
Content-Length: 0
```

Closes #1283

Should still be better documented I guess.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -18,6 +18,7 @@ import sys
 from textwrap import dedent
 from urllib.parse import urlparse
 
+
 if sys.version_info[:2] < (3, 3):
     raise ValueError("Python < 3.3 not supported: %s" % sys.version)
 
@@ -32,6 +33,7 @@ from tornado.ioloop import IOLoop, PeriodicCallback
 from tornado.log import app_log, access_log, gen_log
 import tornado.options
 from tornado import gen, web
+from tornado.netutil import bind_unix_socket
 
 from traitlets import (
     Unicode, Integer, Dict, TraitError, List, Bool, Any,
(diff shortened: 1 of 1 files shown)


yuvipanda:
I don't think we should do this like this.

After 0.8 is out, we should do a systematic overhaul of differences between bind addresses and connect addresses, and properly separate / unify them. We shouldn't have a separate option for a unix socket listener - we already have too many.

yuvipanda:
I proposed https://github.com/jupyterhub/jupyterhub/issues/1289

minrk:
While this will let the Hub bind, it isn't enough to work because hub_ip, etc. are still being used elsewhere by other services to connect to the Hub, and the Hub isn't listening there anymore.

willingc:
@minrk @Carreau Should we close this PR? Or leave open for 0.9?

yuvipanda:
I'd like us to take a slightly different approach that I outlined in https://github.com/jupyterhub/jupyterhub/issues/1289. What do you think, @Carreau?

PR #1290, opened 2017-07-28
Add support for limiting the number of concurrent spawns

yuvipanda:
This lets admin throttle maximum number of concurrent pending spawns.

It'll fail them early, which lets us maintain a consistent level of service for the users we 
can service. Users can try again soon!

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -545,16 +545,14 @@ class JupyterHub(Application):
     ).tag(config=True)
 
     concurrent_spawn_limit = Integer(
-        None,
-        allow_none=True,
+        0,
         help="""
         Maximum number of concurrent users that can be spawning at a time.
 
-        If more than this many users attempt to spawn at a time, they
-        enter an exponential delay with a timeout.
+        If more than this many users attempt to spawn at a time, their
+        request is rejected with a 429 error asking them to try again.
 
-        If set to `None`, no concurrent limit is enforced. If set to 0,
-        then no spawning is allowed.
+        If set to 0, no concurrent_spawn_limit is enforced.
         """
     ).tag(config=True)
 
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -531,16 +531,30 @@ class JupyterHub(Application):
     @default('authenticator')
     def _authenticator_default(self):
         return self.authenticator_class(parent=self, db=self.db)
-    
+
     allow_named_servers = Bool(False,
         help="Allow named single-user servers per user"
     ).tag(config=True)
-    
+
     # class for spawning single-user servers
     spawner_class = Type(LocalProcessSpawner, Spawner,
         help="""The class to use for spawning single-user servers.
 
         Should be a subclass of Spawner.
+        """
+    ).tag(config=True)
+
+    concurrent_spawn_limit = Integer(
+        None,
+        allow_none=True,
+        help="""
+        Maximum number of concurrent users that can be spawning at a time.
+
+        If more than this many users attempt to spawn at a time, they
+        enter an exponential delay with a timeout.
+
+        If set to `None`, no concurrent limit is enforced. If set to 0,
+        then no spawning is allowed.
         """
     ).tag(config=True)
 
@@ -1245,6 +1259,7 @@ class JupyterHub(Application):
             statsd=self.statsd,
             allow_named_servers=self.allow_named_servers,
             oauth_provider=self.oauth_provider,
+            concurrent_spawn_limit=self.concurrent_spawn_limit,
         )
         # allow configured settings to have priority
         settings.update(self.tornado_settings)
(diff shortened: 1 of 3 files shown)


jupyterhub/handlers/base.py
```diff
@@ -366,6 +366,18 @@ def spawner_class(self):
     def spawn_single_user(self, user, server_name='', options=None):
         if server_name in user.spawners and user.spawners[server_name]._spawn_pending:
             raise RuntimeError("Spawn already pending for: %s" % user.name)
+
+        if self.settings['concurrent_spawn_limit'] is not None and \
```
willingc: Let's remove the `\` and put all in parens. [Ref](http://docs.python-guide.org/en/latest/writing/style/#line-continuations)

jupyterhub/handlers/base.py
```diff
@@ -366,6 +366,18 @@ def spawner_class(self):
...
+           self.hub.pending_spawns > self.settings['concurrent_spawn_limit']:
+            self.log.info(
+                'More than %s pending spawns, throttling',
+                self.settings['concurrent_spawn_limit']
+            )
+            raise web.HTTPError(
+                429,
+                "Too many users are starting now, try again shortly")
```
willingc: Questions on error message wording: Does the user/admin have to do anything to try again? Will an action need to be taken on the error message?

jupyterhub/objects.py
```diff
@@ -149,6 +149,15 @@ class Hub(Server):
 
     cookie_name = 'jupyter-hub-token'
 
+    pending_spawns = Integer(
```
willingc: There's something about the variable name `pending_spawns` that doesn't feel quite right. `spawns_pending` seems to be a bit better or something else without `pending` perhaps.

yuvipanda: pending_spawn_count? spawn_pending_count? We already use `spawner._spawn_pending`, so perhaps `spawn_pending_count` to be consistent with that?

jupyterhub/handlers/base.py
yuvipanda: The user should just try again in a few minutes - it's a pure rate limit, nothing for either of them to do. We should definitely word this better!

jupyterhub/objects.py
willingc: `spawn_pending_count` works for me.

jupyterhub/handlers/base.py
willingc: Just so I understand...The user would see the error message. Then, the user would wait a few minutes and press the start server button again. Is that correct?

yuvipanda: Yep, that is correct!

willingc: Perhaps for the error message:

`"User startup rate limit exceeded. Try to start again in a few minutes."`

jupyterhub/handlers/base.py
```diff
@@ -402,6 +414,7 @@ def finish_user_spawn(f=None):
                 spawner.add_poll_callback(self.user_stopped, user)
             finally:
                 spawner._proxy_pending = False
+                self.hub.spawn_pending_count -= 1
```
minrk: The `hub` object is only meant to contain connection info (a tiny wrapper around URLs), not runtime info like this. Maybe store this in self.settings, instead?

jupyterhub/app.py
```diff
@@ -544,6 +544,20 @@ def _authenticator_default(self):
...
+        None,
+        allow_none=True,
+        help="""
+        Maximum number of concurrent users that can be spawning at a time.
+
+        If more than this many users attempt to spawn at a time, they
+        enter an exponential delay with a timeout.
+
+        If set to `None`, no concurrent limit is enforced. If set to 0,
+        then no spawning is allowed.
```
minrk: A Hub with no spawning allowed doesn't make sense (current 429 try again later would not be appropriate in this case). limit=0 should be no limit (and the default; I'd like to avoid `allow_none=True` on number traits in general).

jupyterhub/app.py
```diff
@@ -544,6 +544,20 @@ def _authenticator_default(self):
...
     ).tag(config=True)
 
+    concurrent_spawn_limit = Integer(
+        None,
+        allow_none=True,
+        help="""
+        Maximum number of concurrent users that can be spawning at a time.
+
+        If more than this many users attempt to spawn at a time, they
+        enter an exponential delay with a timeout.
```
minrk: Exponential delay isn't implemented. When the limit is reached, it fails with 429 and it is incumbent on the user to retry later. I think the current behavior is sensible, but we should update the message to be accurate.

yuvipanda: Ah, leftover from my first implementation. Fixed.

jupyterhub/app.py
yuvipanda: Changed.

jupyterhub/handlers/base.py
yuvipanda: I think we should factor stuff out of basehandler at some point - using settings for this feels icky, but is probably the right thing to do for now.

yuvipanda:
@willingc yup!

minrk:
Nice!

PR #1297, opened 2017-08-01
prevent saving auth_state if auth_state is not enabled.

minrk:
allows Authenticators to always return auth_state without having to check enable_auth_state themselves.

Came up in https://github.com/jupyterhub/oauthenticator/pull/97#issuecomment-319304667


yuvipanda:
I like it!

PR #1298, opened 2017-08-01
[WIP] Default pre_spawn_start loads env from auth_state.env if defined

minrk:
Authenticators that return auth_state with `env` defined will automatically set environment for the Spawner.

### DIFF
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -269,11 +269,23 @@ class Authenticator(LoggingConfigurable):
                 of auth state that will be persisted.
         """
 
+    @gen.coroutine
     def pre_spawn_start(self, user, spawner):
         """Hook called before spawning a user's server
 
-        Can be used to do auth-related startup, e.g. opening PAM sessions.
+        Can be used to do auth-related startup, e.g. opening PAM sessions
+        or specifying auth-related environment variables.
+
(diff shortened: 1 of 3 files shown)


yuvipanda:
This puts default control of what's in env in the hands of authenticator authors. I think this means most authenticators will only define env and nothing else :D I think instead we should have a `env_from_auth_state` list or something that lets people specify a dict mapping from auth state to env. That way it's explicitly inserted into env based on requirements rather than showing up by default. We could also let people specify '*' or something like that to merge all the auth_state keys into env...

I mostly just don't want to force authenticator authors to dictate what goes into env :)

https://github.com/yuvipanda/paws/blob/master/paws/values.yaml#L12 is what I'm using now!

minrk:
That makes sense. Any good ideas for how to do that? I'm thinking a Python function might be the best option rather than trying to come up with something declarative but sufficiently expressive. I'm also AOK completely punting on this and having nothing built-in for 0.8 and waiting to see what people end up needing down the line. I'd rather do that than commit to something we don't like just before release.

yuvipanda:
I think I'd prefer we punt this to post 0.8. we could push it out in a
point release. Making sure we document how people can do it before that is
good enough I think.

minrk:
Sounds good to me. Marking as 0.9.

PR #1301, opened 2017-08-02
add active_server_limit

### LINKED ISSUE #1282: Allow limiting total hub resources or the number of notebook servers
Right now jupyterhub *almost* allows to replace `tmpnb` by using `tmpauthenticator`. Since jupyterhub is much more developed and configurable than `tmpnb` (or at least it seems so), it would be a great way of replacing `tmpnb`. However, while it is relatively easy to limit resources of a single server, it is currently hard/impossible to configure the total amount of servers that run in parallel, opening a possibility of a very simple DOS attack.

Therefore it would be useful it it was possible to restrict the total number of single user servers running in parallel, since it would allow anyone to simply run their own configurable tmpnb's.

minrk:
This is the total user limit, which deployments like tmpnb will want to set to limit resource usage. If hit, throws the same 429 error as concurrent_spawn_limit.

For managing this, I also added properties on Spawner for the various active/ready/pending states so they are easier to check. I used these properties to implement a `users.count_active_users` method for collecting and reporting totals of the various active/pending counts, which is used for limit enforcement rather than adding more running bookkeeping, which I think is brittle with many different avenues for spawn failure. I verified that computing this value takes only a few milliseconds for 5k users (db is never touched), so I think it's okay.

closes #1282

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -552,10 +552,35 @@ class JupyterHub(Application):
         help="""
         Maximum number of concurrent users that can be spawning at a time.
 
-        If more than this many users attempt to spawn at a time, their
-        request is rejected with a 429 error asking them to try again.
+        Spawning lots of servers at the same time can cause performance
+        problems for the Hub or the underlying spawning system.
+        Set this limit to prevent bursts of logins from attempting
+        to spawn too many servers at the same time.
 
-        If set to 0, no concurrent_spawn_limit is enforced.
+        This does not limit the number of total running servers.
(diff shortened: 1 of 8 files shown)


yuvipanda:
Note on the naming - Should this be 'concurrent_server_limit'? I guess we do want to limit users rather than servers, but what if one user ends up with 500 servers in the future (with multiple servers per user)

minrk:
Good call on the naming. I called it `active_server_limit`. I think we can cross the per-user server limit bridge when we come to it.

PR #1303, opened 2017-08-02
consolidate server APIHandlers

### LINKED ISSUE #1256: 'Stop server' raises an alert if it gets back a 202 Accepted
If stopping a server is slow, the JupyterHub API responds with a '202 Accepted'. The UI should do something with this - right now it just throws up an error box.

minrk:
no need for separate handlers for named and default servers

avoid claiming that empty responses are JSON.

Fixes 202 error handling by setting content-type: text/plain when sending empty responses.

Errors were raised in jQuery because response was empty but content-type header claimed it was application/json.

closes #1256

### DIFF
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -156,9 +156,9 @@ class UserAPIHandler(APIHandler):
         yield gen.maybe_future(self.authenticator.delete_user(user))
         # remove from registry
         del self.users[user]
-        
+
         self.set_status(204)
-    
+
     @admin_only
     def patch(self, name):
         user = self.find_user(name)
@@ -194,6 +194,8 @@ class UserServerAPIHandler(APIHandler):
         options = self.get_json_body()
         yield self.spawn_single_user(user, options=options)
         status = 202 if user.spawner._spawn_pending else 201
+        if status == 202:
+            self.write(json.dumps({"message": "Server is slow to start"}))
         self.set_status(status)
 
     @gen.coroutine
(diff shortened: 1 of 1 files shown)


jupyterhub/apihandlers/users.py
```diff
@@ -183,79 +183,49 @@ class UserServerAPIHandler(APIHandler):
     """
     @gen.coroutine
     @admin_or_self
-    def post(self, name):
+    def post(self, name, server_name=''):
```
willingc: Class docstring needs to be updated since allow_named_servers can be True.

jupyterhub/apihandlers/users.py
```diff
@@ -183,87 +183,49 @@ class UserServerAPIHandler(APIHandler):
...
-                raise web.HTTPError(400, "%s's server is already running" % name)
+                raise web.HTTPError(400, "%s's server %s is already running" % (name, server_name))
 
         options = self.get_json_body()
-        yield self.spawn_single_user(user, options=options)
-        status = 202 if user.spawner._spawn_pending else 201
-        if status == 202:
-            self.write(json.dumps({"message": "Server is slow to start"}))
+        yield self.spawn_single_user(user, server_name, options=options)
+        status = 202 if spawner.pending in {'spawn', 'proxy'} else 201
```
willingc: May want to comment this line re: dictionary of the two pending states 'spawn' and 'proxy'. I'm not quite sure where the 'proxy' state is set.

minrk: The 'proxy' state is gone now, so I've removed it!

jupyterhub/apihandlers/users.py
minrk: Good catch, done.

jupyterhub/apihandlers/users.py
willingc: Good to know ;-)

willingc:
Thanks!

PR #1306, opened 2017-08-04
remove bower

minrk:
use npm to fetch dependencies and a simple postinstall script to copy into components


.travis.yml
```diff
@@ -21,6 +21,7 @@ install:
 
 # running tests
 script:
+  - cd /tmp
   - pytest -v --maxfail=2 --cov=jupyterhub jupyterhub/tests
```
willingc: Path of `jupyterhub/tests` needs to be updated.

willingc: Will give it a look.

minrk:
Not worth fiddling with this just before release. Marking as 0.9.

.travis.yml
willingc: @minrk I am able to get the tests to run locally and on Travis by removing `cd /tmp` and adding:
```
script:
    pytest -v --basetemp=tmp --maxfail=2 --cov=jupyterhub jupyterhub/tests
```
[ref](https://docs.pytest.org/en/latest/tmpdir.html#the-default-base-temporary-directory)

Though when making this change, one test fails `test_static_files` in `test_pages.py`.  [test output link](https://travis-ci.org/willingc/jupyterhub/jobs/261573175)

willingc: A bit more info: 

Did the location of CSS files change (or stay the same) with the bower-lite changes? I commented out part of the `test_static_files` that was causing the test failure and all passed.

```
 #r = yield async_requests.get(ujoin(base_url, 'static', 'css', 'style.min.css'))
 #r.raise_for_status()
 #assert r.headers['content-type'] == 'text/css'
```

[Passing tests on Travis](https://travis-ci.org/willingc/jupyterhub/jobs/261604492)

minrk: Thanks for investigating. The generated files haven't moved, and the cd to `/tmp` shouldn't be necessary, I just added it for debugging Travis. I, too, don't have issues running the tests locally, it is only Travis that seems to have the issue. I assumed it would be because the generated files aren't being installed in the working dir, and the installed version is what's used, not the local one. I can't figure out why the changes here would have any effect on that, since it doesn't (knowingly) affect when or how the css files are generated.

willingc:
@minrk I think that I have resolved this issue with the test. Heading out for an appointment but the changes below pass Travis. Short story the second `npm install` is needed to get the css in the right static directory.

https://travis-ci.org/willingc/jupyterhub/builds/262344046

screen shot 2017-08-08 at 10 45 27 am

.travis.yml
chavellacinema160: Remove

minrk:
Figured it out, just needed to add a check for the css outputs when looking for data_dir.

PR #1312, opened 2017-08-05
reverse generated-tokens default logic

minrk:
user-provided tokens are added in exactly one place, so switch default handling of tokens to `generated=True` and explicitly distrust user tokens.

Adds `JupyterHub.trust_user_provided_tokens` flag so that users can avoid the extra hashing
if they know they are providing good tokens.


willingc:
Thanks @minrk. Looks good. Well documented too.

PR #1319, opened 2017-08-07
Test against postgres, mysql (again)

minrk:
It works!

<del>Now that we have dropped the test thread, this should be doable. Almost all passing, and I think the remaining failures are leaking data from one test run to the next, since the temp database is cleared per file, while the real databases persist for the session (useful to test!).
</del>

The main changes here are reducing some unnecessarily long strings from Unicode(1023) to Unicode(255), which is the max a default mysql setup can handle on unique columns (InnoDB max length is 767 = 256 * 3 - 1). This change doesn't require a db migration, because any existing db that was able to create the columns won't be affected, they will just have higher limits than they really need.

### DIFF
--- a/jupyterhub/oauth/store.py
+++ b/jupyterhub/oauth/store.py
@@ -52,14 +52,7 @@ class JupyterHubSiteAdapter(AuthorizationCodeGrantSiteAdapter):
 class HubDBMixin(object):
     """Mixin for connecting to the hub database"""
     def __init__(self, session_factory):
-        self.session_factory = session_factory
-        self._local = threading.local()
-
-    @property
-    def db(self):
-        if not hasattr(self._local, 'db'):
-            self._local.db = scoped_session(self.session_factory)()
-        return self._local.db
+        self.db = session_factory()
 
 
 class AccessTokenStore(HubDBMixin, oauth2.store.AccessTokenStore):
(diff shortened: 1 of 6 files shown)


jupyterhub/oauth/store.py
```diff
@@ -77,7 +70,6 @@ def save_token(self, access_token):
         if user is None:
             raise ValueError("No user for access token: %s" % access_token.user_id)
         orm_access_token = orm.OAuthAccessToken(
-            generated=True,
```
minrk: This is removed because it's redundant after #1312

jupyterhub/oauth/store.py
```diff
@@ -52,14 +52,7 @@ def user_has_denied_access(self, request):
...
     def __init__(self, session_factory):
-        self.session_factory = session_factory
-        self._local = threading.local()
-
-    @property
-    def db(self):
-        if not hasattr(self._local, 'db'):
-            self._local.db = scoped_session(self.session_factory)()
-        return self._local.db
+        self.db = session_factory()
```
minrk: This avoids giving OAuth a separate session from the rest of the app (again, related to the test threads that are now gone)

jupyterhub/tests/mocking.py
```diff
@@ -200,17 +200,20 @@ def init_signal(self):
...
         pass
-    
+
     @gen.coroutine
     def initialize(self, argv=None):
-        self.db_file = NamedTemporaryFile()
         self.pid_file = NamedTemporaryFile(delete=False).name
-        self.db_url = self.db_file.name
+        self.db_file = NamedTemporaryFile()
+        self.db_url = os.getenv('JUPYTERHUB_TEST_DB_URL') or self.db_file.name
```
willingc: Assuming the env variable will let us test a basic Postgres and mysql db in Travis going forward.

minrk: Yup, this env variable is in use now to test with postgres/mysql on Travis.

PR #1325, opened 2017-08-08
Update services getting started docs

### LINKED ISSUE #1078: Accessing the Jupyter Server API when using Jupyterhub
The documentation is unclear on how to access the [Jupyter Notebook API
](https://github.com/jupyter/jupyter/wiki/Jupyter-Notebook-Server-API) when using Jupyterhub. 

[Jupyterhub Documentation](http://jupyterhub.readthedocs.io/en/latest/rest.html) states : 
```
Note that the token authorizes JupyterHub REST API requests.
 The same token does not authorize access to the Jupyter Notebook REST API provided by notebook servers managed by JupyterHub.
```
I setup a system wide jupyter_notebook_config.py in /etc/jupyter using "c.NotebookApp.token" expecting to be able to access the notebook API as specified in the docs. 
```
curl -H "Authorization: Token TOKEN" -H "Content-Type: application/json"  -XGET http://localhost:8000/user/username/api
```
All attempts to do so seem to push me to the Jupyterhub login page which is confusing since the jupyterhub documentation states that you cannot authenticate to the Notebook REST API with a Jupyterhub token.

willingc:
Partially addresses #962
Closes #1078 

- provide more clarity on process steps using cull idle server as an example
- deprecate 'jupyterhub token' @minrk if you are cool doing so.
- add example for requests/API from #962 
- add note about tokens and auth to single-user servers in 0.8 #1078
- add small section on using API tokens for service activities that are not tied to a specific user #962

### DIFF
--- a/docs/source/rest.md
+++ b/docs/source/rest.md
@@ -68,6 +68,26 @@ r.raise_for_status()
 users = r.json()
 ```
 
+This example provides a slightly more complicated request, yet the
+process is very similar:
+
+```python
+import requests
+
+api_url = 'http://127.0.0.1:8081/hub/api'
+
+data = {'name': 'mygroup', 'users': ['user1', 'user2']}
+
+r = requests.post(api_url + '/groups/formgrade-data301/users',
+    headers={
+             'Authorization': 'token %s' % token,
+            },
+    json=data
+)
+r.raise_for_status()
+r.json()
+```
+
 Note that the API token authorizes **JupyterHub** REST API requests. The same
 token does **not** authorize access to the [Jupyter Notebook REST API][]
 provided by notebook servers managed by JupyterHub. A different token is used
--- a/docs/source/rest.md
+++ b/docs/source/rest.md
@@ -34,12 +34,26 @@ generating an API token is:
 openssl rand -hex 32
 ```
 
-Alternatively, use the `jupyterhub token` (*deprecated in version 0.8.0*)
-command to generate a token for a specific hub user:
+This `openssl` command generates a potential token that can then be
+added to JupyterHub using `.api_tokens` configuration setting in
+`jupyterhub_config.py`.
+
+
+Alternatively, use the `jupyterhub token` command to generate a token
+for a specific hub user by passing the 'username':
 
 ```bash
 jupyterhub token <username>
 ```
+
+This command generates a random string to use as a token and registers
+it for the given user with the Hub's database.
+
+In [version 0.8.0](./changelog.html), a TOKEN request page for
+generating an API token is available from the JupyterHub user interface:
+
+![Request API TOKEN page](images/api-token-request.png)
+
 
 ## Add API tokens to the config file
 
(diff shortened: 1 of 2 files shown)


docs/source/rest.md
```diff
@@ -24,16 +24,28 @@ Hub.
...
 
-    jupyterhub token USERNAME
+As of [version 0.6.0](./changelog.html), the preferred way of
+generating an API token is:
+
+```bash
+openssl rand -hex 32
+```
+
+Alternatively, use the `jupyterhub token` (*deprecated in version 0.8.0*)
```
minrk: `jupyterhub token` is not deprecated. It does two things:

1. generate a random string
2. register it for the given user with the database

The `openssl` command only generates a *potential* token, but it still must be added to the db via the `.api_tokens` mechanism. It's totally fine to recommend using `jupyterhub token`.

It would probably also be appropriate to add a mention here of the new Token request page in 0.8.

docs/source/services-basics.md
```diff
@@ -1,36 +1,131 @@
...
+openssl rand -hex 32
 ```
 
-As of [version 0.6.0](./changelog.html), the preferred way of doing this is to first generate an API token:
+Alternatively, use the `jupyterhub token` (*deprecated in version 0.8.0*)
+command to generate a token for a specific hub user:
 
 ```bash
-openssl rand -hex 32
+jupyterhub token <username>
```
minrk: Here, I would remove the mention of `jupyterhub token`. While it's not deprecated, it only generates *user* tokens, and we want to encourage use of services.

docs/source/services-basics.md
```diff
@@ -1,36 +1,131 @@
...
+
+### Pass token using a configuration file
 
-and then write it to your JupyterHub configuration file (note that the **key** is the token while the **value** is the username):
+You can write the API token to your JupyterHub
+configuration file, `jupyterhub_config.py` (note that
+the **key** is the 'token' while the **value** is the 'username'):
 
 ```python
 c.JupyterHub.api_tokens = {'token' : 'username'}
```
minrk: Let's remove this mention of using `.api_tokens` for services, and leave only the `.services` config below.

docs/source/services-basics.md
```diff
@@ -1,36 +1,131 @@
...
+        'admin': True,
+        'command': 'python cull_idle_servers.py --timeout=3600'.split(),
+    }
+]
+```
+
+where:
+
+- `'admin': True` indicates that the Service has 'admin' permissions, and
+- `'command'` indicates that the Service will be managed by the Hub.
```
minrk: perhaps "...will be launched as a subprocess, managed by the Hub.

docs/source/services-basics.md
```diff
@@ -1,36 +1,131 @@
...
 
 ```bash
-openssl rand -hex 32
+jupyterhub token <username>
 ```
 
+### Pass environment variable with token to the Hub
+
+In the case of `cull_idle_servers`, it is passed as the environment
+variable called `JPY_API_TOKEN`.
```
minrk: JUPYTERHUB_API_TOKEN

PR #1328, opened 2017-08-09
Update pin on sqlalchemy for travis

willingc:
When investigating a failed test in this [travis build](https://travis-ci.org/jupyterhub/jupyterhub/jobs/262762543), the following error was happening during the pip install of requirements:

screen shot 2017-08-09 at 12 36 47 pm

For some reason the existing entry in `requirements.txt` as well as `sqlalchemy>=1.1` and `sqlalchemy>=1.2` all throw the same error. 

@minrk Would you mind looking at the test failures in the linked build? Thanks.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,5 +4,5 @@ tornado>=4.1
 jinja2
 pamela
 python-oauth2>=1.0
-sqlalchemy>=1.0
+sqlalchemy>=1.2.0b2
 requests


minrk:
Looks like we may have introduced a race condition that can get triggered when mysql is slow. I'll investigate.

minrk:
Looks like we may have introduced a race condition that can get triggered when mysql is slow. I'll investigate.

minrk:
Oh, I didn't realize what this was doing - it's actually bumping the hard limit of the minimum version to install JupyterHub for *everyone*, not just pinning for Travis (requirements.txt is JupyterHub's actual strict dependency list). Is there a reason that JupyterHub doesn't work with sqlalchemy 1.0 anymore?

willingc:
@minrk Yes, JupyterHub works with sqlalchemy 1.0. I'm working on going back to the original requirements.txt but also trying to figure out WTH is up with travis.

PR #1331, opened 2017-08-10
verify requirements installed in travis builds

willingc:
@minrk This is probably a better solution than #1328

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,7 @@ before_install:
 install:
   - pip install -U pip
   - pip install --pre -r dev-requirements.txt .
+  - pip freeze
 
 # running tests
 script:
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,5 +4,5 @@ tornado>=4.1
 jinja2
 pamela
 python-oauth2>=1.0
-SQLAlchemy>=1.0,<1.2
+SQLAlchemy>=1.0
 requests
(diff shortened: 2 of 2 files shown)


minrk:
Is 1.2 the source of a known problem? Do we want to pin it down? I don't understand what issue we are solving by preventing our users from being allowed to install sqlalchemy 1.2 with JupyterHub. Master is currently pinned *to* 1.2b2 and passing, so it seems like we should put it back to `>= 1.0`.

willingc:
Two separate issues: users and travis testing. 

I don't think there is an issue with 1.2, and I agree we do want to have the users use whatever upward version. So I'm fine with >=1.0 for users.

The issue is really more on Travis with the install of the environment (https://github.com/jupyterhub/jupyterhub/pull/1328#issue-249139606). I'm not all that happy with Travis displaying this error. Likely, better to avoid it. I know from my experience with Django that database testing (particularly with mysql) is pretty finicky with versions. Most of the time Django devs will pin a specific version for each requirement.

For now, how about I modify it back to `>=1.0` and if we run into any issues with testing down the road we could always specify a particular version in the `travis.yml` to install. Thoughts?

willingc:
Now that I'm reviewing the oauthenticator release PR. I'm noticing that you had a `pip freeze` in the `.travis.yml` to display the versions currently installed. I'm going to amend this PR to add that so even with the error message that we will know what definitely is being used in the test container.

minrk:
Sounds like a good idea!

PR #1340, opened 2017-08-14
Fix broken jupyterhub getting started link

zonca:


willingc:
Good catch. Thanks for the fix @zonca

PR #1344, opened 2017-08-16
Start drafting 0.8 changelog

minrk:
Lots of good stuff in this release!

cc @willingc

### DIFF
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -7,11 +7,89 @@ command line for details.
 
 ## [Unreleased] 0.8
 
+JupyterHub 0.8 is a big release!
+
+Perhaps the biggest change is the use of OAuth to negotiate authentication
+between the Hub and single-user services.
+Due to this change, it is important that the single-user server
+and Hub are both running the same version of JupyterHub.
+If you are using containers (e.g. via DockerSpawner or KubeSpawner),
+this means upgrading jupyterhub in your user images at the same time as the Hub.
+In most cases, a
+
+    pip install jupyterhub==version
+
+in your Dockerfile is sufficient.
+
 #### Added
+
+- JupyterHub now defined a `.Proxy` API for custom
+  proxy implementations other than the default.
+  The defaults are unchanged,
+  but configuration of the proxy is now done on the `ConfigurableHTTPProxy` class instead of the top-level JupyterHub.
+  TODO: docs for writing a custom proxy.
+- Single-user servers and services
+  (anything that uses HubAuth)
+  can now accept token-authenticated requests via the Authentication header.
+- Authenticators can now store state in the Hub's database.
+  To do so, the `.authenticate` method should return a dict of the form
+
+  ```python
+  {
+      'username': 'name'
+      'state': {}
+  }
+  ```
+
+  This data will be encrypted and requires `JUPYTERHUB_CRYPT_KEY` environment variable to be set
+  and the `Authenticator.enable_auth_state` flag to be True.
+  If these are not set, auth_state returned by the Authenticator will not be stored.
+- There is preliminary support for multiple (named) servers per user in the REST API.
+  Named servers can be created via API requests, but there is currently no UI for managing them.
+- Add `LocalProcessSpawner.popen_kwargs` and `LocalProcessSpawner.shell_cmd`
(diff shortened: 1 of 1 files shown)


docs/source/changelog.md
```diff
@@ -7,12 +7,89 @@ command line for details.
...
+      'username': 'name'
+      'state': {}
+  }
+  ```
+
+  This data will be encrypted and requires `JUPYTERHUB_CRYPT_KEY` environment variable to be set
+  and the `Authenticator.enable_auth_state` flag to be True.
+  If these are not set, auth_state returned by the Authenticator will not be stored.
+- There is preliminary support for multiple (named) servers per user in the REST API.
+  There is currently no UI for creating named servers, but they can be created
```
willingc: I would flip this sentence. Multiple (named) servers can be created via API requests, but there is currently no UI to do so.

docs/source/changelog.md
```diff
@@ -7,12 +7,89 @@ command line for details.
...
+  for customizing how user server processes are launched.
+- Add `Authenticator.auto_login` flag for skipping the "Login with..." page explicitly.
+- Add `JupyterHub.hub_connect_ip` configuration
+  for the ip that should be used when connecting to the Hub.
+  This is promoting (and deprecating) `DockerSpawner.hub_ip_connect`
+  for use by all Spawners.
+- Add `Spawner.pre_spawn_hook(spawner)` hook for customizing
+  pre-spawn events.
+- Add `JupyterHub.active_server_limit` and `JupyterHub.concurrent_spawn_limit`
+  for limiting the total number of active users and the number of pending Spawns, respectively.
```
willingc: Perhaps: ...active users on a hub server...

docs/source/changelog.md
```diff
@@ -7,12 +7,89 @@ command line for details.
...
+
 #### Changed
 
+- more arguments to spawners are now passed via environment variables (`.get_env()`)
+  rather than CLI arguments (`.get_args()`)
+- internally generated tokens no longer get extra hash rounds,
+  significantly speeding up authentication.
+  The hash rounds were deemed unnecessary because the tokens were already
+  generated with high entropy.
+- `JUPYTERHUB_API_TOKEN` env is no longer removed during single-user start.
```
willingc: `JUPYTERHUB_API_TOKEN` env is available at all times (?), and it is no longer removed during...

docs/source/changelog.md
```diff
@@ -7,12 +7,89 @@ command line for details.
...
+
 #### Fixed
 
+So many things fixed!
+
+- Collisions are checked when users are renamed
+- Fix bug where OAuth authenticators could not logout users
+  due to being redirected right back through the login process.
+- Errors in config files prevent JupyterHub from starting
+  instead of skipping the configuration.
```
willingc: configuration and logging a warning (? not sure if this is true. if not something for 0.9)

docs/source/changelog.md
```diff
@@ -7,12 +7,89 @@ command line for details.
...
+- Raise 403 error on unauthorized user rather than redirect to login,
+  which could cause redirect loop.
+- Set httponly on cookies because it's prudent.
```
willingc: `httponly`

docs/source/changelog.md
```diff
@@ -7,12 +7,89 @@ command line for details.
...
+  To do so, the `.authenticate` method should return a dict of the form
+
+  ```python
+  {
+      'username': 'name'
+      'state': {}
+  }
+  ```
+
+  This data will be encrypted and requires `JUPYTERHUB_CRYPT_KEY` environment variable to be set
```
minrk: We don't! https://github.com/jupyterhub/jupyterhub/issues/1349

docs/source/changelog.md
minrk: Fixed. It's a traitlets flag where errors in config files can either be treated as warnings to be logged or fatal errors. An error in configuration now prevents the Hub from starting, with a (hopefully) informative traceback. In 0.7, an error in configuration would result in the Hub *warning* about failed config and proceeding to launch with default config, which is undesirable. Note should be clearer now.

willingc: Thanks for the explanation.

PR #1347, opened 2017-08-17
handle and test a few unlikely cases when Spawners reuse tokens

### LINKED ISSUE #1343: hub doesn't check if the reused token is valid
This is a follow-up of a discussion with @minrk in gitter.

Observed when starting a container using `DockerSpawner` that failed on changing permissions due to mounting a read-only volume in `/home/jovyan` using hub 0.7.2 and dockerspawner 0.7.0.

The initial spawn failed to start, that triggered token removal since the spawn failed, but on the other hand the container was still reused by the spawner and still used the old token.

If I understand @minrk's explanation correctly, the hub should check that the reused hub token is still valid.

minrk:
- test that .will_resume preserves tokens (worked, but wasn't tested)

If a Spawner reuses a token, validate it in the db:

- verify that it's in the db
- if it doesn't map onto the right user, revoke the token
- if it's not in the db, insert it as a user-provided token

The most likely case is prior unclean shutdown of something like DockerSpawner, where a spawn failed and thus the token was revoked from the db, but the container was in fact created
and the token-reuse logic reused the defunct token.

closes #1343

### DIFF
--- a/jupyterhub/tests/test_spawner.py
+++ b/jupyterhub/tests/test_spawner.py
@@ -15,11 +15,14 @@ from unittest import mock
 import pytest
 from tornado import gen
 
-from ..user import User
 from ..objects import Hub, Server
+from .. import orm
 from .. import spawner as spawnermod
 from ..spawner import LocalProcessSpawner, Spawner
-from .. import orm
+from ..user import User
+from ..utils import new_token
+from .mocking import MockSpawner
+from .test_api import add_user
 from .utils import async_requests
 
 _echo_sleep = """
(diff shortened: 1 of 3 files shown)


jupyterhub/tests/test_spawner.py
```diff
@@ -270,3 +273,90 @@ def stop():
...
+    assert user.spawner is spawner
+    # first start: gets a new API token
+    yield user.spawn()
+    api_token = spawner.api_token
+    found = orm.APIToken.find(app.db, api_token)
+    assert found
+    assert found.user.name == user.name
+    assert user.api_tokens == [found]
+    yield user.stop()
+    yield user.spawn()
```
willingc: I would suggest putting a comment before the second spawn
```# second spawn```

File: jupyterhub/tests/test_spawner.py [hunk 2]
```diff
@@ -270,3 +273,90 @@ def stop():
...
+    # verify re-use of API token
+    assert spawner.api_token == api_token
+    # verify that a new token was not created
+    assert user.api_tokens == [found]
+
+
+@pytest.mark.gen_test
+def test_spawner_insert_api_token(db, app):
+    user = add_user(app.db, app, name='tonkee')
```
willingc: ```# setup: add a user to db; check that no single user server running; no api_tokens exist for user```

jupyterhub/tests/test_spawner.py
```diff
@@ -270,3 +273,90 @@ def stop():
...
+    if user.running:
+        yield user.stop()
+    assert user.api_tokens == []
+
+    spawner = user._new_spawner(name='', spawner_class=MockAPITokenSpawner)
```
willingc: `# setup: set up a spawner for the user`

jupyterhub/tests/test_spawner.py
```diff
@@ -270,3 +273,90 @@ def stop():
...
+    user.spawners[''] = spawner
+    assert user.spawner is spawner
+
+    api_token = new_token()
```
willingc: `# test: generate a new token; test token is not in the db`

jupyterhub/tests/test_spawner.py
```diff
@@ -270,3 +273,90 @@ def stop():
...
+    assert not orm.APIToken.find(app.db, api_token)
+    user.spawner.use_this_api_token = api_token
```
willingc: `# set user's spawner (the mock spawner in the test) to use the newly generated token which is not in the db

jupyterhub/tests/test_spawner.py
```diff
@@ -270,3 +273,90 @@ def stop():
...
+    # Spawner's provided API token is inserted into the db
+    # This shouldn't happen unless there are bugs elsewhere (in the Spawner),
```
willingc: `# Under normal circumstances, the spawner's provided API token would not be inserted into the db unless there is a bug somewhere else (in the Spawner)`

jupyterhub/tests/test_spawner.py
```diff
@@ -270,3 +273,90 @@ def stop():
...
+    yield user.stop()
+    yield user.spawn()
+    # verify re-use of API token
+    assert spawner.api_token == api_token
+    # verify that a new token was not created
+    assert user.api_tokens == [found]
+
+
+@pytest.mark.gen_test
+def test_spawner_insert_api_token(db, app):
```
willingc: @minrk this test was confusing to me

Perhaps add a docstring
`"""Token provided by spawner is not in the db. Insert token into db as a user-provided token."""`

jupyterhub/tests/test_spawner.py
minrk: Thanks! should be better, now.

minrk:
Thanks for the review. I think it should be clearer (and simpler) now.

PR #1350, opened 2017-08-17
allow failures on python: nightly

minrk:
since they break stuff sometimes.


willingc:
Thanks for doing this. I was going to do the same thing :-)

PR #1352, opened 2017-08-18
build jupyterhub/singleuser on this repo

minrk:
rather than dockerspawner.

The tags are tied to JupyterHub version, so it makes a bit more sense to build here
rather than there.

Plus, recent updates to dockerspawner eliminate any assumptions in dockerspawner about what the image does other than the fact that it launches `jupyterhub-singleuser`, and even that can be overridden with `c.Spawner.cmd`. So literally any image with JupyterHub installed should work.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,3 +4,7 @@ jupyterhub_cookie_secret
 jupyterhub.sqlite
 jupyterhub_config.py
 node_modules
+docs
+.git
+dist
+build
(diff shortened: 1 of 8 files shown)


minrk:
The Dockerfile here takes its two inputs as build args, so anyone can:

```bash
docker build -t singleuser \
      --build-arg BASE_IMAGE=jupyter/scipy-notebook \
      --build-arg JUPYTERHUB_VERSION=0.8.0 \
      .
```

to get any docker image based on just about any other docker image, as long as it has `python3`

Issue #1351, opened 2017-08-17
What URLs need to be exposed for the singleuser instances


DeepHorizons:
Hey all. I'm running into a problem with communication between JupyterHub and the singleuser instances. I know the singleuser instance needs access to the hub API, but it seems it cannot communicate with it.

I am trying to run JupyterHub behind a reverse proxy (Apache2) with authentication (Shibboleth). The spawner, however, spawns the singleuser instances on different servers than the hub and needs to talk to it over a public address. If I run an unauthenticated version of the hub, everything works just fine. But if I run it behind the authenticated version the singleuser instance fails to communicate to the hub with the error `...Failed to connect to Hub API...`

The authenticated version allows all access to `/hub/api.*` and I have verified that the servers can access it with `curl` without authentication but I am still getting an error. I do see some requests to `/user/<username>/api` in the log on the singleuser instance, are there any other URLs that need to be excluded from authentication?

minrk:
They should only communicate on `/hub/api`

DeepHorizons:
After further evaluation, it seems it may be a problem with SSL/TLS. If I point the spawner to `http`, I see a 301 in the apache logs, which is expected. Apache is configured to redirect all http traffic to the https interface. But I never see anything after that. And if I point the spawner to `https`, nothing shows up either.

Maybe this is because it is failing the identification of the address? My cert is signed by someone who isn't in the default trusted root certificates. But I added the intermediate cert to the store. Will Jupyter use the system CA store? Or do I need to manually specify that.

minrk:
It's using `requests` right now, so I would expect it would find your system-installed certs.

minrk:
I opened #1353 to avoid hiding the original error when this happens. But if you reach into jupyterhub on the node and modify the same block to raise the original connection error instead, you might get a better hint at what's going wrong.

DeepHorizons:
It does seem to be a certificate issue.
```
requests.exceptions.SSLError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:645)
```

After doing some reading on `requests`, it seems it does not in fact use the system certs.

http://docs.python-requests.org/en/master/user/advanced/#ca-certificates

"By default, Requests bundles a set of root CAs that it trusts, sourced from the Mozilla trust store... From Requests version 2.4.0 onwards, Requests will attempt to use certificates from certifi if it is present on the system..."

[...]

PR #1353, opened 2017-08-20
log error when failing to connect to Hub

minrk:
for better diagnosis of issues like #1351


DeepHorizons:
This causes an exception for me. HubAuth doesn't seem to have a `log` object.

minrk:
@DeepHorizons indeed, that was a typo. Fixed in #1354.

PR #1356, opened 2017-08-21
rework spawn futures to fix races

minrk:
1. set _proxy_pending before first wait to ensure that there is never a gap between setting spawn flags
2. always call `finish_user_spawn` to reduce the number of finalization cases
3. wait for proxy to finish on the slow_spawn timeout, not just start, because we are only interested in the total duration for page responsiveness

cc @yuvipanda I think this should fix what you saw in your latest stress test

### DIFF
--- a/jupyterhub/apihandlers/base.py
+++ b/jupyterhub/apihandlers/base.py
@@ -111,9 +111,10 @@ class APIHandler(BaseHandler):
             for name, spawner in user.spawners.items():
                 if spawner.ready:
                     servers[name] = s = {'name': name}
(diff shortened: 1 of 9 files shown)


yuvipanda:
Thanks @minrk. Do we test this by simply doing a perf test against this?

minrk:
I think so. I'm not sure I could trigger it reliably in a simple test because I think it requires tornado itself to be under heavy load.

willingc:
Thanks @minrk. Nice debug log entries too.

PR #1359, opened 2017-08-23
Fix AttributeError on User class

### LINKED ISSUE #1358: orm.User.spawner attribute doesn't exist?
On master 58751067dbb3c4a0e9726f3e27456b3bfb502b84, I'm seeing the below error immediately after login:
```
      File "C:\Miniconda3\envs\jupyterhub\lib\site-packages\tornado\web.py", line 1511, in _execute
        result = yield result
      File "C:\Miniconda3\envs\jupyterhub\lib\site-packages\tornado\gen.py", line 1055, in run
        value = future.result()
      File "C:\Miniconda3\envs\jupyterhub\lib\site-packages\tornado\concurrent.py", line 238, in result
        raise_exc_info(self._exc_info)
      File "C:\Miniconda3\envs\jupyterhub\lib\site-packages\tornado\gen.py", line 1069, in run
        yielded = self.gen.send(value)
      File "c:\dev\src\jupyterhub\jupyterhub\handlers\login.py", line 87, in post
        if user.spawner:
      File "c:\dev\src\jupyterhub\jupyterhub\user.py", line 225, in __getattr__
        raise AttributeError(attr)
    AttributeError: spawner
```
On `user.User` the `spawner` attr is present:
https://github.com/jupyterhub/jupyterhub/blob/58751067dbb3c4a0e9726f3e27456b3bfb502b84/jupyterhub/user.py#L211-L218

..but apparently not the [`orm.User`](https://github.com/jupyterhub/jupyterhub/blob/58751067dbb3c4a0e9726f3e27456b3bfb502b84/jupyterhub/orm.py#L106-L171):
https://github.com/jupyterhub/jupyterhub/blob/58751067dbb3c4a0e9726f3e27456b3bfb502b84/jupyterhub/user.py#L220-L225

dhirschfeld:
Fixes #1358

### DIFF
--- a/jupyterhub/user.py
+++ b/jupyterhub/user.py
@@ -221,8 +221,7 @@ class User(HasTraits):
     def __getattr__(self, attr):
         if hasattr(self.orm_user, attr):
             return getattr(self.orm_user, attr)
-        else:
-            raise AttributeError(attr)
+        return object.__getattribute__(self, attr)
 
     def __setattr__(self, attr, value):
         if not attr.startswith('_') and self.orm_user and hasattr(self.orm_user, attr):


dhirschfeld:
Not sure if this is the *correct* fix, but in case it is I thought I'd put in a PR so you can just merge it in...

minrk:
Does this actually fix it? I wouldn't expect it to because `User` has a `.spawner` property, which means `__getattr__` shouldn't be called.

dhirschfeld:
It fixes it in the sense that I can proceed to the login page and the spawner starts my image. Have some other config? related issues after that, but nothing that seems related to my *fix*

dhirschfeld:
Closing as I can no longer reproduce the error.

PR #1362, opened 2017-08-24
Improve help comments for SSL key/certs configuration parameters

stuartcampbell:
The "Use with ssl_key" and "Use with ssl_cert" were just in the wrong sections.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -291,13 +291,13 @@ class JupyterHub(Application):
     ssl_key = Unicode('',
         help="""Path to SSL key file for the public facing interface of the proxy
 
-        Use with ssl_cert
+        Use with ssl_key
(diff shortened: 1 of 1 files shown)


minrk:
I think this is backwards. The comments are supposed to be saying that "ssl_key should be used with ssl_cert" and vice versa. With this PR, the help strings are saying that 'ssl_key should be used with ssl_key', which is not what we want.

We could propbably communicate that more clearly, though. Do you want to take a stab at updating these help strings to signal that when setting each one, you should also set the other?

stuartcampbell:
Ah I see now, yes I'll have a crack at it

stuartcampbell:
How's this ?

minrk:
Perfect, thank you!

PR #1368, opened 2017-08-27
Provide more detailed error message in case of version mismatch

### LINKED ISSUE #1367: If the singleuser server running is a Jupyter notebook server & doesn't have a version header, fail explicitly
Currently, if you have a JupyterHub 0.8 server but a Jupyterhub <0.8 singleuser server, it'll end up in a redirect loop. This is IMO unideal, since a lot of people will run into it. We should fail with an explicit error that has an useful error message.

minrk:
this is the most likely cause of redirect loops when using docker, so record the spawner version and check it when a redirect is detected.

In the event of a redirect and mismatch, fail with a message explaining the version mismatch and how to fix it.

The redirect threshold is also lower in the case of a version mismatch.

closes #1367

### DIFF
--- a/jupyterhub/_version.py
+++ b/jupyterhub/_version.py
@@ -35,8 +35,11 @@ def _check_version(hub_version, singleuser_version, log):
         else:
             # log warning-level for more significant mismatch, such as 0.8 vs 0.9, etc.
             log_method = log.warning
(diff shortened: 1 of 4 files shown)


yuvipanda:
I like this more than failing explicitly on version mismatch!

PR #1375, opened 2017-08-28
Prevent "extra" from being used before definition.

athornton:
Prevent "extra" from being used before definition.


minrk:
Thanks!

PR #1380, opened 2017-08-30
add —cull-users to cull_idle_servers

minrk:
allows deleting idle users in addition to servers for temp-user cases such as binder/tmpnb

cc @yuvipanda

### DIFF
--- a/examples/cull-idle/cull_idle_servers.py
+++ b/examples/cull-idle/cull_idle_servers.py
@@ -40,8 +40,11 @@ from tornado.options import define, options, parse_command_line
 
 
 @coroutine
-def cull_idle(url, api_token, timeout):
-    """cull idle single-user servers"""
+def cull_idle(url, api_token, timeout, cull_users=False):
+    """Shutdown idle single-user servers
+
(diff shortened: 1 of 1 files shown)


akhmerov:
Should there perhaps be separate timeouts for culling servers and culling users? It makes sense to set server timeout reasonably short (maybe 10 minutes) for limiting resource usage, while it isn't bad if users last for a few days.

minrk:
I'm not sure. The target use case is tmpnb-style deployments where users only exist as long as they have a running server. What would be a use case where users should be deleted eventually but not immediately?

akhmerov:
Probably you're right. I was considering a situation where an instance has relatively limited CPU/RAM resources, so it makes sense to make a tight timeout for stopping the servers. Having a separate timeout for wiping their state would in that situation avoid unpleasant experience when the user is distracted briefly, and loses all the work.

minrk:
Deleting the user from the database shouldn't have a substantially different effect from stopping the server in terms of lost work. It doesn't perform any additional actions such as deleting the user from the system or their home directory, just dropping the row in the jupyterhub db.

PR #1384, opened 2017-08-31
restore db access on Spawner

minrk:
Shouldn’t be strictly necessary, but doesn’t hurt

cc @JenCabral who found the bug via sudospawner


willingc:
Thanks @JenCabral for finding the bug :cake:

@minrk Thanks! I'll add @JenCabral to the contributors list too.

PR #1392, opened 2017-09-04
update docs to preferred method of writing to log file

minrk:
extra_log_files config is unreliable because it doesn't capture all output, and shouldn't be recommended except in advanced use cases.

Piping output is much more robust and reliable.

### DIFF
--- a/docs/source/reference/config-examples.md
+++ b/docs/source/reference/config-examples.md
@@ -49,9 +49,6 @@ c.JupyterHub.cookie_secret_file = pjoin(runtime_dir, 'cookie_secret')
 c.JupyterHub.db_url = pjoin(runtime_dir, 'jupyterhub.sqlite')
 # or `--db=/path/to/jupyterhub.sqlite` on the command-line
 
-# put the log file in /var/log
-c.JupyterHub.extra_log_file = '/var/log/jupyterhub.log'
-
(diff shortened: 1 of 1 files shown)


docs/source/reference/config-examples.md
```diff
@@ -79,7 +76,8 @@ export GITHUB_CLIENT_ID=github_id
 export GITHUB_CLIENT_SECRET=github_secret
 export OAUTH_CALLBACK_URL=https://example.com/hub/oauth_callback
 export CONFIGPROXY_AUTH_TOKEN=super-secret
-jupyterhub -f /etc/jupyterhub/jupyterhub_config.py
+# send log output to /var/log
+jupyterhub -f /etc/jupyterhub/jupyterhub_config.py &>> /var/log/jupyterhub.log
```
yuvipanda: Maybe add a comment about &>>? I had to go look that up, hadn't encountered that before :)

willingc: Perhaps:
`# append log output to log file in /var/log`

minrk: 👍 done

For background: `>>` appends stdout to a file (common), `&>>` is the same after merging stdout and stderr, which is important for us because Python logging happens on stderr.

willingc:
Thanks @minrk.

PR #1393, opened 2017-09-04
improve reporting of spawn failure

### LINKED ISSUE #1389: User not informed their server didn't start
With my current setup of Jupyterhub, sometimes the singleuser server will fail to start. One thing I have noticed however is that while the hub will see that the server did not start, the error will not be propagated to the user, and the user will wait the until the timeout period passes.

The two cases I am currently seeing is
1. The server fails before the spawners `start()` method returns. In this case the hub logs an error, raises an assertion exception in `start()`, and then just kinda does nothing.
2. The spawners`start()` returns, but the server fails while Jupyterhub waits for it. Here the only log is `Spawner failed to start`.

In both cases I can see the hub return a `500` for `/hub/user/<user>` but then it continues returning `200` after that.

My questions are:
1. Is there a specific exception [type] that should be raised in the `start()` method to tell the user there was a problem and they should try again?
2. Is there any specific reason the hub isn't telling the user the server didn't start during the http_timeout period?

minrk:
- add Spawner._spawn_future and wait for it on `/hub/user/:name`, so that if a spawner fails, the error will be raised on waiting pages
- render pending page if triggered spawn doesn't finish, rather than starting redirect loop to `/user/:name`
- point to user logs when spawner exits prematurely, since this usually means an error raised during `jupyterhub-singleuser` startup
- preserve exception from last failed spawn and prevent visits to `/user/:name` from triggering a new spawn if the last attempt failed. Users must visit `/hub/home` and click `Start My Server` if their last spawn failed (or spawn via API or direct visits to /spawn)
- update 'Start My Server' link to visit `/hub/spawn`, which clears the `_spawn_future` exception if there was one, instead of being a simple redirect to `/user/:name`, in order to distinguish between implicit and explicit spawn requests
- Include 'You can try restarting your server from the home page' link on failed spawn error pages.

closes #1389

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -488,7 +488,8 @@ class BaseHandler(RequestHandler):
             if status is not None:
                 toc = IOLoop.current().time()
                 self.statsd.timing('spawner.failure', (toc - tic) * 1000)
-                raise web.HTTPError(500, "Spawner failed to start [status=%s]" % status)
+                raise web.HTTPError(500, "Spawner failed to start [status=%s]. The logs for %s may contain details." % (
+                    status, spawner._log_name))
 
             if spawner._waiting_for_response:
                 # hit timeout waiting for response, but server's running.
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -376,6 +376,9 @@ class BaseHandler(RequestHandler):
 
     @gen.coroutine
     def spawn_single_user(self, user, server_name='', options=None):
+        # in case of error, include 'try again from /hub/home' message
+        self.extra_error_html = self.spawn_home_error
+
         user_server_name = user.name
         if self.allow_named_servers and not server_name:
             server_name = default_server_name(user)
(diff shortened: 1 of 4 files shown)


jupyterhub/handlers/base.py
```diff
@@ -479,7 +490,8 @@ def finish_user_spawn():
             if status is not None:
                 toc = IOLoop.current().time()
                 self.statsd.timing('spawner.failure', (toc - tic) * 1000)
-                raise web.HTTPError(500, "Spawner failed to start [status=%s]" % status)
+                raise web.HTTPError(500, "Spawner failed to start [status=%s]. The logs for %s may contain details." % (
```
willingc: @minrk a general question re: logs.

 if I am using multiple servers per user, will I be able to tell which one failed to be spawned?

minrk: I believe so, yes. When we `raise HTTPError`, this is logged as `500 POST /users/:name/server`. If one HTTP endpoint can prompt the creation of different servers, then this might not be enough.

PR #1399, opened 2017-09-06
add trailing slash on /user/name

minrk:
proxies may not route `/user/name` correctly, only `/user/name/...`, so make sure that `/user/name` is redirected to `/user/name/`

this manifests as a redirect loop between /user/name and /hub/user/name when a route exists but /user/name is still being routed to the Hub.

I think there's an edge-case bug in CHP that's causing this, but we can easily work around it on the Hub side in the meantime.

cc @yuvipanda this covers all instances of redirect loops seen in the current instance on datahub

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -786,6 +786,13 @@ class CSPReportHandler(BaseHandler):
         self.statsd.incr('csp_report')
 
 
+class AddSlashHandler(BaseHandler):
+    """Handler for adding trailing slash to URLs that need them"""
+    def get(self, *args):
+        src = urlparse(self.request.uri)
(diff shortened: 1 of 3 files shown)


minrk:
Also fixed in CHP 3.0: https://github.com/jupyterhub/configurable-http-proxy/pull/123

The edge case was triggered on `/user/name?redirects=1`, but not `/user/name` because of some mishandling of query parameters in CHP's routing

jupyterhub/handlers/base.py
```diff
@@ -786,6 +786,13 @@ def post(self):
         self.statsd.incr('csp_report')
 
 
+class AddSlashHandler(BaseHandler):
```
willingc: Great idea to make this a class.

PR #1408, opened 2017-09-11
Updated the reference flask service example to include token auth

DeepHorizons:
The example flask service was missing showing how to authenticate a user by token.
This adds that functionality along with a slight change on where to look for the authentication implementation.


minrk:
Thanks!

PR #1413, opened 2017-09-12
Allow non integral memory byte specifications

### LINKED ISSUE #1089: Fractional memory / CPU limits / guarantees fail

Set memory limit (or guarantee, or cpu limit / guarantee) to a non-integral spec:

```python
c.Spawner.mem_limit = "1.5G"
```

(In supported spawners) memory limit is set to 1.5 gigabytes of RAM

**What actually happens**

JupyterHub refuses to start, with:

```
[E 2017-04-18 05:39:02.270 JupyterHub app:1527]
      File "/usr/local/lib/python3.4/dist-packages/jupyterhub/app.py", line 1524, in launch_instance_async
        yield self.initialize(argv)
      File "/usr/local/lib/python3.4/dist-packages/jupyterhub/app.py", line 1315, in initialize
        yield self.init_spawners()
      File "/usr/local/lib/python3.4/dist-packages/jupyterhub/app.py", line 1084, in init_spawners
        self.users[orm_user.id] = user = User(orm_user, self.tornado_settings)
      File "/usr/local/lib/python3.4/dist-packages/jupyterhub/user.py", line 128, in __init__
        config=self.settings.get('config'),
      File "/usr/local/lib/python3.4/dist-packages/kubespawner/spawner.py", line 29, in __init__
        super().__init__(*args, **kwargs)
      File "/usr/local/lib/python3.4/dist-packages/jupyterhub/spawner.py", line 345, in __init__
        super(Spawner, self).__init__(**kwargs)
      File "/usr/local/lib/python3.4/dist-packages/traitlets/config/configurable.py", line 84, in __init__
        self.config = config
      File "/usr/local/lib/python3.4/dist-packages/traitlets/traitlets.py", line 585, in __set__
        self.set(obj, value)
      File "/usr/local/lib/python3.4/dist-packages/traitlets/traitlets.py", line 574, in set
        obj._notify_trait(self.name, old_value, new_value)
      File "/usr/local/lib/python3.4/dist-packages/traitlets/traitlets.py", line 1139, in _notify_trait
        type='change',
      File "/usr/local/lib/python3.4/dist-packages/traitlets/traitlets.py", line 1176, in notify_change
        c(change)
      File "/usr/local/lib/python3.4/dist-packages/traitlets/traitlets.py", line 819, in compatible_observer
        return func(self, change)
      File "/usr/local/lib/python3.4/dist-packages/traitlets/config/configurable.py", line 186, in _config_changed
        self._load_config(change.new, traits=traits, section_names=section_names)
      File "/usr/local/lib/python3.4/dist-packages/traitlets/config/configurable.py", line 153, in _load_config
        setattr(self, name, deepcopy(config_value))
      File "/usr/local/lib/python3.4/dist-packages/traitlets/traitlets.py", line 585, in __set__
        self.set(obj, value)
      File "/usr/local/lib/python3.4/dist-packages/traitlets/traitlets.py", line 559, in set
        new_value = self._validate(obj, value)
      File "/usr/local/lib/python3.4/dist-packages/traitlets/traitlets.py", line 591, in _validate
        value = self.validate(obj, value)
      File "/usr/local/lib/python3.4/dist-packages/jupyterhub/traitlets.py", line 71, in validate
        return int(num) * ByteSpecification.UNIT_SUFFIXES[suffix]
    ValueError: invalid literal for int() with base 10: '1.5'
```

0.72.

yuvipanda:
Fixes #1089

### DIFF
--- a/jupyterhub/traitlets.py
+++ b/jupyterhub/traitlets.py
@@ -4,7 +4,7 @@ Traitlets that are used in JupyterHub
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
-from traitlets import List, Unicode, Integer, TraitError
+from traitlets import List, Unicode, Float, TraitError
 
 
 class URLPrefix(Unicode):
@@ -33,7 +33,7 @@ class Command(List):
         return super().validate(obj, value)
 
 
-class ByteSpecification(Integer):
+class ByteSpecification(Float):
     """
     Allow easily specifying bytes in units of 1024 with suffixes
 
@@ -62,11 +62,14 @@ class ByteSpecification(Integer):
         If it has one of the suffixes, it is converted into the appropriate
         pure byte value.
         """
-        if isinstance(value, int):
-            return value
-        num = value[:-1]
+        if isinstance(value, float) or isinstance(value, int):
+            return float(value)
+        try:
+            num = float(value[:-1])
+        except ValueError:
+            raise TraitError('{val} is not a valid memory specification. Must be an float or a string with suffix K, M, G, T'.format(val=value))
         suffix = value[-1]
-        if not num.isdigit() and suffix not in ByteSpecification.UNIT_SUFFIXES:
-            raise TraitError('{val} is not a valid memory specification. Must be an int or a string with suffix K, M, G, T'.format(val=value))
+        if suffix not in ByteSpecification.UNIT_SUFFIXES:
+            raise TraitError('{val} is not a valid memory specification. Must be an float or a string with suffix K, M, G, T'.format(val=value))
         else:
-            return int(num) * ByteSpecification.UNIT_SUFFIXES[suffix]
+            return float(num) * ByteSpecification.UNIT_SUFFIXES[suffix]
--- a/jupyterhub/traitlets.py
+++ b/jupyterhub/traitlets.py
@@ -4,7 +4,7 @@ Traitlets that are used in JupyterHub
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
-from traitlets import List, Unicode, Float, TraitError
+from traitlets import List, Unicode, Integer, TraitError
 
 
 class URLPrefix(Unicode):
@@ -33,7 +33,7 @@ class Command(List):
         return super().validate(obj, value)
 
 
-class ByteSpecification(Float):
+class ByteSpecification(Integer):
     """
     Allow easily specifying bytes in units of 1024 with suffixes
 
@@ -48,7 +48,7 @@ class ByteSpecification(Float):
         'K': 1024,
         'M': 1024 * 1024,
         'G': 1024 * 1024 * 1024,
-        'T': 1024 * 1024 * 1024 * 1024
+        'T': 1024 * 1024 * 1024 * 1024,
     }
 
     # Default to allowing None as a value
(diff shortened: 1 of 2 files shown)


Carreau:
+1

yuvipanda:
On second thoughts, this is actually wrong. It should output an Integer at all times - just a version in bytes (since float bytes don't make any sense). Only the string parsing code needs to support floats.

jupyterhub/traitlets.py
```diff
@@ -62,11 +62,14 @@ def validate(self, obj, value):
         If it has one of the suffixes, it is converted into the appropriate
         pure byte value.
         """
-        if isinstance(value, int):
-            return value
-        num = value[:-1]
+        if isinstance(value, float) or isinstance(value, int):
+            return float(value)
+        try:
+            num = float(value[:-1])
```
willingc: The K8 docs often use `Mi` suffix instead of simply `M`. [ref](https://kubernetes.io/docs/tasks/configure-pod-container/assign-memory-resource/) More of an FYI than a need to necessarily change since the error handling below will look at the suffix.

willingc: Looking at the k8 docs is an integer expected?

> The memory resource is measured in bytes. You can express memory as a plain integer or a fixed-point integer with one of these suffixes: E, P, T, G, M, K, Ei, Pi, Ti, Gi, Mi, Ki. For example, the following represent approximately the same value:
128974848, 129e6, 129M , 123Mi

jupyterhub/traitlets.py
```diff
@@ -62,11 +62,14 @@ def validate(self, obj, value):
...
+        except ValueError:
+            raise TraitError('{val} is not a valid memory specification. Must be an float or a string with suffix K, M, G, T'.format(val=value))
         suffix = value[-1]
-        if not num.isdigit() and suffix not in ByteSpecification.UNIT_SUFFIXES:
-            raise TraitError('{val} is not a valid memory specification. Must be an int or a string with suffix K, M, G, T'.format(val=value))
+        if suffix not in ByteSpecification.UNIT_SUFFIXES:
+            raise TraitError('{val} is not a valid memory specification. Must be an float or a string with suffix K, M, G, T'.format(val=value))
         else:
-            return int(num) * ByteSpecification.UNIT_SUFFIXES[suffix]
+            return float(num) * ByteSpecification.UNIT_SUFFIXES[suffix]
```
minrk: Shouldn't this be `int(float(num) * ByteSpec...)` to cast it back to int after scaling?

jupyterhub/traitlets.py
```diff
@@ -33,7 +33,7 @@ def validate(self, obj, value):
         return super().validate(obj, value)
 
 
-class ByteSpecification(Integer):
+class ByteSpecification(Float):
```
minrk: If it should be cast back to int after multiplying, I think this can go back to Integer.

jupyterhub/traitlets.py
```diff
@@ -62,11 +62,14 @@ def validate(self, obj, value):
         If it has one of the suffixes, it is converted into the appropriate
         pure byte value.
         """
-        if isinstance(value, int):
-            return value
-        num = value[:-1]
+        if isinstance(value, float) or isinstance(value, int):
```
minrk:
```python
if isinstance(value, (int, float)):
    return int(value)
```

jupyterhub/traitlets.py
minrk: Since this isn't strictly for kubernetes, we have some freedom in what we choose. Docker, for instance, allows non-integer with a multiplier, e.g. `1.5G`.

Ultimately, though, we are always providing an integer, e.g. `1.5G` should produce the integer `15e8`. So if Kubernetes requires an int, we will still be giving it one.

yuvipanda:
Thank you for fixing this up, @minrk!

PR #1416, opened 2017-09-14
avoid error if another traitlets Application is initialized

minrk:
encountered when doing db debugging in IPython

### DIFF
--- a/jupyterhub/alembic/env.py
+++ b/jupyterhub/alembic/env.py
@@ -12,9 +12,16 @@ config = context.config
 # Interpret the config file for Python logging.
 # This line sets up loggers basically.
 if 'jupyterhub' in sys.modules:
+    from traitlets.config import MultipleInstanceError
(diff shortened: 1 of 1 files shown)


jupyterhub/alembic/env.py
```diff
@@ -12,9 +12,16 @@
...
 if 'jupyterhub' in sys.modules:
+    from traitlets.config import MultipleInstanceError
     from jupyterhub.app import JupyterHub
+    app = None
     if JupyterHub.initialized():
-        app = JupyterHub.instance()
+        try:
+            app = JupyterHub.instance()
+        except MultipleInstanceError:
+            # could have been another Application
```
willingc: It's not entirely clear to me what other Application had an instance. Another JupyterHub running for a different purpose or group of users?

minrk: Where I ran into this, the Application was IPython.

PR #1418, opened 2017-09-14
Fixes (and tests!) for oauth state handling

minrk:
oauth state is required for redirecting properly, but wasn't being set.

Includes tests that verify that oauth state is handled properly.

Fixes https://github.com/data-8/nbgitpuller/issues/16


yuvipanda:
Awesome!

PR #1424, opened 2017-09-15
wrap control panel button in a span

### LINKED ISSUE #1423: Control Panel button is slightly bigger than Logout button in notebook and terminal view
Login as admin and open a notebook or terminal window

All buttons in header-container have the same height.

**What actually happens**
Control Panel button is 4 pixels higher than the Logout button.

image

```
jupyterhub==0.7.2
notebook==5.1.0
```

phill84:
make sure the same style is applied to all buttons in header-container

fix #1423


minrk:
Thanks!

PR #1439, opened 2017-09-21
avoid oauth state cookie collisions

### LINKED ISSUE #1438: Redirect to login url on mismatched oauth state
A few of our users have reported seeing the error "oauth state does not match" from auth.py#L782. One reported that the error was cleared when "erased all the caches" in her browser. Should the hub redirect to the login URL instead of displaying the error?

helm chart version v0.5.0-4f7cb71.

minrk:
in case of multiple simultaneous

- state arg is strictly required now
- default cookie name in case of no collision is unchanged
- in case of collision, randomize cookie name with a suffix and store
cookie_name in state
- expire state cookies after 10 minutes, not 1 day, in case oauth doesn't complete

closes #1438

### DIFF
--- a/jupyterhub/services/auth.py
+++ b/jupyterhub/services/auth.py
@@ -13,8 +13,10 @@ authenticate with the Hub.
 import base64
 import json
 import os
+import random
 import re
 import socket
+import string
 import time
 from urllib.parse import quote, urlencode
 import uuid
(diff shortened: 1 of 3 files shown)


jupyterhub/services/auth.py
```diff
@@ -531,22 +533,37 @@ def set_state_cookie(self, handler, next_url=None):
...
+        b64_state = self.generate_state(next_url, **extra_state)
         kwargs = {
             'path': self.base_url,
             'httponly': True,
-            'expires_days': 1,
+            # Expire oauth state cookie in ten minutes.
+            # Usually this will be cleared by completed login
+            # in less than a few seconds.
+            # OAuth that doesn't complete shouldn't linger too long.
+            'max_age': 600,
```
willingc: @minrk Could this be shorter - 5 min?

jupyterhub/services/auth.py
```diff
@@ -769,18 +797,19 @@ def get(self):
...
+        # clear cookie state now that we've consumed it
+        self.clear_cookie(cookie_name, path=self.hub_auth.base_url)
+        if isinstance(cookie_state, bytes):
+            cookie_state = cookie_state.decode('ascii', 'replace')
+        # check that state matches
+        if arg_state != cookie_state:
+            app_log.warning("oauth state %r != %r", arg_state, cookie_state)
+            raise HTTPError(403, "oauth state does not match. Try logging in again.")
+        next_url = self.hub_auth.get_next_url(cookie_state)
         # TODO: make async (in a Thread?)
```
willingc: Does this TODO still apply?

jupyterhub/services/auth.py
minrk: Since this could include going through institutional OAuth that might involve looking up your Google password in your password manager or somewhere, I wanted to give it a while for that case.

jupyterhub/services/auth.py
minrk: It does. It's a low priority task that we can do at any time.

jupyterhub/services/auth.py
willingc: Makes sense. Thanks @minrk.

PR #1441, opened 2017-09-21
debug intermittent failure on Travis

minrk:
<del>still figuring this one out</del> got it

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: python
 sudo: false
+cache:
+  - pip
 python:
   - nightly
   - 3.6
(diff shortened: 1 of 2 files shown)


minrk:
@willingc this is the test failure we've seen recently. Turns out it was caused by releasing CHP 3.0, which *fixed* a bug on routing `/user/name?params`. The test was accidentally relying on that buggy behavior and assuming that `/user/name` would be handled by the Hub itself, which it shouldn't have been because the route was setup in the test. Shutting down the server to ensure that `/user/name` is handled by the Hub fixes the issue.

.travis.yml
```diff
@@ -1,5 +1,7 @@
 language: python
 sudo: false
+cache:
+  - pip
```
minrk: Just for efficiency

willingc:
Great!

PR #1442, opened 2017-09-22
[WIP] Added additional statsd collection for the spawner

DeepHorizons:
The spawner was not reporting some statistics that would be useful for debugging and load testing.

* `spawner.failure` counter collects the number of failures for various reasons:
    * `timeout`: When the spawner times out dues to `start_timeout` being exceeded
    * `error`: Any uncaptured errors related to starting the user server
    * `http_timeout`: When the spawner times out dues to `http_timeout` being exceeded
    * `http_error`: Any uncaptured errors related to waiting for the user server to say it's ready
* `spawner.stop` timer for seeing how long it takes a user server to stop

Can anyone else think of any other statistics that should be recorded?

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -545,6 +545,7 @@ class BaseHandler(RequestHandler):
                 spawner._stop_pending = False
             toc = IOLoop.current().time()
             self.log.info("User %s server took %.3f seconds to stop", user.name, toc - tic)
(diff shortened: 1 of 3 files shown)


minrk:
This looks like a solid addition, thanks! If you think of others, feel free to open more PRs. I think @yuvipanda is hoping to migrate to something like prometheus in a future release.

PR #1460, opened 2017-09-29
Enable the asyncio event loop to run with tornado

DeepHorizons:
This allowed me to use a library that used asyncio (asyncssh) to work with tornado. I'm not aware of any negative effects. This is one of two ways to use the asyncio event loop, maybe we would rather use the asyncio event loop instead as such:

```
     def launch_instance(cls, argv=None):
         self = cls.instance()
         AsyncIOMainLoop().install()
         loop = asyncio.get_event_loop()
         loop.add_callback(self.launch_instance_async, argv)
         try:
             loop.run_forever()
         except KeyboardInterrupt:
             print("\nInterrupted")
```

Any thoughts?

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1629,6 +1629,7 @@ class JupyterHub(Application):
     @classmethod
     def launch_instance(cls, argv=None):
         self = cls.instance()
+        IOLoop.configure('tornado.platform.asyncio.AsyncIOLoop')
         loop = IOLoop.current()
         loop.add_callback(self.launch_instance_async, argv)
         try:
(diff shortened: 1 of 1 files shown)


minrk:
👍 to this. It will even be the default behavior in Tornado 5. Despite this being *almost* certainly fine, since JupyterHub 0.8 rc is already out and the release ready to go in the next day or two, I'd like to hold this until after we ship 0.8.0 this week.

DeepHorizons:
Perfectly reasonable, can't wait for the release!

As a side note, after talking with a tornado maintainer, I believe the `AsyncIOMainLoop().install()` is way to go.

yuvipanda:
Awesome. Does this also change the minimum required version of python?

DeepHorizons:
Nope. This should work on Python 3.4 as it has asyncio, although I have not yet tested it personally.

yuvipanda:
@minrk this means we're now running on the asyncio loop by default, right?

PR #1473, opened 2017-10-10
Change username input type to "text" to fix auto-filling

franga2000:
According to the [HTML spec](https://www.w3.org/TR/html5/forms.html#attr-input-type), `type="username"` isn't a valid type for an `<input>` element.

Because of this, Firefox doesn't recognize it when saving and auto-filling credentials - only the password field is saved.

I have changed it to `type="text"`, which fixes the issue.


minrk:
Thanks!

PR #1474, opened 2017-10-11
Replace forward slashes in CILogon oauth id for cookie

zonca:
Fixes https://github.com/jupyterhub/oauthenticator/issues/132

I mostly implemented this to check it really worked...

If you have a suggestion on a more general way to fix this, I can improve this patch.

### DIFF
--- a/jupyterhub/services/auth.py
+++ b/jupyterhub/services/auth.py
@@ -402,7 +402,7 @@ class HubOAuth(HubAuth):
         because we don't want to use the same cookie name
         across OAuth clients.
         """
-        return self.oauth_client_id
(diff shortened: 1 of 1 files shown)


minrk:
I think `/` shouldn't have been allowed in the username at all. These usernames should have failed at `.validate_username`.

https://github.com/jupyterhub/oauthenticator/issues/133 is open to address CILogon problems with the new OAuth 2 setup. I think we are using the wrong field for username in CILogon right now. Email is probably the best choice, but another option is possible, if you have an idea.

minrk:
#1475 disallows `/` explicitly

zonca:
ok very good!

PR #1493, opened 2017-10-19
Rendering logout page when auto login is true

schon:
Rendering logout page When `authenticator.auto_login=True`


andrelu:
Question: When auto_login is enabled should jupyterhub automatically redirect to login page?

PR #1497, opened 2017-10-21
[doc] Updated the rest docs to about change in 0.8

DeepHorizons:
In 0.8, the jupyterhub api token can also be used to make requests to
the jupyter notebook given some conditions. This PR updates that
documentation

### DIFF
--- a/docs/source/reference/rest.md
+++ b/docs/source/reference/rest.md
@@ -114,10 +114,12 @@ r.raise_for_status()
 r.json()
 ```
 
-Note that the API token authorizes **JupyterHub** REST API requests. The same
-token does **not** authorize access to the [Jupyter Notebook REST API][]
-provided by notebook servers managed by JupyterHub. A different token is used
-to access the **Jupyter Notebook** API.
+The same API token can also authorize access to the [Jupyter Notebook REST API][]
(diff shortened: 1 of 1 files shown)


docs/source/reference/rest.md
```diff
@@ -114,10 +114,12 @@ r.raise_for_status()
...
 
-Note that the API token authorizes **JupyterHub** REST API requests. The same
-token does **not** authorize access to the [Jupyter Notebook REST API][]
-provided by notebook servers managed by JupyterHub. A different token is used
-to access the **Jupyter Notebook** API.
+The same API token can also authorize access to the [Jupyter Notebook REST API][]
+provided by notebook servers managed by JupyterHub.
+To do so, the following must be true:
+
+1. The token used is tied to an admin user or service
```
minrk: Thanks! There are two cases here:

1. the token is for the same user as the owner of the notebook
2. it's an admin token **and** admin_access is True

i.e. admin tokens can access any notebook if admin_access is on, but you cancan always access your own notebook with your own tokens.

DeepHorizons:
@minrk Added your comment, let me know if its good

minrk:
Excellent, thanks!

PR #1507, opened 2017-10-27
add `jupyterhub --upgrade-db` to trigger upgrade on launch

### LINKED ISSUE #1506: Do not make a backup of the db when doing jupyterhub upgrade-db if no upgrade was performed
I'd like to run jupyterhub upgrade-db unconditionally every time JupyterHub starts in z2jh. This lets us do cross version upgrades without making the user do manual, hard, reach-into-k8s-guts things.

Right now, it seems to create a backup copy of the db every time it is run, regardless of anything changing or not. It would be nice if this didn't happen, and instead it only made a backup copy whenever it actually changed things.

minrk:
Upgrades the database (if needed) on start.

This is opt-in, for uses like the helm chart where explicit 'upgrade-db' steps are hard to insert.

This ought to be safe for sqlite users, where an automatic backup file is created *if an upgrade will occur*.

closes #1506

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -12,7 +12,6 @@ import logging
 from operator import itemgetter
 import os
 import re
-import shutil
 import signal
 import sys
 from textwrap import dedent
@@ -23,7 +22,6 @@ if sys.version_info[:2] < (3, 3):
 
 from jinja2 import Environment, FileSystemLoader
 
-from sqlalchemy import create_engine
 from sqlalchemy.exc import OperationalError
 
 from tornado.httpclient import AsyncHTTPClient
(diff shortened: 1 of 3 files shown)


jupyterhub/app.py
```diff
@@ -101,6 +99,13 @@
     'no-db': ({'JupyterHub': {'db_url': 'sqlite:///:memory:'}},
         "disable persisting state database to disk"
     ),
+    'upgrade-db': ({'JupyterHub': {'upgrade_db': True}},
+        """Automatically upgrade the database if needed on startup.
+
+        Only safe if the database has been backed up.
+        sqlite database files will be backed up automatically.
```
willingc: Perhaps add: Only SQLite...

jupyterhub/app.py
```diff
@@ -636,6 +613,12 @@ def _db_url_changed(self, change):
         """
     ).tag(config=True)
 
+    upgrade_db = Bool(False,
+        help="""Upgrade the database automatically on start.
+
+        Only safe if database is regularly backed up.
+        sqlite databases will be backed up to a local file automatically.
```
willingc: Only SQLite ...

minrk:
Thanks! I'll try to investigate the timeouts later this week.

yuvipanda:
Thanks, @minrk / @willingc! What do you think of a 0.8.1 with this to make the helm chart work easier? :)

minrk:
0.8.1 sounds good. I want to investigate the recent test timeouts to see if there's anything to fix there, but we should be able to ship that this week.

PR #1525, opened 2017-11-07
avoid waiting forever for threads to exit

minrk:
- call loop.stop via `.add_callback` <- this is the real fix. Calling `loop.stop` from another thread doesn't guarantee that the loop will wake
- add 30s timeout to `thread.join()` to avoid hanging forever if thread fails to exit
- assert that threads exit

I believe this fixes the recent test timeouts on Travis


minrk:
Thanks! It seems to have done.

willingc:
I think it fixed the timeouts @minrk. Though we may have an external proxy test failure (though intermittent) now.

PR #1531, opened 2017-11-10
"Start All" button on admin page

cmoscardi:
A companion to the "Stop All" button on the admin page. 

Our use case: an instructor uses this to turn on everyone's resources in advance of a workshop, so when a student logs in, their environment is set up, provisioned, and ready to go (this is especially important in our k8s case where we want to make sure to allocate via autoscaling, seed some material on to everyone's disks, etc.).

A couple points:
- The current implementation clicks all visible "Start" buttons (mirroring the "Stop All" implementation)
- It waits 500ms before clicking the next one, to mitigate server load. Should this be a configurable variable?*
- Visual feedback is provided as each "Start" button is clicked.
- HTML button layout: I left things janky so it's easier to play with, in case you guys have a better layout in mind (I'm not in love with the vertical stacking... maybe if I vertical-centered the buttons...)

*I guess I'd template it in to the HTML and have the JS pick it up or something?

As discussed with @yuvipanda

### DIFF
--- a/share/jupyter/hub/templates/admin.html
+++ b/share/jupyter/hub/templates/admin.html
@@ -33,7 +33,10 @@
       <tr class="user-row add-user-row">
         <td colspan="12">
           <a id="add-users" role="button" class="col-xs-2 btn btn-default">Add Users</a>
-          <a id="stop-all-servers" role="button" class="col-xs-2 col-xs-offset-5 btn btn-danger">Stop All</a>
+          <span class="col-xs-offset-5 col-xs-2">
+            <a id="start-all-servers" role="button" class="btn btn-success" style="width: 100%; margin-bottom: 10px;">Start All</a>
+            <a id="stop-all-servers" role="button" class="btn btn-danger" style="width: 100%;">Stop All</a>
+          </span>
           <a id="shutdown-hub" role="button" class="col-xs-2 col-xs-offset-1 btn btn-danger">Shutdown Hub</a>
         </td>
       </tr>
@@ -74,6 +77,10 @@
 
 {% call modal('Stop All Servers', btn_label='Stop All', btn_class='btn-danger stop-all-button') %}
   Are you sure you want to stop all your users' servers? Kernels will be shutdown and unsaved data may be lost.
+{% endcall %}
+
+{% call modal('Start All Servers', btn_label='Start All', btn_class='btn-success start-all-button') %}
+  Are you sure you want to start all servers? This can slam your server resources.
 {% endcall %}
 
 {% call modal('Shutdown Hub', btn_label='Shutdown', btn_class='btn-danger shutdown-button') %}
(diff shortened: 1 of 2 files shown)


minrk:
I think this ought to be a reasonable starting point. It would be a good deal more complicated to implement proper fast-start, back-off to launch everything without it falling over. Plus, if you are going to click the 'start all' button, that probably means you don't have that many users. How many users do you typically have when you try this, and how long does it take to launch them all in practice, when the autoscaler has to kick in?

What I would be worried about is spawns starting to time out if the autoscaler takes to long to scale up.

As for UI, the admin page is extremely minimal, since it's intended that the vast majority of admin is done via APIs, rather than this page. That said, we are presumably going to get to the point where we need a menu instead of just some huge buttons. Maybe now is that time?

share/jupyter/hub/templates/admin.html
```diff
@@ -76,6 +79,10 @@
   Are you sure you want to stop all your users' servers? Kernels will be shutdown and unsaved data may be lost.
 {% endcall %}
 
+{% call modal('Start All Servers', btn_label='Start All', btn_class='btn-success start-all-button') %}
+  Are you sure you want to start all servers? This can slam your server resources.
```
willingc: ~~slam~~ suggestions: burden or overload your available server resources.

willingc:
I'm wondering if it would be better to have a small companion app that uses the API to execute this.

cmoscardi:
@minrk 
- I've tried it with 20 users on a toy cluster with very small GCE nodes ... it took about 25 minutes total, including autoscaling from 2 to 10 nodes. We also have a large singleuser image (it takes ~10 minutes to spawn the first singleuser instance on a fresh node), so this may be faster with a more reasonably sized image.
- The kubespawner timeout is indeed an issue! I dealt with it by changing the timeout to 30 minutes. I'm not sure how this holds up in the case with a large number (say >50) of concurrent futures hanging around, but it worked fine with n=20. As an aside, https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/269 just made that timeout explicitly configurable in the helm chart - I had been doing it via the extraConfig option (thanks @yuvipanda ).
- A menu makes sense to me.

@willingc 
I'm not sure how common of a use case ours is (or will be), but having this all-in-one place would definitely be nice for us. Of course, this button could pretty much be a browser extension.

share/jupyter/hub/templates/admin.html
cmoscardi: Definitely not my best copy writing :)

cmoscardi:
Just tried with 50+ users autoscaling from 2 to 7 nodes and experienced better results -- about 15 minutes. I suspect it's linear with each node that gets added, or something like this. In any event, as long as a more sizeable kubespawner timeout is set, this seems to work fine.

Menu: May not actually be necessary just yet. I just tweaked it to get the flat button spacing right - I think that's all I was missing before...

voila

willingc:
Thanks @cmoscardi.

PR #1532, opened 2017-11-13
setup.py: run npm with --unsafe-perm

minrk:
this is the equivalent to bower's `--allow-root` that we had in bower-proper

fixes issues when installing jupyterhub from source with sudo or in Docker


cmoscardi:
Oops sorry, I'd forgotten to mention this in my PR! I just ran into it when developing the start all button.

This was also my fix, but I hadn't included it in the PR in case it was undesirable on other platforms.

PR #1533, opened 2017-11-13
If spawner is pending, always route to /hub/home/:user to resolve iss…

DerekHeldtWerle:
…ues with form options. Issue explanation: https://github.com/jupyterhub/jupyterhub/issues/1534

### DIFF
--- a/jupyterhub/handlers/pages.py
+++ b/jupyterhub/handlers/pages.py
@@ -67,6 +67,12 @@ class HomeHandler(BaseHandler):
         if user.running:
             # trigger poll_and_notify event in case of a server that died
             yield user.spawner.poll_and_notify()
+
+        # send user to /hub/user/:name if spawner is pending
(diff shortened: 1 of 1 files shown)


jupyterhub/handlers/pages.py
```diff
@@ -67,6 +67,12 @@ def get(self):
         if user.running:
             # trigger poll_and_notify event in case of a server that died
             yield user.spawner.poll_and_notify()
+
+        # send user to /hub/user/:name if spawner is pending
+        if user.spawner.pending:
+            url = url_path_join(self.hub.base_url, user.url)
+            self.redirect(url)
```
minrk: This should not be a redirect, but instead setting the url variable used immediately below. We shouldn't ever automatically redirect away from `/hub/home`, because that's how you get to the stop button, admin page, etc.

Issue #1087, opened 2017-04-17
Jupyterhub cross origin requests


y2kbowen:
I have read the documentation but still having an issue with jupyterhub cross origin requests.  I am running jupyter 4.3.0

I am making a request from my web application which when running under test has the  origin http://localhost:5000.  I am making the call from the client code to log into jupyterhub with the call:

```
POST http://10.7.135.5/hub/login HTTP/1.1
Content-Type: application/x-www-form-urlencoded
Accept: application/json, text/plain, */*
Referer: http://localhost:5000/home
Accept-Language: en-US
Origin: http://localhost:5000
Accept-Encoding: gzip, deflate
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.79 Safari/537.36 Edge/14.14393
Content-Length: 23
Host: 10.7.135.5
Connection: Keep-Alive
Pragma: no-cache

username=kenb&password=
```

This will return a cors exception on my browser. So I read this [issue (issue 79)](https://github.com/jupyter/jupyter/issues/79) and updated the c.NotebookApp.allow_origin = '*' in my jupyter_notebook_config.py.  

I am still having the same problem. I run JupyterHub in a container and I was attempting to make the change without rebuilding the image by logging in and updating configuration through a bash shell.  

So, what reads this config file? Does JupyterHub read this file in addition to its own config? Here are the processes running in my container:

```
/opt/conda/bin/python /opt/conda/bin/jupyterhub --ip= --config=config/jupyterhub/jupyterhub_config.py

/opt/conda/bin/configurable-http-proxy --ip  --port 8754 --api-ip 0.0.0.0 --api-port 8755 --default-target http://127.0.0.1:8081 --error-
target http://127.0.0.1:8081/hub/error

/opt/conda/bin/python /opt/conda/bin/jupyterhub-singleuser --user=kenb --port=56840 --cookie-name=jupyter-hub-token-kenb --base-url=/user/kenb
 --hub-host= --hub-prefix=/hub/ --hub-api-url=http://127.0.0.1:8081/hub/api --ip=0.0.0.0 --notebook-dir=notebooks --NotebookApp.default_url=/lab --debug --disable-user-config
```

[...]

PR #1539, opened 2017-11-16
fix headers for CORS

minrk:
- add `authorization` to default `Access-Control-Allow-Headers`
- allow overriding `Access-Control-Allow-Headers` just like everything else
- ensure case-insensitive comparison for proper header checks

With this fix, config to enable token-authenticated cross-origin requests on both the Hub itself and running notebooks

```python
origin = 'http://localhost:9999'

c.Spawner.args = [f'--NotebookApp.allow_origin={origin}']
c.JupyterHub.tornado_settings = {
    'headers': {
        'Access-Control-Allow-Origin': origin,
    },
}
```

And [an example](https://gist.github.com/minrk/924ef40af06cae06116c1c7b60b963e3) of a page that does cross-origin access of both the notebook server and Hub, for testing.

cf #1087

### DIFF
--- a/jupyterhub/apihandlers/base.py
+++ b/jupyterhub/apihandlers/base.py
@@ -18,14 +18,14 @@ class APIHandler(BaseHandler):
         return '; '.join([super().content_security_policy, "default-src 'none'"])
 
     def set_default_headers(self):
-        self.set_header('Content-Type', 'application/json')
         super().set_default_headers()
+        self.set_header('Content-Type', 'application/json')
 
     def check_referer(self):
         """Check Origin for cross-site API requests.
-        
+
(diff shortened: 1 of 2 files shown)


jupyterhub/apihandlers/base.py
```diff
@@ -18,14 +18,14 @@ def content_security_policy(self):
         return '; '.join([super().content_security_policy, "default-src 'none'"])
 
     def set_default_headers(self):
-        self.set_header('Content-Type', 'application/json')
         super().set_default_headers()
+        self.set_header('Content-Type', 'application/json')
```
willingc: +1

Issue #1477, opened 2017-10-12
Secure cookies not working behind AWS ALB


psyvision:
Hello,

I'm running with the following version:

JupyterHub 0.8
Notebook 5.1.0

In a custom Docker container in AWS Container Service.

Jupyterhub is configured with SSL and in AWS it runs behind an AWS application load balancer using SSL.

If I send a request to login direct to the server running the Docker container (e.g. https://10.2.0.10:33212/jupyterhub/hub/login) then I get a "jupyter-hub-token" cookie back marked as Secure:

```
set-cookie:jupyter-hub-token="2|1:0|10:1507813141|17:jupyter-hub-token|44:......"; expires=Sat, 11 Nov 2017 12:59:01 GMT; HttpOnly; Path=/jupyterhub/hub/; Secure
```

However, if I navigate to the same page via the AWS application load balancer's DNS (e.g. https://jupyterhub.internal/jupyterhub/hub/login) then I do not get a secure cookie:

```
set-cookie:jupyter-hub-token="2|1:0|10:1507812660|17:jupyter-hub-token|44:......"; expires=Sat, 11 Nov 2017 12:51:00 GMT; HttpOnly; Path=/jupyterhub/hub/
```

I've added some extra logging and observe the following, this happens with a direct request:

```
[I 2017-10-12 12:58:31.058 JupyterHub jwtauthenticator:15] Protocol: https
[I 2017-10-12 12:58:31.058 JupyterHub jwtauthenticator:16] X-Forwarded-Proto: https
```

The following is observed when going via the AWS ALB:

```
[I 2017-10-12 12:50:59.977 JupyterHub jwtauthenticator:15] Protocol: http
[I 2017-10-12 12:50:59.977 JupyterHub jwtauthenticator:16] X-Forwarded-Proto: https,https
```

The code for logging these is:
```
self.log.info("Protocol: %s", self.request.protocol)
self.log.info("X-Forwarded-Proto: %s", self.request.headers.get("X-Forwarded-Proto", ""))
```

The Secure flag is added to the cookie if `self.request.protocol=="https"`.

The AWS ALB will only accept requests on https/443 and the Docker container will also only accept requests via port 443.  I note that in `app.py` the tornado HTTPServer is started with `xheaders` set to `True` so it should work, as per http://www.tornadoweb.org/en/stable/httpserver.html#http-server:

```
[...]

PR #1540, opened 2017-11-16
allow override of cookie options

minrk:
via `tornado_settings['cookie_options']`

for cases where default options are incorrect or insufficient (e.g. expiry or https detection fails)

@psyvision this is me punting on #1477, giving you an escape hatch for when the default cookie options are wrong, rather than coming up with a more robust detection for https.

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -265,6 +265,8 @@ class BaseHandler(RequestHandler):
             kwargs['secure'] = True
         if self.subdomain_host:
             kwargs['domain'] = self.domain
+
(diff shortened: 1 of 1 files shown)


psyvision:
Very much appreciated, thank you!

psyvision:
@minrk what setting will I need to set (presumably in my jupyterhub_config.py) to set this?

Carreau:
IFAICT you should set `c.JupyterHub.tornado_settings = <a dict>`. with  `<a dict>` being `{'cookie_options':{<what you need>}}` where `<what you need>="X-Forwarded-Proto":...` but I haven't followed exactly the previous issue.

PR #1555, opened 2017-11-21
base dockerfile on alpine linux

ankitml:


minrk:
Thanks! I think this is off to the right start. I think we just want the one alpine dockerfile, though. It should set the CMD of the image and probably run the Hub itself, possibly as `nobody` or another non-root user. Users can make a derivative image that does:

1. install authenticator/spawner packages
2. add jupyterhub_config.py

and they should be set. An example of such a Dockerfile would be great in the examples directory.

dockerfiles/Dockerfile.base
```diff
@@ -0,0 +1,5 @@
+FROM python:3.6.3-alpine3.6
```
minrk: Let's call this Dockerfile.alpine. And merge this with Dockerfile.config. They don't need to be separate files.

ankitml:
Have updated the PR according to the changes suggested here. However, I need to test it again as my local docker-compose like setup breaks with this change.

ankitml:
@minrk So, in the dockerfile which extends from this alpine base image, lets call it example 1 of how to use the base image. I am not able to install anything, this is because of user being nobody. So basically anyone who extends this image will have to switch user back to root and then back to nobody for any pip installation or apk add. Is this an acceptable pattern in extending docker images?

ankitml:
@minrk I was able to test its working on a minikube cluster, looks like it can be merged. Do you suggest any changes?

dockerfiles/Dockerfile.alpine
```diff
@@ -0,0 +1,8 @@
+FROM python:3.6.3-alpine3.6
+RUN pip3 install --no-cache jupyterhub==0.8.0
+ENV LANG=en_US.UTF-8
+
+USER nobody
+ENTRYPOINT ["jupyterhub"]
```
minrk: This should be CMD

dockerfiles/Dockerfile.alpine
```diff
@@ -0,0 +1,8 @@
+FROM python:3.6.3-alpine3.6
+RUN pip3 install --no-cache jupyterhub==0.8.0
```
minrk: Can you add above here:

```
ARG JUPYTERHUB_VERSION=0.8.1
```

then

```
RUN pip3 install --no-cache jupyterhub==${JUPYTERHUB_VERSION}
```

holms:
Would be really grateful if this could be merged :)

ankitml:
Thanks for the suggestions @minrk . Have made the changes. However I havent tested them, they look benign to me.

minrk:
Thanks!

PR #1577, opened 2017-12-07
add session-id cookie

### LINKED ISSUE #1491: User logout does not invalidate cookies for services
It seems that if you log in as a user that has access to the server, and then logout, this does not invalidate the cookie or token for that service and so you can still gain access to it. This is an annoyance with testing, and potentially a security issue (if a user was using JupyterHub on a public machine).

1. Create a `jupyterhub_config.py` with a service, e.g.

```
c = get_config()

c.Authenticator.whitelist = [
    'instructor1'
]

c.JupyterHub.load_groups = {
    'formgrader-course101': [
        'instructor1'
    ]
}

c.JupyterHub.services = [
    {
        'name': 'course101',
        'url': 'http://127.0.0.1:9999',
        'command': [
            'jupyterhub-singleuser',
            '--group=formgrader-course101',
            '--debug',
        ],
        'user': 'grader-course101',
        'cwd': '/home/grader-course101'
    }
]
```

2. Go to the service url, e.g. `/services/course101`. Observe that you get redirected to the login page (good!).

3. Now log in as the user, and go to the service url. Observe you can access it.

4. Go back to the user's server, and click "log out". Go back to the service url and observe that you can still access it.

After the user logs out, they should no longer have access to the service.

**What actually happens**

After the user logs out, they still have access to the service.

0.8.0

minrk:
session cookie is stored cleartext, domain-wide. It contains nothing but a UUID. All services can read the session ID. The session ID serves two purposes, which are related in purpose, but mostly unrelated technically (i.e. it's fine to use one without the other):

1. Hub-side, OAuth tokens are associated with a session id. When a user logs out with the Hub, all tokens with that session id are revoked, and the session id is cleared. This closes #1491.
2. service-side, session id, if present, is added to the auth cache key. If session id changes, even if the auth cookie doesn't, the Hub will be asked again. This should result in the above revocation taking immediate effect, rather than waiting for cache expiry.

If I have done this right, this should be implemented in a backward-compatible way in both directions. If single-user servers are v0.8.0 and do not have session-id cache keys, the only consequence is that revoked tokens take cache_max_age to expire (same as now). Similarly, if newer single-user servers are used with older Hubs, session id will always be empty and revoked tokens will take cache_max_age to expire.

A caveat: Right now, it is not possible for external services (i.e. those not on `/services/`) to use OAuth with the Hub. This is a mostly artificial limitation at this point (there is currently no way to register an oauth client with a URL outside the Hub's domain). However, the service-side part of this logic *cannot* be used by servers on other domains without third-party cookie shenanigans that I think we shouldn't attempt. They will always behave the same as now, where tokens take cache_max_age to expire and cannot be revoked immediately.

Closes #1491

### DIFF
--- a/jupyterhub/alembic/env.py
+++ b/jupyterhub/alembic/env.py
@@ -30,10 +30,7 @@ if 'jupyterhub' in sys.modules:
 else:
     fileConfig(config.config_file_name)
 
-# add your model's MetaData object here
-# for 'autogenerate' support
-# from myapp import mymodel
-# target_metadata = mymodel.Base.metadata
+# add your model's MetaData object here for 'autogenerate' support
 from jupyterhub import orm
 target_metadata = orm.Base.metadata
 
--- a/jupyterhub/alembic/env.py
+++ b/jupyterhub/alembic/env.py
@@ -34,7 +34,8 @@ else:
 # for 'autogenerate' support
 # from myapp import mymodel
 # target_metadata = mymodel.Base.metadata
-target_metadata = None
+from jupyterhub import orm
+target_metadata = orm.Base.metadata
 
 # other values from the config, defined by the needs of env.py,
 # can be acquired:
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -264,7 +264,7 @@ class BaseHandler(RequestHandler):
         session_id = self.get_session_cookie()
         if session_id:
             # clear session id
-            self.clear_cookie(SESSION_COOKIE_NAME)
+            self.clear_cookie(SESSION_COOKIE_NAME, **kwargs)
 
             if user:
                 # user is logged in, clear any tokens associated with the current session
@@ -285,7 +285,8 @@ class BaseHandler(RequestHandler):
 
         # clear hub cookie
         self.clear_cookie(self.hub.cookie_name, path=self.hub.base_url, **kwargs)
-        self.clear_cookie('jupyterhub-services', path=url_path_join(self.base_url, 'services'))
+        # clear services cookie
+        self.clear_cookie('jupyterhub-services', path=url_path_join(self.base_url, 'services'), **kwargs)
 
     def _set_cookie(self, key, value, encrypted=True, **overrides):
         """Setting any cookie should go through here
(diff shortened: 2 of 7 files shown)


jupyterhub/handlers/base.py
```diff
@@ -255,10 +260,39 @@ def clear_login_cookie(self, name=None):
         kwargs = {}
         if self.subdomain_host:
             kwargs['domain'] = self.domain
+        user = self.get_current_user_cookie()
+        session_id = self.get_session_cookie()
+        if session_id:
+            # clear session id
```
willingc: maybe: `# clear cookie storing session id`

jupyterhub/alembic/env.py
```diff
@@ -34,7 +34,8 @@
 # for 'autogenerate' support
 # from myapp import mymodel
 # target_metadata = mymodel.Base.metadata
-target_metadata = None
+from jupyterhub import orm
```
willingc: Do we still want to support the comments in line 33-36? If so, I would combine lines 33 & 34 for greater clarity.

jupyterhub/services/auth.py
```diff
@@ -493,7 +517,7 @@ def token_for_code(self, code):
 
     def _encode_state(self, state):
         """Encode a state dict as url-safe base64"""
-        # trim trailing `=` because 
+        # trim trailing `=` because = is itself not url-safe!
```
willingc: 👍

jupyterhub/tests/test_services_auth.py
```diff
@@ -371,7 +372,7 @@ def test_oauth_service(app, mockservice_url):
 @pytest.mark.gen_test
 def test_oauth_cookie_collision(app, mockservice_url):
     service = mockservice_url
-    url = url_path_join(public_url(app, mockservice_url) + 'owhoami/')
+    url = url_path_join(public_url(app, mockservice_url), 'owhoami/')
```
willingc: Good catch.

jupyterhub/alembic/env.py
minrk: Done.

PR #1578, opened 2017-12-07
remove , which was turning timestamp into tuple

johnkpark:
I noticed that the browser was caching static files for Jupyterhub, and when I opened the chrome inspector I noticed that the timestamp that's supposed to prevent caching was getting set to something else `(%`.

screen shot 2017-12-07 at 11 19 18 am

It turns out that there was an extra `,` which was turning the timestamp into a tuple. With the fix:

screen shot 2017-12-07 at 11 19 40 am

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1308,7 +1308,7 @@ class JupyterHub(Application):
         if os.path.isdir(os.path.join(parent, '.git')):
             version_hash = ''
         else:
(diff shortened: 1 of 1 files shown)


willingc:
Thanks for the catch @johnkpark and the PR to fix. This looks like a [stray comma from 3 years ago](https://github.com/johnkpark/jupyterhub/blame/b21b0427d1a9e40095df3a22063747325ed5f170/jupyterhub/app.py#L1260). Thanks!

PR #1581, opened 2017-12-10
Add RED prometheus metrics for all requests

yuvipanda:
This patch introduces Prometheus for exposing metrics
about JupyterHub's operation. We expose a standard /metrics
endpoint that can be queried without authentication. We
take on prometheus_client as an unconditional dependency
to both simplify code & because it is a pure python package
with no dependencies itself.

The first pass adds 'RED' style metrics for all HTTP requests.
http://rancher.com/red-method-for-prometheus-3-key-metrics-for-monitoring/
has some info on the RED method, but to summarize:

  For each request type, record at least the following metrics

   Rate – the number of requests, per second, your services are serving.
   Errors – the number of failed requests per second.
   Duration – The amount of time each request takes expressed as a time interval.

This instantly gives us a lot of useful metrics in a very
compact form.

Also adds following metrics:
- Spawn times
- Proxy addition times

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -5,6 +5,7 @@
 
 import copy
 import re
+import time
 from datetime import timedelta
 from http.client import responses
 from urllib.parse import urlparse, urlunparse, parse_qs, urlencode
@@ -22,6 +23,7 @@ from .. import orm
 from ..objects import Server
 from ..spawner import LocalProcessSpawner
 from ..utils import url_path_join
+from ..metrics import SPAWN_DURATION_SECONDS
 
 # pattern for the authentication token header
 auth_header_pat = re.compile(r'^(?:token|bearer)\s+([^\s]+)$', flags=re.IGNORECASE)
@@ -388,6 +390,7 @@ class BaseHandler(RequestHandler):
     @gen.coroutine
     def spawn_single_user(self, user, server_name='', options=None):
         # in case of error, include 'try again from /hub/home' message
+        spawn_starttime = time.perf_counter()
         self.extra_error_html = self.spawn_home_error
 
         user_server_name = user.name
@@ -397,6 +400,11 @@ class BaseHandler(RequestHandler):
 
         if server_name in user.spawners and user.spawners[server_name].pending:
             pending = user.spawners[server_name].pending
+            SPAWN_DURATION_SECONDS.labels(
+                status='already-pending'
+            ).observe(
+                time.perf_counter() - spawn_starttime
+            )
             raise RuntimeError("%s pending %s" % (user_server_name, pending))
 
         # count active servers and pending spawns
@@ -415,6 +423,11 @@ class BaseHandler(RequestHandler):
                 '%s pending spawns, throttling',
                 spawn_pending_count,
             )
+            SPAWN_DURATION_SECONDS.labels(
+                status='throttled'
+            ).observe(
+                time.perf_counter() - spawn_starttime
+            )
             raise web.HTTPError(
                 429,
                 "User startup rate limit exceeded. Try again in a few minutes.",
@@ -423,6 +436,11 @@ class BaseHandler(RequestHandler):
             self.log.info(
                 '%s servers active, no space available',
                 active_count,
+            )
+            SPAWN_DURATION_SECONDS.labels(
+                status='too-many-users'
+            ).observe(
+                time.perf_counter() - spawn_starttime
             )
             raise web.HTTPError(429, "Active user limit exceeded. Try again in a few minutes.")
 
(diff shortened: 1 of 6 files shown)


yuvipanda:
I've left the statsd metrics as is for now, but we should probably remove them at some point. They aren't very exhaustive and are also of dubious accuracy I think.

jupyterhub/handlers/base.py
```diff
@@ -388,6 +390,7 @@ def active_server_limit(self):
     @gen.coroutine
     def spawn_single_user(self, user, server_name='', options=None):
         # in case of error, include 'try again from /hub/home' message
+        spawn_starttime = time.perf_counter()
```
willingc: Let's rename to `spawn_start_time` since it's less error prone than the double `tt`

jupyterhub/handlers/base.py
```diff
@@ -22,6 +23,7 @@
 from ..objects import Server
 from ..spawner import LocalProcessSpawner
 from ..utils import url_path_join
+from ..metrics import SPAWN_DURATION_SECONDS, PROXY_ADD_DURATION_SECONDS
```
willingc: Do you think `ADD_PROXY_DURATION_SECONDS` is clearer to indicate that it's how long adding a proxy takes?

jupyterhub/handlers/base.py
```diff
@@ -456,13 +474,30 @@ def finish_user_spawn():
...
+            )
+            proxy_add_starttime = time.perf_counter()
             spawner._proxy_pending = True
             try:
                 yield self.proxy.add_user(user, server_name)
+
+                PROXY_ADD_DURATION_SECONDS.labels(
+                    status='success'
+                ).observe(
+                    time.perf_counter() - proxy_add_starttime
```
willingc: `add_proxy_start_time`

jupyterhub/log.py
```diff
@@ -120,3 +121,4 @@ def log_request(handler):
         if location:
             ns['location'] = ' → {}'.format(location)
     log_method(msg.format(**ns))
+    prometheus_log_method(handler)
```
willingc: Add to docstring: `- log prometheus metrics`

jupyterhub/metrics.py
```diff
@@ -0,0 +1,41 @@
...
+
+PROXY_ADD_DURATION_SECONDS = Histogram(
+    'proxy_add_duration_seconds',
+    'duration for adding user routes to proxy',
+    ['status']
+)
+
+def prometheus_log_method(handler):
+    """
+    Tornado log handler for recording RED metrics
```
willingc: Is this more accurate?  Log RED metrics as a bit more information to Tornado's default log

jupyterhub/metrics.py
```diff
@@ -0,0 +1,41 @@
...
+
+    We record the following metrics:
+       Rate – the number of requests, per second, your services are serving.
+       Errors – the number of failed requests per second.
+       Duration – The amount of time each request takes expressed as a time interval.
+
+    We use a fully qualified name of the handler as a label,
+    rather than every url path to reduce cardinality.
```
willingc: Excellent point re: cardinality 👍

requirements.txt
```diff
@@ -6,3 +6,4 @@ pamela
 python-oauth2>=1.0
 SQLAlchemy>=1.1
 requests
+prometheus_client
```
willingc: `>=0.0.21`

jupyterhub/handlers/base.py
yuvipanda: I think general prometheus guidelines are to do things like PROXY_ADD_DURATION_SECONDS and PROXY_REMOVE_DURATION_SECONDS, so they sort together. I can change it if you feel strongly, but would like to keep it as is!

jupyterhub/handlers/base.py
yuvipanda: Done!

jupyterhub/handlers/base.py
yuvipanda: Done!

jupyterhub/log.py
yuvipanda: Done!

jupyterhub/metrics.py
yuvipanda: It doesn't add to the tornado default log, rather it is just a convenient hook function that lets us read timing & request info. I've added a couple of lines here explaining how this is used!

jupyterhub/metrics.py
yuvipanda: Ty!

requirements.txt
yuvipanda: Done!

yuvipanda:
Thank you for the review, @willingc! I've updated and pushed new commits to account for most of 'em!

jupyterhub/handlers/base.py
willingc: Let's leave as is but clarify in a doc string or comment since I initially interpreted as adding seconds vs. adding user to the proxy.

jupyterhub/metrics.py
```diff
@@ -1,5 +1,8 @@
 """
 Prometheus metrics exported by JupyterHub
+
+Read https://prometheus.io/docs/practices/naming/ for naming
```
willingc: :heart:

jupyterhub/handlers/base.py
yuvipanda: I added more docs on metrics.py, and renamed the spawn metric so it now follows a '<noun>_<verb>_<type-suffix>' pattern. Great catch on confusion!

minrk:
Great! Let's add a deprecation warning when statsd is enabled for 0.9, but otherwise leave them alone.

Issue #1584, opened 2017-12-11
config c.LocalProcessSpawner.mem_limit Issue


prabhu1984:
Hi,

As per http://jupyterhub.readthedocs.io/en/latest/api/spawner.html instructions, I've set config c.LocalProcessSpawner.mem_limit to 15GB as below on jupyterhub_config.py.

```
c.LocalProcessSpawner.mem_limit = 15728640
c.LocalProcessSpawner.mem_guarantee = 1048576
c.LocalProcessSpawner.cpu_limit = 2.0
c.LocalProcessSpawner.cpu_guarantee = 0.5
```

But, when I tried to test by loading some test file to cross the 15gb boundary, process didn't fail. Actually, I am able to load some 48GB data into memory and stopped the test. 

![Image](https://github.com/prabhu1984/sparkmagic/blob/master/examples/images/Testing_MEM_CPU_Limit.png?raw=true)

Please let me know, what I'm doing wrong. Why is memory limit not applied?

Thanks!

minrk:
LocalProcessSpawner doesn't implement mem_limit because it needs to be able to create cgroups and that isn't implemented. You will have to use a Spawner that supports these, such as DockerSpawner or KubeSpawner. For non-containerized servers, I believe SystemdSpawner implements mem_limit.

willingc:
Thanks @minrk for the explanation. 

Documentation recommendation: Let's highlight (bold) the sentence in the docs re: spawner must support... http://jupyterhub.readthedocs.io/en/latest/reference/spawners.html#spawners-resource-limits-and-guarantees-optional

Siecje:
In case this might help someone I've implemented a crude limit with this code. https://github.com/jupyterhub/sudospawner/pull/47

ixjlyons:
In addition to @willingc's comment above, I think the spawner docstrings should mention whether or not these features are supported. For example, the [`LocalProcessSpawner` documentation](https://jupyterhub.readthedocs.io/en/latest/api/spawner.html#localprocessspawner) says "This needs to be supported by your spawner for it to work." This seems to be because it doesn't override the `mem_limit` trait so the generic string comes from the base `Spawner` class. You'd have to dig into the source code to determine whether or no the `mem_limit` option is implemented. If overriding a trait attribute just to modify the documentation is an antipattern, maybe it could just be mentioned in the top-level portion of the docstring.

prabhu1984:
Hi @minrk/@willingc, 

[...]

PR #1588, opened 2017-12-18
Clarify resource limits/guarantees in docs and docstrings

willingc:
This PR clarifies that LocalProcessSpawner, the default spawner, does not enforce these limits. A custom spawner must implement the code to enforce the limits.

Closes: #1584


minrk:
Thanks!

PR #1590, opened 2017-12-19
enable external oauth clients

minrk:
Built on #1577

TODO:

- [x] improve tracking of token activity (this gets increasingly important as we expand scope of how tokens can be used)
- [x] alembic upgrade
- [x] support defining oauth clients in config
- [x] allow oauth to come from external (specified) hosts

Once complete, this will allow JupyterHub to serve as a 'whoami' OAuth provider. OAuth tokens do not grant access to perform actions on behalf of users.

Logical extensions not in this PR:

- implement oauth scopes (requires scope-approval page), allowing OAuth tokens to take action. This will probably mean merging OAuth tokens with the existing APIToken. They are only separate now because OAuth tokens do not grant API access.
- allow registering new OAuth clients via UI, not just config.


willingc:
@minrk when this pr lands, will nbgrader then work with master and 0.8.1 (assuming backward compatibility_?

jupyterhub/apihandlers/auth.py
```diff
@@ -49,7 +53,7 @@ def post(self):
                 raise web.HTTPError(403)
         else:
             data = self.get_json_body()
-            # admin users can request 
+            # admin users can request tokens for other usrs
```
willingc: typo users

minrk:
This shouldn't affect backward compatibility, so nothing should change for nbgrader between 0.8.x and this. It's purely a new feature for allowing oauth clients to have redirect URIs other than `/services/:name/oauth_callback`.

yuvipanda:
w00t, this is awesome!

yuvipanda:
This is awesome! This should also allow us do authenticated binderhubs I think?

minrk:
> This should also allow us do authenticated binderhubs I think?

It's a big part of it, yes. However, binder deployment would also need updates to make the nullauthenticator bit optional.

jupyterhub/alembic/versions/56cc5a70207e_token_tracking.py
```diff
@@ -0,0 +1,42 @@
...
+branch_labels = None
+depends_on = None
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    tables = op.get_bind().engine.table_names()
+    # ### commands auto generated by Alembic - please adjust! ###
```
willingc: Under what circumstances would this be adjusted? Different data fields stored for tokens?

jupyterhub/app.py
```diff
@@ -1145,7 +1145,8 @@ def init_services(self):
             if service.managed:
                 if not service.api_token:
                     # generate new token
-                    service.api_token = service.orm.new_api_token()
+                    # TODO: revoke old tokens?
```
willingc: Are there use cases where a managed service would need to keep an old api_token?

jupyterhub/apihandlers/auth.py
```diff
@@ -31,33 +33,43 @@ def get(self, token):
...
         else:
             data = self.get_json_body()
-            # admin users can request 
-            if data and data.get('username') != user.name:
-                if user.admin:
-                    user = self.find_user(data['username'])
-                    if user is None:
-                        raise web.HTTPError(400, "No such user '%s'" % data['username'])
-                else:
+            # admin users can request tokens for other users
```
willingc: can an admin user request a token for another admin user? I assume that is okay.

jupyterhub/app.py
```diff
@@ -1266,6 +1269,26 @@ def init_oauth(self):
             login_url=url_path_join(base_url, 'login')
         )
 
+    def cleanup_oauth_clients(self):
+        """Cleanup any OAuth clients that shouldn't be in the database.
+
+        This should mainly be services that have been removed from configuration or renamed.
```
willingc: +1

jupyterhub/orm.py
```diff
@@ -319,6 +321,11 @@ def service_id(cls):
     hashed = Column(Unicode(255), unique=True)
     prefix = Column(Unicode(16), index=True)
 
+    # token metadata for bookkeeping
```
willingc: +1

yuvipanda:
Not necessarily in this PR, but would be also great to have an example here!

jupyterhub/app.py
minrk: Not one generated here. I think the main bit missing is that we don't track whether this happened, so we don't have enough info to clean it up afterward.

jupyterhub/apihandlers/auth.py
minrk: Yes, any other user.

jupyterhub/alembic/versions/56cc5a70207e_token_tracking.py
minrk: This is an auto-generated comment from `alembic revision --autogenerate`. I should have removed it after making sure it was correct.

minrk:
external oauth example added, review addressed. Should be good to go.

examples/external-oauth/README.md
```diff
@@ -0,0 +1,74 @@
+# Using JupyterHub as an OAuth provider
+
+JupyterHub 0.9 introduces
+Uses `jupyterhub.services.HubAuthenticated` to authenticate requests with the Hub.
```
willingc: ~~Uses~~ using

examples/external-oauth/README.md
```diff
@@ -0,0 +1,74 @@
...
+1. generate an API token:
+
+        export JUPYTERHUB_API_TOKEN=`openssl rand -hex 32`
+
+2. launch the whoami service:
+
+        bash launch-service.sh &
+
+3. Launch JupyterHub:
+
```
willingc: command??

examples/external-oauth/README.md
```diff
@@ -0,0 +1,74 @@
+# Using JupyterHub as an OAuth provider
+
+JupyterHub 0.9 introduces
+Uses `jupyterhub.services.HubAuthenticated` to authenticate requests with the Hub.
+
+There is an implementation each of cookie-based `HubAuthenticated` and OAuth-based `HubOAuthenticated`.
+
+## Run
```
willingc: Running JupyterHub as an OAuth provider

PR #1597, opened 2017-12-27
Template messages

rschroll:
This builds on top of #1596, pulling out small bits of text as "messages" that a user may wish to override.  With these in place, all the user has to provide in the custom template directory are the relevant message templates.

This is done in the two places I was most interested in
1. On the home page (default empty)
2. On the spawn pending page (default to the existing text)

There are likely other places where this would make sense.  If you point the out, I'd be glad to set those up too.

After toying with a couple of other ideas, I ended up with the macro approach.  This seemed to give the most flexibility with the smallest amount of boilerplate.  Another approach would be to define a `base_home.html`, for example, with the content currently in `home.html` with a block added where the message would go.  Then the new (or customized) `home.html` would extend `base_home.html` and define the message block there.  While this might be a little more in line with the usual use of templates, it seemed to me to require a bit more commitment to making those various base files right now.  In contrast, the macro approach lets us easily add messages as desired.  I'd be happy to show what the other approach would look like, should you prefer.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -265,7 +265,10 @@ class JupyterHub(Application):
     ).tag(config=True)
 
     template_paths = List(
-        help="Paths to search for jinja templates.",
+        help="""Paths to search for jinja templates.
+
+        Use the special string '%DEFAULT%' to include the default path in this list.
+        """
     ).tag(config=True)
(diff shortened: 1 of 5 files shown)


rschroll:

PR #1598, opened 2017-12-27
Add configurable spawn pending message

rschroll:
We wanted to be able to provide a custom message to the user when a spawner is pending.  This was a direct approach I took.  I think I like the approach in #1597 better, since it's more general, but I thought I'd throw this out as another approach.  If #1597 seems reasonable, please just go ahead and close this one.


rschroll:

Issue #1481, opened 2017-10-13
Make html templates more extensible


minrk:
There are a few things we can do to make custom templates a more pleasant experience

1. always include the default templates dir and data_files_dir in the template path, so that custom templates can extend them with `{% extend "templates/page.html" %}`
2. make sure that all of our blocks use `{{ super() }}` as appropriate. Right now, many blocks in `page.html` cannot be overridden in a useful way because `{{ super() }}` is not called on the pages.
3. look for potential additional places for blocks that would be useful

minrk:
#1601 begins implementing this

manics:
I think this is done (or is at least happening gradually).

PR #1601, opened 2017-12-29
Base template mechanism

rschroll:
There are several places where I'd like to make slight changes to template for a JupyterHub deploy.  By using the `template_paths` variable, I can provide my own templates, but I'd rather not need to copy the existing versions over.  This provides a mechanism to let the user extend the base templates.

When `base_templates` is `True`, user templates can extend `BASE:name.html` to extend the base version of that template.  (The `BASE:` part is only necessary if the user is also providing a template `name.html`.)  This means that the user can provide a message on the home page by defining their own `home.html`:
```html
{% extends "BASE:home.html" %}

{% block main %}
{{ super() }}
<div class="container">
  <div class="row">
    <p>Hello there!</p>
  </div>
</div>
{% endblock %}
```

Most pages seem to already have enough flexibility to let the user override or add to relevant parts.  The one page I saw that didn't do this well was `spawn_pending.html`, so I added a block around the message there.

This effectively provides the capabilities of #1596 and #1597, and I think it's also a cleaner way to do #1598.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1298,7 +1298,7 @@ class JupyterHub(Application):
         if base_path not in self.template_paths:
             self.template_paths.append(base_path)
         loader = ChoiceLoader([
-            PrefixLoader({'BASE': FileSystemLoader([base_path])}, ':'),
+            PrefixLoader({'templates': FileSystemLoader([base_path])}, '/'),
             FileSystemLoader(self.template_paths)
         ])
         jinja_env = Environment(
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -20,7 +20,7 @@ from urllib.parse import urlparse
 if sys.version_info[:2] < (3, 3):
     raise ValueError("Python < 3.3 not supported: %s" % sys.version)
 
-from jinja2 import Environment, FileSystemLoader
+from jinja2 import Environment, FileSystemLoader, PrefixLoader, ChoiceLoader
 
 from sqlalchemy.exc import OperationalError
 
@@ -271,6 +271,10 @@ class JupyterHub(Application):
     @default('template_paths')
     def _template_paths_default(self):
         return [os.path.join(self.data_files_path, 'templates')]
+
+    base_templates = Bool(False,
+        help="Allow tempates to extend the base versions, by referencing 'BASE:name.html'."
+    ).tag(config=True)
 
     confirm_no_ssl = Bool(False,
         help="""DEPRECATED: does nothing"""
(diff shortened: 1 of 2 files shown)


rschroll:
The first error seems to consistently be `test_api.py::test_shutdown`.  I'm not sure how I could have messed that up.

Let's see if I can rerun them.

rschroll:
Well, it looks like the same problem, but I don't know what's up.  It passes just fine on my local machine (once I disable `test_start_stop_race`, which fails on master and seems to lead to a cascade of errors).

willingc:
@rschroll Just an FYI, I'm seeing the test errors on a different PR that I just submitted. I'm investigating.  

cc/ @minrk

rschroll:
Thanks, @willingc!  I've rebased and things seem to be working now.

rschroll:
Is there any interest in this?  It's fine if there's not -- we'll just modify templates for our own deploys.  But I think this would reduce long term maintenance costs for us.

jupyterhub/app.py
```diff
@@ -272,6 +272,10 @@ class JupyterHub(Application):
     def _template_paths_default(self):
         return [os.path.join(self.data_files_path, 'templates')]
 
+    base_templates = Bool(False,
```
minrk: I think it's okay for the base templates to always be available. This doesn't need to be configurable.

willingc: This makes sense. Much like extending templates with Sphinx.

rschroll:
Thanks for the look-over.  I hadn't seen #1481; otherwise I would have followed that plan.  Yes, we could use "templates/" instead of "BASE:" as the prefix.  I purposefully chose something that would be hard to confuse with a path, but consistency with jupyter is a good reason to change.

The reason I put it in as an option was to support people who wanted to make sure all the templates were replaced by custom versions.  With this new search method, you'll always get something served; I thought error messages in the logs might be desirable in some cases.  But for my use, I'd like it always on, so if you don't see a need for the option, I'll take it out.

I won't have time in the next day or two, but I'll get a fixed version up soon.

rschroll:
Finally got a chance to get the updates in.  Pretty straight-forward, really.

I feel like this behavior should be documented somewhere, but I'm not sure where that should be.  Before, the base_templates option was the obvious place, but that's gone.  It could be added to the template_paths option, but that's slightly oblique.  Other suggestions?

minrk:
Thanks!

I think adding a new doc in the sphinx docs and/or some examples of custom templates is probably the way to go. If you want to give that a go in this PR that would be great, but I'm also happy to merge this and have that in a subsequent PR.

rschroll:
Let's get this in now.  But don't let me get lazy about the docs!

elgalu:
Hi all! thank you so much for investing in improving the template system. I'm running master branch and finally getting rid of boilerplate code in my fork :)

PR #1606, opened 2018-01-02
Add CHP to dev install and move to CONTRIBUTING

willingc:
After running across a bunch of failing tests (when I thought I had CHP installed), I've updated the dev instructions in README and moved the content to CONTRIBUTING.md so that it's easier to use.


minrk:
Great!

PR #1609, opened 2018-01-03
tornado 5 fixes

### LINKED ISSUE #1608: Support tornado 5
Tests are failing with tornado 5a1. This is likely due to tornado 5 running on asyncio by default. It's my suspicion that the issue lies only in the test suite and not the application, which I've been running locally on tornado master from time to time.

minrk:
Mostly minor cleanup fixes to deal with asyncio changes and avoid using private APIs.

Still one issue to figure out that has to do with raising HTTPError in set_default_headers, which used to be allowed but doesn't appear to be anymore.
  
closes #1608

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1625,10 +1625,7 @@ class JupyterHub(Application):
         if not self.io_loop:
             return
         if self.http_server:
-            if self.io_loop._running:
-                self.io_loop.add_callback(self.http_server.stop)
-            else:
-                self.http_server.stop()
+            self.http_server.stop()
         self.io_loop.add_callback(self.io_loop.stop)
 
     @gen.coroutine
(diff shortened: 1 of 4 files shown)


minrk:
The HTTPError issues I was seeing aren't tornado 5 related, they are related to upstream changes in notebook-master. Will fix here, anyway.

jupyterhub/app.py
```diff
@@ -1625,10 +1625,7 @@ def stop(self):
         if not self.io_loop:
             return
         if self.http_server:
-            if self.io_loop._running:
-                self.io_loop.add_callback(self.http_server.stop)
-            else:
-                self.http_server.stop()
+            self.http_server.stop()
```
minrk: The running check is no longer necessary. It was added when tests ran in a thread, so stop had to be passed via callback. This is no longer the case.

minrk:
Thanks, @willingc!

PR #1619, opened 2018-01-12
Update docker volumes link

timfreund:
The old link now returns 404.


willingc:
Thanks @timfreund for taking the time to submit a PR correcting these links :cake:

PR #1621, opened 2018-01-13
Fix spawner/service typo in proxy.py

consideRatio:
I don't fully grasp the code in this file, but this must be a typo.

@minrk could you confirm this?


minrk:
exactly right!

PR #1623, opened 2018-01-17
Fix Authentication state documentation

BerserkerTroll:
This — is how it actually works!

### DIFF
--- a/docs/source/reference/authenticators.md
+++ b/docs/source/reference/authenticators.md
@@ -145,7 +145,7 @@ If such state should be persisted, `.authenticate()` should return a dictionary 
 
 ```python
 {
-  'username': 'name',
+  'name': username,
   'auth_state': {
     'key': 'value',
   }


docs/source/reference/authenticators.md
```diff
@@ -145,7 +145,7 @@ If such state should be persisted, `.authenticate()` should return a dictionary
 
 ```python
 {
-  'username': 'name',
+  'name': username,
```
betatim: can you quote `username` again? I think in the rest of this example all values are quoted to make it a valid dict.

betatim:
LGTM except for that small nitpick on quoting.

docs/source/reference/authenticators.md
BerserkerTroll: Without quotes it is more clear that this is a variable and not a string literal.

willingc:
Thanks @betatim and @BerserkerTroll

Issue #1620, opened 2018-01-12
Accessing Notebook API through the JupyterHub API


MaximMoinat:
Similar to closed issues #1431 and #1078, but both (outdated?) solutions do not work for me.

**Background**
From a javascript application, that generates a piece of R script, I want to be able to create a Jupyter Notebook into a specific directory. Users can then open the file and execute it. This is possible via the Jupyter Notebook API, with a PUT request on `/api/content/<filename>.ipynb?token=xxx` and a body containing the notebook content.

**Problem**
I want to do the same via the JupyterHub API. And according to the the [REST API documentation](https://jupyterhub.readthedocs.io/en/stable/reference/rest.html), it should be possible to send a request to the Notebook API:

> The same API token can also authorize access to the Jupyter Notebook REST API provided by notebook servers managed by JupyterHub if one of the following is true: 
The token is for the same user as the owner of the notebook
The token is tied to an admin user or service and c.JupyterHub.admin_access is set to True

Can the JupyterHub API indeed be used for what I am trying to achieve? 
Or is there maybe a much easier solution?

Note:
I tried requests to `localhost:8081/user/<username>/api/content/<filename>.ipynb` and `localhost:8081/hub/user/<username>/api/content/<filename>.ipynb` already, but both give a redirect response referring to the other url.
What does work, is to look up the port of the spawned notebook and send a request there (e.g. `localhost:49581/user/<username>/api/content/<filename>.ipynb`. However, this is not practical as the port is generated randomly and can't be retrieved via the JupyterHub API.

sigurdurb:
There was a similar question I answered on StackOverflow 

[Maybe it can be of some assistance](https://stackoverflow.com/a/48224603/7267340)

MaximMoinat:
Thanks for your answer, I already tried without success:
`curl -X GET http://<DOMAIN>:<PORT>/user/<username>/api/contents/<filename>.ipnb -H 'Authorization: Token <token>' -H "Content-Type: application/json"`

Any other suggestions?

sigurdurb:
If there is nothing downloaded then you probably have to access the server first to start it. Like it said in my answer for /user notebooks the user has to start it by himself or admin has to go into the control panel.

willingc:
@MaximMoinat Let me see if I understand your desired use:

1. You have a code snippet that you wish to use to create a new Notebook.
2. You wish to have the newly created Notebook placed in a specific user's directory tree at a specific directory.
[...]

PR #1627, opened 2018-01-22
Fix unicode errors with extra_log_file

minrk:
Should fix unicode issues in #1406

<del>
reverts #1620 
</del>

<del>
Should only be merged if extra_log_file is verified to be the cause.
</del>

Undid the revert of #1625, since this can't have been the source of error messages produced by jupyterhub-singleuser. It's still a known unicode bug in extra_log_file, so we should do that part, at least.

cc @ellisonbg


minrk:
Removed the revert, this just fixes the known bug in extra_log_file

PR #1637, opened 2018-01-31
moved suggested configuration file to match canonical jupyter config paths

mpacer:
This modifies the docs to suggest a new location for configuration files. Specifically this will be found by any utilities using jupyter standard configuration paths to perform their functionality (this includes [`jupyter_conf_search`](https://github.com/mpacer/jupyter_conf_search)).

This still uses the standard Unix filesystem location, it just adheres to the recommendations described in: https://jupyter.readthedocs.io/en/latest/projects/jupyter-directories.html#configuration-files

specifically, `etc/jupyter/`:
<table border="0" class="table">
<colgroup>
<col width="52%">
<col width="48%">
</colgroup>
<thead valign="bottom">
<tr class="row-odd"><th class="head">Unix</th>
<th class="head">Windows</th>
</tr>
</thead>
<tbody valign="top">
<tr class="row-even"><td colspan="2"><span class="target" id="index-0"></span><a class="reference internal" href="#envvar-JUPYTER_CONFIG_DIR"><code class="xref std std-envvar docutils literal"><span class="pre">JUPYTER_CONFIG_DIR</span></code></a></td>
</tr>
<tr class="row-odd"><td colspan="2"><code class="docutils literal"><span class="pre">{sys.prefix}/etc/jupyter/</span></code></td>
</tr>
<tr class="row-even"><td><code class="docutils literal"><span class="pre">/usr/local/etc/jupyter/</span></code>
<code class="docutils literal"><span class="pre">/etc/jupyter/</span></code></td>
<td><code class="docutils literal"><span class="pre">%PROGRAMDATA%\jupyter\</span></code></td>
</tr>
</tbody>
</table>

This originally came to my attention while looking into https://github.com/jupyterlab/jupyterlab-latex/issues/49.

### DIFF
--- a/docs/source/getting-started/config-basics.md
+++ b/docs/source/getting-started/config-basics.md
@@ -25,7 +25,7 @@ jupyterhub --generate-config
 This default `jupyterhub_config.py` file contains comments and guidance for all
 configuration variables and their default values. We recommend storing
 configuration files in the standard UNIX filesystem location, i.e.
-`/etc/jupyterhub`.
+`/etc/jupyter/jupyterhub`.
 
 ## Start with a specific config file
 
@@ -36,11 +36,11 @@ jupyterhub -f /path/to/jupyterhub_config.py
 ```
 
 If you have stored your configuration file in the recommended UNIX filesystem
-location, `/etc/jupyterhub`, the following command will start JupyterHub using
+location, `/etc/jupyter/jupyterhub`, the following command will start JupyterHub using
 the configuration file:
 
 ```bash
-jupyterhub -f /etc/jupyterhub/jupyterhub_config.py
+jupyterhub -f /etc/jupyter/jupyterhub/jupyterhub_config.py
 ```
 
 The IPython documentation provides additional information on the


minrk:
I don't think I agree with this. I think `/etc/jupyterhub` is the right place, not putting everything in `/etc/jupyter/jupyterhub`. JupyterHub doesn't inherit Jupyter's path logic, as it's a very different kind of application from all other Jupyter applications. Specifically, it's not a user application with a hierarchy of system-wide and per-user configuration files, so I think it's appropriate for it to be a sibling of `jupyter` rather than a child.

willingc:
Makes sense @minrk. I'll make a PR to the general Jupyter docs to note that the JupyterHub config files typically reside in `/etc/jupyterhub`. 

@mpacer Anywhere else that we should mention this that would have helped you on your search the other day?

mpacer:
I see there being two separate problems here:  

1. How do _libraries_/_applications_ discover the conventions around JupyterHub config files?
2. How do _people_ discover the conventions around JupyterHub config files?

I address 1. more in my comments responding to @minrk and 2. more in my comments responding to @willingc. However, I would appreciate both of your thoughts in both areas.

### How should libraries/applications discover JupyterHub's config convention?

@minrk 
<details> 
<summary>Are there any technical downsides to including it there? I think there may be expository reasons to include it there (both expository and technical). (click for more details)
</summary>
<br/>

One thing I was worried about was that `jupyterhub` might then default to following these links. But, since It looks like `jupyterhub` is not based on `JupyterApp`, it's not looking in the standard jupyter file paths for these purposes (that is what I think you mean by 
> JupyterHub doesn't inherit Jupyter's path logic, as it's a very different kind of application from all other Jupyter applications. 

In addition to the doc changes that I describe below, I do think there are advantages to keeping it there for consistency with the rest of the Jupyter ecosystem.

Any tools that are working on standard jupyter filepaths will be easier to integrate with JupyterHub if it adheres to the same conventions. This includes `jupyter_conf_search` and any other tools that would be intended to work across all jupyter apps (e.g., a hypothetical jupyter configuration manager). 
</details>
<br/>
<details>
<summary> Can we encode JupyterHub's config convention in either the <code>jupyter</code> or <code>jupyterhub</code> packages so that it can be imported by other packages? Which of these do you think makes the most sense? (click for more details)
</summary>
<br/>

One of the hardest aspects of coördinating configuration between many applications is how hard it manually encode the conventions used by those different applications. This is even worse for conventions that allow for hierarchical inheritance, environment-specificity, and working directory relativity. Jupyter user applications have a config system that has all of those properties, but applications have few issues discovering and adhering to this convention. That is the power of encoding this convention in the `jupyter` package — any application that needs to adhere it can import it… whatever it happens to be in that version of `jupyter`, in that specific environment, in that specific directory. 

I have no qualms with JupyterHub using a different convention from Jupyter user applications. JupyterHub has different kinds of problems and different solutions to address those problems. It makes sense to separate those concerns. I do have qualms with the not being able to import JupyterHub's convention directly. 

I see two primary packages that could provide an importable version of JupyterHub's convention: `jupyter` and `jupyterhub`. Of these two solutions do you have thoughts as to which would be preferable? 

I have some preliminary thoughts on each solution:

- `jupyter`

   It would not make sense to use the same api in the `jupyter` package as is currently used to surface other paths. Instead it would be better to make a separate api. In that manner JupyterHub to adhere to a convention that isolates its configuration files from the rest of the Jupyter user application config files while still making it discoverable. There are a few advantages to this
   1. the same package could be used to find configuration for both JupyterHub and Jupyter user applications
   2. this is the location for the current importable convention
   3. `jupyter` is a small package compared to `jupyterhub` 
   3. this provides a pattern that can be generalised in case this issue arises for another Jupyter application

- `jupyterhub`
    
    On the other hand, it's not clear that we should rely on the `jupyter` package to communicate these conventions for other configuration systems. It seems like it could be a bit of a can-of-worms and introduce a cumbersome coupling of the `jupyter` package to other packages.  It also seems like it is conceptually cleaner to keep the configuration convention specific to JupyterHub inside the `jupyterhub` package. This would mean more dependencies for any libraries that wanted to work "across the whole Jupyter ecosystem", but I think the possibility of importing these conventions at all is more than worth having more dependencies.

</details>  
<br/>

### How should people discover JupyterHub's config convention?

@willingc 

>  I have a broader question of how many talks/tutorials might be out there that has the /etc/jupyterhub location and how this might impact troubleshooting/support.

I agree that that is an issue. Unfortunately, I'm not sure about all of where it would need to be changed…I do not know the full variety of places where all the JupyterHub resources live. I also don't know how to begin figuring that out exhaustively. It seems like that would be a good thing to know independent of the issues around this PR.

I have tried to write out clearly why I perceive this issue as being a lot thornier than can be solved with a note mentioning that JupyterHub has a different convention. I go into much greater detail below, but the challenge seems to arise from a conflict between a few seemingly reasonable assumptions:  
1. Jupyter's general ecosystem documentation should be as simple as possible.
2. All Jupyter applications follow Jupyter conventions (including path and location conventions).
3. JupyterHub is a Jupyter application.
4. > [JupyterHub is] a very different kind of application from all other Jupyter applications.

<details>
<summary> There are at least 3 pages where a change would make things less confusing: <a href="http://jupyter.readthedocs.io/en/latest/projects/jupyter-command.html"> <code>jupyter</code> command</a>, <a href="http://jupyter.readthedocs.io/en/latest/projects/config.html">Jupyter common config system</a>, & <a href="http://jupyter.readthedocs.io/en/latest/projects/jupyter-directories.html"> Jupyter directories </a>. 
 Each page poses a slightly different flavour of the problem (click for more details). </summary> 

*  http://jupyter.readthedocs.io/en/latest/projects/jupyter-command.html 
   <details>
   <summary> 
   The <code> jupyter </code> command docs imply a bunch of conventions for Jupyter applications, without listing JupyterHub as an exception. This is probably the easiest fix: I'd take @willingc's suggestion further and mention that `jupyterhub` is an exception in multiple ways (including its command line invocation).
   </summary>

   <br/>

   > Commands like `jupyter notebook` start Jupyter applications. The jupyter command is primarily a namespace for subcommands. A command like `jupyter-foo` found on your PATH will be available as a subcommand `jupyter foo`.
   > ⋮
   > `--config-dir`
   > Show the location of the config directory.
   > ⋮
   > --paths
   > Show all Jupyter directories and search paths.

    Because `jupyterhub` does not fit the `jupyter-foo` `jupyter foo` pattern, strictly speaking this does not say that JupyterHub should follow this pattern. However, it does suggest that all jupyter applications do follow that pattern, so the exception suggests that JupyterHub should not be thought of as a "Jupyter application" as described in this area.

   I am not sure how to mention that caveat without ruining the flow of the docs. That said, I think this page is salvagable even if we only say "`jupyterhub` is an exception to this command pattern" and then link to the JupyterHub docs.

   </details>

* http://jupyter.readthedocs.io/en/latest/projects/config.html
   <details>
   <summary> 
    The Common Jupyter configuration system docs directly state that "The Jupyter applications have a common config system, and a common config directory." Introducing a caveat makes the notion of a "common config system" much harder to reason about without explaining the reasoning behind the common config system in addition to the exception. This could result in improved (but less simple) docs.
   </summary>

   <br/>
   
   As I see it, the problem with this particular page is not just that we would need to include an exception in multiple locations in the docs themselves. The "take-away" from this page is "Jupyter applications have a common config system". Perhaps I am reading too much into it, but I take this to say that the common config system is an intentional design decision to treat configuration in a common manner across the Jupyter ecosystem.

   I don't see how to convey that jupyterlab is not to be included in this common design without explaining the idea that @minrk described here that `jupyterhub` is a fundamentally different _kind_ of application from other Jupyter applications. 

   This does make clearer how we could salvage this while making the docs for this page even stronger.

   @minrk's cleanly points out that 
   > Specifically, [`jupyterhub`] is not a user application with a hierarchy of system-wide and per-user configuration files…
   
   That holds the beginnings of an **excellent** explanation of why the common config system exists:  
   - some Jupyter applications are user applications  
       - Jupyter's user applications benefit from a hierarchy of system-wide and per-user configuration
       - the reason for this hierarchy is… (I won't go into the details here, as that is a separate conversation)
   - some Jupyter applications (notably, JupyterHub) are _not_ user applications
       - non-user applications are better used with an explicit configuration file passed in via command line arguments
       - it is better to keep these configuration files separate because… (I won't go into the details here, because I'm not sure yet (see my first comment to @minrk))

   However, this is a much larger change than just mentioning an exception and will require a lot more work and iteration to get right. Still, either way, it seems worth it.

   </details>

* http://jupyter.readthedocs.io/en/latest/projects/jupyter-directories.html
   details>
   <summary> 
   The Jupyter directories page lists the path conventions for configuration, runtime and data files. Jupyterhub is an exception to this pattern, but (I think) only in regards to its convention around configuration paths. This can probably be addressed by mentioning the exception, with more or less detail.
   </summary>
   </br>

   The directories page is already complex, and I do not think adding an exception to it would notably increase that complexity. All that said, I do have a few questions that would affect how to implement the change.

   Would this also benefit from the explanation described above for the "common config system" documentation? It already links to the common config system page; would it benefit from a comment saying that the reasoning behind the differing conventions is explained in the linked page? 

   Is there a way to use JupyterHub on a windows machine? Or do we only need to consider the Unix case?

   If there is a way to use it on a windows machine, where should that configuration file be placed? I think, if it is available on windows, it would make sense to include an additional row in the table. This way the information would mirror the structure of the standard Jupyter application.

   If there is no way to use it on a windows machine, I would mention that explicitly and directly link to the page explaining the JupyterHub ["config-basics"](http://jupyterhub.readthedocs.io/en/latest/getting-started/config-basics.html) page.

</details>
</br>
<details> 
<summary> Separately from the Jupyter ecosystem docs, some parts of the <a href="http://jupyterhub.readthedocs.io/en/latest/getting-started/config-basics.html">JupyterHub config basics</a> page could be improved. </summary>
<br/>

Some changes will depend on how we answer the questions about the general Jupyter ecosystem docs. Some changes should be made, regardless of how approach the docs on the general Jupyter pages.

For example, the JupyterHub links to the IPython configuration documentation for more information. IPython's config docs conflict with the Jupyter common config docs (e.g., they have a notion of "profiles" which does not have an analogous in the Jupyter ecosystem docs).  Also IPython is a "user application with a hierarchy of system-wide and per-user configuration files".  As @minrk points out, that description does not describe JupyterHub, meaning we should remove that link to avoid encouraging people to think about the IPython & JupyterHub config systems in analogy to one another. 

</details>
<br/>
<details>
<summary> A bonus if you've clicked on any of these detail areas 😃: </summary>
Thank you for taking the time to read through this. I tried to make it as digestable as possible by allowing it to be consumed gradually, only introducing greater detail when it was needed.

Enjoy:
details kitten

</details>

yuvipanda:
Mostly just wanted to +1 @minrk. If we need to standardize, standardizing on `/etc/jupyterhub` makes sense to me.

mpacer:
@yuvipanda I agree with that having heard @minrk’s reasoning…  I’m now trying to think through how to beat promulgate that info given that `etc/jupyterhub` is the most appropriate standard.

minrk:
> Are there any technical downsides to including it there?

Part of the reason it's separate from `/etc/jupyter` is that often authentication tokens are in jupyterhub config that user's shouldn't be able to read, so `/etc/jupyterhub` may be a private directory users can't read, whereas `/etc/jupyter` must be readable by all users in order to work. Further, a user may look in `/etc/jupyter` and see that all files in there may affect their environment. JupyterHub config would not be included in that.

> Can we encode JupyterHub's config convention in either the jupyter or jupyterhub packages so that it can be imported by other packages? Which of these do you think makes the most sense?

I think there's one point of confusion about JupyterHub's convention not being importable. It's not importable because it doesn't exist. Unlike Jupyter applications, JupyterHub has no default config path (except the current working directory). JupyterHub does *not* load config in /etc/jupyterhub by default. JupyterHub config files are loaded explicitly via `jupyterhub -f /path/to/some/config.py`. The docs point to `/etc/jupyterhub` only because that's the Linux standard location for system-wide application config, and the docs cover setting up a JupyterHub instance putting everything where a Linux user might expect it (config in /etc, logs in /var/log, etc.). If jupyterhub starts loading `/etc/jupyterhub` by default, I definitely agree that it should done in an importable way that you can use, and that would be in `jupyterhub` itself, not `jupyter_core`.

--------

As for documentation, I don't think any of the user-facing documentation of the common Jupyter configuration should mention JupyterHub. At most, perhaps a small note in the common config docs that says something like:

> Note: JupyterHub is not a Jupyter Application and does not use this configuration system. See the JupyterHub docs for configuring JupyterHub `<link>`.

JupyterHub is well outside the scope of users configuring their own Jupyter environment (even users in a JupyterHub environment do not need any access to JupyterHub's own configuration). Instead, I think what's appropriate is to include these clarifications and caveats in the JupyterHub docs. That way, users reading about JupyterHub are informed that JupyterHub is not a Jupyter application (from a config perspective) and does its own thing, but users only exploring the docs for configuring their own environment don't get extra information they don't need. It's probably appropriate to have this info at the very beginning of configuring JupyterHub docs:

1. link to jupyter config docs for configuring user environments
2. make it clear that JupyterHub is *not* a Jupyter Application (in terms of the definition available in the config docs)
3. clearly state how jupyter looks for config (i.e. `jupyterhub_config.py` in the CWD, or `-f /path/to/any/config.py`)

A possible place for the more detailed distinction is some documentation on configuring user environments with JupyterHub. JupyterHub deployments are one of the main cases that these system-wide configuration files (like /etc/jupyter) are used, because a sysadmin is setting up environments on behalf of multiple users. That might be an opportune place for making it clear that JupyterHub does not use these files, but the Jupyter applications that it manages will.

> Is there a way to use JupyterHub on a windows machine? Or do we only need to consider the Unix case?

It is technically possible, though it is not supported, and not documented. The defaults do not work on Windows. We should make it clear in the docs that Windows is not supported by the default Spawner and Authenticator.

mpacer:
@minrk thank you for responding in such detail!

> Part of the reason it's separate from `/etc/jupyter` is that often authentication tokens are in jupyterhub config that user's shouldn't be able to read, so `/etc/jupyterhub` may be a private directory users can't read, whereas `/etc/jupyter` must be readable by all users in order to work. Further, a user may look in `/etc/jupyter` and see that all files in there may affect their environment. JupyterHub config would not be included in that.

Those seem like excellent reasons and good explanations of why it has a different system from Jupyter applications. My hunch is that this info would be good to include in the documentation section that you describe later regarding configuring user environments with JupyterHub. 

> Unlike Jupyter applications, JupyterHub has no default config path (except the current working directory). 

I'd say that that is a default (and is consistent with the `jupyter_core` implicit inclusion of the current working directory when finding config files.

> If jupyterhub starts loading `/etc/jupyterhub` by default, I definitely agree that it should done in an importable way that you can use, and that would be in `jupyterhub` itself, not `jupyter_core`.

That makes sense, is there a reason for it not to do that? 

The one that jumps out to me is that any kind of default "magic" can trip people up pretty easily. The chances of being tripped up by unknown defaults increased in the context of more complicated tasks (like running a JupyterHub). 

>  At most, perhaps a small note in the common config docs that says something like:
> > Note: JupyterHub is not a Jupyter Application and does not use this configuration system. See the JupyterHub docs for configuring JupyterHub <link>.

That sounds great!

> That way, users reading about JupyterHub are informed that JupyterHub is not a Jupyter application (from a config perspective) and does its own thing, but users only exploring the docs for configuring their own environment don't get extra information they don't need. It's probably appropriate to have this info at the very beginning of configuring JupyterHub docs:
> 
> 1. link to jupyter config docs for configuring user environments
> 2. make it clear that JupyterHub is not a Jupyter Application (in terms of the definition available in the config docs)
> 3. clearly state how jupyter looks for config (i.e. jupyterhub_config.py in the CWD, or -f /path/to/any/config.py)

That also sounds great!

> A possible place for the more detailed distinction is some documentation on configuring user environments with JupyterHub. JupyterHub deployments are one of the main cases that these system-wide configuration files (like /etc/jupyter) are used, because a sysadmin is setting up environments on behalf of multiple users. That might be an opportune place for making it clear that JupyterHub does not use these files, but the Jupyter applications that it manages will.

I think including an explanation of that sort would be exceptionally useful.

minrk:
>> If jupyterhub starts loading /etc/jupyterhub by default, I definitely agree that it should done in an importable way that you can use, and that would be in jupyterhub itself, not jupyter_core.

> That makes sense, is there a reason for it not to do that?

Not really. Let's do that now!

Thanks for thinking carefully about this! I think we have some actions to take:

1. (code) make jupyterhub look in /etc/jupyterhub for config by default
2. (jupyterhub docs) clearly state where JupyterHub looks for config (CWD now, CWD and /etc/jupyterhub soon), and discuss configuration in more detail
3. (jupyterhub docs) add some documentation on configuring user environments via things like /etc/jupyter, linking to Jupyter config docs. We have some small notes to this effect, but I think a dedicated new doc with examples is a good idea.
4. (jupyter docs) get a minimal "If you are looking to configure JupyterHub, you should be looking over here <link>" note or two, strategically placed in the Jupyter config docs.

Relevant locations:

- [configuration basics](https://jupyterhub.readthedocs.io/en/latest/getting-started/config-basics.html#configuration-basics) currently describes the CWD loading, and is probably the place for more detail. It also has the link to IPython config docs that probably leads to some of this confusion.
- [installation basics](https://jupyterhub.readthedocs.io/en/latest/installation-basics.html#folders-and-file-locations) mentions file/folder locations as *recommendations* for places, but could be more clear about where we look vs UNIX conventions we recommend that *users* follow.

willingc:
@mpacer I'm doing some repo cleanup today. I'm going to go ahead and close this PR. Thanks for clarifying things.

PR #1639, opened 2018-01-31
parallelize startup poll

### LINKED ISSUE #1633: Parallelize user server status check
At startup, the hub takes a long time to check the status of user servers when there are thousands of accounts.

minrk:
puts each check for a running spawner in a coroutine and runs them all concurrently.

Note: this will only improve performance when a large number of Spawners are running and `yield spawner.poll()` takes a nontrivial amount of time. This is because these are coroutines, not threads. If instantiating Spawners themselves takes a long time, performance will not be affected.

closes #1633

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1204,8 +1204,6 @@ class JupyterHub(Application):
     def init_spawners(self):
         db = self.db
 
-        user_summaries = ['']
-
         def _user_summary(user):
             parts = ['{0: >8}'.format(user.name)]
             if user.admin:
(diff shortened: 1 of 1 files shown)


jupyterhub/app.py
```diff
@@ -1226,39 +1224,49 @@ def user_stopped(user, server_name):
             yield self.proxy.delete_user(user, server_name)
             yield user.stop(server_name)
 
+        @gen.coroutine
+        def check_spawner(user, name, spawner):
+            status = 0
+            if spawner.server:
+                try:
+                    status = yield spawner.poll()
```
yuvipanda: in KubeSpawner, poll doesn't actually have to be a coroutine. Is there something we can do that optimizes based on that info?

minrk: If poll isn't a coroutine, making these concurrent will have no benefit whatsoever, so it's better for it to be a coroutine if it's going to take any time at all.

PR #1642, opened 2018-02-01
Fixes for tests with tornado 5

minrk:
I believe these are *not* compatibility fixes for tornado in JupyterHub in general, as they only affect the tests.

Things started failing because until recently, the tests were never running with asyncio. This was because importing the singleuser server triggered a pyzmq registration which overrode the asyncio registration we have at the module level (which occurred earlier than the notebook import). pyzmq 17 prerelease no longer does this, so asyncio was really being used, revealing a couple of small bugs. These are only in testing utilities involving background threads and cleanup.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -31,7 +31,6 @@ from tornado.log import app_log, access_log, gen_log
 import tornado.options
 from tornado import gen, web
 from tornado.platform.asyncio import AsyncIOMainLoop
-AsyncIOMainLoop().install()
 
 from traitlets import (
     Unicode, Integer, Dict, TraitError, List, Bool, Any,
@@ -1643,6 +1642,7 @@ class JupyterHub(Application):
     @classmethod
     def launch_instance(cls, argv=None):
         self = cls.instance()
+        AsyncIOMainLoop().install()
         loop = IOLoop.current()
         loop.add_callback(self.launch_instance_async, argv)
         try:
--- a/jupyterhub/tests/conftest.py
+++ b/jupyterhub/tests/conftest.py
@@ -45,9 +45,7 @@ def io_loop(request):
 
     def _close():
         io_loop.clear_current()
-        if (not ioloop.IOLoop.initialized() or
-                io_loop is not ioloop.IOLoop.instance()):
-            io_loop.close(all_fds=True)
+        io_loop.close(all_fds=True)
 
     request.addfinalizer(_close)
     return io_loop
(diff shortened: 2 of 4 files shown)


jupyterhub/tests/conftest.py
```diff
@@ -45,9 +45,7 @@ def io_loop(request):
 
     def _close():
         io_loop.clear_current()
-        if (not ioloop.IOLoop.initialized() or
-                io_loop is not ioloop.IOLoop.instance()):
-            io_loop.close(all_fds=True)
+        io_loop.close(all_fds=True)
```
willingc: Nice. I had isolated `initialized` as a problem yesterday and tried without one of the if conditions. Should have removed both. As an FYI as we think about Tornado 5.0 (as it has quite a few changes with IOLoop), pytest-asyncio is a thing now (based on pytest-tornado). I've also played a bit with MonkeyType and pyAnnotate after having a discussion with Guido about typing/debugging async.

jupyterhub/app.py
```diff
@@ -1643,6 +1642,7 @@ def launch_instance_async(self, argv=None):
     @classmethod
     def launch_instance(cls, argv=None):
         self = cls.instance()
+        AsyncIOMainLoop().install()
```
willingc: Not quite sure why this moves into the class method, other than its functionality changed with the deprecation.
http://www.tornadoweb.org/en/latest/asyncio.html#tornado.platform.asyncio.AsyncIOMainLoop

jupyterhub/tests/conftest.py
minrk: I think this bit of checking was from when we had multiple loops running in threads. We don't need to do all that now that we have just one loop.

I've been thinking about whether pytest-asyncio is appropriate, too. I think we still need to explicity run the tornado eventloop, I'm not sure.

jupyterhub/app.py
minrk: This is a change I've been meaning to do for some time. I don't think triggering configuration at import time is generally the right thing to do. The fact that this was at the module level is the reason we've never been running tests with asyncio, despite all of the code in jupyterhub indicating that we are. Another bit of configuration in the notebook package was overriding that choice with pyzmq's eventloop at a later point, and thus taking precedence. Moving it here ensures that the choice takes place.

minrk:
Thanks!

jupyterhub/app.py
willingc: Cool. Thanks for the explanation.

jupyterhub/tests/conftest.py
willingc: Quite likely, we may have a set of tests that use pytest-Tornado and gradually introduce others with pytest-asyncio.

PR #1643, opened 2018-02-01
avoid instantiating idle Spawner objects during startup

minrk:
only instantiate Spawners when they are requested.

This changes startup time on my laptop with 5k users of whom 500 are running from 3 minutes to less than 3 seconds. With no running users, startup completes in <del>200ms</del> 8 seconds (also down from 3 minutes). I'll post startup perf tests soon.

<del>
This is a WIP because I think we need to investigate other times to avoid instantiating the objects prematurely before this is helpful. My current suspicion is that the entire startup delay will just be inherited by the first call to the /users API (e.g. by the culler), which might be worse than having it consistently at startup.
</del>

Related to #1633

Follow-up to #1639 which only solves the case where Spawner.poll spends a significant amount of time waiting.

### DIFF
--- a/jupyterhub/apihandlers/base.py
+++ b/jupyterhub/apihandlers/base.py
@@ -8,6 +8,7 @@ from http.client import responses
 
 from tornado import web
 
+from .. import orm
 from ..handlers import BaseHandler
 from ..utils import url_path_join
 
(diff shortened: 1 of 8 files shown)


minrk:
I've done some basic timing with [this script](https://github.com/minrk/jupyterhub-startup-perf), and found this to be a bit hairy. The main costs to performance are:

1. instantiating orm.User objects
2. instantiating higher-level User object wrappers around orm.User 
3. instantiating Spawner objects
4. invalidating cached orm objects on commit (#1291)

I've been measuring the following times:

1. startup time (and specifically time spend initializing Spawners, which is usually most of it)
2. the time of the first GET /api/users because delayed instantiation at startup may all land here, as all users are looked at
3. the time of the *second* GET /api/users to see the relative time of the first request, and if the time of an average request changes.

The big startup cost right now is that we instantiate all User wrappers and their Spawners. These are the avenues I have pursued:

1. During startup, only instantiate those Users who have possibly running Spawners via a db query (Spawner.server is None via query, rather than after instantiation). This has a huge impact on startup time (200ms vs 3 minutes for 5k idle users), but shifts much of that time to the first GET /api/users request, as all the User wrappers are still instantiated there.
2. avoid creating user wrappers in GET /api/users, allowing models to be created directly from ORM objects for non-running users. This has a big improvement on the first request, but makes the *average* request take much longer, as the users are no longer cached in memory, so get recreated on each request. This throws a wrench in the one-session-per-request optimization we've been kicking around, as this suggests that would make GET all users *very* expensive. Apparently the in-memory cache we've been maintaining has a huge benefit performance-wise.
3. make instantiating Users a little cheaper by not using traitlets (little difference)

Following these bits of info, I think the following is the best course of action:

1. restore instantiating all Users during JupyterHub startup. This ensures that we have the fast cache of all users in memory. It slows startup, but ensures that this cost is only paid during startup, rather than at a later time when users may be interacting with the Hub.
2. delay Spawner instantiation until use. Since instantiating Spawners is the bigger cost and they can be more safely delayed, that will be the source of the performance improvement. This should bring startup costs into a more reasonable zone (~10-20s vs ~3 minutes for 5k users).

minrk:
This is working now with the latest timing, fixing the few bugs I introduced.

PR #1648, opened 2018-02-06
Add option to redirect to running Jupyter server

rschroll:
Currently, on login or `/hub`, users get redirected to their server, if it is running.  This is the desired behavior in most cases, but we are using the control panel as a central store of information and would like to have users always go through there.  This adds a redirect_to_server option, which defaults to True.  If it is set to False, the user will not get redirected to a running server in these cases.


minrk:
Great!

PR #1653, opened 2018-02-09
Add PyCharm folder to gitignore

dhirschfeld:


mpacer:
In general we try not to include editor specific paths or files in the project level gitignore. You can always add it yourself in either a global git ignore or a project specific git ignore inside .git/info/exclude. See more at https://git-scm.com/docs/gitignore

dhirschfeld:
Is there any reason/benefit for that policy? It seems to me that it just puts a completely unnecessary hurdle in the way of someone wanting to contribute, especially a new contributor or someone unfamiliar with git.

Telling a Windows user to manually edit a file in a folder which they can't even see (because it's hidden by default) seems pretty user hostile.

I'm experienced enough that it doesn't faze me either way but even I didn't know about that because I've never had to before - it seems most projects add any and all editor specific folders to the `.gitignore` so that it Just Works for anyone wanting to contribute, irrespective of whatever tools they're using.

I can appreciate the policy from a development-purity aspect but are there any practical advantages to it? Is there ever a case where you wouldn't want `.idea` (or `.vscode`, etc...) in the `.gitignore`?

minrk:
I think it's fine to include things like this if people are really running into them. We already include editor-specific things like `*~`, which is from emacs, I believe.

I think mostly what we do is avoid *including* project config files, rather than ignoring various sources of hidden files.

PR #1660, opened 2018-02-13
expand admin[-access] support

### LINKED ISSUE #1629: Spawner.options_form is not shown when servers are started from the admin panel
which prevents servers from spawning, because the `user_options` is empty and throws `KeyError` when the spawner tries to get values from it.

1. Enable `admin_access` and add (an) admin user(s)
2. Add an `options_form` to spawner and use some field value from the form in the spawner by accessing `user_options`.
3. Log in as admin and go to admin panel
4. Push the `start server` button.

The form is shown.

**What actually happens**

An `API request failed (500): error` error raised, because an attempt to use the `user_options` in the Spawner ended with the `KeyError` exception.

From the `jupyterhub/jupyterhub:latest` docker image.

minrk:
- [x] admin request for /user/:name spawns user's server, not redirecting to admin's server
- [x] admins can see/submit spawn form for users (if admin access is enabled)
- [x] spawn links in admin panel should trigger spawn form if defined

closes #1629

### DIFF
--- a/share/jupyterhub/static/js/admin.js
+++ b/share/jupyterhub/static/js/admin.js
@@ -6,6 +6,8 @@ require(["jquery", "bootstrap", "moment", "jhapi", "utils"], function ($, bs, mo
 
     var base_url = window.jhdata.base_url;
     var prefix = window.jhdata.prefix;
+    var admin_access = window.jhdata.admin_access;
+    var options_form = window.jhdata.options_form;
 
     var api = new JHAPI(base_url);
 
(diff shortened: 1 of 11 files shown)


consideRatio:
Wieee!!! @minrk this is a great improvement!

share/jupyterhub/static/js/admin.js
```diff
@@ -121,7 +135,10 @@ require(["jquery", "bootstrap", "moment", "jhapi", "utils"], function ($, bs, mo
         });
     });
 
+<<<<<<< HEAD:share/jupyterhub/static/js/admin.js
```
willingc: Looks like some merge conflict cruft.

minrk:
Rebased to fix the erroneous conflict marker

willingc:
Thanks  @minrk.

PR #1661, opened 2018-02-14
Additional information about which ports can be used and which are reserved for internal use only.

GladysNalvarte:


jupyterhub/app.py
```diff
@@ -290,7 +290,11 @@ def _template_paths_default(self):
         """
     ).tag(config=True)
     ip = Unicode('',
-        help="The public facing ip of the whole application (the proxy)"
+        help="""The public facing ip of the whole application (the proxy)
```
willingc: Perhaps:

"...the whole JupyterHub application (specifically referred to as the proxy)."

jupyterhub/app.py
```diff
@@ -325,10 +329,19 @@ def _domain_default(self):
         return urlparse(self.subdomain_host).hostname
 
     port = Integer(8000,
-        help="The public facing port of the proxy"
+        help="""The public facing port of the proxy.
+        
+        This is the port on which the proxy will listen. 
+        This is the only port through which jupyterhub 
+        should be accessed.
```
willingc: ...accessed by users.

jupyterhub/app.py
```diff
@@ -383,10 +396,19 @@ def _deprecated_proxy_api(self, change):
...
+        
+        This is the internal port of the hub itself. It should never be accessed directly.
+        See JupyterHub.port for the public port to use when accessing jupyterhub.
+        It is rare that this port should be set except in cases of port conflict.
+        """
     ).tag(config=True)
     hub_ip = Unicode('127.0.0.1',
         help="""The ip address for the Hub process to *bind* to.
+        
+        The hub listens on localhost only by default. This address must be accessible from
```
willingc: Perhaps for clarity:

By default, the hub listens on localhost only.

GladysNalvarte:
@willingc thanks for your comments

minrk:
Thanks!

Issue #1502, opened 2017-10-24
Username in JupyterHub header [customizing templates]


scbarton:
As a supervisor I log into a lot of different JupyterHub accounts, and I'd like to be able to see the username of the current login. I wrote a short iPython extension below, which adds the username to the kernel_logo_widget element. It works with the %load_ext magic, but I'm not sure how to make it work when automatically loaded.

I've tried adding the following to ipython_kernel_config.py:

    c.IPKernelApp.extensions = [ 'NameInHeader' ]

with the idea that the extension would load with the kernel. This loads the extension, but maybe too soon because the header is not changed.  Is there a better way to accomplish what I'm trying to do?

```
# NameInHeader.py

def load_ipython_extension(ipython):
    
    import re, os
    from IPython.display import display, Javascript

    uname= os.getenv('USER')

    s='''var s = document.getElementById('kernel_logo_widget').innerHTML;
         var news= '<span style=\\'color:#919191\\'>_user_ </span>' + s;
         document.getElementById('kernel_logo_widget').innerHTML=news;
    '''

    func=re.compile('_user_')
    s=func.sub(uname, s)

    display(Javascript(s))
```

minrk:
A custom template is probably what is desired here, and while technically possible now, customizing templates needs some work to be really useful (#1481).

Siecje:
I am trying to add the current user to the notebook list page and the notebook page itself.

I can't modify the notebook template here because {{ user }} is not available.
https://github.com/jupyter/notebook/blob/master/notebook/templates/page.html

JupyterHub could add the current user as a literal somewhere but I'm not sure where do to that.
https://github.com/jupyterhub/jupyterhub/blob/c9db504a49deb57d3494ce331fc6c904cf0a9e57/jupyterhub/singleuser.py#L141

manics:
I'm closing this as it's quite old and there's no concrete action point. If you're still having problems please post on the [Jupyter Community Forum](https://discourse.jupyter.org/) or open a new issue. Thanks!

PR #1663, opened 2018-02-14
 Move right navbar into a separate file.

Siecje:
I want to remove the logout link and add the current user. Instead of modifying `page.html` now I can modify `navbar_right.html`. This is much easier for an automated deployment.

Relevant issues
#1481
#1502

My personal use case is to add the username and remove the logout link so for me navbar_right.html will contain.

```HTML
<ul class="nav navbar-nav navbar-right">
  {% if user %}
  <li><span class="navbar-text">{{ user.name }}</span></li>
  {% endif %}
</ul>
```

This won't apply to the Jupyter pages though, when viewing list of notebooks and the notebook page itself.

### DIFF
--- a/share/jupyter/hub/templates/home.html
+++ b/share/jupyter/hub/templates/home.html
@@ -8,7 +8,7 @@
       {% if user.running %}
       <a id="stop" role="button" class="btn btn-lg btn-danger">Stop My Server</a>
       {% endif %}
-      <a id="start"role="button" class="btn btn-lg btn-success" href="{{ url }}">
+      <a id="start" role="button" class="btn btn-lg btn-success" href="{{ url }}">
       {% if not user.running %}
       Start
       {% endif %}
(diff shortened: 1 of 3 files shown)


minrk:
Rather than moving this to a file, I think we should just make it an overrideable block (see the issues you linked for reference, and specifically #1601 which initiates the extensible template process).

Siecje:
What would the usage look like?

To change something in the base template you would have to edit `page.html`.

I think I am missing something.

minrk:
Since #1601 has landed, you can extend any template by creating a directory and adding it to `template_paths`. To override a single block in `page.html` make a `page.html` in your template path and give it just the empty block you want to override:

```html
{% extends "templates/page.html" %}
{% block navbar_right %}
{% endblock %}
```

would replace the navbar_right block on every page with an empty block, effectively deleting it. What's missing in jupyterhub that's already present in the notebook templates is the block markers themselves.

Siecje:
Thanks @minrk

minrk:
We're still missing the block you need to override this particular element in the jupyterhub templates. If you'd like to submit a PR to add the block around the relevant area, then I think you should be set.

rschroll:
FWIW, there's already a `login_widget` block that contains the logout button.  This might be all you need to override.

That said, I think there's a strong case to be made for adding more, and more-descriptive, blocks for the various bit.

PR #1665, opened 2018-02-16
Add documentation for new templates system

rschroll:
This is to be applied on top of #1601.

### DIFF
--- a/docs/source/reference/templates.md
+++ b/docs/source/reference/templates.md
@@ -1,0 +1,34 @@
+# Templates
+
+The pages of the JupyterHub application are generated from [Jinja](http://jinja.pocoo.org/) templates.  These allow the header, for example, to be defined once and incorporated into all pages.  By providing your own templates, you can have complete control over JupyterHub's appearance.
+
+## Custom Templates
+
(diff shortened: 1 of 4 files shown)


rschroll:
I tried to build the docs locally, but ran into problems with a missing theme.  I think everything is reasonable, but do double check me.

docs/source/reference/templates.md
```diff
@@ -0,0 +1,34 @@
+# Templates
+
+The pages of the JupyterHub application are generated from [Jinja](http://jinja.pocoo.org/) templates.  These allow the header, for example, to be defined once and incorporated into all pages.  By providing your own templates, you can have complete control over JupyterHub's appearance.
+
+## Custom Templates
+
+JupyterHub will look for custom templates in all of the paths in the `JupyterHub.template_paths` configuration option, falling back on the [default templates](https://github.com/jupyterhub/jupyterhub/tree/master/share/jupyter/hub/templates) if no custom template with that name is found.  This means you can override as many or as few templates as you desire.
```
minrk: Can you add a little note that template_paths is new in 0.9?

minrk: Oops, rather than `{% extends templates/... %}` support is new. This config isn't the new part.

rschroll:
Try this on for size.

minrk:
Perfect, thanks!

PR #1670, opened 2018-02-21
Add blocks for navbar components in the page template

rschroll:
This makes one block containing the whole nav bar, one containing the elements of the left list, and one containing the elements of the right list.  I hope this gives enough flexibility without cluttering the markup too much, but I'm not married to any of these ideas.

See also #1663, #1481, #1502.

### DIFF
--- a/share/jupyter/hub/templates/page.html
+++ b/share/jupyter/hub/templates/page.html
@@ -82,7 +82,8 @@
   </div>
 </noscript>
 
-<nav class="navbar navbar-default">
+{% block nav_bar %}
+  <nav class="navbar navbar-default">
     <div class="container-fluid">
       <div class="navbar-header">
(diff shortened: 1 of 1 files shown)


share/jupyter/hub/templates/page.html
```diff
@@ -82,7 +82,8 @@ <h4 class="modal-title" id="{{key}}-label">{{title}}</h4>
   </div>
 </noscript>
 
-<nav class="navbar navbar-default">
+{% block nav_var %}
```
minrk: nav_bar

minrk:
one typo, but 👍

rschroll:
I made that same typo in one of the sub-blocks, too, and was proud to have caught it before I pushed the PR.

minrk:
Thanks!

PR #1684, opened 2018-02-27
Add `Spawner.get_options_form` for async support.

### LINKED ISSUE #1681: Can't use `User.get_auth_state` to configure `Spawner.options_form`
I'm using JupyterHub 0.8.1 (on Kubernetes). I am attempting to create a custom `Spawner.options_form` whose content is based on group membership information returned by a custom Authenticator. The groups are successfully retrieved and stored using the `auth_state` interface (introduced in 0.8.0), but I don't think it's possible to use this data to configure the form from my Spawner. The fundamental issue (as I understand it), is that the `User.get_auth_state` method is async (returns a Future), whereas the `Spawner.options_form` method is synchronous.

I've reviewed the approach linked from here: https://github.com/jupyterhub/jupyterhub/issues/1542. This Authenticator/Spawner pair is creative, interesting and useful, but it does not utilize the `auth_state` interface to store information from the Authenticator and pass that to the Spawner. 

A similar (less specific) issue is open on the `wrapspawner` project, here: https://github.com/jupyterhub/wrapspawner/issues/7. I left some remarks there about what I see and why it doesn't appear to me to be possible to use the existing interfaces to accomplish what I described. 

I have attempted to directly block and obtain the result of the Future returned from `User.get_auth_state`, but I was unsuccessful in doing so. I tried:
1. `self.user.get_auth_state().result()`, but that doesn't work: `Exception: DummyFuture does not support blocking for results`
2. `IOLoop.current().run_sync(self.user.get_auth_state)`, but that failed with: `RuntimeError: This event loop is already running`

I considered attempting to pass the information (group memberships) between Authenticator and Spawner through some other route. In the issue on `wrapspawner` I pointed out that `Authenticator.pre_spawn_start` and `Spawner.pre_spawn_hook` are both called after the `Spawner.options_form` has been rendered. Furthermore, I don't see an existing place in the interfaces that will be durable across restarts (if the timing of one of the two `pre_spawn_*` hooks was correct, I believe it could be used to pass state in-memory, but that it would not be durable across restarts). 

Based on this investigation, it appears to me that the only option is to change the interface of `Spawner.options_form` to be async (return a Future), and to introduce `yield` to the call-sites that invoke it. This does not appear to be a massive change, but it also does not seem trivial. I'm also unsure as to whether `Traitlets` properties can return Futures, or whether the change to async will require converting `options_form` away from `Traitlets`.

Have I missed something? Is there some other way to access `auth_state` from `options_form`? Is there some less invasive (localized) way to allow `options_form` to retrieve `auth_state`?

emmanuel:
Fixes #1681. The implementation is heavily influenced by some patterns in the kubespawner, specifically: `Union([Unicode(), Callable()])`, and supporting sync or async callables.

I have locally run these changes, but I've not yet added any tests. I found no existing tests for the `Spawner.options_form`, so I would appreciate any guidance about how to approach testing this change (e.g., are there any examples of similar tests I can draw from? which file should the tests go in?).

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -17,17 +17,17 @@ from tempfile import mkdtemp
 
 from sqlalchemy import inspect
 
-from tornado import gen
+from tornado import gen, concurrent
 from tornado.ioloop import PeriodicCallback
 
 from traitlets.config import LoggingConfigurable
 from traitlets import (
-    Any, Bool, Dict, Instance, Integer, Float, List, Unicode,
+    Any, Bool, Dict, Instance, Integer, Float, List, Unicode, Union,
     observe, validate,
 )
 
 from .objects import Server
-from .traitlets import Command, ByteSpecification
+from .traitlets import Command, ByteSpecification, Callable
 from .utils import random_port, url_path_join, exponential_backoff
 
 
@@ -227,7 +227,10 @@ class Spawner(LoggingConfigurable):
         help="Enable debug-logging of the single-user server"
     ).tag(config=True)
 
-    options_form = Unicode(
+    options_form = Union([
+            Unicode(),
+            Callable()
+        ],
         help="""
         An HTML form for options a user can specify on launching their server.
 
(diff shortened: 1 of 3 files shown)


jupyterhub/spawner.py
```diff
@@ -247,8 +250,32 @@ def name(self):
...
 
+    @gen.coroutine
+    def get_options_form(self):
+        """Get the options form
+
+        Returns:
+          (Future(str)): the content of the options form presented to the user
+          prior to starting a Spawner.
+
+        .. versionchanged:: 0.8.2
```
minrk: This can be `.. versionadded: 0.9`

emmanuel: Great, thanks for the feedback! Updated.

minrk:
A+ thanks!

minrk:
If you'd like to give testing a go, that would most logically be in `test_spawner` or `test_pages`. [Here is the existing test for options_form](https://github.com/jupyterhub/jupyterhub/blob/ec8d008678669813f3433a0912bdf8c511da43be/jupyterhub/tests/test_pages.py#L149) in test_pages. The various mock Spawner classes used in tests are defined [in mocking](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/tests/mocking.py#L125), if you'd like to extend the FormSpawner class, or add a new one.

PR #1691, opened 2018-03-01
talk to PAM in a thread

minrk:
since PAM can be slow sometimes, we don't want to block the rest of the application while we wait for it

cc @fterbo


minrk:
Thanks, @willingc!

PR #1709, opened 2018-03-08
Suggest retry timing when we throttle server starts

### LINKED ISSUE #1706: Provice explicit Retry-After guidance when throttling users
When we throttle server starts because too many spawns are pending, we should explicitly provide users with guidance on how long to wait before retrying. This should be a uniform distribution with a min, max that is configured.

In general, if average server start time is X, we want users to try again, on average, after X. This lets us maximize throughput of users starting while minimizing user wait time.

yuvipanda:
Fixes #1706

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -616,6 +616,31 @@ class JupyterHub(Application):
         """
     ).tag(config=True)
 
+    throttle_retry_suggest_min = Integer(
+        30,
+        help="""
+        Minimum seconds after which we suggest the user retry spawning.
+
+        When `concurrent_spawn_limit` is exceeded, we recommend users retry
+        after a random period of time, bounded by throttle_retry_suggest_min
+        and throttle_retry_suggest_max.
+
+        throttle_retry_suggest_min should ideally be set to the median
+        spawn time of servers in your installation.
+        """
+    )
+
+    throttle_retry_suggest_max = Integer(
+        60,
+        help="""
+        Minimum seconds after which we suggest the user retry spawning.
+
+        When `concurrent_spawn_limit` is exceeded, we recommend users retry
+        after a random period of time, bounded by throttle_retry_suggest_min
+        and throttle_retry_suggest_max.
+        """
+    )
+
     active_server_limit = Integer(
         0,
         help="""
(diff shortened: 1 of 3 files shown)


yuvipanda:
I think it might be nice to autorefresh page with JS after that much time, but that's a different patch!

jupyterhub/handlers/base.py
```diff
@@ -500,17 +501,25 @@ def spawn_single_user(self, user, server_name='', options=None):
...
+            retry_time = int(random.uniform(
+                self.settings['throttle_retry_suggest_min'],
+                self.settings['throttle_retry_suggest_max']
+            ))
+            self.set_header('Retry-After', str(retry_time))
+            self.log.info(
+                '%s pending spawns, throttling. Retry in %s seconds',
+                spawn_pending_count, retry_time
             )
+            self.set_status(429, "Too many users trying to log in right now. Try again in a {}s".format(retry_time))
```
betatim: -> "Try again in {}seconds"?

jupyterhub/app.py
```diff
@@ -606,6 +606,31 @@ def _authenticator_default(self):
         """
     ).tag(config=True)
 
+    throttle_retry_suggest_min = Integer(
```
betatim: nit: can we make this a tuple instead of two integers? Saves some repeated docs and wouldn't have made me initially think that this is in min(utes) instead of the min(imum).

willingc: If we anticipate these values changing on the fly while running, the tuple's immutability might be a downside.

Perhaps:

`spawn_throttle_retry_wait_floor` and `spawn_throttle_retry_wait_ceiling`

jupyterhub/app.py
```diff
@@ -606,6 +606,31 @@ def _authenticator_default(self):
...
+        30,
+        help="""
+        Minimum seconds after which we suggest the user retry spawning.
```
willingc: Smallest period (floor) in seconds to suggest a user wait before retrying.

When `concurrent_spawn_limit` is exceeded, spawning is throttled. We
suggest users wait a random period of time, bounded by the floor and ceiling,
before retrying.

The floor should ideally be set to the median spawn time of your installation.

jupyterhub/app.py
```diff
@@ -606,6 +606,31 @@ def _authenticator_default(self):
...
+
+        throttle_retry_suggest_min should ideally be set to the median
+        spawn time of servers in your installation.
+        """
+    )
+
+    throttle_retry_suggest_max = Integer(
+        60,
+        help="""
+        Minimum seconds after which we suggest the user retry spawning.
```
willingc: Largest period (ceiling) in seconds to suggest a user wait before retrying.

When `concurrent_spawn_limit` is exceeded, spawning is throttled. We
suggest users wait a random period of time, bounded by the floor and ceiling,
before retrying.

PR #1720, opened 2018-03-14
Updated CircleCI config file to 2.0 

### LINKED ISSUE #1699: Update CircleCI config to 2.0
CircleCI, which we use to test our docker builds, has updated its config spec to 2.0. We should update our circleci config to 2.0, following [this guide](https://circleci.com/docs/2.0/migrating-from-1-2/) .

glenak1911:
Closes #1699

I've updated the config file according to [the documentation](https://circleci.com/docs/2.0/migrating-from-1-2/). I also had to consult a few other examples to try and get the syntax/semantics correct. 

I'd like to test this, but I'm not sure how to go about doing that(don't want to break any builds), as well as have this reviewed for accuracy. Any pointers would be helpful.


minrk:
If you'd like to test, you can try enabling circleci on your own fork and push these same changes to your master branch. The docker login / upload step won't succeed because it won't have the credentials, but it should at least get that far.

.circleci/config.yml
```diff
@@ -0,0 +1,37 @@
+# Python CircleCI 2.0 configuration file
+# Updating CircleCI configuration from v1 to v2
+# Check https://circleci.com/docs/2.0/language-python/ for more details
+#
+version: 2
+jobs:
+
+  test:
```
minrk: I was wrong. We do need a build job here. Or we need to use a custom [workflow](https://circleci.com/docs/2.0/workflows/).

.circleci/config.yml
```diff
@@ -0,0 +1,37 @@
...
+              docker login -u $DOCKER_USER -p $DOCKER_PASS -e unused@example.com
+              docker push jupyterhub/jupyterhub-onbuild
+    release:
+      filters:
+        tags:
+          only: /.*/
+        steps:
+          - checkout
+          - run:
+            name: Push tagged branch to corresponding image
```
minrk: name and command here need to be indented one level beyond `run`:

```yaml
- run:
    name: ...
```

.circleci/config.yml
minrk: I think all of the jobs need `machine: true` in the new 2.0 config

.circleci/config.yml
```diff
@@ -0,0 +1,37 @@
...
+          command: |
+            docker build -t jupyterhub/jupyterhub .
+            docker build -t jupyterhub/jupyterhub-onbuild:${CIRCLE_TAG:-latest} onbuild
+
+  deploy:
+    hub:
+      branches:
+        only: master
+      steps:
+        - checkout
```
minrk: I don't think the deployments can have a checkout step

.circleci/config.yml
```diff
@@ -0,0 +1,37 @@
...
+  test:
+    steps:
+      - run:
+          name: Test image build
+          command: |
+            docker build -t jupyterhub/jupyterhub .
+            docker build -t jupyterhub/jupyterhub-onbuild:${CIRCLE_TAG:-latest} onbuild
+
+  deploy:
+    hub:
```
minrk: Deploy can't have sub-keys anymore. It can only have `steps`

minrk:
I think we can actually remove the deploy and release steps and move those over to docker's own automated builds, if you'd like to try that. That should simplify things because then we only have the builds and no conditional login/push steps.

minrk:
Ignore that last comment, I forgot about jupyterhub-onbuild, which would be a little tricky on docker. Let's try to get this working as-is here.

glenak1911:
@minrk thanks for the reviews. I've made some changes, including adding workflows. It's still failing on my fork with the following:

`Configuration errors: 1 error occurred:

* Error migrating config to version 2: 2 errors occurred:

* in job 'workflows': steps is not a list
* in job 'workflows': steps is not a list`

The errors aren't very verbose, and I'm not sure if there is anything missing that is readily apparent to you.

glenak1911:
I have to update my branch on my fork with the changes I made. Will post that later this evening.

.circleci/config.yml
```diff
@@ -0,0 +1,79 @@
...
+      enabled: true
+    steps:
+      - checkout
+      - run:
+          command: docker login -u $DOCKER_USER -p $DOCKER_PASS -e unused@example.com
+      - run:
+          command: docker push jupyterhub/jupyterhub-onbuild:$CIRCLE_TAG
+
+
+  workflows:
```
minrk: workflows should be dedented so it aligns with 'jobs' rather than as an item under 'jobs'

glenak1911:
@minrk Thanks for all the help! I tested this out in my fork and it looks good! I think this is complete.

package-lock.json
```diff
@@ -0,0 +1,671 @@
+{
```
minrk: Oops! This file shouldn't be committed.

minrk: We should add it to .gitignore. Feel free to do that here if you want.

.circleci/config.yml
```diff
@@ -0,0 +1,79 @@
...
+      - run:
+          command: docker build -t jupyterhub/jupyterhub .
+      - run:
+          command: docker build -t jupyterhub/jupyterhub-onbuild:${CIRCLE_TAG:-latest} onbuild
+
+  deploy:
+    machine:
+      enabled: true
+    steps:
+      - checkout
```
minrk: deploy and release shouldn't need checkout steps, only test

package-lock.json
glenak1911: ah understood, I've added updated the gitignore

.circleci/config.yml
glenak1911: sounds good, I've removed the checkout step from deploy and release

minrk:
Thanks!

PR #1722, opened 2018-03-14
Simplify root/login redirect behavior

### LINKED ISSUE #1721: audit default URLs and redirects
Currently, there are too many different initial states that can affect how users arrive at the Hub and where they end up. In particular, the biggest source of confusion is the "if the server is running, default to their server, otherwise default to /hub/home" behavior. This is implemented separately and slightly differently in the root handler and the login handler.

The current state of things, as far as I can tell:

- `/` -> `/hub/`
  - not logged in: redirect to `/hub/login`
  - logged in
    - ready: send to running server at `/user/:name`
    - not ready: send to /hub/home
- `/hub/login`
  - not running: trigger launch
  - ready: send to running server at `/user/:name`
  - not ready (e.g. slow spawn): send to `/hub/home`

Two common sources of confusion with this:

1. where the user ends up on login is sensitive to whether spawn is quick or not
2. visiting the same URL with the Hub in the same state sends the user to a different location, depending on whether the user was already logged in or not:
  1. not logged in -> login -> triggers spawn -> /user/:name
  2. logged in -> doesn't trigger spawn -> /hub/home

Proposal:

- always or never send users to their server from `/hub/` and `/hub/login`. #1648 added `.redirect_to_server` setting which covers the 'never' case. We can use this to implement the 'always' case as well, rather than the current state-dependent behavior.
- ensure /hub/ and /hub/login use the same code for picking destinations (`.get_next_url()`)
- remove implicit spawn from /hub/login, instead relying on `/hub/user/:name` handler to trigger spawn (or render spawn form), so that requests sending users to /hub/home don't trigger spawn

This should result in more consistent behavior that's easier for everyone to follow.

cc @ppLorins who brought up default-url confusion most recently

minrk:
- ignore spawner state when determining redirect destination
- remove implicit spawn from login handler (rely on redirect to user.url for spawn)
- settings.redirect_to_server determines if login sends users to /user/:name vs /hub/home
- visiting `/hub/` should result in the same destination regardless of login state or spawner state

closes #1721

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -392,20 +392,35 @@ class BaseHandler(RequestHandler):
     def get_next_url(self, user=None):
         """Get the next_url for login redirect
 
-        Defaults to hub base_url /hub/ if user is not running,
-        otherwise user.url.
+        Default URL after login:
+
+        - if redirect_to_server (default): send to user's own server
+        - else: /hub/home
         """
         next_url = self.get_argument('next', default='')
         if (next_url + '/').startswith('%s://%s/' % (self.request.protocol, self.request.host)):
             # treat absolute URLs for our host as absolute paths:
             next_url = urlparse(next_url).path
-        if not next_url.startswith('/'):
+        if next_url and not next_url.startswith('/'):
+            self.log.warning("Disallowing redirect outside JupyterHub: %r", next_url)
             next_url = ''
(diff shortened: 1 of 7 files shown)


minrk:
@yuvipanda this will be relevant to your stress tests because a user login no longer implies spawn except via redirect to /user/:name. So unless you follow that redirect with cookies, you will need to make an explicit request to spawn servers.

PR #1731, opened 2018-03-18
Allow checking PAM account stack during PAM auth

vilhelmen:
Hello everybody, this patch lets the PAM authenticator do an optional account check during auth, which enables JH to respect PAM-accessible user access management. The current PAM auth pipeline checks that a user is __authenticated__, not whether they are __authorized__ for the given service.

If you have something AD/LDAP/SSO-like configured at the system level that speaks PAM and has service access restrictions (e.g. students in a department or single course may use the machine), JH's PAM authenticator won't enforce them, as it stops at authentication.

One workaround is to maintain a separate user whitelist and enforce that in the auth stack via pam_listfile.so, but it's a nasty hack as the file containing whitelisted usernames has to be constantly updated instead of relying on the preexisting, external, PAM-able system. The option to open a session on login does not trigger authorization checks, and that setting turns itself off when it fails anyway.

JH does have whitelisting features and LDAP plugins, but they add additional configuration and points of failure on systems where PAM already provides authentication and authorization.

This patch adds the configuration setting c.PAMAuthenticator.check_account to enable the check, and it is disabled by default. I've tested this patch out in the wild and also added a test, and all relevant tests are passing. 2 non-auth tests (one proxy and one app I believe) consistently failed on this branch and master, but that's probably docker's fault.

On a related note, I noticed that c.PAMAuthenticator.encoding is never used, is that a mistake? I forwarded the encoding to pamela for the account check as that seemed like the intended purpose. I can fix the missing encoding uses, or remove my own, depending on whether we're supposed to use the encoding setting or not.

### DIFF
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -550,6 +550,19 @@ class PAMAuthenticator(LocalAuthenticator):
         """
     ).tag(config=True)
 
+    check_account = Bool(True,
+        help="""
+        Whether to check the user's account status via PAM during authentication.
+
+        The PAM account stack performs non-authentication based account 
+        management. It is typically used to restrict/permit access to a 
+        service and this step is needed to access the host's user access control.
+
+        Disabling this can be dangerous as authenticated but unauthorized users may
+        be granted access and, therefore, arbitrary execution on the system.
+        """
+     ).tag(config=True)
+
     def __init__(self, **kwargs):
         if pamela is None:
             raise _pamela_error from None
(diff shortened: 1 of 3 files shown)


minrk:
Great, thanks!

> On a related note, I noticed that c.PAMAuthenticator.encoding is never used, is that a mistake?

Yes, I think it is. It would be great if you added them here.

vilhelmen:
Tweaked the tests a bit and sprinkled the encoding setting all over.

Do we want to have a discussion about switching this to on by default? I would argue that this is the correct/original intended behavior and that the current implementation is incorrect, perhaps to the point of being a hole in security, but I'm also aware that fixing the PAM auth pipeline could uncover issues in existing server configurations. I think enabling this by default would be in the project's best interest, but I will defer to your judgement.

minrk:
Enabling it by default sounds like the right thing to do, thanks.

vilhelmen:
I'm glad you agree. I flipped the setting default, spookied-up the help text a bit to (hopefully) dissuade people from turning it off without knowing what they're doing, and tweaked the tests to reflect the new default. Should be good to go!

minrk:
Thanks!

PR #1733, opened 2018-03-19
Add a spawner_class option to the reference example

willingc:
Remind the person creating the configuration file that `spawner_class` should be set when uses a spawner other than the default spawner. [Clarification to assist end users]


minrk:
Thanks!

PR #1739, opened 2018-03-21
Add Authenticator.blacklist

ndiy:
introduce blacklist to supplement white list

### DIFF
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -88,6 +88,18 @@ class Authenticator(LoggingConfigurable):
         """
     ).tag(config=True)
 
+    blacklist = Set(
+        help="""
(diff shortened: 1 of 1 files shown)


jupyterhub/auth.py
```diff
@@ -205,6 +217,19 @@ def check_whitelist(self, username):
             return True
         return username in self.whitelist
 
+    def check_blacklist(self, username):
+        """Check if a username is blocked to authenticate based on blacklist configuration
+
+        Return True if username is allowed, False otherwise.
+        No blacklist means any username is allowed.
+
+        Names are normalized *before* being checked against the whitelist.
```
minrk: s/whitelist/blacklist

PR #1743, opened 2018-03-21
[MRG] Expose `auth_state` via /api/users/<name>

### LINKED ISSUE #1724: Update user's auth state from a running notebook server
I have a hub where we forward the oauth2 token from the login to the user's notebook server ([like so](https://github.com/jupyterhub/oauthenticator/blob/master/examples/auth_state/jupyterhub_config.py)). There they can use it to make auth'ed requests to a few services.

My question is what do others do to keep the token that the notebook server has fresh? The oauth2 token is only valid for ~24h which is much shorter than the "logout time" for the hub (14days by default?).

Right now I have a nb server extension that periodically refreshes the token but as soon as you do that the auth state that the hub has is outdated (the oauth refresh token there becomes invalid). So I am thinking of writing a hub service that the nbserver extension can talk to. Once it refreshes the tokens it then uses the service to update what the hub's user object contains.

But this sounds quite convoluted so wondering what others do in this situation.

betatim:
Closes #1724.

This let's you get the auth state via the hub's API. It required making several methods async and adds a new key.

I would also like to modify the POST method so that users (admins) can modify their own (a user's) auth state.

The use case is allowing a service to refresh the auth tokens obtained during login to the hub and store the fresh tokens in the auth state so that newly spawned user servers pick them up.

---

When running the tests I noticed that one of them (`test_app.py::test_load_groups`) creates a `jupyterhub.sqlite` in the root directory. As a result it gets confused if there already is one there that has the wrong schema version. Is this on purpose or hard to fix?

### DIFF
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -170,7 +170,10 @@ class UserAPIHandler(APIHandler):
             if self.find_user(data['name']):
                 raise web.HTTPError(400, "User %s already exists, username must be unique" % data['name'])
         for key, value in data.items():
-            setattr(user, key, value)
+            if key == 'auth_state':
+                await user.save_auth_state(value)
+            else:
+                setattr(user, key, value)
         self.db.commit()
         user_ = self.user_model(user)
         user_['auth_state'] = await user.get_auth_state()
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -112,7 +112,8 @@ class UserAPIHandler(APIHandler):
         # this means users can't see their own auth state unless they
         # are admins, Hub admins often are also marked as admins so they
         # will see their auth state but normal users won't
-        if user.admin:
+        requestor = self.get_current_user()
+        if requestor.admin:
             user_['auth_state'] = await user.get_auth_state()
         self.write(json.dumps(user_))
 
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -24,9 +24,7 @@ class SelfAPIHandler(APIHandler):
             user = self.get_current_user_oauth_token()
         if user is None:
             raise web.HTTPError(403)
-        user_ = self.user_model(user)
-        user_['auth_state'] = await user.get_auth_state()
-        self.write(json.dumps(user_))
+        self.write(json.dumps(self.user_model(user)))
 
 
 class UserListAPIHandler(APIHandler):
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -110,7 +110,12 @@ class UserAPIHandler(APIHandler):
     async def get(self, name):
         user = self.find_user(name)
         user_ = self.user_model(user)
-        user_['auth_state'] = await user.get_auth_state()
+        # auth state will only be shown if the requestor is an admin
+        # this means users can't see their own auth state unless they
+        # are admins, Hub admins often are also marked as admins so they
+        # will see their auth state but normal users won't
+        if user.admin:
+            user_['auth_state'] = await user.get_auth_state()
         self.write(json.dumps(user_))
 
     @admin_only
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -17,20 +17,20 @@ class SelfAPIHandler(APIHandler):
 
     Based on the authentication info. Acts as a 'whoami' for auth tokens.
     """
-    def get(self):
+    async def get(self):
         user = self.get_current_user()
         if user is None:
             # whoami can be accessed via oauth token
             user = self.get_current_user_oauth_token()
         if user is None:
             raise web.HTTPError(403)
-        self.write(json.dumps(self.user_model(user)))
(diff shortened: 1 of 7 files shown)


minrk:
> Is this on purpose or hard to fix?

Not on purpose. If you specify `db_url` in a temporary directory [as here](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/tests/test_app.py#L71), it should be avoided.

minrk:
Making `user_model` async by default is going to have big performance consequences for actions like 'get all users', which retrieves the model for all users. The user model is also used in a number of cases that probably shouldn't include auth state. I think we should be more judicious about where we add the auth state and only do it in particular circumstances. In particular:

1. including auth_state in the response should be opt-in via an additional parameter
2. auth_state should only be added on single-user endpoints (GET /user and GET /users/:name), not user-list endpoints (GET /users)
3. don't add auth_state to the default user_model, leave the user model as a synchronous call

A PATCH request to `/users/:name` is probably the right thing for updating auth state, since we have this endpoint already for renames, modifying admin, etc. Adding auth_state handling there should be relatively simple.

One thing we should think about: Should users be view and/or modify this info about themselves? Or should this only be available to admins.

jupyterhub/apihandlers/users.py
```diff
@@ -168,7 +172,9 @@ def patch(self, name):
         for key, value in data.items():
```
minrk: This is where auth_state would need special handling:

```python
if 'auth_state' in data:
    auth_state = data.pop('auth_state')
    await user.set_auth_state(...
```

betatim: Nods

betatim:
> including auth_state in the response should be opt-in via an additional parameter

even for GET /user and GET /users/:name or were you thinking of the other methods?

jupyterhub/tests/test_api.py
```diff
@@ -420,6 +422,41 @@ def test_make_admin(app):
...
+    auth_state = {'secret': 'hello'}
+    db = app.db
+    name = 'admin'
+    user = find_user(db, name)
+    assert user is not None
+
+    with mock.patch('jupyterhub.user.encrypt') as mock_encrypt:
+        encrypt_future = tFuture()
+        encrypt_future.set_result(b"encrypted_state")
+        mock_encrypt.return_value = encrypt_future
```
betatim: This whole mocking setup looks pretty ugly to me. Is there a way to make it nicer? Maybe just mock the fetching of the crypto key from the environment?

jupyterhub/tests/test_api.py
```diff
@@ -13,6 +13,7 @@
 import requests
 
 from tornado import gen
+from tornado.concurrent import Future as tFuture
```
betatim: Why is the tornado `Future` different from the `concurrent.futures` one? If I use `tFuture` in the mock below it works, but with `Future` I get an error that `object of type Future can't be await'ed`.

(this needs to be made nicer before merging)

minrk: There are three kinds of Future:

- asyncio.Future (The New Way)
- tornado.Future (with tornado 5, these are really asyncio.Future)
- concurrent.futures.Future (The Slightly Less New Way) \*

concurrent Futures are not awaitable, so must be wrapped in `asyncio.wrap_future`. Why? I don't know, but that's how it is. Our `utils.maybe_future` wrapper turns all of these things (and more!) into awaitable Futures.

\* - *multiprocessing.AsyncResult (The Old Way)*

jupyterhub/tests/test_api.py
minrk: Do you want to enable auth state in the default MockHub? Then you won't need to mock-in enabled auth state. Otherwise, you could create a decicated fixture in conftest.

jupyterhub/tests/test_api.py
betatim: Ah ok. Will switch the mock to returning a `maybe_future`. Thanks for the history lesson.

minrk:
> even for GET /user and GET /users/:name or were you thinking of the other methods?

I was thinking everything, but I'm not certain. This is the first time sensitive info of any kind is available from the REST API, so I want to make sure we are careful about it. Limiting the number of cases where it might be exposed is part of that.

We have to be careful with GET /user. This is the endpoint used for oauth, which means it is executed *by others on your behalf*. I'm thinking that maybe auth_state should *only* be accessible on GET /users/:name and *only* by admins (not even yourself).

betatim:
Restricting who gets the auth state to only admins only when they hit `/users/:user` sounds like a good plan to me. We can restrict it even further so that even an admin has to set a parameter in the request to receive the auth state. I have no preference, you decide.

I think all this is implemented and has a test now. If not that is a bug/oversight.

I will investigate why the spawn limit test is failing/timing out.

betatim:
Test passed by restarting travis :-/

minrk:
I like this! since it's admin-only on a single endpoint, I think adding an opt-in url parameter for auth_state is probably overkill. I do want to get to scopes in oauth, which will let us clearly specify whether a token or service has auth_state access or not.

betatim:
Yay! This contribution was made possible by http://openhumans.org/!

gedankenstuecke:
Well done @betatim! 🎉😍

willingc:
cc/ @summerswallow

PR #1767, opened 2018-03-30
allow spawners to provide pending_messages and inform users.

clkao:
This is prereq for https://github.com/jupyterhub/kubespawner/issues/148 and https://github.com/jupyterhub/kubespawner/issues/141

### DIFF
--- a/share/jupyterhub/templates/spawn_pending.html
+++ b/share/jupyterhub/templates/spawn_pending.html
@@ -12,7 +12,12 @@
       <p><i class="fa fa-spinner fa-pulse fa-fw fa-3x" aria-hidden="true"></i></p>
       <a role="button" id="refresh" class="btn btn-lg btn-primary" href="#">refresh</a>
     </div>
-  </div>
+    <pre>
+      {% for m in messages %}
+        {{ m }}
+      {% endfor %}
+    </pre>
+</div>
 </div>
 
 {% endblock %}
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -882,7 +882,7 @@ class UserSpawnHandler(BaseHandler):
                 self.log.info("%s is pending %s", spawner._log_name, spawner.pending)
                 # spawn has started, but not finished
                 self.statsd.incr('redirects.user_spawn_pending', 1)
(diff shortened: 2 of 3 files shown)


share/jupyterhub/templates/spawn_pending.html
```diff
@@ -12,7 +12,12 @@
       <p><i class="fa fa-spinner fa-pulse fa-fw fa-3x" aria-hidden="true"></i></p>
       <a role="button" id="refresh" class="btn btn-lg btn-primary" href="#">refresh</a>
     </div>
-  </div>
+    <pre>
+      {% for m in messages %}
```
minrk: This probably shouldn't be in `pre`. They should either be in `<p>` tags or passed as-is for HTML messages.

minrk:
Thanks for the PR! This doesn't provide the basis needed for either of the above linked issues, though. https://github.com/jupyterhub/kubespawner/issues/141 is for failed launches, which don't show on the pending-spawn page. And jupyterhub/kubespawner#148 asks for a mechanism for streaming events, not a static spawner-provided message that doesn't change. While this would improve the pending-message case somewhat because a Spawner could provide a static description describing what's going on, if we want to have a full event-stream of messages, then I think we probably don't want to implement two versions of the same thing for Spawners to deal with.

clkao:
For failed launches, I was mostly referring to slow span that eventually timed out. Otherwise Exceptions are already displayed. So the launch progress message was sufficient for my use case.

clkao:
For streaming messages we probably need to introduce websocket to the codebase first.  I thought we can start with static messages, and maybe a watcher for the message array that emits events over websocket in the future.

minrk:
The reason I'm hesitant about adding this message API is that it doesn't address what's requested in jupyterhub/kubespawner#148 and if we do implement a streaming message API, that would immediately replace this API. I'm hesitant to introduce a new API when we already have a plan to deprecate it immediately. Since the structure (necessarily async, probably an async generator) would be quite different, I'm not sure anything in this API could be reused.

minrk:
#1771 implements the async generator and event-stream

minrk:
closing in favor of #1771

Issue #1602, opened 2017-12-29
Post new group using Rest API interface


sangramga:
In order to add a group, POST method of `/hub/api/groups/{name}` URI  is to be called. But as per general standards of writing REST API services, we should call POST method of `/hub/api/groups` and mention group name only in the JSON body of POST request and not in URI.
Hence there should be a `post(self)` method in `GroupListAPIHandler`.

As these standards are followed while calling POST method in adding users `/hub/api/users` (usernames is mentioned in JSON body) but not in groups.  It would be better to follow same standards in all API's!!

minrk:
> there should be a post(self) method in GroupListAPIHandler.


Indeed there should be. Would you like to make a PR?

sangramga:
@minrk Ok. Just took a look at the code. I would make a PR.

sangramga:
@minrk We need to also change the API Documentation to include this feature.

minrk:
@sangramga good catch. Would you like to make a PR?

sangramga:
@minrk I would be glad. I'll take a look and get back to you.

PR #1768, opened 2018-03-31
Added Post groups API in group list handlers

sangramga:
Solves issue #1602 
Its possible to add/post multiple groups using the post method of "/groups" API.

### DIFF
--- a/jupyterhub/tests/test_api.py
+++ b/jupyterhub/tests/test_api.py
@@ -979,28 +979,12 @@ def test_add_multi_group(app):
     r_names = [group['name'] for group in reply]
     assert names == r_names
 
-    # for name in names:
-    #     user = find_user(db, name)
-    #     assert user is not None
-    #     assert user.name == name
-    #     assert not user.admin
-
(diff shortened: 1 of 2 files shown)


jupyterhub/tests/test_api.py
```diff
@@ -966,6 +966,42 @@ def test_groups_list(app):
...
+    assert r.status_code == 201
+    reply = r.json()
+    r_names = [group['name'] for group in reply]
+    assert names == r_names
+
+    # for name in names:
+    #     user = find_user(db, name)
+    #     assert user is not None
+    #     assert user.name == name
+    #     assert not user.admin
```
minrk: Do you want to restore or delete these commented blocks?

sangramga: Oh sorry. I will delete those comments.

minrk:
Thanks!

sangramga:
@minrk Also need to make changes to REST API Documentation(Swagger) to include the above features.

Issue #1754, opened 2018-03-27
Proxy.check_routes causing performance issue


minrk:
It's possible that there's a performance issue in check_routes. On investigating why we have some launch failures in Binder, I found that they correlated with windows of the Hub being nonresponsive for several seconds (often 7-10):

<img width="1031" alt="screen shot 2018-03-27 at 10 57 24" src="https://user-images.githubusercontent.com/151929/37957325-f056f0a4-31ad-11e8-909b-55442c408fb9.png">

The "Deleting stale routes" messages tell me (though not with 100% confidence) that we are in Proxy.check_routes, which could have a performance issue because it is a database lookup that checks all users and running spawners.

We need to profile this, because the recent startup optimizations may well have already alleviated the issue.

minrk:
I ran some profiling based on the [startup perf tests](https://github.com/minrk/jupyterhub-startup-perf) and couldn't get check_routes to take more than 250 milliseconds with 5k users even on 0.8.1, so I'm not quite sure how it could take 7-10 seconds with 100-200 users like we have on mybinder.org. With 100 active users, I'm seeing check_routes take closer to 5-6 milliseconds, so I'm not quite sure what could be up. This may be correlated with #1753 where the issue only occurs for the case where check_routes and cull-idle are occurring concurrently.

forbxy:
jupyterhub run in only one process，so if many user request check_Oauth,home page or other handler, check_routes will be slowly

PR #1775, opened 2018-04-05
proxy performance tweaks

### LINKED ISSUE #1770: proxy.check_routes can call again by timer before the last call stop
user route add is slow,so if we lost a large number of user's routes，proxy.check_routes will still running in next timer clock.

minrk:
related to #1754

This shouldn't significantly affect performance, but it might improve some edge cases. The main thing this should do is improve diagnostics when check_routes is costly.

- use semaphore to limit concurrent requests to the proxy
- add lock to prevent concurrent calls to check_routes (closes #1770)
- a bit more debug logging in check_routes
  - log (info) that check_routes is starting
  - log (debug) when fetching routes

### DIFF
--- a/jupyterhub/proxy.py
+++ b/jupyterhub/proxy.py
@@ -305,7 +305,7 @@ class Proxy(LoggingConfigurable):
 
         hub = self.app.hub
         if '/' not in routes:
-            self.log.warning("Adding missing default route")
+            self.log.warning("Adding default route")
             futures.append(self.add_hub_route(hub))
         else:
             route = routes['/']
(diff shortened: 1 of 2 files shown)


jupyterhub/proxy.py
```diff
@@ -453,9 +491,7 @@ def _check_process():
             if status is not None:
                 e = RuntimeError(
                     "Proxy failed to start with exit code %i" % status)
-                # py2-compatible `raise e from None`
-                e.__cause__ = None
-                raise e
+                raise e from None
```
willingc: Nice update.

willingc:
Thanks @minrk.

PR #1779, opened 2018-04-06
limit concurrent cull requests

minrk:
avoids simultaneous request for deletion of lots of users, which can slow down the Hub

### DIFF
--- a/examples/cull-idle/cull_idle_servers.py
+++ b/examples/cull-idle/cull_idle_servers.py
@@ -26,7 +26,7 @@ Or run it manually by generating an API token and storing it in `JUPYTERHUB_API_
     python cull_idle_servers.py [--timeout=900] [--url=http://127.0.0.1:8081/hub/api]
 """
 
-from datetime import datetime, timedelta, timezone
+from datetime import datetime, timezone
 from functools import partial
 import json
 import os
@@ -39,6 +39,7 @@ except ImportError:
 import dateutil.parser
 
 from tornado.gen import coroutine, multi
+from tornado.locks import Semaphore
 from tornado.log import app_log
 from tornado.httpclient import AsyncHTTPClient, HTTPRequest
 from tornado.ioloop import IOLoop, PeriodicCallback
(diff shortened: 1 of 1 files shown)


examples/cull-idle/cull_idle_servers.py
```diff
@@ -90,7 +91,21 @@ def cull_idle(url, api_token, inactive_limit, cull_users=False, max_age=0):
...
     now = datetime.now(timezone.utc)
     client = AsyncHTTPClient()
-    resp = yield client.fetch(req)
+
+    if concurrency:
+        semaphore = Semaphore(concurrency)
+        @coroutine
+        def fetch(req):
+            """client.fetch wrapped in a semaphore to limit concurrency"""
+            yield semaphore.acquire()
```
Carreau: Can we use `async with`  or is it 3.6+ only ?

Carreau: Nah, 3.6+ never mind. Have been running 3.6 for too long.

betatim:
LGTM.

One kind of related question: there are a few copies of the culler script floating around (here and th z2jh guide). Is it worth thinking about distributing a service like this as part of juypterhub so we can reduce the number of copies? Was thinking about this because this would be useful for mybinder.org as well.

minrk:
@betatim yeah, there are tradeoffs.

1. make it part of JupyterHub
  - pro: distributed with jupyterhub, always compatible with jupyterhub
  - con: cannot distribute fixes without making jupyterhub releases
2. make it a standalone package
  - pro: easy to reuse in multiple places
  - con: another package to maintain/distribute

At this point, making it a simple lightweight package is probably the thing to do.

minrk:
Follow-up, in case it wasn't clear already: we *do* use this in mybinder.org, since it is inherited from z2jh.

betatim:
My confusion comes about because there is https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/images/hub/cull_idle_servers.py and the copy in this repo. So right now I think: we use `cull_idle_servers.py` in mybinder.org but we use the version from the z2jh repo (not the version from this repository).

minrk:
@betatim exactly right, yes. I think a standalone package is where we should end up. For now, I'll copy this script a few more times before one of us gets bothered enough to make it a real package.

PR #1788, opened 2018-04-10
Options form template variables

rcthomas:
We are working on a custom spawner options form template for a custom spawner, and the form is a bit involved so putting it into `options_form` and hacking on it is a bit unwieldy.  With `options_form` though you can dynamically customize/personalize the form content, similar to how `batchspawner` does it.  But if the form is in a separate template, it's not quite the same.

This PR adds a `template_vars` configurable dict to `Spawner` and then when the form is rendered those variables are substituted in by Jinja2.  These variables can be computed when the spawner is instantiated --- that is the main idea, some other way of substituting in variables since we can't do .format().  I've tested it locally and it works.

### DIFF
--- a/jupyterhub/handlers/pages.py
+++ b/jupyterhub/handlers/pages.py
@@ -65,6 +65,8 @@ class SpawnHandler(BaseHandler):
     Only enabled when Spawner.options_form is defined.
     """
     async def _render_form(self, message='', for_user=None):
+        # Note that 'user' is the authenticated user making the request and
+        # 'for_user' is the user whose server is being spawned.
         user = for_user or self.get_current_user()
         spawner_options_form = await user.spawner.get_options_form()
         return self.render_template('spawn.html',
@@ -72,6 +74,7 @@ class SpawnHandler(BaseHandler):
             spawner_options_form=spawner_options_form,
             error_message=message,
             url=self.request.uri,
+            spawner=for_user.spawner
         )
 
     @web.authenticated


rcthomas:
I guess a more succinct clarification is that extending the spawn.html template is an alternative to the `options_form` string, and this PR allows you to substitute variables into that template if you do that.

minrk:
Ah, I understand now. The goal is in general extending the page template namespace for pages related to a particular Spawner. Would it be sufficient to just put `spawner` into the namespace for such pages, then you can do `{{spawner.thing}}` etc.?

rcthomas:
You mean pass `spawner=user.spawner` in the call to `render_template` here:

https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/handlers/pages.py#L70

instead of the extra dictionary argument I'm adding?  Yes, I think that would be great.  At the moment I am only thinking about doing this when rendering spawn.html.  Should it also be done elsewhere?

rcthomas:
Should have done my test-deployment check off another branch, but I confirm it works.

jupyterhub/spawner.py
```diff
@@ -299,6 +299,15 @@ def options_from_form(self, form_data):
         their server.
         """)
 
+#   template_vars = Dict(
```
willingc: @minrk Would you rather have this uncommented and default to False?

Thanks @rcthomas.

rcthomas: If you like I can resubmit this as a separate, cleaner PR and close this one.  Sorry about the mess... 😄

willingc: Leave it be for now @rcthomas. No apology needed; we appreciate your effort :cookie:

jupyterhub/handlers/pages.py
```diff
@@ -72,6 +72,7 @@ class SpawnHandler(BaseHandler):
             spawner_options_form=spawner_options_form,
             error_message=message,
             url=self.request.uri,
+            spawner=user.spawner
```
minrk: careful, this should be `for_user.spawner`

minrk: Feel free to add a comment clarifying that 'user' is the authenticated user making the request and 'for_user' is the user whose server is being spawned.

jupyterhub/spawner.py
minrk: If spawner works, feel free to delete this bit. If you like, you can rebase and squash out the changes, but I have no objection to deleting them in a new commit. Your choice! Either way, it doesn't need to be a new PR.

jupyterhub/handlers/pages.py
rcthomas: Hey @minrk if that's the case, do we actually need to bother at all?  We're already passing `for_user` through to be rendered.  I just checked, {{ for_user.spawner.whatever }} works...

yuvipanda:
https://github.com/jupyterhub/kubespawner/pull/137 for related work in the KubeSpawner project :)

jupyterhub/handlers/pages.py
minrk: let's pass spawner anyway because I think eventually we are going to need to use this form for named servers, where user.spawner isn't specific enough.

minrk:
Thanks!

PR #1790, opened 2018-04-11
raise 409 conflict on duplicate actions

minrk:
Makes it easier for clients (e.g. BinderHub) to retry failed actions and ignore failure due to duplicate transactions.


minrk:
I think the failure is caused by the missing commits fixed by #1792, which is extracted from #1771

willingc:
There are two failing tests but I'm going to merge. We can keep an eye out on the `test_start_stop_race` and `test_get_proxy`

PR #1792, opened 2018-04-12
add missing commits for deprecated ip/port

minrk:
Removes deprecated handling of Spawners that don't return (ip, port), which seems to have been causing sql issues in tests since the switch to asyncio, which has slightly different behavior for how finished coroutines are handled.


minrk:
Closing as these changes were merged as part of #1771

PR #1800, opened 2018-04-13
[WIP] API for managing tokens

minrk:
new endpoints:

- /users/:name/tokens for listing and creating new tokens
- /users/:name/tokens/:token-id for retrieving and deleting tokens

The token id should be considered an opaque identifier for tokens, and free to change in minor versions of JupyterHub. Currently it's a prefix for the table ('a' for api tokens, 'o' for oauth tokens) and the integer row id. It could also be a UUID if we determine that exposing row ids in the api token tables is undesirable.

Creating new tokens via the `/authorizations` endpoint is deprecated. The one feature not ported from the old endpoint is creating a token with a username and password. I'm not sure how useful this was, as it only worked with username+password Authenticators, and couldn't be made to work with OAuth clients.


minrk:
This is ready to go

PR #1801, opened 2018-04-13
fix and test deletion cascades

minrk:
we were leaving some remnants in the database by assuming that some items would be deleted, when they were actually being detached.

- ensure foreign keys are enabled on sqlite (for `ondelete='CASCADE'` support)
- fix deletion cascades where relationships were causing dissociation
  instead of deletion
- handle possible failure to delete oauth tokens

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -195,11 +195,6 @@ class BaseHandler(RequestHandler):
         orm_token = orm.OAuthAccessToken.find(self.db, token)
         if orm_token is None:
             return None
-        if orm_token and not orm_token.client_id:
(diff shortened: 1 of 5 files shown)


jupyterhub/handlers/base.py
```diff
@@ -195,11 +195,10 @@ def get_current_user_oauth_token(self):
         orm_token = orm.OAuthAccessToken.find(self.db, token)
         if orm_token is None:
             return None
-        else:
-            orm_token.last_activity = \
-                orm_token.user.last_activity = datetime.utcnow()
-            self.db.commit()
-            return self._user_from_orm(orm_token.user)
+        orm_token.last_activity = \
```
willingc: Perhaps remove \ and make a long line

PR #1803, opened 2018-04-13
typo catching 404 in proxy.delete_route

minrk:
status code is .code, not .status_code


willingc:
:tada:

Push the green button @minrk when you are ready.

PR #1804, opened 2018-04-15
/user/someone-else redirection retain query strings

### LINKED ISSUE #1769: Query strings are lost in between /user/redirect/ and /user/my-username/
I hoped to use [nbgitpuller](https://github.com/data-8/nbgitpuller) in order to be able to write something like `https://my-z2jh-deployment.com/user/redirect/git-pull?repo=my-repo-url` in order to first have the user login, and then do something specific, such as triggering nbgitpuller.

This would had worked fine if I wanted to do something that didn't require a query string in the url, like: `https://my-z2jh-deployment.com/user/redirect/terminals/1`.

But... Query strings are lost in between the redirection from `/user/redirect/` to `/user/my-username/` :'(

query-truncated

### Update 1
I think of this as a bug, because I guess this is not the indented behavior.

consideRatio:
Closes #1769 - A redirect from `/user/someone-else-user-account` to `/user/my-user-account` now retain query strings on redirection.

I simply copy-pasted the "retain query string"-logic from this redirection that functioned properly already.
https://github.com/jupyterhub/jupyterhub/blob/58069d015be486085351c98c14fd2181a58528c1/jupyterhub/handlers/base.py#L1011-L1017

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -985,6 +985,9 @@ class UserSpawnHandler(BaseHandler):
             # logged in as a different user, redirect
             self.statsd.incr('redirects.user_to_user', 1)
             target = url_path_join(current_user.url, user_path or '')
+            if self.request.query:
+                # FIXME: use urlunparse instead?
+                target += '?' + self.request.query
             self.redirect(target)
         else:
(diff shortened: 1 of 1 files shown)


minrk:
Perfect, thanks!

PR #1807, opened 2018-04-16
cull_idle_servers.py: Fix problem when age/inactive undefined

rkdarst:
I was testing the new cull_idle_servers.py, and found two problems:

1. In my test server, inactive and age could be undefined, setting value to `'unknown'` which results in the timedelta comparisons failing.  This PR fixes this so far.
2. Sometimes, it would try to delete a non-running server.

I don't know enough about jupyterhub internals to know what the right answer is (is jupyterhub returning something strange or is `cull_idle_servers.py` not handling it right)?  For the second issue, I'm not sure what the expected internal state is to be able to tell when to not cull.  This is what the user state looks like:

    {'kind': 'user', 'name': 'darstr1', 'admin': True, 'groups': [], 'server': None, 'progress_url': '/dev/hub/api/users/USERNAME/server/progress', 'pending ': None, 'created': '2018-04-14T19:14:43.199512Z', 'started': None, 'last_activity': '2018-04-16T20:47:57.478030Z'}

Is `started` the relevant thing to condition on when deleting?

I'm sorry if things are confusing, I am still trying to understand these things.  I may be interpreting things wrong and the solutions may be wrong.

### DIFF
--- a/examples/cull-idle/cull_idle_servers.py
+++ b/examples/cull-idle/cull_idle_servers.py
@@ -66,6 +66,8 @@ def format_td(td):
 
     as HH:MM:SS
     """
+    if td is None:
+        return "unknown"
     if isinstance(td, str):
         return td
     seconds = int(td.total_seconds())
@@ -129,7 +131,7 @@ def cull_idle(url, api_token, inactive_limit, cull_users=False, max_age=0, concu
             age = now - parse_date(server['started'])
         else:
             # started may be undefined on jupyterhub < 0.9
-            age = 'unknown'
+            age = None
 
         # check last activity
         # last_activity can be None in 0.9
@@ -142,7 +144,8 @@ def cull_idle(url, api_token, inactive_limit, cull_users=False, max_age=0, concu
             # for running servers
             inactive = age
 
-        should_cull = inactive.total_seconds() >= inactive_limit
+        should_cull = (inactive is not None and
+                       inactive.total_seconds() >= inactive_limit)
         if should_cull:
             app_log.info(
                 "Culling server %s (inactive for %s)",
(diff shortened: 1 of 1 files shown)


examples/cull-idle/cull_idle_servers.py
```diff
@@ -66,6 +66,8 @@ def format_td(td):
 
     as HH:MM:SS
     """
+    if td is None:
```
minrk: Can we return `'unknown'` if td is None?

minrk:
Thanks! I think you're right that the bug is in cull_idle_servers.

I think they key here is indeed that servers are being investigated for culling when they aren't running. These fixes do make sense, but I want to add one more: `handle_server` shouldn't be called for a non-running non-pending server.

So this bit:

```python
servers = user.get(
            'servers',
            {
                '': {
                    'started': user.get('started'),
                    'last_activity': user['last_activity'],
                    'pending': user['pending'],
                }
            }
        )
```

should be 

```python
if 'servers' in user:
    servers = user['servers']
else:
    servers = {}
    # add default server if it's active
    if user['server']:
        servers[''] =  {
            'started': user.get('started'),
            'last_activity': user['last_activity'],
            'pending': user['pending'],
            'url': user['server'],
        }
```

That way, `handle_server` won't even be called if the server isn't running, which is what we want, I think, and the underlying source of the bug.

rkdarst:
First commit amended to solve the comment and also solve the problem in the user-culling section.

New commit added based on your suggestion, and now the `DELETE` problem seems solved.  Thanks for the tip!

I have done minimal testing, but haven't tested all code paths... still, probably better than before.

Anything else I can do/test?

minrk:
Looks great, thanks!

willingc:
Thanks for the contribution @rkdarst :cake:

PR #1808, opened 2018-04-16
Initial Windows support (LocalAuthenticator and base Spawner)

adelcast:
Add limited Windows support. The PR will make jupyterhub work on Windows when using the LocalAuthenticator and when using the default spawner, which will launch the singleuser server as a process on the system.  To get correct account environments, jupyterhub needs to run as a service (System user). If ran as another user, it will work but the processes will be spawned on temp directories. 

Jupyterhub on Windows uses the auth_state to keep the authentication token. Which means that C.Authenticator.enable_auth_state needs to be enabled, and JUPYTERHUB_CRYPT_KEY needs to be set.


adelcast:
I believe the Jupyterhub project doesn't intend to emphasis the Windows case. Since I had no choice to run it on *nix, I implemented the support and posted the PR in case is of interest to the community. I am fine if only some (or none) commits are taken in. On another note, I didn't find documentation on linters, so this code is not linted. If I need to run pylint, please let me know and I can update the PR.

dhirschfeld:
I'd tend to agree with @minrk that Windows support is probably better off as separate packages rather than having switch statements all over the base class implementations.

On Windows I'd highly recommend to just use `DockerSpawner` so am curious why you didn't choose to use that? In the enterprise environment where I'm running JupyterHub we use the `LDAPAuthenticator` which works well. If you don't have LDAP in your environment maybe the `WinAuthenticator`  will be useful as a standalone package.

adelcast:
Hi @minrk , ok, that sounds good, I'll go ahead and create a WinAuthenticator and a WinLocalProcessSpawner on their own repos as I believe others can benefit from the work I did.

I think I can pull that off, but still there are 2 commits in this PR that affect core jupyterhub that will prevent Windows from working:

1) init_secrets: skip permissions check on Windows
2) ConfigurableHTTPPRoxy.stop: kill child processes on Windows case

For 1), The permission model is different on Windows, so we need to skip the permission check (the switch on os.chmod can be removed, since that command won't affect Windows). We could either skip the permission check if Windows or add a config parameter to the configuration file, to skip the check (which will make this a config option, not a Windows specific option).

For 2). there is already code that sets Shell=True if os.name == 'nt'. However, that's broken since on stop, only the shell process is killed, not the proxy itself.

what do you think?
-

adelcast:
@dhirschfeld , I did look at the DockerSpawner. The problem is that using it requires Windows 10 (Anniversary edition) and a global docker configuration to use Linux containers. That puts a limitation on the systems that could use Jupyterhub on Windows...and adds the extra configuration for the type of containers selection which might conflict with other apps running on the system.

I do plan to use the LDAPAuthenticator (and verified that it works on Windows), but wanted to create a reference implementation for a default Authenticator on Windows that behaves in a similar way as the Posix one (creates processes as the logged in user, loads its profile so they get their own sandbox, etc).

Also, during authentication you have access to login/password, that's why I added the step that creates the Auth token inside the WinAuthenticator. For LDAP, we need to either do the same thing for Windows (create the Auth token) or pass the login/pass as auth_state to the spawner, since the LocalWindowsSpawner needs that to be able to spawn as another user. That's one of the things that is different between Windows and *nix: you can't impersonate a user, unless you have their credentials.

dhirschfeld:
> *it requires Windows 10 (Anniversary edition) and a global docker configuration to use Linux containers*

You do have to be running the latest Windows version. Not sure why you would use Linux containers though? I'm running Windows Containers for everything.

Also, I think in the *very* latest you can happily run linux containers alongside windows containers:
https://stefanscherer.github.io/sneak-peek-at-lcow/

adelcast:
Ok, I added the custom authenticator and spawner to github:

https://github.com/ni/jupyterhub-winauthenticator
https://github.com/ni/jupyterhub-winlocalprocessspawner

however, for the WinAuthenticator and WinLocalProcessSpawner to work, core Jupyterhub needs 2 modifcations (listed above), @minrk any direction on what's sensible to do?

dhirschfeld:
@minrk - the permissions check below is actually a problem on Windows and one I've been meaning to follow up.
https://github.com/jupyterhub/jupyterhub/blob/6b1046697af076d1c24977616466982ecd3875a6/jupyterhub/app.py#L885-L891

***xref:*** [Gitter conversation](https://gitter.im/jupyterhub/jupyterhub?at=59cd90cb7b7d98d30d4c64fa) on Sep 29 2017

Whilst Windows isn't officially supported, if it's not difficult to support and doesn't clutter up the code *too* much it would be good to allow it to work.

I think the approach in faac969a8d41953ff498e6eb20ef574b9d8da62d is reasonable and is similar in spirit to the existing hack:
https://github.com/jupyterhub/jupyterhub/blob/6b1046697af076d1c24977616466982ecd3875a6/jupyterhub/proxy.py#L486-L489

I'll put in a PR with just that change and we can discuss the pros/cons there...

dhirschfeld:
I'll have a look at 5e7712d3eb6793ba3907e223d45d6ac92bbdb884 also...

adelcast:
hi @dhirschfeld, any thoughts on [5e7712d ](https://github.com/jupyterhub/jupyterhub/commit/5e7712d3eb6793ba3907e223d45d6ac92bbdb884) ? is this something that you were planning on submitting on a separate PR or should I do it? Either way works for, just wanted to follow up...

dhirschfeld:
@adelcast - feel free to open a PR for that change as I'm pretty stretched for time.

I was planning on testing the code to try and identify the problem myself. I assume you're saying the existing code leaves an orphaned process (the proxy) on Windows? That being the case I would assume that you wouldn't be able to re-run jupyterhub on the same port if the proxy was still running and that's something I haven't observed.

adelcast:
@dhirschfeld , correct, without this fix if I shutdown the Hub via the "Shutdown Hub" button on the Admin console, the configurable-http-proxy process will stay up. Running jupyterhub again gives me an error:

[W 2018-05-15 12:52:02.104 JupyterHub proxy:458] Running JupyterHub without SSL.  I hope there is SSL termination happening somewhere else...
[I 2018-05-15 12:52:02.104 JupyterHub proxy:460] Starting proxy @ http://127.0.0.1:8090/jupyterhub/
[D 2018-05-15 12:52:02.105 JupyterHub proxy:461] Proxy cmd: ['configurable-http-proxy', '--ip', '127.0.0.1', '--port', '8090', '--api-ip', '127.0.0.1', '--api-port', '8001', '--error-target', 'http://127.0.0.1:8081/jupyterhub/hub/error'] [D 2018-05-15 12:52:02.109 JupyterHub proxy:493] Proxy started and appears to be up
[D 2018-05-15 12:52:02.112 JupyterHub proxy:566] Proxy: Fetching GET http://127.0.0.1:8001/api/routes
12:52:02.115 - warn: [ConfigProxy] 403 GET /api/routes
[E 2018-05-15 12:52:02.117 JupyterHub app:1629]
      File "C:\Python\lib\site-packages\jupyterhub\app.py", line 1627, in launch_instance_async
        yield self.start()
      File "C:\Python\lib\site-packages\tornado\gen.py", line 1099, in run
        value = future.result()
      File "C:\Python\lib\site-packages\tornado\gen.py", line 1107, in run
        yielded = self.gen.throw(*exc_info)
      File "C:\Python\lib\site-packages\jupyterhub\app.py", line 1575, in start
        yield self.proxy.check_routes(self.users, self._service_map)
      File "C:\Python\lib\site-packages\tornado\gen.py", line 1099, in run
        value = future.result()
      File "C:\Python\lib\site-packages\tornado\gen.py", line 1107, in run
        yielded = self.gen.throw(*exc_info)
      File "C:\Python\lib\site-packages\jupyterhub\proxy.py", line 296, in check_routes
        routes = yield self.get_all_routes()
      File "C:\Python\lib\site-packages\tornado\gen.py", line 1099, in run
        value = future.result()
      File "C:\Python\lib\site-packages\tornado\gen.py", line 1107, in run
        yielded = self.gen.throw(*exc_info)
      File "C:\Python\lib\site-packages\jupyterhub\proxy.py", line 603, in get_all_routes
        resp = yield self.api_request('', client=client)
      File "C:\Python\lib\site-packages\tornado\gen.py", line 1099, in run
        value = future.result()
    tornado.httpclient.HTTPError: HTTP 403: Forbidden

[D 2018-05-15 12:52:02.122 JupyterHub application:647] Exiting application: jupyterhub

λ 12:52:02.404 - info: [ConfigProxy] Proxying http://127.0.0.1:8090 to (no default)
12:52:02.410 - info: [ConfigProxy] Proxy API at http://127.0.0.1:8001/api/routes
12:52:02.415 - error: [ConfigProxy] Uncaught Exception Error: listen EADDRINUSE 127.0.0.1:8090
    at Object._errnoException (util.js:1022:11)
    at _exceptionWithHostPort (util.js:1044:20)
    at Server.setupListenHandle [as _listen2] (net.js:1351:14)
    at listenInCluster (net.js:1392:12)
    at doListen (net.js:1501:7)
    at _combinedTickCallback (internal/process/next_tick.js:141:11)
    at process._tickCallback (internal/process/next_tick.js:180:9)
    at Function.Module.runMain (module.js:686:11)
    at startup (bootstrap_node.js:187:16)
    at bootstrap_node.js:608:3
12:52:02.417 - error: [ConfigProxy] Uncaught Exception Error: listen EADDRINUSE 127.0.0.1:8001
    at Object._errnoException (util.js:1022:11)
    at _exceptionWithHostPort (util.js:1044:20)
    at Server.setupListenHandle [as _listen2] (net.js:1351:14)
    at listenInCluster (net.js:1392:12)
    at doListen (net.js:1501:7)
    at _combinedTickCallback (internal/process/next_tick.js:141:11)
    at process._tickCallback (internal/process/next_tick.js:180:9)
    at Function.Module.runMain (module.js:686:11)
    at startup (bootstrap_node.js:187:16)
    at bootstrap_node.js:608:3
12:52:05.266 - warn: [ConfigProxy] Interrupted

dhirschfeld:
Thanks for the info - I'll test it out to confirm.

I usually just shutdown my hub by madly hitting `Ctrl-C` in the terminal!

PR #1809, opened 2018-04-17
don't expire objects on commit

minrk:
after much sqlalchemy digging, we may have a sustainable solution for #1291.

The key points:

1. expire-on-commit, sqlalchemy's default behavior, means that on every commit, all of our local objects are expired.
   This means not only that we must re-retrieve information when we next use it,
   but also that expiry can itself be costly, as we can have many objects in memory to be marked as expired on every commit.
1. To enable this optimization, we must use sqlachemy's orm-level `relationships` for setting and propagating foreign-key relationships.
   This is what allows local updates of `User.api_tokens` lists, etc. to always be up-to-date.
   This means never setting foreign key columns directly, always assigning foreign keys via relationship
   (`APIToken(user=user)` not `APIToken(user_id=user.id)`).
1. relationships cover deletion in the `user.groups.remove(group)` direction, but not the `db.delete(group)` direction.
   We seem to have to add an sqlalchemy event listener for the deletion event to handle this.
   The current implementation expires relationships to a deleted entity.

Using sqlalchemy's relationships basically nullifies database-level ondelete behavior. This makes deletion quite a bit more costly, as sqlalchemy will propagate deletions in Python as one big transaction, rather than relying on the database-level implementation, which is much more efficient. E.g. for deleting a user:

1. query to find all spawners, api tokens, oauth tokens, codes, etc. for the user
2. delete everything that would be orphaned
3. keep querying in case that cascades more steps
4. finally delete the user

This could be a single delete statement with passive-deletes enabled.

There is a passive-delete option that can be used on relationships to skip this and rely on the database, but then we lose local propagation of deletions and are back to inconsistent state.

### DIFF
--- a/jupyterhub/oauth/store.py
+++ b/jupyterhub/oauth/store.py
@@ -70,11 +70,12 @@ class AccessTokenStore(HubDBMixin, oauth2.store.AccessTokenStore):
 
         """
 
-        user = self.db.query(orm.User).filter(orm.User.id == access_token.user_id).first()
+        user = self.db.query(orm.User).filter_by(id=access_token.user_id).first()
         if user is None:
             raise ValueError("No user for access token: %s" % access_token.user_id)
+        client = self.db.query(orm.OAuthClient).filter_by(identifier=access_token.client_id).first()
         orm_access_token = orm.OAuthAccessToken(
-            client_id=access_token.client_id,
+            client=client,
             grant_type=access_token.grant_type,
             expires_at=access_token.expires_at,
             refresh_token=access_token.refresh_token,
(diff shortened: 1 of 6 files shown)


minrk:
Since this is a little risky, we have a couple of options to mitigate that risk:

1. make expire_on_delete a flag deployments can set, in case they want to remove this optimization
2. expire all objects periodically, to ensure that any database inconsistency can only last a finite period of time, e.g. five minutes.

Expiring all objects is not very costly in absolute terms. It is mostly costly because we do it on *every commit*, which is quite frequent.

expire_on_commit also strongly encodes the notion that the Hub owns the database, as it ensures that certain changes made directly to the database are not reflected in the Hub until restart.

minrk:
This should be ready for testing

meeseeksmachine:

https://discourse.jupyter.org/t/confusion-of-the-db-instance/3878/3

PR #1810, opened 2018-04-17
service.stop is not async

minrk:


jupyterhub/app.py
```diff
@@ -1487,7 +1487,7 @@ def write_pid_file(self):
         if managed_services:
             self.log.info("Cleaning up %i services...", len(managed_services))
             for service in managed_services:
-                futures.append(asyncio.ensure_future(service.stop()))
+                service.stop()
```
willingc: Perhaps add a debug log entry here outputting the service name.

minrk: Good idea. Added.

Issue #1755, opened 2018-03-28
Add state of spawners to API


rkdarst:
I'm interested in different cull times for different types of servers. I'm using `wrapspawner.ProfileSpawner`, and would like different timeouts depending on the server type in use (larger memory=shorter cull time).  Right now I have hacked this in the cull script to access the hub database directly and get the info on the spawner the user is using.  If spawner info could be passed via the API, this would be easier.

I could work on a PR sometime, if you interested.  Please let me know what should be included (I would start with the `state` column from the spawners table - that's enough for me).  It could be admin only.

rkdarst:
This was fixed in #1813.

PR #1813, opened 2018-04-18
add server.state to server model if requester is an admin

rkdarst:
In  #1755, I asked if the spawner state could be passed through the API.  My use case is `cull_idle_servers.py`, where I have ProfileSpawner and cull things differently depending on what resources are used (this will require a custom culler, but that's just a fact of life).  Right now, I hack it by making a separate independent database connection to read this state.

This is an initial implementation of this, adding the attribute 'state' to servers, or 'server_state' to the user model (unnamed servers).  Only admin users can get this state, since it might be confidential for some spawners.  It seems to work for me, but there is probably more work to do.

Currently, at least two tests fail because the state is non-deterministic, but I'd rather get feedback before going deeper to save myself effort in case this isn't wanted.  (and some tests that pass could start failing later, and other tests could start failing later so that needs fixing).

What do you think?

(another option would be a spawner option for lifetime that gets passed through.  I use batchspawner, which has a maximum total life but doesn't use recent activity.  This option seems like one of the cleanest separations.)

### DIFF
--- a/jupyterhub/apihandlers/base.py
+++ b/jupyterhub/apihandlers/base.py
@@ -90,7 +90,7 @@ class APIHandler(BaseHandler):
             'message': message or status_message,
         }))
 
-    def server_model(self, spawner):
+    def server_model(self, spawner, include_state=False):
         """Get the JSON model for a Spawner"""
         return {
             'name': spawner.name,
@@ -98,6 +98,7 @@ class APIHandler(BaseHandler):
             'started': isoformat(spawner.orm_spawner.started),
             'pending': spawner.pending,
             'ready': spawner.ready,
+            'state': spawner.get_state() if include_state else None,
             'url': url_path_join(spawner.user.url, spawner.name, '/'),
             'progress_url': spawner._progress_url,
         }
@@ -137,7 +138,7 @@ class APIHandler(BaseHandler):
         model.update(extra)
         return model
 
-    def user_model(self, user, include_servers=False):
+    def user_model(self, user, include_servers=False, include_state=False):
         """Get the JSON model for a User object"""
         if isinstance(user, orm.User):
             user = self.users[user.id]
@@ -164,7 +165,7 @@ class APIHandler(BaseHandler):
             # include 'active' servers, not just ready
             # (this includes pending events)
             if spawner.active:
-                servers[name] = self.server_model(spawner)
+                servers[name] = self.server_model(spawner, include_state=include_state)
         return model
 
     def group_model(self, group):
--- a/jupyterhub/tests/test_api.py
+++ b/jupyterhub/tests/test_api.py
@@ -207,6 +207,9 @@ def normalize_user(user):
             for key in ('started', 'last_activity'):
                 server[key] = normalize_timestamp(server[key])
             server['progress_url'] = re.sub(r'.*/hub/api', 'PREFIX/hub/api', server['progress_url'])
+            if (isinstance(server['state'], dict)
+                and isinstance(server['state'].get('pid', None), int)):
+                server['state']['pid'] = 0
     return user
 
 def fill_user(model):
--- a/jupyterhub/tests/test_named_servers.py
+++ b/jupyterhub/tests/test_named_servers.py
@@ -43,6 +43,7 @@ def test_default_server(app, named_servers):
                 'pending': None,
                 'ready': True,
                 'progress_url': 'PREFIX/hub/api/users/{}/server/progress'.format(username),
+                'state': {'pid': 0},
             },
         },
     })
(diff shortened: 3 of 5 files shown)


jupyterhub/tests/test_api.py
```diff
@@ -247,6 +247,7 @@ def test_get_users(app):
         fill_user({
             'name': 'admin',
             'admin': True,
+            'server_state': {},
```
rkdarst: I don't yet know expected behavior, when this should be empty and when None.

jupyterhub/tests/test_named_servers.py
```diff
@@ -43,6 +43,7 @@ def test_default_server(app, named_servers):
                 'url': user.url,
                 'pending': None,
                 'progress_url': 'PREFIX/hub/api/users/{}/server/progress'.format(username),
+                'state': {},
```
rkdarst: In this file some of these have the pid in the tests.  Needs to be ignored or matched somehow - in the right places.

minrk:
Thanks for the PR! This prompted me to open #1817 because I want to stop adding server fields to the top-level user model.

I think the principle is sound, I just have some comments about implementation.

jupyterhub/apihandlers/base.py
```diff
@@ -92,10 +92,12 @@ def write_error(self, status_code, **kwargs):
 
     def server_model(self, spawner):
         """Get the JSON model for a Spawner"""
+        is_admin = self.get_current_user().admin
```
minrk: Let's move this to a False-by-default boolean flag, and pass this as-appropriate on Handlers. See #1817 for an example of such a flag.

jupyterhub/apihandlers/base.py
```diff
@@ -112,6 +114,7 @@ def user_model(self, user):
             'admin': user.admin,
             'groups': [ g.name for g in user.groups ],
             'server': user.url if user.running else None,
+            'server_state': None,
```
minrk: After #1817, we shouldn't add more fields to the user model, only the server.state field.

rkdarst:
Made a quick amend-update implementing your suggestions.  Would a `ApiHandler` or `BaseHandler` attribute to determine if the requestor is an admin be nice, because it seems like there is a lot of passing of flags now.

I also looked two lines above and saw `normalize_user`, so now the tests pass.  You should probably finish #1817 first, then I will fix tests to match those changes, then this can finalize. (one travil failure but it seems unrelated... will check later.)

minrk:
@rkdarst I agree that I should finish #1817 first. I think only the handlers that opt-in to including the server models should also opt-in to the state, so maybe we don't need a *new* flag, and can rely on the one already added there.

Part of the point of that change is that most requests for a user model actually *shouldn't* return the user's server state. Only the user-list and get user by name handlers.

Then the only question remains: should users be able to get their own spawner.state. If not, then we'll need to ensure that server.state is excluded from the one case: get user model and requester is not an admin. This could be done via the flag as here, our cleared-out after the fact:

```python
model = self.user_model(user, include_servers=True)
if not self.get_current_user().admin:
    [ s.pop('state') for s in model['servers'].values() ]
```

Since there's only one handler where this is what we want, maybe that's simpler.

minrk:
I think #1817 is finished, just pending review. So if you'd like to rebase on that to give this a try, go ahead. It's fine to wait for that one to be merged as well.

rkdarst:
Rebased onto #1817 and tests pass.

I minimized the number of places that can get the state.  Including it by default and requiring "insecure" methods to remove it seemed a bit weird, so I didn't use your suggestion... yet.  Instead I renamed the flag to `internal` with the idea that the you can opt-in to the state by saying this is only for internal, not public, use (currently equivalent to how `admin_requestor` was).  I think it's still sort of ugly and inelegant, but...

I've also tried to add it to the REST api docs.

jupyterhub/apihandlers/base.py
```diff
@@ -90,14 +90,15 @@ def write_error(self, status_code, **kwargs):
             'message': message or status_message,
         }))
 
-    def server_model(self, spawner):
+    def server_model(self, spawner, internal=False):
```
minrk: Can we call this `include_state`?

rkdarst:
Updated.  Strangely enough, I saw four recent comments, 3 of which were for outdated code.  All should be handled, I think.

Locally, the test `test_app.py:test_load_groups` failed with a message `SystemExit: Found database schema version 99a28a4418e1 != d68c98b66cd4. Backup your database and run `jupyterhub upgrade-db` to upgrade to the latest schema.`.  It's unrelated and somehow travis doesn't fail, so if this keeps happening I'll make a new issue but this is just an advance warning.

minrk:
Nice, thanks!

minrk:
I think the load_groups is something where some of the tests improperly look in your CWD when running, which they shouldn't. It's not related to this PR.

PR #1817, opened 2018-04-20
Always include server sub-models in user api requests

minrk:
Rather than only when named servers are enabled.

New server fields (started, ready, pending_url) introduced in 0.9 are exclusively in the new user.servers models, rather than duplicated to the top level. Only the pre-existing fields (pending, server.url as user.server, last_activity) are kept at the top level.

- Makes user structure less sensitive to configuration (always have user.servers, not just in certain situations)
- stops adding server-specific state from user model
- Adds explicit 'ready' boolean field in server models, deprecating the current implicit `bool(url)` used to determine if a server is running.

### DIFF
--- a/examples/cull-idle/cull_idle_servers.py
+++ b/examples/cull-idle/cull_idle_servers.py
@@ -127,6 +127,19 @@ def cull_idle(url, api_token, inactive_limit, cull_users=False, max_age=0, concu
                 log_name, server['pending'])
             return False
 
+        # jupyterhub < 0.9 defined 'server.url' once the server was ready
+        # as an *implicit* signal that the server was ready.
+        # 0.9 adds a dedicated, explicit 'ready' field.
+        # By current (0.9) definitions, servers that have no pending
+        # events and are not ready shouldn't be in the model,
+        # but let's check just to be safe.
+
+        if not server.get('ready', bool(server['url'])):
+            app_log.warning(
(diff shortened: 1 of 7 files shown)


examples/cull-idle/cull_idle_servers.py
```diff
@@ -127,6 +127,19 @@ def handle_server(user, server_name, server):
...
             return False
 
+        # jupyterhub < 0.9 defined 'server.url' once the server was ready
+        # as an *implicit* signal that the server was ready.
+        # 0.9 adds a dedicated, explicit 'ready' field.
+        # By current (0.9) definitions, servers that have no pending
+        # events and are not ready shouldn't be in the model,
+        # but let's check just to be safe.
+
+        if not server.get('ready', bool(server['url'])):
```
willingc: We may wish to be more explicit with the warning, but we can do that if folks experience in beta.

jupyterhub/spawner.py
```diff
@@ -81,7 +81,7 @@ def pending(self):
             return 'spawn'
         elif self._stop_pending:
             return 'stop'
-        return False
+        return None
```
willingc: I'll update the docstring in a separate PR.

willingc:
@mirnk would you mind rebasing and merge?

minrk:
Rebased. Thanks @willingc!

PR #1819, opened 2018-04-23
test tornado 4.5 on travis

minrk:
in case we are breaking things with our asyncio changes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,8 +30,13 @@ before_install:
       pip install psycopg2
     fi
 install:
-  - pip install -U pip
+  - pip install --upgrade pip
   - pip install --pre -r dev-requirements.txt .
+  - |
+    # pin tornado
+    if [[ ! -z "$TORNADO" ]]; then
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -30,8 +30,13 @@ before_install:
       pip install psycopg2
     fi
 install:
-  - pip install -U pip
+  - pip install --upgrade pip
   - pip install --pre -r dev-requirements.txt .
+  - |
+    # pin tornado
+    if [[ ! -z "$TORNADO" ]]; then
+      pip install --pre tornado==$TORNADO
```
betatim: so `pip install --pre tornado==` works for the cases where we don't set `$TORNADO`? Every day is a school day!

minrk: Due to the `-z` check above, this install doesn't run if $TORNADO is undefined.

betatim: Thanks, apparently I had my eyes closed when reading this.

PR #1820, opened 2018-04-23
test database upgrades

### LINKED ISSUE #1796: Running jupyterhub upgrade-db with PostgreSQL database fails

Run `jupyterhub upgrade-db` with a PostgreSQL database to upgrade to 99a28a4418e1.

Successful schema update.

**What actually happens**

It fails with an sqlalchemy `ProgrammingError` message that originates here:

https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/alembic/versions/99a28a4418e1_user_created.py#L40

in particular I think that should be `IS NOT NULL` not just `NOT NULL`.  I substituted this live and it allowed the upgrade to proceed.

Latest master.

### LINKED ISSUE #1799: test populated database migration from 0.8 to 0.9
We have tests for database migration, but I think we don't have a test running a fully populated 0.8 database being upgraded to 0.9, which should catch issues like #1796. We only have a long-jump upgrade from 0.5 or so, which doesn't test much, since tables that are new since then don't get upgraded, they get created on first start.

minrk:
We didn't have great coverage of upgrades, as we only upgraded from 0.7.x for postgres/mysql and from 0.6.x for sqlite. This didn't catch bugs in the new 0.8->0.9 upgrades.

Refactors how we run the database upgrade tests a bit:

- move virtualenv creation for old jupyterhub to the test itself
  for easier parameterization
- populate the database as if there are running servers, to better catch issues with existing objects
- test upgrades from 0.7.2 and 0.8.1
- test all versions with mysql, sqlite, postgres
- fix a couple of bugs in upgrades revealed by this

closes #1799
closes #1796

### DIFF
--- a/jupyterhub/tests/test_db.py
+++ b/jupyterhub/tests/test_db.py
@@ -24,7 +24,12 @@ def generate_old_db(env_dir, hub_version, db_url):
     env_pip = os.path.join(env_dir, 'bin', 'pip')
     env_py = os.path.join(env_dir, 'bin', 'python')
     check_call([sys.executable, '-m', 'virtualenv', env_dir])
-    check_call([env_pip, 'install', 'jupyterhub==' + hub_version])
+    pkgs = ['jupyterhub==' + hub_version]
+    if 'mysql' in db_url:
+        pkgs.append('mysql-connector<2.2')
+    elif 'postgres' in db_url:
+        pkgs.append('psycopg2')
+    check_call([env_pip, 'install'] + pkgs)
     check_call([env_py, populate_db, db_url])
 
 


.travis.yml
```diff
@@ -37,22 +39,9 @@ install:
...
-      # use virtualenv instead of venv because venv doesn't work here
-      python -m pip install virtualenv
-      python -m virtualenv old-hub-env
-      ./old-hub-env/bin/python -m pip install jupyterhub==0.7.2 psycopg2 'mysql-connector<2.2'
-      ./old-hub-env/bin/jupyterhub token kaylee \
-        --JupyterHub.db_url=$JUPYTERHUB_TEST_UPGRADE_DB_URL \
-        --Authenticator.whitelist="{'kaylee'}" \
-        --JupyterHub.authenticator_class=jupyterhub.auth.Authenticator
-    fi
-  - pytest -v --maxfail=2 --cov=jupyterhub jupyterhub/tests
```
willingc: Nice deletion.

jupyterhub/tests/test_db.py
```diff
@@ -1,45 +1,70 @@
...
-    upgrade(db_url)
+def generate_old_db(env_dir, hub_version, db_url):
+    """Generate an old jupyterhub database
 
+    Installs a particular jupyterhub version in a virtualenv
+    and runs populate_db.py to populate a database
+    """
+    env_pip = os.path.join(env_dir, 'bin', 'pip')
+    env_py = os.path.join(env_dir, 'bin', 'python')
+    check_call([sys.executable, '-m', 'virtualenv', env_dir])
```
willingc: Can this be 'venv' instead of 'virtualenv'? Might alleviate the need to install virtualenv in requirements

willingc: Test failed locally when running using virtualenv. Test passed when using venv. I'm running all in python36 environment.

minrk: It failed before with venv, I can't recall why. Maybe missing `ensurepip`? `virtualenv` has been added to the test requirements. Have you re-run `pip install -r dev-requirements.txt` from this PR?

PR #1828, opened 2018-04-26
Authenthicator blacklist

rkdarst:
I was looking for low-threshold useful stuff to do and improved #1739: added the test and addressed the comment.  I haven't tested in production, and probably won't in the near future unless someone requests it.

- Closes: #1739

### DIFF
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -274,13 +274,13 @@ class Authenticator(LoggingConfigurable):
             return
 
         blacklist_pass = await maybe_future(self.check_blacklist(username))
+        whitelist_pass = await maybe_future(self.check_whitelist(username))
         if blacklist_pass:
             pass
         else:
             self.log.warning("User %r in blacklist. Stop authentication", username)
             return
 
-        whitelist_pass = await maybe_future(self.check_whitelist(username))
         if whitelist_pass:
             return authenticated
         else:
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -88,6 +88,20 @@ class Authenticator(LoggingConfigurable):
         """
     ).tag(config=True)
 
+    blacklist = Set(
+        help="""
+        Blacklist of usernames that are not allowed to log in.
+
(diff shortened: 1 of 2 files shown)


jupyterhub/tests/test_auth.py
```diff
@@ -103,6 +103,30 @@ def getgrnam(name):
...
+    assert authorized['name'] == 'wash'
+
+    authenticator = MockPAMAuthenticator(blacklist={'wash'})
+    authorized = yield authenticator.get_authenticated_user(None, {
+        'username': 'wash',
+        'password': 'wash',
+    })
+    assert authorized is  None
+
+    authenticator = MockPAMAuthenticator(blacklist={'wash'}, whitelist={'wash', 'kaylee'})
```
minrk: Maybe remove 'wash' from whitelist. I don't want a user being in both whitelist and blacklist to be supported.

jupyterhub/auth.py
```diff
@@ -88,6 +88,18 @@ class Authenticator(LoggingConfigurable):
...
+    blacklist = Set(
+        help="""
+        Blacklist of usernames that are not allowed to log in.
+
+        Use this with supported authenticators to restrict which users can not log in. This is an
+        additional blacklist that further restricts users, beyond whatever restrictions the
+        authenticator has in place.
+
+        If empty, does not perform any additional restriction.
+        """
```
minrk: Can you add 

    .. versionadded: 0.9

to the end of the docstring here and in the new method ?

jupyterhub/tests/test_auth.py
rkdarst: My thought was shouldn't be supported, but it must either default-deny (tested for) or raise an error (not tested)...

minrk: OK

willingc: We might consider (doesn't need to be in this PR) testing several combinations for expected behavior: user not in either list; user in one or the other; user in both

minrk:
Great!

willingc:
@rkdarst Thanks :cake:

Issue #1825, opened 2018-04-26
Multiple syntax errors in cull_idle_servers.py


andreat:
Hi all,

I am trying to use cull_idle_servers.py to cull idle servers. However jupyterhub complains when I try to use it as service, giving multiple syntax errors. I have been able to remove some of them manually, but as soon as "yield" is involved, I have no idea how to proceed.

Note that these errors occur also when cull_idle_servers.py is executed from command line, so this seems to be not really depending on jupyterhub. I am not expert in python; since I am not sure whether the changes I made are correct, I decided to don't make pull requests for them.

In the following the errors I am getting and how I have tried to solve them. 

```
python cull_idle_servers.py
  File "cull_idle_servers.py", line 58
SyntaxError: Non-ASCII character '\xc3' in file cull_idle_servers.py on line 58, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details
```
Solved by replacing `naïve` with `naive`

```
python cull_idle_servers.py
  File "cull_idle_servers.py", line 78
    return f"{h:02}:{m:02}:{seconds:02}"
                                       ^
SyntaxError: invalid syntax
```
Solved by removing `f`

```
python cull_idle_servers.py
  File "cull_idle_servers.py", line 104
    return (yield client.fetch(req))
SyntaxError: 'return' with argument inside generator
```
Solved(?) by splitting `return (yield client.fetch(req))` in two lines: the first for `yield client.fetch(req)` and the second for `return`

```
python cull_idle_servers.py
  File "cull_idle_servers.py", line 189
    resp = yield fetch(req)
SyntaxError: 'return' with argument inside generator
```
At this point I stopped as I have no idea how to proceed.

**Share what version of JupyterHub you are using**

Running `jupyter troubleshoot` from the command line, if possible, and posting
its output would also be helpful.

```
$PATH:
	/usr/local/bin
	/usr/local/sbin
	/usr/local/bin
	/usr/sbin
	/usr/bin
	/sbin
	/bin
	/usr/games
	/usr/local/games

sys.path:
	/usr/local/bin
	/usr/lib/python35.zip
	/usr/lib/python3.5
	/usr/lib/python3.5/plat-x86_64-linux-gnu
	/usr/lib/python3.5/lib-dynload
	/usr/local/lib/python3.5/dist-packages
	/usr/lib/python3/dist-packages

sys.executable:
	/usr/bin/python3

sys.version:
[...]

PR #1829, opened 2018-04-26
cull_idle_servers.py: py<3.6 compatibility

rkdarst:
Remove f-string.

Works on py3.6 at least... not otherwise tested.

### DIFF
--- a/examples/cull-idle/cull_idle_servers.py
+++ b/examples/cull-idle/cull_idle_servers.py
@@ -75,7 +75,7 @@ def format_td(td):
     seconds = seconds % 3600
     m = seconds // 60
     seconds = seconds % 60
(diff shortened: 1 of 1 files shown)


minrk:
👍 anything else that you want to do with this as a WIP?

rkdarst:
Oops.  I was thinking abut figuring out if there was anything else, but now I doubt I'll get around to it... I think nothing more unless someone knows other incompatibilities.

minrk:
Thanks! Would you like to submit another PR specifying `python3 cull_idle_servers...` in the various places we have `python cull_idle_servers...`?  I think it's in the README, docstring, and jupyterhub_config.py in the example dir. Then I think we can close #1825

Issue #1708, opened 2018-03-08
cull_users with a different timeout


gsemet:
If cull_users is given to cull_idle_servers.py, it basically use the same timeout value. I am wondering if a different timeout can be meaningful. For example, after 1 hours, unused servers are closed, after 7 days, innactive users get deleted.

rkdarst:
If you are interested in this, can you comment on pull request #1834 ?  We are trying to see if this is still useful or excess complexity.

willingc:
Closing as #1834 and #1841 have been merged. 👍

Issue #1791, opened 2018-04-12
Make culler installable


minrk:
Summarizing from #1779:

We currently have cull_idle_servers.py as an *example* of consuming the Hub API. But this is common enough functionality that lots of (most?) deployments, including zero-to-jupyterhub, are using it. Since it's only an example script, this means copying the script into your deployment, which is not great.

Our main options are making culling a feature of JupyterHub itself and making the cull-idle-servers service a standalone package. Pros and cons

1. make it part of JupyterHub
   - pro: distributed with jupyterhub, always compatible with jupyterhub
   - pro: could be internal, not a separate process
   - con: cannot distribute fixes or improvements without making jupyterhub releases. We have regularly wanted this.
   - con: cannot be easily modified to provide alternate cull strategies (e.g. additional sources of info)
2. make it a standalone package
   - pro: easy to reuse in multiple places
   - con: another package to maintain/distribute

minrk:
Making it a package would also facilitate testing, which would be nice since this is used in production ~everywhere

GeorgianaElena:
The idle-culler is now a standalone package: https://github.com/jupyterhub/jupyterhub-idle-culler
Thanks @yuvipanda for taking care of this :heart: :sparkles:

Closing this :tada:

PR #1834, opened 2018-04-26
cull_idle_servers: Include a different user timeout

rkdarst:
More low-hanging fruit, not exactly needed for me.  I took inspiration to check on #1708, asking for a different user inactive timeout.  While doing this, I realized there are all kinds of weird interactions between server and user max ages, e.g. if a server keeps running, user max age is never used.  Should users even have a max age, or only inactive limit?  Should admin users be culled?  I thought this would be fast but little things kept coming.  I have a feeling the culler is borderline too complex and needs rethinking.

- The --user-timeout specifies a different inactive time for culling users.
- The --user-max-age specifies a different max age for culling users.
- Don't cull admin users.
- Closes: #1708

### DIFF
--- a/examples/cull-idle/cull_idle_servers.py
+++ b/examples/cull-idle/cull_idle_servers.py
@@ -79,7 +79,7 @@ def format_td(td):
 
 
 @coroutine
-def cull_idle(url, api_token, inactive_limit, cull_users=False, max_age=0, concurrency=10):
+def cull_idle(url, api_token, inactive_limit, cull_users=False, max_age=0, concurrency=10, user_inactive_limit=None, user_max_age=None):
     """Shutdown idle single-user servers
 
     If cull_users, inactive *users* will be deleted as well.
@@ -110,6 +110,10 @@ def cull_idle(url, api_token, inactive_limit, cull_users=False, max_age=0, concu
     resp = yield fetch(req)
     users = json.loads(resp.body.decode('utf8', 'replace'))
     futures = []
+    if not user_inactive_limit:
+        user_inactive_limit = inactive_limit
+    if not user_max_age:
+        user_max_age = max_age
 
     @coroutine
     def handle_server(user, server_name, server):
@@ -159,10 +163,6 @@ def cull_idle(url, api_token, inactive_limit, cull_users=False, max_age=0, concu
 
         should_cull = (inactive is not None and
                        inactive.total_seconds() >= inactive_limit)
-        if should_cull:
-            app_log.info(
-                "Culling server %s (inactive for %s)",
-                log_name, format_td(inactive))
 
         if max_age and not should_cull:
             # only check started if max_age is specified
@@ -176,10 +176,13 @@ def cull_idle(url, api_token, inactive_limit, cull_users=False, max_age=0, concu
 
         if not should_cull:
             app_log.debug(
-                "Not culling server %s (age: %s, inactive for %s)",
+                "Not culling server %s (age: %s, inactive: %s)",
                 log_name, format_td(age), format_td(inactive))
             return False
 
+        app_log.debug(
+            "Culling server %s (age: %s, inactive for %s)",
+            log_name, format_td(age), format_td(inactive))
         req = HTTPRequest(
             url=url + '/users/%s/server' % quote(user['name']),
             method='DELETE',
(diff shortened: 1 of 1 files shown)


examples/cull-idle/cull_idle_servers.py
```diff
@@ -254,28 +263,27 @@ def handle_user(user):
...
-        if should_cull:
-            app_log.info(
-                "Culling user %s (inactive for %s)",
-                user['name'], inactive)
+                       inactive.total_seconds() >= user_inactive_limit)
 
-        if max_age and not should_cull:
-            # only check created if max_age is specified
+        if user_max_age and not should_cull:
+            # only check created if max_user_age is specified
```
willingc: A small nit in the comment: user_max_age

examples/cull-idle/cull_idle_servers.py
```diff
@@ -254,28 +263,27 @@ def handle_user(user):
...
             # so that we can still be compatible with jupyterhub 0.8
             # which doesn't define the 'started' field
-            if age is not None and age.total_seconds() >= max_age:
+            if age is not None and age.total_seconds() >= user_max_age:
                 app_log.info(
-                    "Culling user %s (age: %s, inactive for %s)",
-                    user['name'], format_td(age), format_td(inactive))
+                    "Culling user %s (age: %s, inactive for %s vs user_max_age %s)",
```
willingc: I would remove `vs` since it confuses Perhaps `"Culling user %s (age: %s, inactive for: %s, user_max_age limit: %s)"`

examples/cull-idle/cull_idle_servers.py
```diff
@@ -312,6 +320,11 @@ def handle_user(user):
            help="""Cull users in addition to servers.
                 This is for use in temporary-user cases such as tmpnb.""",
            )
+    define('user_timeout', default=0, type=int,
+           help="""Inactive time for culling users.  Defaults to --timeout.""",
+           )
+    define('user_max_age', default=0,
+           help="The maximum age (in seconds) of users that should be culled even if they are active.  Defaults to --max-age.")
```
willingc: Perhaps: The age limit (in seconds) where users, active and inactive, will be culled if user's age exceeds the limit. Defaults to --max_age

rkdarst:
@minrk, yeah, I really don't know.

Thinking quickly, it could be: potentially large set of users (public github oauth login) with some sort of persistent data.  So, if same person logs in, they can see their old stuff, but they might stop using the service forever.  In this case, different inactive timeout could be useful.  Then, I had to make a different max_age because otherwise max_age wouldn't be usable in this case.  (something else would have to delete persistent storage...)

In my own service, it could be useful to delete users that haven't logged in in 6 months just to keep the database tidy.  But that's not *really* needed, and maybe the culler isn't the right place.

Anyway, I won't put much more work in until someone thinks this is useful - it's already taken more time than I expected.

minrk:
Yeah, I can imagine that in theory. I think it's fine to merge this as it is, but I wouldn't put too much more work into it until someone has a demonstrated rather than theoretical use case.

gsemet:
Hello ! Thanks for this PR, that would be perfect. We have a set of user base that want to use jupyter sporadically, so the instance should be shutdown after for instance 6h of inactivity.
But the persistant volume should stay if the user goes on vacation for some days or even weeks.
But after a while, say 3 months, if he never logged in, we can delete its persistant volume safely. I would like to send a warning email first, but that s completely out of jupyter scope.

betatim:
How about creating a new service that deals with cleaning up users from the database instead of adding this on to the idle server culler?

As far as I am concerned there is no downside to having a second service that runs once a day or even less often to take care of users. It would also keep the scripts cleaner because you don't have to juggle executing at the shortest timeout period but only doing the infrequent tasks every N iterations.

rkdarst:
@betatim, that is genius.  In fact, no changes are needed: run the same, existing culler now, one with short timeouts and no `--cull-users`, and a separate one with longer and `--cull-users` enabled.  Is there anything this can't do?

If so, instead I'll adjust the docs to mention this and we can call it done enough.

betatim:
> Will this “cleaning from database” culling also kill the PVC (I think it should)?

PVCs are not something JupyterHub should deal with as a hub doesn't have to use k8s. Not sure where to place a service that assumes you use k8s should go. I'd probably use #1791 as a reason to start a package that contains "useful JupyterHub services" some of which assume you are using k8s (clean up PVCs) and some that always work (current culler).

rkdarst:
#1841 adds a line to the script docstring about running the same culler multiple times and will close this PR.  I think there's no need for this code anymore.  Comment if otherwise.

PR #1835, opened 2018-04-26
further scrub potentially sensitive url params

minrk:
include oauth codes, state, and tokens in things we avoid logging


willingc:
@minrk I've been testing this locally and have run into some errors with the active_server_limit and start_stop_race tests which I believe is due to the pytest-tornado plugin and not the PR.

The code looks good to me.

PR #1840, opened 2018-04-30
allow authenticator to identify current user

gweis:
I have the use case, where I want to use OAuth access tokens to identify the current user for an API request, when using oauthenticator.

Would something like this be useful?

### DIFF
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -234,6 +234,11 @@ class Authenticator(LoggingConfigurable):
             return True
         return username not in self.blacklist
 
+    def get_current_user(self, handler):
+        """Return username for current request.
+        """
+        return None
+
     async def get_authenticated_user(self, handler, data):
         """Authenticate the user who is attempting to log in
 
@@ -590,8 +595,8 @@ class PAMAuthenticator(LocalAuthenticator):
         help="""
         Whether to check the user's account status via PAM during authentication.
 
-        The PAM account stack performs non-authentication based account 
-        management. It is typically used to restrict/permit access to a 
+        The PAM account stack performs non-authentication based account
+        management. It is typically used to restrict/permit access to a
         service and this step is needed to access the host's user access control.
 
         Disabling this can be dangerous as authenticated but unauthorized users may
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -258,6 +258,13 @@ class BaseHandler(RequestHandler):
 
     def get_current_user(self):
         """get current username"""
+        user = self.authenticator.get_current_user(self)
+        if user is not None:
+            user = self.find_user(user)
+        if user is not None:
+            user.last_activiy = datetime.utcnow()
+            self.db.commit()
+            return user
         user = self.get_current_user_token()
         if user is not None:
             return user


minrk:
I think we do want something like this, but I would like to confine it to the login process. I'm reluctant to allow Authenticators to override all user identification across the board, which can mess with cookies and the like. Instead, I think it should only be able to do this at login time, i.e. if JupyterHub can't identify the user they are redirected to LoginHandler, which gets called. Right now, this may be possible by defining `auto_login = True` and implementing what you have in `.authenticate`, but maybe we can streamline this process.

What do you think?

gweis:
Certainly agree. It shouldn't be something that is being used everywhere.

Let me explain the setup here. We have a running Keycloak instance, that does all our OAuth stuff. Basic user login to jupyter hub via OAuthenticatcator and Keycloak works really well (even bringing in groups / scopes from keycloak is fairly straight forward)

What I do want to do now, is to allow another website (in our setup on a different domain) to talk to the jupyter hub API. This Webapp is already an OAuth client to keycloak. To let this web app talk to the jupyterhub API I would need to register it as OAuthClient in jupyterhub. And then a user has to go through a double OAuth dance (Keycloak and hub + keycloak) to login, and then to allow API access. That seems to be a bit convoluted.

What I'd like to do is use the Keycloak as token source, and use these OAuth tokens (which are already available and granted) to talk to jupyterhub.

 Essentially, the authenticator (in oauth case), knows everything about the token issuer, and knows how to verify tokens (may even be a JWT). This feature really should not be used by an authenticator that can't identify a user in a secure and reliable way. Not sure how to enforce this though, but I think in this use case it's legitimate to trust our tokens.

Open for any other ideas though. I really could not get my head around the jupyterhub OAuth Provider setup, which in my mind does not really fit here. Jupyterhub is not the source for user accounts, but it manages user / OAuth client trust (which is already done by Keycloak here).

minrk:
That's an interesting use case. Allowing Authenticators to override get_current_user means they can short-circuit things like setting session-ids, etc. which I think they shouldn't have the ability to do.

There's a current limitation of `.get_current_user` that it must be blocking and should thus be as inexpensive as we can make it. Checking upstream with another service in a blocking request on every request is not ideal.

Here's what I'm thinking:

Support Authenticator authenticating users on the request-token endpoint (`POST /users/tokens`). We've had the ability to request tokens with username+password via the previous token creation API, but it wasn't generic or extendable to other authenticators, and I doubt anyone used it. Making it generic so that any Authenticator can hook in would be good, I think. Then we can add support for this in OAuthenticators if given an access token directly, rather than just on oauth code redirect.

This way, you'll have to request and store a Hub token for talking to the Hub, but your Authenticator will allow you to make that request with a keycloak token.

How does that sound?

gweis:
Agreed, that checking with an upstream service in get_current_user is not ideal. That's why I am using JWT access tokens, which can be verified locally :) (Yes, this probably narrows the use case further or opens it up to other JWT users cases like using Auth0?).

So you suggest, to allow using access tokens to allow minting new API tokens only? Yes that would certainly works. My only concern here is, that this may create a lot of API tokens. (Esp. if the Webapp is begin used fro multiple devices / browsers, and also depends on local storage time / availability.) May not be a problem, but I have now idea about how, when (or even if), API tokens get cleared in jupyterhub.

To extend this idea. I guess it might even be possible to auto approve a jupyterhub oauth client in a similar way. Not sure if this would help here (haven't thought it through). And I think this would require a change in the REST API as well.

minrk:
> isn't this exactly what #1859 does?

Forgot to auto-link the issues, but yes, this is what #1859 is for. It won't work with OAuthenticator today, but my plan is to add support for this in oauthenticator, which should amount to checking the `data` argument for a `token` field rather than the `code` url param. Then it should work. The Authenticate would look something like this:

```python
def get_token(handler, data):
    if 'token' in data and self.allow_token_auth:
        return data['token']
    else:
        get_token_for_code_via_oauth(code)

def authenticate(handler, data):
    token = get_token(handler, data)
    get_user_for_token(token)
```

instead of unconditionally getting the token from the code. Then you'll be able to request JupyterHub tokens via the API with an upstream access token. This would have to be opt-in because it is a significant change in security policy to allow anything that can authenticate as you with the upstream service (e.g. with a GitHub OAuth token) to authenticate with JupyterHub.

 In the same way that adding `.get_current_user` would support this, so would implementing support in `.authenticate`. And doing it in `.authenticate` has the benefit of confining it to a specific permission, rather than allowing Authenticators to take over all authentication, which has various consequences.

It may make sense to allow `.get_current_user`-type overrides on API-handlers only, since those authentications do not have things like cookie and session side effects. It would mean that you could use this auth to talk to the Hub itself, but you would not then be able to use the same thing to talk to a running notebook server because the Authenticator is not available in that environment. Only true API tokens can be used there.

> My only concern here is, that this may create a lot of API tokens.

Indeed it could, and this is a valid concern. API Tokens are manually managed, so you could easily end up with a proliferation of tokens if you aren't storing and reusing them. Similarly, there is no good way to revoke/replace prior tokens as part of this process. The upside is that 0.9 adds APIs and UI for managing tokens, but that's not a full consolation.

> To extend this idea. I guess it might even be possible to auto approve a jupyterhub oauth client in a similar way.

I think we should be very very strict about approving oauth clients, and not allow any shortcuts.

gweis:
Thanks for taking the time to explain things.

Yes, changing the way `.authenticate` handles things in OAuthenticator would help here. I wasn't sure whether this is appropriate though, because `.authenticate` seems to  be meant to do the initial authentication only (given the way it is used within JupyterHub).

Yes, this method does not give you access to a running Notebook server. Did not realise that JupyterHub API tokens can be used to talk to the notebook servers as well. That certainly changes things.

Having the ability to override `.get_current_user` could be useful and would solve my current use case. However, I probably want a way to talk to the Notebook server API as well at a later stage. I'll dig around a bit more, now that I better understand that bit, to find a way to improve user experience.

Thanks

Gerhard

Just some thoughts, how I got here: 
I was thinking along the lines, that any `Authenticator` could implement authentication (identification), in whatever way is appropriate for the 'authentication protocol' in use. E.g. Cookie Auth and Token Auth handling could be moved into their own Authenticator classes. Didn't think about the distinction of web access and API access.

gweis:
Closing this request, to find a solution that better fits with JupyterHub authenticator work.

PR #1841, opened 2018-05-01
cull_idle_servers: Add note about running with different timeouts

rkdarst:
- Instead of creating many options for different timeouts of users and
  servers, just add a note that the whole culler can be run multiple
  times with different options.  See discussion in #1834.
- Closes: #1834

### DIFF
--- a/examples/cull-idle/cull_idle_servers.py
+++ b/examples/cull-idle/cull_idle_servers.py
@@ -24,6 +24,12 @@ Or run it manually by generating an API token and storing it in `JUPYTERHUB_API_
 
     export JUPYTERHUB_API_TOKEN=`jupyterhub token`
     python3 cull_idle_servers.py [--timeout=900] [--url=http://127.0.0.1:8081/hub/api]
+
+This script uses the same ``--timeout`` and ``--max-age`` values for
+culling users and users' servers.  If you want a different value for
+users and servers, you should add this script to the services list
+twice, just with different ``name``s, different values, and one with
+the ``--cull-users`` option.
 """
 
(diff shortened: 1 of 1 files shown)


examples/cull-idle/cull_idle_servers.py
```diff
@@ -24,6 +24,11 @@
 
     export JUPYTERHUB_API_TOKEN=`jupyterhub token`
     python3 cull_idle_servers.py [--timeout=900] [--url=http://127.0.0.1:8081/hub/api]
+
+This script uses the same timeouts for users and users' servers.  If
```
willingc: Hi @rkdarst, I'm a bit confused by this and would like to help improve the comment. 

If I am interpreting correctly, the `timeout` value is used for the culling of users and single user servers.

If you wish to cull users with a different timeout from the single user servers, you would enter a second item in the services list which sets a different `name` and `timeout` and uses `--cull-users`:
```
c.JupyterHub.services = [
         {
             'name': 'cull-idle',
             'admin': True,
             'command': 'python3 cull_idle_servers.py --timeout=3600'.split(),
         },
        {
             'name': 'cull-idle-users',
             'admin': True,
             'command': 'python3 cull_idle_servers.py --timeout=4800 --cull-users'.split(),
         },

     ]
```

Is this what you were thinking?

rkdarst: Yes, that is exactly what it means.  I'm about to push a clarification, I wrote the previous one relatively fast.  For the full story, see the referenced issue, where someone asked for different timeouts for users and servers and someone realized that changes weren't actually needed.

PR #1847, opened 2018-05-02
Update the contributing instructions for running tests.

willingc:
Be more explicit about setting the ASYNC_TEST_TIMEOUT as some tests will fail locally if the 5 sec default is used.


minrk:
Nice!

PR #1848, opened 2018-05-03
Switch docs theme to be consistent with z2jh and binderhub

### LINKED ISSUE #1610: API docs list duplicate attributes
Looks like all the attributes in the API docs are listed twice:

https://jupyterhub.readthedocs.io/en/latest/api/proxy.html#module-jupyterhub.proxy

willingc:
Switching the theme to alabaster with similar settings as z2jh and binderhub makes sense from a maintainability standpoint.

Closes #1610 

[Rendered docs](http://test-jupyterhub.readthedocs.io/en/docs-theme/)


minrk:
Nice! The corner logo looks very low res. Is there a higher-res version?

docs/source/conf.py
```diff
@@ -67,34 +66,39 @@
...
+html_theme = 'alabaster'
 
-#html_theme_options = {}
-#html_theme_path = []
-#html_title = None
-#html_short_title = None
-#html_logo = None
-#html_favicon = None
+html_logo = '_static/images/logo/favicon.ico'
+html_favicon = '_static/images/logo/logo.png'
```
betatim: I think these two lines are exactly the wrong way around :) and what causes the logo in the RH corner be "very low res".

minrk:
Yay!

PR #1850, opened 2018-05-03
Use URLs for config

### LINKED ISSUE #1289: Clarify bind and connect addresses throughout JupyterHub
Currently we use `ip`, `port`, `connect_ip`, `connect_port` for various things. It's a little confusing, and not very clear what is what when reading code.

I propose for 0.9, we do the following:

1. Deprecate `ip` and `port`, `connect_ip` and `connect_port`. They'll still work
2. Turn `ip` and `port` into `bind_address`, which can be ip:port combo, unix domain socket, just ip, etc.
3. Turn `connect_ip` and `connect_port` into `connect_address`, which would do the same
4. Use this across the JupyterHub codebase, clarifying for Proxies and Singleuser servers as well.

### LINKED ISSUE #1446: hub_connect_* assumes HTTP, can't use HTTPS
While trying out 0.8.0rc1, I've hit a roadblock relating to the new `hub_connect_*` settings. On boot I see the message

```
Adding default route for Hub: / => http://...
```

My problem is that my server redirects to HTTPS, but tornado's httpclient does not redirect by default, resulting in
```
tornado.httpclient.HTTPError: HTTP 301: Moved Permanently
```
Services don't start and the browser just reports that the page isn't redirecting properly.

The `http` is comming from `Server.proto` in `objects.py`, but the `proto` trait is not configurable.

As discussed in #1289, I believe `connect_url` should be added.

minrk:
deprecates ip/port config in favor of full URLs

- `JupyterHub.bind_url` for the public-facing URL of the whole thing (sets ip, port, proto, base_url)
- `JupyterHub.hub_bind_url` supports unix sockets as `unix+http://%2Ftmp%2Fhub.sock` (no other URLs support this yet)
- `JupyterHub.hub_connect_url` for the connect URL of the Hub, used by services, spawners, and proxy
- `Spawner.start` may return a URL in addition to `(ip, port)` tuple

Actually deprecated config (with warnings):

- `JupyterHub.hub_connect_port`

Superseded but not deprecated (no warnings, no plans to ever remove):

- `JupyterHub.ip`
- `JupyterHub.port`
- `JupyterHub.hub_ip`
- `JupyterHub.hub_port`
- `JupyterHub.hub_connect_ip` which is still useful for the common (e.g. docker) case of connect and bind IPs being different, but ports being the same. I don't think replacing these with redundant specifications of full URLs is desirable.

Among other things this allows specifying certain connections to be https, which wasn't possible in the ip/port cases.

closes #1289
closes #1287
closes #1446

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -19,8 +19,10 @@ import sys
 from textwrap import dedent
 from urllib.parse import urlparse
 
+
 if sys.version_info[:2] < (3, 3):
     raise ValueError("Python < 3.3 not supported: %s" % sys.version)
+
 
 from dateutil.parser import parse as parse_date
 from jinja2 import Environment, FileSystemLoader, PrefixLoader, ChoiceLoader
@@ -33,6 +35,7 @@ from tornado.log import app_log, access_log, gen_log
 import tornado.options
 from tornado import gen, web
 from tornado.platform.asyncio import AsyncIOMainLoop
+from tornado.netutil import bind_unix_socket
 
 from traitlets import (
     Unicode, Integer, Dict, TraitError, List, Bool, Any,
@@ -421,6 +424,13 @@ class JupyterHub(Application):
         interfaces if the proxy or user servers are in containers or on a different host.
 
         See `hub_connect_ip` for cases where the bind and connect address should differ.
+        """
+    ).tag(config=True)
+
+    hub_socket = Unicode('',
+        help="""Set the tornado application to listen on a unix socket.
+
+        If set, take precedence over the `hub_port` and `hub_ip` settings.
         """
     ).tag(config=True)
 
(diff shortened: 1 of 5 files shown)


jupyterhub/app.py
```diff
@@ -440,14 +484,49 @@ def _deprecated_proxy_api(self, change):
...
+        Spawners, services, and the proxy will use this URL
+        to talk to the Hub.
+
+        Only needs to be specified if the default hub URL is not
+        connectable (e.g. using a unix+http:// bind url).
+
+        .. seealso::
+            JupyterHub.hub_connect_ip
+            JupyterHub.hub_bind_url
+        .. versionadded:: 0.9
```
willingc: small nit. I would add a space between the two directive blocks.

PR #1852, opened 2018-05-03
Attach an info field to the service

### LINKED ISSUE #1851: It would be helpful to be able to add configurable metadata to services
In my use case, we have services by their design have a specific URL, but I'd like to pass along additional information to present a more friendly interface to the end user. In our specific case we have a tool which is associated with a database. It is more efficient at the low level to use an shortened form to refer to the database. However, I want to present to the end user a descriptive name.

My propose solution is to add a "info" trait to the service object (I initially called it metadata and found out that it collides with a property).

So to deploy a service my YAML section of my config.yaml looks like:
```
  services:
    whoami:
      apiToken: < blah>
      url: http://whoami.kubernetes:8080
      admin: false
      info:
        databases:
          my_db: "My Database"
          your_db: "Your Database as saved last January"
```
I've already made the change to jupyterhub in my local copy and will be pushing a pull request shortly.

Thought I'd open an issue in the event there are some greater issues at play here.

summerswallow-whi:
This issue is described in greater detail in #1851

closes #1851

### DIFF
--- a/docs/rest-api.yml
+++ b/docs/rest-api.yml
@@ -689,6 +689,11 @@ definitions:
         description: The command used to start the service (if managed)
         items:
           type: string
+      info:
+        type: object
+        description: |
+          Additional information a deployment can attach to a service.
+          JupyterHub does not use this field.
   Token:
     type: object
     properties:
--- a/jupyterhub/apihandlers/services.py
+++ b/jupyterhub/apihandlers/services.py
@@ -23,6 +23,7 @@ def service_model(service):
         'prefix': service.server.base_url if service.server else '',
         'command': service.command,
         'pid': service.proc.pid if service.proc else 0,
+        'info': service.info
     }
 
 class ServiceListAPIHandler(APIHandler):
--- a/jupyterhub/services/service.py
+++ b/jupyterhub/services/service.py
@@ -175,6 +175,13 @@ class Service(LoggingConfigurable):
         If unspecified, an API token will be generated for managed services.
         """
     ).tag(input=True)
+
+    info = Dict(
+        help="""Provide a place to include miscellaneous information about the service,
+        provided through the configuration
+        """
+    ).tag(input=True)
+
     # Managed service API:
     spawner = Any()
 
--- a/jupyterhub/tests/test_api.py
+++ b/jupyterhub/tests/test_api.py
@@ -1500,6 +1500,7 @@ def test_get_services(app, mockservice_url):
             'pid': mockservice.proc.pid,
             'prefix': mockservice.server.base_url,
             'url': mockservice.url,
+            'info': {},
         }
     }
 
@@ -1526,6 +1527,7 @@ def test_get_service(app, mockservice_url):
         'pid': mockservice.proc.pid,
         'prefix': mockservice.server.base_url,
         'url': mockservice.url,
+        'info': {},
     }
 
     r = yield api_request(app, 'services/%s' % mockservice.name,


minrk:
What does the info field accomplish? It is not used anywhere.

betatim:
From reading the issue I think this is a way to configure the service from the jupyterhub config. Is that the right impression? If it is "config" maybe renaming the property as such helps others discover it.

The thing I am missing/can't find is how the info from this field gets passed to the service?

Not sure what I think about having configuration for services in the hubs config. When I have used services (in a z2jh helm chart deployment) I created a new deployment for each service and gave them access to the configmaps they need or configured them via `values.yaml`.

summerswallow-whi:
To answer the question @betatim, this is about being able to configure via values.yaml but a way to get the configuration of data for the hub's consumption. We are endeavoring to make the hub a portal for all our data services, but there is some information the hub should get so that our custom pages can provide a better user experience.

In my use case, we have services by their design have a specific URL, but I'd like to pass along additional information to present a more friendly interface to the end user. In our specific case we have a tool which is associated with a database. It is more efficient at the low level to use an shortened form to refer to the database. However, I want to present to the end user a descriptive name.

My propose solution is to add a "info" trait to the service object (I initially called it metadata and found out that it collides with a property).

So to deploy a service my YAML section of my config.yaml looks like:

```
  services:
    drn:
      apiToken: < blah>
      url: http://drn.kubernetes:8080
      admin: false
      info:
        databases:
          my_db: "My Database"
          your_db: "Your Database as saved last January"
```

The way the templates are set up for helm. This is already passed to jupyterhub and only requires a trait be added to the service to be included as part of the service.

In our welcome page, we have the following handler.
```
class WelcomeHandler(BaseHandler):
    @gen.coroutine
    def get(self):
        user = self.get_current_user()
        auth_state = yield user.get_auth_state()

	datasets = {k: v for k, v in self.services['drn'].info['databases'].items() if k in auth_state['datasets']}

	data=dict(name=' '.join([auth_state['first_name'],auth_state['last_name']]), datasets=datasets,
                  notebook = user.url)
	html = self.render_template('welcome.html', data=data)
	self.finish(html)
```
So our service is called drn, it has extensively be coded so the url begins with an abbreviated form of the db such as `my_db`. In this case there are two databases associated with this service.  The point is in rendering the html welcome page I will present it like:

select your database "My Database" or "Your Database as saved last January"
this will redirect the user to the correct url for the service.

This is all about making the environment more usable.

minrk:
That's the bit I was missing. That this info is exclusively useful only via and to the additional new features in #1854. So it's not even that it's influencing how *jupyterhub* presents this information, it's making it available only to your own custom extensions.

It doesn't seem right to me to put something in services that services don't use. Since this information is exclusively used by your custom welcome handler, can the information maybe live there, instead?

summerswallow-whi:
I'm open to suggestions, but I guess in sort of the computer science aesthetic point of view, I thought "info" is a "trait" of the service so it seemed to be a natural fit. Plus it falls through from the values.yaml, through jupyterhub_config.py into the Service trait, without any significant code modification. Because the code is already plumbed for it, I thought it would be the appropriate place for it.

Not all traits currently are _used_ by the service. Afaik the service does not use the `url` or `admin` but _jupyterhub_ does.

minrk:
What makes this below the bar for me is that it's used nowhere in JupyterHub *at all*. Since it's only information for your custom extension, placing that information in the extension itself makes the most sense to me.

summerswallow-whi:
I understand where you are coming from. Maybe I'm too old school, but I liken this to the practice of old callbacks where in addition to passing a handler to the callback, the callback also accepts user data whose sole purpose is to be passed to the handler.

Other work arounds I've tossed around are either more convoluted or use unnecessarily uses the network. This seems the simplest and most elegant solution to the problem.

betatim:
To me it also feels a bit "niche" :-/

Having thought about how I'd solve this problem I think I would add a handler to the service that you can hit up to get this metadata or even have the service serve a HTML page that users look at. This appeals to my sense of "pretty" because it keeps all the configuration of the service and how it wants itself presented in one place (the service's config). But it does require you to perform a network request to pull in this info (or cache it once or every few hours).

summerswallow-whi:
@betatim, we had actually considered your approach before, while there is a certain aesthetic to keeping presentation information with the service there are some practical issues. I also apologize ahead of time for being long winded.

But first I should perhaps more clearly articulate our use case. Our target audience are clinical researchers who want to access tools like Jupyter Notebook, R Studio and some of our homegrown tools. Originally the design was to place a landing server in front where Jupyter would be one of many services that accept our authentication and serve up notebooks, the landing server would handle authentication. As we looked into the integration of JupyterHub it turns out JupyterHub encompasses all the features especially with the external services that our landing server would encompass. Also the fewer servers the few places we need to serve. In this portal, all our data science services would go through JupyterHub, it would provide authentication. In my other PR which seems to not be as controversial as this, we needed a way to put a custom  welcome page and terms of service page, for user experience and auditing purposes as we do work with sensitive data. From this welcome page, the user can navigate to their notebook or any one of our external services.

So naturally one solution to the problem is simply hard code the service in my welcome page. Problem is that the welcome page is a custom python code and would require me to build a new docker image, upload it to our container storage. Not a huge deal, but not as easy as changing my custom.yaml and reinstalling via helm.

So if the presentation information is kept in the service (which like you said has a certain aesthetic as far as data ownership), the question is how does the welcome handler retrieve it. Ideally in this situation Hub would gather the information when the service is first registered by the proxy so it would happen once. However, code would have to be written to do this. So you might say let the handler deal with it after all it is just a customization. The welcome page is going to get a lot of traffic, and if there are a lot of  services that's a number of async calls to network operations to render a welcome page. If the services (as many are) involved in heavy numerical calculations, the services might not be so responsive. This could lead to a poor user experience. As you suggested this information could be cached, which would require more complexity (albeit not a huge burden) on the part of the customization. Furthermore, in some cases, the service could be a third party service which makes customizing it to add a handler more work.

If you look at what the proposed change is. It is to add a trait to service to allow user data to be assigned to a service from the config file for JupyterHub. It seems the solution is so elegant in its simplicity. The way the hub code is written, hub will already read the config file and insert it into this trait if it is defined. By adding this user configurable piece of information it also adds flexibility for other applications none of has thought of.

Finally, if the goal is to keep stuff out of Hub, the case could be made to not have external services in the first place after all, you could work around not having external service. I get the feeling  that external services were added to as a convenience to JupyterHub users. I think the next step in the evolution of external services is to allow those implementing the service to have the tools necessary to implement a good user experience.

----
The quick summary is while certainly there are other ways to accomplish this, the propose change provides an elegantly simple solution on the part of JupyterHub and a very convenient way for people like me trying to build an ecosystem around JupyterHub with a pleasant user experience.

summerswallow-whi:
Probably wouldn't make a difference to the reasoning above, but a co-worker suggested that perhaps the trait could be labeled vendor_info in keeping with the tradition in communications and software of having "vendor" specific fields.

minrk:
This will make me feel better about adding the info field: Can you add it to the JSON model for services [here](https://github.com/jupyterhub/jupyterhub/blob/0.9.0b2/jupyterhub/apihandlers/services.py#L17)? That way, at least, it's used somewhere in JupyterHub.

summerswallow:
Sounds reasonable. Kind of mirrors what you do with users so it makes sense.

summerswallow-whi:
Can a dict be added to the services model or does it need to be converted to a string to be added?
I've made a change assuming a dict is permissible.

minrk:
Dict is definitely fine.

PR #1861, opened 2018-05-04
Changelog for 0.9

minrk:
With this and #1859 and #1850 I think we can cut our first beta for 0.9 and switch our focus to documentation and testing.


minrk:
@willingc great! I waffled quite a bit about sections, and ended up going with the same structure we had before. I think grouping by "Spawner" "Authenticator" "REST API" might also be good.

PR #1863, opened 2018-05-05
Add post spawn hook

jamescurtin:
There is a `pre_spawn_hook` that is executed before the spawner starts; this PR implements a corresponding `post_spawn_hook` that is executed after the spawner stops.

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -515,6 +515,15 @@ class Spawner(LoggingConfigurable):
         """
     ).tag(config=True)
 
+    post_spawn_hook = Any(
+        help="""
+        An optional hook function that you can implement to do work after
+        the spawner stops.
+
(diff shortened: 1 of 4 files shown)


jupyterhub/spawner.py
```diff
@@ -515,6 +515,15 @@ def my_hook(spawner):
         """
     ).tag(config=True)
 
+    post_spawn_hook = Any(
```
minrk: Can we call it `post_stop_hook`? I would expect `post_spawn` to be after start finishes (we probably should have called `pre_spawn` `pre_start`).

minrk:
I made a comment about the name of the hook, but this makes sense.

jamescurtin:
👍 I've updated the name of the hook. Thanks for the review!

minrk:
Thanks!

PR #1864, opened 2018-05-05
Add Docs about sudo (and remove it from the wiki)

echarles:
Add Docs about sudo (and remove it from the wiki)


willingc:
Thanks @echarles. I'm going to go ahead and merge.

As an FYI, I'm going to move this entire section under configuration examples as that's a more logical location.

PR #1872, opened 2018-05-07
Allow extra variables to be passed into templates

rschroll:
We make JupyterHub deploys fairly often, and we use custom templates for the JH control panel.  These vary slightly, but predictably, for deploy to deploy.  Right now, we upload slightly different versions to each deploy.  It would be easier if we could parameterize the templates.

This patch adds a `template_vars` config value, which takes a dictionary of variable names and values.  These are added to the variables passed in to every template construction.

One possible worry is that this could overwrite an existing variable name.  In some cases this might be useful as a feature, but that would seem to be pretty rare to me.  One way to avoid this would be to pass the dictionary directly into the template, instead of each of its entries.  This would cause the user to have to use `{{template_vars[var]}}` instead of just `{{var}}`.  I think this is pretty ugly, but if we're worried about this problem, it'd be an easy change.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -287,6 +287,10 @@ class JupyterHub(Application):
     def _template_paths_default(self):
         return [os.path.join(self.data_files_path, 'templates')]
 
+    template_vars = Dict(
+        help="Extra variables to be passed into jinja templates",
+    ).tag(config=True)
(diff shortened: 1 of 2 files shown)


minrk:
I'm not worried about the collisions. Thanks!

PR #1876, opened 2018-05-07
refactor sudo example config

willingc:
Follow up to #1864


docs/source/index.rst
```diff
@@ -52,14 +52,16 @@ Contents
...
-* :doc:`reference/sudo`
 * :doc:`reference/websecurity`
 * :doc:`reference/authenticators`
 * :doc:`reference/spawners`
 * :doc:`reference/services`
 * :doc:`reference/rest`
 * :doc:`reference/upgrading`
 * :doc:`reference/config-examples`
+* :doc:`reference/config-examples`
+* :doc:`reference/config-examples`
```
minrk: reference/config-examples occurs three times here.

willingc: good catch. airplane wifi and refresh

PR #1878, opened 2018-05-08
fix listing of OAuth tokens on tokens page

gweis:
After spawning a new notebook server the token listing page returns with an error 500. The logs show an AttributError with unknown attribute client_id.

orm.OAuthClient does not have an attribute client_id


minrk:
Thanks!

Issue #1879, opened 2018-05-09
Default post_stop_hook logs an exception during shutdown 


paccorsi:
**How to reproduce the issue**
Run Jupyterhub without implementing a `post_stop_hook` in the config.

**What you expected to happen**
A `post_stop_hook` with a value `None` (the default value) should not be called.

**What actually happens**
The default value of `post_stop_hook` is None. When it is called as a function, during shutdown, it logically raises an exception:
```
post_stop_hook failed with exception: object at 0x7fdf90804ef0>
    1   │ Traceback (most recent call last):
    2   │ │ File "/opt/conda/lib/python3.6/site-packages/jupyterhub/spawner.py", line 707, in run_post_stop_hook
    3   │ │ │ return self.post_stop_hook(self)
    4   │ TypeError: 'NoneType' object is not callable
```
This exception is then caught by the try/catch block and logged.

**Share what version of JupyterHub you are using**
Jupyterhub 0.9.0b1

I believe an easy fix would be to check the value of `post_stop_hook`, like it was done with `pre_spawn_hook`: https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/spawner.py#L701

PR #1881, opened 2018-05-10
Check the value of post stop hook

paccorsi:
The issue is detailed in: #1879

This PR checks the value of `post_stop_hook`. If it has the value None, the function call is skipped, thus preventing the logged exception described in the issue.

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -703,10 +703,11 @@ class Spawner(LoggingConfigurable):
 
     def run_post_stop_hook(self):
         """Run the post_stop_hook if defined"""
-        try:
-            return self.post_stop_hook(self)
-        except Exception:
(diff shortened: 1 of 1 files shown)


jupyterhub/spawner.py
```diff
@@ -703,10 +703,11 @@ def run_pre_spawn_hook(self):
 
     def run_post_stop_hook(self):
         """Run the post_stop_hook if defined"""
-        try:
-            return self.post_stop_hook(self)
-        except Exception:
-            self.log.exception("post_stop_hook failed with exception: %s", self)
+        if self.post_stop_hook:
```
betatim: nitpick: can we explicitly check for `None` with `if self.post_stop_hook is not None:`?

dhirschfeld:
FWIW this looks good to me.
:+1: to merging.

PR #1895, opened 2018-05-15
avoid creating one huge transaction cleaning up oauth clients

minrk:
some deployments that create temporary users (e.g. binder) may have a *very* large number of defunct oauth clients left in the database.

Commit at most 100 deletions at a time to avoid executing it all in one massive transaction that could fail.


minrk:
Appears to have been an intermittent network issue on docker cloud

PR #1896, opened 2018-05-16
Fix typos in auth config documentation

thoralf-gutierrez:
The correct env var is JUPYTERHUB_CRYPT_KEY, see https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/crypto.py#L24


betatim:
Thanks for noticing it and then taking the time to fix it!

PR #1899, opened 2018-05-18
ConfigurableHTTPProxy.stop: kill child processes on Windows case

adelcast:
On the Windows case, the configurable-http-proxy is spwaned using a
shell. To stop the proxy, we need to terminate both the main process
(shell) and its child (proxy).

### DIFF
--- a/jupyterhub/proxy.py
+++ b/jupyterhub/proxy.py
@@ -516,13 +516,26 @@ class ConfigurableHTTPProxy(Proxy):
         self._check_running_callback = pc
         pc.start()
 
+    def _kill_proc_tree(self, pid):
+        import psutil
+        parent = psutil.Process(pid)
+        children = parent.children(recursive=True)
+        for child in children:
+            child.kill()
+        psutil.wait_procs(children, timeout=5)
+
     def stop(self):
         self.log.info("Cleaning up proxy[%i]...", self.proxy_process.pid)
         if self._check_running_callback is not None:
             self._check_running_callback.stop()
         if self.proxy_process.poll() is None:
             try:
-                self.proxy_process.terminate()
+                if os.name == 'nt':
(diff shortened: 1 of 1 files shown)


adelcast:
cherry-pick from https://github.com/jupyterhub/jupyterhub/pull/1808. This change is needed to prevent a zombie configurable-http-proxy process to stick around after the hub is shutdown. Proxy.py already has code to launch the proxy using shell=True on the Windows case. This is just a fix to that case.

@dhirschfeld

minrk:
Does Windows have anything like `exec` that replaces the process? Or is there any way that we can get the subprocess spawning to work properly without `shell=True`? I don't love pulling in an additional dependency just for an unsupported platform.

minrk:
Docker build also appears to be failing due to the added dependency

dhirschfeld:
The dep could be a "soft" dep - i.e. import it only inside the function and don't explicitly specify it as a dep. In a conda package it could be specified with a `[win]` selector.

On holiday right now, but it is on my list of things to investigate when I'm back next week...

adelcast:
I also researched another option: call Popen with creationflags=subprocess.CREATE_NEW_PROCESS_GROUP, then terminate via "self.proxy_process.send_signal(signal.CTRL_BREAK_EVENT)". According to the documentation, that should do the trick, but I do see my console sometimes being stuck on "Terminate Batch Job Y/N",  so I am not very confident on the solution. @dhirschfeld I like your soft dependency solution better, let me know what you conclude, to see what's the best way forward.

minrk:
Putting the import in `_kill_proc_tree` makes sense. That should be enough.

adelcast:
sounds good, just posted an updated review

minrk:
Thanks!

PR #1903, opened 2018-05-21
Update README.md

aisensiy:
The default localhost does not work for docker. `0.0.0.0` is necessary to forward docker inside to outside.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ not, Jupyter Notebook version 4 or greater must be installed.
 
 The JupyterHub docker image can be started with the following command:
 
-    docker run -p 8000:8000 -d --name jupyterhub jupyterhub/jupyterhub jupyterhub
+    docker run -p 8000:8000 -d --name jupyterhub jupyterhub/jupyterhub jupyterhub --ip=0.0.0.0
 
 This command will create a container named `jupyterhub` that you can
 **stop and resume** with `docker stop/start`.


willingc:
Hi @aisensiy, Thanks for the PR. Instead of adding the ip for a specific use case, it would be better to leave the command as is, and add a sentence after that directs readers to the documentation http://jupyterhub.readthedocs.io/en/latest/getting-started/networking-basics.html#set-the-proxy-s-ip-address-and-port and http://jupyterhub.readthedocs.io/en/latest/quickstart-docker.html#starting-jupyterhub-with-docker

Thanks.

README.md
```diff
@@ -151,7 +151,7 @@ not, Jupyter Notebook version 4 or greater must be installed.
 
 The JupyterHub docker image can be started with the following command:
 
-    docker run -p 8000:8000 -d --name jupyterhub jupyterhub/jupyterhub jupyterhub
+    docker run -p 8000:8000 -d --name jupyterhub jupyterhub/jupyterhub jupyterhub --ip=0.0.0.0
```
willingc: Please leave command as is. Recommendation is to add a sentence following this command with links to documentation.

aisensiy: So we give a command and then write some sentence to tell our friends the code above does not work if you just want to access it from browser. You do want access the docker through http please bind ip `0.0.0.0`?

This does not give a good experience for newbies. Many people try new things through docker because it will not pollute local env. And the one line docker command should work for them. I am a newbie here but after press the command into my terminal I just confuse for no public port map to localhost. After add the `-p` argument and enter again. I try to open the `localhost:8000` in my browse to see if it work or not and it doesn't. 

Look at command in [rancher](https://rancher.com/quick-start/):

    $ sudo docker run -d --restart=unless-stopped -p 80:80 -p 443:443 rancher/rancher

That is a fully worked command. They do give argument like `-p`. After you copy that into your terminal and press Enter all the staff should work.

minrk: The default command ought to work because the default ip is already to bind on all interfaces. A typo in recently added config changed this to localhost only by default. That is fixed in #1910.

aisensiy: @minrk Yeah, that is good.

Issue #1836, opened 2018-04-27
Add short announcement text in templates


rkdarst:
I was thinking that it would be convenient to have a configure that could include a short announcement in various templates (login, spawn, maybe control panel).  It could be used for things like "maintenance on Tuesday, down 14-16", "new kernels available", and so on.  This can be done more easily with the new template inheritance, but a single way to do something more lightweight would be nice.

- Good idea, or better to just require to use the existing template system.
- Which templates should it render on?
- Should some other template magic be done instead?
- Can it be changed without restarting server (maybe templates do this)?

I started something a few week ago, but switched to template inheritance instead.  Should I fix up and submit or archive?

rabernat:
I was just wondering if something like this existed and I found my way to this issue.

Would be very valuable for large-scale jupyterhub deployments that need to push announcements to users.

rschroll:
Another option would be to host the announcement remotely, as JSON or similar, and use a bit of AJAX to load and display it client-side.  This way you can update the announcement without touching anything on JupyterHub.  The obvious downside is another server in the mix.

PR #1913, opened 2018-05-22
Add customizable announcement text on home,login,logout,spawn

rkdarst:
- Using the new template_vars setting (#1872), allow the variable
  `announcement` to create a header message on all the pages in the
  title, or the variables `announcement_{home,login,logout,spawn}` to
  set variables on these single pages.
- This is not the most powerful method of putting an announcement into
  the templates, because it requires a server restart to change.  But
  the invasiveness is very low, and allows minimal message
  without having to touch the templates themselves.
- Closes: #1836

Usage:
```
c.JupyterHub.template_vars = {'announcement_spawn': 'this is a message',
                              'announcement': 'blah'}
```

This at least needs testing and documentation before it can be used.  More importantly, needs consideration if it is a good idea.  Also some designer could make it prettier and fit better.  This idea could be in other places in the spirit of "make the default templates more customizable".

My big picture opinion is that this, or something like it, would be useful to include, but if people start going deep enough they would instead directly modify templates to do this.  https://jupyterhub.readthedocs.io/en/latest/reference/templates.html#example is the alternative and I think it is clear enough now.

### DIFF
--- a/docs/source/reference/templates.md
+++ b/docs/source/reference/templates.md
@@ -60,12 +60,14 @@ text about the server starting up, place this content in a file named
 {% endblock %}
 ```
 
-## Simple configuration
+## Page Announcements
 
-The most powerful way to control templates is to directly extend them
-as you see above.  However, some simple configuration is possible:
+To add announcements to be displayed on a page, you have two options:
 
-### Announcement text
+- Extend the page templates as described above
+- Use configuration variables
+
+### Announcement Configuration Variables
 
 If you set the configuration variable `JupyterHub.template_vars =
 {'announcement': 'some_text}`, the given `some_text` will be placed on
--- a/docs/source/reference/templates.md
+++ b/docs/source/reference/templates.md
@@ -59,3 +59,33 @@ text about the server starting up, place this content in a file named
 <p>Patience is a virtue.</p>
 {% endblock %}
 ```
+
+## Simple configuration
+
+The most powerful way to control templates is to directly extend them
+as you see above.  However, some simple configuration is possible:
+
+### Announcement text
+
+If you set the configuration variable `JupyterHub.template_vars =
+{'announcement': 'some_text}`, the given `some_text` will be placed on
+the top of all pages.  The more specific variables
+`announcement_login`, `announcement_spawn`, `announcement_home`, and
+`announcement_logout` are more specific and only show on their
+respective pages (overriding the global `announcement` variable).
+Note that changing these varables require a restart, unlike direct
+template extension.
+
+You can get the same effect by extending templates, which allows you
+to update the messages without restarting.  Set
+`c.JupyterHub.template_paths` as mentioned above, and then create a
+template (for example, `login.html`) with:
+
+```html
+{% extends "templates/login.html" %}
+{% set announcement = 'some message' %}
+```
+
+Extending `page.html` puts the message on all pages, but note that
+extending `page.html` take precedence over an extension of a specific
+page (unlike the variable-based approach above).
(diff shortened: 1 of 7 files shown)


share/jupyterhub/templates/home.html
```diff
@@ -2,6 +2,13 @@
 
 {% block main %}
 
+{% if announcement_home or announcement %}
```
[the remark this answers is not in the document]
rkdarst: Nice idea.  I was able to simplify a bit more even and it seems to work (pending tests).  Another advantage is that there is now an `announcement` block which can be inherited separately in other templates - so that option (0) using your own templates becomes even easier.  Option (0) should also allow you to easily change the message without restarting!

rkdarst:
I *think* (untested) these latest changes should allow you to do the following, which is probably the easiest way to set a simple message using a template:

```
{% extends "page.html" %}
{% set announcement = "this is some custom local announcement"%}
```

Before I go forward too much more, does this look good?

share/jupyterhub/templates/page.html
```diff
@@ -140,6 +140,16 @@ <h4 class="modal-title" id="{{key}}-label">{{title}}</h4>
   </nav>
 {% endblock %}
 
+
+{% block announcement %}
+{% if announcement %}
+<div class="container text-center">
```
minrk: Can this get the class 'announcement', too, for easier CSS? Then I think this is good to go.

rkdarst:
I updated the PR.  I started some tests (using the framework in `test_pages.py`), but am having trouble making the tests actually work (the `get_page()` function not doing what I expect).  Should I forget them, or push and you can comment on what is wrong?

rkdarst:
Oh, also let me add documentation first... obviously.

rkdarst:
Re-pushed.  The second commit adds tests, but the tests just don't work - I'm not doing the HTTP simulation correctly somehow.  If you have hints, I can fix them, or I can just remove the tests.

rkdarst:
Finally got around to making the tests pass.  It's a relatively large number of lines for a small change, but maybe this block can be reused for other template tests, too.

Please also see if the docs updates are good - it's late and I should check them once more...

share/jupyterhub/templates/home.html
```diff
@@ -1,4 +1,5 @@
 {% extends "page.html" %}
+{% if announcement_home %}{% set announcement = announcement_home %}{% endif %}
```
willingc: A bit easier to read if these are broken into 3 lines.

jupyterhub/tests/test_pages.py
```diff
@@ -517,3 +517,84 @@ def test_oauth_token_page(app):
 def test_proxy_error(app, error_status):
     r = yield get_page('/error/%i' % error_status, app)
     assert r.status_code == 200
+
+
+
+@pytest.mark.gen_test
+def test_page_contents(app):
```
willingc: Probably more maintainable to break this into multiple tests.

docs/source/reference/templates.md
```diff
@@ -59,3 +59,33 @@ text about the server starting up, place this content in a file named
 <p>Patience is a virtue.</p>
 {% endblock %}
 ```
+
+## Simple configuration
```
willingc:
```
## Page Announcements

To add announcements to be displayed on a page, you have two options:

- Extend the page templates as described above
- Use configuration variables

### Announcement Configuration Variables
```

minrk:
@willingc thanks for the review! I added some changes to address it.

@rkdarst thanks! This is going to probably be the last code change for 0.9.

willingc:
Thanks @rkdarst for all your work on this :tada:

rkdarst:
Thanks, all!  I'm sorry I didn't get back to fixing up the patch - my work took me on a different path the last weeks and I didn't have time to think much.  I have another JupyterHub deployment this summer, so hopefully I can do more then...

Congrats on 0.9 and thanks for all the hard work!

willingc:
Thanks you @rkdarst. Your effort got us 90% there :sunny:

PR #1915, opened 2018-05-24
Clarify error message on implicit respawns.

rkdarst:
- This message is presented when the last spawn failed, along with a
  HTTP 500.  The current text is quite confusing, especially when the
  problem may just be solvable by trying to respawn again.

Note: I'm not sure if this message is presented other places, too, and thus this obvious change may not be sufficient.  This message would have saved me an hour of debugging.

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -40,7 +40,8 @@ reasons = {
     'timeout': "Failed to reach your server."
         "  Please try again later."
         "  Contact admin if the issue persists.",
-    'error': "Failed to start your server.  Please contact admin.",
+    'error': "Failed to start your server on the last attempt.  "
+        "  Please contact admin if the issue persists.",
 }
 
 # constant, not configurable


rkdarst:
Also it would be nice to present the previous error message to the user again, but I couldn't figure out how to do that easily.  If someone can give me a hint, I will.

rkdarst:
Extra bonus: this shouldn't be a HTTP 500, because the server is properly doing what it is supposed to and catching the error.  You can pick your own favorite HTTP status code, and the message can be changed to "Last spawn failed, not respawning automatically".

I may figure this out eventually but I'd be happy for pointers...

minrk:
Thanks!

It should probably be a 5XX because the server failed to take the action that the browser requested it to. It's not an *unhandled* error, but it is a server-side error.

rkdarst:
I still think that 500 is a bit misleading, because it not something that has gone wrong now, but has gone wrong before and we are choosing to not automatically repeat (I see 500 and think "someone messed up programming").  But I can't find a better more specific code except 503, so ...

But... I did realize something more useful.  My spawner has an `options_form`, so I never get an implicit spawn.  So to me, just going to the JupyterHub base_url never did any spawning, so go to base_url and see 500 and thus think there is something wrong with the state deep inside JupyterHub.

This is kind of rambling, but I will make a PR to discuss the second paragraph.

Issue #1747, opened 2018-03-22
Documentation improvements: Getting Started / Security Settings


nxg:
There are a couple of points missing from the [‘Getting Started’ documentation](http://jupyterhub.readthedocs.io/en/latest/getting-started/security-basics.html).  These are each points where explicitness would be useful.

1: In the section ‘Enabling SSL encryption’, note that if the configuration settings `c.JupyterHub.ssl_key` and `c.JupyterHub.ssl_cert` are omitted, then JupyterHub will run without SSL (setting them to `None` does not have the same effect, and is an error).  This could be included either at the top of this section or in the subsection ‘If SSL termination happens outside of the Hub’.  Running without SSL is of course not advised if the proxy is exposed, but is the desired configuration if the server is running behind a reverse proxy.

This behaviour – running without SSL if these are absent – is broadly implied by the documentation here, but documenting this is useful (a) because being explicit is always useful, and (b) because this indicates that this behaviour is intended/supported, and not just an accident.

2: In such a proxied configuration, the JupyterHub proxy is probably exposed on a URL such as `https://example.com/jupyter/`.  In this case, it appears that it is both necessary and sufficient to set

    c.JupyterHub.base_url = '/jupyter/'

It isn't hard to guess this, but, as before, explicit text such as the above is good, and it is useful to have reassurance that this is sufficient configuration for this situation.

3: The current version of JupyterHub (I'm using 0.8.0) can only listen to a single IP address and port at once; users may be familiar with other internet servers which can bind to multiple IP:port combinations.  It would, again, be useful to make this briefly explicit, by adding at the end of section ‘Set the Proxy’s IP address and port’ of [the Networking basics](http://jupyterhub.readthedocs.io/en/latest/getting-started/networking-basics.html), some text such as ‘Note that the values of `c.JupyterHub.ip` and `c.JupyterHub.port` are single string values, rather than arrays – JupyterHub can listen on only a single IP:port.’

minrk:
These sound like great changes! Would you like to make a PR? I think making it explicit that running with SSL via external termination means JupyterHub should have no SSL config is a great addition.

> the JupyterHub proxy is probably exposed on a URL such as `https://example.com/jupyter/`

This doesn't appear to be true in practice (most often, we find JupyterHub's deployed on a whole domain of their own with no prefix), but it is still certainly a case worth documenting.

> The current version of JupyterHub (I'm using 0.8.0) can only listen to a single IP address and port

[...]

PR #1923, opened 2018-05-27
Documentation clarifications (adding explicitness).

nxg:
Addresses issue #1747.

These additions aren't perfect -- it's unfortunate that I've added
mention of reverse proxies on two separate pages.  I don't _think_
these can reasonably be put on the same page -- perhaps a cross
reference?


minrk:
Thanks!

PR #1926, opened 2018-05-28
tilde is a safe character in user URLs

### LINKED ISSUE #1717: Users with some special characters can't connect to their servers with Chrome
Chrome is substituting a few url encoded characters for the literal characters in URLs. The tilde is such a character, so it will substitute `%7E` for `~` in the request. This will make the notebook respond with a 404 to the request. Using another browser (eg Firefox), does not run into the same issue.

Create an user with a tilde in it and attempt to access your server in a jupyterhub deployment with Google Chrome.

Get a 200 response with the notebook server interface.
**What actually happens**
Get a 404 response.
z2jh@0.6 so `jupyterhub==0.8.1`

See https://phabricator.wikimedia.org/T189525

minrk:
Chrome unconditionally reverts any not-strictly-necessary escaping in URLs (this seems wrong?)

closes #1717

### DIFF
--- a/jupyterhub/tests/test_spawner.py
+++ b/jupyterhub/tests/test_spawner.py
@@ -11,6 +11,7 @@ import sys
 import tempfile
 import time
 from unittest import mock
+from urllib.parse import urlparse
 
 import pytest
 from tornado import gen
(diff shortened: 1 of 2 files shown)


betatim:
LGTM. Mostly looked at the test to understand what this is about and that looks solid to me.

PR #1942, opened 2018-06-04
note where nginx config files are typically created.

minrk:


docs/source/reference/config-proxy.md
```diff
@@ -31,7 +31,8 @@ openssl dhparam -out /etc/ssl/certs/dhparam.pem 4096
 ## nginx
 
 The **`nginx` server config file** is fairly standard fare except for the two
-`location` blocks within the `HUB.DOMAIN.TLD` config file:
+`location` blocks within the `HUB.DOMAIN.TLD` config file,
```
willingc: It would probably be helpful for minimizing confusion to clarify the filename of the typical nginx config file from the nginx docs or recommend `jupyterhub.conf`.

From nginx docs:

> By default, the configuration file is named nginx.conf and placed in the directory /usr/local/nginx/conf, /etc/nginx, or /usr/local/etc/nginx.

minrk: I specifically want to avoid the confusion that nginx.conf should be edited. That's ~never the case (single-purpose nginx in a container being the exception), as nginx config is conf.d-style and the main `nginx.conf` is generally managed by the package manager and shouldn't be edited. Updated to hopefully clarify.

PR #1946, opened 2018-06-06
Configure max inactive duration

dtaniwaki:
I run JupyterHub with `cull_idle_servers.py` and collect metrics via statsd. However, since the max duration which JupyterHub considers a server is inactive is hard-coded and cannot make a consistent logic for both metrics and culling idle servers.

I'm thinking using it like below.
```
max_inactive_duration = 60 * 60
c.JupyterHub.max_inactive_duration = max_inactive_duration
c.JupyterHub.services = [
    {
        'name': 'cull-idle',
        'admin': True,
        'command': ('python3 cull_idle_servers.py --timeout=%d' % (max_inactive_duration * 2)).split(),
    }
]
```

So I'd like to configure max inactive server duration by a config file. Could you consider merging this change?

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -274,6 +274,9 @@ class JupyterHub(Application):
     service_check_interval = Integer(60,
         help="Interval (in seconds) at which to check connectivity of services with web endpoints."
     ).tag(config=True)
+    active_user_window = Integer(30 * 60,
+        help="Duration (in seconds) to determine the number of active users."
+    ).tag(config=True)
 
     data_files_path = Unicode(DATA_FILES_PATH,
         help="The location of jupyterhub data files (e.g. /usr/local/share/jupyterhub)"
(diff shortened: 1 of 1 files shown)


minrk:
Thanks for the PR! But I don't think we should make this change. The timeout for culling is not hardcoded, it is already a configurable value specified in your config file (in the line passed to the cul_idle_servers.py, which you have edited here to reference a different variable). What is done here is just moving the config value to a different line.

There are two mostly unrelated things in this PR. There *is* a hardcoded 'active window' of 30 minutes in the statsd metric used to determine how many users are 'currently active'. This is unrelated to the idle timeout used to cull servers, which can range from 10 minutes to multiple days, and doesn't have anything to do with how we measure 'recent activity'. Maybe you would be more interested in the `users.running` metric, which is more related to the idle culler?

If you want to just make the activity window configurable, that would be great. How about `c.JupyterHub.active_user_window` (in seconds)? This value should not be used to specify the idle timeout in cull-idle-severs, however.

dtaniwaki:
@minrk Thank you for your feedback!
I agree with you. The idle time of culling servers is not directly related to the duration to determine inactive users. I'm updating the PR with least changes as you explained.

minrk:
Perfect, thank you!

PR #1954, opened 2018-06-06
Blacken python doc build files

willingc:
This PR takes the `*.py` files in the Docs directory and runs them through [*Black*](https://github.com/ambv/black). *Black* is an opinionated python formatter for consistency and ideally smaller diffs/file lengths over time. By default, Black extends file line length to 88, and adds consistency in code block formatting.

This PR used `master` of *Black* so that the command line option --skip-string-normalization could be used.

Some large projects such as pytest have been using *black*. Full disclosure: I'm one of the co-maintainers.

@minrk Thoughts? For now, I would only do the doc files and possibly recommend blackening some of the test files. I would leave the code files as they are since it's pretty late in the release cycle.

### DIFF
--- a/docs/sphinxext/autodoc_traits.py
+++ b/docs/sphinxext/autodoc_traits.py
@@ -7,14 +7,18 @@ from sphinx.ext.autodoc import ClassDocumenter, AttributeDocumenter
 
 class ConfigurableDocumenter(ClassDocumenter):
     """Specialized Documenter subclass for traits with config=True"""
+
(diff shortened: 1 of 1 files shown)


willingc:
Totally agree with you on your assessment. And pleasantly surprised that you have been using black. I am with you on the workflow about changing files as I'm always wary of changing code without a thorough review. Do you want to open an issue if its not already there about diff-radius support?

PR #1958, opened 2018-06-08
Add error message text

### LINKED ISSUE #1155: npm local install suggestion
Hi,

This is not a bug, it's a suggestion for occasional python users like me, who are not familiar with node.js and npm related commands.

Every single time when I start a new virtualenv, and install jupyterhub, runnung `jupyterhub` tells me: 

    Failed to find proxy ['configurable-http-proxy']
    The proxy can be installed with `npm install -g configurable-http-proxy`
 
Which is fair enough, except since many of us, non node.js users, don't have a local copy of node in our environment, and npm is installed globally. This means running `npm install -g configurable-http-proxy` returns errors complaining about file/folder access permissions, which is again, fair enough since the npm is installed globally.

This means every time I need to search for the error online, and try to fix it somehow. After a few times doing it, it's not too hard to find the answer, but still annoying, since what I really expect, is to install my python package (jupyterhub in this case) and use it. And I'm certain many other users encounter the same problem.

So my suggestion is to include something along the lines:

    You need to set these two environment variables to install npm packages locally:
    $ PATH="$HOME/.node_modules/bin:$PATH"
    $ export npm_config_prefix=~/.node_modules

Thanks for the greate work BTW.

willingc:
Closes #1155

### DIFF
--- a/jupyterhub/proxy.py
+++ b/jupyterhub/proxy.py
@@ -488,7 +488,8 @@ class ConfigurableHTTPProxy(Proxy):
         except FileNotFoundError as e:
             self.log.error(
                 "Failed to find proxy %r\n"
-                "The proxy can be installed with `npm install -g configurable-http-proxy`"
+                "The proxy can be installed with `npm install -g configurable-http-proxy`."
+                "Please refer to the nodejs documentation if you need to install `npm`."
                 % self.command
             )
             raise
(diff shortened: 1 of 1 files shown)


betatim:
From a quick read of the issue one thing that stood out was that finding the right page in the npm docs was the stumbling block. What do you think of including the docs page Min linked to in the error message? I am a bit +-0 for links in error messages as they are bound to bit rot but they sure are helpful :-/

Maybe we can suggest a search query that is basically "guaranteed" to get you to the right page?

willingc:
Good catch @betatim. I modified the text.

PR #1961, opened 2018-06-08
Add link to authenticators used with Shibboleth

### LINKED ISSUE #332: setting up jupyterhub behind shibboleth authentication
We are setting up jupyterhub behind SSO (shibboleth) in our organisation. Is there a shibboleth authentication plugin available for jupyterhub?

If not,  the setup we have in mind is the following: the authentication step and authentication cookies are managed by the fronted web server (apache/mod_shib) which also acts as a reverse proxy to jupyterhub. All the traffic from reverse proxy should be trusted by the jupyterhub the user name comes in a header for each request. The unix users exist locally. What's best way to add such support (as an option) to jupyter hub? Is it in Base Handler? e.g.: https://github.com/jupyter/jupyterhub/blob/498c062ee001d5926b601f67648f1993826852c5/jupyterhub/handlers/base.py#L159

Does this approach make sense for you? If it does, would you be interested in having such an option integrated in the upstream jupyterhub codebase?

willingc:
Closes #332

### DIFF
--- a/docs/source/reference/authenticators.md
+++ b/docs/source/reference/authenticators.md
@@ -38,6 +38,8 @@ with any provider, is also available.
 
 - ldapauthenticator for LDAP
 - tmpauthenticator for temporary accounts
+- For Shibboleth, [jhub_shibboleth_auth](https://github.com/gesiscss/jhub_shibboleth_auth)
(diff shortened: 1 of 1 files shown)


willingc:
Thanks @betatim

Issue #1965, opened 2018-06-11
What is the meaning of the `name` parameter in the `_new_spawner` method of the `User` class?


krinsman:
Simple question (in reference to [`user.py`](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/user.py#L186)):

Does the `name` parameter in the `_new_spawner` method of the `User` class refer to:

- the name of the user (the way `self.name` does)
- the name of the notebook server?

It seems to be referring to the name to the notebook server, although I am probably misunderstanding.

If it refers to the name of the notebook server, then why not call the parameter `server_name`? In all of the other methods of the `User` class which refer to the name of a named server, the corresponding parameter is called `server_name`. So `server_name` would be more consistent with the rest of the code.

Also, if `name` doesn't refer to the name of the user, the way `self.name` does, then it is somewhat confusing giving it such a similar designation when they are referring to very different things.

krinsman:
Same question about the `name` parameter of the `stop_single_user` method of the `BaseHandler` class:

https://github.com/jupyterhub/jupyterhub/blob/cb9ce4d3af9f3b2a7e4fdb94786b91e99cd954b0/jupyterhub/handlers/base.py#L719

I'm basically 80% sure at this point that it is referring to the notebook server's name, here and above, not the user's name (since the corresponding attribute of an instance of a `Spawner` object is `name`).

However, this took a long time to confirm by searching through the (very large) source code.

minrk:
Following #1966: `name` is the name of the spawner here, so it sets `spawner.name`. Your #1966 helps clarify this by calling it `server_name` in the user methods, but it shouldn't go so far as to rename it at the Spawner level where it really is the name of the object itself.

krinsman:
@minrk To reiterate what I said in the pull request, I think your logic about keeping the designation of the `name` attribute of the `Spawner` class as is makes a lot of sense sense. Certainly the code is a lot simpler when the `name` attribute is kept as is in the `Spawner` class. 

Mostly I was interested in seeing if I could create a "proof of concept" for my conjecture that the name of the spawner corresponds to the name which the user gives to their corresponding notebook server.

Thank you again for your time and help!

Issue #1935, opened 2018-05-31
Named servers via UI


rcthomas:
There are a few issues posted about named servers already, but I'm really interested in seeing about what we can contribute on named servers through the UI (not via the API).

I imagine if a `server_name` value came in from the options form it could be picked off and set as the `server_name` parameter in the call to `spawn_single_user` done [here.](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/handlers/pages.py#L141-L142)

Or maybe it's better to add a `default_server_name` method to `Spawner` that is just the empty string and let various `Spawner`s determine `default_server_name` however they want.  This could involve form options or be independent of them altogether.  We would just need to call that method, get the answer and pass it to `spawn_single_user`.

But both above seem obvious so I must be missing something.  There are maybe other things missing like an admin page for all the servers for each user that should be done, or other logic I'm not thinking about?

willingc:
Thanks @rcthomas for the suggestion. Can you clarify a bit what you would like added to the options form?

rcthomas:
Well I am not sure that's the right approach.  :)  I could see how it could be used though.

More open-ended, is there a "to do" list outlining what things need to be done to get named servers working compatibly with the UI?  We have some personnel over the summer and we might be able to make some progress contributing to that effort.

minrk:
I don't think named servers UI is necessarily hard, we just haven't had the time or priorities to do it yet, so if you are volunteering to work on it, that would be great!

I think doing it via the options form is definitely one valid option. We should short-circuit it a little, because the server name shouldn't get passed through the overrideable options_from_form. Something along the lines of adding the server name input to the configurable options form if named servers are enabled, and always showing the spawn form if named servers are enabled, rather than just when options_form is defined.

e.g.

```python
form  = """
    <label for="jupyterhub-servername">Server Name</label>
    <input type="text" name="jupyterhub-servername" class="form-control">
""" + spawner.options_form
```

and pop `jupyterhub-servername` out of the formdata before calling `.options_from_form` on the rest.

The other main UI feature we need is a page for listing/managing/linking to existing servers (running and not). The admin UI will also need to gain support for named servers.

There's another option, and that's to make the "new named server" a specifically different action from the "spawn my (default) server" button we have now. Then the server name is picked earlier, rather than in the spawn form. This may be the best option. We would need to mock up the new "start/stop/delete/new" servers UI that works both for the default only-one server and the named servers case. This would probably replace the current 'home' page.

[...]

PR #1966, opened 2018-06-11
Increased consistency of references to names of notebook servers.

krinsman:
This is a proposed solution to this issue: https://github.com/jupyterhub/jupyterhub/issues/1965

The `name` attribute of the ORM spawner class, and the name property of the (non-ORM) spawner class are not used much outside of those classes, so it seemed possible to change their designations to the (seemingly) more unambiguous `server_name` without having to modify much else (it occurs twice in the code for the base API handler I think).

Also I changed the parameter designation from `name` to `server_name` in several methods of classes which appeared to be referring to the name of a notebook server.

The designation `server_name` is not my idea; it was already used in most places in the code. The idea of this pull request is to make thinking about the code referring to the names of notebook servers slightly easier, and thereby facilitate the process of implementing named servers for the UI. Thus this is also intended as a first step to addressing the related issue: https://github.com/jupyterhub/jupyterhub/issues/1935


jupyterhub/orm.py
```diff
@@ -204,7 +204,7 @@ class Spawner(Base):
     server = relationship(Server, cascade="all")
 
     state = Column(JSONDict)
-    name = Column(Unicode(255))
+    server_name = Column(Unicode(255))
```
minrk: This is the name of the Spawner, so I think it's redundant to use `spawner.server_name` here. Let's leave the Spawner.name field alone.

krinsman: Your logic makes sense to me -- I appreciate the clarification and feedback!

krinsman:
OK I screwed this up by merging my other branch into this accidentally. Anyway, this is basically already part of that other PR already, so I might as well close this one and focus on that one only.

PR #1970, opened 2018-06-12
Hub not found at localhost

JasonJWilliamsNY:
Added a command that worked for me to fix the situation that localhost:8000 is unable to reach the hub even though the published command for Docker exposes the correct port.


minrk:
Thanks for the PR! This is a recently introduced bug that's been fixed (#1910)

Both `--ip=0.0.0.0` and `--port=8000` are the defaults, so issuing this command should not result in a change to the default behavior with the current version of the jupyterhub/jupyterhub image.

PR #1977, opened 2018-06-13
Correct 'conda' installation instructions

kpfleming:
JupyterHub packages are in the 'conda-forge' channel of Anaconda packages; if the Anaconda installation doesn't already have 'conda-forge' enabled, `conda install jupyterhub` fails.

Rather than adding instructions to enable 'conda-forge' in Anaconda, this patch modifies the installation command to specify that channel.


minrk:
Thanks!

willingc:
Good catch. Thanks @kpfleming.

Issue #1982, opened 2018-06-14
User with a tilde in the name is being left as active without user activity


chicocvenancio:
**Describe the bug**
User with a tilde in the name is being left as active without user activity, preventing culling.

**To Reproduce**
1. Login with an user with a tilde 
2. Start a server, 
3. Close browser 
4. Start cull_idle_servers.py with debug logging
5. See user's server not being culled due to "being active"
**Expected behavior**
User to become inactive once browser is closed

**Additional context**
This seems related to the continous add and removal of the route to the user server by the proxy. Log excerpt:

> [I 2018-06-14 18:32:31.720 JupyterHub proxy:240] Adding user ~TestJH to proxy /paws/user/~TestJH/ => http://[REDACTED]:8888
> [I 2018-06-14 18:33:31.755 JupyterHub proxy:299] Checking routes
> [W 2018-06-14 18:33:31.755 JupyterHub proxy:322] Adding missing route for /paws/user/~TestJH/ (<jupyterhub.objects.Server object at 0x7f8d89d0a240>)
> [W 2018-06-14 18:33:31.756 JupyterHub proxy:361] Deleting stale route /paws/user/%7ETestJH/
> [I 2018-06-14 18:33:31.756 JupyterHub proxy:240] Adding user ~TestJH to proxy /paws/user/~TestJH/ => http://[REDACTED]:8888
> [I 2018-06-14 18:34:31.713 JupyterHub proxy:299] Checking routes
> [W 2018-06-14 18:34:31.714 JupyterHub proxy:322] Adding missing route for /paws/user/~TestJH/ (<jupyterhub.objects.Server object at 0x7f8d89d0a240>)
> [I 2018-06-14 18:34:31.714 JupyterHub proxy:240] Adding user ~TestJH to proxy /paws/user/~TestJH/ => http://[REDACTED]:8888
> [I 2018-06-14 18:35:31.770 JupyterHub proxy:299] Checking routes
> [W 2018-06-14 18:35:31.771 JupyterHub proxy:322] Adding missing route for /paws/user/~TestJH/ (<jupyterhub.objects.Server object at 0x7f8d89d0a240>)
> [W 2018-06-14 18:35:31.771 JupyterHub proxy:361] Deleting stale route /paws/user/%7ETestJH/
> [I 2018-06-14 18:35:31.771 JupyterHub proxy:240] Adding user ~TestJH to proxy /paws/user/~TestJH/ => http://[REDACTED]:8888
[...]

PR #1983, opened 2018-06-14
add test case for user with tilde

willingc:
This adds a test case for the proxy and user in issue #1982.


minrk:
Thanks for the test! I pushed an additional commit that should fix it. Then I think we're ready for 0.9.0.

minrk:
Will release 0.9.0 as soon as Travis confirms that these tests now pass

minrk:
Tests passed! Thanks for the last minute test. Time to cut 0.9

PR #1984, opened 2018-06-14
mark tilde as safe in proxy routespec quoting FIX:#1982

chicocvenancio:
~~not yet tested.~~ passes `pytest -v jupyterhub/tests/test_proxy.py` including with #1983


willingc:
Thanks @chicocvenancio

PR #1989, opened 2018-06-15
Remove encoding from open file for long description

willingc:
Match the example used in pypi example project (https://github.com/di/markdown-description-example/blob/master/setup.py).

I'm unsure if this was the issue for markdown rendering on pypi or if it was the versions of setuptools/twine used (https://dustingram.com/articles/2018/03/16/markdown-descriptions-on-pypi)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -83,7 +83,7 @@ for d, _, _ in os.walk('jupyterhub'):
     if os.path.exists(pjoin(d, '__init__.py')):
         packages.append(d.replace(os.path.sep, '.'))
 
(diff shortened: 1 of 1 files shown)


minrk:
I uploaded with twine 1.10 and the docs say that I needed to upload with 1.11, so I'm pretty sure that's it.

I think we shouldn't remove the encoding specification because if the file is ever opened with a different encoding (utf8 will still almost always be the default) and there are unicode characters, the result would be the wrong text.

willingc:
Closing this PR as it just brings you back to the original code.

Issue #2004, opened 2018-06-21
Culler raises an exception for a hub deployed at a different base URL


betatim:
Freshly deployed vanilla hub using the zero2jupyterhub chart (v0.7-8b9cb79) on google cloud. Noen standard parts are that this uses an nginx-ingress to route traffic to the hub which is at example.com/foobar.

When the culler runs the following error appears. I suspect the "ends with f" is somehow related to the base URL being `/foobar` (I did pick exactly this name for the test deployment it isn't a placeholder).
```
[E 180621 12:33:35 ioloop:792] Exception in callback functools.partial(<function wrap.<locals>.null_wrapper at 0x7fd7bd74f950>, <Future finished exception=HTTP 599: Port number ended with 'f'>)
    Traceback (most recent call last):
      File "/usr/local/lib/python3.6/dist-packages/tornado/ioloop.py", line 759, in _run_callback
        ret = callback()
      File "/usr/local/lib/python3.6/dist-packages/tornado/stack_context.py", line 276, in null_wrapper
        return fn(*args, **kwargs)
      File "/usr/local/lib/python3.6/dist-packages/tornado/ioloop.py", line 780, in _discard_future_result
        future.result()
      File "/usr/local/lib/python3.6/dist-packages/tornado/gen.py", line 1107, in run
        yielded = self.gen.throw(*exc_info)
      File "/usr/local/bin/cull_idle_servers.py", line 111, in cull_idle
        resp = yield fetch(req)
      File "/usr/local/lib/python3.6/dist-packages/tornado/gen.py", line 1099, in run
        value = future.result()
      File "/usr/local/lib/python3.6/dist-packages/tornado/gen.py", line 1113, in run
        yielded = self.gen.send(value)
      File "/usr/local/bin/cull_idle_servers.py", line 105, in fetch
        return (yield client.fetch(req))
      File "/usr/local/lib/python3.6/dist-packages/tornado/gen.py", line 1099, in run
        value = future.result()
    tornado.curl_httpclient.CurlError: HTTP 599: Port number ended with 'f'
```

Will test/debug a bit more and add details.

chicocvenancio:
@betatim are you certain the `base_url` is set as `/foobar` and not `foobar`? I would expect this error if the base_url is set without the starting slash.

minrk:
Which helm chart complains about the absolute base URL? I just successfully launched the z2jh chart with `hub.baseUrl = 'thiswontwork'` without complaint.

[...]

PR #2008, opened 2018-06-22
managed services always talk to hub on localhost

minrk:
When the Hub listens on all ips by default, connections use the hub's hostname.

In some cases (e.g. certain kubernetes deployments) the hub’s container’s hostname is not connectable from within itself, preventing managed services from connecting to the hub.

This ensures that managed service processes talk to the hub over localhost in this case, rather than via the hostname.

cc @betatim I believe #2004 is caused by the the z2jh [workaround for this](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/4dda7eb9aadafb8bf5a9eab4c0fd132d8c3b662e/images/hub/jupyterhub_config.py#L256) where it sets the Hub URL and can do it incorrectly, where the internal Hub URL setting should always do the right thing for base URLs.

With this patch and removing z2jh's workaround, I think #2004 should not happen.

### DIFF
--- a/jupyterhub/objects.py
+++ b/jupyterhub/objects.py
@@ -63,6 +63,9 @@ class Server(HasTraits):
     @validate('connect_url')
     def _connect_url_add_prefix(self, proposal):
         """Ensure connect_url includes base_url"""
+        if not proposal.value:
+            # Don't add the prefix if the setting is being cleared
+            return proposal.value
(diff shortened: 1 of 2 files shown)


betatim:
Superficially looks good to me. I'd have to think more about the hub deep copy by reading surrounding code. Maybe if someone else also has a look we can merge this. I won't get around to this again till next week.

PR #2009, opened 2018-06-24
proxy.py: Respect base_url in add_hub_route

BerserkerTroll:


jupyterhub/proxy.py
```diff
@@ -365,8 +365,8 @@ def validate_routespec(self, routespec):
 
     def add_hub_route(self, hub):
         """Add the default route for the Hub"""
-        self.log.info("Adding default route for Hub: / => %s", hub.host)
-        return self.add_route('/', self.hub.host, {'hub': True})
+        self.log.info("Adding default route for Hub: %s => %s", hub.base_url[:-4], hub.host)
+        return self.add_route(hub.base_url[:-4], self.hub.host, {'hub': True})
```
betatim: why the `[:-4]`?

BerserkerTroll: To strip `hub/` from the `base_url`. The suffix `hub/` is somehow appended to the `base_url` even though `base_url` is defined in a config file without this suffix. I think https://github.com/jupyterhub/jupyterhub/blob/02a52a02896d65bd0793135dbab7cba4e680802e/jupyterhub/app.py#L550-L552 and https://github.com/jupyterhub/jupyterhub/blob/02a52a02896d65bd0793135dbab7cba4e680802e/jupyterhub/app.py#L1098 are responsible for this.

minrk:
This is now working! I had to make it a bit more complex to handle host-based routing and update the test expectations, but it should be all set.

PR #2012, opened 2018-06-25
c.LocalProcessSpawner.shell_cmd configuration option does not work

josemonsalve2:
If you want to be able to change the shell_cmd option of the LocalProcessSpawner, this tag is needed in the class.

example config

```
c.LocalProcessSpawner.shell_cmd = ['bash', '-l', '-c']
```

Otherwise it will generate the following error during configuration parsing: 

```
[... JupyterHub configurable:168] Config option `shell_cmd` not recognized by `LocalProcessSpawner`.
```


minrk:
Good catch!

PR #2014, opened 2018-06-28
add python 3.7 to travis

willingc:


betatim:
Looks like Travis don't support 3.7 yet? Their docs mention 3.7-dev. Do you know if this will change soon, if yes let's keep this PR open and merge it as soon as 3.7 support is there.

willingc:
@betatim That's cool. Python 3.7 was released today so hopefully Travis will support soon.

minrk:
I learned from digging through travis that if we add:

```
dist: xenial
sudo: true
```

then `python: 3.7` will work. It's not available for the `trusty` image yet, though.

minrk:
Picking up xenial worked for me on pyzmq: https://github.com/zeromq/pyzmq/pull/1185/commits/603bafd2112ca8671569eb7ef5bb42e2a1a15a2a

willingc:
Removed the test env in the last commit and it passes. Unfortunately, it does fail when the env is present.

PR #2016, opened 2018-06-29
mention get_env and get_args in spawner reference

minrk:
these are important and usually required (especially get_env) for custom Spawner implementations


betatim:
LGTM

PR #2027, opened 2018-07-06
_ServiceSpawner: add 'SYSTEMROOT' to environment if Windows

adelcast:
Python 3 cannot be started without SYSTEMROOT environment variable.
Otherwise, CryptAcquireContext() is unable to find a dll.

https://bugs.python.org/issue20614

### DIFF
--- a/jupyterhub/services/service.py
+++ b/jupyterhub/services/service.py
@@ -41,6 +41,7 @@ A hub-managed service with no URL::
 
 import pipes
 import shutil
+import os
 from subprocess import Popen
 
 from traitlets import (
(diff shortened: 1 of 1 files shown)


adelcast:
@dhirschfeld might be interested

adelcast:
yep, it is broken for Windows. I created a generic Windows spawner that takes care of setting the environment: https://github.com/ni/jupyterhub-winlocalprocessspawner. But, for basic services, seems like my fix is enough (and there is no need to have a dependency on WinLocalProcessSpawner). In my case, I need services to run cull_idle_services.py.

PR #2036, opened 2018-07-13
pass requesting handler to spawner

minrk:
allows Spawners to implement logic such as processing GET params to select inputs

USE WITH CARE because this gives authors of links the ability to pass parameters to spawn without user knowledge or input.

This should only be used for things like selecting from a list of all known-good choices, e.g. a profile list, NEVER to set free-field values such as resource requests, image, etc.

cf https://github.com/jupyterhub/kubespawner/issues/189

cc @yuvipanda

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -602,7 +602,7 @@ class BaseHandler(RequestHandler):
 
         self.log.debug("Initiating spawn for %s", user_server_name)
 
-        spawn_future = user.spawn(server_name, options)
+        spawn_future = user.spawn(server_name, options, handler=self)
 
         self.log.debug("%i%s concurrent spawns",
             spawn_pending_count,
--- a/jupyterhub/handlers/pages.py
+++ b/jupyterhub/handlers/pages.py
@@ -111,7 +111,11 @@ class SpawnHandler(BaseHandler):
             if user.spawner._spawn_future and user.spawner._spawn_future.done():
                 user.spawner._spawn_future = None
             # not running, no form. Trigger spawn by redirecting to /user/:name
-            self.redirect(user.url)
+            url = user.url
+            if self.request.query:
+                # add query params
+                url += '?' + self.request.query
+            self.redirect(url)
 
     @web.authenticated
     async def post(self, for_user=None):
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -161,6 +161,7 @@ class Spawner(LoggingConfigurable):
     admin_access = Bool(False)
     api_token = Unicode()
     oauth_client_id = Unicode()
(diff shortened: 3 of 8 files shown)


minrk:
> I'm guessing this will preserve query params across refreshing redirects and stuff too?

That's up to the handler that triggers the redirect (e.g. PrefixRedirectHandler), but I believe we've fixed preserving params across redirects on all of our redirect handlers (there are lots, so this might not be true for 100%, but it's a bug anywhere it isn't).

> What happens when the user clicks 'stop' and then 'start' in the control panel?

Then `self.handler` will be the new handler from the start request with no params. If params should be set once and preserved, they can be stored on the Spawner and persisted across launches in state if that's appropriate, so you could have code like:

```python

_foo = None

def start(self):
    # get foo from url param
    foo = self.handler.get_argument('foo')
    if foo is None:
        # not defined on the current handler,
        # use previous value if there was one
        foo = self._foo
    else:
        # defined, remember it for next time
        self._foo = foo
    
def get_state(self):
    state = super().get_state()
    if self._foo is not None:
        state['foo'] = self._foo
    return state

def load_state(self, state):
    super().load_state(state)
    if 'foo' in state:
        self._foo = state['foo']
```

But it would be up to the Spawner to determine if such params should be preserved as defaults from one request to the next, and it can be determined on a param by param basis.

yuvipanda:
I like it, @minrk

minrk:
It works!

jmatuskey:
I used this to do a dynamic form w/ KubeSpawner for an image profile specified in URL query args. I modified jupyterhub to pass through URL query args but for Kubespawner I just used the config.
```
from kubespawner.spawner import KubeSpawner

c.JupyterHub.spawner_class = KubeSpawner

orig = [
    {
        'display_name': 'Training Env - Python',
        'kubespawner_override': {
            'image_spec': 'training/python:label',
            'cpu_limit': 1,
            'mem_limit': '512M',
        }
    }, {
        'display_name': 'Training Env - Datascience',
        'kubespawner_override': {
            'image_spec': 'training/datascience:label',
            'cpu_limit': 4,
            'mem_limit': '8G',
        }
    }
]

def dynamic_profile_list(spawner, **kwargs):
    return orig + [
        {
            'display_name': spawner.handler.get_query_argument("image", "unknown"),
            'default': True,
            'kubespawner_override': {
                'image_spec': spawner.handler.get_query_argument("image", "unknown"),
                'cpu_limit': spawner.handler.get_query_argument("cpu_limit", 2),
                'mem_limit': spawner.handler.get_query_argument("mem_limit", "2G")
            }
        }
    ]
    
c.KubeSpawner.profile_list = dynamic_profile_list

c.KubeSpawner.start_timeout = 60 * 5
```

PR #2050, opened 2018-07-23
Switch protocols to https in docs links

Carreau:
Chrome will start to show insecure website for http next week


betatim:
LGTM and thanks for helping keep documentation nice!

Restarted the build as it seems likely this is a flakey test, not a real failure.

Carreau:
> Restarted the build as it seems likely this is a flakey test, not a real failure.

This is an actual failure. For whatever reason an old version of attrs get installed that is incompatible with JsonSchema...

willingc:
Remember RTD is using environment.yml to build the docs. You will need to add/change attrs in that file located in docs directory. cc/@betatim @Carreau

PR #2053, opened 2018-07-23
Make attrs dependency explicit and require a higher version

betatim:
An attempt to fix the builds of #2050


betatim:
This fixes the builds. Not sure why we don't get a new enough version of `attrs` without explicitly requiring it though.

betatim:
Closing in favour of #2050

Issue #370, opened 2016-01-06
Enable SSL/TLS for internal communication


minrk:
It should be possible to use SSL/TLS/PKI for all internal communication, but I think http is hardcoded a few places.

It would be doubly awesome if we could generate/use certs for this, a la DOCKER_CERT_PATH, DOCKER_TLS_VERIFY.

minrk:
@rgbkrk said I should cc @smashwilson and @carolynvs for ideas about implementing this.

The gist:

- one proxy, one hub, N single-user servers
- hub connects to proxy, users
- proxy connects to hub, users
- users connect to hub

Awesome enough: HTTPS connections everywhere, just encryption, no auth
Double Awesome: authenticate with certs as well

carolynvs:
I was just thinking about this the other day! I love the idea. :smile:

tgmachina:
Hey folks, I wanted to outline/discuss some of the stuff I have been working       
on towards this goal. I brought up this topic in the security group, and I         
thought this would be a better place to have this discussion.                      
                                                                                   
The overall changes were in line with what @yuvipanda and I discussed:             
   * Create an ephemeral CA for creating and signing certs                         
   * Dole out certs to all components that need them                               
                                                                                   
The guiding requirements I kept for these changes were:                            
   * When internal SSL is disabled, JupyterHub works as it always has              
   * Enabling internal SSL should require no more than switching on a flag         
   * Bi-directional verification, always: all components should be able to         
   verify that they are talking within the realm of trust of the ephemeral CA.  
                                                                                   
Most components in JupyterHub support the use of SSL, so most of the               
development work involved getting certs into the right place via appropriate       
options parameters. As @minrk pointed out, there are a few places where            
http was hard-coded and I made sure those pieces changed to react to the           
environment they were in--SSL or otherwise.                                        
                                                                                   
For cert creation and management, I created a small, python utility that wraps             
pyopenssl that I import into JupyterHub.                                                                         
                                                                                   
With these changes, I have JupyterHub working with all communication legs running SSL. On        
startup, with the `internalSSL` flag on, JupyterHub will generate a CA (if         
none exists), generate certs for the proxy and hub, and sign/chown certs for       
each notebook that gets started.                                                   
                                                                                   
I also had to make some small changes to `configurable-http-proxy` to forward   
requests over SSL as well.                                                         
                                                                                   
I am still working on larger scale deployment here at work as well as starting  
the release process so that I can submit a PR back to you guys.                    
[...]

PR #2055, opened 2018-07-23
End to end ssl

tgmachina:
These changes are in service of turning on SSL internally for JupyterHub (issue #370).

To generate certs, I created a separate, module [Certipy](https://github.com/LLNL/certipy) that wraps pyOpenSsl. Certipy has also been uploaded to PyPI.

To test these changes, I added 3 separate test files that turn on internalSSL via monkey patching. Tests intermittently fail when run together, but pass individually. @minrk had initially suggested making all requests to the hub `public_url`, but I wasn't able to work out a good scheme for that. I set up the 3 separate test files to get these changes through review ASAP, but I'd be happy to iterate on changing these tests.

These changes must be used in conjunction with [changes made to CHP](https://github.com/jupyterhub/configurable-http-proxy/pull/170) in order to work (likely automated github testing will fail because of this). I wasn't entirely sure how to specify a certain CHP version as a requirement.

Happy to iterate on whatever changes you see necessary. Thanks for your help and patience!


jupyterhub/spawner.py
```diff
@@ -666,6 +673,58 @@ def format_string(self, s):
...
+        ca = key_pair['cafile']
+
+        try:
+            user = pwd.getpwnam(self.user.name)
+            uid = user.pw_uid
+            gid = user.pw_gid
+            home = user.pw_dir
+
+            # Create dir for user's certs wherever we're starting
+            out_dir = "{home}/.jupyter/certs".format(home=home)
```
minrk: Let's call this `.jupyterhub/jupyterhub-certs` just to be clear

tgmachina:
For the case where the proxy lives separately from the hub, would it be worthwhile to enable a command like `jupyterhub --generate-config` except to set up the initial certificate authorities/certs that `internal_ssl` uses, for example, something like: `jupyterhub --generate-certs`? That would allow users to get and move the certs they need manually.

As for your assessment of `.move_certs`, that is correct. I have to think about the container case a bit more too, because its also important to get the appropriate domain/ip info into the cert in order for ssl verification to work. That capability exists with the use of `alt_names` at certificate creation, but I am unsure if the info is available or known before container start. I'll think about that some more and work on clarifying this API, good call.

I'm working through these changes now. I think the only thing I'm curious about are the docs. I was thinking of expanding the docstrings for `.create_certs` and `.move_certs` as well as including more notes within the `proxy.md`, `spawner.md` and `websecurity.md` docs. Seem reasonable?

minrk:
> something like: jupyterhub --generate-certs?

That's worth a try. I'm not 100% sure what will work best.

> because its also important to get the appropriate domain/ip info into the cert in order for ssl verification to work

oooh, that makes things more difficult because it might be hard/impossible to know this before the server starts in some cases. I'm not quite sure how to deal with that.

> I was thinking of expanding the docstrings for .create_certs and .move_certs as well as including more notes within the proxy.md, spawner.md and websecurity.md docs. Seem reasonable?

Yes, absolutely!

tgmachina:
Ok, just as an update, here's what I'm adding/changing/pondering:

* Nearly done with updating the docs, just pending the changes below.
* The `Proxy` class will have reference to its own certs either:
    * Created and signed on Proxy init using the `Hub` CA (basically how its done now).
    * Created by some third-party mechanism (Let's Encrypt, Certipy, self-signed, etc) for which the signing CA file is specified in the proxy config so that trust can be propagated appropriately.
* `Docker`/`Kube` cases w.r.t `internal_ssl`:
    * Swarm/Kube managed deployments: as I understand them, the networking can be set such that containers are connected with `tls`. If containers/pods running the single-user notebooks can't reach each other, this seems like it would have about the same effect as using `internal_ssl`.
    * `Docker` does have a mechanism for specifying an `ip` address prior to launch. This would entail drawing a random, unused `ip` from `Docker's` reserved subnet for container `ips`. I can investigate what enabling that would entail.

tgmachina:
Ok, I've updated CHP and JupyterHub accordingly. Overview of the changes:
* External proxy certs can be specified via config (for when the proxy is externally managed)
* CHP has two CAs: one for the API and the other for all client connections
* Trust is set up via Certipy's "trust_graph". Basically what component trusts what can be specified in a more human readable fashion.

Sorry for the delay in getting that out. I'll be a bit busy over the coming weeks, but I'll keep an eye out as always and look for anything else that needs changing/adding.

minrk:
Terrific! I'll try to get a review of this soon

minrk:
For the docker networking case, I don't think the pre-chosen ip ought to be an issue because the container name also works as a connectable hostname within docker networks, and that's known and chosen by us prior to start.

That should leave docker with the following cases:

1. internal docker (use container name as connect hostname instead of ip)
2. docker-exposed network (static public hostname/ip of machine)
3. exposed localhost docker (use 127.0.0.1)

I'm going to try to write a PR to dockerspawner to get a handle on this.

In the meantime, what would be a super useful thing to add is some tests that verify that various of these connections are actually rejected when requests aren't made with the right certs. I don't see those in the PR, unless I'm missing something.

minrk:
@tgmachina is it alright if I update this PR with changes as I test? I might want to tweak a few things, but I don't want to conflict with work you are doing.

minrk:
I've been able to get a version of this running with DockerSpawner with the proxy local to the hub. I had to make one adjustment to get it to work:

- make `move_certs` async. That's a minor change, but it requires moving the call to a different stage, since `.get_env` shouldn't have side effects, it should only report environment variables based on already-created state.

The next hiccup I'm hitting is running the proxy separate from the Hub, which is a best practice in container deployments. This is tricky right now due to the order of SSL initialization. I think setting up the proxy certs needs to happen in `.init_internal_ssl` rather than `init_proxy`, because `jupyterhub --generate-certs` needs to result in creating all the certs required.

tgmachina:
Hey @minrk! Got it, so `.create_certs` and `.move_certs` should be in their own phase outside of `.get_env`. Was that the change you were going to send back? Feel free to push any changes that you need to, no problem. I've been revamping the tests and working on adding more to test that connections fail as you asked.

tgmachina:
My altering of the app fixture to respond to `ssl_enabled` in tests was flawed. I've fixed the fixture and updated the tests accordingly. The main change was ensuring that when a test required making a request to a component, that the correct certs were attached when the `ssl_enabled` flag is turned on.

In addition to fixing the existing tests, I added a `test_internal_ssl_connections.py` file that has tests to ensure that connections to components with `internal_ssl` enabled, fail without the right certs.

minrk:
Okay, after resolving the subdomain issues with trusted_alt_names, I believe this is totally working. I just need to update the tests with real-database backends, because `test_api` can't actually be run multiple times against the same database.

consideRatio:
Wooooooooooooo wow this PR! Massive work done here! I've been keeping an eye on it, well done @tgmachina and @minrk !!! :tada:

tgmachina:
Wow! Thanks so much for all your help @minrk!

tgamblin:
This is awesome!  Thanks @tgmachina and @minrk!

PR #2057, opened 2018-07-24
proxy: make process existance check Windows friendly

adelcast:
Currently, to check if the proxy is running, os.kill(pid,0) is used,
which doesn't work on Windows. Wrapped call into a new function that
adds a Windows case.


adelcast:
@dhirschfeld

dhirschfeld:
LGTM.

`psutil` is already a windows dep and the import is behind an `os.name == 'nt'` check :+1:

image

dhirschfeld:
I'll ping @minrk since I can't merge PRs

minrk:
Thanks!

PR #2060, opened 2018-07-26
Update dependencies used by ReadTheDocs

betatim:
Update the dependencies used by RTD per @willingc's comment. I also added a comment to the doc's dependency files to write down this knowledge that RTD uses the `environment.yml`. Till just now I was confused about why we had both :)


willingc:
@betatim We have both for historical reasons and if folks with to use conda or pip to build locally.

PR #2062, opened 2018-07-27
Ensure request uri with trailing slash

### LINKED ISSUE #2005: Redirect to login page broken when using baseUrl
Freshly deployed vanilla hub using the zero2jupyterhub chart (v0.7-8b9cb79) on google cloud. None standard parts are that this uses an nginx-ingress to route traffic to the hub which is at example.com/foobar.

http://example.com/foobar redirects to http://example.com/foobar/hub/foobar but http://example.com/foobar/ redirects to http://example.com/foobar/hub/login.

Via @minrk and @chicocvenancio

> Yeah, I'm guessing it's not redirecting correctly without the trailing slash. I think there's a check for if not startswith(prefix) and it probably incorrectly returns False on the edge case where the trailing slash is missing
> I think it's PrefixRedirectHandler 
>
> `base_url` is defined as having the trailing slash https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/handlers/base.py#L63

chaoleili:
When request uri matching with base_url in PrefixRedirectHandler,
it's better to ensure uri with tariling slash. That's will avoid
redirecting /foobar to /foobar/hub/foobar.

closes #2005

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -901,6 +901,11 @@ class PrefixRedirectHandler(BaseHandler):
     """
     def get(self):
         uri = self.request.uri
+        # Since self.base_url will end with trailing slash.
+        # Ensure uri will end with trailing slash when matching
+        # with self.base_url.
+        if not uri.endswith('/'):
+            uri += '/'
         if uri.startswith(self.base_url):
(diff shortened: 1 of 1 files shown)


minrk:
Thanks!

PR #2066, opened 2018-08-01
Change GET method of UserSpawnHandler to accommodate named servers

krinsman:
This isn't meant as a serious proposal yet. I just want to compare the logic of [the POST request](https://github.com/jupyterhub/jupyterhub/blob/f2941e36312a2a1134f04e06f39bd9ca248cc046/jupyterhub/apihandlers/users.py#L373) in [`UserServerAPIHandler`](https://github.com/jupyterhub/jupyterhub/blob/f2941e36312a2a1134f04e06f39bd9ca248cc046/jupyterhub/apihandlers/users.py#L344) (in `apihandlers/users`) with the logic of [the GET request](https://github.com/jupyterhub/jupyterhub/blob/5f97487184bb8e308ed68d024a02ae0ef1bc4fe1/jupyterhub/handlers/base.py#L1012) in [`UserSpawnHandler`](https://github.com/jupyterhub/jupyterhub/blob/5f97487184bb8e308ed68d024a02ae0ef1bc4fe1/jupyterhub/handlers/base.py#L878). 

Specifically, to see where the logic of the former fails in the implementation of the latter.

See https://github.com/jupyterhub/jupyterhub/issues/1935


krinsman:
Also I realize I forgot to change some of the `server_name`'s back to `name`'s. To clarify, the ones which should be changed back (for reasons of avoiding redundancy and backwards-incompatibility) are those occurring in `orm.py` and `spawner.py` (i.e. any instance of the `name` attribute of the spawner class)? And the `server_name`'s which can remain `server_name`'s are variables or parameters whose values are set equal to `spawner.name` for some spawner `spawner` ?

krinsman:
Latest commit intended to address: https://github.com/jupyterhub/jupyterhub/pull/1966/files#r194695527

jupyterhub/handlers/base.py
```diff
@@ -978,14 +979,21 @@ def _fail_api_request(self, user):
                 port = 443 if host_info.scheme == 'https' else 80
             if port != Server.from_url(self.proxy.public_url).connect_port and port == self.hub.connect_port:
                 self.log.warning("""
-                    Detected possible direct connection to Hub's private ip: %s, bypassing proxy.
-                    This will result in a redirect loop.
-                    Make sure to connect to the proxied public URL %s
-                    """, self.request.full_url(), self.proxy.public_url)
-
+                Detected possible direct connection to Hub's private ip: %s, bypassing proxy.
```
minrk: Looks like this got dedented by accident

jupyterhub/handlers/base.py
```diff
@@ -978,14 +979,21 @@ def _fail_api_request(self, user):
...
+                """, self.request.full_url(), self.proxy.public_url)
+                    
             # logged in as valid user, check for pending spawn
-            spawner = user.spawner
-
+            if server_name is None:
+                spawner = user.spawner
+            else:
+                spawner = user.spawners[server_name]
+                #if self.allow_named_servers:
```
minrk: This `allow_named_servers` check should still be here

jupyterhub/handlers/base.py
```diff
@@ -994,7 +1002,7 @@ def _fail_api_request(self, user):
                 and spawner._spawn_future.exception()
             ):
                 # Condition: spawner not active and _spawn_future exists and contains an Exception
-                # Implicit spawn on /user/:name is not allowed if the user's last spawn failed.
+                # Implicit spawn on /user/:user_name is not allowed if the user's last spawn failed.
```
minrk: leave `:name` here

minrk:
Would it be possible to add a test for what didn't work before?

jupyterhub/handlers/base.py
krinsman: You're right.

jupyterhub/handlers/base.py
krinsman: Do you want me to change all of the `user_name`'s back to `name`, or just this one?

jupyterhub/handlers/base.py
krinsman: I agree it should still be there. When I added it, it caused the tests to fail.
 
(The first time I added it, the tests failed due to a 400 HTTP Error being raised. When I deliberately changed HTTP error on line 995 to be a 500 HTTP error, the tests failed again, but this time due to a 500 HTTP error. So I am fairly confident, albeit not certain, that it was this block that caused the tests to fail.)

The test failure was something about /user/river or something like that. (I want to say `test_auth.py`, but my memory isn't fresh.) 

My guess would have been that the tests wouldn't use named servers for the UI, so I was somewhat confused why this section of the code was reached in the first place.

Anyway, I figured you wouldn't look at a failing PR, so I commented that block out.

krinsman:
@minrk I'm sorry, I don't think I understand the question.

What do you mean by "what didn't work before"? 

In what sense would you like a test for it? (I.e. what would you want the test to discover about it?)

krinsman:
@minrk The original question on the comment got deleted when I changed the `user_name` on line 997 back to `name`, so to record it here: 

Should I change all of the `user_name`'s back to `name`, or just the one that was already changed back?

krinsman:
Let me try this again; I think I coded myself into a dead end here.

PR #2067, opened 2018-08-01
Add an example simple announcement service

rcthomas:
This is a simple service that allows admin users to manage basic dynamic announcements that appear when JupyterHub renders pages.  Maybe it's interesting as an example.

### DIFF
--- a/examples/service-announcement/announcement.py
+++ b/examples/service-announcement/announcement.py
@@ -1,0 +1,76 @@
+
+import argparse
+import datetime
+import json
+import os
+
+from jupyterhub.services.auth import HubAuthenticated
+from tornado import escape, gen, ioloop, web
+
+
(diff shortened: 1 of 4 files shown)


examples/service-announcement/announcement.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def initialize(self, storage):
+        """Create storage for announcement text"""
+        self.storage = storage
+
+    @web.authenticated
+    def post(self):
+        """Update announcement"""
+        user = self.get_current_user()
+        if user is None or not user.get("admin", False):
```
minrk: You can get this behavior already by setting the class attributes:

```python
hub_users = []
allow_admin = True
```

rcthomas:
I could be wrong but I don't think the travis failures are because of these changes...

minrk:
Nice, thanks!

PR #2080, opened 2018-08-10
prepare to release 0.9.2

minrk:
mostly docs and small bugfixes, with one new feature: `Spawner.consecutive_failure_limit` from #2042

### DIFF
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -9,6 +9,16 @@ command line for details.
 
 ## 0.9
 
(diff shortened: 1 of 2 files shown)


willingc:
Do you want to pin `cryptography` to 2.3 due to its latest CVE or let it float?

minrk:
> Do you want to pin cryptography to 2.3 due to its latest CVE or let it float?

Since it's not a direct dependency, I don't think we should pin it. Good to check, though.

Issue #1956, opened 2018-06-07
Named server backend implementation questions


krinsman:
I have some dumb questions about what changes and where need to be made to JupyterHub internals in order to enable the named servers feature. There are obviously a lot of them, so if it is only possible to answer one or two of them (due to time constraints and similar issues), I would still be immensely grateful if you did so. My goal is to avoid doing something massively stupid and consequently later making a garbage PR which even more completely wastes your time.

Here are my guesses about what at a minimum might need to be changed; please feel free to correct anything and everything that is wrong:

- Some additions to the [orm](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/orm.py) allowing each user to be mapped to multiple spawners (each spawner for a different notebook server), and allowing each spawner to be associated with a specific server name and a different remote host for that notebook server. (This is probably wrong since I don't understand this code as well as the `spawner.py` code, which I also don't understand that well.)
- Some additions to the [JupyterHub application](https://github.com/jupyterhub/jupyterhub/blob/14c8d7dc46017100abd1ac9455d90a27b4b27367/jupyterhub/app.py#L677) which cause different behavior to occur when `self.allow_named_servers` is set to `True`. (As far as I can tell, based on the source code and some tests, it is currently a "dummy attribute", inasmuch as whether or not it is true makes no difference in the program's behavior.
- Some additions to the [User](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/user.py#L129) class which will again cause different behavior when `self.allow_named_servers` is set to `True`. My guess is that the most additions will need to occur in the [`spawn`](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/user.py#L334) and [`spawner`](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/user.py#L226) methods, but additions might also need to be made to the [`progress_url`](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/user.py#L325) and `stop` methods, since they have `server_name` as possible parameter values.

Here are some actual, concrete questions:

[...]

PR #2089, opened 2018-08-16
UI for Named Servers (Home and Admin Pages)

krinsman:
**EDIT:** Most of this was dumb. See below.

<strike>Spawn a named server by making a GET request (by typing the URL in the browser) to the route:

```
<hub_base_url, e.g. localhost:8000/hub>/user/user_name/arbitrary_user_path?server=server_name
```

If the user isn't logged in, they will be re-directed to a log-in page. If applicable, the user enters their credentials. If a server with the name `server_name` does not already exist, it is created. Then the user is re-directed to (in most cases):

```
<single-user_base_url, e.g. localhost:8000/>/user/user_name/server_name/tree
```

which is the page for the newly created server `server_name`. 

However, there are still edge cases where the re-direction doesn't work as expected and one gets a 404 (although the named server `server_name` is still created and can still be accessed afterwards at the route `<single-user_base_url>/user/user_name/server_name/tree`). I have tried several times to reliably reproduce one of the corner cases so as to document it, but have not yet succeeded. 

What does always work (i.e. will guarantee avoiding one of the corner cases) is choosing `arbitrary_user_path` so that `arbitrary_user_path == server_name`.

That case works more reliably than the case when `abitrary_user_path != server_name`, which required more extensive changes to the code in order to work when the user hadn't logged in yet .

I don't know if this will pass the Travis build tests, but it did work with my testing setup, which can be found [here](https://github.com/krinsman/jupyterHubTestingWorkflow). Perhaps the problem is that my setup was not sufficiently realistic.</strike>

Any feedback is welcome, although constructive criticism would be preferred.


jupyterhub/handlers/base.py
```diff
@@ -1105,7 +1115,15 @@ def _fail_api_request(self, user):
                 url_parts = urlparse(target)
                 query_parts = parse_qs(url_parts.query)
                 query_parts['redirects'] = redirects + 1
-                url_parts = url_parts._replace(query=urlencode(query_parts))
+                # https://stackoverflow.com/a/21997659/9899076
+                # https://docs.python.org/3/library/urllib.parse.html#urllib.parse.parse_qs
+                url_parts = url_parts._replace(query=urlencode(query_parts, doseq=True))
+                # prevent redirect loop when user_path != server_name
+                if user_path != server_name:
```
minrk: What's the case addressed by this? Can you make a test for it? I don't follow the `split("/")[:3]` part, and I'm not sure it's going to be correct with respect to trailing slashes and requests for `/user/foo/servername/notebooks/somenotebook.ipynb`

minrk:
One additional point, based on the title: I don't think we need GET requests to be able to specify server name in order to have a backend for named-server UI. That should be entirely built on the existing REST API, which ought to be sufficient for the named server UI.

My envisioning of the named server UI, based on #1956 is something like the following:

- the handler serves a HTML page and doesn't trigger spawn at all
- on that page is some HTML and javascript, including button(s) that talk to the REST API
- e.g. a Launch button that makes a POST to the REST API and redirects the user when the launch has finished.

krinsman:
@minrk To clarify, which handler do you think should serve an HTML page and not trigger a spawn at all? It wasn't clear to me what the antecedent in that sentence was.

It seems to me that `LoginHandler` already serves an HTML page and doesn't trigger a spawn at all. Maybe the problem is that right now `LoginHandler` automatically redirects to `UserSpawnHandler` instead of `HomeHandler`? And that the home page rendered by `HomeHandler` doesn't have buttons sending requests handled by `UserSpawnHandler` and causing spawns of named servers?

Maybe it should be changed so that, when `self.allow_named_servers = True`, `LoginHandler` redirects to `HomeHandler` instead of `UserSpawnHandler`? 

(It appears to be the case to me that right now `LoginHandler` redirects automatically to `UserSpawnHandler`, regardless of the value of `self.allow_named_servers`, although this isn't entirely clear to me, so please correct me if that is wrong.)

Then it seems like we have the same ideas in mind, since then on the home page there would be HTML and Javascript for buttons which talk to the API and make the requests necessary to spawn the servers. If the admins want the users to be able to name the servers themselves, the HTML and Javascript can include forms that modify the API requests to include the custom names, otherwise the buttons can contain pre-coded logic for making the names themselves (e.g. the use case at NERSC).

To me it doesn't matter much on a personal level whether the spawn occurs via GET or POST. 

To me, it seems like if all server spawns should happen via POST, then `UserSpawnHandler` should be modified to only have a POST method and no GET method. Because right now, with or without named servers, the default behavior of LoginHandler redirecting automatically to UserSpawnHandler which then spawns a server (with name `''`) via GET request is already violating the paradigm from the API (where servers are spawned only via POST). So again, from my naive outsider perspective, it doesn't seem like this PR is violating the API neither any more nor any less than the current UI does.

(And from my naive outsider perspective, "unnamed servers" are de facto just a special case of named servers. The only difference, as I see it, when `self.allow_named_servers = False`, is not actually that named servers are disabled. To me it seems that the only difference is that the length of `user.spawners` is forced to be exactly one, and that the name of the only server in it is `''`. It's a stupid/null/trivial name, but technically a name. In fact, my beliefs are so radical that even having both a `user.spawner` attribute and a `user.spawners` attribute seems redundant to me. Regardless of whether or not there are multiple servers per user, each server is spawned via a call to `spawn_single_user`.)

This actually seems like a possible instance of [Hyrum's law](http://www.hyrumslaw.com/) (which I literally learned about for the first time yesterday) in action. I might be depending upon observable behaviors of JupyterHub which weren't promised in the contract (the API). So that might make this PR fundamentally flawed. (I hope not.)

That or maybe this is a case of a leaky abstraction, where I (as the user of the code) am utilizing the technical implementation of un-named servers as named servers and thereby ruining the abstraction.

My original attempt had been to modify `UserSpawnHandler` to use a POST method instead of a GET method, but that was evidently a major breaking change (at least as I unsuccessfully tried to implement it). In particular, getting the authorization flow to work correctly was a mystery to me since the internals of JupyterHub's authorization/authentication process remain mostly unfamiliar to me. 

In contrast, when keeping `UserSpawnHandler` as a GET-only handler, and specifying the server name via an additional query argument, it was possible for me to piggy-back on all of the existing re-direction and authorization/authentication logic without breaking anything. I.e. it was possible for me to do this without having to understand the internals of how cookies, tokens (what is the difference between a cookie and a token? and are there multiple types of tokens), and the proxy work in JupyterHub.

I won't be at JupyterCon next week, but to the best of my knowledge @rcthomas will be there. Maybe it would be worthwhile to discuss and try to iron out the philosophy of named servers with him there? It's just a suggestion; it's not really clear to me personally either way.

jupyterhub/handlers/base.py
krinsman: Oh you're definitely right, the `split("/")[:3]` wouldn't be correct for opening specific files at all. That wasn't something that had occurred to me. So that's probably just wrong. 

The **TL;DR** is that with the code for this section (as it is in the current master), there would be a redirect loop (and no spawn whatsoever I think) when `user_path != server_name`. But for some reason it would work (no redirect loop + successful spawn) when `user_path == server_name`. 

Basically all of the changes in this part are probably bad, or at least unnecessary, *except* the additional `doseq=True` argument in `urlencode`.**END TL;DR**

A better solution would have been to figure out why everything worked (with the re-direct code as it is in master) when `user_path == server_name` and then make sure the same thing happened when `user_path != server_name`. But after several hours I gave up and went with the hacky solution of re-directing/changing the URL to `user_path == server_name` in the cases when that wasn't true originally, and then everything "worked".

I.e. I didn't actually fix the problem, I just forcibly converted all of the cases where the problem occurred into special instances of the nice case where it didn't occur.

As you already point out above though that isn't a good idea since it is functionality-diminishing, e.g. inasmuch as it makes it impossible to access individual files on a named server.

Also @rcthomas told me it was already enough for NERSC's use case if it worked only when `user_path == server_name`. (So why bother having an implementation for when `user_path != server_name`? Maybe instead just raise an exception that says it's supposed to fail?)

And in any case my implementation of the forcible conversion to `user_path == server_name` (namely the `split("/")[:3]` thing) is implemented in a stupid way. I tried writing a regex for it but gave up after a couple of hours of failing to do so. My idea had been to use the same capturing groups as in the Tornado handler to parse the string, but that apparently doesn't work. (Or if it can be made to work, only with a lot of difficulty, since according to StackOverflow doing that would violate the spirit or intent of regex or something.) So basically I just gave up and hard coded in the `split("/")[:3]` part (`['', 'hub', 'user', 'user_name' ]`) to implement the forced conversion. Which is obviously dumb since it hard-codes a lot of assumptions which aren't stated explicitly and which can't be expected to hold in all circumstances.

There is a little bit more to the code in this section than the stupid parts above though. One thing I noticed is that, during the re-direct loops, every time the page got re-directed, an extra pair of square brackets was added to the query argument. So `server_name` (which remember is the query argument) would start out as `test`, but by the end of the re-direct loop it would be `'['['['test']']']'` or `` `[`[`[`test`]`]`]` `` or something else terrible like that. It's a result of a quirk of `parse_qs` documented at the two webpages in the comments above the line

    url_parts = url_parts._replace(query=urlencode(query_parts, doseq=True))

in particular, the extra `doseq=True` argument to `urlencode` is the fix. This doesn't matter for the code as it is in master, since the relevant GET requests don't have string-valued query arguments. But for this PR, which does introduce string-valued query arguments to that request, it seems that omitting the extra `doseq=True` argument would lead to unintended bugs. In particular, the addition of the spurious square brackets seemed to be part of what was causing the re-direct loop (since there may have even been a re-direct loop even when `user_path == server_name` without the fix), although they weren't the entire source of the problem.

krinsman:
Oh duh I just realized the reason why it works really nicely when `user_path == server_name`.

When `user_path == server_name`, the whole URL path (ignoring the query arguments) is the same as the URL for the root path of the named server as established by the API.

So it doesn't just work more nicely, it also makes a lot more sense as a GET request in the first place.

Given that valid motivation for the `user_path != server_name` case does not really exist, and that the implementation for it is bad, I will update the PR soon to remove that block.

minrk:
> To clarify, which handler do you think should serve an HTML page and not trigger a spawn at all?

UserSpawnHandler. Essentially, I think we should stop GET requests from triggering spawn implicitly if the server isn't running, instead showing a form/button for spawning.

krinsman:
**copy/pasted from Gitter -- will probably delete this here soon:**

> @minrk I think I finally understand what you mean by using the API to spawn named servers. Right now the GET query argument thing is still in the PR because it is useful for rapidly adding named servers via the UI (since I am not familiar with any spawners with an options form which I could hack with). It seems like I managed to get the home page working with named servers, but now two of the tests are failing.

> The reason seems to be that I chose to to mimic UserSpawnHandler's functionality for the default server with named server. Namely, UserSpawnHandler re-spawns the default server whenever it receives a request for `/hub/user/user_name/*`. This makes the home page a lot easier to implement for the default server.

> So I tried to do the same thing with the named servers -- re-spawn (if it had already been spawned before and is still in the ORM) the named server server_name whenever the hub receives a request for `/hub/user/user_name/server_name/*`.

> Is that a good idea? Or should I try to brute force a completely new implementation of the home page buttons for the named servers using only Javascript and no server side request handling? 

> (I wanted to keep the 'My Server' and 'Start My Server' buttons, as well as their analogs, in the same CSS class, but if there really should be no expanded request handling by UserSpawnHandler for named servers, then it seems difficult to get the JavaScript to work without splitting those buttons into two different CSS classes.)

krinsman:
@minrk You were right, it is definitely possible to get this to work only using Javascript and the API. I hope you like it!

krinsman:
Just added the UI for the admin page.

**Home page:**

home_page

**Admin page:**

admin_page

krinsman:
@minrk Is this PR merge-able after being cleaned up? 

It would be helpful to know if it is a waste of time to spend more time cleaning this PR up before beginning to do so.

Some things I have in mind regarding what might need to be cleaned up -- please let me know how (in)accurate this is:

- Removing the query argument based named server spawning -- it helped me to test and prototype the code changes faster, that's great, but as you already correctly argued it wouldn't serve any purpose in master.
- Fix the git rebase, since right now in my forked branch the commit history is messed up (I might have to close this PR and re-open a new one). 
- Better indents/code styling for the HTML and Javascript code? I don't know what the JupyterHub style conventions for that are.
- Remove any unnecessary debug logging that I might have forgotten to get rid of.

minrk:
I think that we can work on the design a bit after it's merged, but I think this should be just about set. If you want to cleanup the commits, feel free to rebase and force push to this branch. There is never a need to open a new pull request.

minrk:
@krinsman did you want to do the commit cleanup? I can do that and merge if you don't have time.

krinsman:
@minrk Yes you're right, I want to do it but don't have the time. If you're able to do that, then it would be greatly appreciated, although I feel terrible about burdening you with something that is my fault.

minrk:
@krinsman not a problem! I'll swatch this PR and merge it, then we can proceed to do some UI work.

minrk:
Landing this one after squashing. Thanks, @krinsman!

choldgraf:

https://discourse.jupyter.org/t/how-do-i-enable-multi-server-ui-with-jupyterhub-on-kubernetes/618/1

ablekh:
@choldgraf I have submitted a relevant ticket a couple of weeks ago (still Open/Unresolved): https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1186.

It would be great to hear from @minrk , @yuvipanda or others who could provide some answers on this ...

animesh2706:
Hello contributors,
I am getting 404 after creating named server on jupyterhub(1.0.0) with dockerspawner. 
In Jupyterhub_config.py, I have made allow named server as True,  is there any other configuration needed? I can see the list of named server but while clicking on it, I get 404.

Any help is appreciated. Thanks.

PR #2122, opened 2018-08-31
Address the pylint warnings

SivaMaplelabs:
Pylint score = 10

### DIFF
--- a/examples/bootstrap-script/jupyterhub_config.py
+++ b/examples/bootstrap-script/jupyterhub_config.py
@@ -1,9 +1,14 @@
-# Example for a Spawner.pre_spawn_hook
-# create a directory for the user before the spawner starts
-
+"""
+Example for a Spawner.pre_spawn_hook
+create a directory for the user before the spawner starts
+"""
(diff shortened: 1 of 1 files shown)


SivaMaplelabs:
@willingc @minrk Please review. Thanks

examples/bootstrap-script/pylintrc
```diff
@@ -0,0 +1,378 @@
+[MASTER]
```
minrk: Was this file committed by accident? I don't think we should have a full pylintrc in one example directory.

minrk: Instead of adding this large config file that's not part of the example, a better fix is probably to add this line to the top:

```python
c = get_config()  # noqa: F821
```

This ensures the variable is defined and the linter satisfied for flake8 (the linter used in this repo)

SivaMaplelabs: @minrk Have changed. Thanks

minrk:
Thanks!

PR #2127, opened 2018-09-04
switch to oauthlib from python-oauth2

### LINKED ISSUE #1849: switch to oauthlib
I was investigating implementing an oauth confirmation page and found some significant impediments in our current oauth library, which isn't a very active project.

I think we should switch to [oauthlib](http://oauthlib.readthedocs.io/en/latest/) which is a more widely used oauth implementation library. We would need to do most of the same things we have now, just translated to a slightly different API, but it shouldn't be too much.

This should be a prerequisite for oauth scopes and login confirmation.

This is for post-0.9.

minrk:
This gives us a better level API to access things like oauth scopes and rendering a confirmation page,
which we are going to need for requesting scopes beyond the current scope of 'identify-only'.

This should pave the way to eliminating the distinction between OAuth tokens and API tokens.

Still lots of details to get through, but the basic login process works at this point.

TODO:

- [x] update requirements
- [x] implement oauthlib provider 
- [x] implement oauth confirmation page (only for authenticating services and cross-user requests)
- [x] update tests
- [x] error handling

TODO (not in this PR):

- [ ] implement oauth scopes (#1057)
- [ ] eliminate the separation of APIToken and OAuthAccessToken. The only distinction now is that APITokens have 'all' scopes and OAuthAccessTokens only have 'identify'. If we implemented scopes, these could be tracked in a single APIToken table.

closes #1849

### DIFF
--- a/jupyterhub/apihandlers/auth.py
+++ b/jupyterhub/apihandlers/auth.py
@@ -144,35 +144,89 @@ class OAuthHandler(BaseHandler):
             ._replace(query=urlencode(query_list))
         )
 
+    def add_credentials(self, credentials=None):
+        """Add oauth credentials
 
-class OAuthAuthorizeHandler(OAuthHandler):
-    """Implement OAuth provider handlers
+        Adds user, session_id, client to oauth credentials
+        """
+        if credentials is None:
+            credentials = {}
+        else:
+            credentials = credentials.copy()
 
-    OAuth2Handler sets `self.provider` in initialize,
-    but we are already passing the Provider object via settings.
-    """
+        session_id = self.get_session_cookie()
+        if session_id is None:
+            session_id = self.set_session_cookie()
+
+        user = self.get_current_user()
+
+        # Extra credentials we need in the validator
+        credentials.update({
+            'user': user,
+            'handler': self,
+            'session_id': session_id,
+        })
+        return credentials
+
+    def send_oauth_response(self, headers, body, status):
+        """Send oauth response
+
+        headers, body, status  are returned by provider method.
+        """
+        self.set_status(status)
+        for key, value in headers.items():
+            self.set_header(key, value)
+        if body:
+            self.write(body)
+
+
+class OAuthAuthorizeHandler(OAuthHandler, BaseHandler):
+    """Implement OAuth authorization endpoint(s)"""
+
+    def _complete_login(self, uri, headers, scopes, credentials):
+        try:
+            headers, body, status = self.oauth_provider.create_authorization_response(
+                uri, 'POST', '', headers, scopes, credentials)
+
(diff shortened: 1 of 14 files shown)


jupyterhub/apihandlers/auth.py
```diff
@@ -144,35 +144,89 @@ def make_absolute_redirect_uri(self, uri):
...
+            'user': user,
+            'handler': self,
+            'session_id': session_id,
+        })
+        return credentials
+
+    def send_oauth_response(self, headers, body, status):
+        """Send oauth response
+
+        headers, body, status  are returned by provider method.
```
willingc: Perhaps add something like:

This method prepares headers, body, status for an oauth response.
The provider method returns headers, body, status in its response.

jupyterhub/apihandlers/auth.py
```diff
@@ -188,59 +242,38 @@ def get(self):
     @web.authenticated
     def post(self):
         uri, http_method, body, headers = self.extract_oauth_params()
+        referer = self.request.headers.get('Referer', 'no referer')
+        full_url = self.request.full_url()
+        if referer != full_url:
+            self.log.error("OAuth POST from %s != %s", referer, full_url)
+            raise web.HTTPError(403, "Authorization form must come from authorization")
```
willingc: Error message is a bit unclear.

jupyterhub/oauth/provider.py
```diff
@@ -307,7 +314,19 @@ def save_bearer_token(self, token, request, *args, **kwargs):
...
             - Client Credentials grant
         """
-        app_log.debug("Saving bearer token %s", token)
+        log_token = {}
+        log_token.update(token)
+        scopes = token['scope'].split(' ')
+        # TODO:
+        if scopes != ['identify']:
+            raise ValueError("Only 'identify' scope is supported")
+        # redact sensitive keys in log
```
willingc: Idea for future: If redaction is going to be done in various places beyond this method, perhaps have a small utility that is called.

minrk: Good idea! I believe the object being redacted is only used in this method, though.

jupyterhub/apihandlers/auth.py
minrk: Added some detail, thanks!

jupyterhub/apihandlers/auth.py
minrk: Done, thanks!

jupyterhub/oauth/provider.py
willingc: If it's only this method, just leave as is :-)

willingc:
:shipit:

PR #2135, opened 2018-09-06
add Windows case when stopping the CHP

adelcast:
Currently, if there is a zombie CHP, Windows will fail with:

Failed to start proxy
      File "c:\\Python\3.6\lib\site-packages\jupyterhub\app.py", line 1859, in start
        await self.proxy.start()
      File "c:\\Python\3.6\lib\site-packages\jupyterhub\proxy.py", line 530, in start
        self._check_previous_process()
      File "c:\\Python\3.6\lib\site-packages\jupyterhub\proxy.py", line 490, in _check_previous_process
        for i, sig in enumerate([signal.SIGTERM] * 2 + [signal.SIGKILL]):
    AttributeError: module 'signal' has no attribute 'SIGKILL'

### DIFF
--- a/jupyterhub/proxy.py
+++ b/jupyterhub/proxy.py
@@ -487,9 +487,14 @@ class ConfigurableHTTPProxy(Proxy):
 
         # if we got here, CHP is still running
         self.log.warning("Proxy still running at pid=%s", pid)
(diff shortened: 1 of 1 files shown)


minrk:
Thanks!

PR #2139, opened 2018-09-10
token expiry fixes

minrk:
typos in token expiry:

- omitted from token model (it's in the spec in docs, but wasn't in the model)
- wrong type when sorting oauth tokens on token page could cause token page to not render

Since the token page can be completely broken due to this bug, it's probably worth doing a 0.9.3 with this fix.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,5 +1,6 @@
 -r requirements.txt
 mock
+beautifulsoup4
 codecov
 cryptography
 pytest-cov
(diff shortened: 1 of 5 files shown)


SivaMaplelabs:
@minrk Will get "AttributeError: 'module' object has no attribute 'fromtimestamp'" error with import datetime in jupyterhub/apihandlers/base.py.
Have already raised PR for the same. Please review and merge.
Thanks,

PR #2142, opened 2018-09-11
make `.get_current_user` async

minrk:
User is identified during the `prepare` stage for all requests, to ensure that `.current_user` is defined.

Use the `.current_user` accessor instead of `.get_current_user` throughout to access the result.

Adds placeholder for `.refresh_user_auth` which calls `Authenticator.refresh_user`, a new method for refreshing user authentication. This is the async part of get_current_user.

related to #2117

### DIFF
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -300,21 +300,27 @@ class Authenticator(LoggingConfigurable):
         Args:
             user (User): the user to refresh
         Returns:
-            auth_data (dict or None):
-                The same return value as `.authenticate`.
-                Any values here will refresh the value
-                for the user.
-                This can include updating `.admin` fields
-                or updating `.auth_state`.
-                Return None if the user's auth data has expired,
+            auth_data (bool or dict):
+                Return **True** if auth data for the user is up-to-date
+                and no updates are required.
+
+                Return **False** if the user's auth data has expired,
                 and they should be required to login again.
+
+                Return a **dict** of auth data if some values should be updated.
+                This dict should have the same structure as that returned
+                by :meth:`.authenticate()` when it returns a dict.
+                Any fields present will refresh the value for the user.
+                Any fields not present will be left unchanged.
+                This can include updating `.admin` or `.auth_state` fields.
         """
-        return {'name': user.name}
+        return True
 
     async def authenticate(self, handler, data):
         """Authenticate a user with login form data
 
-        This must be a tornado gen.coroutine.
+        This must be a coroutine.
+
         It must return the username on successful authentication,
         and return None on failed authentication.
 
@@ -328,12 +334,14 @@ class Authenticator(LoggingConfigurable):
             data (dict): The formdata of the login form.
                          The default form has 'username' and 'password' fields.
         Returns:
-            user (str or dict or None): The username of the authenticated user,
+            user (str or dict or None):
+                The username of the authenticated user,
                 or None if Authentication failed.
+
                 The Authenticator may return a dict instead, which MUST have a
-                key 'name' holding the username, and may have two optional keys
-                set - 'auth_state', a dictionary of of auth state that will be
-                persisted; and 'admin', the admin setting value for the user.
+                key `name` holding the username, and MAY have two optional keys
+                set: `auth_state`, a dictionary of of auth state that will be
+                persisted; and `admin`, the admin setting value for the user.
         """
 
     def pre_spawn_start(self, user, spawner):
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -287,6 +287,30 @@ class Authenticator(LoggingConfigurable):
             self.log.warning("User %r not in whitelist.", username)
             return
 
+    async def refresh_user(self, user):
+        """Refresh auth data for a given user
+
+        Allows refreshing or invalidating auth data.
+
+        Only override if your authenticator needs
+        to refresh its data about users once in a while.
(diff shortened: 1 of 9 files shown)


minrk:
ok, this should be set. `.refresh_user_auth` still needs to be fully implemented, but that should be a separate PR, I think.

jupyterhub/auth.py
```diff
@@ -287,6 +287,30 @@ def check_blacklist(self, username):
...
+            auth_data (dict or None):
+                The same return value as `.authenticate`.
+                Any values here will refresh the value
+                for the user.
+                This can include updating `.admin` fields
+                or updating `.auth_state`.
+                Return None if the user's auth data has expired,
+                and they should be required to login again.
+        """
+        return {'name': user.name}
```
betatim: if I don't include a key in the return value then the value remains the old one or should I return the current value?

jupyterhub/handlers/base.py
```diff
@@ -209,6 +223,44 @@ def get_current_user_oauth_token(self):
...
+
+        Calls `authenticator.refresh_user(user)`
+
+        Called at most once per user per request.
+
+        Args:
+            user (User): the user whose auth info is to be refreshed
+            force (bool): force a refresh instead of checking last refresh time
+        Returns:
+            user (User): the user having been refreshed, or None if
```
betatim: missing end of sentence

jupyterhub/handlers/base.py
```diff
@@ -467,36 +533,55 @@ def get_next_url(self, user=None):
...
+    async def auth_to_user(self, authenticated, user=None):
+        """Persist data from .authenticate() or .refresh_user() to the User database
+
+        Args:
+            authenticated(dict): return data from .authenticate or .refresh_user
+            user(User, optional): the User object to refresh, if refreshing
+        Return:
+            user(User): the constructed User object
+        """
+        if isinstance(authenticated, str):
```
betatim: `.authenticate()` returning just a string is the old interface right? Is there a plan to retire it? If yes maybe we should create an issue to come up with a plan and a helper for this pattern that also produces a warning?

jupyterhub/handlers/base.py
```diff
@@ -51,6 +51,20 @@
 class BaseHandler(RequestHandler):
     """Base Handler class with access to common methods and properties."""
 
+    async def prepare(self):
+        """Identify the user during the prepare stage of each request
+
+        During requests, the current user may be retrieved via
```
willingc: During prepare stage of request, the user is identified by get_current_user. 
During a request, the current...

jupyterhub/handlers/base.py
willingc: ... the user needs to login again to authenticate.

jupyterhub/handlers/base.py
minrk: I don't think we should ever drop support for authenticate returning a string, so this should always work without warning. That's the simplest way to implement an authenticator, and is all you need in most cases.

jupyterhub/auth.py
minrk: omitted keys are left alone. Updated the docstring to clarify.

minrk:
Thanks for the reviews! I clarified some comments and docstrings. I think the simple/common case for refresh_user is a boolean "is it up-to-date or do I need to login again", so I switched the return type for that to be bool, to match the logic. Only a dict can be returned to update fields.

adelcast:
hi @minrk . I am using a a custom Authenticator (uses the HTTP header Remote-User to login) that directly login users if the HTTP header is set (Jupyterhub is behind a reverse proxy). My problem is that I need to invalidate the cookies if the Remote-User is different than the current_user. If the apache cookie expires, the user may login with a different user, but jupyterhub will still use the old-user cookie. 

Looking at this PR, seems like refresh_user_auth can do what I want. However, the code currently doesn't call refresh_user_auth with force=True. What is the design pattern to follow to make sure refresh_user_auth is called periodically? I had to manually modify the call to refresh_user_auth (on base.py) to set force=True.

appreciate any pointers...

minrk:
refresh_user_auth isn't in use yet, but the placeholder is there. The last piece to make this useful is adding a configurable age to auth info [here](https://github.com/jupyterhub/jupyterhub/pull/2142/files#diff-09e41de2f86321f654d5d9f6a9205a85R246) so that when `force=False`, only "old enough" auth is refreshed. Once this is present, setting this timeout to 0 should require auth to be refreshed on every request.

However, there is a piece missing for what you need because refresh_user takes only the user to be refreshed, it does not receive the Handler, which you would need to check the headers.

Could you open a new Issue describing your use case? Then maybe we can write a test that makes sure to implement something like you need so that we can be sure it works.

adelcast:
yes, forgot to mention that I had to change refresh_user to add a handler argument. I have the change working, so I can submit a PR, if that helps. The setting should work for my case (set it to 0), it gives me the external knob to make sure refresh_auth is called on every request.

Do you still want me to open a new issue? between the PR I will be submitting (add handler param to refresh_auth) and the setting for the auth age I think my case is fulfilled...thanks for quick response!

minrk:
Yes, feel free to open an issue or PR. We should make sure to track this for the release.

zflamig:
Just wanted to +1 @adelcast's use case because I am doing the same thing and ran into the same problem. Still working out exactly how to resolve it... looks like the current code has to be modified to default force=True?

PR #2155, opened 2018-09-19
fix oauth client cleanup

### LINKED ISSUE #2125: JH 0.8.1 -> 0.9.2 upgrade `invalid_redirect_uri` error for running users
## Z2JH deployment upgrade failure for running users
We recently released a new Z2JH Helm chart. Apparently, users with a running pod revisiting their server through the hub got to see the error below, if an upgrade from `v0.6` to `0.7.0` had been made in between visits of the running pod. The suspected change that could have caused this is that JupyterHub went from `0.8.1` to `0.9.2`, and the `--upgrade-db` flag (if I recall correctly) did some work when the JupyterHub started again.

```
{“error”: “invalid_redirect_uri”, “error_description”: “Invalid redirect URI”}
```

@minrk had a clue as to why this happened
> It means the db upgrade resulted in invalid oauth for the running users.
I think this might be something to fix in the jupyterhub alembic upgrades

minrk:
- delete oauth clients for servers when they shutdown
- avoid deleting oauth clients for servers still running across an 0.8 -> 0.9 upgrade, when the oauth client ids changed from `user-NAME` to `jupyterhub-user-NAME`

closes #2125

We should probably make an 0.9.4 with this fix, since it greatly impacts upgrade from 0.8 to 0.9 with running servers (as in, it's impossible without this).

### DIFF
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -9,6 +9,12 @@ command line for details.
 
 ## 0.9
 
+### [0.9.4] 2018-09-20
+
+JupyterHub 0.9.4 fixes a single issue that required
+all running user servers to be restarted when performing an upgrade
+from 0.8 to 0.9.
+
 ### [0.9.3] 2018-09-12
 
 JupyterHub 0.9.3 contains small bugfixes and improvements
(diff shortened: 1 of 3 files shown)


consideRatio:
Thanks for addressing this issue @minrk !! :heart:

Issue #2160, opened 2018-09-20
Revamp contributor docs


yuvipanda:
JupyterHub is participating in Outreachy this year (woohoo!). To make the lives of our applicants easier, we should improve & expand out developer contributing documentation.

This issue tracks the various things that can be improved.

- [ ] More details on setting up your development environment (#2159)
- [ ] Using SimpleLocalProcessSpawner for development, so you don't need to create users on your system
- [ ] Using DummyAuthenticator for development, so you don't need to create users on your system
- [ ] Better links to various underlying technologies we use + tutorials on how to learn them.
- [ ] Unified description of where to communicate with the team

yuvipanda:
Trying to figure out what are 'foundational technologies' for JupyterHub.

1. Python3 
2. Tornado (and async programming)
3. Jinja2
4. Whatever the spawner you are using is
5. Whatever the authenticator you are using is
https://github.com/jupyterhub/jupyterhub/pull/2161
Anything I've missed?

minrk:
Additional topics that you may need, depending on where in the code you are working:

- sql/sqlachemy (for the low-level / query stuff)
- javascript (frontend actions)
- html/css (UI)

Do you think dummy/simplespawner for testing ought to be shipped as part of jupyterhub?

mriedem:
#2961 is related to this. As a new contributor trying to get tests passing locally I've been going between https://github.com/jupyterhub/jupyterhub/blob/master/CONTRIBUTING.md and https://jupyterhub.readthedocs.io/en/stable/contributing/setup.html#id2 and running into some issues with both. Note that they are slightly different (one doesn't mention installing `requirements.txt`, only `dev-requirements.txt`). Using `tox` would help with that since most things would be automated. Basically consider taking the scripts portion of the travis config and putting that into a `tox.ini` and then you could update the travis config to just run the tox environment for each travis build. Then local dev / test and CI are aligned. I've also had issues getting `SimpleLocalProcessSpawner` to work locally so I had to fallback to the default `LocalProcessSpawner` (I'm not sure what the issue was but if it makes a difference I wasn't running `jupyterhub` with `sudo`). It's also kind of hard to figure out from those contributing docs that you need sqlite packages installed for tests to work by default, and if the tests setup sqlite automatically (the database, not the package install).

PR #2161, opened 2018-09-20
Add notes on simplespawner & dummyauthenticator

yuvipanda:
The default spawner & auth implementation on JupyterHub requires
users have root on their systems and create new unix users for
each JupyterHub user. This gets complex & hard quickly.

Recommend using simplespawner & dummyauthenticator to 
get started instead. This is what I do, and I think it works well.

Ref #2160

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -108,6 +108,26 @@ to make that happen.
 
 Happy developing!
 
+### Using DummyAuthenticator & SimpleSpawner
+
+The default JupyterHub authenticator & spawner require your
+system to have user accounts for each user you want to log in to
+JupyterHub as. This requires you have root / `sudo` access to your
+machine. It also gets cumbersome & insecure quickly as you
+do development.
+
+[DummyAuthenticator](https://github.com/jupyterhub/dummyauthenticator) allows
+you to log in with any username & password.
+[SimpleSpawner](https://github.com/jupyterhub/simplespawner) lets you start
+servers without having to create unix users for each JupyterHub user.
+Together, these make it much easier to test JupyterHub.
+
+If you are working on parts of JupyterHub that are common to all
+authenticators & spawners, we recommend using both DummyAuthenticator &
+SimpleSpawner. If you are working on just authenticator related parts, use
+only SimpleSpawner. Similarly, if you are working on just spawner
+related parts, use only DummyAuthenticator.
+
 ## Troubleshooting a development install
 
 ### `lessc` not found
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -110,18 +110,22 @@ Happy developing!
 
 ### Using DummyAuthenticator & SimpleSpawner
 
+To simplify testing of JupyterHub,
+it's helpful to use DummyAuthenticator instead of the default JupyterHub authenticator
+and SimpleSpawner instead of the default spawner.
+
 The default JupyterHub [authenticator](https://jupyterhub.readthedocs.io/en/stable/reference/authenticators.html#the-default-pam-authenticator)
 & [spawner](https://jupyterhub.readthedocs.io/en/stable/api/spawner.html#localprocessspawner)
 require your system to have user accounts for each user you want to log in to
 JupyterHub as.
 
 [DummyAuthenticator](https://github.com/jupyterhub/dummyauthenticator) allows
-you to log in with any username & password.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -108,6 +108,26 @@ to make that happen.
 
 Happy developing!
 
+### Using DummyAuthenticator & SimpleSpawner
+
+The default JupyterHub authenticator & spawner require your
+system to have user accounts for each user you want to log in to
+JupyterHub as. This requires you have root / `sudo` access to your
```
consideRatio: Perhaps add a link to the default authenticators / spawners as this text may raise the question about what those are called. I figure a link would be less intrusive to the narration than dropping names of stuff.

> The default JupyterHub [authenticator](https://jupyterhub.readthedocs.io/en/stable/reference/authenticators.html#the-default-pam-authenticator) & [spawner](https://jupyterhub.readthedocs.io/en/stable/api/spawner.html#localprocessspawner) require your
system to have user accounts for each user you want to log in to
JupyterHub as.

---

I'm not sure, but one could consider adding a one to three sentences before this clarifying the point of an authenticator and spawner. Perhaps the links are enough though, yeah, probably, yepp, lets not repeat ourselves too much.

yuvipanda: I agree! Will do when I'm at a computer :) feel free to push a change too if you wanna

betatim:
It would be good to include the config/command-line snippet + `pip install ...` needed to actually do this for the one "best" option (maybe dummy auth?).

My thinking is that as newcomer to the project I know nothing about how the config system works nor what the trade-offs are between the options. We assume a fairly low level of Python expertise at the start of this guide by explaining how to get Python so we should keep serving that audience.

yuvipanda:
@betatim thanks for the feedback! IMO, we can do that as a different PR. The important thing for documentation PRs is that they are incremental & happen quickly, since we can have many of them. I'll happily make the changes as a new PR, so that people who land up in CONTRIBUTING.md today will have a better experience than they have right now.

betatim:
Lets merge this if you open a new PR for the change instead of a new issue.

yuvipanda:
@betatim I opened https://github.com/jupyterhub/jupyterhub/issues/2163. I'll make a PR!

CONTRIBUTING.md
```diff
@@ -108,6 +108,25 @@ to make that happen.
...
+### Using DummyAuthenticator & SimpleSpawner
+
+The default JupyterHub [authenticator](https://jupyterhub.readthedocs.io/en/stable/reference/authenticators.html#the-default-pam-authenticator)
+& [spawner](https://jupyterhub.readthedocs.io/en/stable/api/spawner.html#localprocessspawner)
+require your system to have user accounts for each user you want to log in to
+JupyterHub as.
+
+[DummyAuthenticator](https://github.com/jupyterhub/dummyauthenticator) allows
+you to log in with any username & password.
+[SimpleSpawner](https://github.com/jupyterhub/simplespawner) lets you start
```
willingc: While, SimpleSpawner starts servers without having to create a unix user for each JupyterHub user. ~~Together...JupyterHub~~

CONTRIBUTING.md
```diff
@@ -108,6 +108,25 @@ to make that happen.
 
 Happy developing!
 
+### Using DummyAuthenticator & SimpleSpawner
+
+The default JupyterHub [authenticator](https://jupyterhub.readthedocs.io/en/stable/reference/authenticators.html#the-default-pam-authenticator)
```
willingc: To simplify testing of JupyterHub, it's helpful to use DummyAuthenticator instead of the default JupyterHub authenticator and SimpleSpawner instead of the default spawner.

CONTRIBUTING.md
```diff
@@ -108,6 +108,25 @@ to make that happen.
...
+require your system to have user accounts for each user you want to log in to
+JupyterHub as.
+
+[DummyAuthenticator](https://github.com/jupyterhub/dummyauthenticator) allows
+you to log in with any username & password.
+[SimpleSpawner](https://github.com/jupyterhub/simplespawner) lets you start
+servers without having to create unix users for each JupyterHub user.
+Together, these make it much easier to test JupyterHub.
+
+If you are working on parts of JupyterHub that are common to all
```
willingc: Tip: If you are ...

minrk:
Merged with @willingc's notes applied. Thanks!

PR #2164, opened 2018-09-21
Add user name to control panel in JupyterHub 

### LINKED ISSUE #2157: Display user name in Control Panel
The JupyterHub Control Panel currently only shows actions for stoppin & starting servers. It would be great if it could also show the current user name, since this is not obvious with a bunch of Authenticators (such as LTI).

For reference, https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/handlers/pages.py#L43 is the handler that renders the page. You can access the page by clicking 'Control Panel' on the top right after logging in to your JupyterHub.

sashafierce:
Fixes #2157 . Now username is displayed next to the logout button in control panel.

screenshot_2018-09-21 jupyterhub

### DIFF
--- a/share/jupyterhub/templates/page.html
+++ b/share/jupyterhub/templates/page.html
@@ -123,6 +123,7 @@
               {% block login_widget %}
                 <span id="login_widget">
                   {% if user %}
+		    <p class="navbar-text">{{user.name}}</p>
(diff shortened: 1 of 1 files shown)


minrk:
Awesome, that works great!

PR #2165, opened 2018-09-21
Fix content-type on API endpoints

### LINKED ISSUE #2146: API handlers are returning text/html responses
I've recently upgrade to v0.9.3 from v0.8.0. When I try to make a request to any of the API endpoints (`/hub/api/info` - for instance) I'm getting back the header Content-Type set to text/html instead of application/json as it used to be prior the upgrade.

Should I use some configuration parameter, maybe related to tornado?

Thanks!

minrk:
and includes content-type header checks in tests to catch regressions

closes #2146


minrk:
Released 0.9.4 with this and #2155

PR #2166, opened 2018-09-21
add sample configuration that loads dummyauthenticator and simplespawner

### LINKED ISSUE #2163: Inline documentation about installing simplespawner & dummyauthenticator in CONTRIBUTING.md
https://github.com/jupyterhub/jupyterhub/pull/2161#issuecomment-423422471

> It would be good to include the config/command-line snippet + pip install ... needed to actually do this for the one "best" option (maybe dummy auth?).

> My thinking is that as newcomer to the project I know nothing about how the config system works nor what the trade-offs are between the options. We assume a fairly low level of Python expertise at the start of this guide by explaining how to get Python so we should keep serving that audience.

minrk:
and inline installation of the packages (closes #2163)

Running:

    jupyterhub -f testing/jupyterhub_config.py

loads dummyauth/simplespawn

If they aren't present, the installation command is displayed.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -114,6 +114,12 @@ To simplify testing of JupyterHub,
 it's helpful to use DummyAuthenticator instead of the default JupyterHub authenticator
 and SimpleSpawner instead of the default spawner.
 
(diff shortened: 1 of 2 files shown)


yuvipanda:
Thanks, @minrk

PR #2177, opened 2018-09-24
Example homedir bootstrapping using `mkhomedir_helper`

nils-werner:
Many Linux distributions provide a script that is responsible for user homedir bootstrapping: `/sbin/mkhomedir_helper`. It creates a homedir from `/etc/skel` and adjusts all permissions in one single call.

This PR adds that script as an option in the examples README.


minrk:
Nice!

PR #2179, opened 2018-09-24
back to conda on rtd

minrk:
We can't build the docs without conda on rtd because we need it to get npm for rendering the rest-api docs.

Instead, workaround the sphinx 1.8 regression by skipping the broken PDF builds.

Unfortunately, we can't currently pin the sphinx version when using conda due to https://github.com/rtfd/readthedocs.org/issues/3829

cc @willingc

### DIFF
--- a/readthedocs.yml
+++ b/readthedocs.yml
@@ -4,3 +4,8 @@ conda:
      file: docs/environment.yml
 python:
   version: 3
+formats:
+  - htmlzip
+  - epub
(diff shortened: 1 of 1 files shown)


willingc:
Sounds good @minrk. I ran out of time this morning to address it.  It did seem to build the api correctly using pip when removing the lines about the shell execution in the else statement in conf.py. Totally cool with removing the PDF build and using conda though. Probably prefer it.

PR #2194, opened 2018-09-26
Use python3 -m pip explicitly in docs

### LINKED ISSUE #2191: pip vs pip3 in CONTRIBUTING.md
There is a mix of using pip and pip3 in the CONTRIBUTING.md docs. I followed the instructions on a mac, and found the pip3 to fail but pip to work after installing miniconda according to instructions.

Go with `pip` and `python` over `pip3` and `python3`?

I'm confused about when `python3` is available btw, I'm also confused about other things =D See below. I'm ready to accept this confusion though.

image

yuvipanda:
- Add a documentation convention listing this too.

Fixes #2191

### DIFF
--- a/docs/source/contributing/docs.rst
+++ b/docs/source/contributing/docs.rst
@@ -10,7 +10,6 @@ you get set up on how to contribute documentation to JupyterHub.
 Building documentation locally
 ==============================
 
-
 We use `sphinx <http://sphinx-doc.org>`_ to build our documentation. It takes
 our documentation source files (written in `markdown
 <https://daringfireball.net/projects/markdown/>`_ or `reStructuredText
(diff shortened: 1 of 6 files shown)


choldgraf:
This seems reasonable to me! Gonna merge this later today unless somebody disagrees :-)

docs/source/contributing/docs.rst
```diff
@@ -46,4 +45,35 @@ change renders correctly, it is good practice to test it locally.
...
+
+1. ``pip``
+    This points to python3 pip if you are in an appropriate environment,
+    but in some cases it can actually point to a python2 pip.
+
+2. ``pip3``
+   When present, this always points to a python3 pip, most likely in
+   the same environment as your python3.
+
+3. ``python3 -m pip``
```
willingc: This should really be the first item here. I would remove 1 and 2 and refer to the pip documentation.

choldgraf: that's a good point, +1

betatim: Maybe even remove the other options and write "There are many ways to run `pip`, this is our preferred option: `python -m pip`"?

PR #2197, opened 2018-09-27
Expand DummyAuthenticator docstring

minrk:
and replace default None password with more consistent Falsy empty string for no password set.


yuvipanda:
Thanks @minrk!

Issue #2168, opened 2018-09-21
Move simplespawner into JupyterHub


yuvipanda:
https://github.com/jupyterhub/simplespawner is very useful when developing & debugging JupyterHub. We currently recommend using it in our CONTRIBUTING.md.

@minrk suggested that we move it into JupyterHub itself. The code is pretty small, and I think it makes a lot of sense for us to do that.

Tasks here would be:

1. Move the code for simplespawner into JupyterHub
2. Write a few tests for it
3. Add some documentation for it.

yuvipanda:
This is a good sized task for an outreachy participant.

minrk:
Closed by #2198. Thanks @kshitija08!

PR #2198, opened 2018-09-27
spawners/simplespawner.py

kshitija08:
Move simplespawner into JupyterHub #2168
@yuvipanda @minrk My first pull request. Kindly, let me know if some changes need to be made. I have made a new directory spawners in the jupyterhub master branch and added the simplespawner.py to this new directory.

### DIFF
--- a/testing/jupyterhub_config.py
+++ b/testing/jupyterhub_config.py
@@ -9,10 +9,6 @@ to enable testing without administrative privileges.
 c = get_config() # noqa
 c.JupyterHub.authenticator_class = DummyAuthenticator
 
-try:
-    from jupyterhub.spawners import SimpleSpawner
-    c.JupyterHub.spawner_class = SimpleSpawner
-except ImportError:
-    print("simplespawner not available. Try: `pip install jupyterhub-simplespawner`")
-else:
-    c.JupyterHub.spawner_class = SimpleLocalProcessSpawner
+from jupyterhub.spawners import SimpleSpawner
+c.JupyterHub.spawner_class = SimpleSpawner
+
--- a/testing/jupyterhub_config.py
+++ b/testing/jupyterhub_config.py
@@ -10,7 +10,7 @@ c = get_config() # noqa
 c.JupyterHub.authenticator_class = DummyAuthenticator
 
 try:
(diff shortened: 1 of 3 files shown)


minrk:
This is great! Because it is a Python package, an `__init__.py` is needed in the `spawners` directory to make it importable, and it should contain:

```python
from .simplespawner import SimpleSpawner
```

Additionally, would you like to update the `testing/jupyterhub_config.py` to import simplespawner from this new location? It would be `from jupyterhub.spawners import SimpleSpawner` instead of the current import. The try/except can also go away, because it's now always included with jupyterhub.

kshitija08:
@minrk Thank you. May I know what will the __int__.py would contain ? 
I will update the testing/jupyterhub_config.py file.

testing/jupyterhub_config.py
```diff
@@ -10,7 +10,7 @@
 c.JupyterHub.authenticator_class = DummyAuthenticator
 
 try:
-    from simplespawner import SimpleLocalProcessSpawner
+    from jupyterhub.spawners import SimpleSpawner
 except ImportError:
```
minrk: Almost there! This can now remove the `try:...except:` because the import can't fail anymore:

```python
from jupyterhub.spawners import SimpleSpawner
c.JupyterHub.spawner_class = SimpleSpawner
```

kshitija08: @minrk I have done the suggested changes. Kindly, have a look.

minrk: Thanks! Sorry, I think I was unclear. I mean that the whole `try:...` block could be removed because there is no longer a need for the `jupyterhub-singleuser` package.

So instead of:

```python
try:
    from jupyterhub.spawners import ...
except ImportError:
    ....
```

it should have no try or except at all:

```python
from jupyterhub.spawners import ...
c.JupyterHub....
```

minrk:
@kshitija08 perfect, thanks!

kshitija08:
@minrk Thank you so much for your support and guidance.

willingc:
Congrats @kshitija08 on your first pull request too. :tada:

PR #2205, opened 2018-09-28
bump bootstrap version

willingc:
WIP - We will also need to move from less to sass should we wish to support bootstrap 4.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "prettier": "^1.14.2"
   },
   "dependencies": {
-    "bootstrap": "^3.3.7",
+    "bootstrap": "^4.1.3",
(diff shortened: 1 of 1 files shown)


minrk:
I've looked into the relevant issues, and I don't believe we are affected here, so this shouldn't be urgent for us. Additionally, there will be a 3.4 release of Bootstrap (hopefully shortly), so I think we can do a lower-impact bump to bs 3.4 when it ships, then we can do the work of migrating to to 4.0 and SASS later.

3.4 release PR: https://github.com/twbs/bootstrap/pull/27288

willingc:
Sounds good @minrk. I'll leave this open and we can amend to 3.4 when it is available.

PR #2207, opened 2018-09-28
add server_poll_duration_seconds prometheus metric

### LINKED ISSUE #2185: Add server_poll_duration_seconds prometheus metric
We use [prometheus](https://github.com/prometheus/client_python) to expose metrics about JupyterHub operations. Currently we expose HTTP request stats and time it takes to spawn servers. We should expose more metrics!

One useful metric is 'time taken for JupyterHub to poll if server is running'. Periodically (and at startup), JupyterHub goes through the list of servers it thinks are running & verifies that they *are* running. It does so by polling them through the spawner. It would be great to record how much time this takes.

 It should be a [histogram](https://github.com/prometheus/client_python#histogram) metric. We should observe the timing whenever a server is stopped. This is similar to the SERVER_SPAWN_DURATION_SECONDS metric, which is a histogram that observes the amount of time servers take to start.

- [ ] Declare the metric in metrics.py
- [ ] Observe time it takes for a spawner poll to happen

GeorgianaElena:
Hi! I added a server_poll_duration_seconds metric (regarding https://github.com/jupyterhub/outreachy/issues/16) but I am not sure if I should have labeled the histogram based on the status of the polling operation.

closes #2185

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -32,6 +32,7 @@ from ..utils import maybe_future, url_path_join
 from ..metrics import (
     SERVER_SPAWN_DURATION_SECONDS, ServerSpawnStatus,
     PROXY_ADD_DURATION_SECONDS, ProxyAddStatus,
+    SERVER_POLL_DURATION_SECONDS,
     RUNNING_SERVERS
 )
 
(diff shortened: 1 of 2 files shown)


minrk:
I think it might be useful to use the status as a label, since we may want to distinguish the time to check for a running object (status = None) vs one that has stopped (integer exit code).

GeorgianaElena:
Hi @minrk ! I updated the PR by adding the status as a label for the metric. Please let me know if now it's ok and thanks for the feedback.

GeorgianaElena:
The tests are failing. From my understanding of the logs, it doesn't seem related to metrics but do you think my changes messed up something?

GeorgianaElena:
@minrk, thanks a lot for helping me with this!

minrk:
Excellent, thanks!

consideRatio:
@GeorgianaElena wieeee :tada: !!

PR #2210, opened 2018-10-02
Replace python with sys.executable in most service examples

nils-werner:
There are two ways of executing `jupyterhub` in a virtualenv:

1. Activate the virtualenv first

        source /virtualenv/bin/activate && jupyterhub

    or

2. Run it the executable directly

        /virtualenv/bin/jupyterhub

In both cases jupyterhub will work identically.

However, when using method 2, `$PATH` will not be modified, and calling `python` inside your program might accidentally leave the virtualenv, as can easily be tested using

    from subprocess import call
    call(["which", "python"])

and running it like

    /virtualenv/bin/python test.py                          # yields /usr/bin/python
    source /virtualenv/bin/activate && python test.py       # yields /virtualenv/bin/python

One way to circumvent it is to call `sys.executable` instead of `'python'`.

This PR replaces all relevant instances of `'python'` with `sys.executable` in the `c.JupyterHub.services` examples.

### DIFF
--- a/docs/source/getting-started/services-basics.md
+++ b/docs/source/getting-started/services-basics.md
@@ -88,7 +88,7 @@ c.JupyterHub.services = [
     {
         'name': 'cull-idle',
         'admin': True,
-        'command': 'python3 cull_idle_servers.py --timeout=3600'.split(),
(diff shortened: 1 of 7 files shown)


minrk:
Thanks!

Issue #2184, opened 2018-09-24
Add total_users prometheus metric


yuvipanda:
We use [prometheus](https://github.com/prometheus/client_python) to expose metrics about JupyterHub operations. Currently we expose HTTP request stats and time it takes to spawn servers. We should expose more metrics!

One useful metric is 'total users'. It should be a [gauge](https://github.com/prometheus/client_python#gauge) metric. Incremented when a user is created, and decremented when user is deleted. We should also set it to an appropriate value when JupyterHub starts.

- [x] Declare the metric in metrics.py
- [x] Increment it when user is created
- [x] Decrement it when user is deleted
- [x] Set it to appropriate value when JupyterHub itself starts

kshitija08:
Hi @yuvipanda
I am Kshitija Saharan, a final year undergraduate from IIT Roorkee. I came to know about this project from outreachy's project list. I am totally new to open source development , but I know programming in Python. I have previously worked in data science projects for image classification.
I would love to learn and contribute to this issue. Kindly, help me with the starting of it.

yuvipanda:
@kshitija08 that's great! You might want to look at [this recent PR](https://github.com/jupyterhub/jupyterhub/pull/2176) as an example for adding a new metric, and look at the SERVER_SPAWN_DURATION metric already in the code for an existing timer that is a histogram. You'll probably want to start with the [contribution guide](https://jupyterhub.readthedocs.io/en/latest/index.html#contributing) for JupyterHub.

yuvipanda:
@yuvipanda can you tell me what you mean by 'the appropriate value would be zero'? If you can open a pull request here with what you have that might be easier to discuss!

PR #2215, opened 2018-10-03
Adding TOTAL_USERS metric

kshitija08:
@yuvipanda @minrk I have made a PR. Kindly check. #2184 https://github.com/jupyterhub/jupyterhub/issues/2184


minrk:
Thanks for the PR! This is tracking the number of running servers, however, not the total number of users that exist. The total number of users is the number of `orm.User` items that exist.

The number should be increased when new users are created ([here](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/user.py#L38) and decreased when users are deleted [here](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/user.py#L89).

kshitija08:
@minrk I have added the TOTAL_USERS metric in user.py.
Increment is done when user is created and decrement when it is deleted.

minrk:
That's great, thank you @kshitija08!

Issue #2183, opened 2018-09-24
Add 'check_routes_duration_seconds' metric


yuvipanda:
We use [prometheus](https://github.com/prometheus/client_python) to expose metrics about JupyterHub operations. Currently we expose HTTP request stats and time it takes to spawn servers. We should expose more metrics!

One useful metric is 'Time taken to validate all routes in proxy'. JupyterHub occasionally checks the routes in the proxy, making sure there are no routes that shouldn't be there & that the routes that should be there are present. This happens both at application startup & at regular intervals. We should record how long it takes.

It should be a [histogram](https://github.com/prometheus/client_python#histogram) metric. We should observe the timing whenever a server is stopped. This is similar to the SERVER_SPAWN_DURATION_SECONDS metric, which is a histogram that observes the amount of time servers take to start.

- [ ] Declare the metric in metrics.py
- [ ] Observe time it takes for check_routes function in proxy.py to run

minrk:
@kshitija08 that's great! You might want to look at [this recent PR](https://github.com/jupyterhub/jupyterhub/pull/2176) as an example for adding a new metric, and look at the SERVER_SPAWN_DURATION metric already in the code for an existing timer that is a histogram. You'll probably want to start with the [contribution guide](https://jupyterhub.readthedocs.io/en/latest/index.html#contributing) for JupyterHub.

Issue #2216, opened 2018-10-04
Question: AD/LDAP integration and home users


rainmanh:
Hi,

I have currently done an AD integration of jupyterhub using as a client the classic ldap component.

After a couple of days of lot of fun, I managed to have it running but now I got some questions user home related.

While by default the  AD users got their home defined in *nix attributes as : `/home/username`

I would like to have the specific home user directory for the jupyter user somewhere else so `/opt/home/` instead.

I have spent quite a while playing around with the following directives but I am not having much luck..

```
c.LDAPAuthenticator.create_user_home_dir = True
c.LDAPAuthenticator.create_user_home_dir_cmd = ['mkhomedir_helper']
c.NotebookApp.default_url = 'tree'
```

dhirschfeld:
If you have a problem with `LDAPAuthenticator` you're probably best to ask over at that repo:
https://github.com/jupyterhub/ldapauthenticator

Where do you see the `create_user_home_dir` configuration? AFAICS it doesn't exist in either `LDAPAuthenticator` or `JupyterHub`?

rainmanh:
Hi,

True, apologies for posting in here.

The configuration is coming from here:

https://pypi.org/project/jupyterhub-ldap-authenticator/

dhirschfeld:
It looks like that page isn't up to date. I'd refer to the GitHub README for the correct config.

PR #2219, opened 2018-10-05
Adding CHECK_ROUTES_DURATION_SECONDS metric #2183

kshitija08:
A new **histogram** metric is defined in jupyterhub/metrics.py which is used to calculate "Time taken to validate all routes in proxy". 
A timer is set in jupyterhub/proxy.py which starts at the time of the function call "check_routes" and stops after the execution of the function.

**time.perf_counter() → float** is used for calculating the time. 
Return the value (in fractional seconds) of a performance counter, i.e. a clock with the highest available resolution to measure a short duration. It does include time elapsed during sleep and is system-wide. The reference point of the returned value is undefined, so that only the difference between the results of consecutive calls is valid.
@minrk @yuvipanda I have made the PR. Kindly check.

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -33,7 +33,7 @@ from ..metrics import (
     SERVER_SPAWN_DURATION_SECONDS, ServerSpawnStatus,
     PROXY_ADD_DURATION_SECONDS, ProxyAddStatus,
     SERVER_POLL_DURATION_SECONDS, ServerPollStatus,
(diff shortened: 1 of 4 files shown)


minrk:
Excellent, thank you! The test failures are unrelated. Merging this will also merge #2216

kshitija08:
@minrk Thank you :) Can you tell me why some of the tests failed ? I was just curious to know.

kshitija08:
#2183

Issue #2171, opened 2018-09-21
Add server_stop_duration_seconds prometheus metric


yuvipanda:
We use [prometheus](https://github.com/prometheus/client_python) to expose metrics about JupyterHub operations. Currently we expose HTTP request stats and time it takes to spawn servers. We should expose more metrics!

One useful metric is 'time taken for server to stop'. It should be a [histogram](git) metric. We should observe the timing whenever a server is stopped. This is similar to the SERVER_SPAWN_DURATION_SECONDS metric, which is a histogram that observes the amount of time servers take to start.

- [ ] Declare the metric in metrics.py
- [ ] Observe time it takes for server to stop

yuvipanda:
@glenak1911 go for it! Feel free to ask for help here or on gitter.im/jupyterhub/jupyterhub.

tuhina2020:
hi @yuvipanda @minrk. I would like to contribute as an Outreachy applicant to jupyterhub project 1 "Highly available JupyterHub proxy with Traefik".Is this issue meant for Outreachy applicants ?  Can I get started on this one ?

yuvipanda:
@tuhina2020 welcome! yes, this is meant for outreachy applicants. Please feel free to get started.

Issue #14, opened 2014-08-28
Running kernel became disassociated from dashboard


dsblank:
After running for about 24 hours, a kernel was still running, but the dashboard no longer showed a "shutdown" button, and no longer listed it under the "Running" tab. Looks like the connection between dashboard and kernel timed out, as everything was still working fine.

PR #2220, opened 2018-10-05
Add server_stop_duration_seconds prometheus metric #14

tuhina2020:
@yuvipanda @minrk 

https://github.com/jupyterhub/jupyterhub/issues/2171

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -33,7 +33,7 @@ from ..metrics import (
     SERVER_SPAWN_DURATION_SECONDS, ServerSpawnStatus,
     PROXY_ADD_DURATION_SECONDS, ProxyAddStatus,
     SERVER_POLL_DURATION_SECONDS, ServerPollStatus,
-    RUNNING_SERVERS
+    RUNNING_SERVERS, SERVER_STOP_DURATION_SECONDS, ServerStopStatus
 )
 
 # pattern for the authentication token header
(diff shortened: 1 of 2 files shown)


jupyterhub/handlers/base.py
```diff
@@ -873,10 +873,14 @@ def abort():
...
 
     async def stop_single_user(self, user, server_name=''):
+        start_time = time.perf_counter()
         if server_name not in user.spawners:
             raise KeyError("User %s has no such spawner %r", user.name, server_name)
         spawner = user.spawners[server_name]
         if spawner.pending:
+            SERVER_STOP_DURATION_SECONDS.labels(
+                status=ServerStopStatus.pending
+            ).observe(time.perf_counter() - start_time)
```
minrk: Let's remove this pending record. It doesn't really make sense to record the time it takes to not do anything (the server is not stopping here, the request is being cancelled because a stop is already pending)

jupyterhub/metrics.py
```diff
@@ -100,11 +100,29 @@ def from_status(cls, status):
...
+    'time taken for server stopping operation',
+    ['status'],
+)
+
+class ServerStopStatus(Enum):
+    """
+    Possible values for 'status' label of SERVER_STOP_DURATION_SECONDS
+    """
+    stopped = 'stopped'
+    pending = 'pending'
```
minrk: The only status fields for stop, I think, are 'success' and 'failure'. The failure status should be set in the event of a failure.

tuhina2020:
@minrk please let me know if there are any further changes.

minrk:
@tuhina2020 great, thanks! I think everything should be in order, as soon as the tests pass.

Issue #2105, opened 2018-08-25
/metrics endpoint should be private


shreddd:
**Describe the bug**
Currently the /hub/metrics endpoint is public. It may be more appropriate to have these metrics be protected by a login since this is internal system information. 

**To Reproduce**
1. Go to the /hub/metrics URL on a jupyterhub server without logging in 
2. Metrics will be displayed without requiring a login

**Expected behavior**
Metrics should minimally require a login (and potentially admin access for certain metrics)

**Desktop (please complete the following information):**
 - OS: Any
 - Browser: Any

jhermann:
If you add a local SSL off-loader (e.g. nginx), then auth can be handled there, or else you can block the URL and have some local agent access port 8000 directly (i.e. no metric access from the outside).

yuvipanda:
I'm conflicted here, since these metrics are usually collected by something like Prometheus. Adding auth to it makes that hard. However, I also recognize that this is something many installations consider private info. 

I'd recommend we make an option that lets it be authenticated or not, and turn the default to be authenticated.

minrk:
For implementation reference, to make this authentication configurable, it should:

1. add `authenticate_prometheus = Bool(True, config=True)` trait to the JupyterHub class in `app.py`
2. define a new [auth_decorator](https://github.com/jupyterhub/jupyterhub/blob/0.9.4/jupyterhub/utils.py#L243-L248) that checks if current_user is None and raises 403 *if* current user is None and authenticate_prometheus is True. 
3. apply this new auth decorator to the `get` method on the MetricsHandler

tuhina2020:
Hi @minrk @yuvipanda . Can I get started on this bug ?

consideRatio:
Thanks for your work on this @tuhina2020, I'm closing this as it is my understanding that your PR resolved this issue!

PR #2224, opened 2018-10-07
add metrics_authentication decorator to /metrics API #2105

tuhina2020:
@minrk @yuvipanda 
#2105


jupyterhub/utils.py
```diff
@@ -247,6 +247,14 @@ def admin_only(self):
     if user is None or not user.admin:
         raise web.HTTPError(403)
 
+@auth_decorator
+def metrics_authentication(self):
+    """Decorator for restricting access to metrics"""
+    user = self.get_current_user()
+    authenticate_prometheus = self.authenticate_prometheus()
```
minrk: This should be `self.authenticate_prometheus`, not `self.authenticate_prometheus()` (it's a property, not a method).

jupyterhub/utils.py
```diff
@@ -247,6 +247,14 @@ def admin_only(self):
     if user is None or not user.admin:
         raise web.HTTPError(403)
 
+@auth_decorator
+def metrics_authentication(self):
+    """Decorator for restricting access to metrics"""
+    user = self.get_current_user()
```
minrk: This should be `user = self.current_user`, not `.get_current_user()`

tuhina2020:
@minrk added tests for metrics API

jupyterhub/utils.py
```diff
@@ -247,6 +247,14 @@ def admin_only(self):
     if user is None or not user.admin:
         raise web.HTTPError(403)
 
+@auth_decorator
+def metrics_authentication(self):
+    """Decorator for restricting access to metrics"""
+    user = self.current_user
+    authenticate_prometheus = self.authenticate_prometheus
+    if user is None and not user.authenticate_prometheus:
```
minrk: This check should be:

```python
if user is None and self.authenticate_prometheus:
```

jupyterhub/utils.py
```diff
@@ -247,6 +247,13 @@ def admin_only(self):
     if user is None or not user.admin:
         raise web.HTTPError(403)
 
+@auth_decorator
+def metrics_authentication(self):
+    """Decorator for restricting access to metrics"""
+    user = self.current_user
+    if user is None and not self.authenticate_prometheus:
```
minrk: This should be `and self.authenticate_prometheus`. The idea here is that `if self.authenticate_prometheus`, we should only allow authenticated users.

jupyterhub/tests/test_pages.py
```diff
@@ -667,3 +667,20 @@ def test_server_not_running_api_request(app):
     assert r.status_code == 404
     assert r.headers["content-type"] == "application/json"
     assert r.json() == {"message": "bees is not running"}
+
+@pytest.mark.gen_test
+def test_metrics_no_auth(app):
+    r = yield get_page("metrics", app)
+    r.raise_for_status()
```
minrk: Almost there! Instead of calling `raise_for_status()`, we should *expect* a 403 error here:
```suggestion
    assert r.status_code == 403
```

And then remove the url checks below, since it won't send folks to the login page.

consideRatio:
Thanks for your work on thus @tuhina2020! :heart: :tada:!

PR #2236, opened 2018-10-12
Fix missing spaces in if statement in bootstrap-script and README 

### LINKED ISSUE #2202: Example #3 - Run a shell script  has a bug
In the script in Example #3:

> USER=$1
> if ["$USER" == ""]; then
>     exit 1
> fi

There should be space inside the if condition. A few lines later another if condition is correct:

> if [ -d "$USER_DIRECTORY" ]; then

mpolidori:
Fixes #2202


minrk:
Thanks!

PR #2240, opened 2018-10-16
Small improvement of test coverage

GeorgianaElena:
Hi @minrk ! I've added a few checks regarding https://github.com/jupyterhub/outreachy/issues/8 hoping I will get a little more familiar with the code base.


minrk:
A+, thanks!

PR #2244, opened 2018-10-16
admin_groups for PAMAuthenticator

vilhelmen:
#888 via #1564

Introduces `admin_groups` to PAMAuthenticator, a set of system groups that grant admin status to users. `admin_groups` and `admin_users` are effectively an or between the two with room for additional changes in the future.

Created mock objects for struct_group and struct_passwd, migrated existing mock group object to it.

As discussed earlier in #1564, when an error in group resolution occurs, authentication fails instead of soft-failing and not returning an admin state. If this does fail, it will look like an ordinary username/password mismatch to the user instead of a backend failure.

Is there a way to report a user-facing error message at login failure? Raising an exception just reports a 500 error. Push comes to shove, I suppose a 500 will work. We don't want to suggest it's the user that's wrong when the backend is failing.


jupyterhub/auth.py
```diff
@@ -643,25 +666,50 @@ def authenticate(self, handler, data):
...
                     self.log.warning("PAM Account Check failed: %s", e)
-            else:
-                return username
+                return None
+
+        if username in self.admin_users:
+            admin = True
+        elif self.admin_groups:
+            try:
+                # Most likely source of error here is a group name <-> gid mapping failure
```
minrk: I'd rather fail on errors here.

In fact, would you like to refactor this step into a `def is_admin(username): bool|None` method? That would make it easier to override just this logic in subclasses for those who want to have different behavior on errors.

vilhelmen: That sounds wisest.

I suppose `is_admin` should go in `Authenticator`, do you agree? I'll start moving things around!

vilhelmen:
All done. I think moving admin checks out of authentication was a great idea!

Now with `authenticate` back to doing just that, and `get_authenticated_user` changes the default admin flag from None to a call to `is_admin`, which defaults to `True` if the user is in the `admin_users` list, or `None`, to prevent stripping admin from users granted admin status through the web interface, and to prevent from interfering with any additional checks by other authenticators, since `admin_users` is part of the base authenticator.

jupyterhub/auth.py
```diff
@@ -631,11 +666,51 @@ def _default_executor(self):
...
+        """
+    ).tag(config=True)
+
     def __init__(self, **kwargs):
         if pamela is None:
             raise _pamela_error from None
         super().__init__(**kwargs)
 
+    @run_on_executor
+    async def is_admin(self, handler, data):
```
minrk: `@run_on_executor` wraps a blocking method (no `async` here).

vilhelmen: Thank you. I've fixed that, but now it's a syntax error on the `admin_status = await maybe_future(super().is_admin(handler, data))` in `PAMAuthenticator.is_admin`.

I've tried every combination of `await`, `yield`, and `maybe_future` I can think of and it either is a syntax error, or it's returning a generator instead of the actual result.

jupyterhub/auth.py
```diff
@@ -631,11 +666,51 @@ def _default_executor(self):
...
+        """
+    ).tag(config=True)
+
     def __init__(self, **kwargs):
         if pamela is None:
             raise _pamela_error from None
         super().__init__(**kwargs)
 
+    @run_on_executor
+    def is_admin(self, handler, data):
```
minrk: I see the problem now.

- `@run_on_executor` wraps a *blocking* method and makes it return a Future
- calling `super().is_admin()` can only be called from a true coroutine

So the options are:

1. make the base `.is_admin()` method blocking so that it can be called here without `await`
2. split this method so that `.is_admin` is a true coroutine (`async def is_admin`), and it `awaits` super, then calls `._check_groups`, a private `@run_on_executor` method.

Either one's fine with me.

vilhelmen: Thank you so much, I would have never figured that out on my own. I opted for the former. The basic is_admin in Authenticator should be trivial enough that it doesn't need to be async'd.

vilhelmen:
Gentle bump since the CI failure looks like the misconfigured timeout like #2243 had. Everything is passing for me in a VM with a timeout of 25.

vilhelmen:
On second thought, would it be beneficial to pass `auth_state` along to `is_admin`? It could allow some coordination between `authenticate` and `is_admin`.

Is there one Authenticator for the entire hub, or one for each user session? If it's one per user, it could also be done with variables inside the authenticator, but if not, shared data between the two could be useful. For example, the LDAP authenticator needs to determine a user's DN for authentication. `is_admin` may need the DN again to check admin group memberships, requiring more round trips for DN lookup. Sharing auth data between phases would allow for quicker overall authentication.

May or may not be related to something I'm doing for #2261.

minrk:
> would it be beneficial to pass auth_state along to is_admin

That's a good idea. Instead of calling it with the input, as done here, call it with the `authenticated` dict after the other processing (i.e. at the very end of the method)

vilhelmen:
I'm glad you agree, I'll move over my commit for this later today.

I'm also starting to think it would be beneficial to provide the `authenticated` dict to the other functions in the `get_authenticated_user` flow, like `check_whitelist`. I'm going to investigate if that can be done without completely breaking existing authenticators and submit my findings later.

jupyterhub/auth.py
```diff
@@ -281,8 +281,8 @@ def check_blacklist(self, username):
                 'name': authenticated,
             }
         authenticated.setdefault('auth_state', None)
-        admin_status = await maybe_future(self.is_admin(handler, data))
-        authenticated.setdefault('admin', admin_status)
+        # Leave the default as None, but reevaluate later post-whitelist
+        authenticated.setdefault('admin', None)
```
vilhelmen: I figured leaving a default of None at the beginning was safest, instead of leaving it undefined until the end.

If other `Authenticator` functions start to use the `authenticated` dict, having a default set as soon as possible seems best.

jupyterhub/auth.py
```diff
@@ -282,6 +299,14 @@ def check_blacklist(self, username):
             return
 
         if whitelist_pass:
+            # Have to merge any existing admin status returned by authenticate with is_admin
+            current_admin_status = authenticated['admin']
+            is_admin_status = await maybe_future(self.is_admin(handler, data, authenticated))
+            if current_admin_status is None:  # use whatever is_admin gave
+                authenticated['admin'] = is_admin_status
+            elif is_admin_status is not None:  # Merge Bool results from both
+                authenticated['admin'] = current_admin_status or is_admin_status
```
vilhelmen: Admin status configuration has gotten pretty ugly, but I believe this is the correct logic for merging anything that comes out of `authenticate`.

minrk: Let's simplify:

```python
if authenticated['admin'] is None:
    authenticated['admin']  = await self.is_admin(handler, authenticated)
```

Basically, there's no reason to call or implement `is_admin` if authenticate is returning the admin status. Providing admin status from two sources should be considered mutually exclusive.

jupyterhub/auth.py
```diff
@@ -316,6 +341,23 @@ def check_blacklist(self, username):
         """
         return True
 
+    def is_admin(self, handler, data, authentication, *args, **kwargs):
+        """Authentication helper to determine a user's admin status.
+
+        .. versionadded: 0.9.5
```
minrk: Next release is 1.0

jupyterhub/auth.py
```diff
@@ -316,6 +341,23 @@ def check_blacklist(self, username):
         """
         return True
 
+    def is_admin(self, handler, data, authentication, *args, **kwargs):
```
minrk: Let's not add `*args, **kwargs` here. I don't think that's a good plan for methods that will be overridden by subclasses, since there's no indicator whether this has been implemented by sublasses until/unless args are added, which is effectively the same as if they weren't here at all on the base class.

Instead, let's take two args:

```suggestion
    def is_admin(self, handler, authenticated):
```

where the second arg is the authenticated user dict.

jupyterhub/auth.py
```diff
@@ -631,11 +673,53 @@ def _default_executor(self):
...
+                # fail to authenticate and raise instead of soft-failing and not changing admin status
+                # (returning None instead of just the username) as this indicates some sort of system failure
+
+                admin_group_gids = {getgrnam(x).gr_gid for x in self.admin_groups}
+                user_group_gids  = {x.gr_gid for x in getgrouplist(username, getpwnam(username).pw_gid)}
+                admin_status = len(admin_group_gids & user_group_gids) != 0
+
+            except Exception as e:
+                if handler is not None:
+                    self.log.critical("PAM Admin Group Check failed (%s@%s): %s", username, handler.request.remote_ip, e)
```
minrk: `log.error` for errors

jupyterhub/auth.py
vilhelmen: That's fair. I was actually just thinking of that on my drive in today. I'll get #2268 refactored as well.

minrk:
Excellent, thank you!

vijaysinghrajputv6-collab (commented):
Hello bhai aayat ka password bata do please

vijaysinghrajputv6-collab:
Bhalobasa ka password bata do please

PR #2251, opened 2018-10-18
[wip] Idle culling implementation using notebook activity data

angelikamukhina:
https://github.com/jupyterhub/outreachy/issues/25

### DIFF
--- a/examples/cull-idle/cull_idle_servers.py
+++ b/examples/cull-idle/cull_idle_servers.py
@@ -93,15 +93,22 @@ def cull_idle(url, api_token, inactive_limit, cull_users=False, max_age=0, concu
     auth_header = {
         'Authorization': 'token %s' % api_token,
     }
-    req = HTTPRequest(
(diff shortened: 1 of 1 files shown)


minrk:
@gweis both true! I'm hoping to fix the first with oauth scopes so that a token for idle culling could have access to read activity without write access to the notebook api. The second I've been meaning to get to for some time, and should be improved by https://github.com/jupyter/notebook/pull/4235

gweis:
@minrk Awesome,... looking forward for these changes to be released ;). These would really help to track activity in notebooks (not just user activity, but also active / running cells / kernels etc...)

consideRatio:
Stil PR has been stale for a long time it seems, and I know that there is now a dedicated package that does culling. @angelikamukhina and @minrk, any suggestions on what to do at this point with this PR?

yuvipanda:
I think we can close this now. Thanks for getting this going, @angelikamukhina.

PR #2255, opened 2018-10-19
unpin coverage

minrk:
need run.parallel = True in coveragerc to fix database errors with coverage 5

exclude the resulting .coverage.host.pid files in gitignore


betatim:
LGTM

Should we also do this in repo2docker?

PR #2257, opened 2018-10-20
Added health check endpoint

tmshn:
Health check is very fundamental to make a service to be robust and scalable.

I added health check endpoint to simply return 200 OK to requests.

Note: I'm using zero-to-jupyterhub-k8s so going to fix helm template to add health check.


minrk:
Thanks!

PR #2268, opened 2018-10-26
Share authenticated dict with auth functions

vilhelmen:
In the same vein as #2244 and sharing the `authenticated` dict with the new `is_admin` function, this brings sharing the `authenticated` dict with the `check_whitelist` and `check_blacklist` functions and also updates related function definitions to allow for future changes with fewer problems. A compatibility check is in place until users have had enough time to migrate.

For example, if an LDAP authenticator were to cache the user's DN during `authenticate` and want to do LDAP group whitelisting in `check_whitelist`, it would currently have to do an additional DN resolution since it doesn't have access to the `authenticated` data. Not only do unnecessary lookups have to be done, but it could cause problems as the username passed to the list functions have been modified by normalization.

The internals of `Authenticator` are rather brittle since they lack `*args, **kwargs` in their signatures, so this would be a breaking change for any custom authenticators that override `check_blacklist` or `check_whitelist`, as they get the incorrect number of arguments. I've checked the authenticators under the jupyterhub account, and only LocalAuthenticator overrode a check function, it seems.

To prevent breaking other authenticators, `get_authenticated_user` first tries the new signature and then falls back to the old one and prints deprecation warnings when a TypeError is raised.

This also adds `*args, **kwargs` to `authenticate` so that it can also be changed with less double in the future. If/when the arguments change, we can use the same deprecation check flow that the check functions use.

This also adds a test for the current signatures to make sure they still work until it's decided that users have had enough time to migrate. The migration for end users should really be as simple as adding `*args, **kwargs` to their signature and passing them along to any `super()` function they call.

I made some assumptions about version numbers in the version change docstrings, feel free to modify as necessary. I didn't update the signature for username normalization, but maybe that should update as well before it comes back to bite us?


vilhelmen:
`*args, **kwargs` have been removed since changes to signatures are significant and this would hide whether or not an authenticator supports certain features. Updated version assumptions to 1.0.

minrk:
Sorry for taking so long on this one! I moved where the deprecation/validation occurs so it would fix `.check_whitelist` itself rather than have to adapt each time it's called (which meant the calls in `app.py` wouldn't get updated).

vilhelmen:
Better late than never. I like your fix, it's much better!

PR #2290, opened 2018-11-05
add configuration for shutting down all user spawners on logout

katsar0v:
Added `shutdown_on_logout` new setting. JupyterHub is now calling `spawner.stop()` method of the spawner when a user logs out.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1020,6 +1020,11 @@ class JupyterHub(Application):
 
     statsd = Any(allow_none=False, help="The statsd client, if any. A mock will be used if we aren't using statsd")
 
+    shutdown_on_logout = Bool(
+        False,
+        help="""Shuts down all user servers on logout"""
+    ).tag(config=True)
+
     @default('statsd')
     def _statsd(self):
         if self.statsd_host:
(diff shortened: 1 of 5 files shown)


minrk:
I suspect that'll do it. Would you like to add a test in `juptyerhub/tests/test_pages.py`?

katsar0v:
> I suspect that'll do it. Would you like to add a test in `jupyterhub/tests/test_pages.py`?

Am not sure how to do this honestly. I tested locally with the config changed to dockerspawner and installing dockerspawner from pip.

minrk:
No worries! I'll look into the test.

yuvipanda:
This is currently failing with:

> ERROR: /home/travis/build/jupyterhub/jupyterhub/pytest.ini:2: requires pytest-3.3, actual pytest-3.10.0'

Think we might have to pin this?

minrk:
pytest is [doing string comparison](https://github.com/pytest-dev/pytest/pull/4308) of version numbers, so `'3.10' < '3.3'`. We just have to disable the check while pytest is at 3.10

minrk:
Thanks @katsar0v! Test added and passing.

Issue #2295, opened 2018-11-08
Exception raised during api check


ryanlovett:
We're seeing exceptions of the form:
```
JupyterHub web:1670] Uncaught exception GET /hub/user/{username}/api/contents/foo.ipynb?content=0&_=1541561546585 (10.240.1.119)
    HTTPServerRequest(protocol='https', host='{hostname}', method='GET', uri='/hub/user/{username}/api/contents/foo.ipynb?content=0&_=1541561546585', version='HTTP/1.1', remote_ip='10.240.1.119')
    Traceback (most recent call last):
      File "/usr/local/lib/python3.6/dist-packages/tornado/web.py", line 1592, in _execute
        result = yield result
      File "/usr/local/lib/python3.6/dist-packages/jupyterhub/handlers/base.py", line 961, in get
        if "api" in user_path.split("/") and not user.active:
    AttributeError: 'NoneType' object has no attribute 'active'
```

This also occurs for users we delete via /hub/admin. Seems related to #2294.

**Additional context**
z2jh master

katsar0v:
I think correct line would be `if "api" in user_path.split("/") and user and not user.active:`
Can you reproduce this issue if you replace "/usr/local/lib/python3.6/dist-packages/jupyterhub/handlers/base.py", line 961 ?

minrk:
I think that's exactly right. Would you like to make a PR?

ryanlovett:
> Can you reproduce this issue if you replace "/usr/local/lib/python3.6/dist-packages/jupyterhub/handlers/base.py", line 961 ?

Not easily on the hub we're running. However we're on master z2jh so once it is upstream we can bump the chart version.

katsar0v:
Okay, made a pull request which also passes the builds, but still think that we have to make sure this fixes the issue

ryanlovett:
Thanks! I'll let you know once I can test.

minrk:
Merged the patch. @ryanlovett feel free to close when you confirm the fix.

PR #2298, opened 2018-11-09
improve condition when failing api request, check if user exists, aft…

katsar0v:
Fixes issue #2295 

Improves if logic, checks if user exists and is not NoneType, afterwards check if the user is not active.


ryanlovett:
Thanks for the PR! This will pass None into _fail_api_request which will then generate an exception when attempting to access user.name.

katsar0v:
I think the condition checks if user is not None.

minrk:
Exactly right, `_fail_api_request` will not be called if user is None.

Thanks @katsar0v!

Issue #2299, opened 2018-11-09
"No Hub user identified for request" Errors


clkao:
Similar to #1618, we are seeing flood of this error message from the single user pod:
`No Hub user identified for request" Errors` after `Allowing whitelisted Hub user XXX`

but this doesn't seem to be a stalled browser session, but a freshly spawned pod by user. The error causes lots of 404 /api/authorization/token requests to the hub pod.

from the user pod log: jupyterhub and jupyterhub-singleuser both on version 0.9.2

Meanwhile I noticed that there is an action item in the incident report: 

| Add “No Hub user identified for request” error to Troubleshooting doc.

but I don't see the relevant section in https://github.com/jupyterhub/jupyterhub/blob/master/docs/source/troubleshooting.md

minrk:
If the Hub is getting 404s on `/api/authorizations/token`, then I'm guessing what's happening is some frontend is making requests with an old token or cookie (e.g. one issued to a prior instance of the server, but the server got cleaned up and then restarted, which would invalidate oauth tokens issued to the previous instance). The failing requests should prompt a fresh login, but if these are *API* requests, then they ought to just fail with 403 until the user refreshes their credentials.

#2303 fixes an issue where the auth results of an invalid token would be cached, but the cached result would never be used. The results for the user will be the same, but a bunch of stale-token requests will no longer flood the Hub with checks to see if the same token is valid over and over again.

PR #2303, opened 2018-11-12
HubAuth: allow caching None responses

minrk:
when a token doesn't identify a user, the response is None. These results are cached, but the cache checked for `is None`, causing failed-auth responses to effectively not be cached.

This fixes the flood of Hub requests caused by repeated stale-auth requests in #2299


clkao:
@minrk woot. is the 3.5 ci failure a flaky test? it's getting 202 instead of 500

PR #2307, opened 2018-11-13
getgrnam/getpwnam/getgrouplist refactor

vilhelmen:
I noticed a function in auth was using its own getpwnam without a wrapper when everything else was, I switched it over. I also moved the getgrnam/getpwnam/getgrouplist wrappers under LocalAuthenticator since they're only used by LocalAuthenticator and PAMAuthenticator.

I've noticed that the spawner code doesn't use a wrapper for pwd or os functions that the python docs say are strictly available on unix platforms. Can we just get rid of the wrapper functions in auth entirely?

### DIFF
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -27,30 +27,6 @@ from .utils import maybe_future, url_path_join
 from .traitlets import Command
 
 
-def getgrnam(name):
-    """Wrapper function to protect against `grp` not being available
-    on Windows
-    """
-    import grp
-    return grp.getgrnam(name)
-
-
-def getpwnam(name):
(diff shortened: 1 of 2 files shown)


minrk:
The only goal of these wrappers is to avoid the imports that would fail on Windows when the classes that use them are not in use (PAMAuthenticator and LocalProcessSpawner cannot be used on Windows). Local imports in the methods where they are used is a fine way to accomplish that as well, or local private methods (ultimately, it doesn't matter). The one thing I wouldn't do, though, is make them public methods, which implies a public API on the classes, which I would like to avoid.

vilhelmen:
Makes sense, I'll move them over to private functions!

minrk:
Perfect, thanks!

Issue #2288, opened 2018-11-05
More hooks


minrk:
We currently have a mixture of overrideable methods (`Authenticator.pre_spawn_start`) and hooks which can be set via config (`c.Spawner.pre_spawn_hook`). Subclassing in config is intimidating for lots of folks, so I think we should try to take a systematic approach to adding hooks in at least the common places for deployment customization.

Current potential for hooks:

- normalize_username (currently a method)
- validate_username (currently a method)
- pre_spawn (have this)
- post_stop
- post_auth

others? pre/post counterparts like `pre_stop`, `post_spawn`? I can't necessarily think of a use for them, but maybe they should be defined for completeness?

Question about hooks:

1. should they be transforms (i.e. post_spawn takes the return of spawn and can transform it), or pure event listeners (cannot modify return value)? I can see the utility in transforms, but I can also see the potential complexity of allowing hooks to modify things like that
2. I had another one, but I forgot.

katsar0v:
My ideas for hooks:
- pre / post login
- pre / post logout
- different hooks in services, e. g. service added / runned. Would help with implentation, just need one hook for example and code styling

vpavlin:
@minrk I have a use case for `post_spawn` hook:) We run JupyterHub on OpenShift/Kubernetes and we run nbviewer as a second container in the singleuser pod to provide notebook sharing feature.

The problem is that for that we need a new (unauthenticated) route, so current solution is to poll proxy for routes and for each user create a second route. I would much prefer to drop polling and the `add_route` on `post_spawn` - as that is when I can get an IP of the pod to construct the route target.

Same for `post_stop` - removing the newly created route.

manics:
We have `pre_spawn_hook` `auth_state_hook` `post_stop_hook`.
If there's a need for a new hook I think it makes sense to describe the requirements in a new dedicated issue.

Issue #2261, opened 2018-10-25
Authenticator/Spawner configuration cooperation


vilhelmen:
I'm looking at developing some tighter integration between the authenticator and the spawner and wondering if it fits within the scope of the main project contribution-wise. I'd hate to hide useful features away in custom forks.

What I'm proposing is a designated section of `auth_state` for storing user profile data relevant to the more complex spawners that aids in user configuration for things like mounted homedir disk permissions.

In particular, I'm looking at the interaction between PAMAuthenticator/LDAPAuthenticator, and the Kubernetes (or Docker) spawner. At the moment, PAM/LDAP authenticator has easy access to information such as UID, primary GID, and secondary groups that at the moment has to be looked up all over again even though the authenticator most likely already did and discarded the results.

The current Docker/Kubernetes documentation seems to suggest that all files for all users be set to 1000:1000 and rely on container mount isolation, but this isn't feasible when users need to work on a shared machine with mounted homedirs.

This could be achieved by, for example, `Authenticator.build_profile` (`build_spawn_profile`?) which, like `is_admin` #2244, is called during `get_authenticated_user` and returns this designated `auth_state` subdict and merges it with any `auth_state` generated by `authenticate`. From there, any spawner updated to be made aware of this data can use it for spawning. This would cost likely need to be gated by config options like `c.Authenticator.generate_spawn_profile` and `c.Spawner.use_spawn_profile` or something.

I was thinking something along the lines of UID, primary GID, and a GID map laid out like so:
```
auth_state = {
  'spawn_profile': {
    'uid': 1001,
    'primary_gid': [2000, 'jh_user']
    'groups': {
      2000: 'jh_user',
      3000: 'jh_admin',
      4000: 'postgres_user',
      5000: 'stats_3050'
    },
}
```

Then Docker/kubernetes spawners and containers can use this information for things like prepopulating the container with the user's groups and things like `c.KubeSpawner.uid` or `c.KubeSpawner.fs_gid`

minrk:
[...]

PR #2308, opened 2018-11-13
post_auth_hook processor

vilhelmen:
Does what it sounds like, allows an arbitrary function to be called right before the return in `get_authenticated_user`.

(and an accidental reformat of the include section)

#2261 / #2288 (which I just noticed, I should look at issues more)

I'm not sure if the full authentication dictionary should be passed in, or if the signature should be changed to something like `post_auth_hook(authenticator, handler, name, admin, auth_state)` to prevent changes to the name and admin status.

I feel like if I was the end user, I would be annoyed by the restriction, though we have proper places for overriding and setting the name and admin state.

### DIFF
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -223,7 +223,8 @@ class Authenticator(LoggingConfigurable):
         details from an external system.
 
         This function is called after the user has passed all authentication checks
-        and is ready to successfully authenticate.
+        and is ready to successfully authenticate. This function must return the
+        authentication dict reguardless of changes to it.
 
         This maybe a coroutine.
 
@@ -236,12 +237,12 @@ class Authenticator(LoggingConfigurable):
                     'pw_data': user_data
                     'gid_list': os.getgrouplist(authentication['name'], user_data.pw_gid)
                 }
-
-                if authentication['auth_data'] is None:
-                    authentication['auth_data'] = {'spawn_data': spawn_data}
+                if authentication['auth_state'] is None:
+                    authentication['auth_state'] = {'spawn_data': spawn_data}
                 else:
-                    authentication['auth_data']['spawn_data'] = spawn_data
+                    authentication['auth_state']['spawn_data'] = spawn_data
 
+                return authentication
             c.Authenticator.post_auth_hook = my_hook
 
         """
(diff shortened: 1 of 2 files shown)


jupyterhub/auth.py
```diff
@@ -210,6 +209,53 @@ def validate_username(self, username):
...
+
+        Args:
+            handler (tornado.web.RequestHandler): the current request handler
+            authentication (dict): User authentication data dictionary. Contains the
+                username ('name'), admin status ('admin'), and auth state dictionary ('auth_state').
+        Returns:
+            None
+        """
+        if self.post_auth_hook is not None:
+            return self.post_auth_hook(self, handler, authentication)
```
minrk: The docstring (rightly so) allows this to be a coroutine, but the code doesn't. It needs `await maybe_future(self.post...`).

I think it's probably appropriate here to have the hook return the new `authenticated` structure, to make it clear that mutating is appropriate.

vilhelmen: Hah, you know I'm not good with those! I just copied it from the spawner and renamed things. [Maybe it needs to be fixed there as well?](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/spawner.py#L840)

Are you thinking we pass in the authentication dict and always replace it with the results to force the user to return it? Or maybe do nothing if it returns None?

Should we pass in a deepcopy'd version? Or should we just send it as it is because it's being returned to make the user pay attention to it?

I'll put together a forced return with the same dict for now.

jupyterhub/auth.py
```diff
@@ -210,6 +209,57 @@ def validate_username(self, username):
...
+
+            import os, pwd
+            def my_hook(authenticator, handler, authentication):
+                user_data = pwd.getpwnam(authentication['name'])
+                spawn_data = {
+                    'pw_data': user_data
+                    'gid_list': os.getgrouplist(authentication['name'], user_data.pw_gid)
+                }
+                
+                if authentication['auth_state'] is None:
```
vilhelmen: Do you think we can change the default `auth_state` from None to an empty dict without too many headaches Fumbling with None makes the example more complex.

minrk:
Yeah, there are a couple of async/race leaks that have gotten into the tests recently. Need to chase them down.

vilhelmen:
So long as it's failing by accident and not passing by accident!

PR #2309, opened 2018-11-13
refresh_user: add handler parameter

adelcast:
The current request handler might be needed to determine if the auth
data needs to be refreshed.


minrk:
Thanks!

PR #2313, opened 2018-11-14
setuptools entrypoints instead of scripts

adelcast:
Use setuptools console_scripts functionality to create top level jupyter
& jupyterhub-single user entry point scripts on *nix, and executables on
Windows.


minrk:
Thanks! The `scripts` dir should be removed if we are switching to entrypoints.

adelcast:
oops, removed files

Issue #2120, opened 2018-08-30
Server process being moved to first user's systemd cgroup


cmaves:
**Description**
When running Jupyterhub as a systemd service as described by [Run jupyterhub as a system service](https://github.com/jupyterhub/jupyterhub/wiki/Run-jupyterhub-as-a-system-service), the `/opt/anaconda/bin/jupyterhub` process is started under the services's cgroup (`named=systemd:/system.slice/jupyterhub.service`)

However when the first user logs in, the original `jupyterhub` process is moved to that user's cgroup (i.e. `named=systemd/user.slice/user-1000.slice`). All single-user instances are then spawned under this user's cgroup. 

This bug occurs in the latest version of Jupyterhub (0.9.2 on Anaconda 5.2) and on version 0.7.2 on Anaconda 4.4, on multiple Linux distros.

**To Reproduce**
1. Install Jupyterhub using `conda install -c conda-forge jupyterhub` per the installation guide.
2. Generate the config: `jupyterhub --generate-config -f /etc/jupyterhub/jupyterhub_config.py`
3. Install the service file provided [here](https://github.com/jupyterhub/jupyterhub/wiki/Run-jupyterhub-as-a-system-service#ubuntudebian-anaconda3-with-systemd)
4. Reload systemd service files: `systemctl daemon-reload`
5. Start `jupyterhub.service`
6. Login as a user

**Expected behavior**
The server process (`/opt/anaconda/bin/jupyterhub`) should always remain in the `named=systemd:/system.slice/jupyterhub.service` cgroup. The single-user instance would preferably get added to their respective user's cgroup, but remaining in the service cgroup would be acceptable.

**Desktop (not really relevant as this is a server issue):**
 - OS: Mac OS X, Arch Linux
 - Safari, Chrome, Firefox

**Additional context**
This bug is an issue because many systems use systemd's cgroups as a resource limiting mechanism. This bug loads every single-user jupyterhub instance under one user, preventing effective resource management, and allowing other users to consume resources of another user.

cmaves:
After looking into, the issue is the `PAMAuthenticator.pre_spawn_start` is responsible for starting the pam:session using `pamela.open_session(username)`. This function is run from the main process. Instead `pamela.open_session(username)` needs to be run in the preexec_fn when spawning the single user process. `pamela.close_session(username)` needs to by the singleuser process upon exit.

jhermann:
[...]

PR #2321, opened 2018-11-20
[WIP] Moved PAM session creations to a preexec_fn

cmaves:
PAM sessions are intended to be created within the spawned child
process, not in the parent process. This can create issues with
certain PAM session modules (i.e. pam-systemd.so)

### DIFF
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -356,6 +356,9 @@ class Authenticator(LoggingConfigurable):
                 persisted; and `admin`, the admin setting value for the user.
         """
 
+    def pre_exec_fn(self, user, spawner):
+        """Hook called before exec is called in the spawned child process """
+
     def pre_spawn_start(self, user, spawner):
         """Hook called before spawning a user's server
 
@@ -746,8 +749,7 @@ class PAMAuthenticator(LocalAuthenticator):
         return username
 
 
-    @run_on_executor
-    def pre_spawn_start(self, user, spawner):
+    def pre_exec_fn(self, user, spawner):
         """Open PAM session for user if so configured"""
         if not self.open_sessions:
             return
@@ -755,8 +757,9 @@ class PAMAuthenticator(LocalAuthenticator):
             pamela.open_session(user.name, service=self.service, encoding=self.encoding)
         except pamela.PAMError as e:
             self.log.warning("Failed to open PAM session for %s: %s", user.name, e)
-            self.log.warning("Disabling PAM sessions from now on.")
-            self.open_sessions = False
+            # this will be ineffective in the child process
+            # self.log.warning("Disabling PAM sessions from now on.")
+            # self.open_sessions = False
 
     @run_on_executor
     def post_spawn_stop(self, user, spawner):
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -357,7 +357,10 @@ class Authenticator(LoggingConfigurable):
         """
 
     def pre_exec_fn(self, user, spawner):
-        """Hook called before exec is called in the spawned child process """
+        """Hook called before Popen is called by LocalProcessSpawner """
+
+    def post_exec_fn(self, user, spawner):
+        """Hook called after child process closes by LocalProcessSpawner """
 
     def pre_spawn_start(self, user, spawner):
         """Hook called before spawning a user's server
@@ -757,12 +760,8 @@ class PAMAuthenticator(LocalAuthenticator):
             pamela.open_session(user.name, service=self.service, encoding=self.encoding)
         except pamela.PAMError as e:
             self.log.warning("Failed to open PAM session for %s: %s", user.name, e)
-            # this will be ineffective in the child process
-            # self.log.warning("Disabling PAM sessions from now on.")
-            # self.open_sessions = False
 
-    #@run_on_executor
-    def post_spawn_stop(self, user, spawner):
+    def post_exec_fn(self, user, spawner):
         """Close PAM session for user if we were configured to opened one"""
         if not self.open_sessions:
             return
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -761,7 +761,7 @@ class PAMAuthenticator(LocalAuthenticator):
             # self.log.warning("Disabling PAM sessions from now on.")
             # self.open_sessions = False
 
-    @run_on_executor
+    #@run_on_executor
     def post_spawn_stop(self, user, spawner):
         """Close PAM session for user if we were configured to opened one"""
         if not self.open_sessions:
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -1167,7 +1167,8 @@ class LocalProcessSpawner(Spawner):
         """
         fn = set_user_setuid(name)
         def preexec_fn():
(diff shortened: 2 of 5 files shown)


jupyterhub/auth.py
```diff
@@ -746,19 +749,19 @@ def authenticate(self, handler, data):
...
-    @run_on_executor
-    def pre_spawn_start(self, user, spawner):
+    def pre_exec_fn(self, user, spawner):
         """Open PAM session for user if so configured"""
         if not self.open_sessions:
             return
         try:
             pamela.open_session(user.name, service=self.service, encoding=self.encoding)
         except pamela.PAMError as e:
             self.log.warning("Failed to open PAM session for %s: %s", user.name, e)
```
minrk: Since this is now called in the subprocess, I'm not sure `self.log.warning` is going to work. Probably switch this to a lower-level print to stderr.

jupyterhub/auth.py
```diff
@@ -356,6 +356,9 @@ def is_admin(self, handler, authentication):
                 persisted; and `admin`, the admin setting value for the user.
         """
 
+    def pre_exec_fn(self, user, spawner):
+        """Hook called before exec is called in the spawned child process """
```
minrk: ...when using a LocalProcessSpawner and its derivatives.

btw, this should probably return a pure callable rather than being an instance method calling the method. I wouldn't trust e.g. accessing the database-connection-having user object in the pre_exec_fn.

Instead, let's have something like:

```python
@staticmethod
def _preexec_fn(name, service, encoding):
    ...

def make_preexec_fn(self, user, spawner):
    return partial(self._preexec_fn, user.name, self.service, self.encoding)
```

jupyterhub/auth.py
```diff
@@ -746,19 +749,19 @@ def authenticate(self, handler, data):
...
         except pamela.PAMError as e:
             self.log.warning("Failed to open PAM session for %s: %s", user.name, e)
-            self.log.warning("Disabling PAM sessions from now on.")
-            self.open_sessions = False
+            # this will be ineffective in the child process
+            # self.log.warning("Disabling PAM sessions from now on.")
+            # self.open_sessions = False
 
-    @run_on_executor
+    #@run_on_executor
```
minrk: This needs to come back, btw. It's important that all calls to pam happen off the main thread.

jupyterhub/auth.py
```diff
@@ -746,19 +749,19 @@ def authenticate(self, handler, data):
...
     def post_spawn_stop(self, user, spawner):
```
minrk: As I understand it, to fix the issue, the close session also needs to happen in the subprocess as well, not just the open. How do we accomplish that? There isn't an equivalent to preexec_fn for the end of a process.

Do we instead need to create a two-layer process, where the parent does:

```python
pamela.open_session()
p = start_real_subprocess(args, env)
wait(p)
pamela.close_session()
```

jupyterhub/auth.py
cmaves: The reason I removed it was because it caused this check failed with: `jupyterhub/tests/test_internal_ssl_api.py::test_bad_spawn`.

jupyterhub/auth.py
cmaves: This is believe this is how `sshd` accomplishes it. It has process that watches the shell process.

I think creating a `PAMAuthenticator.post_exec_stop()` function and moving the PAM session logic into it would be best and making the two-layer process look like this:
```
if self.auth:
   self.auth.pre_exec_fn(self.user, self)
Popen.run(args, env) # will wait for process to end on its own
if self.auth:
   self.auth.post_exec_stop(self.user, self)
```

This also has the advantage of allowing me to avoid the `@run_on_executor` statement that has caused issue with one check, because the PAM session will already be occurring off of the main thread.

cmaves: It also avoid doing a lot with `pre_execfn` kwarg which the Python docs say is bad.

jupyterhub/auth.py
minrk: Then that would be an issue that needs to be fixed in the test. This method must be allowed to be async. Making it a blocking method isn't an okay fix for the problem, it just ensures the async behavior isn't tested.

But this is hugely useful info! I've been wondering what the source of these errors has been, and I can look into a fix that doesn't assume an async method is synchronous, if you like.

cmaves: The most recent commit restructures this into a function that runs on a intermediate process. This make the run_on_executor statement unnecessary because it is already running of the main thread.

jupyterhub/auth.py
cmaves: The instance method no longer calls this method, so this should be resolved.

jupyterhub/auth.py
cmaves: I've confirmed it does work in a subprocess.

jupyterhub/spawner.py
```diff
@@ -1308,12 +1358,17 @@ def get_env(self):
         we return the exit code of the process if we have access to it, or 0 otherwise.
         """
         # if we started the process, poll with Popen
-        if self.proc is not None:
-            status = self.proc.poll()
-            if status is not None:
-                # clear state if the process is done
+        if isinstance(self.proc_state, Queue_class):
```
minrk: This isinstance check appears irrelevant, since proc_state is always either None or this Queue, right?

jupyterhub/spawner.py
```diff
@@ -1308,12 +1358,17 @@ def get_env(self):
...
+            try:
+                status = self.proc_state.get(False)
+                print('PROCESS HAS FINISHED (POLL) with code', status, file=sys.stderr)
```
minrk: debug statement

jupyterhub/spawner.py
```diff
@@ -1308,12 +1358,17 @@ def get_env(self):
...
+                self.proc_state = status
                 self.clear_state()
-            return status
+                return status
+            except queue.Empty:
+                return None
+        elif self.proc_state is not None:
+            return self.proc_state
```
minrk: This branch can never occur, I think

jupyterhub/spawner.py
```diff
@@ -1254,6 +1270,36 @@ def get_env(self):
 
         return {"keyfile": key, "certfile": cert, "cafile": ca}
 
+    # intended to be run as intermediate process that
+    # calls self.authenticator.pre_exec_fn/post_exec_fn()
+    # after before and after starting the child process
+    # needed by PAMAuthenticator to handle PAM_sessions
+    def start_monitor_process(self, q, cmd, popen_kwargs):
+        if self.authenticator:
+                self.authenticator.pre_exec_fn(self.user, self)
```
minrk: Does this pre_exec_fn need to run as the user (i.e. after setuid)?

cmaves: Nope. PAM session modules may need elevated privileges to run such mounting drives.

consideRatio:
Hi @cmaves and @minrk, I'm looking through PRs and hoping to nudge them towards a merge, closure, or other concrete action. Any input on this?

cmaves:
From my point of view this pull request is dead. The core issue (#2120) still exists, but the required changes were more expansive and requires changes to how processes are spawned and forked. This should probably be done by somebody with more involvement then myself.

Issue #2323, opened 2018-11-22
[Feature Request] Allow option to configure table prefix


akhilputhiry:
I am trying to install jupyter hub in one of my existing database. Database already has a table called `users`. If you can provide an option to specify `prefix` it would be very helpful.

@yuvipanda

minrk:
Makes sense. Might be a bit tricky, but if there's a general option to sqlalchemy to add a table prefix, that would be useful.

PR #2327, opened 2018-11-25
[WIP] Support prefix in database tables

akhilputhiry:
https://github.com/jupyterhub/jupyterhub/issues/2323


akhilputhiry:
@minrk @yuvipanda How do we handle the alembic migrations with table prefixes? Let me know your thoughts.

PR #2330, opened 2018-11-27
Modified JSON-DICT Implementation from TEXT to Text

Deepakdubey90:


Deepakdubey90:
@minrk , Implemented the JSON field Implemented  as large-binary not Text becacuse oracle does not having data type as Text.

jupyterhub/orm.py
```diff
@@ -46,7 +46,7 @@ class JSONDict(TypeDecorator):
 
     """
 
-    impl = TEXT
+    impl = LargeBinary
```
minrk: Does it work to use `Text` instead of `TEXT`?

minrk: I understand that `Text` resolves to TEXT most places and CLOB on oracle.

minrk: Have you been able to test this with `Text`?

Deepakdubey90:
Yes

minrk:
Great! Are you planning to update the pull request? It is still using LargeBinary currently.

minrk:
Went ahead and made the LargeBinary -> Text substitution. Thanks for opening this @Deepakdubey90 !

PR #2343, opened 2018-12-03
Fix markdown link into rst

clemens-tolboom:


betatim:
Thanks for helping out fixing things up!

PR #2344, opened 2018-12-03
Spawner.py: do not quote Popen args

adelcast:
On Windows, Popen calls list2cmdline, which makes sure quotes are added
if needed. Currently, double quoting is happening, which breaks options
like notebook_dir, when full paths are used.

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -5,6 +5,7 @@ Contains base Spawner class & default implementation
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
+import ast
 import asyncio
 import errno
 import json
@@ -34,6 +35,23 @@ from .objects import Server
 from .traitlets import Command, ByteSpecification, Callable
 from .utils import iterate_until, maybe_future, random_port, url_path_join, exponential_backoff
 
+
+def _quote_safe(s):
+    """pass a string that is safe on the command-line
+
+    traitlets may parse literals on the command-line, e.g. `--ip=123` will be the number 123 instead of the *string* 123.
+    wrap valid literals in repr to ensure they are safe
+    """
+
+    try:
+        val = ast.literal_eval(s)
+    except Exception:
+        # not valid, leave it alone
+        return s
+    else:
+        # it's a valid literal, wrap it in repr (usually just quotes, but with proper escapes)
+        # to avoid getting interpreted by traitlets
+        return repr(s)
 
 class Spawner(LoggingConfigurable):
     """Base class for spawning single-user notebook servers.
@@ -831,7 +849,7 @@ class Spawner(LoggingConfigurable):
         args = []
 
         if self.ip:
-            args.append('--ip="%s"' % self.ip)
+            args.append('--ip=%s' % _quote_safe(self.ip))
 
         if self.port:
             args.append('--port=%i' % self.port)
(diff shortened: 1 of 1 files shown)


minrk:
Unfortunately, it's important that the quotes are preserved on the command-line (double-quotes is intended here) due to weirdness in traitlets handling cli arguments that could look like numbers. What's going wrong when quoting paths on Windows?

adelcast:
With quotes this is how my spawning looks like:

[I 2018-12-04 13:43:09.771 JupyterHub winlocalprocessspawner:55] Spawning 'C:\Python\3.6\python.exe' -m jupyterhub.singleuser --port=38730 '--notebook-dir="C:\ProgramData\notebooks"' '--NotebookApp.default_url="/lab"' --show-control-btn=True '--config=C:\Program Files\\Shared\\Jupyterhub\jupyter_notebook_config.py'

and the error:

[C 2018-12-04 13:43:10.998 SingleUserNotebookApp application:91] No such notebook dir: 'C:\\Program Files\\Shared\\Jupyterhub\\"C:\\ProgramData\\notebooks"'

Popen processes the args via the function list2cmdline, which takes care of adding quotes when needed. My understanding is that you don't quote since Popen will do that for you. Since Popen is receiving a list, it already knows where to put the quotes, as each element on the list represents one argument.

I could change the PR to have a Windows case that removes the quotes.

minrk:
Even on Windows, though, it is important that some of these get quotes added, we just might not be adding them correctly. Perhaps you can replace `"%s"` with `%r`?

adelcast:
Doesn't work either. As I mentioned above, this has to do with the way Popen works. Inside Popen, it calls list2cmdline, to create the correct cmd line. I printed a dump of subprocess.list2cmdline(cmd) and got:

"C:\Program Files\National Instruments\Shared\Skyline\Python\3.6\python.exe" -m jupyterhub.singleuser --port=7310 "--notebook-dir=\"C:\ProgramData\National Instruments\Skyline\JupyterHub\notebooks\"" --NotebookApp.default_url=\"/lab\" --show-control-btn=True "--config=C:\Program Files\National Instruments\Shared\Skyline\Jupyterhub\jupyter_notebook_config.py"

Running this exact line fails, since --notebook-dir has the extra quotes, which trip many things down the line (determining if a path is absolute  or not, etc). 

Another option would be to remove the quotes, if present, inside _notebook_dir_validate in singleuser.py. I just tried it and it works. Would that be acceptable?

minrk:
If you did repr, the line would be `--notebook-dir="C:\\ProgramData...` wouldn't it?

I don't like the double-quoting, but it is necessary for certain fields that could be valid numbers. Maybe we can do a better job and protect against the few cases where that could happen and revert the extra encoding.

Can you add this:

```python
import ast
def _quote_safe(s):
    """pass a string that is safe on the command-line

    traitlets may parse literals on the command-line, e.g. `--ip=123` will be the number 123 instead of the *string* 123.
    wrap valid literals in repr to ensure they are safe
    """

    try:
        val = ast.literal_eval(s)
    except Exception:
        # not valid, leave it alone
        return s
    else:
        # it's a valid literal, wrap it in repr (usually just quotes, but with proper escapes)
        # to avoid getting interpreted by traitlets
        return repr(s)
```

and pass the strings through this function? It should result in no quotes most of the time, but apply the safe wrapping in the edge cases where we need to.

adelcast:
Sorry for the late reply, but I was out last week.

The plain repr adds the problematic double escaping: "--notebook-dir=\"'C:\\ProgramData\\National Instruments\\Skyline\\JupyterHub\\notebooks'\""....

However, using your more generic solution, the problem was solved! I like your solution a whole lot more since it is specifically fixing the traitlets issue (adding quotes when traitlets may get confused). How do you want to move forward with this? Close this PR then add the _quote_safe function on another one?

minrk:
If you'd like to update the PR to add and use `_quote_safe`, I'm happy to merge it.

adelcast:
ok, just updated the PR

minrk:
Thanks!

PR #2346, opened 2018-12-04
push last_activity to the hub

minrk:
This adds periodic publishing of last_activity from single-user servers to the hub

This removes the reliance on network activity tracked in the proxy, which we won't have anymore when we switch to traefik.

It's an alternative to the culler script talking directly to the user servers, which is more complicated and slower (but also more scalable) and requires admin-access to be enabled, which may be undesirable.

### DIFF
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -708,7 +708,7 @@ class ActivityAPIHandler(APIHandler):
                         server_name,
                         isoformat(last_activity),
                     )
-                    spawner.last_activity
+                    spawner.last_activity = last_activity
                 else:
                     self.log.debug(
                         "Not updating server activity on %s/%s: %s < %s",
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -4,16 +4,17 @@
 # Distributed under the terms of the Modified BSD License.
 
 import asyncio
-from datetime import datetime
+from datetime import datetime, timedelta, timezone
 import json
 
 from async_generator import aclosing
+from dateutil.parser import parse as parse_date
 from tornado import  web
 from tornado.iostream import StreamClosedError
 
 from .. import orm
 from ..user import User
-from ..utils import admin_only, iterate_until, maybe_future, url_path_join
+from ..utils import admin_only, isoformat, iterate_until, maybe_future, url_path_join
 from .base import APIHandler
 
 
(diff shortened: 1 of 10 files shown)


jupyterhub/apihandlers/users.py
```diff
@@ -587,6 +588,139 @@ def on_finish(self):
             await self.send_event(failed_event)
 
 
+def _parse_timestamp(timestamp):
+    """Parse and return a utc timestamp
+
+    - raise HTTPError(400) on parse error
+    - handle and strip tz info for internal consistency
+      (we use naïve utc timestamps everywhere)
```
willingc: Not sure if the special character will render properly in autodoc but it may.

jupyterhub/apihandlers/users.py
```diff
@@ -587,6 +588,139 @@ def on_finish(self):
...
+        if not last_activity_timestamp and not servers:
+            raise web.HTTPError(
+                400,
+                "body must contain at least one of `last_activity` or `servers`"
+            )
+
+        if servers:
+            # validate server args
+            servers = self._validate_servers(user, servers)
+            # at this point we know that the servers dict
```
willingc: Nice explanation 👍

jupyterhub/singleuser.py
```diff
@@ -385,20 +391,30 @@ def _validate_static_custom_path(self, proposal):
             path = list(_exclude_home(path))
         return path
 
-    async def check_hub_version(self):
-        """Test a connection to my Hub
-
-        - exit if I can't connect at all
-        - check version and warn on sufficient mismatch
-        """
+    hub_http_client = Any()
```
willingc: Down the road it may make sense in the contributor guide to highlight this pattern (traitlet and dynamic validation) or improve the traitlets doc for this.

jupyterhub/tests/test_api.py
```diff
@@ -1549,6 +1549,92 @@ def normalize_token(token):
...
+        app,
+        "users/{}/activity".format(user.name),
+        headers={"Authorization": "token {}".format(token)},
+        data=json.dumps({"last_activity": utcnow().isoformat()}),
+        method="post",
+    )
+    r.raise_for_status()
+
+
+@mark.parametrize(
```
willingc: NIce!

jupyterhub/tests/test_api.py
```diff
@@ -1549,6 +1549,92 @@ def normalize_token(token):
...
+        ("", False),
+        ("exists", True),
+        ("exists", False),
+        ("nope", True),
+        ("nope", False),
+    ],
+)
+async def test_update_server_activity(app, user, server_name, fresh):
+    token = user.new_api_token()
+    sp = user.spawners["exists"]
```
willingc: @minrk, how is `sp` being used? A flag for spawner present? I'm not seeing it used in test.

dhirschfeld:
> *...when we switch to traefik.*

What's switching to `traefik`? Is there some discussion I can read or issue I can follow? I just want to figure out if I'll need to update my setup.

I haven't been following jupyterhub development so closely recently so this is (interesting) news to me. I'm interested in why *traefic* in particular - if there's any discussion around that fact. I've come across `traefik` in the context of k8s but there seem to be a baffling plethora of routers/proxies to choose from in that space!

jupyterhub/apihandlers/users.py
minrk: it should, since we're running sphinx with Python 3

minrk:
> I just want to figure out if I'll need to update my setup.

This shouldn't be something that you need to deal with, unless you want to. CHP will continue to work indefinitely, and likely continue to be the default for *jupyterhub* for some time. We're adding another option we are excited about.

> What's switching to `traefik`?

"switching" might have been the wrong word. jupyterhub 0.8 introduced an API for the Proxy implementation (like Spawner or Authenticator), making room for additional implementations that may be more scalable/resilient/whatever-you-want than configurable-http-proxy. https://github.com/jupyterhub/traefik-proxy is the first adopter of this API, which we hope to make the *recommended* implementation (assuming it works well), at least in some cases, and default for certain distros like zero-to-jupyterhub where we can eliminate the proxy as a single-point-of-failure because traefik+etcd can have multiple instances, persist the routing table across restarts, etc. Deploying traefik+etcd is more complicated than is worth doing for simple cases, so it isn't a replacement for CHP.

One feature of CHP that isn't part of the proxy *spec* is tracking activity. The notebook server is better at tracking activity anyway (more granular data), so adding the ability to push data from the notebook server to the Hub gets the information from a smarter source.

On Binder, we've also found that the notebook server's internal culling/self-shutdown mechanisms are much better than the coarse timers that the cull_idle script uses, which ultimately come from CHP.

minrk:
> I'm interested in why _traefik_ in particular

I missed this point in my first reply. We didn't do an exhaustive evaluation of the now many candidates, thanks to kubernetes. But traefik is one we've heard about and looking into it found it to be a good match:

1. it seems popular and active (good user base, Q&A, docs, etc.)
2. Being go, it's a simple single binary, so easy to install anywhere (even easier than configurable-http-proxy!)
3. its configuration is simple toml configuration (friendly for us Python folks)
4. it supports the required features (websockets, prefix routing, adding/removing routes without dropping existing connections)
5. it supports distributed configuration hosted from a key-value store like etcd and multiple instances for scaling (the main features we are looking for in a larger-scale/HA substitute for CHP)
5. it supports letsencrypt out of the box with two lines of config, making it really handy for getting working https without having to install a separate letsencrypt agent. We are already using traefik *for this feature alone* in the-littlest-jupyterhub

jupyterhub/tests/test_api.py
minrk: Updated to remove this since it's not actually needed.

Thanks for the review, @willingc!

dhirschfeld:
Thanks for taking the time to go into more detail, it's very useful/interesting - much appreciated!

It sounds like I'll definitely want to get up to speed with traefik...

PR #2347, opened 2018-12-04
Don’t mix no_patience and bad_spawn

minrk:
this introduces a race between the early RuntimeError being tested and the no_patience causing handlers to return early if async start isn’t complete.

With tornado coroutines, an early RuntimeError could be guaranteed to resolve promptly, but asyncio isn’t as consistent, possibly causing some of the recent flaky tests.

### DIFF
--- a/jupyterhub/tests/test_api.py
+++ b/jupyterhub/tests/test_api.py
@@ -718,7 +718,7 @@ def test_never_spawn(app, no_patience, never_spawn):
 
 
 @mark.gen_test
-def test_bad_spawn(app, no_patience, bad_spawn):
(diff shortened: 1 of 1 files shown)


dsblank:
I want a t-shirt that says "Don’t mix no_patience and bad_spawn"... words of wisdom for many situations!

betatim:
Thanks for chasing this down and fixing it!

minrk:
These tests passed first time, which is a soft indicator that we've tracked down at least one of the sources of flaky tests here. The "coroutine from another loop" one is *probably* unrelated to this, but I still don't have a good grasp of that one (some sense of a long-running thing leaking from one test module to the next)

PR #2353, opened 2018-12-06
Fix PAMAuthenticator is_admin

vilhelmen:
I deployed #2244's patch today and found out that is, in fact, bad. `os.getgrouplist` returns a list of gids directly, not a list of group objects as I and the tester assumed.

I must have gotten the os/grp/pwd functions mixed up during the final refactor and did not give this the final version the thorough testing I should have, as we had already pivoted to LDAP. I'm sorry :C


minrk:
Thanks!

PR #2355, opened 2018-12-11
accept token in Authorization header in services, not just url

chang-zhijie:


chang-zhijie:
Notebook can be authenticated by carrying a token in the url or header, but in jupyterhub, carrying a token in the header does not set a cookie.

minrk:
Thanks!

PR #2365, opened 2018-12-20
Change id selector to class selector for named servers

rcthomas:
The logic for starting/stopping/connecting to named servers is pretty repurpose-able for other ways of specifying the server name through the UI.  We are experimenting with a setup where we have a set of pre-defined names in hidden fields in another kind of UI layout for spawning.  Since there is more than one element involved for starting a new server, just changing this from ID to class selector appears to work just fine for the standard named-server implementation and our own.


minrk:
Thanks!

PR #2375, opened 2019-01-08
Limit max named server per user

dtaniwaki:
I'd like to allow users to have more than one server but don't want them to make hundreds of servers because the resource is limited. This config allows us to set the limit on the number of named servers per user.

### DIFF
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -350,8 +350,13 @@ class UserServerAPIHandler(APIHandler):
     @admin_or_self
     async def post(self, name, server_name=''):
         user = self.find_user(name)
-        if server_name and not self.allow_named_servers:
-            raise web.HTTPError(400, "Named servers are not enabled.")
+        if server_name:
+            if not self.allow_named_servers:
+                raise web.HTTPError(400, "Named servers are not enabled.")
+            if self.named_server_limit_per_user > 0 and server_name not in user.orm_spawners:
+                named_spawners = list(user.all_spawners(include_default=False))
+                if self.named_server_limit_per_user <= len(named_spawners):
+                    raise web.HTTPError(400, "Named servers can not be created any more.")
         spawner = user.spawners[server_name]
         pending = spawner.pending
         if pending == 'spawn':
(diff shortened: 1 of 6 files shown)


share/jupyterhub/templates/home.html
```diff
@@ -26,10 +26,12 @@ <h2>
 
   <p>
   In addition to your default server,
-  you may have additional servers with names.
+  you may have additional {% if named_server_limit_per_user > 0 %}{{ named_server_limit_per_user }} {% endif %}server(s) with names.
   This allows you to have more than one server running at the same time.
   </p>
 
+  {% set named_spawners = user.all_spawners(include_default=False)|list %}
```
minrk: Is there a reason this was created as a separate variable when its use hasn't changed?

jupyterhub/apihandlers/users.py
```diff
@@ -350,8 +350,13 @@ class UserServerAPIHandler(APIHandler):
     @admin_or_self
     async def post(self, name, server_name=''):
         user = self.find_user(name)
-        if server_name and not self.allow_named_servers:
-            raise web.HTTPError(400, "Named servers are not enabled.")
+        if server_name:
+            if not self.allow_named_servers:
+                raise web.HTTPError(400, "Named servers are not enabled.")
+            if self.named_server_limit_per_user > 0 and server_name not in user.orm_spawners:
+                named_spawners = list(user.all_spawners(include_default=False))
```
minrk: No need to cast to list before computing len. Dicts have a len.

jupyterhub/apihandlers/users.py
```diff
@@ -350,8 +350,13 @@ class UserServerAPIHandler(APIHandler):
...
+                if self.named_server_limit_per_user <= len(named_spawners):
+                    raise web.HTTPError(400, "Named servers can not be created any more.")
```
minrk: Let's include the limit in the error:

```suggestion
                    raise web.HTTPError(
                        400,
                        "User {} already has a the maximum of {} named servers."
                        "  One must be deleted before a new server can be created".format(
                            name,
                            self.named_server_limit_per_user
                        )
```

share/jupyterhub/templates/home.html
dtaniwaki: I was going to disable buttons to add new server based on the length of this list, but I realized it needs many changes in the JS. I think we can improve it in another PR, then we will need to use this variable.

jupyterhub/apihandlers/users.py
dtaniwaki: I remember that I wrapped it with `list()` because `user.all_spawners(include_default=False)` returns `gen()` instead of `dict`.

dtaniwaki:
@minrk I updated the PR. Would you review it again?

minrk:
Excellent, thanks!

PR #2377, opened 2019-01-08
add upgrade to pip install

minrk:
to ensure version compatibility

fixes failures due to too-new pytest-cov


minrk:
also pinning oauthlib <3 which is causing invalid redirect_uri errors

PR #2382, opened 2019-01-16
update requirements: use oauthlib<3.0 because of redirect issue

katsar0v:
Fixes a redirecting issue when spawning containers with dockerspawner


minrk:
Thanks! Just did this in #2377

PR #2387, opened 2019-01-19
Fix travis CI failure

dtaniwaki:
`pytest-cov` 2.6.1 which was release 2 weeks ago contains newer `pytest` version constraint. 
https://github.com/pytest-dev/pytest-cov/pull/253
So we should pin `pytest` newer than 3.6.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -7,7 +7,7 @@ cryptography
 html5lib  # needed for beautifulsoup
 pytest-cov
 pytest-asyncio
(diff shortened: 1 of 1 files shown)


dtaniwaki:
This PR may not be necessary after https://github.com/jupyterhub/jupyterhub/pull/2377

willingc:
Hi @dtaniwaki, Thanks for the PR. 👍  It looks as if #2377 fixed the issue. Do you want to go ahead and close this PR?

Issue #2129, opened 2018-09-04
JupyterHub spawns multiple server for one user


Jamesits:
**Describe the bug**
A Linux user with one UID but multiple username can make JupyterHub (or the systemd spawner? I'm not familiar with the spawner so not very sure if I'm reporting to the correct repository) spawn multiple servers, one per username, effectively bypass some security and resource limitations.

**To Reproduce**
1. Link a UID to multiple username
2. login using all those usernames from different browser sessions
3. try open different notebooks from different sessions
4. look at the running tab, see if they are the same server

**Expected behavior**
There should be only one server per UID.

**Screenshots**
Nope

**Server**
 - OS: Fedora 28
 - JupyterHub: 0.9.2

**Desktop (please complete the following information):**
 - OS: Windows 10 1803
 - Browser: Firefox
 - Version 61.0.2

PS: If you wonder why this kind of setup will be done in production, this is common in systems with kerberos authentication, where you might have multiple names mapped to the same uid. Users can login using `username`, `domain\username` or `username@domain`.

minrk:
If multiple usernames map onto one user, this should be handled in `.authenticate` or possibly `.normalize_username`. What Authenticator are you using?

willingc:
Would you be able to share your debug log with us (please remove any private information)? Thanks.

minrk:
Since this is PAM, my first thought was to have a default normalize_username that runs through username->userid->username, which should get a deterministic answer:

```python
uid = pwd.getpwnam(username).pw_uid
username = pwd.getpwuid(uid).pw_name
```

Can you run that snippet for your deployment and see if it works?

But then I read about some use cases where different users *should* have the same UID, so I'm not sure about this as the default behavior.

Regardless of what we end up doing by default, in the meantime you can implement this as configuration for your deployment:

```python
# in jupyterhub_config.py
from jupyterhub.auth import PAMAuthenticator

class MyPAMAuthenticator(PAMAuthenticator):

    def normalize_username(self, username):
        # pass through uid to ensure that all names that
        # correspond to one uid map to the same jupyterhub user
        uid = pwd.getpwnam(username).pw_uid
        return super().normalize_username(pwd.getpwuid(uid).pw_name)

c.JupyterHub.authenticator_class = MyPAMAuthenticator
```

Jamesits:
@minrk Sorry for the late reply, I've been on a trip for some day with no computer access.

The snippet
```python
[...]

Issue #1671, opened 2018-02-22
PAMAuthenticator shouldn't normalize usernames


christiaanjs:
When I try to authenticate with PAM with a mixed-case username (in this case it is `swanecAdm`) I get this error:
```
    Traceback (most recent call last):
      File "/usr/lib/anaconda3/lib/python3.6/site-packages/tornado/web.py", line 1512, in _execute
        result = yield result
      File "/usr/lib/anaconda3/lib/python3.6/site-packages/jupyterhub/handlers/login.py", line 94, in post
        yield self.spawn_single_user(user)
      File "/usr/lib/anaconda3/lib/python3.6/site-packages/jupyterhub/handlers/base.py", line 475, in spawn_single_user
        yield gen.with_timeout(timedelta(seconds=self.slow_spawn_timeout), finish_spawn_future)
      File "/usr/lib/anaconda3/lib/python3.6/site-packages/jupyterhub/handlers/base.py", line 445, in finish_user_spawn
        yield spawn_future
      File "/usr/lib/anaconda3/lib/python3.6/site-packages/jupyterhub/user.py", line 439, in spawn
        raise e
      File "/usr/lib/anaconda3/lib/python3.6/site-packages/jupyterhub/user.py", line 378, in spawn
        ip_port = yield gen.with_timeout(timedelta(seconds=spawner.start_timeout), f)
      File "/usr/lib/anaconda3/lib/python3.6/types.py", line 248, in wrapped
        coro = func(*args, **kwargs)
      File "/usr/lib/anaconda3/lib/python3.6/site-packages/jupyterhub/spawner.py", line 968, in start
        env = self.get_env()
      File "/usr/lib/anaconda3/lib/python3.6/site-packages/jupyterhub/spawner.py", line 960, in get_env
        env = self.user_env(env)
      File "/usr/lib/anaconda3/lib/python3.6/site-packages/jupyterhub/spawner.py", line 947, in user_env
        home = pwd.getpwnam(self.user.name).pw_dir
    KeyError: 'getpwnam(): name not found: **swanecadm'**
```

This is caused by the case normalization in [`Authenticator.normalize_username`](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/auth.py#L189).

As a hack, commenting out this line fixes the problem and I can login. Maybe `PAMAuthenticator` should override this behaviour. I am happy to have a go at fixing this myself.

minrk:
[...]

Issue #1617, opened 2018-01-11
HUB API refuses connections 80% of times


BerserkerTroll:
**UPD**
I suspect that the reason is that k8s load balancer does not notice Pods removals and still redirects to their dead IPs.
But I will keep this bugreport open till I am 100% sure.
**END OF UPD**

I am deploying JupyterHub into k8s cluster. k8s version is 1.8.6. I use version `0.8.1` of the jupyterhub docker image (with KubeSpawner added to it) and `jupyterhub/singleuser:0.9` as a single user notebook image. My current cluster consists of 2 nodes: master and working node and I use [Flannel](https://github.com/coreos/flannel) as container network interface. One working node means that both hub and single user servers run on the same node.

The problem is that when I login to hub, most of the time it gives me `500 : Internal Server Error`
```
Failed to connect to Hub API at 'http://10.98.186.233:80/hub/api'. 
Is the Hub accessible at this URL (from host: jupyter-<...>)?
```
I would understand if this always happened, which would mean that I use inkorekt Hub API IP, BUT!!! if I go to Control Panel—My Server, nine times I will see this error, but for the tenth time I will get my directory list as if nothing happened before!

**I am not sure who to blame**: JupyterHub or k8s cluster infrastructure. Currently what I have is just hub and single user notebook, the CPU is not loaded and there is no memory pressure. Both hub and notebook are running on the same node, so it can't be networking issue. At least not physical networking problem. Could it be the problem in the Flannel overlay network? Or k8s load balancing is broken? Currently I have decided to submit a bugreport here, because I have no problems accessing JupyterHub login page, which means that the "routing" to its service/endpoint works fine.

A fragment from the single user Pod log:
```
[E 2018-01-11 13:02:58.173 SingleUserNotebookApp auth:249] Error connecting to http://10.98.186.233:80/hub/api: HTTPConnectionPool(host='10.98.186.233', port=80): Max retries exceeded with url: /hub/api/authorizations/token/d83ae03f-e10d-40f5-b27e-0bbe0601f24f (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f5f76891748>: Failed to establish a new connection: [Errno 111] Connection refused',))
[...]

PR #2397, opened 2019-01-28
pam_normalize_username option: round-trip usernames through PAM to normalize

rkdarst:
I have had this semi-completed for months now.  I realized that #2129 (multiple servers per user, root problem was that usernames weren't normalized.).  This basically takes the solution of that issue and makes it a built-in option, which I think will be quite helpful to many people.

Other thoughts:
* I don't know how this can be tested - can we assume PAM is available on the test runners?  And the change is so short that mocking it or some such is almost short-circuting all relevant logic.  Advice requested.
* I seem to recall some other issue about normalizing usernames to lowercase breaks something or the other.  Could this PAM normalization be used as an alternative for lowercase normalization?  Currently lowercase normalization is also done even if `pam_normalize_username` is true.

I have gently tested this myself, but it's not yet in any of my production systems.

### DIFF
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -787,7 +787,9 @@ class PAMAuthenticator(LocalAuthenticator):
             import pwd
             uid = pwd.getpwnam(username).pw_uid
             username = pwd.getpwuid(uid).pw_name
-        return super().normalize_username(username)
+            username = self.username_map.get(username, username)
+        else:
+            return super().normalize_username(username)
 
 class DummyAuthenticator(Authenticator):
     """Dummy Authenticator for testing
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -678,6 +678,16 @@ class PAMAuthenticator(LocalAuthenticator):
         """
     ).tag(config=True)
 
+    pam_normalize_username = Bool(False,
+        help="""
+        Round-trip the username via PAM lookups to make sure it is unique
+
+        PAM can accept multiple usernames that map to the same user,
+        for example DOMAIN\\username in some cases.  To prevent this,
+        convert username into uid, then back to uid to normalize.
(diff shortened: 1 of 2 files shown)


rkdarst:
#1617 is an issue with lowercase normalization, but I thought I remembered something more recently, too.

jupyterhub/auth.py
```diff
@@ -769,6 +779,15 @@ def post_spawn_stop(self, user, spawner):
...
 
+    def normalize_username(self, username):
+        """Round-trip the username to normalize it with PAM
+
+        PAM can accept multiple usernames as the same user, normalize them."""
+        if self.pam_normalize_username:
+            import pwd
+            uid = pwd.getpwnam(username).pw_uid
+            username = pwd.getpwuid(uid).pw_name
+        return super().normalize_username(username)
```
minrk: Should we skip `super()` in the pam case and make it `else: super()`? This would fix case-sensitivity issues. How does the DOMAIN\name case occur?

rkdarst: That's a great idea - done.  I also added the username_map support, it's probably not needed with PAM but I guess can't hurt to be consistent...

In my case the weird usernames of DOMAIN\username and email@domain came about because we linked with university ActiveDirectory, and however the linking was done these turned out to be valid usernames for authentication, but not the canonical names in the filesystem.

rkdarst:
I wonder if pam_normalize_username should be the default?  It seems like it would at least not hurt in most cases.  Disadvantages: would it ever fail?  Take longer to do the lookups?  Most importantly, it changes from the previous behavior so if someone upgrades, their usernames may no longer match, so it would have to be very well documented.  If you ask me I'd say safer to not change.

Zsailer:
@rkdarst minor correction just for reference. The relevant issue is #1671.

PR #2418, opened 2019-02-15
Adopt black, pre-commit

### LINKED ISSUE #2416: Autoformat with black?
I'm not a fan of style-checking in CI that requires human intervention, but auto-formatters like [black](https://pypi.org/project/black/) allow us to eliminate code style as a review point, even if the auto-formatter doesn't always make the best decisions. I think projects with style-enforcement are generally contributor-hostile, but projects with auto-formatting can eliminate contributor stress about "does my style match?"

The biggest problem with adopting an auto-formatter in an existing project is that it'll make one huge PR that's not meaningful, which is a bit of a mess for version control history and understanding the code base. If we adopt black, I would aim to enable the skip-string-normalization feature, since that would cause the biggest churn (and I frankly happen to disagree with the design of black there).

What do folks think? If there's interest, I can investigate picking up https://pre-commit.com which offers some pretty handy hooks for quick fixes.

/cc @willingc

minrk:
pre-commit hook runs various autoformatting commands, including:

- consistent ordering of Python imports
- black for Python formatting
- whitespace/trailing newline fixes
- json/yaml validation
- flake8 (just for correctness)

The goal of this is *not* to bug folks about things they need to do, but rather to autoformat so they don't need to. So the one I'm least sure about is flake8, since it doesn't fix things, it just tells you about issues it finds, which may or may not need fixing. We have a pretty minimal flake8 configuration, though, so I think it should be okay.

There's one big-ol' commit that results from running:

    pre-commit run --all-files

to get everything happy with the hook, preventing future PRs from mixing formatting updates with actual changes.

I also added to travis config a message when linting fails that running `pre-commit run` *should* fix whatever is failing.

I added installation/setup steps for pre-commit to CONTRIBUTING.md

closes #2416


docs/source/conf.py
```diff
@@ -1,12 +1,13 @@
...
 #
-import sys
 import os
 import shlex
+import sys
 
-# For conversion from markdown to html
 import recommonmark.parser
 
+# For conversion from markdown to html
```
willingc: Let's remove this since it was for the import in line 7.

jupyterhub/apihandlers/auth.py
```diff
@@ -1,25 +1,23 @@
...
-from datetime import datetime
 import json
-from urllib.parse import (
-    parse_qsl,
-    quote,
-    urlencode,
-    urlparse,
-    urlunparse,
-)
+from datetime import datetime
```
willingc: I'm not really a fan of the multi-line imports, but could be helpful for new contributors.

minrk: This is from https://github.com/asottile/reorder_python_imports, and it gives its [reasons why](https://github.com/asottile/reorder_python_imports#why-this-style) to minimize git conflicts (fewer things per line means fewer chances of conflict). The two choices appear to be this and [isort](https://github.com/timothycrosley/isort/), and looking at them both, isort has one deal-breaker for me: no ability to `#noqa` a line or section for manual ordering when import order matters (https://github.com/timothycrosley/isort/issues/679).

We could also skip the import-ordering and do black-only.

docs/source/conf.py
minrk: Yeah, looks like a few import-related comments got moved to weird places. Will have to check those

jupyterhub/apihandlers/auth.py
willingc: Thanks for the links @minrk. Let's keep the import-ordering as you have it. I agree that I like your choice more than isort.

Looks like the PR needs a rebase and then go ahead and merge. Thanks for doing this. I think it will keep things tidy as more folks add contributions.

minrk: Just finished merging the outstanding PRs I wanted to before landing this one to avoid conflicts. Will rebase this one in the AM and merge.

Thanks for reviewing!

PR #2431, opened 2019-02-21
Entrypoint is necessary to build the API docs of most authenticators

Carreau:


Carreau:
Merging this to try if it fixes some of the RTD issues.

PR #2433, opened 2019-02-22
document admin_group

Carreau:
... I was going to implement it but it's there.


Carreau:
(at least it's on master...)

PR #2434, opened 2019-02-22
Add documentation on running the proxy separate from the hub [wip]

rkdarst:
- It took me a bit longer than I would have liked for me to figure out
  how to run the proxy separate from the hub.  When I had to do this a
  second time for a different hub, it also took me too long.
- This adds a page dedicated to running the proxy separate from the
  hub, since it is relatively easy and has a high usability
  improvement.
- Currently work in progress.

=====

I have been meaning to write this for a while.  I got something started, it's not perfect but better than what I could find so far.  Should this be continued and made mergeable?  (it's possible that there is some other good docs that I just can't find now...).

I'd welcome help to make this better, since I don't think I know everything relevant... in particular, docs on any aids to do this more better.  The docker section should probably be improved at least.  Anyone, feel free to push directly here.


minrk:
great to have docs on this! It looks like this is linking to a page that hasn't been added yet. Maybe a missing `git add`?

rkdarst:
Oops... normally I'm better at this.  Somehow I even remember doing it right this time.  Anyway, it's there now.

minrk:
Looks like you're the first to run into the `pre-commit` checks! Want to try running the setup in CONTRIBUTING.md and see if that helps? If that's installed, these things should be fixed automatically. Also if you find this annoying to deal with, that's useful feedback as well.

rkdarst:
I was watching the pre-commit stuff, just waiting to run into it...

It was a bit annoying because "pre-commit run" was failing with a PermissionDenied error.  In the end, it's because I was on a cluster and my cache dir was redirected to a Lustre filesystem, where some call (xattrs?) failed.  I eventually fixed it, and this probably counts as a "personal problem" - let the user who does this beware.

`pre-commit run` was not helpful in fixing it, because the commits were already there.  But `pre-commit run --all-files` fixed old commits - I will add this to the docs.

Now it passes locally.

minrk:
Yeah, I'm new to pre-commit, too. Indeed, I think `pre-commit run --all-files` is usually what you want to do. The `pre-commit run` only looks at *staged* files (what I really want it to run on in my workflow is all tracked, unchanged files, not only staged files). There are good not-messing-stuff-up reasons they do what they do, but it's a little surprising.

I turned it up to eleven as a bit of an experiment. e.g. the trailing newline check that caught you here is super unimportant (I would never reject a PR because of that), but the hope is that pre-commit helps the fixes be automatic rather than something folks need to pay attention to. I actually wish it wouldn't set an error code and just make the fixes quietly. We're still learning if the automatic fixes are nicer than ignoring the unimportant issues like that. It may be that we end up cutting back to just the code formatting.

Might be worth reporting the permission issue to pre-commit and recording here in more detail in case others run into it.

PR #2444, opened 2019-02-27
Contributing: hint on pre-commiting commits that are already there

rkdarst:
Hint based on my experience on using `pre-commit`.

When it did fix something, I had to `git add ` the changes and try the commit again.  I'm not sure if that was because I use `git add` first or just an artifact of my workflow.  Either way, seems like figuring that out is part of basics of git, so I didn't mention that.


minrk:
I think you're right, it only works on staged things, so when it makes changes it doesn't stage them so you get a chance to review before staging the new changes. Not *quite* achieving my goal of "autoformat and forget about it"

PR #2446, opened 2019-02-28
persist user_options

minrk:
remember user_options from the previous run

this allows user options set via spawn form (or previous api request) to be re-used when restarting e.g. a named server via the api

### DIFF
--- a/jupyterhub/orm.py
+++ b/jupyterhub/orm.py
@@ -216,6 +216,7 @@ class Spawner(Base):
 
     started = Column(DateTime)
     last_activity = Column(DateTime, nullable=True)
+    user_options = Column(JSONDict)
 
     # properties on the spawner wrapper
     # some APIs get these low-level objects
--- a/jupyterhub/orm.py
+++ b/jupyterhub/orm.py
@@ -3,7 +3,6 @@
 # Distributed under the terms of the Modified BSD License.
 import enum
 import json
-import pickle
 from base64 import decodebytes
 from base64 import encodebytes
 from datetime import datetime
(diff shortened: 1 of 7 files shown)


jupyterhub/orm.py
```diff
@@ -3,6 +3,9 @@
 # Distributed under the terms of the Modified BSD License.
 import enum
 import json
+import pickle
```
willingc: @minrk Trying to reason through if there are security implications given that I think, if I'm understanding correctly, that the information is coming from a form with user input data.

What would be an unrecognized object? A new field in the option form offered by the deployer of the hub?

minrk: There are indeed questions about using pickle. Injecting malicious pickle data into the database would indeed be an attack, but if you can inject data into the database, you can also already grant yourself access to everything on JupyterHub and all user servers, so I'm not that concerned about adding one more thing malicious folks with write access to the db can do.

The reason I added this is the upload-form test was failing because the form upload is returning bytes. Initially, I started supporting bytes only, but realized folks could be doing whatever they want, so I chose pickle. That said, after your comment I've switched this PR back to extending JSON only to raw bytes, and clarified the requirement that user_options be JSON+bytes.

willingc: Sounds reasonable. Probably good to go conservative first and if there is strong interest beyond that then to revisit.

meeseeksmachine:

https://discourse.jupyter.org/t/user-options-are-not-returned-by-users-rest-api-after-jupyterhub-is-restarted/12641/1

PR #2447, opened 2019-02-28
Allow for proper logging of IP addresses when using separate TLS termination

mpetroff:
When using separate TLS termination, all IP addresses are currently logged as 127.0.0.1, since that's the last IP address in the `X-Forwarded-For` header (added by the Node HTTP proxy).

Trusting localhost (127.0.0.1) X-headers causes Tornado to use the next IP address in the header, the one added by the TLS termination proxy, as the `remote_ip`, resulting in the correct IP address being logged.

The potential risk to this approach is that local users could make local traffic look like it's coming from elsewhere, but I don't think this is a significant risk.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -466,6 +466,18 @@ class JupyterHub(Application):
         """,
     ).tag(config=True)
 
+    trusted_downstream_ips = List(
+        Unicode(),
+        help="""Downstream proxy IP addresses to trust.
+
+        This sets the list of IP addresses that are trusted and skipped when processing
+        the `X-Forwarded-For` header. For example, if an external proxy is used for TLS
+        termination, its IP address should be added to this list to ensure the correct
+        client IP addresses are recorded in the logs instead of the proxy server's IP
+        address.
+        """,
+    ).tag(config=True)
+
     ip = Unicode(
         '',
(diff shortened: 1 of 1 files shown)


jupyterhub/app.py
```diff
@@ -2291,7 +2291,10 @@ def ask():
 
         # start the webserver
         self.http_server = tornado.httpserver.HTTPServer(
-            self.tornado_application, ssl_options=ssl_context, xheaders=True
+            self.tornado_application,
+            ssl_options=ssl_context,
+            xheaders=True,
+            trusted_downstream=['127.0.0.1'],
```
minrk: The http proxy may not be on localhost. Can you add this as a `trusted_downstream_ips = List(Unicode(), default_value=['127.0.0.1'], config=True, help="""explanation...""")` trait so that it can be set when the proxy might be remote?

mpetroff:
I'm terminating TLS with Apache running on the same server as JupyterHub, under Ubuntu 18.04. As the IP addresses were being recorded incorrectly, Fail2ban wasn't working properly. The primary reason I'm using the Apache proxy is to make Certbot auto certificate renewal more convenient.

As requested, I made the trusted downstream IP address list configurable. I left an empty default, since it's slightly more secure, and this seems to be an uncommon issue.

PR #2455, opened 2019-03-06
Added -p 8000:8000 to jupyterhub Docker quick start command.

javabrett:
If port 8000 isn't published, the quick-start command to launch a JupyterHub Docker image isn't very useful.  Adding this means it can be accesses on http://localhost:8000 .


minrk:
Good catch!

Issue #2464, opened 2019-03-11
Question: user-redirect extension


msuperina:
As per [UserUrlHandler](https://github.com/jupyterhub/jupyterhub/blob/b1606f21e6afbcce6b9fd4bbdd07b29448a8f3f5/jupyterhub/handlers/base.py#L1143) 
```
If a user, alice, requests /user/bob/notebooks/mynotebook.ipynb,
she will be redirected to /hub/user/bob/notebooks/mynotebook.ipynb,
which will be handled by this handler,
which will in turn send her to /user/alice/notebooks/mynotebook.ipynb.
Note that this only occurs if bob's server is not already running.
```

I was thinking of creating an extension to redirect all alice requests to `/user/bob/notebooks/mynotebook.ipynb` to `/user/alice/notebooks/mynotebook.ipynb` whether or not bob's server is running.
And I was wondering, is there any scenario I should take into account which could turn this extension into a bad idea ?

minrk:
I don't think this is how it should be done. Instead, the JupyterLab "copy shareable link" should be modified via an extension to copy the actually correct shareable url (`/user-redirect/notebooks/mynotebook.ipynb`), rather than trying to get the wrong url to do what might sometimes be the right thing.

One of the reasons being, alice visiting bob's URL *may well make sense* if bob has given alice permission to access his server, e.g. for real-time collaboration on documents. Trying to guess what should happen is something I'd like to do less of, not more, since it's a source of various problems and lots of complexity.

Closing as duplicate of #2017

PR #2478, opened 2019-03-20
WIP: Redirect if source user's server is running

bobbychopra:
Redirects notebook URL to user's server if source user's server is running

### DIFF
--- a/jupyterhub/services/auth.py
+++ b/jupyterhub/services/auth.py
@@ -979,5 +979,13 @@ class HubOAuthCallbackHandler(HubOAuthenticated, RequestHandler):
         if user_model is None:
             raise HTTPError(500, "oauth callback failed to identify a user")
         app_log.info("Logged-in user %s", user_model)
+
+        # TODO:: check some flag for user auto-redirected
+        user_server = user_model.get('server', '')
+        if next_url.startswith('/user/') and not next_url.startswith(user_server):
+            def remove_singleuser_server(uri):
+                return uri[uri.find('/', len('/user/'))+1:]
+            new_url = user_server + remove_singleuser_server(next_url)
+            self.redirect(new_url)
         self.hub_auth.set_cookie(self, token)
         self.redirect(next_url or self.hub_auth.base_url)


minrk:
I don't believe that this redirect should be performed. At most, the desired URL should be displayed as a suggested "Did you mean..." link, with an explanation about the `user-redirect` url. Too many automatic redirects guessing at what users wanted has been a major problem for JupyterHub.

Would you like to take a stab at implementing that?

msuperina:
@minrk to give you more context https://github.com/jupyterhub/jupyterhub/issues/2464 was a question in preparation of this PR. 
The motivation behid this question is that we have users who want to share notebooks by copy pasting the url from the browser. 
We have already implemented extensions to generate the /user-redirect link with a single click but those users do not find those extensions obvious enough.

minrk:
What about my proposal above, including the link in the error page? That would make it just one extra click away, but I would like to stick strongly to the fact that this is a result of an error made by the user, and should be treated as such.

bobbychopra:
The suggestion _"Did you mean..." link with an explanation about the user-redirect url_ is an option. It isn't something that we are keen to implement as we feel that user: bob should have accesses the notebook using user: alice url's link if alice's server is stopped or is running

The current behavior is confusing for our users because:
- user: bob **can** accesses the notebook using user: alice url's link if alice's server is stopped, but 
- user: bob **cannot** accesses the notebook and receives a 403 HTTP Error using user: alice url's link if alice's server is running

We get tons of complaints from user: bob why the user:alice's url suddenly stopped working.

minrk:
We can get to consistent behavior with the above proposal, because we can then stop redirecting users from a page they cannot access to a guessed but not necessarily correct URL on their own server. We can fully consistently show the error page with a "Did you mean..." link, which users only click if that's where they wanted to go.

Given that the current redirect has proved to be a huge problem, adding it more places is not something I'm comfortable accepting. We need to move in the opposite direction, but we can still achieve the goal of getting users where they intend to go with an appropriate link on the error page.

bobbychopra:
Your last proposal to show the "Did you mean ..." link would make the behavior pretty consistent. 

To confirm:
If user: bob accesses the notebook using user: alice url's link if alice's server is stopped, **we will show HTTP 403** and present a "Did you mean ..." link
If user: bob accesses the notebook using user: alice url's link, if alice's server is stopped, we show HTTP 403 and can present a "Did you mean ..." link

Please let me know if you have a different approach.

minrk:
@bobbychopra yes, that's exactly right. Same 403 error in both cases, same "Did you mean..." link to the user's own server.

minrk:
oh, and the "Did you mean" link should probably be to `/user-redirect/...`, not `/user/bob/...`. That way, it shows users the *correct* shareable link, not just the final destination.

consideRatio:
@bobbychopra thank you for suggesting an improvement to this project!

I agree with Min that it would be more reliable in the long term to have a change that doesn't automatically redirect a user, but instead have a error with a suggestion added to it. My dream outcome would be to have a 403 error indicating that JupyterHub has spawned a server for user X and you tried to access it, but you could start / go to your own server if that is desired instead.

PR #2484, opened 2019-03-25
Fix server name parsing in UserUrlHandler

rcthomas:
Assuming named servers is enabled, the `user_path` when a user tries to go directly to a server seems to come in with a leading "/".  This was leading to the server name always ending up as "" which is the default.  I've changed the split to take this slash into account.

I've tested this gives the desired behavior (e.g. launch button page).  There might be a more right way to extract the server name from the `user_path` parameter.

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -1202,7 +1202,7 @@ class UserUrlHandler(BaseHandler):
             and self.current_user
             and user_name == self.current_user.name
         ):
-            server_name = user_path.split('/', 2)[1]
+            server_name = user_path.lstrip('/').split('/', 1)[0]
             if server_name not in self.current_user.orm_user.orm_spawners:
                 # no such server, assume default
                 server_name = ''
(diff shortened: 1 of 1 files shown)


jupyterhub/handlers/base.py
```diff
@@ -1202,7 +1202,7 @@ def non_get(self, user_name, user_path):
             and self.current_user
             and user_name == self.current_user.name
         ):
-            server_name = user_path.split('/', 1)[0]
+            server_name = user_path.split('/', 2)[1]
```
minrk: `[1]` will fail when this is empty. Instead, let's strip the leading / and still get the first element:

```suggestion
            server_name = user_path.lstrip('/').split('/', 1)[0]
```

jupyterhub/handlers/base.py
```diff
@@ -1292,7 +1292,7 @@ def non_get(self, user_name, user_path):
         server_name = ''
         if self.allow_named_servers:
             # check if url prefix matches an existing server name
-            server_name = user_path.split("/", 1)[0]
+            server_name = user_path.split("/", 2)[1]
```
minrk: same here

```suggestion
            server_name = user_path.lstrip('/').split('/', 1)[0]
```

minrk:
Good catch! I've added a tiny tweak to ensure it gets the right answer even for empty paths.

rcthomas:
Thought there would be a more right way to do it, thanks!

minrk:
Thanks!

PR #2490, opened 2019-03-29
Fix 1.0 date in changelog

mathematicalmichael:
fix date


minrk:
Perfect, thanks!

mathematicalmichael:
just out of curiosity, why did the Travis builds fail?

PR #2494, opened 2019-04-01
include retry link after failed spawn

### LINKED ISSUE #2492: Clearing spawn failure inside spawner
**Describe the bug**

rsdockerspawner is a customer spawner that does automatic resource management. The only way I could figure out to return "no resources" was to [throw a 429](https://github.com/radiasoft/rsdockerspawner/blob/9fb308bddc255216aaa896b753b8d98dd2df670e/rsdockerspawner/rsdockerspawner.py#L464), but this is seen [as an error in user.py](https://github.com/jupyterhub/jupyterhub/blob/689a312756483db77b874115d829f93a9d6fb96b/jupyterhub/user.py#L616). 

There's no way to clear that error, which [ends up getting stored in base.py](https://github.com/jupyterhub/jupyterhub/blob/689a312756483db77b874115d829f93a9d6fb96b/jupyterhub/handlers/base.py#L835).

Also, the error message that is passed to `tornado.web.HTTPError` is not shown to the user, just the generic 429 Too Many Requests error.

While there is a link to the home page, it would be better to have a (big :) button that says "Try Again". People don't see the link.

**To Reproduce**

Login twice with different GitHub user names to http://alpha-jupyter.radiasoft.org. The garbage collector is set to a very short interval so it should produce the error.

**Expected behavior**

The spawner should be able to manage the resources in the spirit of [EAFP](https://en.wikipedia.org/wiki/EAFP). A 429 is a known exception for too few resources.

**Screenshots**
n/a

**Desktop (please complete the following information):**
n/a

**Additional context**
n/a

minrk:
implicit spawn check doesn't make sense anymore, since there are no implicit spawns

instead, render not_running page with error message and retry link:

Screen Shot 2019-04-01 at 12 48 30

closes #2492

### DIFF
--- a/jupyterhub/handlers/pages.py
+++ b/jupyterhub/handlers/pages.py
@@ -290,6 +290,7 @@ class SpawnPendingHandler(BaseHandler):
                 server_name=server_name,
                 spawn_url=spawn_url,
                 failed=True,
+                failed_message=getattr(exc, 'jupyterhub_message', ''),
             )
             self.finish(html)
             return
--- a/jupyterhub/handlers/pages.py
+++ b/jupyterhub/handlers/pages.py
@@ -291,6 +291,7 @@ class SpawnPendingHandler(BaseHandler):
                 spawn_url=spawn_url,
                 failed=True,
                 failed_message=getattr(exc, 'jupyterhub_message', ''),
+                exception=exc,
             )
             self.finish(html)
             return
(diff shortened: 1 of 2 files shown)


robnagler:
Would be nice if the spawner could control the error message.

minrk:
> Would be nice if the spawner could control the error message.

Great idea! I'll look into it.

minrk:
This now includes an `Exception.jupyterhub_message` attribute if defined, in between "Spawn failed..." and "Would you like to retry", e.g. raising
 

```python
class CustomFailure(Exception):
    jupyterhub_message = "Because reasons."
```

gives

Screen Shot 2019-04-05 at 10 25 53

How does that sound? Further customization can be done via custom templates. I'm not sure about allowing the Spawner to *replace* the existing messages, but adding to them makes sense. The custom message is plaintext-only, no HTML or formatting is allowed.

robnagler:
Thank you so much @minrk!

minrk:
Thanks for the feedback!

PR #2497, opened 2019-04-01
[travis] trade mysql-connector for mysql-connector-python

minrk:
mysql-connector is deprecated and seems to have pulled old releases


minrk:
Merging to get CI back to green.

PR #2499, opened 2019-04-02
Remove password from connection string before logging

ryogesh:
Postgresql database password is shown in the log file. 

Added code to strip the password  from the connection_string db_url; the new string is used to log the messages

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1410,7 +1410,13 @@ class JupyterHub(Application):
     def init_db(self):
         """Create the database connection"""
 
-        self.log.debug("Connecting to db: %s", self.db_url)
+        #Do not log the entire DB connection string, strip the password
+        try:
+            db_info = ':'.join(self.db_url.split(':')[0:2]) + '@' + self.db_url.split('@')[1]
+        except IndexError:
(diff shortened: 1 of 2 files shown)


jupyterhub/app.py
```diff
@@ -1410,7 +1410,13 @@ def init_internal_ssl(self):
     def init_db(self):
         """Create the database connection"""
 
-        self.log.debug("Connecting to db: %s", self.db_url)
+        #Do not log the entire DB connection string, strip the password
+        try:
+            db_info = ':'.join(self.db_url.split(':')[0:2]) + '@' + self.db_url.split('@')[1]
```
minrk: Can you try using urlparse to make it more clear that we are redacting the password?

Something like:

```python
urlinfo = urlparse(self.db_url)
if urlinfo.password:
    # avoid logging the database password
    urlinfo = urlinfo._replace(netloc='{}:[redacted]@{}:{}'.format(urlinfo.username, urlinfo.hostname, urlinfo.port))
    db_log_url = urlinfo.geturl()
else:
    db_log_url = self.db_url
```

minrk:
closing in favor of #2511

Issue #2523, opened 2019-04-15
Database password redaction breaks in some circumstances.


athornton:
Hi! Thanks for using JupyterHub.

If you are reporting an issue with JupyterHub, please use the [GitHub issue](https://github.com/jupyterhub/jupyterhub/issues) search feature to check if your issue has been asked already. If it has, please add your comments to the existing issue.

**Describe the bug**
The change introduced to redact a database password (jupyterhub/app.py:1413) breaks under some circumstances.

**To Reproduce**
Create a database user and password containing a square bracket.

urlinfo = urlparse(self.db_url) breaks with "Illegal IPv6 address"

If you percent-encode the password, then it isn't recognized by the database, presumably because you're telling it the password contains "%5D" instead of "]".

**Expected behavior**

Database access to continue to work

**Screenshots**
n/a

**Desktop (please complete the following information):**
n/a

**Additional context**
n/a

**Describe the bug**
The change introduced to redact a database password (jupyterhub/app.py:1413) breaks under some circumstances.

**To Reproduce**
Create a database user and password containing a square bracket.

urlinfo = urlparse(self.db_url) breaks with "Illegal IPv6 address"

If you percent-encode the password, then it isn't recognized by the database, presumably because you're telling it the password contains "%5D" instead of "]".

**Expected behavior**

Database access to continue to work

**Screenshots**
n/a

**Desktop (please complete the following information):**
n/a

**Additional context**
I think this could be fixed with a try/except around the urlparse, and (sigh) do by-hand parsing if it fails.

minrk:
Arg, so frustrating. I guess we need to do the url parsing by hand, at least a little bit.

minrk:
When I test with percent-encoded password, it works. I believe this is the right thing to do. I'll update CI to make sure this works, but can you double-check that the url-encoding is exactly right? I did it with `urllib.quote(password, safe='')`.

athornton:
Sorry, I've been on the road.  I'll get to it this week.

minrk:
@athornton have you been able to test? I’d like to release 1.0 tomorrow.

athornton:
I have not; go ahead and release and I'll get back to it later.

PR #2524, opened 2019-04-16
test postgres with url-encoded password

minrk:
to ensure this is handled correctly

seems to work (#2523)


minrk:
@athornton this tests postgres with url-encoded password. Is there perhaps something not quite right in your URL in #2523?

willingc:
@athornton Do you want to try to test this PR, if you have time, or would you prefer @minrk and I merge?

athornton:
Just merge it, please.  If it's not working on my end, it could be a broken MySQL setup, or a bad design decision to store the whole connect string.  In any event I've worked around it for now and if it comes back to haunt me I'll give it a shot then.

minrk:
@athornton thanks!

PR #2534, opened 2019-04-22
Refactor Logout Handler

distortedsignal:
AS A developer of a Logout handler
I WANT to be able to call a function to kill spawners and
do other backend logout stuff and a separate function to
forward the user along the logout chain.

I believe this PR adds (moderately private) methods to the
Logout Handler to do just that.

### DIFF
--- a/jupyterhub/handlers/login.py
+++ b/jupyterhub/handlers/login.py
@@ -81,8 +81,6 @@ class LogoutHandler(BaseHandler):
         """Log the user out, call the custom action, forward the user
             to the logout page
         """
-        # TODO: when these can all be done concurrently, do all of them
-        #       concurrently?
         await self.default_handle_logout()
         await self.handle_logout()
         await self.render_logout_page()
--- a/jupyterhub/handlers/login.py
+++ b/jupyterhub/handlers/login.py
@@ -19,6 +19,10 @@ class LogoutHandler(BaseHandler):
         return self.settings.get('shutdown_on_logout', False)
 
     async def _shutdown_servers(self, user):
+        """Shutdown servers for logout
+
+        Get all active servers for the provided user, stop them.
+        """
         active_servers = [
             name
             for (name, spawner) in user.spawners.items()
@@ -32,11 +36,21 @@ class LogoutHandler(BaseHandler):
             await asyncio.gather(*futures)
 
     def _backend_logout_cleanup(self, name):
+        """Default backend logout actions
+
+        Send a log message, clear some cookies, increment the logout counter.
+        """
         self.log.info("User logged out: %s", name)
         self.clear_login_cookie()
         self.statsd.incr('logout')
 
-    async def _shutdown_spawners_and_backend_cleanup(self):
+    async def default_handle_logout(self):
+        """The default logout action
+
+        Optionally cleans up servers, clears cookies, increments logout counter
+        Cleaning up servers can be prevented by setting shutdown_on_logout to
+        False.
+        """
         user = self.current_user
         if user:
             if self.shutdown_on_logout:
@@ -44,14 +58,34 @@ class LogoutHandler(BaseHandler):
 
             self._backend_logout_cleanup(user.name)
 
-    async def get(self):
-        await self._shutdown_spawners_and_backend_cleanup()
+    async def handle_logout(self):
+        """Custom user action during logout
 
+        By default a no-op, this function should be overridden in subclasses
+        to have JupyterHub take a custom action on logout.
+        """
+        return
+
+    async def render_logout_page(self):
+        """Render the logout page, if any
+
+        Override this function to set a custom logout page.
+        """
         if self.authenticator.auto_login:
             html = self.render_template('logout.html')
             self.finish(html)
         else:
             self.redirect(self.settings['login_url'], permanent=False)
+
+    async def get(self):
+        """Log the user out, call the custom action, forward the user
+            to the logout page
+        """
+        # TODO: when these can all be done concurrently, do all of them
+        #       concurrently?
+        await self.default_handle_logout()
+        await self.handle_logout()
+        await self.render_logout_page()
 
 
 class LoginHandler(BaseHandler):
--- a/jupyterhub/handlers/login.py
+++ b/jupyterhub/handlers/login.py
@@ -18,27 +18,35 @@ class LogoutHandler(BaseHandler):
     def shutdown_on_logout(self):
         return self.settings.get('shutdown_on_logout', False)
 
-    async def get(self):
+    async def _shutdown_servers(self, user):
+        active_servers = [
+            name
+            for (name, spawner) in user.spawners.items()
+            if spawner.active and not spawner.pending
+        ]
+        if active_servers:
+            self.log.info("Shutting down %s's servers", user.name)
+            futures = []
+            for server_name in active_servers:
+                futures.append(maybe_future(self.stop_single_user(user, server_name)))
+            await asyncio.gather(*futures)
+
+    def _backend_logout_cleanup(self, name):
+        self.log.info("User logged out: %s", name)
+        self.clear_login_cookie()
+        self.statsd.incr('logout')
+
+    async def _shutdown_spawners_and_backend_cleanup(self):
         user = self.current_user
         if user:
             if self.shutdown_on_logout:
-                active_servers = [
-                    name
-                    for (name, spawner) in user.spawners.items()
-                    if spawner.active and not spawner.pending
-                ]
-                if active_servers:
-                    self.log.info("Shutting down %s's servers", user.name)
-                    futures = []
-                    for server_name in active_servers:
-                        futures.append(
-                            maybe_future(self.stop_single_user(user, server_name))
-                        )
-                    await asyncio.gather(*futures)
+                await self._shutdown_servers(user)
 
-            self.log.info("User logged out: %s", user.name)
-            self.clear_login_cookie()
-            self.statsd.incr('logout')
+            self._backend_logout_cleanup(user.name)
+
+    async def get(self):
+        await self._shutdown_spawners_and_backend_cleanup()
+
         if self.authenticator.auto_login:
             html = self.render_template('logout.html')
             self.finish(html)


distortedsignal:
As this is my first PR to JupyterHub (and the Jupyter Project in general), please let me know of any requirements that are missing from the above request, and I would be happy to help provide them.

distortedsignal:
I'm not sure why the tests failed - shouldn't this more-or-less take the same codepath as the other python 3.6 branch? Why would using MySQL as the db backend change the results from a Postgres database?

consideRatio:
@distortedsignal cool thanks for your work! It is a bit unknown domain for me to review so I won't jump into it right now, but I restarted the job just in case it flaked out for some reason.

distortedsignal:
Thanks @consideRatio! I've got a J.Hub project in the works, and I think that this could help me out immensely. I'll monitor the job closely to respect your time. :)

consideRatio:
Oh gosh I couldn't stop myself and had a look as it was only one file modified, I better get some sleep soon though ;D

Is it correct that this change makes the Logout action shut down servers, while they previously wasn't? The use of the word refactored made me think this was only a developer convenience, but as this would affect the user experience I have some concrete ideas!

As this behavior change would break how the users expect things to work, I think we must do something more. Making the software cause users to be surprised can be quite a big thing. So, to me, it would make sense that **a) this behavior would be configurable**, and **b) that it would default to the current behavior**: to not shut down the user servers when we logout.

I hope @minrk could provide some directions, but I also know he isn't working full time on this at the moment.

@distortedsignal if you want to learn more about how things are configured in the Jupyter world, this presentation is a great introduction - _traitlets_ are central: https://www.youtube.com/watch?v=_gYEVTaNuKU

distortedsignal:
Good questions!

I think that the git differ is not doing me a whole lot of favors here, and that the changeset is getting a little mangled. If I could explain the changes in this PR in English, I would say that this adds and removes no functionality. If aggressive function inlining is turned on, this should compile to the same bytecode as the previous code. Previously, the code (more or less) looked like this:
```py
async def get(self):
    if self.current_user:
        if self.shutdown_on_logout:
            # Do a bunch of shutdown stuff
        # Do a bunch of user logout stuff

    # Do a bunch of user forwarding stuff (in the original code, this starts with if self.authenticator.auto_login: )
```
My changes alter the flow to look like this:
```py
async def get(self):
    await self._shutdown_spawners_and_backend_cleanup()

    # Do a bunch of user forwarding stuff
```
And then, inside of the `_shutdown_spawners_and_backend_cleanup` function, there is code that is equivalent to shutdown stuff and logout stuff from the previous version.

Does that make sense?

distortedsignal:
Mostly, I'm making this PR so that I can put the `_shutdown_spawners_and_backend_cleanup()` will be in the Logout Handler, so that I can call it in the future. I think that function is a _huge_ win.

consideRatio:
Ahhhh I missed the conditional, I was indeed reading in the diff ^^ Thanks for helöing me get it right!

I also notice there was such "shutdown servers on logout" option already :D

Regarding the names of the functions btw, what does it really mean to shutdown a spawner though? Im thinking of the name: _shutdown_spawners_and_backend_cleanup

The backens cleanup function seem to entail only clearing a cookie atm, and the shutdown spawners part seem to be the conditional shutdown of servers plus the backend cleanup.

Perhaps the full _shutdown_spawners_and_backend_cleanup should be named something related to logout as it contains the conditional with the shutdown-servers-on-logout setting still.

The backend cleanup is pretty much log stuff and clear cookies right, perhaps it doesnt even need to be its on function?

Judgement calls, never easy! what do you think?

betatim:
I think splitting things up like this isn't controversial. Can we come up with a few use-cases for this to help with naming the various methods that are introduced?

Right now there are two things I think we should ponder:
* the naming of the methods is IMHO a bit awkward. Maybe via some actual use-cases we can find better ones.
* should all the methods be async or should we keep the "backend cleanup" one not async? I don't know a good way of deciding this.

distortedsignal:
Lots of scatterbrained content ahead. 

Oh, the names are 100% awkward and clunky. I’m fully aware of my limitations in that realm. I’m willing to bend to the will of the council on that. 

Use case: a logout where the admin wants to send the users to a “app selection screen” after deauthentication. Would “deauthenticate” or “deauthorize” be a good fit for the _shutdown_spawners_and_backend_cleanup function? Another use case would be if someone wanted to change the authentication procedure, they could override the deauth function to allow a user to remain logged in after they click logout. I’m not sure why someone would want that, but maybe?

I think we should keep the three functions in _backend_logout_cleanup (clear cookies, increment logout counter, write to log) in its own function. You more-or-less want those done atomically, from what I understand. Also, yeowch, that needs a better name.

As for the synchronicity (thanks autocorrect!) of the functions, I’m not sure I have an opinion. I’m _super new_ to async python, so I’ll be ok with whatever you tell me to do. 

Thanks for the help!

minrk:
Thanks for the PR, this is interesting! I'm thinking a bit about the implementation, though. This creates some private methods, which by definition shouldn't be called by any third-party code. However, it sounds like they are being created with the specific goal of calling them in a different way from third-party code. Since they are private, it would be totally okay to remove, rename, reorganize these private methods in between patch releases, which would break any downstream packages that call them. If they should be callable by subclasses, they should be public methods. In thinking about what the public pieces are and what the private pieces are, we have:

1. do what jupyterhub would have done already
2. do something extra
3. render a different page

It seems that we probably don't want to facilitate *not* doing what jupyterhub does by default, but we do want to facilitate doing something extra and custom page rendering. That would suggest doing something like:

```python
async def _default_handle_logout(self):
    """The default logout action

    Cleans up servers, clears cookies, etc.
    Private; not to be called or modified by subclasses.
    """
    # shutdown, clear cookies go here

async def handle_logout(self):
    """override in subclass to add extra logout actions"""
    return

async def render_logout_page(self):
    """render a logout page, if any

    override for custom logout page rendering
    """
    if self.authenticator.auto_login:
	    html = self.render_template('logout.html')
	    self.finish(html)
    else:
            self.redirect(self.settings['login_url'], permanent=False)

async def get(self):
    """Should not generally be overridden by subclasses

    override handle_logout and/or render_logout_page instead
    """
    await self._default_handle_logout()
    await self.handle_logout()
    await self.render_logout_page()
```

In this scenario, *adding* actions can be done by defining `handle_logout` and page rendering can be changed by overriding render_logout_page. Additionally, `await super().get()` could be used to add actions to the beginning/end of this process. The 'default' action is a private method, indicating that `get()` should *not* be overridden unless it calls `super()`. I think this is probably what we want.

How does that sound?

Ultimately, maybe what's really desired here is a logout hook, e.g. `Authenticator.logout_hook` in configuration that is called during the handler, rather than requiring a custom Handler?

distortedsignal:
@minrk I agree with 90% of what you said. However, I don't think we ever want to be calling `await super().get()` because it could redirect, and getting back from a redirect is challenging (I ran into this a bit ago when trying to write my own logout handler). I think having people write their own custom Handlers is fine, but the barrier to entry is high. This might give some people pause, but I think the way you propose both makes sense and is clear. I'll change the PR.

distortedsignal:
How do we test these changes? Since I'm introducing new public methods, from what I understand of good software dev practice, I need to write some test cases. I can't find the `LogoutHandler` in any `test_{something_here}` files - where are these tests?

jupyterhub/handlers/login.py
```diff
@@ -18,33 +18,75 @@ class LogoutHandler(BaseHandler):
...
             html = self.render_template('logout.html')
             self.finish(html)
         else:
             self.redirect(self.settings['login_url'], permanent=False)
 
+    async def get(self):
+        """Log the user out, call the custom action, forward the user
+            to the logout page
+        """
+        # TODO: when these can all be done concurrently, do all of them
```
minrk: Let's remove this todo. There's no safe assumption that they can be done concurrently, so let's preserve their order.

minrk:
Logout is tested [in test_pages](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/tests/test_pages.py#L544). A test case for a custom LogoutHandler is probably tricky, since it needs to modify the Authenticator at Hub creation time, which happens per-module, not per-test. If you'd like to try to write a test for that, that would be fine, but it would be a bit of work to do the test scaffolding. That can be another PR.

The new methods are indeed [tested](https://codecov.io/gh/jupyterhub/jupyterhub/compare/7742bfdda5f466ed8f291c49bcb774a2bfad9c1e...77d4c1f23dfab070f5a78cd56a0525f4e8ba4609/diff) (see the codecov/patch link in the "All check have passed" above), so I think this is okay to merge. Thanks!

distortedsignal:
@minrk the Travis check failed after you removed a comment - is it reasonable to rely on that check?

consideRatio:
I restarted the job @distortedsignal

minrk:
Failure's not related to this PR, it's an updated test dependency. I'll work on that separately. Thanks!

distortedsignal:
Big thanks guys. It’s been an honor. Let me know if there’s more auth stuff you’d like someone to look at.

consideRatio:
Wieeeee merge! Thanks for your work on this @distortedsignal, I think I'll have use of this myself quite soon ;D

nsshah1288:
Hi @minrk @consideRatio , I am able to override the logout page nicely by overriding 
**async def render_logout_page(self)** function, as you all suggested. 

In jupyterhub/handlers/login.py, I added the following: 

```
class OverrideLogoutHandler(LogoutHandler):
    async def render_logout_page(self):
        """Render the logout page, if any
        Override this function to set a custom logout page.
        """
        html = self.render_template('custom_logout_page.html')
        self.finish(html)
```

I was wondering, do you know of a better way of doing this without touching the jupyterhub source code? Perhaps something like is explained here: https://jupyter-notebook.readthedocs.io/en/stable/extending/handlers.html

I am using zero-to-jupyterhub Helm setup hosted on AWS EKS. Thanks!

distortedsignal:
@nsshah1288 I'm not an expert here in any sense (you tagged the right people!), but you might be able to change the LogoutHandler class by changing the Authenticator you're using. I wrote the [JupyterHub-SAMLAuthenticator](https://github.com/hewlettpackard/jupyterhub-samlauthenticator), and changed the LogoutHandler by changing the class returned in the Authenticator's `get_handlers` function.

So if you don't want to hack in the actual J.Hub code, you can set that up in the config file the same way that you change the Authenticator.

So let's say you're using the DummyAuthenticator. You would do that by doing something like the following (I believe):
```py
from jupyterhub.auth import DummyAuthenticator
from jupyterhub.handlers.login import LoginHandler

class OverrideLogoutHandler(LogoutHandler):
    # Logic goes here
    pass

class Nsshah1288Authenticator(DummyAuthenticator):
    def get_handlers(self, app):
        return [
            ('/login', LoginHandler),
            ('/logout', OverrideLogoutHandler)
            ('/hub/login', LoginHandler),
            ('/hub/logout', OverrideLogoutHandler)
        ]
```

If either Erik or Min would like to critique this approach, I am all ears.

distortedsignal:
This is by no means guaranteed to be valid Python code (can Python classes have numbers in the class name? Who knows?), but it might get you started on a path that might be helpful.

PR #2542, opened 2019-05-02
Instrument JupyterHub to record events with jupyter_telemetry

yuvipanda:
This instruments JupyterHub to record events about user
server starts & stops using the [jupyter_telemetry](https://github.com/jupyter/telemetry)
library.

https://github.com/jupyter/telemetry/blob/master/design.md
has a bunch of useful information here on what we are
doing and why.

You can test this out to see events in stderr with the
following config:

```python
import logging
c.EventLog.handlers_list = lambda: [logging.StreamHandler()]
c.EventLog.allowed_schemas = [
    'hub.jupyter.org/server-action'
]
```

This produces events like:

```
{"__timestamp__": "2019-07-09T23:49:50.449746Z", "__schema__": "hub.jupyter.org/server-action", "__version__": 1, "action": "start", "username": "yuvipanda", "servername": ""}
```


yuvipanda:
If I add this to my `jupyterhub_config.py`:

```python
def make_eventlog_sinks(eventlog):
    return [logging.FileHandler('events.log')]

c.EventLog.handlers_maker = make_eventlog_sinks
```

I see the following events in `events.log`

```
{"timestamp": "2019-05-02T23:09:51.090341Z", "schema": "hub.jupyter.org/server-action", "version": 1, "action": "start", "username": "abcd", "servername": ""}
{"timestamp": "2019-05-02T23:09:55.450531Z", "schema": "hub.jupyter.org/server-action", "version": 1, "action": "stop", "username": "abcd", "servername": ""}
```

yuvipanda:
We already support metrics via prometheus, so what's the difference between those and events? There's a lot online about this, but I like this simple analogy [from Katy Farmer](https://thenewstack.io/what-is-the-difference-between-metrics-and-events/):

> I want to keep track of my piggy bank closely. Right now, there’s only one metric I care about: total funds. Anyone can put money into my piggy bank, so I want to report the total funds at a one-minute interval. This means that every minute, my database will receive a data point with the timestamp and the amount of total funds in my piggy bank.
> Now, I want to track specific events for my piggy bank: deposits and withdrawals. When a deposit occurs, my database will receive a data point with the “deposit” tag, the timestamp and the amount of the deposit. Similarly, when a withdrawal occurs, my database will receive a data point with the “withdrawal” tag, the timestamp and the amount of the withdrawal.
> Imagine now that this is the same basic idea behind online banking. We could add more metadata to add detail to the events, like attaching a user ID to a deposit or withdrawal.

So while prometheus metrics let us answer questions like 'how many users have been active in the last 12h on average?', events let us answer questions like 'which users have been active in the last 12h?'. This is a lot more powerful, but also a lot more work to store and process. It also raises powerful questions around privacy and data collection we need to carefully consider. Most people also do not need this, but the ones who do need it badly.

yuvipanda:
Now, the admin has to explicitly whitelist which schemas they want information recorded about, like:

```
c.EventLog.allowed_schemas = [
    'hub.jupyter.org/server-action'
]
```

or no events will be recorded.

minrk:
Great to see this started! Want to make a sketch of what sort of events we should emit?

jupyterhub/events.py
```diff
@@ -0,0 +1,133 @@
...
+        """
+    )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.log = logging.getLogger(__name__)
+        # We don't want events to show up in the default logs
+        self.log.propagate = False
+        self.log.setLevel(logging.INFO)
```
minrk: I'm curious. Does log-level have any meaning for the event logger? Or is this just to make the default calls to `.emit()` pass through?

jupyterhub/handlers/base.py
```diff
@@ -994,6 +1003,11 @@ def abort():
                 SERVER_STOP_DURATION_SECONDS.labels(
                     status=ServerStopStatus.success
                 ).observe(toc - tic)
+                self.event_log.emit('hub.jupyter.org/server-action', 1, {
+                        'action': 'stop',
+                        'username': user.name,
+                        'servername': server_name
```
minrk: Should these include durations like the prometheus metrics?

minrk: Is it worth including the requesting user in these events (not always the same as the server owner)?

minrk:
Great to see this started! Want to make a sketch of what sort of events we should emit? Should we start with a check on every prometheus metric and see which of those make sense?

yuvipanda:
re: Adding more fields, I think we should think about the different use cases people have for the data that is collected here.

1. Usage / Analysis. You wanna know how much people are using your hub. This is useful for capacity planning, and cost analysis. I *think* for most people this is the most common use case.
2. Audit. You wanna know *who* is doing *what*, for auditing purposes. This operates as a security boundary almost - you wanna make sure that when your admin users access someone else's server, that is recorded.
3. Operational. This gives you information at a level that's more granular than metrics. Useful for technical staff operating the service. Server start / stop times fall under this, I think.

There's probably more categories, but this is a broad start.

We should collect the least amount of info possible, and have admins do extra work to opt in explicitly to whatever data they need to store for specific purposes. However, we should also make our dev's lives easier, and not require explicit code checks wherever the events are being emitted.

So I think the way to do this is:

1. Annotate the JSON Schema with information on what data category a field is
2. Require admins to explicitly state which categories of data they wanna record data for which schemas. 
3. Have code in EventLog that filters data in or out based on this info.

This shifts the burden of what data to keep onto the admins of the hub instance, which is where it belongs IMO. This helps us default to least amount of data stored without having to do a lot of work in the dev side.

What do you think?

Zsailer:
This is awesome—great work @yuvipanda! 

How about making one of the "data sinks" a "telemetry JupyterHub service"? Essentially, abstract out the storage piece of the telemetry system even further. The event logger inside JupyterHub would route all event data it emits to this service's endpoint. This would enable people to write custom event storing backends (e.g. remote backends) and easily hook them up to JupyterHub's telemetry system. It would also allow *other* services to talk to the same event logging backend as JupyterHub. 

What do you think?

yuvipanda:
@Zsailer absolutely! 

We already do this on mybinder.org. If you see in https://github.com/jupyterhub/mybinder.org-deploy/blob/master/mybinder/values.yaml#L82, we send the data to Google Stackdriver, which is the remote log storage system we use on mybinder.org. We then run this code as a separate service (https://github.com/jupyterhub/mybinder.org-deploy/blob/master/images/analytics-publisher/archiver.py) to get this data, scrub it and publish it to http://archive.analytics.mybinder.org. 

So if your organization already has a way to collect events / analytics, you can send them there directly. Example sinks that already exist are [Kafka](https://pypi.org/project/python-kafka-logger/), or [AWS Cloudwatch](https://pypi.org/project/watchtower/), or an [ELK Stack](https://pypi.org/project/python-logstash/), you can send these events there directly. Of course, if you don't have any such infrastructure, you can log to file or something like rsyslog easily too. Since we depend on Python's logging infrastructure, we get all these 'for free' - often with heavy performance improvements around async that we don't have to worry about :)

If I understand you correctly, in addition to this, we should also have a standard JupyterHub service that notebooks servers (and other sources) can directly write to in an *authenticated* manner. This allows centralized configuration and routing. Is that right? I think it's a brilliant idea, and we should definitely have that. Ideally, we should try and use a event / log router that already exists so we don't have to write this - but it is pretty simple to write at low enough volumes to not be a problem I think.

Zsailer:
> standard JupyterHub service that notebooks servers (and other sources) can directly write to in an authenticated manner. 

Yes! 

Although I now realize I was blending the discrete pieces in my head (a consequence of my newness to this stuff ;)). There is actually two separate pieces that I was not separating in my head. 

Let me see if I can say these back to you... :) 

1. There's the piece that "emits" events to a remote backend. In your implementation here, that's the eventlog handler. It decides what to do with the events that it receives. In this PR's example, it writes to file. In mybinder.org, you're using google cloud's `CloudLoggingHandler` to send those events to a cloud backend. If someone wanted to write a custom Handler to route events to their own remote service, they can use `logging.handlers.HTTPHandler`. 
    >Ideally, we should try and use a event / log router that already exists so we don't have to write this - but it is pretty simple to write at low enough volumes to not be a problem I think.

    Here, you're saying that hopefully someone has already written a more robust HTTPHandler that's optimized to handle a bunch of logging requests?

2. The second piece is a JHub service that exposes an endpoint to the Hub. Any *authenticated* service can POST its events to this endpoint and have them added to JupyterHub's event log. (The single-user notebook server is one example?) This service is essentially a portal through which other services talk to JHub's event log. It authenticates the service with the Hub and grants access to the event log.

yuvipanda:
@Zsailer yep, you got it right :) One clarification though.

>> Ideally, we should try and use a event / log router that already exists so we don't have to write this - but it is pretty simple to write at low enough volumes to not be a problem I think.
> Here, you're saying that hopefully someone has already written a more robust HTTPHandler that's optimized to handle a bunch of logging requests?

Here, I was talking about your point (2) above - a JupyterHub service that provides an endpoint that other services / notebooks can send telemetry to via HTTP. This would be a HTTP service that does the following:

1. Receive a POST from notebook server / other services over HTTP
2. Authenticate the request so we  know it is a legit request
3. Validate the event maybe? Notebook Servers are untrusted, so we'd definitely at least wanna make sure the 'username' fields are populated by this service than notebooks. We might also wanna validate it against the schema.
4. Route it to its destination (file, HTTP, Cloud, etc).

The reason I was hoping this would be an external service and not our python code is because there are lots of battle tested log router servers - like [fluent-bit](https://fluentbit.io) or [logstash](https://www.elastic.co/products/logstash). These can handle many hundreds of thousands of messages without issues, so we could avoid writing our own.

However, now that I've actually written it out, it sounds way too complicated for very little benefit! So let's totally ignore the 'already exists' part of my previous reply.

Thank you for helping me think this through, @Zsailer!

Zsailer:
@yuvipanda great stuff! 

Just for fun, I think I'm going to work on a prototype of the "telemetry service" we've talked about. I'll try to get it working against this branch. :)

yuvipanda:
@Zsailer awesome! Please do. I think we should eventually split the Eventlogging code into a common library that can be used in the different packages

meeseeksmachine:

https://discourse.jupyter.org/t/jupyter-telemetry-system/1217/1

yuvipanda:
Capsule fields that are added automatically for all events now are formatted as `__<name>__`, so they don't clash with properties defined in the event schema. It also lets us add more capsule items in the future without clashing with event schema properties.

jupyterhub/events.py
yuvipanda: It's to make the default calls to .emit pass through.

jupyterhub/app.py
```diff
@@ -2123,6 +2127,12 @@ def _log_cls(name, cls):
         _log_cls("Authenticator", self.authenticator_class)
         _log_cls("Spawner", self.spawner_class)
 
+        self.event_log = EventLog(parent=self)
+
+        for schema_file in glob(os.path.join(here, 'event-schemas','*.json')):
```
Zsailer: Should the schema source/definitions be a configurable trait?

jupyterhub/events.py
```diff
@@ -0,0 +1,131 @@
...
+    """
+    del record['message']
+    return json.dumps(record, **kwargs)
+
+
+class EventLog(Configurable):
+    """
+    Send structured events to a logging sink
+    """
+    handlers_maker = Callable(
```
Zsailer: See my comments [here](https://github.com/jupyterlab/jupyterlab-telemetry/pull/1/files#r294020974). 

Wondering if we can make change this trait to be a list of "event sinks". As it stands, I think this trait puts too much work on the shoulders of admins. Maybe we can create a standard "EventSink" interface. Admins can use this interface to hook their event sink service into Jupyterhub's event-logger by creating an EventSink for their service. Then, they just list it in the `event_sinks` trait.

yuvipanda: Yeah, I agree that a callable isn't a viable solution. It's here as a hack copied from binderhub due to a limitation of traitlest - you can only set traitlets to values that are pickleable! This means most objects containing a Logger instance can't be set to traitlet values in any form. We should find a way to work around that - requiring admins to pass in a callable isn't an interface we should ship with.

What would our own custom EventSink class provide that the python Logger class does not? If we have our own layer, someone needs to write a wrapper (and documentation) for every logging service out there. This is already written for the standard python logging module, so unless we have very strong reasons we should just stick to that IMO...

jupyterhub/app.py
yuvipanda: Great question!

The various extensions that can emit events from within JupyterHub (which is really what is hit by this code path), are:

1. JupyterHub Spawners
2. JupyterHub Authenticators
3. JupyterHub Proxies

All three will probably have their own schemas. They will have access to the same `event_log` object created here, and can register schemas based on where they have them stored. 

Is there a use case for registering schemas that no code is going to emit events with? If not, I think it's ok to leave that be not a traitlet. We can always expose that later if we have a use case.

jupyterhub/handlers/base.py
yuvipanda: Maybe! I want to keep the first version of these as simple as possible, though - since the focus is on the eventlogging infrastructure itself.. We can easily add more later on. Is that ok?

yuvipanda:
Picking this back up, the following things have happened since:

1. @Zsailer has cherry-picked the parts of the EventLog class here to the
   [jupyter_telemetry](https://github.com/jupyter/telemetry) library
2. We've iterated on the code there, and made a quick release (as
   jupyter_telemetry==0.0.1)
3. There's a parallel PR going on in https://github.com/jupyter/notebook/pull/4752
   for the notebook package
4. I've expanded the documentation on the schema itself, and moved its location
   in accordance with the [draft naming convention recommendations](https://github.com/jupyter/telemetry/blob/master/design.md#schema-naming-recommendations)

So my goal for this PR is to get v1 of this schema in. We can add more fields and what not later as needed easily. 

Would you have time to look through this again, @minrk?

minrk:
I'm happy with this as it is, I'd just like to add tests and docs, then 👍 

- [ ] start a place in the docs to list telemetry events (maybe auto-generate a list from schemata? That doesn't need to happen just yet, but it's probably what we want in the long run)
- [ ] test event emission and validate with schema

jupyterhub/event-schemas/server-actions/v1.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+  JupyterHub can perform various actions on user servers via 
+  direct interaction from users, or via the API. This event is 
+  recorded whenever either of those happen.
+
+  Limitations:
+
+  1. This does not record all server starts / stops, only those
+     explicitly performed by JupyterHub. For example, a user's server
+     can go down because the node it was running on dies. That will
```
willingc:
```suggestion
     can stop when its host node dies. In this situation, a stop event will not be recorded since the stop was caused by the node's failure and not an action initiated by JupyterHub. Since nodes do die in practice, a complete record of all starts and stops is not captured.
```

jupyterhub/event-schemas/server-actions/v1.yaml
```diff
@@ -0,0 +1,59 @@
...
+  action:
+    enum:
+    - start
+    - stop
+    description: |
+      Action performed by JupyterHub.
+
+      This is a required field.
+
+      Possibl Values:
```
willingc:
```suggestion
      Possible Values:
```

jupyterhub/event-schemas/server-actions/v1.yaml
```diff
@@ -0,0 +1,59 @@
...
+      
+      2. stop
+         A user's server was successfully stopped
+  username:
+    type: string
+    description: |
+      Name of the user whose server this action was performed on.
+
+      This is the normalized name used by JupyterHub itself,
+      which is derived from the authentication provider used but
```
willingc:
```suggestion
      which is derived from the authenticator and its supported authentication provider. The name may not 
```

jupyterhub/event-schemas/server-actions/v1.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+      This is the normalized name used by JupyterHub itself,
+      which is derived from the authentication provider used but 
+      might not be the same as used in the authentication provider.
+  servername:
+    type: string
+    description: |
+      Name of the server this action was performed on.
+
+      JupyterHub supports each user having multiple servers with
```
willingc: While JupyterHub supports multiple named servers for a user, this field specifies the name of the server where the action was performed.

jupyterhub/event-schemas/server-actions/v1.yaml
```diff
@@ -0,0 +1,59 @@
+"$id": hub.jupyter.org/server-action
+version: 1
+title: JupyterHub server events
+description: |
+  Record actions on user servers made via JupyterHub.
+
+  JupyterHub can perform various actions on user servers via
```
willingc:
```suggestion
  JupyterHub can perform actions on user servers. These actions may be caused by a user's interaction or an API request. The event is recorded for user interactions and API requests. 
```

Zsailer:
Thanks for the review, @minrk and @willingc! 

I'll pick this PR up and start adding docs+tests over the next couple of days. I just got back from holiday, so I'm catching up on emails+threads. I should update this PR early next week and ping you again.

Zsailer:
This was merged in #2698! 🎉

PR #2548, opened 2019-05-06
Adding the use case of the Elucidata: How Jupyter Notebook is used in…

IamViditAgarwal:
Hi.
I have added the use case of the Jupyter Notebook deployment so that it can also be featured on the Gallery Page of Jupyterhub

Jupyter Notebooks are used heavily and quite frequent in Elucidata and in the Jupyterhub is playing a very crucial role. So as a developer in Elucidata, I am raising this PR, so that my company can also be featured on the Gallery Page


IamViditAgarwal:
@betatim  When the changes will be reflected ?

betatim:
They are online now https://jupyterhub.readthedocs.io/en/latest/gallery-jhub-deployments.html

Issue #2284, opened 2018-11-03
Authorization form must be sent from authorization page


elgalu:
Hi! after upgrading to master branch (around 40 days of new commits) admins are no longer able to impersonate users, we get this error right after clicking "Authorize" on the new page:

https://github.com/jupyterhub/jupyterhub/blob/5b4f0d4304b1ea57b4d398609f2777bd93131ef4/jupyterhub/apihandlers/auth.py#L255

```
403 : Forbidden
Authorization form must be sent from authorization page
```

We are still using GitHubOAuthenticator and Postgres as backend.

Also other strange thing is happening, the admin page is not properly listing all users, but this is probably related to some of our custom page.html templates:

![image](https://user-images.githubusercontent.com/111569/47953972-e1b1c080-df84-11e8-8933-77b528c07f71.png)

elgalu:
**Update** I disabled all custom templates and the listing users issue persists.

Also, when clicking "Add Users" the new user doesn't appear. It will only appear if the user himself logs into the Hub independently of the admin.

Is this an intended changed behavior?

elgalu:
Found the problem, is related to being using http and https in different parts of the config settings

elgalu:
It seems I should switch to `c.JupyterHub.internal_ssl = True` in order to keep `http_method` consistency with my external TLS termination.

hnykda:
Interesting. We should add it to https://github.com/jupyterhub/jupyterhub/blob/master/docs/source/getting-started/security-basics.rst#if-ssl-termination-happens-outside-of-the-hub, no?

minrk:
I think this means something is wrong in the detection of what URL is being used to make the request. How is SSL termination being implemented? Since all of these requests are made by the client with HTTPS, it shouldn't be relevant whether internal SSL is used. That would only come up if the proxy-protocol is not properly forwarding that the request is made via HTTPS.

Is this perhaps on Amazon? I believe folks have seen incorrect proxy headers coming from Amazon load-balancers. I think we need to provide the ability to override public URLs in cases where the proxy-protocol doesn't properly preserve the request protocol.

hnykda:
We have seen this on GCP as well.

minrk:
thanks for the pointer. It would be helpful if someone had a reproducible test case where this occurs (I think it's somewhere in the proxy protocol reporting http when the client request is actually https). Dumping all of the headers for the rejected request would be useful, I think.

hnykda:
[...]

PR #2551, opened 2019-05-07
[WIP] add internal_ssl = True in case of SSL termination

hnykda:
see #2284

Still want to test it on 0.9-dev.


minrk:
See comment in #2284. This shouldn't be required, and is caused by a bug somewhere else in determining the protocol being used to make the request. While this is a workaround in some cases, it shouldn't be required.

hnykda:
Yes, it shouldn't and it breaks everything (hubs are failing because they try to use SSL for connection internally), it doesn't work

Issue #2552, opened 2019-05-07
Getting auth_state to the spawner


rcthomas:
I think this may be similar to the issue that led to the async `get_options_form`.  Before the home page with named servers is rendered, I want to get some information from an external service so I can customize the home page.  For instance, I want staff to be able to see options related to a testbed system but normal users shouldn't see it, since they can't access it.

I've looked at sticking this on `auth_state` and that seems the right place to put it because it's to do with the user.  But, unless I'm mistaken, to get to it later I need to call `get_auth_state` and that's a coroutine.  I could actually do it some other way, make the call from the post-auth hook, but it all boils down (I think) to needing to await something and get it (or something derived from it) into the template.

The home page `get` is already async, I was wondering if putting something there to possibly await `get_auth_state` and allow it to be used during rendering might be possible.  What I was looking for first was some way to get that information to the spawner in some non-blocking way, but I don't see it.

vilhelmen:
Yeah, I think your best bet is to use `post_auth_hook` to fill out `auth_state` data and then use the callable form of `options_form` and use `self.user.get_auth_state()` to get the auth state. You may need an await in there, I'm extremely bad at async python.

rkdarst:
About this issue and #2555, it seems that the root issue is needing to run async stuff before you render the options form.  I also dynamically adjust available profiles in one of my hubs using kubespawner: I have a profile_list which is a callable, and gets run with the spawner as an argument.  The spawner has the user attribute there.  I do this backwards from you: for each profile, I generate the list of users who have access to it and then comparing it with users is simple, but it could be reversed: you get permissions for each user and do what you do.  I don't do it async (but I should...), but from what I can tell, a callable options_form can run async.  I also do refreshing: after a certain delay, I will refresh all the profiles and their allowed users.

So, to summarize, could you put everything you need into a callable async options_form on the spawner?

I have a feeling you have dug into it so much that you probably would have thought of it if it worked, but just thought to mention it here...

rcthomas:
The problem is `options_form` rendering comes too late in the process.  What I need is to customize the home/console page based on the contents of `auth_state`.  This maybe doesn't make sense unless you are doing named servers, which we use to organize how users can spawn on different systems.

It's true I did have the problem of trying to find some async function I could slot the call into, but right before `render_template` was where I needed it.  It could be there's a more universally proper place to put it, something that gets called and could be used to refresh the auth info automatically.

vilhelmen:
[...]

PR #2555, opened 2019-05-08
Add Spawner.auth_state_hook

rcthomas:
Here's an example of what I'm thinking of as a solution for #2552.  Another idea I had was a custom handler, but in #1854 it was agreed not to override existing handlers.  Is there a better way to accomplish this, or, is there a reason it would be bad to directly pass this info to a spawner?

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -628,6 +628,10 @@ class Spawner(LoggingConfigurable):
         """
     ).tag(config=True)
 
+    def handle_auth_state(self, auth_state):
+        """Pass optional auth_state to Spawner."""
+        pass
+
     def load_state(self, state):
         """Restore state of spawner from database.
 
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -628,9 +628,23 @@ class Spawner(LoggingConfigurable):
         """
     ).tag(config=True)
 
-    def handle_auth_state(self, auth_state):
-        """Pass optional auth_state to Spawner."""
-        pass
+    auth_state_hook = Any(
+        help="""
+        An optional hook function that you can implement to pass `auth_state`
+        to the spawner after it has been initialized but before it starts.
+        The `auth_state` dictionary may be set by the `.authenticate()`
+        method of the authenticator.  This hook enables you to pass some
+        or all of that information to your spawner.
+
+        Example::
+
+            def userdata_hook(spawner, auth_state):
+                spawner.userdata = auth_state["userdata"]
+
+            c.Spawner.auth_state_hook = userdata_hook
+
+        """
+    ).tag(config=True)
 
     def load_state(self, state):
         """Restore state of spawner from database.
@@ -957,6 +971,14 @@ class Spawner(LoggingConfigurable):
                 return self.post_stop_hook(self)
             except Exception:
                 self.log.exception("post_stop_hook failed with exception: %s", self)
+
+    def run_auth_state_hook(self, auth_state):
+        """Run the auth_state_hook if defined"""
+        if self.auth_state_hook is not None:
+            try:
+                return self.auth_state_hook(self, auth_state)
+            except Exception:
+                self.log.exception("auth_stop_hook failed with exception: %s", self)
 
     @property
     def _progress_url(self):
--- a/jupyterhub/handlers/pages.py
+++ b/jupyterhub/handlers/pages.py
@@ -68,7 +68,7 @@ class HomeHandler(BaseHandler):
             url = url_path_join(self.hub.base_url, 'spawn', user.escaped_name)
 
         auth_state = await user.get_auth_state()
-        user.spawner.handle_auth_state(auth_state)
+        user.spawner.run_auth_state_hook(auth_state)
 
         html = self.render_template(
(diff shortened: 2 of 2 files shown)


jupyterhub/spawner.py
```diff
@@ -627,6 +627,10 @@ def my_hook(spawner):
         """
     ).tag(config=True)
 
+    def handle_auth_state(self, auth_state):
+        """Pass optional auth_state to Spawner."""
+        pass
```
minrk: Should there perhaps be a configurable auth_state_hook so this can be used via config without subclassing?

minrk:
My only concern with this is the overlap with pre_spawn_hook, which accomplishes much the same thing since it is called at the same time with the same info available, except that `get_auth_state` must be called in the hook. What do you think?

rcthomas:
I can look again but I think `run_pre_spawn_hook` runs too late for what I'm trying to make happen.  There are things I want to happen before the notebook is spawned that depend on `auth_state` and the spawner seems like the right place to put those.  Maybe there's a better place?

vilhelmen:
My dream that I'm slowly forcing^H^H^H^H^H^H^H suggesting to @minrk is to get `auth_state` everywhere, acting like a user profile for all sorts of runtime configuration, but what we've got now works pretty well. The user object is generally reachable when needed, and a surprising number of configuration settings support callables. Calling `get_auth_state()` in whatever spawner functions may really be best.

Worst case, you could always subclass your spawner, override init, and after calling super call get_auth_state and save the dict to self, unless there's some async funkiness involved.

rcthomas:
Two things in my mind led to this PR.  First I was looking for a way to stick some extra (not even secret) stuff onto `user` to carry around. Specifically, "this user can do this" or "this user can charge to this time allocation."  I overlooked `auth_state` for a while because the name and the encryption made me think it wasn't the appropriate place for that, but looking back at old issues convinced me it was OK.  So that solved that problem, but it made me think the `user` concept needs to be extendable. I see some other issues/comments on roles and things that indicate others have thought about this.

The second thing that led to this implementation was that I needed to `await` that `get_auth_state()` somewhere.  (I looked around at a lot of places to inject it, I guess that is "async funkiness" as you say.)  It might be better to do this in the base handler instead of in just the one method I personally happened to need it in (which was the `get()` for rendering the home page)?

What I've done here isn't that good even because some stuff in `auth_state` could change while the user is off getting coffee, but I think the auth can be refreshed.  That's already a shortcoming of this PR.

vilhelmen:
I'm all for flexibility in the system, I've been wanting to rename `auth_state` to `profile` to make it a little less spooky sounding and get it more actively used as a general "stick user data here for everything else to reference" blob.

You should, in theory, be able to immediately await that `get_auth_state()` call since the `option_form()` callable is awaited in the spawner, but I tend to shy away from the async bits because I end up needing Min to bail me out in my PRs. Multithreaded C? Sure, any time. Async python? Uggghhh.

I've been curious about the auth state refresh system (I believe the function is called something like that), and that may be a way to go, but I've never used it. Not having read the code, it should work just fine with named servers in JH 1.0. I'm not sure if your spawner tracks total session runtime, but the auth state refresh function should be called on a timer roughly often enough to use it as a manual timekeeper.

I've been meaning to propose some user onboarding handler, basically a `new_user_add` callback to act like a non-auth `auth_state` initialization system, but I don't think that'd really help in this case.

jupyterhub/spawner.py
```diff
@@ -953,6 +971,14 @@ def run_post_stop_hook(self):
             except Exception:
                 self.log.exception("post_stop_hook failed with exception: %s", self)
 
+    def run_auth_state_hook(self, auth_state):
+        """Run the auth_state_hook if defined"""
+        if self.auth_state_hook is not None:
+            try:
+                return self.auth_state_hook(self, auth_state)
+            except Exception:
+                self.log.exception("auth_stop_hook failed with exception: %s", self)
```
willingc:
```suggestion
                self.log.exception("auth_state_hook failed with exception: %s", self)
```

jupyterhub/spawner.py
```diff
@@ -627,6 +627,24 @@ def my_hook(spawner):
         """
     ).tag(config=True)
 
+    auth_state_hook = Any(
+        help="""
+        An optional hook function that you can implement to pass `auth_state`
+        to the spawner after it has been initialized but before it starts.
+        The `auth_state` dictionary may be set by the `.authenticate()`
+        method of the authenticator.  This hook enables you to pass some
+        or all of that information to your spawner.
```
willingc:
```suggestion
        or all of the `auth_state` information to your spawner.
```

rcthomas:
I have been running this in production for a while here and it seems to be OK, but I would be happy to learn that there's a better place (or more places) where the hook should be called.

minrk:
Thanks for your patience on this one @rcthomas and thanks @willingc for the review!

hvpareja:
Hi there,

sorry @rcthomas for reopening this threat, but I think the documentation is not helping me to leverage this new feature.

I am very interested to use the auht_state_hook, but I fail to understand where exactly needs to be placed. Could you extend the same example you wrote in the documentation, please?

Thanks

PR #2560, opened 2019-05-13
Jupyterhub: use previous exit strategy for Windows

adelcast:
Windows doesn't have support for signal handling so it can't use the
signal handling capabilities of asyncio. Use the previous atexit
strategy on the Windows case instead.

Signed-off-by: Alejandro Del Castillo <alejandro.delcastillo@ni.com>

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -2420,28 +2420,54 @@ class JupyterHub(Application):
             pc.start()
 
         self.log.info("JupyterHub is now running at %s", self.proxy.public_url)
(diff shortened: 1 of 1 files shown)


adelcast:
@minrk does this look like a sane strategy? or are there changes that need to be addressed?

minrk:
This does seem reasonable. I don't like to have too many divergences for Windows since they are all untested and unsupported, but if this is needed, I'm okay.

adelcast:
Thanks @minrk! Without it, JupyterHub would just not start on Windows, so even using a container spawner approach would fail....with this I can migrate to 1.0.

PR #2564, opened 2019-05-14
typo

julienchastang:
small typo


consideRatio:
Thanks :D I appreciate these small fixes in the same way as it feels good to really polish something.

:heart:

Issue #2566, opened 2019-05-16
Visual UX: Make the announcement text a banner


consideRatio:
The announcement would preferably be formatted as a banner, just like it is on Discourse and GitLab for example.

### GitLab announcement banner
![image](https://user-images.githubusercontent.com/3837114/57852037-6017d680-77e2-11e9-8ba2-efd0b6b61f17.png)

### Discourse announcement banner
![image](https://user-images.githubusercontent.com/3837114/57852049-673ee480-77e2-11e9-8540-89ae7da78787.png)

### Current JupyterHub announcement
![jupyterhub_announcement](https://user-images.githubusercontent.com/3837114/57852011-4d050680-77e2-11e9-81f0-b153b43d265e.gif)

consideRatio:
Hmmm...

Well, we can configure the announcement as HTML as well though, so we have more options not wrapping this to a banner, but at the same time more work. Hmmm...

https://github.com/jupyterhub/jupyterhub/blob/8a6de3006c51854b539eaf5c1c61179217f06996/share/jupyterhub/templates/page.html#L149-L155

Anyhow, I experimented by adding css classes `alert` and `alert-warning` as described in the [twitter-bootstrap docs](https://getbootstrap.com/docs/4.0/components/alerts/).

![jupyterhub_announcement_with_alert](https://user-images.githubusercontent.com/3837114/57860258-dae9ed00-77f4-11e9-9ec9-216b8b860d89.gif)

rcthomas:
We definitely had feedback on our deployment from users that the ordinary announcement text is too unobtrusive.  In response, we did exactly what you did in your follow-up, @consideRatio, and applied a bootstrap css class to the announcement.

(Also in our case however, we still got the feedback that people are expecting those kinds of messages through our normal center-wide communications channels outside JupyterHub: our MOTD and Live Status pages for instance).

That said:
* If we could get the styling to match JupyterHub's color scheme that might be enough.
* Managing the announcements by editing templates or another service is OK, but something in the admin interface to update/change the announcement would be neat too (I know that's separate from this issue).

prateek2408:
> Hmmm...
> 
> Well, we can configure the announcement as HTML as well though, so we have more options not wrapping this to a banner, but at the same time more work. Hmmm...
> 
> https://github.com/jupyterhub/jupyterhub/blob/8a6de3006c51854b539eaf5c1c61179217f06996/share/jupyterhub/templates/page.html#L149-L155
> 
[...]

PR #2570, opened 2019-05-16
Make announcements visible without custom HTML

consideRatio:
Fixes https://github.com/jupyterhub/jupyterhub/issues/2566 to some
degree by making the announcement stand out using twitter-bootstrap
classes `alert` and `alert-warning`. Perhaps we could theme twitter
bootstrap or this alert specifically with jupyter related colors as well
though?

See the issue for details on how the fix looks visually.


minrk:
Makes sense!

Issue #2575, opened 2019-05-21
Stop My Server button in control panel should disappear after server stopped.


aar0nTw:
After we press `Stop My Server`, the button doesn't have any obvious update since user press it. I think if `Stop My Server` can be removed after server stopped will be better.

**To Reproduce**
1. Go to control panel
2. Click `Stop My Server`
3. Server stopping and `My Server` blue button will disabled.
4. Server stopped, `My Server` will became `Start my server`, but `Stop My Server` still there, make user confused.

**Expected behavior**
`Stop My Server` button should disappear after server stopped.

**Screenshots**
![Screen Shot 2019-05-21 at 10 29 57 AM](https://user-images.githubusercontent.com/935988/58065534-a5f4e600-7bb8-11e9-934a-c298909a48e2.png)

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Chrome
 - Version 75.0.3770.38 (Official Build) beta (64-bit)

PR #2577, opened 2019-05-22
Hide Stop My Server red button after server stopped.

aar0nTw:
Patch the issue #2575, hide stop sever button after server stopped for better ui response.


minrk:
Thanks!

PR #2586, opened 2019-05-24
Wrap body parameter type in schema to conform to OpenAPI spec

sbrunk:
I ran into to an issue while trying to generate a REST client from the rest-api.yml (twilio/guardrail#309).

Seems like the [OpenAPI spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameterIn) mandates a `schema` field for body parameters:

If [`in`](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameterIn) is `"body"`:

Field Name | Type | Description
---|:---:|---
<a name="parameterSchema"></a>schema | [Schema Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md##schemaObject) | **Required.** The schema defining the type used for the body parameter.

In a few cases the body parameter types are directly specified in `parameters`. This PR simply moves them under a `schema` field.

### DIFF
--- a/docs/rest-api.yml
+++ b/docs/rest-api.yml
@@ -7,7 +7,7 @@ info:
   license:
     name: BSD-3-Clause
 schemes:
-  - [http, https]
+  [http, https]
 securityDefinitions:
   token:
     type: apiKey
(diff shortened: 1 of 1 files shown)


minrk:
merged as #2590

PR #2588, opened 2019-05-27
Define default values for HubAuth ssl traitlets

cmd-ntrf:
The default values are taken from environment variables defined by `Spawner.get_env`.

## Use-case

I am using `HubAuth` with batchspawner to send the singleuser port number back to the Hub before launching singleuser server. After activating internal_ssl, the HubAuth stopped working because I was not initializing `client_ca`, `certfile` and `keyfile`. I presumed HubAuth would read the JupyterHub SSL environment variables like singleuser does, but it did not. So I added, the initialization by reading the environment variable myself, but it would be nicer if HubAuth was in charge.

The example usage of HubAuth with batchspawner is available here:
https://github.com/cmd-ntrf/batchspawner/blob/singleuser_fix/batchspawner/singleuser.py

### DIFF
--- a/jupyterhub/services/auth.py
+++ b/jupyterhub/services/auth.py
@@ -216,7 +216,7 @@ class HubAuth(SingletonConfigurable):
         return self.hub_host + url_path_join(self.hub_prefix, 'login')
 
     keyfile = Unicode(
-        '',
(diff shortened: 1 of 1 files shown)


minrk:
Makes sense, thanks!

PR #2590, opened 2019-05-28
Change API description to be a valid OpenAPI spec

sbrunk:
This change should make the JupyterHub API spec a valid [OpenAPI 2.0 spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md). You can check this for instance using https://editor.swagger.io/. Use File -> import URL to import the new [version in this PR](https://raw.githubusercontent.com/sbrunk/jupyterhub/openapi-compliance/docs/rest-api.yml) or  [current master](https://raw.githubusercontent.com/jupyterhub/jupyterhub/master/docs/rest-api.yml) for comparison to see validation results.

This should make the [generated API documentation](https://jupyterhub.readthedocs.io/en/stable/_static/rest-api/index.html) more accurate. It' also helpful for generating correct clients for different programming languages and it should make it easier to move to OpenAPI 3.0 in the future.

Changes:
* Add missing responses (doesn't include all possible responses yet)
* Refactor invalid multi in body parameters into a single parameter
* Change form type into valid formData
* Fix use of required fields
* Apply a few other minor fixes (indentation etc.)

This supersedes (and includes) #2586

### DIFF
--- a/docs/rest-api.yml
+++ b/docs/rest-api.yml
@@ -1,4 +1,4 @@
-# see me at: http://petstore.swagger.io/?url=https://raw.githubusercontent.com/jupyter/jupyterhub/master/docs/rest-api.yml#/default
+# see me at: http://petstore.swagger.io/?url=https://raw.githubusercontent.com/jupyterhub/jupyterhub/master/docs/rest-api.yml#/default
 swagger: '2.0'
 info:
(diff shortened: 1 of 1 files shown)


minrk:
Nice, thanks!

PR #2601, opened 2019-06-04
[WIP] Add support for Jupyter Server

yuvipanda:
Use case:

1. We want to run something like https://github.com/QuantStack/voila or similar on JupyterHub
2. We *don't* want the Notebook API to be available, for security (and other) reasons

This applies to any server extension we might wanna run without running the Jupyter Notebook API itself.

## Current State

This adds jupyter_server support to jupyterhub-singleuser. It'll just use the notebook package if available, falling back to jupyter_server if not. I don't like this - it should probably be configurable somewhere.

Most of this is done with dynamic importing of libraries, since the code for Jupyter Notebook and Jupyter Server is pretty much the same in all the ways the singleuser server deals with it.

## TODO

1. Activity tracking?
2. Does user configuration disabling work with ServerApp?

## Alternative approaches considered

- Put this code in the jupyter_server package instead. There is practically 100% code sharing between what NotebookApp needs and what ServerApp needs, and magically patching jupyterhub-singleuser doesn't seem ideal. 
- Use voila and others purely as notebook serverextensions. You can already do this, but this is insecure - end users can easily call the notebook API directly to execute code.
- Put this code in voila. Putting it in jupyter_server makes it more generally useful to all the apps that can use it. There's a Jupyter Enhancement Proposal for Jupyter Server (https://github.com/jupyter/enhancement-proposals/blob/master/jupyter-server/jupyter-server.md), and @Zsailer / @SylvainCorlay  are working on it.
- Split the SingleUserNotebookApp class into two levels. Put everything common to both in base class, and subclass / customize in other forms. This is promising, and something I'll explore.

### DIFF
--- a/jupyterhub/singleuser.py
+++ b/jupyterhub/singleuser.py
@@ -6,6 +6,7 @@ import asyncio
 import json
 import os
 import random
+import importlib
 from datetime import datetime
 from datetime import timezone
 from textwrap import dedent
@@ -20,10 +21,13 @@ from tornado.httpclient import HTTPRequest
 from tornado.web import HTTPError
 from tornado.web import RequestHandler
 
+use_serverapp = os.environ.get('USE_JUPYTER_SERVER', 'False') == 'True'
+
+required_package = 'jupyter_server' if use_serverapp else 'notebook'
 try:
-    import notebook
+    parent_module = importlib.import_module(required_package)
 except ImportError:
-    raise ImportError("JupyterHub single-user server requires notebook >= 4.0")
+    raise ImportError("JupyterHub single-user server requires {}".format(required_package))
 
 from traitlets import (
     Any,
@@ -38,14 +42,19 @@ from traitlets import (
     TraitError,
 )
 
-from notebook.notebookapp import (
-    NotebookApp,
-    aliases as notebook_aliases,
-    flags as notebook_flags,
+app_name = 'jupyter_server.serverapp' if use_serverapp else 'notebook.notebookapp'
+app_module = importlib.import_module(app_name)
+
+NotebookApp = getattr(app_module, 'ServerApp' if use_serverapp else 'NotebookApp')
+notebook_aliases = app_module.aliases
+notebook_flags = app_module.flags
+
+LoginHandler = getattr(importlib.import_module(required_package + '.auth.login'), 'LoginHandler')
+LogoutHandler = getattr(importlib.import_module(required_package + '.auth.logout'), 'LogoutHandler')
+IPythonHandler = getattr(
+    importlib.import_module(required_package + '.base.handlers'),
+    'JupyterHandler' if use_serverapp else 'IPythonHandler'
 )
-from notebook.auth.login import LoginHandler
-from notebook.auth.logout import LogoutHandler
-from notebook.base.handlers import IPythonHandler
 
 from ._version import __version__, _check_version
 from .log import log_request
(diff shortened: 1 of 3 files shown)


jupyterhub/singleuser.py
```diff
@@ -38,14 +48,17 @@
     TraitError,
 )
 
-from notebook.notebookapp import (
-    NotebookApp,
-    aliases as notebook_aliases,
-    flags as notebook_flags,
+app_module = importlib.import_module(app_name)
```
blink1073: There's a new recommended pattern for this (which I think is bizarre, personally): https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly

yuvipanda:
The failures are at least superficially related to https://github.com/jupyter/jupyter_server/issues/52

jupyterhub/singleuser.py
yuvipanda: @blink1073 that's only for importing source files right? We're importing a module directly here...

Zsailer:
I've opened a [PR in notebook](https://github.com/jupyter/notebook/pull/4653) that refactors the Notebook API into a Jupyter Server Extension. 

If that gets merged, we wouldn't have to mess with dynamic importing. Rather, the `SingleUserNotebookApp` could become a `SingleUserServer` and append the Notebook API like any other extension. 

Do you think it's worth waiting for that change to happen (assuming it gets accepted)?

jupyterhub/singleuser.py
blink1073: Ah, I see what you mean, thanks!

jupyterhub/spawner.py
```diff
@@ -931,7 +931,7 @@ def get_args(self):
             args.append('--notebook-dir=%s' % _quote_safe(notebook_dir))
         if self.default_url:
             default_url = self.format_string(self.default_url)
-            args.append('--NotebookApp.default_url=%s' % _quote_safe(default_url))
+            args.append('--SingleUserNotebookApp.default_url=%s' % _quote_safe(default_url))
```
minrk: 👍

jupyterhub/singleuser.py
```diff
@@ -22,8 +23,17 @@
 
 try:
     import notebook
+    use_serverapp = False
+    server_package = 'notebook'
+    app_name = 'notebook.notebookapp'
 except ImportError:
-    raise ImportError("JupyterHub single-user server requires notebook >= 4.0")
+    try:
+        import jupyter_server
```
minrk: Can we make this explicit and based on an env variable (e.g. JUPYTERHUB_SERVER_PACKAGE) rather than try/except? Unconditional try/except makes it impossible to pick when both packages are present.

jupyterhub/singleuser.py
```diff
@@ -22,8 +23,17 @@
 
 try:
     import notebook
+    use_serverapp = False
+    server_package = 'notebook'
+    app_name = 'notebook.notebookapp'
```
minrk: Maybe use fully qualified name for the App class `notebook.notebookapp.NotebookApp` rather than having the module name and class name logic in two different places. Then we can import with:

```python
from traitlets.utils.importstring import import_item
app_name = 'notebook.notebookapp.NotebookApp'
...
ServerApp = import_item(app_name)
```

minrk:
This is great! I am a little concerned about how this is effectively using JupyterHub to make commitments in JupyterServer about API-compatibility with nbclassic (i.e. relative import locations of LoginHandler, `init_` methods for subclassing, etc.). It might be useful to jot down which APIs we use here to see how tightly we are coupling these two things.

PR #2605, opened 2019-06-07
add activity_resolution config

minrk:
limits last_activity update interval to 30 seconds by default to avoid a db commit on every authenticated request, which can have serious performance cost


meeseeksmachine:

https://discourse.jupyter.org/t/identifying-jupyterhub-api-performance-bottleneck/1289/2

Issue #2616, opened 2019-06-20
Unable to handle usernames containing spaces 


druths:
**Describe the bug**
When a new user is being created and the username has a space in it (as will happen with the AzureAd authentication service), the useradd command fails - users can't have spaces in them (or at least need to be handled specially).

**To Reproduce**
1. Setup AzureAd authenticator
2. Login as a new user who has a space in their name
3. Boom

**Expected behavior**
The system should replace the space with some non-whitespace character (e.g., a '.')

katsar0v:
The usernames should be normalized and the authenticator should usually do this. In docker for example you cannot have container with spaces in the name. So I'd rather look in the authenticator, because using system accounts with system processes (which might have spaces in them) works as wished.

druths:
Makes sense to me.  I'll take this issue there.

Derek Ruths
Associate Professor // School of Computer Science // McGill
Chief Architect // Charitable Analytics International
Chief Data Scientist // Ruths.ai
www.derekruths.com
@derekruths

------ Original Message ------

PR #2617, opened 2019-06-20
Fixed issue #2616 - accounts with a whitespace are adjusted

druths:


minrk:
Thanks for the PR! As mentioned in #2616, this should be the responsibility of the Authenticator to normalize names. Coercing at this level, only coming *out* of the database, allows for potentially undesirable collisions and inconsisntence and problems elsewhere (e.g. name in != name out).

willingc:
Thanks for the PR @druths. The change would be better to be addressed in the Authenticator base classes (https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/auth.py).

minrk:
closing as inactive. Feel free to come back to this if there's something to do in jupyterhub instead of in the respective Authenticators.

PR #2625, opened 2019-07-02
Add missing words

remram44:
This page feels a bit like a duplicate from https://jupyterhub.readthedocs.io/en/latest/reference/services.html. I copied the missing words from there.

I reflowed to 72 chars seems that seem to be the format of the file.


betatim:
Thanks for taking the time to make the documentation better!

PR #2626, opened 2019-07-03
Update doc: do not suggest depricated config key

lumbric:
According to changelog JupyterHub.bind_url has been added in 0.9.0.


minrk:
Thanks!

PR #2628, opened 2019-07-04
Update the config used for testing

jtpio:
It looks like it should be `SimpleLocalProcessSpawner` instead of `SimpleSpawner`.


yuvipanda:
Thanks, @jtpio

Issue #2618, opened 2019-06-20
running_servers prometheus metric has negative values


yuvipanda:
Metric for 'running_servers' is incremented on server start and decremented on server stop. However, since servers may already be running when the hub starts - and this isn't initially set for the metric - we end up with negative numbers! The general count is also not accurate at all.

PR #2629, opened 2019-07-04
Fix running_servers prometheus metric

GeorgianaElena:
*RUNNING_SERVERS* is now
* initialized with the number of active servers
* decremented when a user's server is stopped
* incremented after the user's server has been successfully added to the proxy

Closes https://github.com/jupyterhub/jupyterhub/issues/2618

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -79,6 +79,7 @@ from .utils import (
     print_ps_info,
     make_ssl_context,
 )
+from .metrics import RUNNING_SERVERS
 
 # classes for config
 from .auth import Authenticator, PAMAuthenticator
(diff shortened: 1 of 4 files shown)


yuvipanda:
w00t, thank you for working on this!

A question - does the count get set only when a user's server spawns? Does it get set at startup as well?

GeorgianaElena:
Unfortunately it gets set on user server spawn. Do you think setting it in ```JupyterHub.init_users``` it's more appropriate?

yuvipanda:
@GeorgianaElena I think setting it in init_users in addition to where it's set now is a good idea

GeorgianaElena:
Thanks, @yuvipanda. I ended up setting it in init_spawners (in init_users, it made ```assert self.tornado_settings```(from ```_users_default```) to fail).
Quick question: I think *total_users* metrics might have the same problem as running_servers and might need to be set at startup with the number of users in the db. Do you think it is the case?

yuvipanda:
@GeorgianaElena yeah, I think doing that for total_users is also a good idea.

I'm going to merge this now! <3 thank you!

PR #2631, opened 2019-07-05
Fixed docs and testing code to use refactored SimpleLocalProcessSpawner

danlester:
The docs at [Setting up a development install](https://jupyterhub.readthedocs.io/en/stable/contributing/setup.html) suggest installing the separate jupyterhub-simplespawner package for easy testing, but this now appears to have been refactored into the main jupyterhub repo (and the SimpleSpawner class also renamed to SimpleLocalProcessSpawner).

This PR fixes the documentation and also the sample testing/jupyterhub_config.py file so that it refers to the correct class.

### DIFF
--- a/docs/source/contributing/setup.rst
+++ b/docs/source/contributing/setup.rst
@@ -112,13 +112,12 @@ happen.
 
 Happy developing!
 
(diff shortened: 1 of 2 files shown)


minrk:
great, thanks!

PR #2632, opened 2019-07-05
corrected docker network create instructions in dockerfiles README

bartolone:
Hi jupyterhub folks!

While browsing documentation, I noticed that dockerfiles/README.md includes an incorrect step for testing:
put both containers on the same network (e.g. docker create network jupyterhub; docker run ... --net jupyterhub)

... should be 'docker network create' rather than 'docker create network'. Cutting and pasting the incorrect version leads to an unhelpful error message, so I thought this correction might be helpful to relative newbies to Docker like myself.

I think this follows your contribution guidelines since it isn't part of your official documentation but rather a file that resides on github, but happy to resubmit if there's another preferred way I should do that.

Thanks for your time!


minrk:
Thanks!

PR #2637, opened 2019-07-10
Fix total_users prometheus metric

GeorgianaElena:
We might get more accurate results if we initialize *total_users* metric at startup with the number of users stored in the db.


minrk:
Yes indeed!

Issue #2128, opened 2018-09-04
Cannot stop my server when a backslash is in username


Jamesits:
**Describe the bug**
Cannot stop my server when a backslash character is in username.

**To Reproduce**
1. Login with a user named `foo\bar`
2. start server
3. stop my server

**Expected behavior**
The server should stop.

**Screenshots**
![image](https://user-images.githubusercontent.com/1297257/45033727-e4576e80-b044-11e8-8fba-85cf66e39064.png)

**Server**
 - OS: Fedora 28
 - JupyterHub: 0.9.2

**Desktop (please complete the following information):**
 - OS: Windows 10 1803
 - Browser: Firefox
 - Version 61.0.2

nicorikken:
I can confirm this. We use the genericOAuthenticator, resulting in usernames like `Company\UserId`. This is causing ever more issues.
As the backslash is not URL-encoded, it is replaced with a forward slash by the browser, resulting in an incorrect path and thus the error.
URL-encoding the backslash with `%5c` and repeating the call does succeed.

nicorikken:
I think URL-encoding the `user` variable in the javascript code will solve this issue https://github.com/jupyterhub/jupyterhub/blob/38cf95523f2ec7ce553136b630c915b0395cbca2/share/jupyterhub/static/js/home.js#L61

nicorikken:
In the templated html and the javascript the `user` is not properly URL-encoded anywhere, also causing issues when spawning a new container, so this issue is much wider than just stopping the server.

nicorikken:
@Jamesits I proposed some UI changes to fix this. Perhaps you can test this for your use-case? https://github.com/jupyterhub/jupyterhub/pull/2640

nicorikken:
I misinterpreted the issue. In the UI `window.jhdata.user` the user is known without the backslash, e.g. `companyuserid`. The backend is serving pages which included the backslash, so there is a mismatch between the two.
The javascript injected into the page does contain the backslash:
```
      window.jhdata = {
        base_url: "/hub/",
        prefix: "/",
        user: "company\userid",
        admin_access: false,
        options_form: true,
      }
```
So now it seems to me that the username should be properly escaped in this javascript insert.

nicorikken:
[...]

PR #2640, opened 2019-07-12
Escape usernames in the frontend

nicorikken:
To cope with special characters like backslash, to address issue
https://github.com/jupyterhub/jupyterhub/issues/2128

### DIFF
--- a/jupyterhub/user.py
+++ b/jupyterhub/user.py
@@ -351,6 +351,11 @@ class User:
         return quote(self.name, safe='@~')
 
     @property
+    def unicode_escaped_name(self):
+        """My name, escaped for use in javascript inserts, etc."""
+        return self.name.decode('unicode_escape')
+
+    @property
     def proxy_spec(self):
         """The proxy routespec for my default server"""
         if self.settings.get('subdomain_host'):
--- a/jupyterhub/user.py
+++ b/jupyterhub/user.py
@@ -1,5 +1,6 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
+import codecs
 import warnings
 from collections import defaultdict
 from datetime import datetime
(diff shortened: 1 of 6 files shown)


nicorikken:
**I haven't check its workings yet, this is based on assumptions. Please test.**

nicorikken:
I've come to the realization that in regards to spawning, this issue is probably deeper, as the username is not properly escaped in the python handlers and the HTML templating.

nicorikken:
I was mistaken. `window.jhdata.user` doesn't include the backlash. So this makes me question where the backslash error is happening.

minrk:
I think your final code is right, that this should be using encodeURIComponent instead of escape. Can you rebase to remove the Python-side changes that are now unused, and just the js fixes? Then I'm happy to merge.

jupyterhub/user.py
```diff
@@ -350,6 +351,11 @@ def escaped_name(self):
         """My name, escaped for use in URLs, cookies, etc."""
         return quote(self.name, safe='@~')
 
+    @property
+    def unicode_escaped_name(self):
+        """My name, escaped for use in javascript inserts, etc."""
```
willingc:
```suggestion
        """The user name, escaped for use in javascript inserts, etc."""
```

share/jupyterhub/static/js/home.js
```diff
@@ -103,7 +103,7 @@ require(["jquery", "moment", "jhapi", "utils"], function(
   $(".new-server-btn").click(function() {
     var row = getRow($(this));
     var serverName = row.find(".new-server-name").val();
-    window.location.href = "../spawn/" + user + "/" + serverName;
+    window.location.href = "../spawn/" + escape(user) + "/" + serverName;
```
willingc: @minrk did you want this to be encodeURIComponent instead of escape?

nicorikken:
Right... that was a lot of back-and-forth. @willingc @minrk thanks for responding so quickly, I didn't notice while I was still fixing this code. Now it is neatly rounded. The main issue was that the Python string needed to be properly JSON-encoded. My previous version using unicode encoding didn't cover all cases, so I went this route.

- I chose a substring `[1:-1]` to strip the outer JSON quotes, to remain compatible with the `user.name` and `user.escaped_name` properties. The outer quotes are added in the templating.
- The parameter is renamed to `json_escaped_name` to better reflect its purpose.
- A few tests are added for edge-cases, although it seems to break the build somehow.

share/jupyterhub/static/js/home.js
nicorikken: My bad, this introduced errors, it was a bad attempt.

jupyterhub/user.py
nicorikken: Thanks, resolved.

nicorikken:
Done! Ready to merge now.
It was nice to work on JupyterHub and get to know its internals.

nicorikken:
By the way, I tested this on our own infrastructure, where the issue occurred as well. This PR solved that.

minrk:
That works, thanks!

PR #2641, opened 2019-07-13
Some suggestions from reading through the docs

ericdill:
I think this should be marked as a work in progress, but I didn't want to forget to open this PR :) I end up reading the docs on planes and then always forget to push the changes.

Some of these are more questions than notes and should ultimately be removed from this PR, but I wasn't sure what the best way to ask the questions was.


docs/source/reference/spawners.md
```diff
@@ -119,7 +119,7 @@ def clear_state(self):
 
 Some deployments may want to offer options to users to influence how their servers are started.
 This may include cluster-based deployments, where users specify what resources should be available,
-or docker-based deployments where users can select from a list of base images.
+or docker-based deployments where admins can select from a list of base images.
```
manics: The original is correct, users select from a list of base images (the list is configured by the admin)

docs/source/reference/authenticators.md
```diff
@@ -31,6 +31,7 @@ popular services:
 - Okpy
 - OpenShift
 
+Note: Where is this generic implementation available?
```
manics: https://github.com/jupyterhub/oauthenticator/blob/master/oauthenticator/generic.py

docs/source/reference/spawners.md
ericdill: ok thanks will revert

docs/source/contributing/setup.rst
```diff
@@ -75,6 +75,8 @@ happen.
 
       npm install -g configurable-http-proxy
 
+   Question: Is this supposed to say ACCESS instead of EACCES?
```
minrk: EACCES is correct. it is the name of the error code (see `errno.EACCES` in Python)

docs/source/contributing/setup.rst
```diff
@@ -75,6 +75,8 @@ happen.
...
+   Question: Can we recommend `conda install nodejs` instead of running sudo npm install?
```
minrk: There's loads of ways to do this (`nvm` is often my personal choice). We could be more specific about what and why:

. `sudo` may be required to perform a system-wide install, or nodejs can be installed in a user-writeable location with e.g. `conda install nodejs`.

docs/source/contributing/setup.rst
```diff
@@ -87,6 +89,7 @@ happen.
    The second line needs to be run every time you open a new terminal.
 
 4. Install the python packages required for JupyterHub development.
+   Question: If miniconda is the recommended way to get python should these installation instructions be updated to use ``conda`` instead of ``pip``?
```
minrk: I'd leave this as pip. Using conda to get Python doesn't preclude using pip. Conda 4.6 and later behaves pretty well with pip as a peer.

docs/source/getting-started/authenticators-users-basics.md
```diff
@@ -101,6 +101,9 @@ popular services:
 - Okpy
 - OpenShift
 
+NOTE: Open issue asking for more details on this generic implementation.
+It's not clear if this is a different implementation or if the JupyterHub OAuth
+_is_ the generic implementation.
```
minrk: It's called `GenericOAuthenticator`, which is a generic configurable OAuth implementation. JupyterHub doesn't ship with support for external OAuth providers.

docs/source/getting-started/networking-basics.md
```diff
@@ -12,7 +12,8 @@ This section will help you with basic proxy and network configuration to:
 The Proxy's main IP address setting determines where JupyterHub is available to users.
 By default, JupyterHub is configured to be available on all network interfaces
 (`''`) on port 8000. *Note*: Use of `'*'` is discouraged for IP configuration;
-instead, use of `'0.0.0.0'` is preferred.
+instead, use of `'0.0.0.0'` is preferred. (reminder to open issue asking for a
+link to this statement for futher reading if interested)
```
minrk: `*` is a Jupyter-specific alias for '' that is not helpful or necessary. This was provided as a shortcut in Jupyter notebook, so this is clarifying that the shortcut should not be used, but JupyterHub does still technically handle it.

docs/source/getting-started/networking-basics.md
```diff
@@ -41,6 +42,10 @@ port.
 
 ## Set the Proxy's REST API communication URL (optional)
 
+Note to open issue about the following paragraph. The wording is a little unclear. 
+Is there a second port that the Hub service uses? Or is the api_url the only
+available port here?
```
minrk: The Hub exclusively talks to the proxy via its REST API URL. This is the only URL by which the proxy can be configured.

docs/source/getting-started/security-basics.rst
```diff
@@ -79,6 +79,9 @@ it is reasonable to run the hub without SSL.
 To achieve this, simply omit the configuration settings
 ``c.JupyterHub.ssl_key`` and ``c.JupyterHub.ssl_cert``
 (setting them to ``None`` does not have the same effect, and is an error).
+Reminder to ask: Does "simply omit" mean "make sure they are commented in 
+the configuration file" or "make sure they are set to an empty string" or
+something else?
```
minrk: Omit means "not set", which is commented-out or removed from the file altogether (the file is Python, so comments are the same as not being there at all. This sentence is an attempt to state explicitly that setting them  to None or an empty value is not the same as not setting them.

docs/source/reference/config-sudo.md
```diff
@@ -71,6 +71,7 @@ rhea ALL=(JUPYTER_USERS) NOPASSWD:JUPYTER_CMD
 ```
 
 It might be useful to modify `secure_path` to add commands in path.
+Question: What is `secure_path` and where can one find more information about it?
```
minrk: from [sudo docs](https://www.sudo.ws/man/1.8.14/sudoers.man.html):

> secure_path: Path used for every command run from sudo. If you don't trust the people running sudo to have a sane PATH environment variable you may want to use this. Another use is if you want to have the “root path” be separate from the “user path”. Users in the group specified by the exempt_group option are not affected by secure_path. This option is not set by default.

docs/source/reference/config-sudo.md
```diff
@@ -153,6 +154,7 @@ then you will need to give `node` permission to do so:
 sudo setcap 'cap_net_bind_service=+ep' /usr/bin/node
 ```
 However, you may want to further understand the consequences of this.
+Question: Is there a reference for further reading?
```
minrk: [capabilities doc](http://man7.org/linux/man-pages/man7/capabilities.7.html)

docs/source/reference/config-user-env.md
```diff
@@ -43,6 +43,8 @@ sudo python3 -m pip install numpy
 to install the numpy package in the default system Python 3 environment
 (typically `/usr/local`).
 
+TODO: Get a link from the conda team for a description of what "appropriate permissions for users" is
+
 You may also use conda to install packages. If you do, you should make sure
 that the conda environment has appropriate permissions for users to be able to
 run Python code in the env.
```
minrk: The appropriate permissions are described in the previous paragraph:

> ... must be *readable and executable* by your users. If you want users to be able to install additional
> packages, it must also be *writable* by your users.

The appropriate permissions are not sensitive to which installation tools are used.

docs/source/reference/spawners.md
```diff
@@ -222,6 +222,9 @@ discover these resource limits and guarantees, such as for memory and CPU.
 For the limits and guarantees to be useful, **the spawner must implement
 support for them**. For example, LocalProcessSpawner, the default
 spawner, does not support limits and guarantees. One of the spawners
+
+TODO provide a link to the `systemdspawner`
```
minrk: https://github.com/jupyterhub/systemdspawner

docs/source/contributing/setup.rst
ericdill: Do you want me to mention `nvm` here? Or was that more of an FYI for the sake of this PR?

docs/source/getting-started/authenticators-users-basics.md
ericdill: Right above this list it says that JupyterHub's OAuthenticaator supports that list of services. Here you're saying that JupyterHub doesn't ship with support for external OAuth providers. Is it accurate to say that:

JupyterHub ships with a generic reference implementation, [GenericOAuthenticator](https://github.com/jupyterhub/oauthenticator/blob/master/oauthenticator/generic.py#L33) It is left up to the user to implement support for any OAuth provider that is not listed above.

ericdill:
Thanks @minrk that's really helpful. I made updates based on your review. I only had one additional clarification question that I left on the in-line conversation [here](https://github.com/jupyterhub/jupyterhub/pull/2641#discussion_r303944990).

docs/source/reference/config-sudo.md
```diff
@@ -163,7 +165,8 @@ instructions](http://ubuntuforums.org/showthread.php?t=992706).
 
 ### Shadow group (FreeBSD)
 
-**NOTE:** This has not been tested and may not work as expected.
+**NOTE:** This has not been tested on FreeBSD and may not work as expected on
```
willingc: Add:

Do not use in production without verifying that it works properly.

docs/source/reference/proxy.md
```diff
@@ -204,7 +204,7 @@ setup(
 ```
 
 If you have added this metadata to your package,
-users can select your authenticator with the configuration:
+admins can select your authenticator with the configuration:
```
willingc: Good distinction.

docs/source/reference/separate-proxy.md
```diff
@@ -14,8 +14,8 @@ because the proxy is automatically managed by the hub.  This is great
 for getting started and even most use, but everytime you restart the
 hub, all user connections also get restarted.  But it's also simple to
 run the proxy as a service separate from the hub, so that you are free
-to reconfigure the hub while only interrupting users who are currently
-actively starting the hub.
+to reconfigure the hub while only interrupting users who are in the middle of
```
willingc:
```suggestion
to reconfigure the hub while only interrupting users who are waiting for their notebook server to start.
```

docs/source/reference/services.md
```diff
@@ -201,8 +201,9 @@ below.
 The reference, or base, implementation is the [`HubAuth`][HubAuth] class,
 which implements the requests to the Hub.
 
-To use HubAuth, you must set the `.api_token`, either programmatically when constructing the class,
-or via the `JUPYTERHUB_API_TOKEN` environment variable.
+To use HubAuth, you must set the `.api_token` instance variable. This can be
+done either programmatically when constructing the class, or via the
+`JUPYTERHUB_API_TOKEN` environment variable. (TODO: Where is this API TOKen set?)
```
willingc: https://github.com/jupyterhub/jupyterhub/blob/071e375d5fd038e70c29c619d199bf2c9c090d5c/docs/source/reference/rest.md#add-api-tokens-to-the-config-file

willingc: The examples directory in the repo's root directory has configuration examples for different cases.

docs/source/reference/urls.md
```diff
@@ -123,6 +123,8 @@ Handling this URL is the most complicated condition in JupyterHub,
 because there can be many states:
 
 1. server is not active
+NOTE: These next two don't render properly in "EPUB" format when downloaded
```
willingc: I would go ahead and remove them since they are confusing as written. 

Perhaps something like: 

Handling this URL depends on two conditions: whether a requested user is found as a match; the state of the requested user's notebook server.

docs/source/getting-started/services-basics.md
```diff
@@ -67,7 +67,7 @@ c.JupyterHub.services = [
 Upon restarting JupyterHub, you should see a message like below in the
 logs:
 
-```
+```none
```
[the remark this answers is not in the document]
ericdill: My understanding is that Sphinx will default the code blocks to be Python if you dont specify anything. At least that's what is happening with the docs as rendered by readthedocs:

Screen Shot 2019-07-16 at 11 28 55 AM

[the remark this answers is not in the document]
willingc: Cool. Thanks for pointing out. One of these days I will have to dive in and see if this a recommonmark issue or a sphinx issue. :wink:

ericdill:
Thanks for all the feedback @willingc   I'll be able to respond to it tomorrow

docs/source/contributing/setup.rst
ericdill: Ah good idea. I'll move that up to the Sys Reqs section

docs/source/reference/authenticators.md
```diff
@@ -113,7 +112,8 @@ to uid to username), which is useful in case you use some external
 service that allows multiple usernames mapping to the same user (such
 as ActiveDirectory, yes, this really happens).  When
 `pam_normalize_username` is on, usernames are *not* normalized to
-lowercase.
+lowercase. NOTE: Earlier it says that usernames are normalized using PAM.
```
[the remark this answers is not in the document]
ericdill: The two sentences that make this unclear are 

> When using `PAMAuthenticator`, you can set `c.PAMAuthenticator.pam_normalize_username = True`, which will normalize usernames using PAM

and 

> When `pam_normalize_username` is on, usernames are *not* normalized to lowercase.

I think the second sentence here needs to be changed.

------

OHH NEVERMIND I GET IT NOW. 🤦‍♂ 

Changing the last sentence to the following would clarify it, at least for me. Is this a sensible change?

> When `pam_normalize_username` is on, usernames are not normalized to *lowercase* but to however they are represented in the system.

docs/source/reference/services.md
ericdill: Is there a way to crosslink to that markdown heading?

docs/source/reference/config-sudo.md
ericdill: definitely worth making that as in-your-face as possible :)

ericdill:
Thanks so much for the feedback @willingc . I've updated based on your comments, though I've got a few questions in return 😁

ericdill:
bump @willingc -- can you have a look at the questions I left here? This might be good to go as-is

cmd-ntrf:
The following note has been left in the text:
> (TODO: Where is this API TOKen set?)
https://github.com/jupyterhub/jupyterhub/blob/main/docs/source/reference/services.md?plain=1#L247

It should probably be removed.

PR #2647, opened 2019-07-16
Some theme updates; no double NEXT/PREV buttons.

Carreau:
- Install pip in the docs conda env (or conda complains).
- Do not override page.html, the next/previous buttons are now handled by
  alabaster_jupyterhub (this actually remove the duplicated next/prev
  buttons)
- use alabaster_jupyterhub when building locally, this make it easy for
  new contributor to get the _exact_ same appearance than on
  readthedocs.


docs/environment.yml
```diff
@@ -4,6 +4,7 @@ name: jhub_docs
 channels:
   - conda-forge
 dependencies:
+- pip
```
willingc: It's weird that it is not picking up pip on line 18 for installation.

Carreau: Well on line 18 it's a pip subsection, and that's because there is line 18 that it complains. 

The reason to have to include pip on it's own is that you may want to pin pip version I believe: 

```
- pip==18.3:
    - entrypoints
    ...
```

Would be weird, so you need:
```
  - pip 
  - pip:
      - entrypoints
```

Carreau:
Also thanks , those double next were driving me nuts :-) And it was good seeing you at SciPy !

PR #2648, opened 2019-07-17
fix: spawn redirect for users whose names need escaping

nicorikken:
The 302 redirect served after the spawn POST was not escaping the username, causing issues with a backslash.


nicorikken:
As far as I can tell, this solves all redirect issues with usernames. I've tried the dialogs of starting and stopping a server using a username with a backslash.

nicorikken:
The Travis CI build fails, but I don't see why.

minrk:
Great! I think the test was just a bit flaky. Should be okay to merge shortly.

PR #2650, opened 2019-07-17
[Discussion] Add options to validate HTML before sending to the page.

Carreau:
While obviously this is a lot of extra work; This is mostly to get the
conversation started; and think as whether this should be set for – for
example – testing.

That would allow to catch earlier miss-matched tags; or issues in
templates.

I'm not quite sure AdvancedHTMLParser is the best for that; but I
haven't found a better library.


minrk:
This sounds great for testing. I'm less sure about instantiating and running a validator on every page served. What do you think?

It would be interesting to see what happens if there's a validation error on the error page that would be attempted when this error is raised.

PR #2658, opened 2019-07-24
Update flask hub authentication services example in doc

cmd-ntrf:
The flask example in the documentation was still using the input argument `cookie_cache_max_age` when instantiating `HubAuth` object. `cookie_cache_max_age` is deprecated sinceJupyterHub 0.8 and should be replaced by `cache_max_age`.


betatim:
Thanks for helping to keep examples up to date!

PR #2662, opened 2019-07-26
Update spawn-form example

kinow:
Hi,

The current example does not match the screen shot. I think it was due a Bootstrap update. I used the first example from [Bootstrap Forms component](https://getbootstrap.com/docs/4.3/components/forms/) to update the example.

Before:

image

After:

image

Thanks
Bruno


GeorgianaElena:
Awesome! Thanks a lot, @kinow :confetti_ball:

Issue #2598, opened 2019-06-03
add configurable limit to the kernel runtime 


rochaporto:
We rely on jupyterhub for a few multi user notebook deployments, backed by Kubernetes and exposing different sets of resources.

In cases where the resources are limited (as is the case for GPUs), it would be great to be able to configure a max runtime for the user server so that everyone gets their share. This is similar to what kaggle offers with session limits.

We rely on the zero2jupyterhub helm chart to setup and config the server, so maybe a good option would be to add this in the profile definition.

rochaporto:
Culling seems to be doing already what was described above:
https://github.com/jupyterhub/jupyterhub/blob/master/examples/cull-idle/cull_idle_servers.py#L331

Apart from the separation between different server types: cpu only, cpu+gpu, etc. Would be nice to set different maxAge values for each of these. Is this already possible somehow? I couldn't figure this out from the code and config, maybe it needs an additional param to set the server profile(s) the cull service should aim for? Could then deploy more than one, with different configs.

vilhelmen:
I don't think there's anything in place to support this. I'll take a closer look, though. One route would be to attach custom `maxAge` and `maxIdle` to spawner objects that the culler can receive through the API, but that seems annoyingly complicated. A better method would be making the culler more configurable and allowing users to map some sort of container name to lifetimes. It all depends on what the culler can get through the API.

vilhelmen:
Yeah, the culler currently only receives the following snippets of spawner data from `/users`, it would need to be extended to transmit some sort of container id data:

```
{
    "kind": "user",
    "name": "XXX",
    "admin": false,
    "groups": [],
    "server": "/user/XXX/",
    "pending": null,
    "created": "2019-06-05T19:02:28.458702Z",
    "last_activity": "2019-06-05T19:35:56.697797Z",
    "servers": {
      "": {
        "name": "",
        "last_activity": "2019-06-05T19:35:56.633000Z",
        "started": "2019-06-05T19:02:36.859734Z",
        "pending": null,
        "ready": true,
        "state": {
          "pod_name": "jupyter-XXX"
        },
        "url": "/user/XXX/",
        "progress_url": "/hub/api/users/XXX/server/progress"
      }
    }
```

vilhelmen:
[...]

PR #2664, opened 2019-07-30
WIP: Internalize idle culler

vilhelmen:
Just a proof of concept for #2598. Probably not best to jam it into the polling startup, but it's just a start. Needs tests/docs/etc.

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -13,6 +13,7 @@ import shutil
 import signal
 import sys
 import warnings
+from datetime import datetime
 from subprocess import Popen
 from tempfile import mkdtemp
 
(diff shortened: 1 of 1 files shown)


minrk:
I don't think it should be done this way. In particular, I don't think that culling should be internal to JupyterHub, but even if it were, I think it should be done in a way with a single global callback, not one per Spawner.

Making it internal also means it cannot be customized or modified by deployments, unlike the API consumer implementation.

vilhelmen:
I was wondering about using a single function, there could be a concerning amount of overhead with one per spawner. Correctly looping through the spawner dict could be tricky, though.

Would you reconsider if it was moved to a single callback a loop and used a user-provided function (with a default provided)? Or should we close this and go back to work on modifying API data to provide similar functionality?

minrk:
I think we should stick to the API approach. Thanks for exploring this one, though!

Issue #2667, opened 2019-08-01
/hub/admin should redirect to login


clkao:
**Describe the bug**

When a user is not logged in, visiting /hub/admin page simply shows 403. we should redirect to the login page, and show 403 only if the user is indeed not admin.

**To Reproduce**
1. Go to /hub/admin without login
4. See error 403

**Expected behavior**
1. Go to /hub/admin without login
2. See login page
3. Upon login: if the user is admin, show the /hub/admin page, otherwise show 403

clkao:
Note that for external auth configured, the user creation logic should also be part of the admin login. So it still works if somehow this is the admin's first time visiting jupyterhub

PR #2670, opened 2019-08-05
/hub/admin redirect to login

GeorgianaElena:
This should close https://github.com/jupyterhub/jupyterhub/issues/2667.
 */hub/admin* without login redirects to the login page and throws a 403 if the user is not an admin.


betatim:
Looks good to me, so merged!

PR #2672, opened 2019-08-05
Replace header logo: jupyter -> jupyterhub

consideRatio:
I like the header to reflect that we are in the JupyterHub domain rather than Jupyter generic domain or for example the JupyterLab domain. This is somewhat related to #2671 that adds a JupyterHub logo to the spawning page options.

I think if we merge something like this PR, it would reduce the need for #2671.

## Current
image

## This PR
image

Perhaps @jupyterhub/designers got input on this?

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -592,7 +592,9 @@ class JupyterHub(Application):
 
     @default('logo_file')
     def _logo_file_default(self):
-        return os.path.join(self.data_files_path, 'static', 'images', 'jupyterhub-80.png')
+        return os.path.join(
+            self.data_files_path, 'static', 'images', 'jupyterhub-80.png'
+        )
 
     jinja_environment_options = Dict(
         help="Supply extra arguments that will be passed to Jinja environment."
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -592,7 +592,7 @@ class JupyterHub(Application):
 
     @default('logo_file')
     def _logo_file_default(self):
-        return os.path.join(self.data_files_path, 'static', 'images', 'jupyter.png')
+        return os.path.join(self.data_files_path, 'static', 'images', 'jupyterhub-80.png')
 
     jinja_environment_options = Dict(
         help="Supply extra arguments that will be passed to Jinja environment."
--- a/jupyterhub/tests/test_pages.py
+++ b/jupyterhub/tests/test_pages.py
@@ -611,7 +611,7 @@ async def test_static_files(app):
     r = await async_requests.get(ujoin(base_url, 'logo'))
     r.raise_for_status()
     assert r.headers['content-type'] == 'image/png'
-    r = await async_requests.get(ujoin(base_url, 'static', 'images', 'jupyter.png'))
+    r = await async_requests.get(ujoin(base_url, 'static', 'images', 'jupyterhub-80.png'))
     r.raise_for_status()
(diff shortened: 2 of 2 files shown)


betatim:
I think I prefer this over #2671

willingc:
@consideRatio I like the look of this over the Server options page in #2671. Would you mind clarifying the use cases that we are trying to solve? Thanks.

- Stock unmodified JupyterHub: Users and admins will now see JupyterHub in the top left of the UI. Server options page would not have the JupyterHub logo
- Customized JupyterHub: Users and admins would now see JupyterHub or their custom logo in the top left. Server options page would not have the JupyterHub logo.

consideRatio:
Thanks for input!

@willingc I've been confused myself about what is jupyterhub and what is the singleuser-server, these PRs were created to make us discuss what we can do about such confusions.

I like this PR far more than the other, because the other is like a splash screen for arriving users but arriving users may arrive either to login options or to the login button screen depending on how the JupyterHub is setup. I figure I'll close the other, that PR (#2671) was meant to splash the JupyterHub vs Singleuser-server distinction for users arriving after having the auto-login option configured for JupyterHub, meaning this was their first actual screen in interaction with JupyterHub as they bypassed the big login button screen.

willingc:
> I like this PR far more than the other, because the other is like a splash screen for arriving users but arriving users may arrive either to server options screen or to the login button screen depending on how the JupyterHub is setup.

I agree with you that it would be helpful to users to know where they are :wink: This PR is a reasonable approach to me.

betatim:
Travis is unhappy because of formatting of the code. Needs some more black.

I'd be happy to merge this once the bots are happy. Also pinging @minrk in case he wants to chime in with a historical view on things.

betatim:
Also, do you need to add the actual image to the PR as well? Is there a way for us to test if all assets are present?

consideRatio:
@betatim the image is a 80px height sized png image, and used in a 28px height setting. Perhaps that is fine, I'm not 100% on that though.

minrk:
Historical view was that jupyterhub was a thin wrapper around the notebook and folks deploying jupyterhub didn't want users to view it as a separate thing, but rather a lightweight launching page for the Jupyter application(s) they are familiar with. We received feedback along the lines of "I don't want my users to know that JupyterHub is involved at all, or to be confused by ask what is JupyterHub and how is it different from Jupyter?" So the design idea was that JupyterHub is something that only deployers are aware of and users never see anything related to JupyterHub as distinct from Jupyter. This was perhaps more true when JupyterHub had literally no UI beyond 2 buttons on one page: start and stop (not counting admin UI). It now has things like named server management, token management, etc., so asserting JupyterHub's identity may be more appropriate.

I definitely prefer this over #2671

betatim:
One case where it would have been useful to have a clear label like this PR proposes is when I try and describe to people that they need to go to the admin panel and that this is the hub wide admin panel, not admin/settings/something else for their particular user.

I'd be +0 on adding this to make it easier for people to describe where they are or how they can tell they are in the right place. Inn particular with written/over the phone instructions.

willingc:
@betatim It may make sense to have some "debug" (where am I) content in the footer of the page.

betatim:
@minrk from what you wrote I can't tell if you are -1, 0 or +1. I think you are not -1 so I'll merge as we can revert it again if my guess was wrong. Until stuff is released I think it is Ok to merge and revert again.

minrk:
Sorry! Yes, +1 to merge.

Issue #2668, opened 2019-08-03
Configuring tokens is quite confusing


katsar0v:
So i had following config with jupyterhub `0.9.4`, which worked pretty fine:

```
c.JupyterHub.service_tokens = {
    '675982d420fbed3bb6403ecd8032be284ace814fdab37815b7d24f741c84237d' : 'root'
}

c.Authenticator.admin_users = {'root'}

c.JupyterHub.services = [
    {
        'name': 'usage-monitor',
        'admin': True,
        'command': 'python /srv/jupyterhub/monitor_usage.py --every=900'.split()
    },
    {
        'name': 'cull-idle',
         'admin': True,
         'command': 'python /srv/jupyterhub/cull_idle_servers.py --cull_every=60 --timeout=691200'.split()
    }
]
```

Updating jupyterhub throws a 403, I get:
```
[W 2019-08-02 13:34:33.045 JupyterHub log:174] 403 GET /hub/api/users (root@172.18.0.1) 46.77ms
```
and response from api:
```
[ERROR]: There was an error: {"status": 403, "message": "Forbidden"})
```
The documentation for 1.0.0 is confusing or not up to date and it is not really clear how to use these tokens for accessing the api via scripts (curl, custom python requests etc ), here is a curl as well:
```
(base) katsarov@katsarov-ThinkPad-T470s:~$ curl http://172.18.0.4:8080/hub/api/users -H "Content-Type: application/json" -H "Authorization: token 675982d420fbed3bb6403ecd8032be284ace814fdab37815b7d24f741c84237d" -v
*   Trying 172.18.0.4...
* TCP_NODELAY set
* Connected to 172.18.0.4 (172.18.0.4) port 8080 (#0)
> GET /hub/api/users HTTP/1.1
> Host: 172.18.0.4:8080
> User-Agent: curl/7.58.0
> Accept: */*
> Content-Type: application/json
> Authorization: token 675982d420fbed3bb6403ecd8032be284ace814fdab37815b7d24f741c84237d
> 
< HTTP/1.1 403 Forbidden
< Server: TornadoServer/6.0.3
< Content-Type: application/json
...
```

I think this part of the documentation or the tokens generally (https://jupyterhub.readthedocs.io/en/stable/reference/rest.html) should be updated explaining what api tokens and service tokens are and which to use when.

At the end I got the working config for my issue:
```
c.JupyterHub.services = [
    {
        'name': 'usage-monitor',
        'admin': True,
[...]

PR #2679, opened 2019-08-10
Documentation update: hint for using services instead of service tokens.

katsar0v:
This is a small pull request, as already spoken with @minrk, it would make sense to hint users to consider using services instead of service_tokens

Issue: https://github.com/jupyterhub/jupyterhub/issues/2668

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -811,14 +811,14 @@ class JupyterHub(Application):
 
     api_tokens = Dict(
         Unicode(),
-        help="""PENDING DEPRECATION: consider using service_tokens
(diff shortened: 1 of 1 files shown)


minrk:
Great! It looks like this pulls in some old commits we shouldn’t include. Could you rebase the PR on the latest master?

katsar0v:
> Great! It looks like this pulls in some old commits we shouldn’t include. Could you rebase the PR on the latest master?

Updated it, sorry ^^

GeorgianaElena:
Thanks for working on this @katsar0v :confetti_ball:

Issue #2657, opened 2019-07-24
SpawnPendingHandler doesn't seem to be working right with named servers


rcthomas:
In the `SpawnPendingHandler` there are places where the spawn_url is set like:

https://github.com/jupyterhub/jupyterhub/blob/24c08292899492bb0bd5e2b7fdbf7e09cf1a4057/jupyterhub/handlers/pages.py#L286

We've got a pre-spawn hook that fires to make sure the user's home directory is not over quota.  If the user's over quota then it fires and we get the `not_running` template.  The user is advised via message there they should make some room and then come back and try again.  The problem is the button points to the default server "", not the named server the user was trying to start.

Is this the desired behavior or is it maybe a bug that the `spawn_url` should be set using say, `user.server_url(server_name)`?

GeorgianaElena:
Fixed in https://github.com/jupyterhub/jupyterhub/pull/2682

PR #2682, opened 2019-08-12
Add `server_name` to `spawn_url`

rcthomas:
The `server_name` is missing from `spawn_url` here in `SpawnPendingHandler` when the `not_running.html` template is rendered.  This causes the template to render the retry button with `/hub/spawn/<user>` with the default server as the target, and not the named server that the user probably intended to spawn.  The user then clicks and starts the wrong server.  In the `UserUrlHandler` this is already taken care of, I think it just didn't get propagated to the `pages` module.

Addresses #2657


GeorgianaElena:
Thanks for fixing this, @rcthomas :four_leaf_clover:

consideRatio:
@rcthomas thanks for writing an excellent summary :) I'm working on the changelog and has become happy by reading many great summaries in PR that makes it so much easier to summarize changes!

:heart:

Issue #2683, opened 2019-08-13
pre_spawn_start exceptions may lead to redirect loop


clkao:
**Describe the bug**

if we override authenticator `pre_spawn_start` with a coroutine and throw exceptions from it, it could lead to the redirect loop issue with an eventual 500 error, instead of showing the exception.  however, exceptions `options_from_form` are rendered properly as error message.

**To Reproduce**
1. override authenticator pre_spawn_start and throw exception from it
2. try to launch user server from hub
3. gets infinite redirect
4. reload and see 500 error

**Expected behavior**

errors are displayed in the spawner page

cc/@aar0nTw

aar0nTw:
#2684 fix this problem, @GeorgianaElena 👍 
cc/@clkao

PR #2684, opened 2019-08-14
Handle pre_spawn_start possible exceptions

GeorgianaElena:
This should fix https://github.com/jupyterhub/jupyterhub/issues/2683 :eyes:

### DIFF
--- a/jupyterhub/handlers/pages.py
+++ b/jupyterhub/handlers/pages.py
@@ -177,12 +177,11 @@ class SpawnHandler(BaseHandler):
             # otherwise it may cause a redirect loop
             if f.done() and f.exception():
                 exc = f.exception()
-                if exc:
-                    raise web.HTTPError(
-                        500,
-                        "Error in Authenticator.pre_spawn_start: %s %s"
-                        % (type(exc).__name__, str(exc)),
-                    )
+                raise web.HTTPError(
+                    500,
+                    "Error in Authenticator.pre_spawn_start: %s %s"
+                    % (type(exc).__name__, str(exc)),
+                )
             self.redirect(pending_url)
 
     @web.authenticated
--- a/jupyterhub/handlers/pages.py
+++ b/jupyterhub/handlers/pages.py
@@ -172,7 +172,19 @@ class SpawnHandler(BaseHandler):
                 spawner._spawn_future = None
             # not running, no form. Trigger spawn and redirect back to /user/:name
             f = asyncio.ensure_future(self.spawn_single_user(user, server_name))
-            await asyncio.wait([f], timeout=1)
+            done, pending = await asyncio.wait([f], timeout=1)
+            # If spawn_single_user throws an exception, raise a 500 error
+            # otherwise it may cause a redirect loop
+            if f in done:
+                future, = done
+                exc = future.exception()
+                if exc:
+                    raise web.HTTPError(
+                        500,
+                        "Error in Authenticator.pre_spawn_start: %s %s"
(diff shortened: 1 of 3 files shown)


jupyterhub/handlers/pages.py
```diff
@@ -172,7 +172,19 @@ def _render_form(self, for_user, spawner_options_form, message=''):
...
+            if f in done:
+                future, = done
+                exc = future.exception()
+                if exc:
+                    raise web.HTTPError(
+                        500,
+                        "Error in Authenticator.pre_spawn_start: %s %s"
+                        % (type(exc).__name__, str(exc)),
+                    )
+                    return
```
minrk: No need for return after raise
```suggestion
```

jupyterhub/handlers/pages.py
```diff
@@ -172,7 +172,19 @@ def _render_form(self, for_user, spawner_options_form, message=''):
                 spawner._spawn_future = None
             # not running, no form. Trigger spawn and redirect back to /user/:name
             f = asyncio.ensure_future(self.spawn_single_user(user, server_name))
-            await asyncio.wait([f], timeout=1)
+            done, pending = await asyncio.wait([f], timeout=1)
+            # If spawn_single_user throws an exception, raise a 500 error
+            # otherwise it may cause a redirect loop
+            if f in done:
```
minrk: We only need to take this branch if f is done and failed:

```suggestion
            if f.done() and f.exception():
```

jupyterhub/handlers/pages.py
```diff
@@ -172,7 +172,19 @@ def _render_form(self, for_user, spawner_options_form, message=''):
...
+                future, = done
+                exc = future.exception()
```
minrk: No need for separate `future` referring to the same `f`

GeorgianaElena:
Addressed the feedback! Thanks a lot @minrk.

jupyterhub/handlers/pages.py
```diff
@@ -172,7 +172,17 @@ def _render_form(self, for_user, spawner_options_form, message=''):
                 spawner._spawn_future = None
             # not running, no form. Trigger spawn and redirect back to /user/:name
             f = asyncio.ensure_future(self.spawn_single_user(user, server_name))
-            await asyncio.wait([f], timeout=1)
+            done, pending = await asyncio.wait([f], timeout=1)
+            # If spawn_single_user throws an exception, raise a 500 error
+            # otherwise it may cause a redirect loop
+            if f.done() and f.exception():
+                exc = f.exception()
+                if exc:
```
minrk: This `if exc` is no longer needed, since we know it's true from above.

GeorgianaElena: ups... sorry, I missed it :disappointed:

jupyterhub/handlers/pages.py
```diff
@@ -172,7 +172,16 @@ def _render_form(self, for_user, spawner_options_form, message=''):
                 spawner._spawn_future = None
             # not running, no form. Trigger spawn and redirect back to /user/:name
             f = asyncio.ensure_future(self.spawn_single_user(user, server_name))
-            await asyncio.wait([f], timeout=1)
+            done, pending = await asyncio.wait([f], timeout=1)
```
aar0nTw: I thought [line 216](https://github.com/jupyterhub/jupyterhub/pull/2684/files#diff-f14530788c189c34e88032abd7c584e0R216) in post handler need rewrite in same way as well.

PR #2695, opened 2019-08-21
Fix failing travis tests

GeorgianaElena:
Posgresql and MySql tests are failing on Travis :( and I believe it's related to the Xenial migration (https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment).

- [x] fix postgres test
- [x] fix mysql test

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,8 @@ env:
     - MYSQL_HOST=127.0.0.1
     - MYSQL_TCP_PORT=13306
 services:
-  - postgres
+  - postgresql
+  - mysql
   - docker
 
 # installing dependencies
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,6 @@ env:
     - MYSQL_TCP_PORT=13306
 services:
   - postgresql
-  - mysql
   - docker
 
 # installing dependencies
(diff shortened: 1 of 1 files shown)


GeorgianaElena:
While the tests pass with these changes, I cannot find a clear explanation for the mysql related one  :face_with_head_bandage:

* The posgres test makes sense to be passing now, because according to travis (https://docs.travis-ci.com/user/reference/xenial#services-disabled-by-default) with the xenial adoption all services are disabled and postgresql needs to be started explicitly ( https://docs.travis-ci.com/user/database-setup/#postgresql)
* The mysql test, I cannot explain... because it's using mysql inside docker and it's not obvious for me how the xenial migration would impact it :(. The strange thing is that, the test is failing just like before [this fix](https://github.com/jupyterhub/jupyterhub/pull/2543). Using an even older version of mysql-connector-python (i.e. 8.0.11) fixes the test, so the only reason I can think of is that the change that introduced this bug propagated to older versions (but I'm not sure).

GeorgianaElena:
Update:
* On trusty the test still passes with mysql-connector-python 8.0.15
* On xenial the test passes with mysql-connector-python 8.0.11

I don't understand what's causing this... Two possible fixes to make the tests green again would be:
* Continue to run the mysql test on trusty (``dist: trusty``)
* Pin mysql-connector-python to 8.0.11 on xenial

However, both solutions seem *hacky* :disappointed: and I'm not sure what's the better approach

manics:
Could it be a bug in MySQL server (5.6 vs 5.7)? https://docs.travis-ci.com/user/database-setup/#mysql

minrk:
mysql version could definitely be relevant.

Let's merge this as tests are passing, and we can dig in more later.

Thanks @GeorgianaElena!

Issue #2621, opened 2019-06-25
Link to hub given as HTTP and not HTTPS


TomJDavey:
Hi there,

While trying to configure JupyterHub locally, running in the terminal gives the following link:

"JupyterHub is now running at http://127.0.0.1:port/"

To which my browser responded "The connection was reset". I found that using HTTPS and ignoring the warnings solved this issue. 

My specs are: Ubuntu 19.04, Firefox and I used openssl to get a key and certificate.
JupyterHub version: 1.0.0

I believe it makes sense to change this to HTTPS for future users.

minrk:
Can you share your configuration? It's possible that we should be updating bind_url if ssl configuration is detected.

kinow:
Quick PR submitted, just used existing crt/key file from the `internal-ssl` configuration and defined `ssl_key` and `ssl_cert` in `jupyterhub_config.py`.

oguttu:
I am having the same problem on windows. How do I correct it?

Ignatiusdork:
Hi
I would like to contribute, but am new to open source, and i would like to know how to get more familiar with it . Thanks

PR #2697, opened 2019-08-22
updated bind url (public facing URL of app) to https from http

AashrayAnand:
Updated the bind URL to https from http

Original Issue: #2621


minrk:
It would not be accurate to change this to https, since the default is to not use https. This PR also has no changes in it right now, perhaps there was a mistake?

PR #2702, opened 2019-08-23
Fix mistypos

rlukin:


betatim:
Thanks a lot for fixing up tpyos!

PR #2704, opened 2019-08-23
Create a warning when creating a service implicitly from service_tokens

katsar0v:
Create a warning when creating a service implicitly from service_tokens by checking `services` list. If the service is implicitly created, warn the user.

Issue: https://github.com/jupyterhub/jupyterhub/issues/2668

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1673,6 +1673,12 @@ class JupyterHub(Application):
                     raise ValueError("Token name %r is not in whitelist" % name)
                 if not self.authenticator.validate_username(name):
                     raise ValueError("Token name %r is not valid" % name)
+            if kind == 'service':
+                if not any(service["name"] == name for service in self.services):
+                    self.log.warn(
+                        "Warning: service '%s' not in services, creating implicitly. It is recommended to register services using services list."
+                        % name
+                    )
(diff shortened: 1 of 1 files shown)


betatim:
This looks good. One more thing to do is fix the formatting of the code. This is why the tests fail https://travis-ci.org/jupyterhub/jupyterhub/builds/575995012?utm_source=github_status&utm_medium=notification

We use `black` and https://github.com/jupyterhub/jupyterhub/blob/master/CONTRIBUTING.md#setting-up-a-development-environment has instructions for how to set it up. 

After that this looks good to be merged.

jupyterhub/app.py
```diff
@@ -1673,6 +1673,9 @@ def init_hub(self):
                     raise ValueError("Token name %r is not in whitelist" % name)
                 if not self.authenticator.validate_username(name):
                     raise ValueError("Token name %r is not valid" % name)
+            if kind == 'Service':
```
betatim: should it be Service or service?

katsar0v: Will correct it, my mistake

katsar0v:
Will make the needed adjustments. Thank you for your feedback

betatim:
Nice work! Merged.

PR #2712, opened 2019-08-30
ORM: allow removed MySQL variables to not exist

chicocvenancio:
In current versions of MySQL and MariaDB `innodb_file_format`
and `innodb_large_prefix` have been removed. This allows them to not
exist and makes sure the format for the rows are `Dynamic` (default
for current versions).

Starting Jupyterhub with a current version of MySQL or MariaDB results in a `KeyError`. This was a bug introduced by me, hopefully I get it right this time.

### DIFF
--- a/jupyterhub/orm.py
+++ b/jupyterhub/orm.py
@@ -770,8 +770,8 @@ def mysql_large_prefix_check(engine):
         ).fetchall()
     )
     if (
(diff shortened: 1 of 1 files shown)


chicocvenancio:
Hey @minrk, could you take a look? #1866 is now bitting me.

minrk:
Thanks, makes sense!

PR #2721, opened 2019-09-06
Add JupyterHub.init_spawners_timeout

### LINKED ISSUE #2525: Slow startup is back
We used to do a lot of work during startup that scaled by the number of users (instantiating too many Spawner objects, even when they weren't running). This was fixed a few releases ago to only instantiate Spawner objects that might be running, calling `.poll()` on them, and the startup polling is now concurrent across running Spawners.

In 1.0, #2297 adds an *http* poll for each apparently running server to protect against partially alive Spawners, verifying that the Spawners are alive at their expected URLs. As discovered on mybinder.org today, this can significantly increase Hub restart time when there are lots of running servers.

The most logical solution is to delay the correct-url check until after the Hub begins accepting connections. This sounds good, since partial launches are rare, but it would also introduce some potentially complex race conditions with launches that might start after the Hub begins accepting connections, but before it has finished recovering from the database into a consistent state.

minrk:
If init_spawners takes too long (default: 10 seconds) to complete, app start will be allowed to continue while finishing in the background.

Adds new `check` pending state for the initial check.

Checking lots of spawners can take a long time, so allowing this to be async limits the impact on startup time at the expense of starting the Hub in a not-quite-fully-ready state.

If we determine that this is super reliable, we can reduce the default timeout to zero.

closes #2525

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -11,8 +11,10 @@ import re
 import signal
 import socket
 import sys
+import time
 from concurrent.futures import ThreadPoolExecutor
 from datetime import datetime
+from datetime import timedelta
 from datetime import timezone
 from functools import partial
 from getpass import getuser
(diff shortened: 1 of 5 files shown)


yuvipanda:
This keeps biting me, so I love it

jupyterhub/app.py
```diff
@@ -2156,11 +2195,61 @@ def _log_cls(name, cls):
...
+
+            # don't allow a zero timeout because we still need to be sure
+            # that the Spawner objects are defined and pending
+            await gen.with_timeout(
+                timedelta(seconds=max(init_spawners_timeout, 1)), init_spawners_future
+            )
+        except gen.TimeoutError:
+            self.log.warning(
+                "init_spawners did not complete within %i seconds. "
+                "Allowing to complete in the background.",
```
willingc: Looks great @minrk. Do we also log after it completes in the background?

minrk: Yeah, [in the callback](https://github.com/jupyterhub/jupyterhub/pull/2721/files#diff-a9c49d54d1416c0a6f0a70bc047bb72dR2213-R2216)

Issue #2724, opened 2019-09-07
Moving Profile selection to core JupyterHub


rkdarst:
I discussed the multiple profile selection tools ([ProfileSpawner](https://github.com/jupyterhub/wrapspawner), the one in [KubeSpawner](https://github.com/jupyterhub/kubespawner), etc) with @minrk at the Oslo JupyterHub sprint.  His proposal was that profiles could be merged directly into JupyterHub since they have been proven useful by now.  This could directly select the spawner class, and avoid some of the problems with wrapspawner.

This would need at least the features:
- Select a spawner
- Pass options to those spawners
- Support options forms (maybe?)

What do you all think about this?
- Who is for and against?
- What functions in wrapspawner would still be needed?  For what functionality would wrapspawer still be needed?  Could it be removed?
- What configurability would be needed in JupyterHub, e.g. dynamic profile lists, etc.

We talked about implementation, it could be:
- add a `c.JupyterHub.profile_list = List` configurable
  - List of `{"id": str, "display_name": str, "description": str, "spawner_class": Spawner, "spawner_options": dict, "default": bool}`
- add a `c.JupyterHub.profile_list_template = String` configurable - jinja template that renders the form.
- anywhere that `options_form` is rendered, if `JupyterHub.profile_list` is nonempty, render that instead
- `profile_list_template` can include "extra options form" at the bottom if you need profiles *and* options for those profiles.  These are somehow passed on to the chosen spawner_class.  Details on how this works or if it is needed is unknown.

(This issue is copied from https://github.com/jupyterhub/wrapspawner/issues/33 to notify people there, please use this location as the master copy.)

rcthomas:
* For/against: Most definitely **for** it and have thought it was the way to go for a long time.  Thanks for working on this.

* Functions/functionality: All.  That's not meant to be a joke, what wrapspawner does is already scoped very well (select a spawner at runtime, configure spawners in the config file).

* Configurability: Of all the deployments I think we've possibly done the most with what we have to enable configurability and user experience.  A lot of it is still experimental but we are open to new approaches.  I'll elaborate below, this is where I have the most comments.

(1) The UI doesn't have to be a one-size fits all solution for everyone, keep it override-able.  We didn't use ProfileSpawner and its default options form because our view is that there are some things that should just be a single click.  We were happy customizing the home/console page for this.

[...]

PR #2735, opened 2019-09-18
Non empty default server names (useful for profiles?)

krinsman:
Added configurable default server name attribute to better match behavior described for user-redirect in [urls.md in the docs](https://github.com/jupyterhub/jupyterhub/blob/071e375d5fd038e70c29c619d199bf2c9c090d5c/docs/source/reference/urls.md#user-redirect). In particular something in this direction should allow default servers besides `""`, and thus allow `/url-redirect` to redirect to a user server besides `""`. This is helpful e.g. for people using profile spawners where there is a default profile, and so by default someone would want to launch a notebook using the default profile rather than using the `""` server. Thus this PR is _somewhat_ related to this JupyterHub issue:  https://github.com/jupyterhub/jupyterhub/issues/2724

That being said I feel like this is a somewhat kludgy implementation (which is why it was important to me that this does nothing by default, which it does at least according to my testing), so suggestions for improvement would be much appreciated.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -911,6 +911,22 @@ class JupyterHub(Application):
         """,
     ).tag(config=True)
 
+    default_server_name = Unicode(
+        "",
+        help="If named servers are enabled, default name of server to spawn or open, e.g. by user-redirect.",
+    ).tag(config=True)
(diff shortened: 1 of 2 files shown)


minrk:
Makes sense, thanks!

krinsman:
Oh wow, thank you!

PR #2739, opened 2019-09-22
Fix deprecated call and true comparison

kinow:
Hi,

Just replacing a `if var == True` by simply `if var` (faster, but not important here I think), and two deprecated calls.

The deprecated calls were fixed according to the deprecation message, that says to access the properties directly now, instead of using the `.server` member.

Formatting performed by `black` + `git commit` hook.

Cheers
Bruno

### DIFF
--- a/jupyterhub/objects.py
+++ b/jupyterhub/objects.py
@@ -213,8 +213,4 @@ class Hub(Server):
         return url_path_join(self.url, 'api')
 
     def __repr__(self):
-        return "<%s %s:%s>" % (
-            self.__class__.__name__,
-            self.server.ip,
-            self.server.port,
-        )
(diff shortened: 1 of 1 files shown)


jupyterhub/handlers/base.py
```diff
@@ -328,7 +328,7 @@ def _record_activity(self, obj, timestamp=None):
 
         user._auth_refreshed = now
 
-        if auth_info == True:
+        if auth_info:
```
minrk: This isn't correct. True is a distinct value from a truthy dict.

kinow: Roger, prevent any chance of a... __bool__ I think? Slipping and passing this test, or any other possible value like a non empty array I think. Let me update this commit.

kinow: commit dropped, push-force'd

PR #2747, opened 2019-09-26
Reduce verbosity for "Failing suspected API request to not-running server"

rkdarst:
- API requests to non-running servers are not uncommon when you cull
  servers and people leave tabs open and active.  It returns with 503
  and *logs all headers*, which can take up half of our total log
  lines
- This adds logic to reduce verbosity of these logs by making a note
  of the failure on the handler and then overriding the log handling.

Unresolved questions:

- Is there a better way to do this?  This is kind of hacky and
  probably there's a more elegant way.
- I welcome nits on naming of the variable and the chosen override log
  level, since I'm not very happy with it now.

### DIFF
--- a/jupyterhub/log.py
+++ b/jupyterhub/log.py
@@ -130,6 +130,8 @@ def log_request(handler):
         log_method = access_log.info
     elif status < 500:
         log_method = access_log.warning
+    elif getattr(handler, 'failing_non_running_server', False):
+        log_method = access_log.debug
     else:
         log_method = access_log.error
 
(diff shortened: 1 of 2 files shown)


jupyterhub/log.py
```diff
@@ -162,7 +164,14 @@ def log_request(handler):
         location='',
     )
     msg = "{status} {method} {uri}{location} ({user}@{ip}) {request_time:.2f}ms"
-    if status >= 500 and status != 502:
```
minrk: I think we could just add `and status not in {502, 503}` for this. The reason 502 is here is really the same as what you are after.

rkdarst: I saw that and it looked *so* tempting to filter out 503, but didn't know if 503 had some other meaning.  If you approve I'll just replace this whole diff with that.

It looks like the two uses of 503 are related to stopped servers, so I'll go ahead and do this.  thanks!

minrk: We log this extra stuff for debugging purposes when the error is likely the fault on the Hub side. 502 and 503 are errors that are generally "not my fault," so don't warrant the extra debugging.

It's possible that we should only be doing it for 500 exactly, which is the case for unhandled server-side errors

PR #2750, opened 2019-09-26
wait for proxy before accepting requests

### LINKED ISSUE #2745: hub->proxy web-request timebombs - how to not explode?
## Short version based on findings

I've been working on the Z2JH projects CI system a lot recently, which often fails seemingly at random. I figured it out I think.

proxy.py's function `api_request` used to communicate with the proxy can _fail hard_ and crash the whole jupyterhub process. This often happens even though both the proxy and the hub pod in z2jh has become ready in the kubernetes sense, which they only become after they have started listening to web endpoints.

The hub pod has a function (`check_routes`) that is run on startup and is scheduled to run regularly after this, but it is done in an async manner. This leaves the hub able to start operating even before it has ever run a `check_routes` call to completion. The `check_routes` function makes calls to the proxy api, but if for example, the proxy is unavailable somehow, we can end up with a timebomb because the hub will have a async webrequests that in 20 seconds will timeout and fail hard within the `api_request` function in proxy.py.

I consider some action points in https://github.com/jupyterhub/jupyterhub/issues/2745#issuecomment-535190293.

---

## Long version from the start

I'm debugging something in depth, and I believe that that the z2jh CI pipeline is unstable because of a unexpected timing.

From the code block below, the hub pod can and do crash in the z2jh CI pipeline, and I can't figure out why, but end up with some interesting observation I hope someone can shed light on.

https://github.com/jupyterhub/jupyterhub/blob/5b13f96162b77b4ae23306327cd700b8faeb299c/jupyterhub/proxy.py#L758-L774

Looking at the crash logs below. We can see that this code block started running, but only five seconds later it errors because of a 20 second timeout on a request. I don't get that. Is it reusing some connection? What is going on here? I figure 

```
    [I 2019-09-25 12:03:01.117 JupyterHub app:2337] Hub API listening on http://0.0.0.0:8081/hub/

    ### NOTE: The hub started up 20 seconds earlier...

    [I 2019-09-25 12:03:16.796 JupyterHub proxy:281] Removing user testuser-c947c528-06d0-4d0f-be5e-ceb6f1c32dfc from proxy (/user/testuser-c947c528-06d0-4d0f-be5e-ceb6f1c32dfc/)
    ## NOTE: The log line below is generated directly from the code above.
    ##       This is just before making the delete request.
    [D 2019-09-25 12:03:16.796 JupyterHub proxy:765] Proxy: Fetching DELETE http://10.99.19.102:8001/api/routes/user/testuser-c947c528-06d0-4d0f-be5e-ceb6f1c32dfc
    [D 2019-09-25 12:03:16.799 JupyterHub user:724] Stopping testuser-c947c528-06d0-4d0f-be5e-ceb6f1c32dfc
    [I 2019-09-25 12:03:16.800 JupyterHub spawner:1758] Deleting pod jupyter-testuser-2dc947c528-2d06d0-2d4d0f-2dbe5e-2dceb6f1c32dfc
    [E 2019-09-25 12:03:21.132 JupyterHub app:2482]
        File "/usr/local/lib/python3.6/dist-packages/jupyterhub/app.py", line 2480, in launch_instance_async
            await self.start()
        File "/usr/local/lib/python3.6/dist-packages/jupyterhub/app.py", line 2405, in start
            await self.proxy.check_routes(self.users, self._service_map)
        File "/usr/local/lib/python3.6/dist-packages/jupyterhub/proxy.py", line 62, in locked_method
            return await method(*args, **kwargs)
        File "/usr/local/lib/python3.6/dist-packages/jupyterhub/proxy.py", line 315, in check_routes
            routes = await self.get_all_routes()
        File "/usr/local/lib/python3.6/dist-packages/jupyterhub/proxy.py", line 804, in get_all_routes
            resp = await self.api_request('', client=client)
        File "/usr/local/lib/python3.6/dist-packages/jupyterhub/proxy.py", line 773, in api_request
            result = await client.fetch(req)
        tornado.curl_httpclient.CurlError: HTTP 599: Connection timed out after 20000 milliseconds
```

## Relevant JupyterHub configuration
Z2JH configures this [in its jupyterhub_config.py](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/d25bcd9e38aca2734f90e51c2bc9a0bf9d90ca3e/images/hub/jupyterhub_config.py#L5-L15) which I think is relevant. The error was after all a `tornado.curl_httpclient.CurlError` error.

```python
from tornado.httpclient import AsyncHTTPClient
from kubernetes import client
from jupyterhub.utils import url_path_join

from z2jh import get_config, set_config_if_not_none

# Configure JupyterHub to use the curl backend for making HTTP requests,
# rather than the pure-python implementations. The default one starts
# being too slow to make a large number of requests to the proxy API
# at the rate required.
AsyncHTTPClient.configure("tornado.curl_httpclient.CurlAsyncHTTPClient")
```

## Related reference
- https://www.tornadoweb.org/en/stable/httpclient.html#tornado.httpclient.AsyncHTTPClient.configure
- [`request_timeout`](https://www.tornadoweb.org/en/stable/httpclient.html#request-objects) is the configuration defaulting to 20 which makes the request timeout after 20 seconds (20000 milliseconds).
- For a full view of the logs I've been dissecting, see [this TravisCI job's logs](https://travis-ci.org/jupyterhub/zero-to-jupyterhub-k8s/jobs/589410196).

## Background context
Z2JH CI runs a pytest to spawn a user called `test_api_request_user_spawn`, and this fails often for the reason above. The events leading up to this are described below. Currently, this test fails about 50% of the time in my development branch where I use a `kind` k8s cluster instead of a `minikube` k8s cluster.

- The `jupyterhub_user` pytest fixture used by the test function is instantiated and triggers the creation of a JupyterHub user through the JupyterHub API
- `test_api_request_user_spawn` use the JupyterHub API to spawn a user
- `test_api_request_user_spawn` calls `_wait_for_user_to_spawn` that probes the JupyterHub API for a registered server until it finds one
- `test_api_request_user_spawn` cleans up by calling `_delete_server` which calls the JupyterHub API to stop the user server
  - The Hub runs and crashes like described above
- pytest conclude `test_api_request_user_spawn` failed, the function scope is closing down which causes the `jupyterhub_user` fixture finish up by calling the JupyterHub API to delete the user, but JupyterHub has crashed so this function errors.

minrk:
ensures the Hub doesn't appear "ready" before it can talk to the proxy

closes #2745

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -2452,6 +2452,20 @@ class JupyterHub(Application):
             loop.stop()
             return
 
+        # start the proxy
+        if self.proxy.should_start:
+            try:
+                await self.proxy.start()
+            except Exception as e:
+                self.log.critical("Failed to start proxy", exc_info=True)
+                self.exit(1)
+        else:
+            self.log.info("Not starting proxy")
+
+        # verify that we can talk to the proxy before listening.
+        # avoids delayed failure if we can't talk to the proxy
+        await self.proxy.get_all_routes()
+
         ssl_context = make_ssl_context(
             self.internal_ssl_key,
             self.internal_ssl_cert,
@@ -2488,16 +2502,6 @@ class JupyterHub(Application):
         except Exception:
             self.log.error("Failed to bind hub to %s", self.hub.bind_url)
             raise
-
-        # start the proxy
-        if self.proxy.should_start:
-            try:
-                await self.proxy.start()
-            except Exception as e:
-                self.log.critical("Failed to start proxy", exc_info=True)
-                self.exit(1)
-        else:
-            self.log.info("Not starting proxy")
 
         # start the service(s)
         for service_name, service in self._service_map.items():


jupyterhub/app.py
```diff
@@ -1457,14 +1457,12 @@ def init_internal_ssl(self):
...
-                self.internal_ssl_cert,
-                cafile=self.internal_ssl_ca,
-            )
-            AsyncHTTPClient.configure(None, defaults={"ssl_options": ssl_context})
+        # Configure the AsyncHTTPClient. This will affect anything using
+        # AsyncHTTPClient.
+        self.ssl_context = make_ssl_context(
+            self.internal_ssl_key, self.internal_ssl_cert, cafile=self.internal_ssl_ca
+        )
+        AsyncHTTPClient.configure(None, defaults={"ssl_options": self.ssl_context})
```
consideRatio: This is no longer part of the `if` clause as it were before, is that as it should be?

minrk:
Revised the PR to make a smaller check and updated the title to reflect it.

consideRatio:
LGTM! Merge?

minrk:
@consideRatio go ahead!

PR #2751, opened 2019-09-26
Reduce verbosity for "Failing suspected API request to not-running server" (new)

rkdarst:
- API requests to non-running servers are not uncommon when you cull
  servers and people leave tabs open and active. It returns with 503
  and logs all headers, which can take up half of our total log lines
- This avoids logging headers for all 502 and 503 return statuses.
  #2747 presented an alternative (more complex) implementation, but this
  turned out to be appropriate.
- Closes: #2747

### DIFF
--- a/jupyterhub/log.py
+++ b/jupyterhub/log.py
@@ -162,7 +162,7 @@ def log_request(handler):
         location='',
     )
     msg = "{status} {method} {uri}{location} ({user}@{ip}) {request_time:.2f}ms"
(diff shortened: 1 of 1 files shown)


minrk:
For future reference, you can force-push to existing branches to update open PRs

rkdarst:
> For future reference, you can force-push to existing branches to update open PRs

I do it for almost every PR... in this case I thought it worth saving
old and new implementations for reference, not that old was worth
much...

PR #2752, opened 2019-09-27
Log JupyterHub version on startup

consideRatio:
Typically useful when reading logs, perhaps should go another place, or include even more information etc, but the gist is perhaps captured.


minrk:
👍

PR #2755, opened 2019-10-01
Expose spawner.user_options in REST API.

danielballan:
This adds `spawner.user_options` to the `server_model`, meaing that an admin or owner who requests `GET /users/{name}/` can see how each running server was spawned.

The motivating use case is the Hub Service [https://github.com/danielballan/jupyterhub-share-link](jupyterhub-share-link) which enables short-term sharing between users on the same Hub. See the [discourse thread](https://discourse.jupyter.org/t/request-for-early-feedback-on-jupyterhub-share-link/1543/16) for details and discussion.

The service aims to share the _environment_ as well as a notebook document. It does this by encoding some information about the sender's server in the share link and attempting to start a similar server for the recipient (or find one that is already running). This approach requires, of course, that `allow_named_servers = True`.

Currently, the Hub Service relies on a small server extension that understands how to extract the relevant details from the environment (i.e. `JUPYTER_IMAGE_SPEC`). It is fairly coupled to specific spawners. I believe that this PR would remove the need for a server extension and also generalize it to support most spawners.

### DIFF
--- a/jupyterhub/apihandlers/base.py
+++ b/jupyterhub/apihandlers/base.py
@@ -141,6 +141,7 @@ class APIHandler(BaseHandler):
             'ready': spawner.ready,
             'state': spawner.get_state() if include_state else None,
             'url': url_path_join(spawner.user.url, spawner.name, '/'),
+            'user_options': spawner.user_options,
             'progress_url': spawner._progress_url,
         }
 
--- a/jupyterhub/tests/test_named_servers.py
+++ b/jupyterhub/tests/test_named_servers.py
@@ -57,6 +57,7 @@ async def test_default_server(app, named_servers):
                         username
                     ),
                     'state': {'pid': 0},
+                    'user_options': {},
                 }
             },
         }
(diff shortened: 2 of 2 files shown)


meeseeksmachine:

https://discourse.jupyter.org/t/request-for-early-feedback-on-jupyterhub-share-link/1543/19

danielballan:
Does the change proposed here raise any thorny issues? Before I proceed with more development on jupyterhub-share-link, it would be useful to know if I should plan on Hub adding this API or something similar.

minrk:
Sorry for neglecting this. I don't think there are any issues here.

danielballan:
Thanks for taking a look.

animesh2706:
Hi Daniel, Thanks for sharing your work. 
Please suggest exactly what changes required in my current configuration in order to enable link sharing.
All my users will be part of same hub. 

I am currently running jupyterhub 1.0 with dockerspawner. 
I have enabled allow_named_servers to True.

I could find base.py in apihandlers directory.
Please suggest.

Thanks.

PR #2763, opened 2019-10-05
Link services

rcthomas:
Adds a list element to the navbar that presents a list of links to accessible services.  If no services are accessible the list element is not shown.

For a service to be accessible it must have a URL.  Also, if the service runs with admin privileges, then the user must be an admin for it to be accessible.

An issue is that while a service may have a URL, it may not be desirable to expose that service as a link.  For instance the service may just be a REST API that doesn't need to be surfaced to normal users.  One way to address this would be to add a property to service configuration that would prevent a service from appearing in the list if set.  I could add that the PR if it sounds like a good suggestion but am open to alternatives.

Another issue is that, in principle, services may be accessible without users being authenticated at all but the login page won't present them.  I'm not terribly worried about this inconsistency but thought I would mention it; I'm willing to let it go for now.

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -1107,10 +1107,21 @@ class BaseHandler(RequestHandler):
             logout_url=self.settings['logout_url'],
             static_url=self.static_url,
             version_hash=self.version_hash,
+            services=self.get_accessible_services(user),
         )
         if self.settings['template_vars']:
             ns.update(self.settings['template_vars'])
         return ns
+
+    def get_accessible_services(self, user):
+        accessible_services = list()
+        for service in self.services.values():
+            if not service.url:
+                continue
+            if service.admin and not user.admin:
+                continue
+            accessible_services.append(service)
+        return accessible_services
 
     def write_error(self, status_code, **kwargs):
         """render custom error pages"""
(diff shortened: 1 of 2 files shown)


jupyterhub/handlers/base.py
```diff
@@ -1107,11 +1107,22 @@ def template_namespace(self):
...
         if self.settings['template_vars']:
             ns.update(self.settings['template_vars'])
         return ns
 
+    def get_accessible_services(self, user):
+        accessible_services = list()
+        for service in self.services.values():
+            if not service.url:
+                continue
+            if service.admin and not user.admin:
```
minrk: This isn't necessarily a correct condition. service.admin means that the service itself has admin permissions (e.g. launching servers for users), not that only admin users can use it. I'd just remove it for now.

consideRatio:
@rcthomas could you update the description of the PR to reflect the changes since the PR was created? I will help out with writing the changelog I figured, and isn't confident the current description reflects the PR anymore.

rcthomas:
@consideRatio Sure thing, thanks.  All that's happening here now is:

Add a list element to the navbar that presents authenticated users with links to services that have URLs.

minrk:
This is great, thanks!

PR #2767, opened 2019-10-08
add `service.oauth_no_confirm` configuration

minrk:
allows services to be explicitly blessed to skip the extra oauth confirmation page

added in 1.0

This confirmation page is unhelpful for many admin-managed services,
and is mainly intended for cross-user access.

The default behavior is unchanged, but services can now opt-out of confirmation
(as is done already for the user's own servers).

Use with caution, as this eliminates users' ability to confirm that a service
should be able to authenticate them.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -2034,7 +2034,7 @@ class JupyterHub(Application):
         for service in self._service_map.values():
             if service.oauth_no_confirm:
                 self.log.warning(
-                    "Allowing service %s to complete OAuth without confirmation",
+                    "Allowing service %s to complete OAuth without confirmation on an authorization web page",
                     service.name,
                 )
                 oauth_no_confirm_whitelist.add(service.oauth_client_id)
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -2030,6 +2030,15 @@ class JupyterHub(Application):
         else:
             version_hash = datetime.now().strftime("%Y%m%d%H%M%S")
 
+        oauth_no_confirm_whitelist = set()
+        for service in self._service_map.values():
+            if service.oauth_no_confirm:
+                self.log.warning(
+                    "Allowing service %s to complete OAuth without confirmation",
+                    service.name,
+                )
+                oauth_no_confirm_whitelist.add(service.oauth_client_id)
+
         settings = dict(
             log_function=log_request,
             config=self.config,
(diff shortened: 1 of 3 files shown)


jupyterhub/app.py
```diff
@@ -2030,6 +2030,15 @@ def init_tornado_settings(self):
         else:
             version_hash = datetime.now().strftime("%Y%m%d%H%M%S")
 
+        oauth_no_confirm_whitelist = set()
+        for service in self._service_map.values():
+            if service.oauth_no_confirm:
+                self.log.warning(
+                    "Allowing service %s to complete OAuth without confirmation",
```
willingc:
```suggestion
                    "Allowing service %s to complete OAuth without confirmation on an authorization web page",
```

jupyterhub/services/service.py
```diff
@@ -197,6 +202,21 @@ class Service(LoggingConfigurable):
...
     ).tag(input=True)
 
+    oauth_no_confirm = Bool(
+        False,
+        help="""Skip OAuth confirmation when users access this service.
+
+        By default, when users authenticate with a service using JupyterHub,
+        they are prompted to confirm that they want to grant that service
+        access to their credentials.
+        Setting oauth_no_confirm=True skips the confirmation for this service.
```
willingc:
```suggestion
        Setting oauth_no_confirm=True skips the confirmation web page for this service.
```

jupyterhub/services/service.py
```diff
@@ -197,6 +202,21 @@ class Service(LoggingConfigurable):
...
+        Useful for admin-managed services that are considered part of the Hub,
```
willingc:
```suggestion
        Skipping the confirmation page is useful for admin-managed services that are considered part of the Hub
```

jupyterhub/services/service.py
```diff
@@ -197,6 +202,21 @@ class Service(LoggingConfigurable):
...
+        which shouldn't need extra prompts for login.
```
willingc:
```suggestion
        and shouldn't need extra prompts for login.
```

willingc:
Hmm...not sure what is causing the nightly test to fail on the assert.

The docs error seems to be due to bad formatting in a docstring unrelated to this PR.

minrk:
The failure was indeed one of the docstrings edited here. Should be fixed now, I think.

PR #2770, opened 2019-10-10
Update to default slow_spawn_timeout

ryogesh:
With heavy containers, first launch fails with a warning "User <name> is slow to become responsive (timeout=10)" 
Hence setting this timeout default to 30 seconds making it consistent with Spawner.start_timeout  and Spawner.http_timeout
Further, setting this config  e.g. c.JupyterHub.tornado_settings = {'slow_spawn_timeout': 30} is not easily available on the documentation

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -714,7 +714,7 @@ class BaseHandler(RequestHandler):
 
     @property
     def slow_spawn_timeout(self):
-        return self.settings.get('slow_spawn_timeout', 10)
+        return self.settings.get('slow_spawn_timeout', 30)
 
     @property
     def slow_stop_timeout(self):


minrk:
Apologies for not documenting this. Indeed, it should be covered. It shouldn't be 30 seconds, though, and best practice these days is usually to make it *shorter*, using either 0 or 1 seconds. This timeout is not the same as the others because it shouldn't affect success, it only affects whether to respond with a `202 Accepted` status for spawns that are slow. 30 seconds can result in slow requests being dropped by proxies, so this number actually must be lower than 20 to be reliable (this is deployment specific). If the recently added progress pages go well, the default may be dropped to 0.

ryogesh:
@minrk Thanks for the detailed explanation. 
Didn't realize the difference of this timeout. I agree on it being closer to 0.

PR #2771, opened 2019-10-10
Documentation/Configure User Environment: Adding Info for SystemUserS…

rschauf:
…pawner to put server names together properly

### DIFF
--- a/docs/source/reference/config-user-env.md
+++ b/docs/source/reference/config-user-env.md
@@ -179,3 +179,16 @@ The number of named servers per user can be limited by setting
 ```python
 c.JupyterHub.named_server_limit_per_user = 5
 ```
+
+### The SystemUserSpawner
+
+Spawned containers must have unique names. To be able to spawn more than
+one container per user, the spawner must have a rule to construct unique 
+container names. This can be achieved by adding a template string to the 
+`jupyterhub_config.py` togehter with the allow_named_servers tag mentioned above.
+
+To set up correct server name construction add:
+
+```
+c.SystemUserSpawner.name_template = '{prefix}-{username}-{servername}'
+```


betatim:
This looks good to me. One thing I am wondering is if we have a general rule on including (or not) spawner specific things.

https://github.com/jupyterhub/dockerspawner (the home of system user spawner) doesn't have its own docs page so .. :-/

docs/source/reference/config-user-env.md
```diff
@@ -179,3 +179,16 @@ The number of named servers per user can be limited by setting
...
+
+Spawned containers must have unique names. To be able to spawn more than
+one container per user, the spawner must have a rule to construct unique 
+container names. This can be achieved by adding a template string to the 
+`jupyterhub_config.py` togehter with the allow_named_servers tag mentioned above.
+
+To set up correct server name construction add:
+
+```
+c.SystemUserSpawner.name_template = '{prefix}-{username}-{servername}'
```
betatim: Is it worth making this whole section more generic by replacing `SystemUserSpawner` with `DockerSpawner` and tweaking the text to say "The whole docker spawner family (`DockerSpawner`, `SystemUserSpawner` and ...) support this"?

rschauf: This is a good question. I am as a user not familiar with all the details of the configuration of the jupyterhub features. The opposite is the case. When setting up jupyterhub for the first time, it was really hard for me to get it running, as the docu didn't cover everything I had to know to get it running. The same way here, I do not know if this template also works for the other spawner types from the dockerspawner family. therefore I just wrote what I knew. Of course, if this is valid for all spawner types, this should be mentioned here.

betatim: From a look at the code it looks like the whole group of spawners in the `dockerspawner` package supports this. So we could include all of them. I don't think it is a general thing for all spawners that exist though.

rschauf: ok, then I better should take back the PR and generate a new one with all spawner types mentioned ... 😄

rschauf: .... oops, where is the revert button for this action? Or can you reject it/close it?

betatim: You can add another commit to this PR which updates it. No need to make a new PR.

GeorgianaElena: Hey @rschauf! :wave: 
The `name_template` configurable should be `{prefix}-{username}-{servername}` by default in the latest `dockerspawner` version when `JupyterHub.allow_named_servers` is true. Did you run into issues with this when not specifying it explicitly?

Also, when a new dockerspawner version will be released (it should be soon), the `name_template` default value will be `{prefix}-{username}__{servername}` (see https://github.com/jupyterhub/dockerspawner/pull/386 for context).

manics:
Dockerspawner has it's own docs page:
https://jupyterhub-dockerspawner.readthedocs.io/en/latest/
I think the any relevant docs should be kept there, otherwise we have a risk of duplication of docs, or things going out of date.

Issue #2758, opened 2019-10-02
named_server_limit_per_user not working?


vilhelmen:
I don't seem to be able to get named server limits working on our 1.0.0 test server. I've tried the recently introduced Z2JH `hub.namedServerLimitPerUser` as well as raw configuration via `c.JupyterHub.named_server_limit_per_user` and neither seem to be working. I don't have any concurrent spawn limits enabled, as that seemed like a site-wide thing. I'm able to keep creating servers without any restriction both as admin and as a regular user.

I tried waiting for the container to fully start before creating more in case it was a race condition, but that didn't seem to fix it either.

No warnings/errors in the logs, but it does seem to be tracking totals correctly:
```
[D 2019-10-02 21:47:00.809 JupyterHub base:787] 0/64 concurrent spawns
[D 2019-10-02 21:47:00.809 JupyterHub base:792] 4 active servers
```

Config dump, but there's not much of interest in there:
```
hub:
  image:
    name: 'XXXX/jupyter/zero-to-jupyterhub-k8s/dev/hub'
    tag: 'latest'
  imagePullPolicy: Always
  imagePullSecret:
    XXXX
  nodeSelector:
    workset: cpu
  allowNamedServers: True
  namedServerLimitPerUser: 3
  extraConfig:
    1: |
       c.KubeSpawner.profile_list.append({'display_name': 'Debug Image', 'kubespawner_override': {'image': 'XXXX/jupyter/docker-stacks/dev/singleuser-dev:latest-shim'}})
    2: |
       c.JupyterHub.named_server_limit_per_user = 3

singleuser:
  cmd: null
  extraEnv:
    DEVDEBUG: "yespls"
  image:
    name: 'XXXX/jupyter/docker-stacks/dev/singleuser-minimal'
    tag: 'latest-shim'
  imagePullPolicy: Always
  imagePullSecret:
    XXXX

  profileList:
   -  display_name: 'Minimal'
      default: False
      kubespawner_override:
        image: 'XXXXX/jupyter/docker-stacks/dev/singleuser-minimal:latest-shim'
        cpu_guarantee: 0.5
        cpu_limit: 2
        mem_limit: '4G'
        node_selector:
          workset: cpu
  XXXX

prePuller:
  hook:
    enabled: false
  continuous:
    enabled: false

debug:
  enabled: true

auth:
  type: 'ldap'
  XXX

proxy:
  secretToken: 'XXX'
  service:
    loadBalancerIP: XXX
[...]

PR #2772, opened 2019-10-11
fix named server checks

bitnik:
Currently checks for named servers (if named servers are allowed and limit for named servers) are only checked in API handler (`UserServerAPIHandler`). They are not checked in `SpawnHandler` which is called when user interacts with buttons in home page. So even named servers are not allowed, user can enter this url `/hub/spawn/<user_name>/<server_name>` manually and jhub creates this named server. User can even create as many many named servers in the same way (https://github.com/jupyterhub/jupyterhub/issues/2758).

In this PR I copied code piece, which does these checks, from `UserServerAPIHandler` to `SpawnHandler`. And also fixed the test `test_named_server_spawn_form()`, it was not using the fixture which enables named servers.

I have 1 question for my education: Isn't it better to raise 409 error when user tries to launch more than allowed number of named servers?

### DIFF
--- a/jupyterhub/handlers/pages.py
+++ b/jupyterhub/handlers/pages.py
@@ -118,6 +118,23 @@ class SpawnHandler(BaseHandler):
             if user is None:
                 raise web.HTTPError(404, "No such user: %s" % for_user)
 
+        if server_name:
+            if not self.allow_named_servers:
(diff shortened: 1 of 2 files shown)


minrk:
Thanks!

PR #2773, opened 2019-10-12
Handle the protocol when ssl is enabled and log the right URL

### LINKED ISSUE #2621: Link to hub given as HTTP and not HTTPS

kinow:
Uses traitlets `@validate` and `@default` (thanks @minrk !) to define the `.bind_url` of a JupyterHub application, also handling when SSL is enabled - in which case, we want the `bind_url` to have `https`, not `http`.

(updated the description of this PR to reflect the latest comments, so some of the discussion below is out of sync)

Closes #2621

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -576,6 +576,22 @@ class JupyterHub(Application):
         """,
     ).tag(config=True)
 
+    @validate('bind_url')
+    def _validate_bind_url(self, proposal):
+        """ensure protocol field of bind_url matches ssl"""
+        v = proposal['value']
+        proto, sep, rest = v.partition('://')
+        if self.ssl_cert and proto != 'https':
+            return 'https' + sep + rest
+        elif proto != 'http' and not self.ssl_cert:
+            return 'http' + sep + rest
+        return v
+
+    @default('bind_url')
+    def _bind_url_default(self):
+        proto = 'https' if self.ssl_cert else 'http'
+        return proto + '://:8000'
+
     subdomain_host = Unicode(
         '',
         help="""Run single-user servers on subdomains of this host.


kinow:
> This is great @kinow! Thanks a lot for the PR tulip .

Thank you for taking your time to review it.

> I believe we can make the Hub log the right protocol without introducing a new property, by updating `bind_url` if ssl is enabled, just before the proxy is initialized. What do you think?
> (`proxy.public_url` is initialized with `bind_url` [here](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/app.py#L2053))

You are right! I didn't want to change anything as I was afraid it could break something. But I tried your suggestion, and it seems to work fine.

Furthermore, now both `app` and `proxy` both have `https` protocol in their URL's, which seems more consistent, and with less risk of another log line showing http instead of https elsewhere.

Updated the tests, splitting in both `test_proxy.py` and `test_app.py` (but happy to merge them in a single test if that's better).

Thanks!
Bruno

kinow:
>Updated the tests, splitting in both test_proxy.py and test_app.py (but happy to merge them in a single test if that's better).

Scratch that. My unit test was changing `app.ssl_key` and `app.ssl_crt`, then calling `app.init_proxy()` again. Which works for that unit test, but then breaks other tests due to the `app` fixture in `conftest.py` being created with scope module (so my small unit test caused other tests to fail with 403).

There is no mechanism in place to test external SSL, so perhaps this change could be tested manually only? :disappointed_relieved:

jupyterhub/app.py
```diff
@@ -2048,6 +2048,8 @@ def cleanup_oauth_clients(self):
     def init_proxy(self):
         """Load the Proxy config"""
         # FIXME: handle deprecated config here
+        if self.ssl_cert and self.ssl_key:
+            self.bind_url = self.bind_url.replace('http', 'https')
```
minrk: to avoid replacing 'http' elsewhere in the url:

```suggestion
            self.bind_url = self.bind_url.replace('http://', 'https://', 1)
```

kinow: Ooohh, hadn't thought about it! Good spot!

GeorgianaElena:
I'm thinking if it wouldn't be better to make sure we have the correct protocol in ``bind_url`` right from the start (when we set it the default value).
Also, I think maybe we should also have a validation in place that can take care of the case when in the config we have something like:
```
c.JupyterHub.ssl_cert = "'/path/to/my.cert"
c.JupyterHub.bind_url = "http://:9000"
```
(or the other way arround).
Since we're choosing the protocol based on ssl_cert and ssl_key, having this config won't break things, but it would not log what is actually happening.

@minrk , @kinow what do you think?

kinow:
Hmm, it is a good point, that could prevent users of raising issues or asking on gitter/etc why they configured one `bind_url` but got another value. Should it fail to initialize the hub if the values are no consistent?

I think there are other values that may lead to confusing scenarios @GeorgianaElena.

```
c.JupyterHub.port = 9999
c.JupyterHub.bind_url = 'http://:8888'
```

Results in: `[I 2019-10-17 22:40:43.192 JupyterHub app:2422] JupyterHub is now running at http://:9999/`

If you remove the `port`, then the `bind_url` value is used, and you get `[I 2019-10-17 22:41:18.305 JupyterHub app:2422] JupyterHub is now running at http://:8888`.

GeorgianaElena:
I don't think it should fail, maybe just log a message about ``bind_url`` being changed. Or maybe not log anything at all, the same way it's done in the example you provided , where ``port/ip/base_url`` overwrite ``bind_url`` silently if different ([ref](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/app.py#L546)), making sure the docs cover these cases.

minrk:
bind_url is a replacement for previous configuration that set only one field at a time (port, ip). We could put a warning if both bind_url and ip or port is set.

minrk:
> I'm thinking if it wouldn't be better to make sure we have the correct protocol in ``bind_url`` right from the start (when we set it the default value).

The traitlets way to do this would be to have a `@validate` method in setting bind_url that sets it, e.g.

```python
@validate('bind_url')
def _validate_bind_url(self, proposal):
    """ensure protocol field of bind_url matches ssl
    proto, sep, rest = proposal.partition('://')
    if self.ssl_cert and proto != 'https':
        return 'https' + sep + rest
    elif proto != 'http' and not self.ssl_cert:
        return 'http' + sep + rest

@default('bind_url')
def _bind_url_default(self):
    proto = 'https' if self.ssl_cert else 'http'
    return proto + '://:8000'
```

kinow:
@minrk 

>The traitlets way to do this would be to have a @validate method in setting bind_url that sets it, e.g.

Done! Pretty much the code you shared, with an extra `return` for when none of the two `if` pass in the validate method (otherwise I was getting `None` locally).

Tested with SSL and without, and the console logs displayed the correct protocol. Ready for review again. Thanks!

minrk:
Thanks!

PR #2774, opened 2019-10-12
Remove unused  setupegg.py

kinow:
Hi,

Looks like the `setuptools` module is no longer used, but instead a command is executed calling `setup.py`. This PR simply removes the unused import and the related comment :+1:

### DIFF
--- a/setupegg.py
+++ b/setupegg.py
@@ -1,7 +1,0 @@
-#!/usr/bin/env python
-"""Wrapper to run setup.py using setuptools."""
-# Import setuptools and call the actual setup
-import setuptools
-
-with open('setup.py', 'rb') as f:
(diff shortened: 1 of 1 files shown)


kinow:
ps: quite sure the test that failed is not related and just needs a kick on Travis

minrk:
If you'd like, you can just remove `setupegg.py`. It's from an ancient way of avoiding setuptools that we don't anymore.

kinow:
> If you'd like, you can just remove `setupegg.py`. It's from an ancient way of avoiding setuptools that we don't use anymore.

Oh, that's even easier. Dropped previous commit, and added a new one removing `setupegg.py`. Updated pull request title as well. Thanks!

minrk:
Thanks!

PR #2775, opened 2019-10-12
Fix header project name typo

kinow:
Hi,

Was looking at the requirements in `setup.py` for another issue when I noticed this typo. Searched and found another one in this repo, and the same issue in other jupyter-related projects (PR's on the way).

Cheers
Bruno


betatim:
Thanks for finding these!

Issue #2757, opened 2019-10-02
Display version number in web interface


vilhelmen:
I know you can ask the API what version is running, but I think it'd be a quick, useful addition to show the server's version number in the admin panel somewhere. it would make it a bit easier for users to quickly find the version number without having to check container contents or use the API.

Or do we already do this? I feel like I saw it somewhere before, but I can't find it now.

kinow:
Haven't seen that on the web UI. But if added, perhaps it would be a good idea to have a setting to disable it too, to hide the server signature. Even if there are other ways to tell what version of the hub is running.

vilhelmen:
I was actually thinking about that. Nginx has a setting called `server_tokens` that controls whether it reports the version or not, and we could use something similar.

kinow:
Same reference I had in mind when I wrote this. We had a request from security team some time ago, and it was super easy to handle that in Nginx, while not so easy for other Java servers we had in our stack (though we were using older versions).

PR #2776, opened 2019-10-12
Display server version on admin page

vilhelmen:
`server_tokens` is taken from a similar nginx setting, but I understand that the name could be confusing. TBH I question the need to turn it off since it's behind the admin page in the first place, maybe we should remove the setting to prevent configuration bloat? See #2757 

I borrowed some classes from the navbar for styling, hopefully that won't make it shoot to the top of the page someday.

image

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1185,6 +1185,10 @@ class JupyterHub(Application):
         False, help="""Shuts down all user servers on logout"""
     ).tag(config=True)
 
+    server_tokens = Bool(
+        True, help="""Display JupyterHub version information on admin page"""
+    ).tag(config=True)
+
     @default('statsd')
     def _statsd(self):
         if self.statsd_host:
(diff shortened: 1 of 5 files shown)


share/jupyterhub/static/less/admin.less
```diff
@@ -1,3 +1,9 @@
 i.sort-icon {
   margin-left: 4px;
 }
+
+.version_footer {
+  position: fixed;
+  bottom: 0;
+  width: 100%;
+}
```
kinow: I think this one caused the build to fail during the lint execution.

https://travis-ci.org/jupyterhub/jupyterhub/jobs/597187918

vilhelmen: Oh, thank you! I was only watching the 3.6 tests and forgot to check back in on the others. I'll do some fiddling.

jupyterhub/app.py
```diff
@@ -1185,6 +1185,10 @@ def _log_file_changed(self, change):
         False, help="""Shuts down all user servers on logout"""
     ).tag(config=True)
 
+    server_tokens = Bool(
+        True, help="""Display JupyterHub version information on admin page"""
```
minrk: I think we can skip this option. This makes sense to do all the time, I think.

vilhelmen: I agree, all users can ask the version from the API as it is. I'll keep it wrapped up as a separate commit for easier revert. Ideally we can have this report other component versions, too!

PR #2783, opened 2019-10-16
Log proxy class

GeorgianaElena:
Now that we have multiple proxies, I believe it would be nice to log which proxy class is being used :)


minrk:
Perfect, thanks!

PR #2785, opened 2019-10-16
chown jupyterhub dir in user home

rpwagner:
When using the default spawner with `internal_ssl` enabled, if the destination directory for the user certs is created, only the cert directory is currently set to the running user and group. This adds the parent directory to the `chown` steps.


minrk:
Thanks!

PR #2790, opened 2019-10-20
Allow admins to customize /user-redirect/ behavior

yuvipanda:
/user-redirect/ is used to help link to a particular url
in the logged in user's authenticated notebook. For example,
if I'm logged in as user 'yuvipanda' and hit the URL
/hub/user-redirect/git-pull, it'll redirect me to
/user/yuvipanda/git-pull. This is extremely useful in
connecting hub links to notebook server extensions, such
as nbgitpuller.

Admins might want to customize how this redirection is done -
for example, redirect users to different running servers
based on the nbgitpuller repository they are linking from.
Adding a hook here helps accomplish that.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1264,15 +1264,15 @@ class JupyterHub(Application):
         help="""
         Callable to affect behavior of /user-redirect/
 
-        A callable that receives two parameters - the authenticated user and
-        the tornado request object for the current request passed to
-        `/user-redirect/<path>`, and returns a URL to redirect the user to.
-        This can be used to customize how the `user-redirect` URL works.
+        Receives 3 parameters:
+        1. path - URL path that was provided after /user-redirect/
+        2. request - A Tornado HTTPServerRequest object representing the
+           current request.
+        3. user - The currently authenticated user.
+        4. app - The JupyterHub object
 
-        Could be async or not.
-
-        To get the default behavior of /user-redirect/ leave this property unset
-        or return None from your callable.
+        It should return the new URL to redirect to, or None to preserve
+        current behavior.
         """,
     ).tag(config=True)
 
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1269,7 +1269,8 @@ class JupyterHub(Application):
         2. request - A Tornado HTTPServerRequest object representing the
            current request.
         3. user - The currently authenticated user.
-        4. app - The JupyterHub object
+        4. base_url - The base_url of the current hub, to allow for relative
+                      redirects
 
         It should return the new URL to redirect to, or None to preserve
         current behavior.
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -76,7 +76,7 @@ from .oauth.provider import make_provider
 from ._data import DATA_FILES_PATH
 from .log import CoroutineLogFormatter, log_request
 from .proxy import Proxy, ConfigurableHTTPProxy
-from .traitlets import URLPrefix, Command, EntryPointType
+from .traitlets import URLPrefix, Command, EntryPointType, Callable
 from .utils import (
     maybe_future,
     url_path_join,
@@ -1258,6 +1258,21 @@ class JupyterHub(Application):
         """
     ).tag(config=True)
 
+    user_redirect_hook = Callable(
+        None,
+        allow_none=True,
+        help="""
+        Callable to affect behavior of /user-redirect/
+
+        A callable that receives two parameters - a handler and the path passed to
+        `/user-redirect/<path>`, and returns a URL to redirect the user to. This
+        can be used to customize how the `user-redirect` URL works.
+
+        To get the default behavior of /user-redirect/ leave this property unset
+        or return None from your callable.
+        """
+    ).tag(config=True)
+
     def init_handlers(self):
         h = []
         # load handlers from the authenticator
@@ -2144,6 +2159,7 @@ class JupyterHub(Application):
             trusted_alt_names=self.trusted_alt_names,
             shutdown_on_logout=self.shutdown_on_logout,
             eventlog=self.eventlog,
+            app=self
         )
         # allow configured settings to have priority
         settings.update(self.tornado_settings)
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1264,7 +1264,7 @@ class JupyterHub(Application):
         help="""
         Callable to affect behavior of /user-redirect/
 
-        Receives 3 parameters:
+        Receives 4 parameters:
         1. path - URL path that was provided after /user-redirect/
         2. request - A Tornado HTTPServerRequest object representing the
            current request.
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1264,9 +1264,12 @@ class JupyterHub(Application):
         help="""
         Callable to affect behavior of /user-redirect/
 
-        A callable that receives two parameters - a handler and the path passed to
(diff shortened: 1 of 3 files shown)


yuvipanda:
This is to tie into nbgitpuller. My hope is to do be able to launch different repos in different environments determined by the repo URL in the nbgitpuller URL. This goes through url-redirect, and so more control there would be great.

jupyterhub/handlers/base.py
```diff
@@ -1470,20 +1474,31 @@ class UserRedirectHandler(BaseHandler):
...
-        if self.request.query:
-            user_url = url_concat(user_url, parse_qsl(self.request.query))
+    async def get(self, path):
+        # If hook is present to generate URL to redirect to, use that instead
+        # of the default. The configurer is responsible for making sure this
+        # URL is right. If None is returned by the hook, we do our normal
+        # processing
+        url = None
+        if self.app.user_redirect_hook:
+            url = await self.app.user_redirect_hook(self, path)
```
minrk: Do you think it's worth having `user` as an explicit argument to the hook? Rather than relying on the hook knowing the ins and outs of our BaseHandler API (which is a bit of a mess that I'd like to minimize exposing as public)? Further, what about being even more conservative and *only* passing `user, path` without exposing the Handler at all? What would be missing there that's useful? (first possible example: access to Authenticator instance)

jupyterhub/app.py
```diff
@@ -2144,6 +2159,7 @@ def init_tornado_settings(self):
             trusted_alt_names=self.trusted_alt_names,
             shutdown_on_logout=self.shutdown_on_logout,
             eventlog=self.eventlog,
+            app=self
```
minrk: `pre-commit run --all-files` to appease the autoformatter gods

jupyterhub/handlers/base.py
minrk: This `await` makes the assumption that the hooks are always async. Can add `utils.maybe_future` to allow for non-async, or add to the configurable's docstring that hooks must be async.

jupyterhub/tests/test_pages.py
```diff
@@ -398,6 +398,36 @@
...
+    assert query == urlencode(
+        {'next': ujoin(app.hub.base_url, '/user-redirect/redirect-to-terminal')}
+    )
+
+    # We don't actually want to start the server by going through spawn - just want to make sure
+    # the redirect is to the right place
+    r = await get_page('/user-redirect/redirect-to-terminal', app, cookies=cookies, allow_redirects=False)
+    r.raise_for_status()
+    redirected_url = urlparse(r.headers['Location'])
+    assert redirected_url.path == ujoin(app.base_url, f'/user/{username}/terminals/1')
```
consideRatio: Nice testing!

jupyterhub/app.py
```diff
@@ -1258,6 +1258,21 @@ def add_url_prefix(prefix, handlers):
...
+        Callable to affect behavior of /user-redirect/
+
+        A callable that receives two parameters - a handler and the path passed to
+        `/user-redirect/<path>`, and returns a URL to redirect the user to. This
+        can be used to customize how the `user-redirect` URL works.
+
+        To get the default behavior of /user-redirect/ leave this property unset
+        or return None from your callable.
+        """
+    ).tag(config=True)
```
consideRatio: Will `path` include the query parameters as well? I become uncertain, because of the terminology when parsing an url gives you `.path` and `.query` for example.

```
    path = urlparse(r.url).path
    assert path == ujoin(app.base_url, '/hub/login')
    query = urlparse(r.url).query
```

jupyterhub/handlers/base.py
yuvipanda: Good idea! I'm explicitly passing in user and the request object (Rather than just path), since you might wanna get access to the query string, etc. It does mean you need to deal with base_url and what not yourself - maybe we should pass 'app' as well?

jupyterhub/app.py
yuvipanda: Done I think

jupyterhub/handlers/base.py
yuvipanda: Done!

jupyterhub/tests/test_pages.py
yuvipanda: Thanks :D

jupyterhub/app.py
yuvipanda: Yep, Path didn't contain query parameters. I'm just passing in the request object directly now, although I might have to pass in the handler too.

yuvipanda:
@minrk I'm now passing:

1. path
2. request object
3. user object
4. app object

(4) feels a bit of a cheat, but for most redirects you need stuff like base_url and friends. If you think we should just pass some other kinda object instead I'm happy to work on that too.

In the long run, I wanna explore pluggy for writing hooks that can be versioned in backwards compatible ways. I use it in TLJH and love it.

minrk:
👍 to exploring pluggy for hooks.

The only thing I'm sketchy about is this implicitly making the JupyterHub app object a public API, which has never been true before, and that's largely on purpose. It certainly is a convenient handle for base_url, authenticator, etc., but it's also a handle on lots more, and I don't super want to be forced to make that API public.

base_url could be avoided if redirect-url is assumed to be *relative* to base_url, i.e. 

```python
url = hook()
if not (url + '/').startswith(base_url):
    # enforce redirect within base_url. Redirect outside JupyterHub is not allowed.
    url = url_path_join(base_url, url)
```

what do you think?

yuvipanda:
@minrk I just updated to pass base_url rather than app. I'd like to continue allowing redirects outside of base_url - for example, to preserve some nbgitpuller links even if they are now sent to a different hub. What do you think?

jupyterhub/app.py
```diff
@@ -1258,6 +1258,25 @@ def add_url_prefix(prefix, handlers):
         """
     ).tag(config=True)
 
+    user_redirect_hook = Callable(
+        None,
+        allow_none=True,
+        help="""
+        Callable to affect behavior of /user-redirect/
+
+        Receives 3 parameters:
```
minrk:
```suggestion
        Receives 4 parameters:
```

yuvipanda:
@minrk done!

I'll try run black manually again. I tried and failed to have pre-committ work again yesterday :(

yuvipanda:
@minrk done!

jupyterhub/app.py
```diff
@@ -1258,6 +1258,23 @@ def add_url_prefix(prefix, handlers):
...
+        Callable to affect behavior of /user-redirect/
+
+        Receives 4 parameters:
+        1. path - URL path that was provided after /user-redirect/
+        2. request - A Tornado HTTPServerRequest representing the current request.
+        3. user - The currently authenticated user.
+        4. base_url - The base_url of the current hub, for relative redirects
+
+        It should return the new URL to redirect to, or None to preserve
+        current behavior.
```
consideRatio: :heart: for a nice help string

yuvipanda:
@minrk not passing 'app' or 'handler' means we don't actually get access even the logger...

I think the 'right' thing to do here is some form of very simple dependency injection, where we can give the callback exactly the params it wants, like py.test's fixtures. Probably not something that should block this PR though.

minrk:
Great!

I agree that we need to figure out a better way to pass what things need without making everything accept `**kwargs` but also ideally without handing over the keys to the whole private-api shebang. That's a future plan. I hadn't thought of doing it py.test-style. That's a very interesting idea! I'd be really interested in seeing what that would look like.

PR #2794, opened 2019-10-25
cull_idle_servers.py: rebind max_age and inactive_limit locally

rkdarst:
- In the cull script, the max_age and inactive_limit are used from the
  outer scope.  In the case that you add extra logic, one may want to
  modify these values.
- In that case, you either have to rename them locally, or access the
  outer scope with "nonlocal", the first of which is too much work,
  the second of which has a high chance of introducing bugs (as it did
  for me).
- This change introduces a fix for everyone.  It doesn't change basic
  functionality, but makes local modifications simpler.

### DIFF
--- a/examples/cull-idle/cull_idle_servers.py
+++ b/examples/cull-idle/cull_idle_servers.py
@@ -117,7 +117,7 @@ def cull_idle(
     futures = []
 
     @coroutine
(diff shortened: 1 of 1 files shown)


minrk:
Makes sense

PR #2795, opened 2019-10-26
The proxy's REST API listens on port `8001`

bnuhero:
By default, the proxy's REST API listens on port `8001` instead of `8081` and the hub service listens on port `8081`.


minrk:
Thanks!

Issue #1585, opened 2017-12-11
Prometheus metrics to add to JupyterHub


yuvipanda:
We started adding prometheus metrics to JupyterHub in #1581.

More metrics to add:

1. Server poll times (Histogram)
2. Currently active servers (Gauge)
3. Proxy route poll times (Histogram)
4. Hub startup time (Histogram)
5. Server stop times (Histogram)
6. Proxy route deletion times (Histogram)
7. Server aliveness latency (Histogram) (time from server starting to being reachable)
8. Total users count (Gauge)

This should be a great start, and I think we can / should make a quick release after this.

yuvipanda:
Clearly did not work on these that week :) Would be great if someone can pick this one up!

yuvipanda:
I want to get some of this into 0.9, will set aside some time for it!

vpavlin:
Hi, this is super interesting issue:). We are running JupyterHub on OpenShift and we'd love to get some metrics from it - as we already use Prometheus to scrape OpenShift, Spark and other services, this would be an awesome addition.

Is there any update on this?

rkdarst:
More suggestions:
- Server age (histogram)
- Server time since last activity (histogram)
- proxy time since last activity (histogram) (likely not that important since the interesting data is in the services)
- services active (gauge)
- servers active (gauge) (maybe already done?)

For KubeSpawner, my own script provides some count of pods by profile, which is computed based on pod name.  This may be hard to do in a general way, but ideally could be adapted to other spawners, too (profilespawner).

The histograms are a bit difficult since they are instantaneous counts, while prometheus histograms are usually envisioned as a bunch of accumulating gauges over time as far as I can tell.  Would be good for someone to study what the proper technique here is.

dtaniwaki:
I decided to use [statsd-exporter](https://github.com/prometheus/statsd_exporter) because those values are not implemented for Prometheus although I know it's not ideal. I'm happy to implement this feature if no one is handling it.

clkao:
@dtaniwaki did you manage to start adding metrics?

dtaniwaki:
@clkao I haven’t started it yet because I didn’t have much time and statsd-exporter works fine.

clkao:
perhaps we should start with identifying desired stats, as some of them are already implemented with statsd, and some are implemented as prometheus in binderhub. Then we can see how best to unify them.

rajat404:
Hi!
[...]

PR #2799, opened 2019-10-30
Add prometheus metric to measure hub startup time

rajat404:
This PR exposes `hub_startup_duration_seconds` prometheus metric, partially completing #1585

Kindly review.


jupyterhub/app.py
```diff
@@ -2265,6 +2267,9 @@ def _log_cls(name, cls):
         self.init_tornado_settings()
         self.init_handlers()
         self.init_tornado_application()
+        HUB_STARTUP_DURATION_SECONDS.observe(
+            time.perf_counter() - hub_startup_start_time
+        )
```
GeorgianaElena: I believe it would be useful to have ``HUB_STARTUP_DURATION_SECONDS`` measure how much it takes to both *initialize* and *start* all the Hub components (including spawners).

jupyterhub/app.py
```diff
@@ -2265,6 +2267,9 @@ def _log_cls(name, cls):
...
 
         # init_spawners can take a while
```
GeorgianaElena: Since ``init_spawners`` can take a while and because we're already logging its duration I believe it would be nice to also expose this duration as a prometheus metric (*INIT_SPAWNERS_DURATION_SECONDS*). This way, if hub startup takes long, we can check INIT_SPAWNERS_DURATION_SECONDS and see if this is the reason. What do you think?

GeorgianaElena:
This is great, thanks @rajat404 :confetti_ball:. Just left a couple of comments inline.

jupyterhub/app.py
GeorgianaElena: Because spawners initialization is part of hub initialization, I think separating them might cause a bit of confusion :thinking:.
Also, what do you think of having *HUB_STARTUP_DURATION_SECONDS* reflect not only the initialization time, but also the entire time it takes to have the Hub up and running, i.e. the sum of init time and start time?

jupyterhub/app.py
rajat404: Done

jupyterhub/app.py
rajat404: My apologies for the late response.
I've exposed `INIT_SPAWNERS_DURATION_SECONDS` metric as you suggested.

Regarding `HUB_STARTUP_DURATION_SECONDS` reflecting time taken for Hub to be up and running - I don't think I have enough context/knowledge to make that call. Can you please share what's normally considered as `Hub Startup` - the initialization, or Hub running?

If it seems confusing, perhaps we can rename/split the metric:
* HUB_INITIALIZATION_DURATION_SECONDS - Initialization time
* HUB_STARTUP_DURATION_SECONDS - Initialization & Hub up and running time

Please suggest which approach to take.

GeorgianaElena: Since this has proven not to be very straightforward, maybe @betatim, can help us decide. @betatim, what do you think, **hub startup time** should be hub initialization time or hub init + hub start time?

minrk:
Thanks @rajat404 and @GeorgianaElena for the review!

rajat404:
@minrk While I'm grateful for the PR merge, I think we still have some open questions on this PR.
Could you perhaps give some inputs on the above thread, and/or check that the current strategy in this PR is correct 🙏 

Just want to ensure that we don't add code with incorrect conventions to the release 😅

minrk:
Sorry for missing the conversation, thanks for working through it @rajat404 and @GeorgianaElena! I think what's best is to have just two metrics:

1. HUB_STARTUP_DURATION_SECONDS, which measures the time from process start to responding to HTTP requests (i.e. the total startup time), which is the most relevant number for answering questions like "how much downtime should I expect from a hub restart?" #2854 does this.
2. INIT_SPAWNERS_DURATION_SECONDS, which measures the spawner initialization, the most variable and longest-running process, which is all set in this PR.

I don't believe we need a third metric for the total Hub initialization time, short of the total hub startup time.

GeorgianaElena:
Thanks for clearing this up @minrk and thanks @rajat404 for adding more metrics to JH :tulip:

PR #2803, opened 2019-10-31
adding docs preview to circleci

choldgraf:
This adds a step to build and preview the documentation in CircleCI. It is a little bit hard for me to test, because I still can't get the jupyterhub docs to build, I keep running into NPM errors :-(


choldgraf:
Phew, got it working - the dependency installation was tricky to figure out in circleci, but some combination of `sudo`-ing the right lines got it to work. For reference, here's the previewed docs:

https://1561-20785986-gh.circle-artifacts.com/0/html/index.html

minrk:
Awesome!

PR #2811, opened 2019-11-08
chore: Update python versions in travis matrix

jgwerner:
- Added python 3.8 to build matrix to support docs, which specify support for python 3.5+
- Run tests with 3.8 only when release tag is not null

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,7 +75,6 @@ after_failure:
       echo "or after-the-fact on already committed files with"
       echo "    pre-commit run --all-files"
     fi
-
 matrix:
   fast_finish: true
   include:
@@ -95,6 +94,7 @@ matrix:
         # password in url is url-encoded (urllib.parse.quote($PGPASSWORD, safe=''))
         - JUPYTERHUB_TEST_DB_URL=postgresql://jupyterhub:hub%5Btest%2F%3A%3F@127.0.0.1/jupyterhub
     - python: 3.7
+      dist: xenial
     - python: 3.8
       if: tag IS present
   allow_failures:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,15 @@
 language: python
+dist:
+  - bionic
+  - xenial
 sudo: false
 cache:
   - pip
 python:
-  - 3.6
   - 3.5
+  - 3.6
+  - 3.7
+  - 3.8
   - nightly
 env:
   global:
@@ -95,6 +100,6 @@ matrix:
         # password in url is url-encoded (urllib.parse.quote($PGPASSWORD, safe=''))
         - JUPYTERHUB_TEST_DB_URL=postgresql://jupyterhub:hub%5Btest%2F%3A%3F@127.0.0.1/jupyterhub
     - python: 3.7
-      dist: xenial
+    - python: 3.8
   allow_failures:
     - python: nightly
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,10 @@
 language: python
-dist:
-  - bionic
-  - xenial
 sudo: false
 cache:
   - pip
 python:
-  - 3.5
   - 3.6
-  - 3.7
-  - 3.8
+  - 3.5
   - nightly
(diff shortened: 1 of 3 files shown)


betatim:
Could you explain a bit what the benefit of adding these additional options is or what current shortcoming it improves on? There are pros and cons to more testing.

For example testing more versions of Python takes more slots in travis (slower builds for this project and all other projects in this GitHub org) and as long as we don't expect dependencies on the Python version I think testing the minimum supported and a current version should do the job. Tests also run against a nightly build of Python which should provide amble warning time if new features break something in JupyterHub. This means it isn't immediately clear to me what we gain from testing more versions.

jgwerner:
@betatim thanks for taking the time to look at this PR! I don't see any big benefits to adding more complexity to the matrix, but rather make the test matrix reflect what is stated in the docs. In my opinion if the docs say >=3.5 then those versions should be tested. I could send another PR just clarifying that JupyterHub _may_ work with `3.7` and `3.8` but that it is not formally tested before merging to master. Testing with nightly builds _may_ provide ample warning with new versions of Python, however, in some cases the nightly build will test with several versions ahead of what is currently tested in the matrix. For example, right now there are tests for Python 3.6 and nightly is 3.9 dev alpha 0, so 3.7 and 3.8 are skipped altogether.

Regarding Ubuntu, the differences between `xenial` and `bionic` also _may_ warrant additional tests.  I added tests for `bionic` since the Dockerfile in this repo uses `bionic` is the source image and left `xenial` because we didn't know if you wanted to keep using that version when testing with Travis.

manics:
@jgwerner is it possible to run the extra tests on tags only? That would fulfill the requirement for documented supported versions to be tested without much impact on Travis builds for the rest of the Github organisation.

betatim:
I like the idea of running things only on tagging/releases.

---

In my experience Python doesn't tend to break things suddenly. This means that things that work in 3.6 will also work in 3.7 and 3.8, etc. The cost of explicitly testing this is high compared to the benefit. Things do get deprecated but the timeline for reacting to these changes is measured in year(s).

The `Dockerfile` in the root of the repo is IMHO provided as a way for people to get started using JupyterHub but isn't meant as a production ready artefact or a place to start from that incorporates best practices. So I'd stick with the [a Linux/Unix based system](https://github.com/jupyterhub/jupyterhub#check-prerequisites) requirement which is intentionally vague. The difference in the different systems where a hub could run (and if it is broken it is considered a bug) are much bigger than the differences between Ubuntu releases. This means the value of testing on two systems that could be considered "the same" compared to the possible supported systems is outweighed by the cost.

The costs of extending the build matrix are
* longer time for the builds to complete (slower feedback),
* using CI slots that other packages could use,
* more maintenance time purely to keep the CI working,
* maintainer toil in the form of clicking "restart build"

jgwerner:
@manics sure we could help with that, I’ll send a new commit within the next couple of days.

jgwerner:
@manics @betatim have you ever considered using tox?

manics:
I've used it in other projects, I've got a love/hate relationship with tox!

I've just tried out [Travis build stages in another project](https://github.com/jupyterhub/oauthenticator/pull/301). It looks like adding conditional jobs is really easy. @betatim what do you think?

jgwerner:
@manics I added an update to test for `tag` using [Travis conditionals](https://docs.travis-ci.com/user/conditions-v1). Do you envision running the additional tests for all tags or just when the tag has a SemVer compatible value?

.travis.yml
```diff
@@ -95,6 +95,7 @@ matrix:
         # password in url is url-encoded (urllib.parse.quote($PGPASSWORD, safe=''))
         - JUPYTERHUB_TEST_DB_URL=postgresql://jupyterhub:hub%5Btest%2F%3A%3F@127.0.0.1/jupyterhub
     - python: 3.7
-      dist: xenial
+    - python: 3.8
```
betatim:
```suggestion
      dist: xenial
    - python: 3.8
```

PR #2812, opened 2019-11-08
docs: Update docs to run tests

jgwerner:
Unlike `pytest-tornado`, `pytest-asyncio` does not support the `async-test-timeout` option:

- Remove async env var from travis config
- Update docs to remove `--async-test-timeout` flag from test instructions


minrk:
Thanks!

PR #2820, opened 2019-11-15
updating to pandas docs theme

choldgraf:
This updates the documentation theme to `pandas_sphinx_theme`. It does *not* rearrange any pages, because we've already done a great job of organizing much of the documentation for this repo :-)

So the difference should only be cosmetic, not structural. One thing I did do was re-organize a few of the `toctree` pieces so that we have fewer "top-level" sections in the docs. Let's take a look at how this looks in the built docs to see if we like it.

Current docs: https://jupyterhub.readthedocs.io/en/latest/
New docs: https://1626-20785986-gh.circle-artifacts.com/0/html/index.html


choldgraf:
ok, I think this is ready to go (Travis might need to be fixed). What do folks think?

consideRatio:
@choldgraf - stealing your attention here! Can you make ma RTD maintainer on kubespawner?

https://readthedocs.org/projects/jupyterhub-kubespawner/builds/9980877/

choldgraf:
I just added descriptions to each index page, as well as including EOLs to try and make travis happy.

@consideRatio : done!

consideRatio:
The documentation looks great! This LGTM!

choldgraf:
Does anybody know what this error in travis (under the "lint" job) means?

`Fix requirements.txt.....................................................Failed`

consideRatio:
https://github.com/jupyterhub/jupyterhub/blob/master/.pre-commit-config.yaml, it is about:

"requirements-txt-fixer - Sorts entries in requirements.txt and removes incorrect entry for pkg-resources==0.0.0"

consideRatio:
https://pypi.org/project/pre-commit-hooks/

choldgraf:
I saw that there, but I wasn't sure what the problem was. Does it mean that we can't pin to a non-versioned dependency for the docs?

consideRatio:
I don't know, I'd try to install this pre-commit fix thing locally to run it locally and see what comes of it. I failed to install it last time i tried =/

choldgraf:
yeah I still haven't figure out how to install the pre-commit thing locally either :-(

choldgraf:
ok I think figured out both how to install the pre-commit thing (I guess it was doing some weird stuff with WSL) and also fixed the error that was popping up

betatim:
I like the new layout.

Two things before we merge:
* @minrk and @willingc you should give your 👍 before we do this
* my usual question: which external links do we break with this move?

choldgraf:
I agree - I am happy to hold off merging to make sure others are happy with the changes

I don't believe that any pages have moved here, so there shouldn't be any broken links (only new ones). As an aside, a quick way to get a list of moved files is to search for the "`→`" character in the "files changed" page - that should find all the files that were renamed or moved

choldgraf:
Rebased to fix the merge conflict, and so we can preview our nifty Institutional FAQ with the new theme :-)

https://1663-20785986-gh.circle-artifacts.com/0/html/getting-started/institutional-faq.html

willingc:
@choldgraf Use your judgement. I'm swamped at the moment. I'm sure it's an improvement. Go for it.

choldgraf:
IMO these are good to go if folks agree that the general theme structure is good, and I'm happy to available to spot-check things that we missed etc if something comes up.

minrk:
Yay!

PR #2824, opened 2019-11-17
Raise error on missing specified config

### LINKED ISSUE #2819: Using --config with a missing config should raise an error
```shell
# this cause no error
jupyterhub --config /asdfasdf
```

I want the above to cause an error! If it doesn't, it cause various delayed errors. As an example in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1463, a failure to find the passed --config led to a harder to debug error. In this case, the `jupyterhub/k8s-hub` image that doesn't ship with configurable-http-proxy, but it doesn't need it either assuming the config file is passed and found.

---

I inspected the current setup, and conclude that there is a default value on a `config_file` Traitlet set to `"jupyterhub_config.py"`, so it may not be obvious if it has been set explicitly or not which is required information for us.

One idea is to try to spot if it has been explicitly set or if we rely on the default value using traitlets know-how, and then run a check very early to see if it exists or not, and error out early if it doesn't. No matter what, we should catch a provided configuration file that doesn't exist somehow!

## Relevant parts of the code base

Where I think --config translates to the JupyterHub class traitlet config_file:
https://github.com/jupyterhub/jupyterhub/blob/703af1dd1eddb266e5fd80774e3baa413728be0c/jupyterhub/app.py#L100-L105

Where the config_file traitlet is defined along with a default value:
https://github.com/jupyterhub/jupyterhub/blob/703af1dd1eddb266e5fd80774e3baa413728be0c/jupyterhub/app.py#L310-L312

Where the config_file is first used:
https://github.com/jupyterhub/jupyterhub/blob/703af1dd1eddb266e5fd80774e3baa413728be0c/jupyterhub/app.py#L2207-L2213

consideRatio:
Closes #2819 by exiting JupyterHub directly with an error if a config
file has been specified for the config_file traitlet, for example
through the -f or --config flag, but isn't available on the file system.

### DIFF
--- a/jupyterhub/tests/test_app.py
+++ b/jupyterhub/tests/test_app.py
@@ -3,13 +3,13 @@ import binascii
 import os
 import re
 import sys
+import time
 from subprocess import check_output
 from subprocess import PIPE
 from subprocess import Popen
 from subprocess import run
 from tempfile import NamedTemporaryFile
 from tempfile import TemporaryDirectory
-import time
 from unittest.mock import patch
 
 import pytest
--- a/jupyterhub/tests/test_app.py
+++ b/jupyterhub/tests/test_app.py
@@ -6,8 +6,10 @@ import sys
 from subprocess import check_output
 from subprocess import PIPE
 from subprocess import Popen
+from subprocess import run
 from tempfile import NamedTemporaryFile
 from tempfile import TemporaryDirectory
+import time
 from unittest.mock import patch
 
 import pytest
@@ -37,6 +39,24 @@ def test_token_app():
             f.write("c.Authenticator.admin_users={'user'}")
         out = check_output(cmd + ['user'], cwd=td).decode('utf8', 'replace').strip()
     assert re.match(r'^[a-z0-9]+$', out)
+
+
+def test_raise_error_on_missing_specified_config():
+    """
+    Using the -f or --config flag when starting JupyterHub should require the
+    file to be found and exit if it isn't.
+    """
+    process = Popen(
+        [sys.executable, '-m', 'jupyterhub', '--config', 'not-available.py']
+    )
+    # wait inpatiently for the process to exit
+    for i in range(100):
+        time.sleep(0.1)
+        returncode = process.poll()
+        if returncode is not None:
+            break
+    process.kill()
+    assert returncode == 1
 
 
 def test_generate_config():
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -57,6 +57,7 @@ from traitlets import (
     Float,
     observe,
     default,
+    validate,
 )
 from traitlets.config import Application, Configurable, catch_config_error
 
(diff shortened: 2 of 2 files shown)


jupyterhub/tests/test_app.py
```diff
@@ -39,6 +41,24 @@ def test_token_app():
...
+def test_raise_error_on_missing_specified_config():
+    """
+    Using the -f or --config flag when starting JupyterHub should require the
+    file to be found and exit if it isn't.
+    """
+    process = Popen(
+        [sys.executable, '-m', 'jupyterhub', '--config', 'not-available.py']
+    )
+    # wait inpatiently for the process to exit
+    for i in range(100):
```
betatim: Can we use something like https://docs.python.org/3/library/subprocess.html#subprocess.run which will give us the return code of the process without having to poll like this. My guess is that you poll and then kill the process in case it doesn't error out and keeps running. If this guess is right maybe we can add a comment about this so that people from the future also realise this.

betatim:
One comment about the test and one question about logging, but otherwise looks good to me.

Is there any precedence in JupyterHub for exiting if an invalid config is found? Using `print` directly seems like that should be replaced with `logging` but I'd look at how the rest of JupyterHub handles this, then copy what is done there.

jupyterhub/tests/test_app.py
consideRatio: I didn't find a timeout flag or similar for the process created using run(), so I got stuck there on test-failure scenarios where the process actually just start up.

> Run the command described by args. Wait for command to complete, then return a CompletedProcess instance.

So this was the path I choose in the end, after trying to use .run() first.

consideRatio:
> Is there any precedence in JupyterHub for exiting if an invalid config is found? Using print directly seems like that should be replaced with logging but I'd look at how the rest of JupyterHub handles this, then copy what is done there.

I found a place where it happened and this example is copied from there, that is for example how I picked up to output to the stderr output. I found it by searching for sys.exit in the jupyterhub source code. Here it is:

https://github.com/jupyterhub/jupyterhub/blob/9fdab027daa32c9017845572ad9d5ba1722dbc53/setup.py#L20-L24

Thiis is from setup.py though, so the practice of using logging/print may be different.

Ah but here is another one in app.py.

https://github.com/jupyterhub/jupyterhub/blob/703af1dd1eddb266e5fd80774e3baa413728be0c/jupyterhub/app.py#L193-L195

jupyterhub/tests/test_app.py
betatim: What do you think of adding a comment here to say exactly what you just said "this looks like it should not be needed (see `subprocess.run()`) but because we want a timeout we do it ourselves"? (In a new Pr as I am merging this one now)

consideRatio: Ah yepp it is a very relevant comment since it answers a actual question raised! I created #2826

PR #2826, opened 2019-11-18
Add inline comment to test

consideRatio:


jupyterhub/tests/test_app.py
```diff
@@ -46,16 +46,20 @@ def test_raise_error_on_missing_specified_config():
...
-    # wait inpatiently for the process to exit
+    # wait inpatiently for the process to exit like we want it to
     for i in range(100):
         time.sleep(0.1)
         returncode = process.poll()
         if returncode is not None:
             break
-    process.kill()
+    else:
+        process.kill()
```
betatim: For my education: looking at this again I wondered what the exit code for a killed JupyterHub would be. I think it is not `1` and not `0`. So this test should fail when the process gets killed.

consideRatio: The test is using the variable `returncode` set to `process.poll()`, this is only cleanup if it never exited during the 10 seconds we waited for it to exit. The else clause made the cleanup happen only if the process never exited, but I found that `process.kill()` can be run no matter what without erroring itself. I don't know either what `kill()` will default to btw.

PR #2828, opened 2019-11-22
add block for scripts included in head

bitnik:
This adds blocks for script files included in head. I think this is useful when one customises the templates, as I work on customised home page right now.


minrk:
Makes sense, thanks!

PR #2830, opened 2019-11-24
Changelog for 1.1.0

consideRatio:
# TODO and help wanted

- [x] Write a release summary to complement the list of merged PRs.

---

I've looked through all the merged PRs since 1.0.0. I have done so by using a list of them created by using @choldgraf's [`github-activity`](https://github.com/choldgraf/github-activity) command line tool.

```
# gave me list of PRs since latest git tag (1.0.0):
github-activity --kind pr -o prs-since-1.0.0.md jupyterhub/jupyterhub

# in the future, when we want to augment with new PRs:
github-activity --kind pr --since 92173c605 -o prs-since-last.md jupyterhub/jupyterhub
```

I sorted all the PRs in three categories for now, _Added, Fixes, and Other_. I sorted PRs relating to documentation, CI, UI, refactoring to the _Others_ category. For anything within _Fixes_ and _Added_, I did not just use the PR title but instead tried to translate it into an entry that can make sense for a JupyterHub user. For example, sometimes technical "what" are described as compared to the outcome of the PR, and the outcome is often more relevant in a changelog - not the technical change itself.

## Review
1. Summaries of Added / Fixed PRs - are they correct and comprehensible enough?
2. What structure should we have in the changelog? Consider some past changelog structures. Should we include all merged PRs categorized like this or not? Should we also add a highlight summary?
   I don't have a strong opinion, but this is what I consider:
   - I like having PR links associated with changes
   - I don't mind having a long list of PRs if they are grouped in a way allowing me to trace down PRs relating to new features and and fixed issues apart from the rest.
   - I think as the author of a changelog, working through a list of PRs is a suitable way to reach the required overview to summarize the release as a whole.

## Help
Go ahead and add commits and commit suggestions!


choldgraf:
Looking through the list of changes there I quite like the structure, thanks for adding the extra context to each line. I'm +1 on "added", "fixed", and "other". What I use for this on Jupyter Book is to associate a tag with each ("enhancement" and "bug") and then have `github-activity` auto-sort for me. Maybe something to consider for the future?

minrk:
Thanks for getting this started! I personally like having two layers: First, a manual, curated description of the changes most likely to be relevant to users, i.e. new feature highlights and especially breaking changes. When there's a comprehensive list of lots of PRs, ~nobody is going to read them all, so it's hard to be sure that folks will be able to see what they need rather than being overwhelmed. It's fine for them to be there, but they don't solve the "I'm upgrading JupyterHub, what do I need to know?" question.

minrk:
I made a pass on what's here, and I don't feel the need to add anything further. I'd be happy to make a 1.1 beta with this changelog as-is (perhaps after #2840)

consideRatio:
Fully agree on the need for an additional summary alongside these lists!

choldgraf:
I agree with an additional summary - I think that the signal-to-noise of the changelog could be improved by tagging PRs as enhancement/bug/etc and then letting `github-activity` separate those out (e.g. most of the "major features" are probably enhancements). That said, I'm a fan of writing a short description of the highlights, followed by the machine-generated changelog.

minrk:
I added an additional summary and updated with the latest PRs. I think we can push this out (with #2854, which finishes an already-merged PR) if anyone wants to have a look.

consideRatio:
This LGTM!!

We have not updated the date in the title, and perhaps a beta release is made before 1.1.0, but none of this is problematic in my mind.

PR #2840, opened 2019-12-03
simplify Dockerfile

minrk:
- use apt to get Python, nodejs
- use two-stage builds for wheels so build dependencies aren't added to the runtime image
- shrinks image by ~70% from ~900MB to ~270
- pin CHP version

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,8 +22,8 @@
 # from your docker directory.
 
 # https://github.com/tianon/docker-brew-ubuntu-core/commit/3c462555392cb188830b7c91e29311b5fad90cfe
-ARG BASE_CONTAINER=ubuntu:bionic-20190612@sha256:9b1702dcfe32c873a770a32cfd306dd7fc1c4fd134adfb783db68defc8894b3c
-FROM $BASE_CONTAINER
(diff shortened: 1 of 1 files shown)


consideRatio:
From what I can tell, this is good.

I'm not 100% regarding the use of CMD over ENTRYPOINT with regards to making sure that we don't end up discarding SIGTERM signals etc.

I'm not confident about this though, and knows a bit too little about how the docker image could be used for this to be something I end up with a clear suggestion about.

Of relevance to such considerations:
- https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes
- https://github.com/Yelp/dumb-init and from the first opinion that I googled my way to it seems favored over the competing option `tini`, see: http://jkamenik.github.io/blog/2017/10/21/docker-details---dumb-init/

minrk:
Thanks! I don't think one needs to override ENTRYPOINT in general, except in special circumstances. It's inconvenient in a lot of ways to do so (especially with `docker run` and `docker exec`, so I'd avoid it unless it solves a specific problem.

I did double check with the image and SIGTERM is handled correctly by the hub without a special entrypoint.

Issue #2366, opened 2018-12-21
Failed spawns are still added to the proxy


hhuuggoo:
**Describe the bug**

If a spawn fails for a named server (for example, due to a timeout), The dead server will still be added to proxy and considered `ready` 

I believe the issue is that the notion of `ready` inside jupyterhub is based on:
https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/spawner.py#L87
which defines it as not being `pending` and having a `server`.  but jupyterhub assigns servers before we even start spawning.
https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/user.py#L410
And if the spawn fails, it's no longer "pending" therefore it becomes ready as soon as it has failed.  Then we go through looking for missing routes and we add them based on this ready flag, periodically.
https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/proxy.py#L322

**To Reproduce**
- Start jupyterhub (I'm using the kubespawner jupyterhub_config)
- set the spawner timeout to be 1 second 
- create a named server
- the server will be added to the page, but that server will not work.

**Expected behavior**
-  failed spawns should not show up and be represented as valiid server

**Desktop (please complete the following information):**

hhuuggoo:
fwiw I believe this issue impacts 0.9.4 (even without named servers).  You do get the error message in the UI that the server failed to start, but if you navigate to home, there are links for your server and stopping it.  The named server UI makes this problem worse because you don't get the visual indication that the spawn failed

consideRatio:
I think it might also, because I recall noticing such issue, and I have not used named servers yet.

consideRatio:
I'm quite confident servers that have not been responding correctly by the hub after spawn won't be get a route added for them in current jupyterhub.

PR #2849, opened 2019-12-12
Cleanup if spawner stop fails

gabber12:
If spawner.stop throws an exception `spawner.server` is not set to None and _stop_pending is set to False leading to redirect to a url bound to fail.


gabber12:
Might relate to #2366 too

minrk:
Good catch, thanks!

PR #2854, opened 2019-12-18
HUB_STARTUP_DURATION_SECONDS measures total startup time

minrk:
which is the main metric of interest for measuring disruptions due to restarts

finishes #2799

cc  @rajat404 @GeorgianaElena


GeorgianaElena:
Thanks @minrk :confetti_ball:

minrk:
Thanks!

PR #2857, opened 2019-12-18
Kernel connection does not work with tornado>=6.0

hoba87:


minrk:
Can you include `pip list` or `conda list`? Tornado 6 is used regularly with jupyterhub and the notebook, so it's likely another package that's out-of-date, such as ipykernel or pyzmq.

Closing this PR as we definitely don't want to pin down tornado in jupyterhub, but feel free to continue troubleshooting your installation here.

hoba87:
Ok thanks, for me probably an outdated ipython (6.2.1) was the issue.

Issue #2852, opened 2019-12-16
Jupyterhub Docker not able to find any jinja templates


nayan1247:
I wanted to run jupyterhub on a docker and used the below -
docker run -p 8000:8000 -d --name jupyterhub jupyterhub/jupyterhub jupyterhub
Unable to find the login/error jinja templates

Tried the below code as well still the same -

FROM jupyterhub/jupyterhub:master

RUN apt-get update --fix-missing && apt-get -y install \
    wget git rsync openssh-server build-essential gcc openjdk-8-jdk-headless && \
    apt install -y vim vim-tiny && \
    rm -rf /var/lib/apt/lists/*

COPY requirements.txt /tmp/requirements.txt

RUN pip install -r /tmp/requirements.txt

RUN echo "127.0.0.1 $(hostname) localhost" > /etc/hosts
RUN echo "echo \"127.0.0.1 $(hostname) localhost\" > /etc/hosts" >> /etc/profile

CMD jupyterhub --ip 0.0.0.0 --port 8000 --Spawner.default_url='/lab'

Gives the following error 
<img width="1129" alt="image" src="https://user-images.githubusercontent.com/3715878/70894088-540f2400-2012-11ea-94c9-5f6b27485b3c.png">

steinad:
I'm seeing this as well. If I set`c.JupyterHub.data_files_path = '/usr/local/share/jupyterhub'` in the config file, I get a login page, but little/no styling.

steinad:
I was able to get around this by specifying `FROM jupyterhub/jupyterhub:1.0.0` in my dockerfile. Using `jupyterhub/jupyterhub:1.0` pulls version `1.0.1dev`, which makes it likely that that version is where the issue is. That being said, I would not expect to get a dev version when specifying `1.0`.

nayan1247:
Thanks @steinad.

I overcame by uninstalling jupyterhub (1.0.1dev) and again installing jupyterhub 1.0.0.
This solved the issue for me and its working normally.
Agree that we shouldn't be seeing a dev version of jupyterhub when mentioning 1.0 or master.

Dmitry1987:
lol same issue, spent half day trying to figure out what's wrong ....

bigbosst:
Same issue here, updated 1.0 got the dev version with missing jinja, added the jinja pathing like @steinad and got the all the missing CSS issues. Switching to 1.0.0 got the stable version and that went back to working like expected.

[...]

PR #2864, opened 2019-12-27
Dockerfile: Copy share/ to the final image

rkdarst:
- When the Dockerfile was turned into a multi-stage build, it seems
  the share/ directory was not copied to the final image.  This
  resulted in certain components (static/components/, static/css/)
  being missing, which resulted in the JupyterHub share directory not
  being findable (in jupyterhub/_data.py).  This led to all kinds of
  weird havoc, like templates not being findable (#2852).
- I am still unsure if this is the right fix, please check this well.
- Closes: #2852

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -89,6 +89,7 @@ RUN npm install -g configurable-http-proxy@^4.2.0 \
 
 # install the wheels we built in the first stage
 COPY --from=builder /src/jupyterhub/wheelhouse /tmp/wheelhouse
+COPY --from=builder /src/jupyterhub/share /src/jupyterhub/share
 RUN python3 -m pip install --no-cache /tmp/wheelhouse/*
 
 RUN mkdir -p /srv/jupyterhub/


rkdarst:
Good point about not needing it in the builder (I guess, I have no idea).  Copying it later also invalidates caches less often (and later caches).

Force-pushed an update... hm, it fails ().  My guess is that share/templates is needed to build, since templates *were* included in the wheel.  So, maybe I'll just revert back to the old way, I guess it's not worth separately copying static/ and templates/ to the image... any better method?

rkdarst:
I just noticed that this copies to /src/jupyterhub/share in the final image... it should probably go into /usr/local/share/jupyterhub/ instead; I'm not sure how this is expected to work.  Could this be behind the continuing problem in #2852?

... but this makes no sense why it works when I built locally!  I checked my reflog and I never did anything else.

[limited time and i/o capacity, the above is not checked thoroughly]

consideRatio:
@rkdarst I went back and tried the current master tag and everything works now! I must have used an outdated master tag from or similar! Thanks for your help fixing this!

rkdarst:
On Sat, Dec 28, 2019 at 05:40:29AM -0800, Erik Sundell wrote:
> @rkdarst I went back and tried the current master tag and everything works now! I must have used an outdated master tag from or similar! Thanks for your help fixing this!

Oh... I thought the problem was in the dockerhub version somehow.
Well, glad it worked out!

rkdarst:
I've been thinking about this more, and I'm less sure this was the right answer.  Can someone else take a look, and if this is well-founded perhaps make another issue about it?

 * Why aren't the things included in the wheel?
* How does this PR get them to the right place in the final image?

consideRatio:
Regarding first point, i dont know and i dont understand what is desired or/and expected.

Regarding second, there are three file systems or areas where file cans reside:
- the build context, which is the local file system and specified in the docker build command
- the build stage's file area
- the final stage's file area
I recall from memory (on mobile atm) that we copy files to the build stage area from the build context, and later copy those to the final stage from the build stage.

diurnalist:
@rkdarst see #3082 for a different fix, I think it undid some of this PR.

PR #2868, opened 2019-12-28
Fixup .travis.yml

### LINKED ISSUE #2862: .travis.yaml cleanup - testing both xenial/bionic and python3.8 on tagged commits
I made the 1.1.0b1 release of this repository, but the build for Python 3.8 specifically for the tagged commit with the only difference being in _version.py.

In https://github.com/jupyterhub/jupyterhub/commit/861a7c5c5e98dbff4f3f71a4d79dd6c743bea907, where I go "back to dev" and the version 1.1.0b2.dev everything is fine again.

Here is the build for Python 3.8 that fails on 1.1.0b1:
https://travis-ci.org/jupyterhub/jupyterhub/jobs/629709780

Here is a build for Python 3.8 that succeeds for 1.1.0b2.dev... Oh... No apparently there is no Python 3.8 build attempted. Then the issue is probably a bigger issue of Python 3.8 compatibility where what I note is simply that we differentiate tests for tagged commits and untagged commits.

The build that does not contain Python 3.8, as usual: https://travis-ci.org/jupyterhub/jupyterhub/builds/629709523

## Action points
- Clean up the [.travis.yml](.travis.yml) file:
  - [ ] remove tests of both xenial and bionic, use only bionic
  - [ ] systematically test for Python 3.8 in both tagged commits and untagged commits. If it works, don't allow it to fail, and if it doesn't, then allow it to fail

## Some traces of changes (oldest first)
https://github.com/jupyterhub/jupyterhub/commit/c944c0e54a1d5b5aa326b11e83a20b37305b8337
https://github.com/jupyterhub/jupyterhub/commit/1dd09094a532f3bce0f6b2523e03cb43fea29f75
https://github.com/jupyterhub/jupyterhub/commit/731faf29c8a505c5052c12714906612f2841c8df

consideRatio:
Closes #2862 that I ran into while releasing the JupyterHub beta and made a tagged commit that Travis built with failures. I ended up doing quite a bit of experimentation and testing until I understood things and got them to work.

## Key findings
- bionic + Python versions less than 3.7 run into issues, see #2870 
- we don't support Python 3.8, see #2869 

## PR change highlights
- We now test with bionic and python 3.8 but allow them to fail
- I removed the `make html` build test as CircleCI tests that already
- I removed plain Py35 and Py36 tests and integrated Py35 and Py36 testing with other test cases
- I added names that describe the tests somewhat instead of relying on environment variables to help us figure it out in a too technical manner, the downside is that it hides some information about the environment variables that you need to inspect the build logs in order to see them. I consider it worth it anyhow as the jobs suddenly are or at least can be made more understandable to a broad audience now.
- I removed some deprecated travis logic (`sudo: false`)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 dist: bionic
 language: python
-sudo: false
 cache:
   - pip
 python:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,18 @@
-dist: bionic
+dist: xenial
 language: python
 cache:
   - pip
-python:
-  # These lines will yield additional jobs, and they will be separated out as
-  # allowed to fail in the Travis web user interface as compared to if we would
-  # add them in the matrix.include below instead.
-  - &allowed_failure_1 3.8
-  - &allowed_failure_2 nightly
 env:
   global:
     - MYSQL_HOST=127.0.0.1
     - MYSQL_TCP_PORT=13306
+
+# request additional services for the jobs to access
 services:
   - postgresql
   - docker
 
-# installing dependencies
+# install dependencies for running pytest (but not linting)
 before_install:
   - set -e
   - nvm install 6; nvm use 6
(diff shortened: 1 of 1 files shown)


consideRatio:
## Detected issues
Apparently we have issues between ubuntu 16.04 and ubutnu 18.04 on TravisCI with internal_ssl tests. [Here is a suite of tests done for bionic as default but with one test configuration for xenial](https://travis-ci.org/jupyterhub/jupyterhub/builds/630344816) that works. The only bionic job that works is the one with Python 3.7 which is the default of bionic. So perhaps it fails on Bionic whenever we change the python version?

It fails on the test:
```
test_connection_notebook_wrong_certs
```

consideRatio:
As I only modify .travis.yml and this allows for some issues to debugged better in referenced issues specifically if someone wants to help resolve them, I'll go ahead and merge this.

Issue #2865, opened 2019-12-28
login_url is hardcode


louis-she:
**Describe the bug**

The `login_url` is hardcode since this commit: https://github.com/jupyterhub/jupyterhub/commit/6b22f80ead47790798707c6e73a8cd7dc4ad865b#diff-a9c49d54d1416c0a6f0a70bc047bb72dL1178-R1180
Before that, this is configurable via `Authenticator`.

**Expected behavior**
`login_url` being configurable

minrk:
Closed folowing rationale in #2872 - this is already configurable via `Authenticator.auto_login = True`

PR #2872, opened 2020-01-01
Make login_url configurable

louis-she:
issue: https://github.com/jupyterhub/jupyterhub/issues/2865
blame: https://github.com/jupyterhub/jupyterhub/commit/6b22f80ead47790798707c6e73a8cd7dc4ad865b#diff-a9c49d54d1416c0a6f0a70bc047bb72dL1178-R1180


minrk:
It may be a bit confusing, but this is deliberate, and the behavior you want should result from setting `Authenticator.auto_login = True`, which causes the fixed and not configurable `/login` url to redirect to the Authenticator's login url. Reverting this change would be effectively setting `auto_login` to True always, and removing *that* configuration option.

Issue #2875, opened 2020-01-02
bug in auth.py


narnish:
Function 'normalize_username' of class 'PAMAuthenticator'
does not return 'username' if pam_normalize_username is True.

```
def normalize_username(self, username):
        """Round-trip the username to normalize it with PAM
        PAM can accept multiple usernames as the same user, normalize them."""
        if self.pam_normalize_username:
            import pwd

            uid = pwd.getpwnam(username).pw_uid
            username = pwd.getpwuid(uid).pw_name
            username = self.username_map.get(username, username)
            **return username**
        else:
            return super().normalize_username(username)

```

rkdarst:
Thanks for reporting this!  It was caused by my last change to the PR
that made this, combined with the fact that we didn't know how to test
the PAM case.  I guess it just goes to show that yes, there really is
nothing too small to test...

Fix in #2876.

PR #2876, opened 2020-01-02
Bugfix: pam_normalize_username didn't return username

rkdarst:
- A trivial bug caused by my last change to #2397 - made possible by
  the fact we didn't have a way to reliable test PAM stuff.
- Thanks to @narnish for noticing.
- Closes: #2875


consideRatio:
:heart:

PR #2881, opened 2020-01-09
trigger auth_state_hook prior to options form, add auth_state to template namespace

### LINKED ISSUE #2552: Getting auth_state to the spawner

### LINKED ISSUE #2860: auth_state hook fires too late; it should go before options form generation
**Describe the bug**

The placement of the auth_state hook in Spawner is later than it should be.  The auth_state is known when the options form is generated, and you might want to generate a different options form depending on the stuff in auth_state, e.g. users in different groups might get offered different container sizes or images.

Currently I'm having to make get_options_form a coroutine so it can call `yield self.user.get_auth_state()` and that seems silly, because all the stuff I am doing there is really stuff that belongs in the auth_state hook.

**To Reproduce**

Uh, I don't know an easier way to do it than to make an options form that depends on user auth_state, and then show that if you set that in auth_state_hook it isn't present in the options form.

**Compute Information**
 - Centos 7
 - JupyterHub master (as of 20 Dec. 2019)

minrk:
- allow auth_state_hook to be async
- trigger it prior to start and options_form serving, rather than on visit to home page

closes #2860
closes #2552
closes #2885

### DIFF
--- a/jupyterhub/handlers/pages.py
+++ b/jupyterhub/handlers/pages.py
@@ -67,8 +67,10 @@ class HomeHandler(BaseHandler):
         else:
             url = url_path_join(self.hub.base_url, 'spawn', user.escaped_name)
 
+        auth_state = await user.get_auth_state()
         html = self.render_template(
             'home.html',
+            auth_state=auth_state,
             user=user,
             url=url,
             allow_named_servers=self.allow_named_servers,
@@ -91,10 +93,12 @@ class SpawnHandler(BaseHandler):
 
     default_url = None
 
-    def _render_form(self, for_user, spawner_options_form, message=''):
+    async def _render_form(self, for_user, spawner_options_form, message=''):
+        auth_state = await for_user.get_auth_state()
         return self.render_template(
             'spawn.html',
             for_user=for_user,
+            auth_state=auth_state,
             spawner_options_form=spawner_options_form,
             error_message=message,
             url=self.request.uri,
@@ -305,6 +309,8 @@ class SpawnPendingHandler(BaseHandler):
         # if spawning fails for any reason, point users to /hub/home to retry
         self.extra_error_html = self.spawn_home_error
 
+        auth_state = await user.get_auth_state()
+
         # First, check for previous failure.
         if (
             not spawner.active
@@ -324,6 +330,7 @@ class SpawnPendingHandler(BaseHandler):
             html = self.render_template(
                 "not_running.html",
                 user=user,
+                auth_state=auth_state,
                 server_name=server_name,
                 spawn_url=spawn_url,
                 failed=True,
@@ -345,7 +352,11 @@ class SpawnPendingHandler(BaseHandler):
             else:
                 page = "spawn_pending.html"
             html = self.render_template(
-                page, user=user, spawner=spawner, progress_url=spawner._progress_url
+                page,
+                user=user,
+                spawner=spawner,
+                progress_url=spawner._progress_url,
+                auth_state=auth_state,
             )
             self.finish(html)
             return
@@ -370,6 +381,7 @@ class SpawnPendingHandler(BaseHandler):
             html = self.render_template(
                 "not_running.html",
                 user=user,
+                auth_state=auth_state,
                 server_name=server_name,
                 spawn_url=spawn_url,
             )
@@ -389,7 +401,7 @@ class AdminHandler(BaseHandler):
 
     @web.authenticated
     @admin_only
-    def get(self):
+    async def get(self):
         available = {'name', 'admin', 'running', 'last_activity'}
         default_sort = ['admin', 'name']
         mapping = {'running': orm.Spawner.server_id}
@@ -438,9 +450,11 @@ class AdminHandler(BaseHandler):
         for u in users:
             running.extend(s for s in u.spawners.values() if s.active)
 
+        auth_state = await self.current_user.get_auth_state()
         html = self.render_template(
             'admin.html',
             current_user=self.current_user,
+            auth_state=auth_state,
             admin_access=self.settings.get('admin_access', False),
             users=users,
             running=running,
@@ -456,7 +470,7 @@ class TokenPageHandler(BaseHandler):
     """Handler for page requesting new API tokens"""
 
     @web.authenticated
-    def get(self):
+    async def get(self):
         never = datetime(1900, 1, 1)
 
         user = self.current_user
@@ -525,8 +539,12 @@ class TokenPageHandler(BaseHandler):
 
         oauth_clients = sorted(oauth_clients, key=sort_key, reverse=True)
 
+        auth_state = await self.current_user.get_auth_state()
         html = self.render_template(
-            'token.html', api_tokens=api_tokens, oauth_clients=oauth_clients
+            'token.html',
+            api_tokens=api_tokens,
+            oauth_clients=oauth_clients,
+            auth_state=auth_state,
         )
         self.finish(html)
 
--- a/jupyterhub/handlers/pages.py
+++ b/jupyterhub/handlers/pages.py
@@ -186,7 +186,7 @@ class SpawnHandler(BaseHandler):
         spawner_options_form = await spawner.get_options_form()
         if spawner_options_form:
             self.log.debug("Serving options form for %s", spawner._log_name)
-            form = self._render_form(
+            form = await self._render_form(
                 for_user=user, spawner_options_form=spawner_options_form
             )
             self.finish(form)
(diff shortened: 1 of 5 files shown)


jupyterhub/handlers/pages.py
```diff
@@ -67,9 +67,6 @@ class HomeHandler(BaseHandler):
         else:
             url = url_path_join(self.hub.base_url, 'spawn', user.escaped_name)
 
-        auth_state = await user.get_auth_state()
-        user.spawner.run_auth_state_hook(auth_state)
-
```
consideRatio: Removing this will make us unable to use a users auth_state in templates to render /hub/home, right?

Should we perhaps have this remain here as well? Here is a render of /hub/home from my jupyterhub for example, and it could make sense in my mind to write out something one has stored in auth_state here.

image

rcthomas:
Haven't been able to verify this change myself.  For our named-server + wrapspawner set up I have set in our config `c.JupyterHub.default_url = '/hub/home'` so unless a user has done something like bookmark their server they are presented with the list of servers they could go to or start up.

I was reviewing the code change and I see that the auth state hook fires before `start()` but I am not 100% whether my setup would hit it as part of the wrapspawner flow.

It's our "new year" (allocation rollover and deploying multiple other new things at once) and I am guessing I'll need a few days before I can test.

jupyterhub/handlers/pages.py
minrk: I don't think so. Only the *spawner* shouldn't be able to take actions based on the auth state, which shouldn't happen on the home page, I think.

rcthomas: Is there another/right way to get the auth state information to the template?

minrk:
It sounds like there are two separate concerns going on here:

1. there are use cases for taking custom actions with auth_state immediately, and
2. there are use cases for delivering auth_state to spawners, e.g. setting environment, etc.

What has been implemented is 2. (by defining it as a method on the Spawner), but it is called in a manner that more indicates 1. as the target use case (on load of the home page, not necessarily prior to spawn, and possible to bypass with spawn links). I think 1. is better handled by a general -post-authentication hook, and shouldn't have to do with Spawners.

So I see two options:

a) move Spawner auth hook calls to the right places (this PR)
b) define an Authenticator hook to make it easier to take generic non-spawner-specific actions on auth_state, e.g. for use on the home page

We can do *both*, if both are really needed, or if the general hook is really the only use case people are expressing a need for, we can move the hook that's been added off of Spawner and onto Authenticator, and call it in more general places.

rcthomas:
Hi @minrk I think you have summarized it correctly.  

I was originally looking for a customizable way to associate some information with the user.  I noticed `auth_state` as the only thing even close, and even looked back through the history of how that evolved.  It had some features I didn't need, like the encryption, and there wasn't a way to get `auth_state` to where I needed it in the template, except by attaching it to the spawner.

That is, not without doing something else like extending `user` to have a way to hold that kind of arbitrary user info.  I looked at that but didn't pursue it because it would touch a lot more code.  But it seems like attaching that to the `user` and even having a way to refresh it on page loads or something might be useful.

minrk:
@rcthomas I think I see that you are using it here:

[link](https://github.com/NERSC/jupyterhub-deploy/blob/30420bfda19c543762d1ef7a15e0c6bdc30e7b51/jupyter-nersc/web-jupyterhub/templates/home.html#L18-L27)

```html
<thead><tr>
    <th></th>
    {% for setup in user.spawner.setups -%}
        {% for arch in setup.architectures if user.spawner.check_roles(arch.roles) %}
            <th class="text-center">{{ arch.description }}</th>
        {% endfor %}
    {%- endfor %}
</tr></thead>
```

And I see that you have extra nontrivial [logic defined on the Spawner](https://github.com/NERSC/jupyterhub-deploy/blob/30420bfda19c543762d1ef7a15e0c6bdc30e7b51/jupyter-nersc/web-jupyterhub/nerscspawner.py#L33-L65) used in the template, and it's been helpful that the Spawner is a customizable object accessible as `user.spawner` in the page template.

But as you described, your 'real' goal is to make it easier to deliver [extra data acquired at authentication time](https://github.com/NERSC/jupyterhub-deploy/blob/30420bfda19c543762d1ef7a15e0c6bdc30e7b51/jupyter-nersc/web-jupyterhub/jupyterhub_config.py#L1179-L1188) to the page template.


Key sources of information and where they are used

1. setups, architectures, profiles, from [jupyterhub_config.py]()
  - used in [spawning]()
  - used in [page template]()
2. 'userdata' from iris, [fetched at auth time]()
  - used in [spawning]()
  - used in [page template]()

Defining all of these on the spawner gives you a consolidated handle on everything, which is certainly convenient. But perhaps the key problem is that where you have stored the 'userdata' (auth_state) is not available at page render time (which this PR currently breaks).

My inclination is to make a more direct fix for this: ensure auth_state is directly available in the template namespace.

This would require a slight change to your code, because rather than relying on `self.userdata` being defined (which you solved by invoking spawner.auth_state_hook prior to page render) in e.g. `user_allocations`, you would *pass* auth_state down through `check_roles(role, auth_state) -> user_allocations(auth_state)` in your calls in the template.

minrk:
@rcthomas I've updated the PR to add `auth_state` to the page template for authenticated pages, so I think this should be solvable now with some updated templates, passing auth_state as an argument where you are currently using `self.userdata` in methods called in the templates.

consideRatio:
## Issue from Travis
```
[E 59:47.909 MockHub web:1788] Uncaught exception GET /@/space%20word/hub/spawn/testuser-39/myserver (127.0.0.1)
    HTTPServerRequest(protocol='http', host='localhost.jovyan.org:8000', method='GET', uri='/@/space%20word/hub/spawn/testuser-39/myserver', version='HTTP/1.1', remote_ip='127.0.0.1')
      File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/tornado/web.py", line 1699, in _execute
        result = await result
      File "/home/travis/build/jupyterhub/jupyterhub/jupyterhub/handlers/pages.py", line 192, in get
        self.finish(form)
      File "/home/travis/build/jupyterhub/jupyterhub/jupyterhub/handlers/base.py", line 174, in finish
        super().finish(*args, **kwargs)
      File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/tornado/web.py", line 1122, in finish
        self.write(chunk)
      File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/tornado/web.py", line 834, in write
        raise TypeError(message)
    TypeError: write() only accepts bytes, unicode, and dict objects
```

## About this code

```python
        if spawner_options_form:
            self.log.debug("Serving options form for %s", spawner._log_name)
            form = self._render_form(
                for_user=user, spawner_options_form=spawner_options_form
            )
            self.finish(form)
```

## Fix
I think the key is that we now have an async _render_form function that needs to be awaited someplace.

minrk:
@consideRatio yup, good catch. Fixed

consideRatio:
This LGTM! Thorough work on this @minrk!

minrk:
Thanks!

rcthomas:
Oh shoot, if I'd known you'd be coming over I'd have cleaned up my code a bit first!  Hope it wasn't too hard to track down, next time maybe I'll just point at my code.  Thanks for looking into it so deeply!

This looks good to me, and I'll make changes to my templates and config and update if there are any issues though I don't anticipate any.  This is much more like what I was hoping for and a lot better than my "well, I guess I need to attach it to the spawner" hack.

athornton:
This sounds very promising.  I'll give it a look when I can.

rcthomas:
Just to close the loop and document, I've been able to adjust my template and spawner to work with the `auth_state` passed in from the handler.  So far this is working properly and is a bit less convoluted than what was happening before.  Thanks!

PR #2882, opened 2020-01-09
LocalProcessSpawner should work on windows by using psutil.pid_exists

ociule:
Hi there!

I'm taking advantage of the policy you've stated previously that you accept patches that make windows implementations like WinLocalProcessSpawner possible, without looking to add win-specific code to jupyterhub.

In this patch, I want to allow subclasses of LocalProcessSpawner  to have a working poll method without reimplementing the whole poll method, but just the call to _signal, which, as its comment says, is not expected to work on Windows.

After this MR, WinLocalProcessSpawner will implement its own win-specific version of the new _is_single_user_process_alive method, instead of the whole poll.

Cheers, hope this helps.

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -18,6 +18,7 @@ from tempfile import mkdtemp
 
 from async_generator import async_generator
 from async_generator import yield_
+import psutil
 from sqlalchemy import inspect
 from tornado.ioloop import PeriodicCallback
 from traitlets import Any
@@ -1468,21 +1469,12 @@ class LocalProcessSpawner(Spawner):
             self.clear_state()
             return 0
 
-        alive = await self._is_single_user_process_alive()
+        alive = psutil.pid_exists(self.pid)
         if not alive:
             self.clear_state()
             return 0
         else:
             return None
-
-    async def _is_single_user_process_alive(self):
-        """Returns True if the process still exists, False otherwise.
-        
-        Allows subclasses to implement this in a platform specific manner.
-        """
-        # send signal 0 to check if PID exists
-        # this doesn't work on Windows, but that's okay because we don't support Windows.
-        return await self._signal(0)
 
     async def _signal(self, sig):
         """Send given signal to a single-user server's process.
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -1468,14 +1468,21 @@ class LocalProcessSpawner(Spawner):
             self.clear_state()
             return 0
 
-        # send signal 0 to check if PID exists
-        # this doesn't work on Windows, but that's okay because we don't support Windows.
-        alive = await self._signal(0)
+        alive = await self._is_single_user_process_alive()
         if not alive:
             self.clear_state()
             return 0
         else:
             return None
+
+    async def _is_single_user_process_alive(self):
+        """Returns True if the process still exists, False otherwise.
+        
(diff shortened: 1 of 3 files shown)


jupyterhub/spawner.py
```diff
@@ -1468,15 +1468,22 @@ def get_env(self):
...
-        # this doesn't work on Windows, but that's okay because we don't support Windows.
-        alive = await self._signal(0)
+        alive = await self._is_single_user_process_alive()
         if not alive:
             self.clear_state()
             return 0
         else:
             return None
 
+    async def _is_single_user_process_alive(self):
```
manics: Since this is intended to be overridable it's part of the public API, so I think the leading `_` could be removed?

ociule: It's intended to be overridable by subclasses indeed. The way I see it's intended not to be called by outside code, so I think it's not part of the public API per se. Making it part of that is not needed for making WinLocalProcessSpawner work. But if you guys think it could be useful for other code, why not.

ociule: Of course, like a public API method, it has the constraint that once we merge it in and WinLocalProcessSpawner reimplements it, there's an involved refactoring process to change the name, for example. Even if it's not meant to be called by outside code.

ociule:
I have discovered that JH uses psutil. It's a much better idea to use psutil.pid_exists for this, which will work on all platforms and eliminates the need for _is_single_user_process_alive

jupyterhub/spawner.py
ociule: I have removed this method in my latest commit, it's not needed, pls see below.

ociule:
psutil was not in requirements.txt yet, even if it's clearly used already, see proxy.py#L690.

ociule:
Well, as the code in proxy.py and utils.py is written, psutil was a requirement only on windows.

So there are several ways to solve this:
 * keep it a requirement on win only, add it as such to requirements.txt (using `sys_platform == 'win32'`), poll should then use proxy.ConfigurableHTTPProxy._check_pid (which should be moved to utils) which uses os.kill(0, pid) on unix and psutil on win.
 * make it a requirement on unix too, use it everywhere, including in proxy.ConfigurableHTTPProxy._check_pid

ociule:
TODO : Should update https://github.com/jupyterhub/jupyterhub/blob/47cf1915ff916fe97f17d5e2930732834b14ea7f/docs/source/reference/spawners.md as needed.

requirements.txt
```diff
@@ -7,6 +7,7 @@ jupyter_telemetry
 oauthlib>=3.0
 pamela
 prometheus_client>=0.0.21
+psutil>=5.6.5
```
minrk: I'd prefer to avoid adding an unconditional dependency on a compiled extension that doesn't have non-Windows wheels that's only needed on Windows. Perhaps this could be a conditional dependency, only used on Windows?

minrk:
Happy to have this change, but let's keep psutil as Windows-only, not a strict dependency.

requirements.txt
ociule: All right, I'll do that.

ociule: FWIW, psutil is also an optional depency on unix, as it's used by utils.print_ps_info (without which it returns None) which is called by app.JupyterHub.log_status

minrk:
Thanks!

PR #2885, opened 2020-01-12
Run auth_state_hook on spawner before spawning

dpoon:
Authenticators may have a [hook function](https://github.com/jupyterhub/jupyterhub/pull/2555/commits) that passes `auth_state` information to the spawner.  It is therefore important to run such a hook in the `SpawnHandler` page handler before spawning.

It might also be a good idea to run `auth_state_hook` in the `SpawnPendingHandler` page handler as well.


minrk:
I think #2881 is meant to fix this one, can you check if it does?

PR #2887, opened 2020-01-14
Fix implementation of default server name

krinsman:
Having the modification to the default user-redirect behavior be contained inside of `UserRedirectHandler` seems more transparent to me. Also this worked as intended in more cases than the previous version when I tested it.

### DIFF
--- a/jupyterhub/tests/test_named_servers.py
+++ b/jupyterhub/tests/test_named_servers.py
@@ -267,3 +267,89 @@ async def test_named_server_spawn_form(app, username, named_servers):
     assert server_name in user.spawners
     spawner = user.spawners[server_name]
     spawner.user_options == {'energy': '938MeV', 'bounds': [-10, 10], 'notspecified': 5}
+
+
+async def test_user_redirect_default_server_name(app, username, default_server_name):
+    name = username
+    server_name = default_server_name
+    app.default_server_name = default_server_name
+    cookies = await app.login_user(name)
+
+    r = await api_request(app, 'users', username, 'servers', server_name, method='post')
+    r.raise_for_status()
+    assert r.status_code == 201
+    assert r.text == ''
+
+    r = await get_page('/user-redirect/tree/top/', app)
+    r.raise_for_status()
+    print(urlparse(r.url))
+    path = urlparse(r.url).path
+    assert path == ujoin(app.base_url, '/hub/login')
+    query = urlparse(r.url).query
+    assert query == urlencode(
+        {'next': ujoin(app.hub.base_url, '/user-redirect/tree/top/')}
+    )
+
+    r = await get_page('/user-redirect/notebooks/test.ipynb', app, cookies=cookies)
+    r.raise_for_status()
+    print(urlparse(r.url))
+    path = urlparse(r.url).path
+    while '/spawn-pending/' in path:
+        await asyncio.sleep(0.1)
+        r = await async_requests.get(r.url, cookies=cookies)
+        path = urlparse(r.url).path
+    assert path == ujoin(
+        app.base_url, '/user/{}/{}/notebooks/test.ipynb'.format(name, server_name)
+    )
+
+
+async def test_user_redirect_hook_default_server_name(
+    app, username, default_server_name
+):
+    """
+    Test proper behavior of user_redirect_hook when c.JupyterHub.default_server_name is set
+    """
+    name = username
+    server_name = default_server_name
+    app.default_server_name = default_server_name
+    cookies = await app.login_user(name)
+
+    r = await api_request(app, 'users', username, 'servers', server_name, method='post')
+    r.raise_for_status()
+    assert r.status_code == 201
+    assert r.text == ''
+
+    async def dummy_redirect(path, request, user, base_url):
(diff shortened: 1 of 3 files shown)


krinsman:
I'm not sure why it's failing only 3.8+dist:bionic, but obviously there's no sense in even considering merging this until the tests pass. Sorry about this.

minrk:
I think this makes sense. Could you add a test for a case that's handled differently after this change?

minrk:
I don't believe the test failure is related to this PR, I think that should have been fixed by #2889.

krinsman:
@minrk You're right, I agree, it's a good idea to add a test for this.

**Main question:** How does one change the configuration settings of the mock JupyterHub?

I tried to set `app.default_server_name` in the tests (where `app` is one of the arguments to the test). I doubt that would actually work, but I don't know either way since Travis didn't run the two tests I added which do that.

**How I attempted to add tests for this:**
Would it make sense to copy the tests [`test_user_redirect`](https://github.com/jupyterhub/jupyterhub/blob/6a872b371e6ae1de462813b7971a81b791e57942/jupyterhub/tests/test_pages.py#L384) and [`test_user_redirect_hook`](https://github.com/jupyterhub/jupyterhub/blob/6a872b371e6ae1de462813b7971a81b791e57942/jupyterhub/tests/test_pages.py#L409) from [test_pages.py](https://github.com/jupyterhub/jupyterhub/blob/6a872b371e6ae1de462813b7971a81b791e57942/jupyterhub/tests/test_pages.py) to [`test_named_servers.py`](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/tests/test_named_servers.py), and then modify the `assert` statements to check for the value of `c.JupyterHub.default_server_name` to be interposed between `username` and the rest of the URL? 

(I also copied some of the lines from [`test_create_named_server`](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/tests/test_named_servers.py#L86) to attempt to create the named server beforehand.)

This is extremely embarrassing to admit, but I don't actually know (yet) how to use the [unittest.mock](https://docs.python.org/3/library/unittest.mock.html) library (this seems to be what JupyterHub is using for tests?). The best I could do so far was cargo-cult programming unfortunately.

> I think this makes sense. Could you add a test for a case that's handled differently after this change?

To be honest this is handled exactly the same as the previous PR was supposed to be handled, it's just that the previous PR didn't work as intended (at least when I tested at NERSC).

This is supposed to work exactly [the way that `/user-redirect/...` currently does](https://github.com/jupyterhub/jupyterhub/blob/071e375d5fd038e70c29c619d199bf2c9c090d5c/docs/source/reference/urls.md#user-redirect), 

except that if `c.JupyterHub.default_server_name = 'test'`, for example, 

then the redirect for user `hortense` for the link `/user-redirect/notebooks/Index.ipynb` will be to `/user/hortense/test/notebooks/Index.ipynb` (instead of the default `/user/hortense/notebooks/Index.ipynb`).

It's up to the JupyterHub administrator to ensure that a server with the name of `'test'` (or whatever the value of `c.JupyterHub.default_server_name` is) can be spawned for every user though; this doesn't take care of that. I think it would be better if that feature was also included, but honestly I wouldn't even know where to begin to check for that.

minrk:
Thanks for writing the tests! I [defined the missing fixture](https://github.com/jupyterhub/jupyterhub/pull/2887/commits/8ee72dd80f7fbead6aa6f9b40f5e915d892d6e90) and imports. The only thing missing is that the user-redirect-hook test doesn't pass its assertion that when the user_redirect_hook only has `url_path_join(user.url, '/terminals/1')` that default_server_name is inserted. Now, given what is changed here, where user.url is only the user's base URL, that makes sense to me, but maybe isn't what you intend? I'd be okay with updating the test assertions to pass with the current behavior, or if you want to take a stab at fixing the behavior to what the test expects, that's fine, too (harder, I think, and perhaps more complicated than we should do).

krinsman:
@minrk Your reasoning sounds solid to me as well, updating the test assertions to pass with the current behavior should be fine. That way the current behavior is implicitly documented.

jupyterhub/tests/test_named_servers.py
```diff
@@ -267,3 +280,89 @@ def named_servers(app):
...
+    r = await get_page(
+        '/user-redirect/redirect-to-terminal',
+        app,
+        cookies=cookies,
+        allow_redirects=False,
+    )
+    r.raise_for_status()
+    redirected_url = urlparse(r.headers['Location'])
+    assert redirected_url.path == url_path_join(
+        app.base_url, 'user', username, server_name, 'terminals/1'
```
consideRatio: This change makes the test pass locally on my computer.

I figure that the user-redirect hook could transition user-redirect/redirect-to-terminal to user-redirect/terminals/1 as well, and then not interfere in that redirect by returning None.

```suggestion
        # the user_redirect_hook has the final say and
        # default_server_name wasn't considered there
        app.base_url, 'user', username, 'terminals/1'
```

minrk: exactly right, thanks!

PR #2901, opened 2020-01-17
fix: 'Non-ASCII character '\xc3'

jgwerner:
Code comments that have non-ASCII characters sometimes create encoding issues when not [adding encoding headers to the source files](https://www.python.org/dev/peps/pep-0263/).

- Change `naïve` to `naive` in all files. The `diaerisis` isn't [used in English that much anymore anyway](https://www.merriam-webster.com/words-at-play/mary-norris-diaeresis).
- Change `-` to `:` in the `metrics.py --> prometheus_log_method()` comments.


consideRatio:
Makes sense and LGTM!

PR #2905, opened 2020-01-21
Add what we need with some margin to Dockerfile's build stage

### LINKED ISSUE #2852: Jupyterhub Docker not able to find any jinja templates

consideRatio:
This fixes #2852 by adding a script referenced by package.json. But is this
enough? Should we perhaps look in MANIFEST.in and copy some more files
listed there?

This is all thanks to people coming together and helping out figuring
out the issue in https://github.com/jupyterhub/jupyterhub/issues/2852.
Thank you @shingo78 for spotting that we missed bower-lite and its role
and all others who reported and helped debug this!

---

I have never learned what a MANIFEST.in file does etc, but I think it relates to what is packaged in a PyPI package for example. So, then it may be a suitable source to align with when we copy things into the Dockerfile's build step where we package the JupyterHub package as well.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,14 +41,10 @@ RUN apt-get update \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/*
 
-# copy only what we need to avoid unnecessary rebuilds
-COPY bower-lite \
-     package.json \
-     pyproject.toml \
-     README.md \
-     requirements.txt \
-     setup.py \
-     /src/jupyterhub/
+# copy everything except whats in .dockerignore, its a
+# compromise between needing to rebuild and maintaining
+# what needs to be part of the build
+COPY . /src/jupyterhub/
 COPY jupyterhub/ /src/jupyterhub/jupyterhub
 COPY share/ /src/jupyterhub/share
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,7 +42,8 @@ RUN apt-get update \
  && rm -rf /var/lib/apt/lists/*
 
 # copy only what we need to avoid unnecessary rebuilds
-COPY package.json \
+COPY bower-lite \
+     package.json \
      pyproject.toml \
      README.md \
      requirements.txt \


consideRatio:
Ping @metonymic-smokey, here is another PR where comprehension is key and either research and learning or lots of past experience is needed. I'm confident I'm currently lacking relevant understanding and is quite dependent on review to solve this properly.

manics:
My understanding is `MANIFEST.in` should contain files that need to be distributed i.e. in the output of `python setup.py sdist|bdist_wheel` which will subsequently be `pip install`ed. If it's required to run the `python setup.py` or equivalent but *not* to `pip install` the final package then I don't think it should be in in `MANIFEST.in`. "equivalent" in this case also means `pip install .` since it's starting from the source.

I think your approach here of copying in the missing source files but not adding them to MANIFEST is correct, and I suspect you're also right that other files may be missing. Since it's a multi-stage Dockerfile it's probably fine to copy in everything for the first build stage since it doesn't take up space in the final image.

consideRatio:
The circle-ci build step running install_jupyterhub fails because the singleuser base image jupyter/base-notebook cannot find module pip any more since latest build of the base-notebook it seems.

https://circleci.com/gh/jupyterhub/jupyterhub/1860?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Im on a conference today, anyone wanting to fix this is very welcome to push commits or add code suggestions

consideRatio:
```
# works
docker run -it --rm jupyter/base-notebook:d4e60350af15 python3 -m pip

# a commit later, this fail
docker run -it --rm jupyter/base-notebook:140e7cc2370a python3 -m pip
```

What went wrong in 140e7cc

consideRatio:
I've narrowed it down to bumping JupyterLab from 1.2.4 to 1.2.5, that leads to a dependency of pip disappears.

consideRatio:
Related issue: https://github.com/jupyter/docker-stacks/issues/988

PR #2907, opened 2020-01-22
Fix --generate-config bug when specifying a filename

### LINKED ISSUE #2906: --generate-config fails since 1.1.0
**Describe the bug**

The function `--generate-config` doesn't work as expected with the release of 1.1.0. Jupyterhub complains about the missing config file, which should be generated. We execute the steps in a Packer build process, which worked until the upgrade of jupyterhub from 1.0.0 to 1.1.0.

The executed command is even listed in the -h explanation, so I'm pretty confident it should work:

```sh
[user@ip-1-1-1-1 ~]$ /opt/anaconda3/bin/jupyterhub -h
Start a multi-user Jupyter Notebook server
[..]

Examples
--------

    generate default config file:
    
        jupyterhub --generate-config -f /etc/jupyterhub/jupyterhub_config.py
    
[...]
```

Please don't hit me if conda or amazon linux may introduce this error, but I personally don't think so.

**To Reproduce**

Just execute jupyterhub with *--generate-config*. Installation has been done via conda. Execute the following commands on a clean linux/centos/amazonLinux1.

```sh
$ wget https://repo.continuum.io/archive/Anaconda3-2019.10-Linux-x86_64.sh -O /tmp/Anaconda3-Linux-x86_64.sh
$ sudo sh /tmp/Anaconda3-Linux-x86_64.sh -b -p /opt/anaconda3
$ sudo /opt/anaconda3/bin/conda config --add channels 'conda-forge'
$ sudo /opt/anaconda3/bin/conda install -y -c conda-forge jupyterhub",
$ sudo /opt/anaconda3/bin/jupyterhub --generate-config -f /etc/jupyterhub/jupyterhub_config.py
ERROR: Failed to find specified config file: /etc/jupyterhub/jupyterhub_config.py
```

**Expected behavior**

The config file should be generated. Failing with "Failed to find specified config file" doesn't make sense.

**Compute Information**

 - Installation is usually performed within a packer build, but problem can be reproduced manually.

```sh
[user@ip-1-1-1-1 ~]$ cat /etc/issue
Amazon Linux AMI release 2018.03
[user@ip-1-1-1-1 ~]$ sudo /opt/anaconda3/bin/jupyterhub --version
1.1.0
```

consideRatio:
This commit fixes #2906 that was introduced due to #2824. See analysis
of issue in https://github.com/jupyterhub/jupyterhub/issues/2906#issuecomment-577303510.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -316,7 +316,7 @@ class JupyterHub(Application):
 
     @validate("config_file")
     def _validate_config_file(self, proposal):
-        if not os.path.isfile(proposal.value):
+        if not self.generate_config and not os.path.isfile(proposal.value):
             print(
                 "ERROR: Failed to find specified config file: {}".format(
                     proposal.value


consideRatio:
The circleci issue is unrelated. It is discussed in #2905

jupyterhub/app.py
```diff
@@ -316,7 +316,7 @@ def _load_classes(self):
 
     @validate("config_file")
     def _validate_config_file(self, proposal):
-        if not os.path.isfile(proposal.value):
+        if not self.generate_config and not os.path.isfile(proposal.value):
```
consideRatio: I've considered if it would make sense to do something specific on generate_config and a file exist, but that is handled already by asking to overwrite or not at a later stage, so if generate_config is true, the file can both exist already and not.

consideRatio: So, this LGTMyself ;)

PR #2917, opened 2020-01-31
rest api: fix schema for remove parameter in rest api

minrk:
this should have been in the docs, but wasn't showing up because it didn't have the right structure

came up in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1558#issuecomment-579092135


consideRatio:
LGTM!

Issue #2926, opened 2020-02-10
Paginate users under /admin template


cbjuan:
**Describe the bug**
When using JupyterHub in an environment with a lot of users (example, 50K users), if I try to access the `/hub/admin` page, the website takes a long time to render every user. 

For example, when loading the URL `/hub/admin` with 50K users, the handler to serve the users for the admin view - https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/handlers/pages.py#L399-  takes about 7 seconds to retrieve the data and pass all the information to the template. Meanwhile, the rendering of the template https://github.com/jupyterhub/jupyterhub/blob/master/share/jupyterhub/templates/admin.html takes about 1 minute to complete the display of every element. Tests made using a local Postgres database.

Could it be possible to improve this behavior? Probably, paginating users or taking a similar approach could work.

**To Reproduce**

1. Have a high amount of users in JupyterHub database (my tests: 50,000)
2. After starting the JupyterHub (which takes a long time too), log in with an admin user and navigate to the admin web view (/hub/admin).
3. Wait until the webpage is fully rendered (~1 min). This is because the template renders each user individually in the table: `{% for user in users %}`.

**Expected behavior**

It would be great if admins of that kind of JupyterHubs don't need to wait so long to get the users. Probably, we can paginate users or make async rendering.

**Compute Information**
 - Operating System: OSX
 - JupyterHub Version [e.g. 22]: Last version available in GitHub https://github.com/jupyterhub/jupyterhub/tree/22f86ad76c437d37bc26c28d9fecfa5457ef5974

Thanks!

betatim:
Sounds like a worthwhile thing to work on. I think the tricky parts will be making a nice API that is backwards compatible and doesn't require lots of JS on the webpage. Do you want to make a suggestion/draft/first attempt as a PR?

cbjuan:
Sure, I'm already working on it.

betatim:
Very nice! For a change like this it probably makes sense to open a PR pretty early when you have a rough sketch. That way others can give feedback and input in design decisions. Otherwise we run the risk of having weeks of work that then gets completely changed during review.

cbjuan:
Ok, sounds good. I think early tomorrow I will have the PR open so we can begin the discussion. I'm currently polishing some details before submitting my approach  💅

cbjuan:
[...]

PR #2929, opened 2020-02-14
Adding pagination in the admin panel

cbjuan:
This WIP PR tries to introduce pagination for the users' table in the admin panel. 

Motivation: As described in #2926, when having a high number of users (for example, 50K), the webpage takes a long time to render the table for the users (minutes) and to allow the right interaction. 

Approach: Introduce pagination on that page to reduce the number of users to display and iterate over the list of users (default 100 users per page). Pagination is added via custom code taking as example this package: `python-paginate` package (https://pypi.org/project/python-paginate/ & https://github.com/lixxu/python-paginate). 

Here is an example of how the admin panel is displayed with pagination:
JupyterHub_paginate

As commented, the PR is WIP, so we can evaluate if the approach is the right one and continue working on it.

If you need the example database I used for this purpose, please, ping me.

Thanks!


cbjuan:
The error in Travis tests are related to a network error when installing `dev-requirements.txt` from the `autoformatting check` Travis job. I can't re-run that check to test it really fails.

manics:
I've restarted the failing job and its green now 🙂

cbjuan:
Yay! Thanks @manics

jupyterhub/handlers/pages.py
```diff
@@ -442,17 +454,34 @@ class AdminHandler(BaseHandler):
...
             running.extend(s for s in u.spawners.values() if s.active)
 
+        total = self.db.query(orm.User.id).count()
+        pagination = Pagination(
+            url=self.request.uri,
+            total=total,
+            record_name='users',
+            display_msg='Displaying {record_name} <b>{start} - {end}</b>. Total {record_name}: {total}',
+            page=page,
+            per_page=per_page,
```
minrk: I think we want

```python
css_framework="bootstrap3"
```

to be explicit. Looking at the code, this is the default, but it will be easier to find if we are explicit for when we bump to bootstrap 4.

share/jupyterhub/templates/admin.html
```diff
@@ -18,11 +18,12 @@
 {% block main %}
 
 <div class="container">
+  {{ pagination.info|safe }}
   <table class="table table-striped">
     <thead>
       <tr>
         {% block thead %}
-        {{ th("User (%i)" % users|length, 'name') }}
+        {{ th("User", 'name') }}
```
minrk: This can preserve the user count by adding it as its own field, rather than using `users | length`. We don't want to lose the info

jupyterhub/handlers/pages.py
```diff
@@ -442,17 +454,34 @@ class AdminHandler(BaseHandler):
...
+        )
+
         auth_state = await self.current_user.get_auth_state()
         html = self.render_template(
             'admin.html',
+            request=self.request,
```
minrk: request isn't used in the template, looks like?

jupyterhub/handlers/pages.py
```diff
@@ -442,17 +454,34 @@ class AdminHandler(BaseHandler):
         # get User.col.desc() order objects
         ordered = [getattr(c, o)() for c, o in zip(cols, orders)]
 
-        users = self.db.query(orm.User).outerjoin(orm.Spawner).order_by(*ordered)
+        users = (
+            self.db.query(orm.User)
+            .outerjoin(orm.Spawner)
+            .order_by(*ordered)
+            .limit(per_page)
+            .offset(offset)
```
minrk: 👍

betatim:
This PR looks like a good v1 of this feature. I think I'd also prefer not depending on a new library. Especially because the "<< < 1 2 3 ... 456 *457* 458 ... 500> >>" style pagers on paginated sites are a pet peeve of mine. Most of the time I don't know which page I want to go to to find in which case you have to click through all the page. Or I do happen to know (roughly) which page I have to go to but that page is usually hidden in the "...".

My ideal "big list of rows" UI is a small search field that allows (very basic) queries for the thing I know (in our case the username?) to narrow down the list of 101 pages to one or two pages.

I recently learned about "keyset pagination" (https://blog.jooq.org/2016/08/10/why-most-programmers-get-pagination-wrong/ and https://www.citusdata.com/blog/2016/03/30/five-ways-to-paginate/) so I am wondering if we should try to do something like that or leave it for a v2?

jupyterhub/handlers/pages.py
cbjuan: Ok, agree about being explicit

share/jupyterhub/templates/admin.html
cbjuan: In this case, I removed it because it's shown similarly in `{{ pagination.info|safe }}`. I can check it to be sure.

jupyterhub/handlers/pages.py
cbjuan: I think so. Will check it

cbjuan:
> This PR looks like a good v1 of this feature. I think I'd also prefer not depending on a new library. Especially because the "<< < 1 2 3 ... 456 _457_ 458 ... 500> >>" style pagers on paginated sites are a pet peeve of mine. Most of the time I don't know which page I want to go to to find in which case you have to click through all the page. Or I do happen to know (roughly) which page I have to go to but that page is usually hidden in the "...".
> 
> My ideal "big list of rows" UI is a small search field that allows (very basic) queries for the thing I know (in our case the username?) to narrow down the list of 101 pages to one or two pages.
> 
> I recently learned about "keyset pagination" (https://blog.jooq.org/2016/08/10/why-most-programmers-get-pagination-wrong/ and https://www.citusdata.com/blog/2016/03/30/five-ways-to-paginate/) so I am wondering if we should try to do something like that or leave it for a v2?
> 
> ps. I really like that you went with 100 as default page size. The usual of 10 or 20 is another thing that drives me mad :)

I agree absolutely on the search as the best feature for this page (even better than pagination). It should be the next step to do after having minimal pagination that let us having thousands of users and manage them in a performant way.

I also agree with both of you, @betatim and @minrk, it would be better if we don't introduce new dependencies, especially in a case like this where the library is not actively developed or widely used. I can try to abstract the features and implement them as @minrk suggested. 

I will dig into Keyset pagination, to check whether it can be a better option.


About how to proceed with these improvements, I can try to continue working on them. In any case, what do you think, can we go with this V1 of pagination and improve it in further PRs? Or is it worth to wait until we redo this?

minrk:
> About how to proceed with these improvements, I can try to continue working on them. In any case, what do you think, can we go with this V1 of pagination and improve it in further PRs? Or is it worth to wait until we redo this?

Let's use this version in terms of the requests and queries, but avoid adding the dependency. If you wanted to leave out the HTML nav and have only the query parameters in the API, that would be an okay first PR, as well.

cbjuan:
The last commit https://github.com/jupyterhub/jupyterhub/pull/2929/commits/7b6ac158ccf4cd3bc4ffafd4ba3681b0a8ddb3df removes the `python-paginate` package and enables a minimal API for pagination in the admin page (as it worked previously with the package), as it was suggested. 

I plan to try to find time to generate the proper UI to use.

jupyterhub/handlers/pages.py
```diff
@@ -591,6 +606,69 @@ def get(self, *args):
         self.finish()
 
 
+class Pagination(BaseHandler):
```
minrk: Pagination isn't a handler:

```suggestion
class Pagination:
```

minrk: let's also move this pagination class to its own top-level `jupyterhub/pagination.py` file. I think we'll likely want to use it on some API endpoints in the future like the user list.

jupyterhub/handlers/pages.py
```diff
@@ -591,6 +606,69 @@ def get(self, *args):
...
+        self._cached = {}
+        self.skip = (self.page - 1) * self.per_page
+        pages = divmod(self.total, self.per_page)
+        self.total_pages = pages[0] + 1 if pages[1] else pages[0]
+
+        self.has_prev = self.page > 1
+        self.has_next = self.page < self.total_pages
+
+    @classmethod
+    def get_page_args(self, handler):
```
minrk: let's document what this returns in the docstring. Looks like page, per_page, and offset?

Since it references `self`, I don't think this is a classmethod, I think it's a regular instance method.

jupyterhub/handlers/pages.py
```diff
@@ -591,6 +606,69 @@ def get(self, *args):
...
+    def __init__(self, *args, **kwargs):
+        """Potential parameters.
+        **url**: URL in request
+        **page**: current page in use
+        **per_page**: number of records to display in the page. By default 100
+        **total**: total records considered while paginating
+        **display_msg**: informative text for pagination
+        **record_name**: name of the record, showed in pagination info
+        """
+        self.page = kwargs.get(self._page_name, 1)
```
minrk: I don't think this needs to be so customizable as to have the name of the argument used to get the page as a class attribute.

I think defining args in init is probably best.

jupyterhub/handlers/pages.py
```diff
@@ -591,6 +606,69 @@ def get(self, *args):
...
+        'Displaying {record_name} <b>{start} - {end}</b>. Total {record_name}: {total}'
+    )
+
+    def __init__(self, *args, **kwargs):
+        """Potential parameters.
+        **url**: URL in request
+        **page**: current page in use
+        **per_page**: number of records to display in the page. By default 100
+        **total**: total records considered while paginating
+        **display_msg**: informative text for pagination
```
minrk: I'm not sure whether the display aspects (display_msg, record_name) belong in our Pagination class, since it doesn't do the rendering like the imported one did. What do you think? It might be simpler to put the text parts directly in the template, but I might be missing a benefit of having it here.

share/jupyterhub/templates/admin.html
```diff
@@ -18,11 +18,12 @@
 {% block main %}
 
 <div class="container">
+  {{ pagination.info|safe }}
```
minrk: pagination.info is currently undefined

share/jupyterhub/templates/admin.html
```diff
@@ -102,6 +103,15 @@
       {% endfor %}
   </tbody>
   </table>
+  {% if pagination.links %}
```
minrk: pagination.links is currently undefined

jupyterhub/handlers/pages.py
cbjuan: In this case, we need to use the `@classmethod` decorator to be able to call it from external methods/classes and get by default the self configurations. In other cases we get errors about not passing 2 arguments, or `'Pagination' object has no attribute 'per_page'`, etc.

jupyterhub/handlers/pages.py
cbjuan: Since we can use this pagination class from other methods (and not only for the admin template), I agree. Removing them.

jupyterhub/handlers/pages.py
cbjuan: In this case, I would prefer to maintain these in this way to don't copy the strings across methods or to allow to use the `Pagination.get_page_args(self)` as it's currently coded and without needing to fully instantiate the class to get the args to later re-instantiate the class again.

share/jupyterhub/templates/admin.html
cbjuan: Now it's defined :)

share/jupyterhub/templates/admin.html
cbjuan: Now it's defined :)

cbjuan:
In the last commits, I have included the information about pagination and the links to navigate between the pages in the admin view. Here are some screenshots about how they are displayed

The message with the pagination info before the table:
Screen Shot 2020-02-28 at 17 09 00

The different ways of working for the links:

When navigating from the page '1'
Screen Shot 2020-02-28 at 17 09 10

When navigating from a page > '1' and no near to the end of the pages list

Screen Shot 2020-02-28 at 17 09 23

When navigating in the last pages:
Screen Shot 2020-02-28 at 17 09 38
 
I hope this fulfills our expectations.

cbjuan:
Seems like the test for config ` python:3.5 + dist:xenial` is failing for a reason is not under my control. Probably it would be great if someone can restart it :)

cbjuan:
The PR lacks of a mechanism that prevents user from trying to render higher page numbers than the total_pages in the pagination.

cbjuan:
> The PR lacks a mechanism that prevents the user from trying to render higher page numbers than the total_pages in the pagination.

After thinking about this 👆, my conclusion is that we should not add that check (or at least there is room for discussion in further PRs). Currently, when navigating via the UI, the user cannot navigate to pages that do not contain users. The check would be good to control access via URLs. In that case, I think if anyone is trying to access a page programmatically, we should not redirect the URL to another page which contains users, because it can introduce misleading behaviors.

cbjuan:
After last improvement introduced in https://github.com/jupyterhub/jupyterhub/pull/2929/commits/6ed41b38edfd9c863994482800d8fb7b4500638d, I think the PR is ready for review again @minrk, @betatim

mriedem:
> I think the PR is ready for review again

Should the `[WIP]` be removed from the title then?

cbjuan:
> > I think the PR is ready for review again
> 
> Should the `[WIP]` be removed from the title then?

Yes!

minrk:
Thanks for waiting on this one! I just merged it with some minor changes to the display of info at the bottom.

cbjuan:
Ok. Cool @minrk. Thanks, everyone for your work and reviews on this PR 🥇

elgalu:
Is it per_page=100 configurable?

Now number of running servers only reflects the current page, e.g. this show 28 but we have hundreds
image

cbjuan:
> Is it per_page=100 configurable?
It's configurable via URL query param up to 250 per page (hardcoded). An example:

https://your-jupyterhub/hub/admin?page=1&per_page=250

meeseeksmachine:

https://discourse.jupyter.org/t/500-internal-server-error/3848/7

PR #2930, opened 2020-02-14
Add favicon to the base page template

JohnPaton:
This was missing before. Giving it its own named block will let users customize it if they wish.

Having the favicon present makes it much easier to navigate a sea of open tabs :)


minrk:
Thanks!

PR #2936, opened 2020-02-18
make init_spawners check O(running servers) not O(total users)

### LINKED ISSUE #2928: Extremely slow hub start up with large numbers of users
**Describe the bug**

We have an instance of JupyterHub with over 50,000 users in it, connected to a custom authenticator. When starting the hub pod it takes over 15 minutes to get to the ready state, where it is accepting requests. During this period of time the Hub is effectively offline for users and they can not start notebooks. This appears to be linear related to users based on the delays we were seeing at 30,000.

**To Reproduce**

1. Create 50k+ users
2. Boot hub, see how long until pod goes to ready state

**Expected behavior**

Ideally this would take a lot less time. 

**Compute Information**
 - Kubernetes 1.15
 - JupyterHub Version - jupyterhub/k8s-hub:0.9.0-beta.3

minrk:
by making the query on Server objects instead of User objects, which is logically equivalent because we were already filtering on `if spawner.server is not None`

avoids lots of ORM work on startup since there are typically a small number of running servers relative to the total number of users

this also means that the users dict is not fully populated at startup, which is a big change. I don't know that that's safe at this point and need to dig into that assumption in a few places.

In my tests, this brings startup with 30k users with only a few running down from 45 seconds to ~1 second.

closes #2928

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1670,9 +1670,12 @@ class JupyterHub(Application):
         # This lets whitelist be used to set up initial list,
         # but changes to the whitelist can occur in the database,
         # and persist across sessions.
+        total_users = 0
         for user in db.query(orm.User):
             try:
-                await maybe_future(self.authenticator.add_user(user))
+                f = self.authenticator.add_user(user)
+                if f:
+                    await maybe_future()
             except Exception:
                 self.log.exception("Error adding user %s already in db", user.name)
                 if self.authenticator.delete_invalid_users:
@@ -1694,6 +1697,7 @@ class JupyterHub(Application):
                         )
                     )
             else:
+                total_users += 1
                 # handle database upgrades where user.created is undefined.
                 # we don't want to allow user.created to be undefined,
                 # so initialize it to last_activity (if defined) or now.
@@ -1704,6 +1708,8 @@ class JupyterHub(Application):
         # The whitelist set and the users in the db are now the same.
         # From this point on, any user changes should be done simultaneously
         # to the whitelist set and user db, unless the whitelist is empty (all users allowed).
+
+        TOTAL_USERS.set(total_users)
 
     async def init_groups(self):
         """Load predefined groups into the database"""
@@ -2005,21 +2011,24 @@ class JupyterHub(Application):
             spawner._check_pending = False
 
         # parallelize checks for running Spawners
+        # run query on extant Server objects
+        # so this is O(running servers) not O(total users)
         check_futures = []
-        for orm_user in db.query(orm.User):
-            user = self.users[orm_user]
-            self.log.debug("Loading state for %s from db", user.name)
-            for name, orm_spawner in user.orm_spawners.items():
-                if orm_spawner.server is not None:
-                    # spawner should be running
-                    # instantiate Spawner wrapper and check if it's still alive
-                    spawner = user.spawners[name]
-                    # signal that check is pending to avoid race conditions
-                    spawner._check_pending = True
-                    f = asyncio.ensure_future(check_spawner(user, name, spawner))
-                    check_futures.append(f)
-
-        TOTAL_USERS.set(len(self.users))
+        for orm_server in db.query(orm.Server):
+            orm_spawners = orm_server.spawner
+            if not orm_spawners:
+                continue
+            orm_spawner = orm_spawners[0]
+            orm_user = orm_spawner.user
+            # instantiate Spawner wrapper and check if it's still alive
+            # spawner should be running
+            user = self.users[orm_spawner.user]
+            spawner = user.spawners[orm_spawner.name]
+            self.log.debug("Loading state for %s from db", spawner._log_name)
+            # signal that check is pending to avoid race conditions
+            spawner._check_pending = True
+            f = asyncio.ensure_future(check_spawner(user, spawner.name, spawner))
+            check_futures.append(f)
 
         # it's important that we get here before the first await
         # so that we know all spawners are instantiated and in the check-pending state
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -2019,7 +2019,6 @@ class JupyterHub(Application):
             if not orm_spawners:
                 continue
             orm_spawner = orm_spawners[0]
-            orm_user = orm_spawner.user
             # instantiate Spawner wrapper and check if it's still alive
             # spawner should be running
             user = self.users[orm_spawner.user]
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1675,7 +1675,7 @@ class JupyterHub(Application):
             try:
                 f = self.authenticator.add_user(user)
                 if f:
-                    await maybe_future()
+                    await maybe_future(f)
             except Exception:
                 self.log.exception("Error adding user %s already in db", user.name)
                 if self.authenticator.delete_invalid_users:
--- a/jupyterhub/tests/test_spawner.py
+++ b/jupyterhub/tests/test_spawner.py
@@ -103,7 +103,8 @@ async def wait_for_spawner(spawner, timeout=10):
 
 
 async def test_single_user_spawner(app, request):
-    user = next(iter(app.users.values()), None)
+    orm_user = app.db.query(orm.User).first()
+    user = app.users[orm_user]
     spawner = user.spawner
     spawner.cmd = ['jupyterhub-singleuser']
     await user.spawn()
--- a/jupyterhub/orm.py
+++ b/jupyterhub/orm.py
@@ -230,7 +230,7 @@ class Spawner(Base):
     user_id = Column(Integer, ForeignKey('users.id', ondelete='CASCADE'))
 
     server_id = Column(Integer, ForeignKey('servers.id', ondelete='SET NULL'))
-    server = relationship(Server, cascade="all")
+    server = relationship(Server, backref='spawner', cascade="all")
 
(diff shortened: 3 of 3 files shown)


minrk:
I've been using [this script](https://gist.github.com/23c6aaec9f5194b7ca965c2e22faa780) to get quick profile measurements of startup with different numbers of users.

sdague:
In our 50,000 user test this dramatically increases startup. Locally with PG on the local machine this is taking 2.5 minutes startup time without this, and 2 seconds with it. On our production system with cloud hosted database it's 15 minutes startup time. There are typically only 100 - 200 servers actually running with 50K+ accounts. So I expect equally impressive numbers.

betatim:
Looks like the only test that fails is

https://github.com/jupyterhub/jupyterhub/blob/3e17b47ec3309b1b15e6c9963107db1981554397/jupyterhub/tests/test_spawner.py#L105-L117

I thought we could fix it by replacing the first line with something like `user = User(app.db.query(orm.User).first(), {})` but that user doesn't have a `MockSpawner` in its spawner attribute, which then leads to other things breaking. I am not quite sure I understand why it doesn't have a mock spawner :-/

betatim:
```diff
diff --git a/jupyterhub/app.py b/jupyterhub/app.py
index 0bd57de7..f146631c 100644
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1672,7 +1672,9 @@ class JupyterHub(Application):
         # and persist across sessions.
         for user in db.query(orm.User):
             try:
-                await maybe_future(self.authenticator.add_user(user))
+                f = self.authenticator.add_user(user)
+                if f:
+                    await maybe_future(f)
             except Exception:
                 self.log.exception("Error adding user %s already in db", user.name)
                 if self.authenticator.delete_invalid_users:
diff --git a/jupyterhub/tests/test_spawner.py b/jupyterhub/tests/test_spawner.py
index 177c1da0..8ffb9312 100644
--- a/jupyterhub/tests/test_spawner.py
+++ b/jupyterhub/tests/test_spawner.py
@@ -102,8 +102,9 @@ async def wait_for_spawner(spawner, timeout=10):
     await wait()


-async def test_single_user_spawner(app, request):
-    user = next(iter(app.users.values()), None)
+async def test_single_user_spawner(app, db, request):
+    orm_user = User(app.db.query(orm.User).first(), {})
+    user = app.users.from_orm(orm_user)
     spawner = user.spawner
     spawner.cmd = ['jupyterhub-singleuser']
     await user.spawn()
```
I think this makes the tests pass

minrk:
@betatim found one example of that assumption, assuming `app.users` is populated, and it was in a test, which is a good sign! I'm digging through the code today to check for other references to the users dict to see if we've made this assumption elsewhere. There *shouldn't* be any, but it's worth an audit.

Just to write it down, the `users` dict is a collection of `User` object wrappers around the database entries for users. These contain the higher-level wrapper APIs and state such as running Spawner objects. It should have these characteristics:

- A `User` wrapper must be present whenever one or more Spawners is active for that user
- otherwise, a `User` wrapper does not need to exist

This means that when checking through our use of the `users` dict and we see cases of iterating through the dict or counting entries, only the number of *active* users or servers can be relevant. Anything operating on *all* users (e.g. the admin page) must start with a database query.

jupyterhub/app.py
```diff
@@ -1670,9 +1670,12 @@ def init_hub(self):
         # This lets whitelist be used to set up initial list,
         # but changes to the whitelist can occur in the database,
         # and persist across sessions.
+        total_users = 0
         for user in db.query(orm.User):
             try:
-                await maybe_future(self.authenticator.add_user(user))
+                f = self.authenticator.add_user(user)
+                if f:
+                    await maybe_future()
```
betatim: for my education: why doesn't the `f` go in the `maybe_future`? Right now it seems like it is a future that isn't awaited, which is a bug??

Or is it because we don't need to/care to wait for the future to resolve. Feels like there should be a way to make that explicit (or a comment?). The `await maybe_future()` is then just a way of saying "I yield, in case someone else wants to do something"?

jupyterhub/tests/test_spawner.py
```diff
@@ -103,7 +103,8 @@ def wait():
 
 
 async def test_single_user_spawner(app, request):
-    user = next(iter(app.users.values()), None)
+    orm_user = app.db.query(orm.User).first()
+    user = app.users[orm_user]
```
betatim: This works because of https://github.com/jupyterhub/jupyterhub/blob/3e17b47ec3309b1b15e6c9963107db1981554397/jupyterhub/user.py#L73 🎉

consideRatio:
image

What did you do here?

jupyterhub/app.py
minrk: >  why doesn't the f go in the maybe_future

Because I forgot to put it there. It definitely needs to be there!

minrk:
@consideRatio sorry, I was typing the comment and never posted it.

I've done a read through of the uses of the `.users` dict, and I think this should be fine and safe now, unless other folks want to have a look themselves.

minrk:
More background! In the database, a User has one or more Spawners associated with it. A running Spawner has one Server associated with it. A not running Spawner has zero Servers associated with it, so when we reconstruct running spawners, we go through all the Spawner objects that have a Server, and reconstruct them and verify that they are still alive.

Initially, we did this in the most obvious/logical/inefficient way possible:

```python
for user in users:
    for spawner in user.spawners:
        if spawner.server:
            # check if it's alive
```

However, this meant visiting all users and all spawners, even those that have no chance of being running.

This PR turns the loop inside out, visiting only the existing Servers directly, and then recreating the User and Spawner wrapper objects *only for the existing Server objects*, since that was the expensive part:

```python
for server in servers:
    spawner = spawner_for_server(server)
    user = user_for_spawner(spawner)
```

jupyterhub/app.py
```diff
@@ -2005,21 +2011,24 @@ def _user_summary(user):
...
-                    f = asyncio.ensure_future(check_spawner(user, name, spawner))
-                    check_futures.append(f)
-
-        TOTAL_USERS.set(len(self.users))
+        for orm_server in db.query(orm.Server):
+            orm_spawners = orm_server.spawner
+            if not orm_spawners:
+                continue
+            orm_spawner = orm_spawners[0]
+            orm_user = orm_spawner.user
```
betatim: It looks like this is an unused variable or is it here because accessing `orm_spawner.user` has a side-effect?

jupyterhub/app.py
```diff
@@ -2005,21 +2011,24 @@ def _user_summary(user):
...
-                    spawner._check_pending = True
-                    f = asyncio.ensure_future(check_spawner(user, name, spawner))
-                    check_futures.append(f)
-
-        TOTAL_USERS.set(len(self.users))
+        for orm_server in db.query(orm.Server):
+            orm_spawners = orm_server.spawner
+            if not orm_spawners:
+                continue
+            orm_spawner = orm_spawners[0]
```
betatim: Why only look at the first spawner instead of iterating over all of them?

betatim: Is it because we look at all Servers and a server only has one spawner? In which case why does `orm_server.spawner` return an array?

jupyterhub/app.py
minrk: no, I just thought I was going to use this variable more than once.

jupyterhub/app.py
minrk: there will always be one or zero because of how these are created. I'm not sure why this backref is a list instead of a singleton. I might need to fix something in the way the backref is created to make it a singleton, and I just forgot to look into this before moving on.

jupyterhub/app.py
betatim:
```suggestion

```

betatim:
Startup goes from 🚂  to 🚅 !

jupyterhub/app.py
minrk: Did some reading about backrefs in sqlalchemy, and now this is properly a 1:1 relationship

cbjuan:
Hello there. Is any estimated ETA to get this PR available on dev versions of the JupyterHub docker image? We're testing [0.9.0-beta.4.n000.h72e54ee](https://hub.docker.com/layers/jupyterhub/k8s-hub/0.9.0-beta.4.n000.h72e54ee/images/sha256-e6ddf7ef9dde382692564b4315a8f0428e4fbaf2e3834022c3a2d86de61ba69a?context=explore) but this PR seems like it's not present (still slow startup)

Also, is there any way to relate which PRs are in each docker image version?

Thanks!

betatim:
It is a bit tricky :) One thing that is useful to know: the last seven characters on most tags, versions, etc in JupyterHub land correspond to a git SHA. The hard part is knowing in which repository.

The image used by the zero2jupyterhub helm chart is built from https://github.com/jupyterhub/zero-to-jupyterhub-k8s/tree/master/images/hub which doesn't track jupyterhub/jupyterhub master but uses the released version. The 72e54ee in the tag you linked corresponds to https://github.com/jupyterhub/zero-to-jupyterhub-k8s/commit/72e54ee9814263a5c8d7a70e3f854bfda0b239d0.

So I'd build my own hub image which uses a new enough version of jupyterhub/jupyterhub and run that in your deployment/test cluster (via updating the tag used by the helm chart).

cbjuan:
Good to know. Thanks @betatim

PR #2938, opened 2020-02-18
[doc] Fix link to SSL encryption from troubleshooting page

kinow:
Hi, not sure if the syntax is 100%. Tested locally with `make html` and confirming that the link redirected to the right page+section.

Thanks!
Bruno


consideRatio:
I verified that the link from [here] was indeed broken, so it will probably be an improvement anyhow :)

consideRatio:
LGTM! Thank you Bruno! :heart:

/ Erik

PR #2939, opened 2020-02-18
[doc] Use fixed commit plus line number in github link

kinow:
Not sure if the best approach, or if we should include some sample code instead.

The link currently points to an empty string. Used `git blame`, looks like the link was added in 2016. Going back to last commit in November 2016, line 94 points to the `prepare` function.

Simply got the most recent commit for that function by going to GitHub UI, clicking on the line number, and then pressing `y`.

This way the docs will always point to this line. Users should be able to learn that that's the function referenced, then just click on `master` in GitHub UI to see the latest if available.

Cheers
Bruno


consideRatio:
:heart: makes sense to me! Thanks! :tada:

Issue #2924, opened 2020-02-07
Option to allow implicit server spawn


charnpreetsingh:
JupyterHub version 1.0.0 introduced a change to no longer implicitly spawn a user user if it's not running. A number of users find the "Server not Running" page confusing, especially coming from a non-technical audience or from past JupyterHub versions where servers launched in the background. It would be nice to add an option to the JupyterHub config to disable/enable implicit spawning of servers.

stevegore:
I'd like to echo that this is a really jarring experience for users. I've had numerous requests to get rid of what is perceived as a completely superfluous extra step.

(I'd also raised an issue [here](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1579) but realise now this is probably a JupyterHub issue.)

minrk:
I'm mixed on whether we should allow this as an option, versus attempting to improve the interstitial page, since this behavior is known to cause lots of very hard to resolve issues (hence the change). It is, however, an easy option to add, so we could punt and make it an option, but that means not improving the default experience for most users and reintroducing known problems for those who opt in to the old behavior.

One option for improving the existing page would be to allow the launch to happen automatically via *javascript* redirect. The advantage of a javascript redirect is that it requires a browser to actually visit the page, which should mitigate the launch triggered by 302-style redirect which is triggered even for e.g. API requests. That was one of the many problems caused by the implicit spawn, which shouldn't be an issue if it's triggered by javascript instead of a Location redirect.

From the other issue:

> However, we have users accessing direct links to notebooks. When they do so, no instance is spawned, and they are presented with the question:

"direct links to notebooks" is ambiguous. Whose notebook server was the link to? Was it user A sharing to user B a link to `/user/a/my/notebook` or `/user/b/my/notebook`? If it was `/user/a/my/notebook` (common, and the result of user A sharing a link from their URL bar), then the answer for what to do is ambiguous: sharing this URL *should* mean offering access to user B to user A's server (that's what the URL means). The "right" way to share a link to a notebook is to share `/hub/user-redirect/path/to/notebook` which does not have this issue at all. There is an issue about making the jupyterlab "get shareable link" behavior create these links properly, but I don't think it's happened yet.

minrk:
I've opened #2941 which implements the javascript redirect, which means that the user interaction step is indeed eliminated (basically the button is clicked automatically), but no other changes, which should leave in-tact most of the issues solved by the removal of the Location-based redirect.

[...]

PR #2941, opened 2020-02-20
Allow implicit spawn via javascript redirect

minrk:
Draft for design discussion.

there are deployments that are frustrated by the extra click required since 1.0 when ambiguous links are shared to notebook servers that aren't running (#2924).

Removing the Location-based redirect is important to fixing lots of issues and I don't want to remove that, but this PR allows deployments to opt-in to removing that user interaction by effectively automatically clicking that one button for the user after a delay.

The javascript redirect is much better than the Location redirect because it will only happen to "real" browser visits, not API requests, scripts and the like.

Screenshot when `JupyterHub.implicit_spawn_seconds` is defined:

Screen Shot 2020-02-20 at 12 45 44

Default behavior is unchanged.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -917,9 +917,36 @@ class JupyterHub(Application):
     def _authenticator_default(self):
         return self.authenticator_class(parent=self, db=self.db)
 
+    allow_implicit_spawn = Bool(
+        False,
+        help="""Allow implicit spawning
+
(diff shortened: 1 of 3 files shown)


minrk:
After thinking more about this, I'm happy with it as an option now that it's a javascript redirect rather than restoring the old server-side redirect behavior.

PR #2944, opened 2020-02-21
make spawner:server relationship explicitly one to one

minrk:
in #2936 I hadn't figured out how to make a backref not have a list on one side, but this is a 1:1 relationship. Turns out `uselist=False` is the way to tell it not to use a list.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -2013,12 +2013,19 @@ class JupyterHub(Application):
         # parallelize checks for running Spawners
         # run query on extant Server objects
         # so this is O(running servers) not O(total users)
+        # Server objects can be associated with either a Spawner or a Service,
+        # we are only interested in the ones associated with a Spawner
         check_futures = []
         for orm_server in db.query(orm.Server):
-            orm_spawners = orm_server.spawner
-            if not orm_spawners:
+            orm_spawner = orm_server.spawner
+            if not orm_spawner:
+                # sanity check for orphaned Server rows
+                # this shouldn't happen if we've got our sqlachemy right
+                if not orm_server.service:
+                    self.log.warning("deleting orphaned server %s", orm_server)
+                    self.db.delete(orm_server)
+                    self.db.commit()
                 continue
-            orm_spawner = orm_spawners[0]
             # instantiate Spawner wrapper and check if it's still alive
             # spawner should be running
             user = self.users[orm_spawner.user]
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -2019,7 +2019,7 @@ class JupyterHub(Application):
         for orm_server in db.query(orm.Server):
             orm_spawner = orm_server.spawner
             if not orm_spawner:
-                # sanity check for orphaned Server rows
+                # check for orphaned Server rows
                 # this shouldn't happen if we've got our sqlachemy right
                 if not orm_server.service:
                     self.log.warning("deleting orphaned server %s", orm_server)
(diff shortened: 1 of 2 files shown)


betatim:
Travis is unhappy :-/

minrk:
@betatim thanks, fixed!

betatim:
If someone understands this please do review and merge (and leave some explanation for me to learn from). I realised that my knowledge of sqlalchemy is equivalent to taking "Sqlalchemy 101" and skipping most of the lectures and homework assignments :-/ -> it will take me a while to read and learn before I have a qualified opinion.

jupyterhub/app.py
```diff
@@ -2013,12 +2013,19 @@ def _user_summary(user):
...
         check_futures = []
         for orm_server in db.query(orm.Server):
-            orm_spawners = orm_server.spawner
-            if not orm_spawners:
+            orm_spawner = orm_server.spawner
+            if not orm_spawner:
+                # sanity check for orphaned Server rows
+                # this shouldn't happen if we've got our sqlachemy right
+                if not orm_server.service:
+                    self.log.warning("deleting orphaned server %s", orm_server)
```
minrk: This is a condition that shouldn't happen (`not server.spawner and not server.service`), but is being cleaned up anyway because I've been known to make mistakes that leave orphaned things in the database.

If the ORM is set up correctly, deleting Spawner and Service objects should mean that associated Server objects also get deleted. There should be no Server objects not associated with Spawners or Services.

Since we are now already iterating over all the Server objects, this bit here is a check to validate that assertion, and delete any Server objects that may still exist without valid relationships (e.g. a Spawner was deleted but its Server was orphaned instead of being deleted at the same time). Basically, it's cleaning up after hypothetical bugs in past versions of the ORM code.

If everything is correct (and always has been correct), this condition will never be met.

jupyterhub/orm.py
```diff
@@ -230,7 +231,12 @@ class Spawner(Base):
     user_id = Column(Integer, ForeignKey('users.id', ondelete='CASCADE'))
 
     server_id = Column(Integer, ForeignKey('servers.id', ondelete='SET NULL'))
-    server = relationship(Server, backref='spawner', cascade="all")
+    server = relationship(
```
mriedem: I was just going to point out that this is similar to how https://github.com/jupyterhub/jupyterhub/blob/84acdd5a7ffb3d79ca04aa86ffcea7e067cce6ee/jupyterhub/orm.py#L175 works which implicitly creates a `Spawner.user` attribute, the only difference being `uselist=False` and `single_parent=True`.

I've worked on another project that also used `uselist=False` in a couple of places, e.g.:

https://github.com/openstack/nova/blob/b6e353a2c84d87bbb1353ff6602f9f13dc62a2fd/nova/db/sqlalchemy/api_models.py#L151

And that's pretty similar to how you're using it here (a `cell_mapping` can have many `instance_mappings` but an `instance_mapping` can only ever be in one `cell_mapping`).

jupyterhub/app.py
```diff
@@ -2013,12 +2013,19 @@ def _user_summary(user):
...
         # so this is O(running servers) not O(total users)
+        # Server objects can be associated with either a Spawner or a Service,
+        # we are only interested in the ones associated with a Spawner
         check_futures = []
         for orm_server in db.query(orm.Server):
-            orm_spawners = orm_server.spawner
-            if not orm_spawners:
+            orm_spawner = orm_server.spawner
+            if not orm_spawner:
+                # sanity check for orphaned Server rows
```
willingc:
```suggestion
                # check for orphaned Server rows
```

willingc:
Doc failure is unrelated. Opening an issue.

PR #2947, opened 2020-02-21
use attribute_mapped_collection for user:spawner

minrk:
allows automatic, direct creation of dict instead of using a property to wrap the default list

discovered this possibility while working on #2944


minrk:
From the test failures, looks like I didn't do this right. This is not a high priority! Just an optimization that I thought was easy, but is apparently slightly less easy :)

PR #2953, opened 2020-02-27
preserve auth type when logging obfuscated auth header

minrk:
Authorization header has the form `"<type> <credentials>"`

rather than checking for "token" only, preserve type value, which could be Bearer, Basic, etc.


betatim:
LGTM. Made a small edit to your comment to make the `<textinbrackets>` appear.

PR #2957, opened 2020-02-28
Add trailing slash to server_name

GeorgianaElena:
Without this trailing '/', traefik-proxy doesn't know how to route to a named server.

As reported by https://github.com/jupyterhub/traefik-proxy/issues/91, traefik-proxy adds the routespec received from JH (that ends with a '/') to the routing table and routes properly to the named server only when the server_url ends with '/' as well.

I think CHP doesn't have this problem because when it builds its routing table it strips the trailing slashes. I believe that instead of making traefik-proxy do the same, since we're adding a trailing slash to the `server_url` for the default servers anyway, we should do the same for the named servers.

### DIFF
--- a/jupyterhub/user.py
+++ b/jupyterhub/user.py
@@ -398,7 +398,7 @@ class User:
         if not server_name:
             return self.url
         else:
-            return url_path_join(self.url, server_name)
+            return url_path_join(self.url, server_name) + '/'
 
     def progress_url(self, server_name=''):
         """API URL for progress endpoint for a server with a given name"""


GeorgianaElena:
@minrk, do you think it makes sense for the server_url to end with a  '/' or is it better to change the routing rules in traefik-proxy?

consideRatio:
@GeorgianaElena I think that whatever distinguish a trailing slash and no trailing slash should update itself to not do so. I recently experienced a similar issue in Grafana, where a cookie was only passed to a route with a trailing `/` (https://github.com/grafana/grafana/pull/22265) but not without.

consideRatio:
Oh @GeorgianaElena I think this comment of yours is very relevant to link explicitly from this PR as well: https://github.com/jupyterhub/traefik-proxy/issues/91#issuecomment-592411745

minrk:
I do think this should be fixed in traefik-proxy if we can. Is it possible for `user/name/server_name` to be routed correctly without the trailing slash?

GeorgianaElena:
If we follow what CHP is doing [here](https://github.com/jupyterhub/jupyterhub/blob/4e8acc71c671d39b0ae6f140d894c983bc448233/jupyterhub/proxy.py#L737) it should be routed correctly. I will open a PR in traefik-proxy and close this one.

I don't understand however what's the reason why there is a trailing slash for default servers but not for named servers. Is there a use case we're breaking if `server_name` ends with a slash?

PR #2960, opened 2020-02-28
Start named servers by pressing the Enter key

jtpio:
Add support for the <kbd>Enter</kbd> key as an alternative to the "Add New Server" button to start new named servers.

named-servers-enter


jtpio:
Yes that would be the case with a `form` tag, but indeed there is no form submitted in this case.

Issue #2954, opened 2020-02-27
cull_idle_servers causes hub to go unresponsive in environments with 50,000 users


sdague:
**Describe the bug**

After addressing #2928 , we discovered there is another issue when the hub starts in a kubernetes environment. 

The cull_idle_servers script does a GET /api/users, then iterates through the users to find servers. In our environment we typically have 100 to 300 servers, with 50,000+ users. The GET /api/users call takes minutes to return. That means that all other requests to the hub are blocked, because the asyncio handler for /users is blocking on the database. 

In a kubernetes environment this also means we miss 3 /api/health heartbeats, which means the service is marked not ready, and the ingress controller no longer has a route to the hub, and starts throwing 503 errors back to the user.

** Options **

It seems like there are a couple of approaches. 

1. Add pagination to the /api/users interfaces. This could work off of the progress in #2929 . 
2. Expose an /api/servers admin interface, because for culling servers we really only care about actual servers. Getting all the servers would be much faster than getting all Users.

**Compute Information**
 - Operating System
 - JupyterHub Version master bc7bb5076ff2deabc7702c75dce208166d14568e

sdague:
@minrk @cbjuan

betatim:
I like the idea of allowing the culler to be inverted the same way as the startup process (so expose a `/api/servers` end point). We could also transition slowly to it for existing culler scripts. The current culler scripts will keep working (for small number of users) and we can add a new culler that uses the new servers endpoint. This is nice because there are several copies/variations of the culler script floating around and people might have even forked it for their own use case.

Maybe another option to think about is if the `/api/users` endpoint could return a random subset of all users (when asked to do so). That way we'd have the culler only see (say) 300 users at a time, the call would be fast and we'd not have to change much in the culler script (maybe make it submit a new parameter in its request specifying that it wants a random subset).

sdague:
Yeah, I think my generally preference is to expose /api/servers as an admin API. I think it might be generally valuable long term as well. And that would provide an alternate cull script that uses that instead.

mriedem:
I'll start working on exposing a `GET /api/servers` admin API PR (first time contributor to this project so it might be a bit slow going to get setup).

mriedem:
[...]

PR #2962, opened 2020-02-28
Add GET /servers API

mriedem:
This adds a `GET /servers` admin API which attempts to be as close to the `GET /users` admin API with just the servers resource portion of that response without needing to query the users table in the database.

A difference from the servers returned in the `GET /users` API is that `GET /servers` will return the `id` and minimal `user` object for each server resource in the response. The `id` is needed to uniquely identify a server in upcoming `GET /servers/{id}` and `DELETE /servers/{id}` APIs without the user context. The user name is just a nicety to use other APIs like `DELETE /users/{name}/servers/{server_name}`.

This is eventually something that is going to be used when culling idle servers without querying users to improve performance in environments with a large number of users (tens of thousands).

Part of issue #2954

### DIFF
--- a/docs/rest-api.yml
+++ b/docs/rest-api.yml
@@ -405,6 +405,16 @@ paths:
           description: The authenticated user's model is returned.
           schema:
             $ref: '#/definitions/User'
+  /servers:
+    get:
+      summary: List servers
+      responses:
+        '200':
+          description: The Hub's server list
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/Server'
   /groups:
     get:
       summary: List groups
@@ -768,6 +778,9 @@ definitions:
   Server:
     type: object
     properties:
+      id:
+        type: integer
+        description: The unique internal ID of the server.
       name:
         type: string
         description: The server's name. The user's default server has an empty name ('')
@@ -787,6 +800,14 @@ definitions:
         description: |
           The URL where the server can be accessed
           (typically /user/:name/:server.name/).
+      user:
+        type: object
+        description: |
+          Contains information about the user that owns the server.
+          Only available on ``GET /servers``.
+          Includes fields:
+
+          - ``name``: The name of the user that owns the server.
       progress_url:
         type: string
         description: |
--- a/docs/rest-api.yml
+++ b/docs/rest-api.yml
@@ -780,7 +780,7 @@ definitions:
     properties:
       id:
         type: integer
-        description: The unique internal ID of the server.
+        description: The unique internal ID of the server. Only available on ``GET /servers``.
       name:
         type: string
         description: The server's name. The user's default server has an empty name ('')
--- a/jupyterhub/apihandlers/base.py
+++ b/jupyterhub/apihandlers/base.py
@@ -133,6 +133,15 @@ class APIHandler(BaseHandler):
 
     def server_model(self, spawner, include_state=False):
         """Get the JSON model for a Spawner"""
+        # TODO(mriedem): Need to convert orm.Spawner to spawner.Spawner
+        # to use the `get_state` method, but how to do that conversion?
+        if hasattr(spawner, 'user') and hasattr(spawner.user, 'url'):
+            url = url_path_join(spawner.user.url, spawner.name, '/')
+        else:
+            url = None
+        progress_url = (
+            spawner._progress_url if hasattr(spawner, '_progress_url') else None
+        )
         return {
             'name': spawner.name,
             'last_activity': isoformat(spawner.orm_spawner.last_activity),
@@ -140,9 +149,9 @@ class APIHandler(BaseHandler):
             'pending': spawner.pending,
             'ready': spawner.ready,
             'state': spawner.get_state() if include_state else None,
-            'url': url_path_join(spawner.user.url, spawner.name, '/'),
+            'url': url,
             'user_options': spawner.user_options,
-            'progress_url': spawner._progress_url,
+            'progress_url': progress_url,
         }
 
     def token_model(self, token):
--- a/jupyterhub/apihandlers/base.py
+++ b/jupyterhub/apihandlers/base.py
@@ -9,6 +9,7 @@ from sqlalchemy.exc import SQLAlchemyError
 from tornado import web
 
 from .. import orm
+from .. import spawner as spawner_obj
 from ..handlers import BaseHandler
 from ..utils import isoformat
 from ..utils import url_path_join
@@ -131,28 +132,29 @@ class APIHandler(BaseHandler):
             json.dumps({'status': status_code, 'message': message or status_message})
         )
 
-    def server_model(self, spawner, include_state=False):
+    def server_model(self, spawner, include_state=False, include_user_and_id=False):
         """Get the JSON model for a Spawner"""
-        # TODO(mriedem): Need to convert orm.Spawner to spawner.Spawner
-        # to use the `get_state` method, but how to do that conversion?
-        if hasattr(spawner, 'user') and hasattr(spawner.user, 'url'):
-            url = url_path_join(spawner.user.url, spawner.name, '/')
-        else:
-            url = None
-        progress_url = (
-            spawner._progress_url if hasattr(spawner, '_progress_url') else None
-        )
-        return {
+        if isinstance(spawner, orm.Spawner):
+            # TODO(mriedem): Would be nice to join the user field from
+            # the Spawner ORM object when we fetch it from the DB so we don't
+            # need the roundtrip.
+            orm_user = orm.User.find_by_id(self.db, spawner.user_id)
+            user = self._user_from_orm(orm_user)
+            spawner = spawner_obj.Spawner(orm_spawner=spawner, user=user)
+        server = {
             'name': spawner.name,
             'last_activity': isoformat(spawner.orm_spawner.last_activity),
             'started': isoformat(spawner.orm_spawner.started),
             'pending': spawner.pending,
             'ready': spawner.ready,
             'state': spawner.get_state() if include_state else None,
-            'url': url,
+            'url': url_path_join(spawner.user.url, spawner.name, '/'),
             'user_options': spawner.user_options,
-            'progress_url': progress_url,
+            'progress_url': spawner._progress_url,
         }
+        if include_user_and_id:
+            server.update({'id': spawner.id, 'user': {'name': spawner.user.name}})
+        return server
 
     def token_model(self, token):
         """Get the JSON model for an APIToken"""
--- a/jupyterhub/apihandlers/base.py
+++ b/jupyterhub/apihandlers/base.py
@@ -135,16 +135,12 @@ class APIHandler(BaseHandler):
     def server_model(self, spawner, include_state=False, include_user_and_id=False):
         """Get the JSON model for a Spawner"""
         if isinstance(spawner, orm.Spawner):
-            # TODO(mriedem): Would be nice to join the user field from
-            # the Spawner ORM object when we fetch it from the DB so we don't
-            # need the roundtrip.
-            orm_user = orm.User.find_by_id(self.db, spawner.user_id)
-            user = self._user_from_orm(orm_user)
+            user = self._user_from_orm(spawner.user)
             spawner = spawner_obj.Spawner(orm_spawner=spawner, user=user)
         server = {
             'name': spawner.name,
-            'last_activity': isoformat(spawner.orm_spawner.last_activity),
-            'started': isoformat(spawner.orm_spawner.started),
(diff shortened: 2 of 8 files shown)


mriedem:
The new `test_get_servers` test is failing because the tests are not idempotent so by the time it runs there are other servers (spawners) in the database. I guess the servers tests should be moved before the users tests?

docs/rest-api.yml
```diff
@@ -768,6 +778,9 @@ definitions:
   Server:
     type: object
     properties:
+      id:
```
mriedem: Question for reviewers: how are API changes like this communicated? Change log? Release notes? IOW, how does someone know that they can or should see these fields, i.e. maybe I don't see them because JupyterHub I'm talking to is old, or maybe because I'm not an admin and so I can't do `GET /servers`?

docs/rest-api.yml
```diff
@@ -768,6 +778,9 @@ definitions:
...
+        type: integer
+        description: The unique internal ID of the server.
```
mriedem: I suppose this should also say:

```
Only available on ``GET /servers``.
```

jupyterhub/apihandlers/base.py
```diff
@@ -131,9 +132,16 @@ def write_error(self, status_code, **kwargs):
...
+            # TODO(mriedem): Would be nice to join the user field from
+            # the Spawner ORM object when we fetch it from the DB so we don't
+            # need the roundtrip.
+            orm_user = orm.User.find_by_id(self.db, spawner.user_id)
+            user = self._user_from_orm(orm_user)
+            spawner = spawner_obj.Spawner(orm_spawner=spawner, user=user)
+        server = {
             'name': spawner.name,
             'last_activity': isoformat(spawner.orm_spawner.last_activity),
             'started': isoformat(spawner.orm_spawner.started),
```
mriedem: Note that with the `Spawner.__getattr__` override now we don't need to touch the `orm_spawner` directly here. Future cleanup I guess.

jupyterhub/apihandlers/servers.py
```diff
@@ -0,0 +1,29 @@
...
+class ServerListAPIHandler(APIHandler):
+    @admin_only
+    def get(self):
+        # TODO(mriedem): Should include_state be conditional on
+        # `spawner.active` like how `user_model` works for
+        # `GET /users`?
+        include_state = True
+        data = [
+            self.server_model(
+                spawner, include_state=include_state, include_user_and_id=True
```
mriedem: I sort of wonder if state and the user-related fields like `url` and `user` should be query parameters which are off by default so that someone calling `GET /servers` that don't care about any of the user information don't need to fetch it from the DB. IOW, I tried to make `GET /servers` similar to how `GET /users` works but maybe that should be opt-in and we should be returning the most minimal representation of each server/spawner by default?

pytest.ini
```diff
@@ -10,3 +10,4 @@ markers =
     services: mark as a services test
     user: mark as a test for a user
     slow: mark a test as slow
+    server: mark as a test for a server
```
mriedem: Note that I'm not sure how useful this is because while it allows running just the server annotated tests, because of how the tests are setup to not be idempotent just running `pytest -m server` will fail, so this isn't very useful right now but seemed like the pattern to copy.

docs/rest-api.yml
mriedem: Should I be updating `docs/source/changelog.md`?

mriedem: Thinking through this a bit more, the root `GET /` route returns the API version so if API changes are documented in the change log properly then I guess people could determine what they should expect that way in client side code though it would be nice if the API reference itself indicated when something was added or changed by version.

As for knowing if `GET /servers` is available that should be pretty easy - if it's not the response will be a `404` so the client needs to fallback to some other behavior.

jupyterhub/apihandlers/servers.py
mriedem: While adding the `DELETE /servers/{id}` API in pull #2967 I realized [1] that it's not going to work well in the current form of the API because you need the server id to delete it but only `GET /servers` in this pull request returns the server id and `GET /servers` is an admin-only API, so this means non-admins can't delete their own servers unlike `DELETE /users/{name}/servers/{server_name}` which is bad UX.

So I think our options are:

1. Continue with adding a dedicated `DELETE /servers/{id}` API which allows `@admin_or_self` and modify the `GET /user` and `GET /users/{name}` responses so they include the id of the user-owned servers.

    * Pros: admin tooling like `cull_idle_servers` would not need to get the user information when listing servers and deleting them making `GET /servers` faster (no extra join from `servers` to `users` in the DB).
    * Cons: more API changes and bloat (more than one way to delete a server), expanded maintenance cost

2. Scrap the idea of a dedicated `DELETE /servers/{id}` API and as long as `GET /servers` returns the user name the client can use the existing `DELETE /users/{name}/servers/{server_name}` API.

    * Pros: re-use existing API to delete a server that `cull_idle_servers` is already using, no new `DELETE /servers/{id}` API or changes to `GET /user` or `GET /users/{name}` to return the server id.
    * Cons: Slight performance hit on `GET /servers` to include the user name since we have to query that from the DB per server that is returned (joining the user when we query the servers might negate that, or we could cache the `user_id -> User` mapping if the same user owns more than one server).

In the case of issue #2954 the client is the `cull_idle_servers` script which runs with an admin token.

[1] https://github.com/jupyterhub/jupyterhub/pull/2967/files#diff-0a701bf780442df25926c4771a5f170eR30

jupyterhub/orm.py
```diff
@@ -220,6 +220,15 @@ def find(cls, db, name):
         """
         return db.query(cls).filter(cls.name == name).first()
 
+    @classmethod
+    def find_by_id(cls, db, id):
```
mriedem: Note that we won't need this if we join the user with the servers when we fetch the servers from the DB. Maybe that's already done though with this?

```python
_orm_spawners = relationship(
    "Spawner", backref="user", cascade="all, delete-orphan"
)
```

jupyterhub/apihandlers/servers.py
mriedem: I've gone with option 2 and dropped pull #2967 once I got `cull_idle_servers` updated to use `GET /servers` in pull #2969.

rkdarst: At least I use `state` in my hubs to adjust culling parameters within the cull_idle script.  It could be reasonable to be off by default, as long as it's obvious how to turn it on again, which I can add later if need be (but I guess it should be the same setting for /users and /servers?).

mriedem: > At least I use `state` in my hubs to adjust culling parameters within the cull_idle script. It could be reasonable to be off by default, as long as it's obvious how to turn it on again, which I can add later if need be

I would think that would be a query parameter if `state` is not returned by default from `GET /servers` so you could do something like:

`GET /servers?include_state=true`

Or something similar to get the `state` field returned if you need it.

> (but I guess it should be the same setting for /users and /servers?).

Yeah this is my main question and if so, it means we should only include `state` for active servers which is how the `GET /users` (`user_model`) code handles that. I'm fine either way, was just looking for direction from maintainers on this wrinkle.

docs/rest-api.yml
betatim: Yes please to updating the changelog. We often forget and then someone has to go through all PRs when they are making the release.

If you look around the code some of the methods/functions where behaviour changed have a note saying when&what changed. It seems like none of the entries in `rest-api.yml` have those so I'd also not add anything for this new route. I think additions (or changes that don't break existing behaviour) are Ok to just announce in the changelog.

jupyterhub/tests/test_api.py
```diff
@@ -176,7 +230,7 @@ def fill_user(model):
     users = [normalize_user(u) for u in users]
     assert users == [
         fill_user({'name': 'admin', 'admin': True}),
-        fill_user({'name': 'user', 'admin': False, 'last_activity': None}),
```
betatim: Why this change?

jupyterhub/tests/test_api.py
```diff
@@ -109,6 +109,50 @@
...
+    """
+    db = app.db
+    r = await api_request(app, 'servers')
+    assert r.status_code == 200
+
+    servers = sorted(r.json(), key=lambda d: d['id'])
+    servers = [normalize_server(server) for server in servers]
+    # Note that there are two servers to start because the `app` fixture
+    # creates a user named "user" and test_referer_check creates an admin user.
+    assert servers == [fill_server(server) for server in servers]
```
betatim: What exactly does this check? At first I thought it checks that we indeed got two servers back from the API, but it doesn't. Then I thought it was checking that the model/JSON fields returned from the API call were as expected, but now I don't think it does that either.

jupyterhub/spawner.py
```diff
@@ -212,6 +213,19 @@ def name(self):
             return self.orm_spawner.name
         return ''
 
+    # pass getattr to ORM spawner
+    def __getattr__(self, attr):
```
betatim: Maybe @minrk can comment on why so far attributes were not automatically forwarded to the ORM class.

My Python intuition (potentially outdated) makes me feel a bit uncomfortable when I see classes implementing dunder methods for convenience. It seems like a good idea at the time but down the road it causes confusion/debugging pain. Usually the only advantage is the saving of a few characters when writing the original code.

jupyterhub/apihandlers/base.py
```diff
@@ -131,19 +132,25 @@ def write_error(self, status_code, **kwargs):
...
+            'started': isoformat(spawner.started),
             'pending': spawner.pending,
             'ready': spawner.ready,
             'state': spawner.get_state() if include_state else None,
             'url': url_path_join(spawner.user.url, spawner.name, '/'),
             'user_options': spawner.user_options,
             'progress_url': spawner._progress_url,
         }
+        if include_user_and_id:
+            server.update({'id': spawner.id, 'user': {'name': spawner.user.name}})
```
betatim: With `DELETE /server/{id}` out of the picture now, what do we need the ID for? To delete the server I need the username and the servername. The ID doesn't seem like something I'd need as a consumer?

If we do remove it we can rename this from the long name of `include_user_and_id` to `include_user`.

minrk:
Would the goal here (making the request to list running users and servers more efficient for large numbers of users) perhaps be more easily accomplished with a `?active=true` query param on `GET /users` that results in a filtered query? That would be a minimal, backward-compatible API change with no new API endpoints or schema changes, just an improved efficiency for a common costly request?

jupyterhub/apihandlers/base.py
mriedem: Yeah `id` can be removed.

jupyterhub/spawner.py
mriedem: I copied this pattern from the `user.User` ORM wrapper and while it makes the client-side code for this object a bit cleaner (and less tightly coupled to the internals of this object) it does introduce some magic. I can see pros/cons both ways. I don't have an issue with removing this.

jupyterhub/tests/test_api.py
mriedem: I think at some point while working on this I had it checking two servers individually like how `test_get_users` works but obviously removed that, I'm not sure why now. It should be changed since if we don't get any servers back this will still pass with a false positive. If we remove the `id` field and sort by something like the user name then we could assert two servers in the order we expect.

jupyterhub/tests/test_api.py
mriedem: When I added the test above the `last_activity` field on this `user` server was no longer None, I'm not entirely sure why. I assumed it was because as a result of the new test that runs before this (and these tests aren't idempotent) that something _touched_ the server so it had an activity set, but I'm not familiar enough with the code to know what might cause that side effect.

mriedem:
> Would the goal here (making the request to list running users and servers more efficient for large numbers of users) perhaps be more easily accomplished with a `?active=true` query param on `GET /users` that results in a filtered query? That would be a minimal, backward-compatible API change with no new API endpoints or schema changes, just an improved efficiency for a common costly request?

I replied in more detail to the same question on the issue:

https://github.com/jupyterhub/jupyterhub/issues/2954#issuecomment-596694346

tl;dr we can probably just abandon this pull request now unless you think there is other general usefulness in having this API (though if you're like me I wouldn't want to add something I need to maintain if there is no clear use case for it yet).

mriedem:
As explained earlier and in detail in issue #2954 this is not something we're going to use now (I'll also close pull #2969) since we're now running the hub-managed `cull-idle` service with `--cull-users`. We ca follow up in issue #2954 with adding any server-side filtering logic to the `GET /users` API if we want to try and speed that up for large user data sets.

PR #2967, opened 2020-03-02
[WIP] Start work on the DELETE /servers/{id} API

mriedem:
This starts the `DELETE /servers/{id}` API based on the
existing `DELETE /users/{name}/servers/{server_name}` API.
TODOs are inline for testing and DRYing up some of that
code, plus docs are obviously needed.

There is also a question of whether or not we want to do
a separate route to stop a server versus delete it to keep
this code a bit less complicated.

Part of issue #2954

### DIFF
--- a/docs/rest-api.yml
+++ b/docs/rest-api.yml
@@ -405,6 +405,16 @@ paths:
           description: The authenticated user's model is returned.
           schema:
             $ref: '#/definitions/User'
+  /servers:
+    get:
+      summary: List servers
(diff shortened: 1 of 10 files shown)


mriedem:
I'm going to close this out due to the comments in pull #2962 here:

https://github.com/jupyterhub/jupyterhub/pull/2962#discussion_r387097765

And pull #2969 shows that we can use the existing APIs to cull servers just fine.

PR #2969, opened 2020-03-03
Change cull_idle to use GET /servers if not culling users

mriedem:
This changes the `cull_idle_servers` script such that if users
are not being culled (the default) and the `GET /servers` API
is available it uses that to get servers directly rather than
go through `GET /users` which could have a lot more records if
servers are being culled but users are not.

In our deployment we have ~50K users but only a couple hundred
servers and since `cull_idle_servers` runs on startup it makes
restarting our single jupyterhub pod instance quite slow:

```
[I 2020-02-27 13:32:07.798 JupyterHub log:174] 200 GET /hub/api/users (cull-idle@127.0.0.1) 1222100.09ms
```

There is a bit of refactoring in this change to have consistent
"log names" and to be graceful about `GET /servers` not being
available.

Closes issue #2954

### DIFF
--- a/docs/rest-api.yml
+++ b/docs/rest-api.yml
@@ -405,6 +405,16 @@ paths:
           description: The authenticated user's model is returned.
           schema:
             $ref: '#/definitions/User'
+  /servers:
+    get:
+      summary: List servers
+      responses:
+        '200':
+          description: The Hub's server list
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/Server'
   /groups:
     get:
       summary: List groups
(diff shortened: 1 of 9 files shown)


mriedem:
Note to reviewers: this is meant to build on pull #2962 so the only thing new in this change is the `cull_idle_servers.py` changes:

https://github.com/jupyterhub/jupyterhub/pull/2969/files#diff-7282c7cf243e8b33d617744112327cbc

mriedem:
As explained in pull #2962 this is not something we're going to use now since we're now running the hub-managed `cull-idle` service with `--cull-users`. We can follow up in issue #2954 with adding any server-side filtering logic to the `GET /users` API if we want to try and speed that up for large user data sets.

PR #2971, opened 2020-03-04
Merge pull request #2971 from mriedem/issues/2970-singleuser-version-logging

### LINKED ISSUE #2970: singleuser server version check spams the logs
**Describe the bug**

We have ~277 active single user servers in our deployment right now and on restart of the hub service we see this for each one:

> Mar 4 09:20:45 hub-7bccd48cd5-mp4fk hub [W 2020-03-04 15:20:45.996 JupyterHub _version:56] jupyterhub version 1.2.0dev != jupyterhub-singleuser version 1.1.0. This could cause failure to authenticate and result in redirect loops! 

My only complaint is that logging that per server is redundant and spams the logs. Can we just log that once per restart of the hub?

**To Reproduce**

Have the jupyterhub and jupyterhub-singleuser services at different minor versions.

**Expected behavior**

Just log the warning once since there is no user/server specific context in the message.

**Compute Information**
 - 1.2.0dev - we're running with a custom build based on b4391d0f796864a5b01167701d95eafce3ad987e so that we can pick up the performance fix for issue #2928.

mriedem:
If your jupyterhub and jupyterhub-singleuser instances
are running at different minor or greater versions a
warning gets logged per active server which can be a lot
when you have hundreds of active servers.

This adds a flag to that version mismatch logging logic
such that the warning is only logged once per hub/singleuser
per restart of the hub server.

Closes #2970

### DIFF
--- a/jupyterhub/_version.py
+++ b/jupyterhub/_version.py
@@ -18,6 +18,15 @@ version_info = (
 
 __version__ = ".".join(map(str, version_info[:3])) + ".".join(version_info[3:])
 
+# Singleton flag to only log the major/minor mismatch warning once.
+_version_mismatch_warning_logged = False
+
+
+def reset_globals():
+    """Used to reset globals between test cases."""
+    global _version_mismatch_warning_logged
+    _version_mismatch_warning_logged = False
+
 
 def _check_version(hub_version, singleuser_version, log):
     """Compare Hub and single-user server versions"""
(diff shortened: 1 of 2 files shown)


jupyterhub/_version.py
```diff
@@ -42,19 +51,26 @@ def _check_version(hub_version, singleuser_version, log):
...
+                do_log = False  # We already logged this warning so don't log it again.
+            else:
+                log_method = log.warning
+                extra = " This could cause failure to authenticate and result in redirect loops!"
+                _version_mismatch_warning_logged = True
+        if do_log:
+            log_method(
+                "jupyterhub version %s != jupyterhub-singleuser version %s." + extra,
+                hub_version,
+                singleuser_version,
```
mriedem: Tim mentioned that `singleuser_major_minor` could be different so only logging once per instance of this warning is probably too heavy so the flag should probably be based on `singleuser_major_minor` using a dict.

minrk: I think the dict is a good idea.

It's probably also good to change this to setting `log_method = log.debug` when it's been seen before, rather than `do_log = False`. Because as it is, if the version doesn't match and it's called for a second time, there's going to be an erroneous debug log saying that the version does match when it doesn't. Logging the mismatch every time at the debug level should still be fine.

mriedem: > Because as it is, if the version doesn't match and it's called for a second time, there's going to be an erroneous debug log saying that the version does match when it doesn't.

I think you might be looking at the nesting wrong here (it's kind of hard to see with the diff in github) but if `do_log` is False we don't log anything. The `else` block below that I think you're looking at is only logged if the versions match which is a different conditional above (L48) and if that is False we don't even have a `do_log` variable.

--

I'm a bit conflicted on what's being asked for direction now. I thought Tim was OK with not repeating the logging for redundant version mismatches as long as I fix this to log each mismatch once. But now it sounds like you (Min) would like me to log everything where the first mismatch for a particular version combination is at `warning` level and any redundancies thereafter are logged at `debug`, but I don't see what the redundant `debug` logging buys us.

minrk:
Sorry for making you wait! I had to step away for another project for a while. This looks AOK to me.

PR #2972, opened 2020-03-04
Update contributor docs

mriedem:
As a new contributor to jupyterhub it took awhile to get
up and running locally mainly because I didn't have sqlite
installed but also because I was flipping between README,
CONTRIBUTING and the actual contributing docs which are all
a little bit different.

This does a few things:

- Updates the contributor sphinx docs to mention that how
  one chooses to isolate their development environment is
  up to them with a link to the detailed forum thread on
  that topic.
- Updates the contributor sphinx docs to mention sqlite and
  database setup in general. While in here some trailing
  whitespaces are cleaned up.
- Leave a comment in CONTRIBUTING.md about the redundant
  information in the docs on getting a development environment
  setup. Long-term we should really get those merged so there
  is a single authoritative document on how to get a dev env
  setup for contributing to jupyterhub.
- Link to the jupyterhub gitter channel for asking questions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,6 +8,12 @@ for a friendly and welcoming collaborative environment.
 
 ## Setting up a development environment
 
+<!--
+https://jupyterhub.readthedocs.io/en/stable/contributing/setup.html
+contains a lot of the same information. Should we merge the docs and
+just have this page link to that one?
+-->
+
 JupyterHub requires Python >= 3.5 and nodejs.
 
 As a Python project, a development install of JupyterHub follows standard practices for the basics (steps 1-2).
@@ -128,4 +134,4 @@ To read more about fixtures check out the
 [pytest docs](https://docs.pytest.org/en/latest/fixture.html)
 for how to use the existing fixtures, and how to create new ones.
 
-When in doubt, feel free to ask.
+When in doubt, feel free to [ask](https://gitter.im/jupyterhub/jupyterhub).
--- a/docs/source/contributing/setup.rst
+++ b/docs/source/contributing/setup.rst
@@ -8,7 +8,7 @@ System requirements
 ===================
 
 JupyterHub can only run on MacOS or Linux operating systems. If you are
-using Windows, we recommend using `VirtualBox <https://virtualbox.org>`_ 
+using Windows, we recommend using `VirtualBox <https://virtualbox.org>`_
 or a similar system to run `Ubuntu Linux <https://ubuntu.com>`_ for
 development.
 
@@ -18,7 +18,7 @@ Install Python
 JupyterHub is written in the `Python <https://python.org>`_ programming language, and
 requires you have at least version 3.5 installed locally. If you haven’t
 installed Python before, the recommended way to install it is to use
-`miniconda <https://conda.io/miniconda.html>`_. Remember to get the ‘Python 3’ version, 
+`miniconda <https://conda.io/miniconda.html>`_. Remember to get the ‘Python 3’ version,
 and **not** the ‘Python 2’ version!
 
 Install nodejs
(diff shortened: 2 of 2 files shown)


betatim:
One question: I thought that sqlite was shipped as part of Python or listed as a dependency of the respective python package (conda, apt, etc). Is that true or have I just been lucky that it always got installed because of something else?

In case it is a pitfall on a particular OS/setup we could add an explicit mention of that in the note. "woah look out on XXX you will have to do this! On other platforms you might have to do this."

Otherwise LGTM

minrk:
Thanks for the improvements! As @betatim mentioned, let's leave out the sqlite3 install step. That should ~never be required.

mriedem:
> One question: I thought that sqlite was shipped as part of Python or listed as a dependency of the respective python package (conda, apt, etc). Is that true or have I just been lucky that it always got installed because of something else?
> 
> In case it is a pitfall on a particular OS/setup we could add an explicit mention of that in the note. "woah look out on XXX you will have to do this! On other platforms you might have to do this."
> 
> Otherwise LGTM

Not as far as I know, and it would seem weird to me if the ubuntu python package required sqlite since that could set a weird precedent, e.g. assuming the backend DB that people are going to use with their python apps.

I'm technically using lubuntu 18.04 so maybe it is a difference in the default binaries installed between lubuntu and for example ubuntu desktop or server but I think either way it's worthwhile to mention that you should make sure to install the native binary for the default database in the hub.

rkdarst:
isn't Ubuntu `sqlite3` just the command line utility?  If it did depend on anything, it should be `libsqlite3-0`, since "all" usage is embedded by linking to the library.  But most likely, it's embedded within Python itself.  I just did a test: `strace -e file -o strace python3 -c 'import sqlite3'`, and I see it loads from `/usr/lib/python3.7/lib-dynload/_sqlite3.cpython-37m-x86_64-linux-gnu.so` which is in the `libpython3.7-stdlib` package.  I see it also loads my system one, maybe it also tries that in case it's useful.

Anyway, I always have the command line utility installed too, since it's useful.  But probably almost never a dependency.

... all the above is untested, and could be wrong.

docs/source/contributing/setup.rst
```diff
@@ -93,21 +99,33 @@ happen.
...
-5. Install the development version of JupyterHub. This lets you edit
+5. Setup a database.
+
+   The default database is ``sqlite`` so if you are just trying
+   to get up and running quickly for local development remember to install
+   the native package for your distribution, for example on Ubuntu:
+
+   .. code:: bash
+
+      sudo apt-get install sqlite3
```
mriedem: Should I just remove this part starting with the "for example..."? The "install the native package for your distribution" is generic enough I think.

manics:
The `sqlite3` Python module has been part of the standard Python library since 2.5:
https://docs.python.org/2.7/library/sqlite3.html

Did you run into problems when the CLI `sqlite3` wasn't installed?

manics:
I've just verified that it's possible to import the Python module without the CLI being present:
```
$ podman run -it --rm python:3.8.2-alpine sqlite3 --help
Error: executable file not found in $PATH: No such file or directory: OCI runtime command not found error
```
```
$ podman run -it --rm python:3.8.2-alpine python -c 'import sqlite3; print(sqlite3.__file__)'
/usr/local/lib/python3.8/sqlite3/__init__.py
```

mriedem:
> The `sqlite3` Python module has been part of the standard Python library since 2.5 (I have to admit I wasn't aware of that either!):
> https://docs.python.org/2.7/library/sqlite3.html

Me neither. 😃 

> Did you run into problems when the CLI `sqlite3` wasn't installed?

To be honest I was running into a few problems with my initial attempt at setting up a development environment, probably because I wasn't starting within a virtual environment and was installing everything globally into my dev VM.

I can try removing the `sqlite3` package and see if I have any issues running the tests.

mriedem:
> I can try removing the `sqlite3` package and see if I have any issues running the tests.

OK yeah that made no difference in running the tests or starting up `jupyterhub` without any `jupyterhub_config.py` file in the project root directory. I also renamed the `jupyterhub.sqlite` file that I had before restarting the service and a new one was created.

Note that I had to do both of the above in the virtualenv I created since that's where all of the jupyterhub dependencies are now installed on my VM.

docs/source/contributing/setup.rst
mriedem: Just as an FYI this is what I have installed now for lubuntu 18.04 (and I think these were part of the base image I used to create this VM):

```
$ dpkg -l | grep sqlite
ii  libsqlite0                             2.8.17-14fakesync1                              amd64        SQLite 2 shared library
ii  libsqlite3-0:amd64                     3.22.0-1ubuntu0.2                               amd64        SQLite 3 shared library
ii  libsqlite3-dev:amd64                   3.22.0-1ubuntu0.2                               amd64        SQLite 3 development files
```

CONTRIBUTING.md
```diff
@@ -8,6 +8,12 @@ for a friendly and welcoming collaborative environment.
 
 ## Setting up a development environment
 
+<!--
```
willingc: Feel free to add a link here to the Documentation hosted on RTD. I could see this `CONTRIBUTING.md` being minimal docs targeted for experienced developers where the more thorough docs could be on RTD.

PR #2978, opened 2020-03-10
SpawnHandler POST with user form options displays the spawn-pending page

danlester:
Previous behavior: 

If there is no form for user options before spawning, SpawnHandler in GET kicks off the spawning and then redirects to the /spawn-pending/ page to show the user the progress stream.

If there is a form, SpawnHandler in POST starts the spawn but then just awaits its finish before redirecting to the hopefully-running server. It doesn’t go via /spawn-pending/ in this case.

New behavior:

spawn-pending is displayed in both cases, whether a user options form was involved or not.

### DIFF
--- a/jupyterhub/handlers/pages.py
+++ b/jupyterhub/handlers/pages.py
@@ -151,17 +151,7 @@ class SpawnHandler(BaseHandler):
 
         spawner = user.spawners[server_name]
 
-        # resolve `?next=...`, falling back on the spawn-pending url
-        # must not be /user/server for named servers,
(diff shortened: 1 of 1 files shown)


meeseeksmachine:

https://discourse.jupyter.org/t/why-does-spawner-not-go-via-spawn-pending-when-there-are-options-from-form/3546/2

minrk:
Thanks for the PR! If you run `black` in the repo to autoformat the code, everything should be happy and we can merge.

danlester:
Oh yes, black has been run now. Thank you!

PR #2980, opened 2020-03-11
Support JupyterLab with LocalSpawner

echarles:
Some users want to get jupyterlab by default and the recommended way is [documented here](https://jupyterlab.readthedocs.io/en/stable/user/jupyterhub.html) with `c.Spawner.default_url = '/lab'`.

This works fine for e.g. `DockerSpawner` but I have found it failing for the basic `LocalSpawner`. When I set c.Spawner.default_url = '/lab' with `LocalSpawner`, the notebook process is launched but return a `404` on the `lab` url. Then manually changing to `tree` makes it work and the classic notebook is shown.

Looking at the code, this sound logical as the `singleuser.py` contains `class SingleUserNotebookApp(NotebookApp)` which launches a classic notebook, and not a jupyterlab.

This PR address this with a simple change `class SingleUserNotebookApp(LabApp)` - The rest of the  `singleuser.py` remains unchanged and still uses the NotebookApp nomenclature.

I have found this change to make the jlab/localspawner tandem work.

### DIFF
--- a/jupyterhub/singleuser.py
+++ b/jupyterhub/singleuser.py
@@ -38,8 +38,12 @@ from traitlets import (
     TraitError,
 )
 
+try:
+    from jupyterlab.labapp import LabApp as App
+except ImportError:
+    from notebook.notebookapp import NotebookApp as App
+
 from notebook.notebookapp import (
-    NotebookApp,
     aliases as notebook_aliases,
     flags as notebook_flags,
 )
@@ -216,7 +220,7 @@ def _exclude_home(path_list):
             yield p
 
 
-class SingleUserNotebookApp(NotebookApp):
+class SingleUserNotebookApp(App):
     """A Subclass of the regular NotebookApp that is aware of the parent multiuser context."""
 
     description = dedent(
@@ -230,7 +234,7 @@ class SingleUserNotebookApp(NotebookApp):
     examples = ""
     subcommands = {}
     version = __version__
-    classes = NotebookApp.classes + [HubOAuth]
+    classes = App.classes + [HubOAuth]
 
     # disable single-user app's localhost checking
     allow_remote_access = True


manics:
Have you thought about whether this should be a mandatory dependency for JupyterHub, or optional?
E.g. something like
```python
try:
    from jupyterlab.labapp import LabApp as App
except ImportError:
    from notebook.notebookapp import NotebookApp as App
```

echarles:
@manics That makes a lot of sense. Thx a lot for pointing me to that.

The last commit brings those less disruptive change and it does the work. CI remains green.

betatim:
I think this change means that lab will be used if it is installed, no matter what. So I think we should change this so that it only happens if users request lab, as before.

minrk:
It shouldn't matter if you are using DockerSpawner or LocalProcessSpawner, since the code to launch the single-user server is the same. It's likely a difference of environment.

I can also confirm that the existing config works with current jupyterhub (1.1.0) and jupyterlab (2.0.1):

```python
c.Spawner.default_url = '/lab'
```

<details>

```
$ pip list
Package            Version   
------------------ ----------
alembic            1.4.1     
appnope            0.1.0     
async-generator    1.10      
attrs              19.3.0    
backcall           0.1.0     
bleach             3.1.1     
certifi            2019.11.28
certipy            0.1.3     
cffi               1.14.0    
chardet            3.0.4     
cryptography       2.8       
decorator          4.4.2     
defusedxml         0.6.0     
entrypoints        0.3       
idna               2.9       
importlib-metadata 1.5.0     
ipykernel          5.1.4     
ipython            7.13.0    
ipython-genutils   0.2.0     
jedi               0.16.0    
Jinja2             2.11.1    
json5              0.9.2     
jsonschema         3.2.0     
jupyter-client     6.0.0     
jupyter-core       4.6.3     
jupyter-telemetry  0.0.5     
jupyterhub         1.1.0     
jupyterlab         2.0.1     
jupyterlab-server  1.0.7     
Mako               1.1.2     
MarkupSafe         1.1.1     
mistune            0.8.4     
nbconvert          5.6.1     
nbformat           5.0.4     
notebook           6.0.3     
oauthlib           3.1.0     
pamela             1.0.0     
pandocfilters      1.4.2     
parso              0.6.2     
pexpect            4.8.0     
pickleshare        0.7.5     
pip                20.0.2    
prometheus-client  0.7.1     
prompt-toolkit     3.0.4     
ptyprocess         0.6.0     
pycparser          2.20      
Pygments           2.6.1     
pyOpenSSL          19.1.0    
pyrsistent         0.15.7    
python-dateutil    2.8.1     
python-editor      1.0.4     
python-json-logger 0.1.11    
pyzmq              19.0.0    
requests           2.23.0    
ruamel.yaml        0.16.10   
ruamel.yaml.clib   0.2.0     
Send2Trash         1.5.0     
setuptools         46.0.0    
six                1.14.0    
SQLAlchemy         1.3.15    
terminado          0.8.3     
testpath           0.4.4     
tornado            6.0.4     
traitlets          4.3.3     
urllib3            1.25.8    
wcwidth            0.1.8     
webencodings       0.5.1     
wheel              0.34.2    
zipp               3.1.0     
```

</details>

without any changes. Note that you must have run `jupyter serverextension enable jupyterlab` for this approach to work. It's likely that this is the difference between your local environment and the one used by default in DockerSpawner.

The alternative is to use:

```python
c.Spawner.cmd = 'jupyter-labhub'
```

to invoke the jupyterlab application wrapper instead.

echarles:
@minrk The trick is indeed `jupyter serverextension enable jupyterlab`. With that, lab is working fine.

As this is an optional step which I do not apply since some time, it would make sense to document at least on https://jupyterlab.readthedocs.io/en/stable/user/jupyterhub.html

Any other place we should mention that?

minrk:
I think that's probably the main place. I think it may also be time to promote jupyterlab support to first-class so that we can have switches based on a single `c.Spawner.jupyterlab = True`

echarles:
Will open a PR on jlab repo to document that.

Regarding `c.Spawner.jupyterlab = True`, do you mean the spawner should enable the server extension via code if the user has not done that?

echarles:
@minrk Looking ahead, I am working with @Zsailer to have jupyterlab be a [server extension](https://jupyter-server.readthedocs.io/en/latest/frontends.html) (work done in https://github.com/jupyterlab/jupyterlab/pull/7416 and https://github.com/jupyterlab/jupyterlab_server/pull/79).

At some point, Lab will not be anymore a old/current serverextension but an extension in the new jupyter-server world. In that new world, I am not sure that the command `jupyter serverextension enable jupyterlab` would have any effect at all (to be tested...).

echarles:
PR for the jupyterlab documentation is opened https://github.com/jupyterlab/jupyterlab/pull/8036

minrk:
> do you mean the spawner should enable the server extension via code if the user has not done that?

Yeah, I mean promoting the responsibility to know how to enable jupyterlab properly to jupyterhub code, wherever that might be, rather than the passthroughs to the underlying code that we have now. That can mean modifying the launch command / class instantiated / default url or other things, as jupyterlab evolves.

echarles:
Opened https://github.com/jupyterhub/jupyterhub/issues/2984 as followup of this PR.

Closing this with Won't fix as we have other plans (docs, config...)

Issue #2958, opened 2020-02-28
ReadTheDocs builds are failing


manics:
The [docs builds](https://readthedocs.org/projects/jupyterhub/builds/) have been failing since [Feb. 26, 2020. 1:05 p.m](https://readthedocs.org/projects/jupyterhub/builds/10508207/). This means https://github.com/jupyterhub/jupyterhub/pull/2956 which should've fixed https://github.com/jupyterhub/jupyterhub/issues/2955 hasn't been deployed

[Most recent failure](https://readthedocs.org/projects/jupyterhub/builds/10520408/):
```
[rtd-command-info] start-time: 2020-02-28T10:48:22.704838Z, end-time: 2020-02-28T10:48:35.821550Z, duration: 13, exit-code: 1
conda env create --quiet --name latest --file docs/environment.yml
Collecting package metadata (repodata.json): ...working... failed

CondaHTTPError: HTTP 500 CONNECTION FAILED for url <https://metachannel.conda-forge.org/conda-forge/jupyterhub,sphinx,recommonmark/noarch/repodata.json>
Elapsed: 00:00.420768
CF-RAY: 56c1d36d1d42ccd6-EWR

A remote server error occurred when trying to retrieve this URL.

A 500-type error (e.g. 500, 501, 502, 503, etc.) indicates the server failed to
fulfill a valid request.  The problem may be spurious, and will resolve itself if you
try your request again.  If the problem persists, consider notifying the maintainer
of the remote server.
```

betatim:
Locally `wget https://metachannel.conda-forge.org/conda-forge/jupyterhub,sphinx,recommonmark/linux-64/repodata.json` takes a moment (10s or so??) to start sending a reply by works (now).

Here the status 500 came very quickly (~0.4s), could it be that network access on RTD is firewalled?

willingc:
@manics I just double-checked RTD. You are correct that there were build failures 3 weeks ago. I just checked and it is working properly. I'm going to close. Please let me know if this has persisted for you.

willingc:
@GeorgianaElena May be the same error but unrelated. PyPI was having issues this morning which may be affecting RTD and even conda-forge.

I tried wiping the container and rebuilding. Still getting a 500 error. Do you have a RTD account? I could add you as a maintainer there.

GeorgianaElena:
Thanks for investigating @willingc :heart:

My handle on RTD is *GeorgianaElena*

willingc:
[...]

PR #3010, opened 2020-04-01
Use pip instead of conda for building the docs on RTD

GeorgianaElena:
As mentioned in [this comment](https://github.com/jupyterhub/jupyterhub/issues/2958#issuecomment-606483892), the metachannel service is down. :confused: 

Using the official conda-forge channel is not an option because it uses [a lot of memory](https://github.com/conda/conda/issues/5003) and makes [RTD unhappy](https://readthedocs.org/projects/georgiana-jh-docs/builds/10734292/). As [@manics mentioned](https://github.com/jupyterhub/jupyterhub/issues/2958#issuecomment-606590081), RTD recommends using pip instead of conda where possible.

However, because we need nodejs 12 to build our rest-api docs (this requirement comes from bootprint), and the latest RTD docker image installs an older nodejs version (nodejs 8), I had to pin `bootprint` back to 1.0.0 to make it work with the RTD nodejs. This way, we don't need conda anymore to install a newer version of nodejs.
 
Fixes https://github.com/jupyterhub/jupyterhub/issues/2958


docs/requirements.txt
```diff
@@ -1,10 +1,16 @@
 # ReadTheDocs uses the `environment.yaml` so make sure to update that as well
 # if you change this file
```
betatim, line 2: Shouldn't we remove this comment because we now don't use conda any more?

docs/requirements.txt
```diff
@@ -1,10 +1,16 @@
...
--r ../requirements.txt
```
betatim, line 3: For my education: why can we no longer reference the requirements file from the parent directory?

readthedocs.yml
```diff
@@ -1,11 +1,19 @@
-name: jupyterhub
-type: sphinx
```
betatim, line 2: Did you remove these on purpose or accidentally/

docs/requirements.txt
GeorgianaElena: We can still reference it, but we don't need all the requirements of the project for the docs, so I removed it to speed up a bit the docs build.

betatim: Ah ok! That sounds like a good thing to do :)

readthedocs.yml
GeorgianaElena: This is on purpose, as in RTD config version 2, the name setting is [per project and for per version anymore](https://docs.readthedocs.io/en/stable/config-file/v2.html#migrating-from-the-web-interface) and [sphinx got its own setting.](https://docs.readthedocs.io/en/stable/config-file/v2.html#sphinx)

docs/requirements.txt
GeorgianaElena: Oh, right. I'll remove it now.

docs/requirements.txt
minrk: I think we should still get all the requirements from the package to ensure that autodoc populates correctly. This also helps reduce maintenance burden because ~*every* time we add a real dependency to the package the docs builds break until we add a duplicate to docs/requirements.txt. This has happened several times across projects before adding the package requirements to the build.

minrk:
👍 other than putting back the install of jupyterhub itself to keep things simple and consistent.

docs/requirements.txt
GeorgianaElena: This makes sense, @minrk, I'll put back the package requirements.

betatim:
Merged! Thanks for spending time on getting the docs into good shape :)

PR #2726, opened 2019-09-09
Jupyter(Hub) conceptual intro

rkdarst:
At the JupyterHub/BinderHub workshop, one of our ideas was to make a conceptual intro to JupyterHub so that people could know what it does, and in particular what it doesn't do (what is handled by other components).  We get many issues that end up misdirected or that have a root cause of not understanding what the components are.

This PR is my initial draft - comments welcome.  I've written it informally and with a certain opinion - it's supposed to be like a teaching giving a first lesson, and not technical reference.  It goes far beyond just JupyterHub, but the reason for having it here is that you *really* need to start knowing this stuff once you start administrating a JupyterHub, before that you can sort of get by without a perfect mental model.

Issues I know of:
- [ ] What is a proper title?
- [x] Where should it go under the table of contents?
- [ ] I used the very verbose "single-user notebook server", I'd like a shorter term that still avoids ambiguity.
- [x] All parts need fact-checking, because I haven't done that yet
- [x] After the content is somewhat stable, we need to add links and references.
- [x] Intro needs improvement and we should make sure it explains the purpose well
- [x] Conclusion needs improvement and guides to what is next

PR #3015, opened 2020-04-08
Fix user_row endblock in admin template

jtpio:
It looks like the `user_row` block was ended right after the `</tr>` element, while it is defined right after the corresponding `<tr>` element above:

https://github.com/jupyterhub/jupyterhub/blob/1f515464fe2edf5d9d8a7617c22ebe67002e0ba0/share/jupyterhub/templates/admin.html#L43-L46

This came up when trying to extend the admin template to add extra information to the user row.

### DIFF
--- a/share/jupyterhub/templates/admin.html
+++ b/share/jupyterhub/templates/admin.html
@@ -96,8 +96,8 @@
           <a role="button" class="delete-server btn btn-xs btn-warning">delete server</a>
         {%- endif -%}
       </td>
-      </tr>
       {% endblock user_row %}
+      </tr>
       {% endfor %}
       {% endfor %}
   </tbody>


betatim:
cc @manics who I think did some work on making our docs build more strict. Maybe you have an idea why this comes up now and if the fix is easy. (The error message is not mega clear to me without opening the mentioned files etc)

manics:
I only changed z2JH https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1611

A new major version of Sphinx (3.0.0) was released on 5 April https://pypi.org/project/Sphinx/3.0.0/ so perhaps it's gotten stricter?

betatim:
https://www.sphinx-doc.org/en/master/changes.html I can spy anything obvious in the change log :-/

jtpio:
> `use :noindex: for one of them`

Maybe we could try that? (from the error message)

rkdarst:
We found a little bit more here: https://github.com/jupyterhub/jupyterhub/pull/2726 (I guess we should make a new issue about it...)

betatim:
Oh wow, I clicked the "update branch" button the GitHub UI in order to find out what it does...turns out the button doesn't ask for confirmation it just does its thing. This is how we got `master` merged into this branch. I had originally planned to rebase this branch on latest master to see if that fixes the builds. (merge master into this branch has AFAIK the same effect as rebasing just less pretty)

Let's see what happens.

betatim:
All the robots are happy, so merging this. Thanks for the patience and to all those who helped debugging things.

jtpio:
Thanks!

PR #3019, opened 2020-04-15
Remove unused imports

stv0g:
There are a bunch more unused variables a imports.

We might want to add pyflakes to the CI


consideRatio:
@stv0g thank you for your, this LGTM. It is still marked as a draft though, what do you think?

stv0g:
@consideRatio I think this is ready to be merged.. I just never managed to fully check all imports.. But I would suggest to continue this effort lateron in other PRs.

consideRatio:
@stv0g I agree! Thank you, I know I end up being confused by a unused import sometimes as they are misleadingly indicative of what goes on in the code file.

PR #3020, opened 2020-04-15
Support kubespawner running on a IPv6 only cluster

stv0g:
This is a relatively minor change which checks if the `(ip, port)` pair of the singleuser container returned by `Spawner.start()` is an IPv6 address.

If its the case, we need to enclose the IP address in brackets (`http://[ipv6]:port`) before passing it to `urlparse` because otherwise the part IPv6 address would be mistakenly treated as a port number.

We detect the presence of an IPv6 by check if it contains a colon.

PR #3021, opened 2020-04-16
Fix docs CI test failure: duplicate object description

rkdarst:
We've been seeing this warning in docs build:
```
Warning, treated as error:
/l/data/git/jupyterhub/jupyterhub/auth.py:docstring of jupyterhub.auth.Authenticator.admin_users:1:duplicate object description of jupyterhub.auth.Authenticator.admin_users, other instance in api/auth, use :noindex: for one of
 them
```

Basic discussion is in #2726.  The problem seems to comes from autodocs-traits: https://github.com/jupyterhub/autodoc-traits/pull/1

This patches requiremens.txt to fix it as a demo, then tries to fix some of the actual errors.  But... it quickly turns into what seems to a large game of fixing many different errors, it seems most of them are at the interface of rst/md: some of the worst are caused by code blocks in markdown, some of which evaluate rst.

I'm pushing this now for others to explore, but not everything is fixed yet.  Anyone, feel free to push to this branch (but I might force-push it myself, too, so pull/push often).

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,7 +1,9 @@
 -r ../requirements.txt
 
 alabaster_jupyterhub
-autodoc-traits
+# Temporary fix of #3021. Revert back to released autodoc-traits when
+# 0.1.0 released.
+https://github.com/jupyterhub/autodoc-traits/archive/75885ee24636efbfebfceed1043459715049cd84.zip
 pydata-sphinx-theme
 recommonmark>=0.6
 sphinx-copybutton
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -20,6 +20,7 @@ extensions = [
     'autodoc_traits',
     'sphinx_copybutton',
     'sphinx-jsonschema',
+    'recommonmark',
 ]
 
 templates_path = ['_templates']
@@ -65,8 +66,6 @@ def setup(app):
     app.add_css_file('custom.css')
     app.add_transform(AutoStructify)
 
-
-source_parsers = {'.md': 'recommonmark.parser.CommonMarkParser'}
 
 source_suffix = ['.rst', '.md']
 # source_encoding = 'utf-8-sig'
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -62,7 +62,7 @@ from recommonmark.transform import AutoStructify
 
 def setup(app):
     app.add_config_value('recommonmark_config', {'enable_eval_rst': True}, True)
-    app.add_stylesheet('custom.css')
+    app.add_css_file('custom.css')
     app.add_transform(AutoStructify)
 
 


manics:
I also ran into problems with RST vs MD when fixing from broken URLs in Z2JH https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1610

For example, if everything is in RST you can use internal Sphinx cross references and you'll get a warning/error if they incorrect. I don't think you can do that with Markdown, so you end up using an absolute URL. Obviously the problems here are different, but do you think converting everything to RST would help?

rkdarst:
I think converting everything to RST would help, but almost all the docs are md, so that's probably too big a task to reasonably do now (especially as a side-effect of something else).  (Though personally, very strongly prefer rst for big stuff like this).

At least part of the problems were caused by some markdown code block with parser `eval_rst`, which I suppose embeds rst within markdown - so that it can the the cross-linking.  It appeared to be valid markdown (but which flavor...?) and valid rst, but the combination made some error.

I could do something more, but I'm not sure what's the right tradeoff for the solution.  It would be good to have someone with decision making respect decide a strategy, then decide what to do.

willingc:
Historically, we had used Markdown because @minrk preferred and expedience won out :smile:

At this point, converting to rst makes good sense. Pandoc does a reasonable job. If used on a section at a time and visual editing after the conversion, it should be a smooth transition.

I would also recommend trying https://sphinx-autoapi.readthedocs.io/en/latest/ instead of autodoc extension. I've been using for over 2 months on another (proprietary) project, and I find it a big improvement over autodoc.

rkdarst:
Really, I'm sort of leading towards removing the `-W` to be able to proceed and make progress.  This isn't a failure in build, but seems to be a change that affects the suitability of `-W` to the project.  But I'm not quite sure what changed or when - I tried downgrading sphinx, docutils, and recommonmark (independently), it still doesn't work.  In fact, the rabbit hole keeps getting deeper...

Incidentally, I recently changed my big sphinx docs site to give some more warnings on errors.  But `-W` was too much, there were too many unimportant warnings that didn't have a reasonable fix or ignore for us.  In the end, I made my own Python script that would search for warnings I wanted to be errors and cause github actions failures on that:
  https://github.com/AaltoScienceIT/scicomp-docs/blob/master/meta/check-warnings.py
integrated with github actions.

willingc:
Feel free to remove -W on CI. I had it in more for my own local development @rkdarst.

stv0g:
Any Ideas why the [current Travis](https://travis-ci.org/github/jupyterhub/jupyterhub/jobs/675773176) builds are failing?

Same seems to happen in #3020

rkdarst:
OK, rabbit hole got shallower.  `source_parsers` is removed in sphinx 3.0, and instead `recommonmark` should be added as an extension - which it wasn't.  I presume that caused all markdown parsing to fail (I guess it was being parsed as rst, no wonder it seemed so bad!).  Then, we also need the new autodoc-traits.  I also fixed the add_stylesheet method (add_css_file added in sphinx 1.4).  And we keep `-W` which is probably good overall.

This seems to let the builds pass!  I propose minimal changes for now to make things pass, then more later.

(the build still failed, but it seemed unrelated - trying again).

rkdarst:
@stv0g, that seems to be unrelated and we got the same problem here.  That's more complicated and I'm not a specialist in that area, though...

willingc:
@rkdarst @stv0g There is a test of spawn pending that is currently failing which is unrelated to docs work.

betatim:
If we at all can keep `-W` I'd like to keep it. I am ok with temporarily disabling it to get ourselves unstuck. My thinking is that it is better (like "eat your vegetables" is) to deal with its occasional finger waggling when you are trying to make a quick&small change. The alternative being that once every few months we are so tangled up that we need a major operation to get ourselves untangled.

Looking at the spawn failure.

rkdarst:
I think keeping `-W` is reasonable now and a good idea, my previous analysis about removing it turned out to be wrong.

I just force pushed, to have requirements.txt now point to jupyterhub/autodocs-traits master instead of my personal repo.  I think this is reasonable to go now, still will need to be changed to autodocs-traits release later.

docs/requirements.txt
```diff
@@ -1,7 +1,9 @@
 -r ../requirements.txt
 
 alabaster_jupyterhub
-autodoc-traits
+# Temporary fix of #3021, revert back to released autodoc-traits when
+# possible.
+https://github.com/jupyterhub/autodoc-traits/archive/master.zip
```
betatim, line 6: Maybe use the SHA of the commit on master right now. That way we are "pinned" to a known state. In case someone pushes more things to master in autodoc-traits. What do you think?

betatim:
I can't work out why the spawn test is failing. It is looking for a particular bit of text inside a `script` tag. From looking at the template (`spawn_pending.html`) that code should be there. If I run the same test from `master`  on a local checkout it (of course) doesn't fail. I've restarted the build again in the hopes of making it go away but I doubt it will. Fresh out of ideas for now. Not familiar enough with BeautifulSoup to know what happens if there are multiple `<script>` tags and a test like https://github.com/jupyterhub/jupyterhub/blob/1f515464fe2edf5d9d8a7617c22ebe67002e0ba0/jupyterhub/tests/test_pages.py#L356-L358 (which is where the failure is)

Of the recent PR only #2978 looks like it might potentially be relevant.

GeorgianaElena:
@betatim, I managed to make the test fail locally too by running:
```
pip install --upgrade --pre -r dev-requirements.txt .
```
Which among others, bumped the version of beautifulsoup to 4.9.0, so there might be a breaking change somewhere.

twalcari:
Yes, it is a breaking change in Beautifulsoup 4.9 indeed. As suggested by the maintainer [here](https://bugs.launchpad.net/beautifulsoup/+bug/1871335), you can fix it by changing `.text` to `.string` in our test.

rkdarst:
> Yes, it is a breaking change in Beautifulsoup 4.9 indeed. As suggested by the maintainer [here](https://bugs.launchpad.net/beautifulsoup/+bug/1871335), you can fix it by changing `.text` to `.string` in our test.

Great find!  Is someone working on this or do you want me to do it here?

GeorgianaElena:
@twalcari, do you want to open a PR to fix the failing spawn test? :rocket:

twalcari:
I created pull request #3025 to address this issue.

rkdarst:
It's green!  Thanks everyone for helping to figure this out.  It's a great day today...

willingc:
Thanks everyone!

PR #3025, opened 2020-04-17
Fix broken test due to BeautifulSoup 4.9.0 behavior change

twalcari:
cfr. https://bugs.launchpad.net/beautifulsoup/+bug/1871335 and PR#3021


GeorgianaElena:
Yaay :tada: Thanks @twalcari!

PR #3026, opened 2020-04-17
Always pass config object to render_template

danlester:
Adds 'config' as a standard Jinja2 template var for all request handlers. This means that custom configuration can be accessed from inside a custom template html file.

So e.g. if I replace home.html in my own templates folder, I can locate custom config settings, or even vary behavior based on standard config settings.

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -1132,6 +1132,7 @@ class BaseHandler(RequestHandler):
             static_url=self.static_url,
             version_hash=self.version_hash,
             services=self.get_accessible_services(user),
+            config=self.config,
         )
         if self.settings['template_vars']:
             ns.update(self.settings['template_vars'])


consideRatio:
What kind of config I asked myself.

https://github.com/jupyterhub/jupyterhub/blob/50e6d451d59f70077336aa74d0a2af8bfefcfd04/jupyterhub/handlers/base.py#L97-L99

Which reference: https://www.tornadoweb.org/en/stable/web.html#tornado.web.RequestHandler.settings

Which reference: https://www.tornadoweb.org/en/stable/web.html#tornado.web.Application.settings

That doesn't mention "config" as a setting. Is this a JupyterHub set setting called config?

danlester:
Yes, that is the JupyterHub 'config' object - everything loaded from the command line args and/or the jupyterhub_config.py config file. (Ultimately, None if this wasn't set for some reason, but I don't think that should happen under normal JupyterHub operation.)

manics:
Are these template functions purely on the JupyterHub side, or is it possible for a template to be inherited by something on the client side which would leak access to any secrets in config? Similarly if there's an error is there any possibility of all variables being dumped and becoming visible to the client?

danlester:
Good questions... There is certainly no way to access the Jinja2 template context from the client side - if the template html and functions do not access the 'config' variable then there can be no leak. So yes, this is all JupyterHub server side.

There is the possibility that the JupyterHub admin has written their own templates that could blindly dump out all template vars. The most likely place is a custom error page as you suggest. There is a standard Jinja2 extension called debug that would most likely be used by enabling in jupyterhub_config.py:
```c.JupyterHub.jinja_environment_options = {'extensions': ['jinja2.ext.debug']}```

And then using the ```{% debug %}``` tag in the debug error template.

That doesn't give too much detail:

```
{'context': {'base_url': '/',
             'config': {'Authenticator': {...},
                        'ConfigurableHTTPProxy': {...},
                        'DockerSpawner': {...},
                        'JupyterHub': {...},
                        'Spawner': {...}}, ...
```

Of course, it would be possible that the admin has written a function to traverse the entire tree which could spit out secrets too. One imagines that would not be available in production...

Please do run this past anyone who has been more involved in either the security or templating side of JupyterHub development if you think it needs more thought!

By the way, it is already possible for the adventurous admin to pass most of the config (although missing anything from the command line) through to the templates with this:
```c.JupyterHub.template_vars = {'config': c}```
so maybe we have to accept we're already giving them the freedom to be dangerous!

manics:
Closed/reopened to retrigger travis with https://github.com/jupyterhub/jupyterhub/pull/3021 to fix the docs.

@danlester Thanks for the additional info, it's been very helpful for me. I'll defer to someone else on the decision of whether to merge or not.

danlester:
Great - no problem if it doesn't make sense to merge. I have worked around it, but would have used this PR if it was available. It would mainly be useful for SingletonConfigurables - it just means you can use the same config within your own extra_handlers as well as your template extensions to the main built-in JupyterHub handlers.

minrk:
I don't think direct access of the traitlets Config object is something that should be encouraged. Config is pretty dynamic and should generally be consumed only by Configurables that resolve things like inheritance, validation, and dynamic and default values. Direct access to Config is a recipe for getting not-useful LazyConfigValues unless you're pretty careful. Accessing the config through the instantiated objects should be a more reliable way to get at the true values of any configuration, and gives you access to default values, ensuring values are defined, etc.

What should be a more robust path to the same goal is making it easy/possible to put *configured objects* into the template namespace. I'm not sure if that's the case at the moment.

danlester:
@minrk Thank you for your input. That makes sense, and I'll close this PR.

Maybe I'll open a separate feature request to find another way to get relevant config.

PR #3027, opened 2020-04-19
Cross site scripting false positive

scivm:
A load balancer in front of my jupyterhub service is adding a :443 into the host header.   This is undesired but valid according to the http spec and causes a false positive when performing the cross site scripting check.  This pull request modifies the value read from the host header to strip the :443 or :80 if a load balancer has added it.   In our case it is an azure load balancer and we cant control it.

https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23

Host = "Host" ":" host [ ":" **port** ] ; Section 3.2.2

Error seen in logs during false positive:

[W 2020-04-17 11:51:29.470 JupyterHub base:60] Blocking Cross Origin API request.  Referer: https://jupyter.myurl.fi/hub/spawn-pending/user, Host: **jupyter.myurl.fi:443**/hub/
[W 2020-04-17 11:51:29.471 JupyterHub log:174] 403 GET /hub/api/users/xxx/server/progress (@myip) 6.40ms

### DIFF
--- a/jupyterhub/apihandlers/base.py
+++ b/jupyterhub/apihandlers/base.py
@@ -39,6 +39,7 @@ class APIHandler(BaseHandler):
 
         - allow unspecified host/referer (e.g. scripts)
         """
+        referer = self.request.headers.get("Referer")
         host_header = self.request.headers.get("Host")
         # http spec allows for port at end of host header but this causes
         # cross site scripting check to give false positive
--- a/jupyterhub/apihandlers/base.py
+++ b/jupyterhub/apihandlers/base.py
@@ -42,8 +42,11 @@ class APIHandler(BaseHandler):
         host_header = self.request.headers.get("Host")
         # http spec allows for port at end of host header but this causes
         # cross site scripting check to give false positive
-        host, port = host_header.split(":", maxsplit=1)
-        if not (port=='443' or port=='80'):
+        if ':' in host_header:
+            host, port = host_header.split(":", maxsplit=1)
+            if not (port=='443' or port=='80'):
+                host = host_header
+        else:
             host = host_header
 
         # If no header is provided, assume it comes from a script/curl.
--- a/jupyterhub/apihandlers/base.py
+++ b/jupyterhub/apihandlers/base.py
@@ -42,11 +42,7 @@ class APIHandler(BaseHandler):
         host = self.request.headers.get("Host")
         # http spec allows for port at end of host header but this causes
         # cross site scripting check to give false positive
-        if (host.endswith(':443')):
-            host = host[:-4]
-        elif (host.endswith(':80')):
-            host = host[:-3]
-        referer = self.request.headers.get("Referer")
+        host, port = host.split(":", maxsplit=1)
 
         # If no header is provided, assume it comes from a script/curl.
         # We are only concerned with cross-site browser stuff here.
--- a/jupyterhub/apihandlers/base.py
+++ b/jupyterhub/apihandlers/base.py
@@ -39,10 +39,12 @@ class APIHandler(BaseHandler):
 
         - allow unspecified host/referer (e.g. scripts)
         """
-        host = self.request.headers.get("Host")
+        host_header = self.request.headers.get("Host")
         # http spec allows for port at end of host header but this causes
         # cross site scripting check to give false positive
         host, port = host.split(":", maxsplit=1)
+        if not (port=='443' or port=='80'):
+            host = host_header
 
         # If no header is provided, assume it comes from a script/curl.
         # We are only concerned with cross-site browser stuff here.
--- a/jupyterhub/apihandlers/base.py
+++ b/jupyterhub/apihandlers/base.py
@@ -40,6 +40,12 @@ class APIHandler(BaseHandler):
         - allow unspecified host/referer (e.g. scripts)
         """
         host = self.request.headers.get("Host")
+        # http spec allows for port at end of host header but this causes
+        # cross site scripting check to give false positive
+        if (host.endswith('.443')):
+            host = host[:-4]
+        elif (host.endswith('.80'))::
+            host = host[:-3]
         referer = self.request.headers.get("Referer")
 
         # If no header is provided, assume it comes from a script/curl.
--- a/jupyterhub/apihandlers/base.py
+++ b/jupyterhub/apihandlers/base.py
@@ -42,9 +42,9 @@ class APIHandler(BaseHandler):
         host = self.request.headers.get("Host")
         # http spec allows for port at end of host header but this causes
         # cross site scripting check to give false positive
-        if (host.endswith('.443')):
+        if (host.endswith(':443')):
             host = host[:-4]
-        elif (host.endswith('.80'))::
+        elif (host.endswith(':80')):
             host = host[:-3]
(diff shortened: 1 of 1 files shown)


betatim:
Seems like a reasonable change to make given it is part of the spec.

Having to "parse" the value of the `Host` header manually with `.endswith()` triggers my "seems fragile and like it might create hard to debug weirdness" senses. Have you investigated if there is something "ready made" that we can use to split the header value into "hostname" and "port"? I don't think https://docs.python.org/3/library/urllib.parse.html contains the answer but it is the equivalent for handling URLs.

Alternatively what do you think of `hostname, port = host_value.split(":", maxsplit=1)` as a way to also handle other ports?

jupyterhub/apihandlers/base.py
```diff
@@ -40,6 +40,12 @@ def check_referer(self):
         - allow unspecified host/referer (e.g. scripts)
         """
         host = self.request.headers.get("Host")
+        # http spec allows for port at end of host header but this causes
+        # cross site scripting check to give false positive
+        if (host.endswith(':443')):
```
betatim, line 45: in general we don't style conditionals with surrounding `()`.

scivm:
I updated the code based on your comments.   I still make changes only to host_headers with 80 and 443 in them.   For example, if someone wants to run on port 8080 they still could.  

It is not in the spec but standard behavior of many clients and servers will truncate the 80 and 443 from a request.    For example if you curl -v https://www.google.com:443, their host header will be www.google.com without the 443.

jupyterhub/apihandlers/base.py
```diff
@@ -39,8 +39,12 @@ def check_referer(self):
 
         - allow unspecified host/referer (e.g. scripts)
         """
-        host = self.request.headers.get("Host")
-        referer = self.request.headers.get("Referer")
+        host_header = self.request.headers.get("Host")
+        # http spec allows for port at end of host header but this causes
+        # cross site scripting check to give false positive
+        host, port = host_header.split(":", maxsplit=1)
+        if not (port=='443' or port=='80'):
```
betatim, line 46: Can you explain a bit why this is needed? `host` will only contain the hostname (which is what you wanted no?) so why set it back to its original value?

jupyterhub/apihandlers/base.py
```diff
@@ -39,8 +39,12 @@ def check_referer(self):
 
         - allow unspecified host/referer (e.g. scripts)
         """
-        host = self.request.headers.get("Host")
-        referer = self.request.headers.get("Referer")
+        host_header = self.request.headers.get("Host")
+        # http spec allows for port at end of host header but this causes
+        # cross site scripting check to give false positive
+        host, port = host_header.split(":", maxsplit=1)
```
betatim, line 45: This line will raise an exception if there is no port in the header value.

jupyterhub/apihandlers/base.py
scivm: I have only seen automatic removal of the port when 80 and 443 are used.  If the user wants to use 8080 then I think they would get an error after the 8080 is removed since we are appending /hub later prior to the comparison.

jupyterhub/apihandlers/base.py
scivm: I will check it.

manics:
I'm slightly concerned by why this is necessary. Do you know if it's Tornado or JupyterHub that's stripping the `80` or `443` in the first place causing the comparison here to fail?

For instance there's a very unlikely edge case where someone runs a `http:80` and `http:443` (***not*** `https:443`) on the same server. These should be treated as different hosts, but here they'd be treated the same. This isn't necessary a blocker to this PR, but I thought I mention it as this this change could affect security so we should at least state our assumptions upfront.

scivm:
@manics  Jupyterhub was not stripping the 80/443 from the hosts header so the comparison against what should have been equal referer failed.   My code strips the 80/443 so that the comparison can succeed.

jupyterhub/apihandlers/base.py
scivm: I look for a : in the host header now prior to try to split it.  According to how I read the spec there can only be 1 :  in the host section.
Host = "Host" ":" host [ ":" port ] ; Section 3.2.2

betatim:
This check compares the value of the referrer and the value of the host headers. Do we understand why the referrer doesn't contain a port but the host header does? That is how I understand the situation you are in @scivm, correct?

manics:
In addition to the above, could you also confirm it's an Azure Load Balancer and not an Azure Application Gateway which includes load balancing? I've been searching Google since this sounds like something that would affect other applications.

scivm:
@betatim I believe it is a standard behavior to remove 80 and 443 ports by clients and servers.  The referer header likely had the port removed in the browser.  It was not every request where the issue was manifesting either.  It would happen during the poll for a user workspace starting and also in the hub control panel when you tried to shutdown your server.

@manics I confirmed that it is Azure Application Gateway which has load balancing and firewall built in.  The load balancer is in default configuration.  The firewall implements a standard OWASP rule set (there are configurable versions but we are using 3.0).  We found for jupyterhub we needed to disable 920230-Multiple URL Encoding Detected, 942370-Detects classic SQL injection probings 2/2, 943120-Possible Session Fixation Attack: SessionID Parameter Name with No Referer for jupyterhub to work.  We are investigating these and they could easily be false positives.

manics:
Could you try modifying the `X-Forwarded-For` header as suggested here:
https://azure.microsoft.com/en-gb/blog/rewrite-http-headers-with-azure-application-gateway/

That might give you a workaround whilst we figure out whether this is OK to merge.

minrk:
> Do we understand why the referrer doesn't contain a port but the host header does?

Referer is set by the browser, the original post says that a Load Balancer is modifying the Host header (in a technically valid, but not normal or common way).

What we really want to check is Origin here, which includes protocol (and thereby port) and why our checks here are fiddly. Unfortunately, browsers don't consistently send Origin (some send Origin for all cross-origin requests, others—firefox—*don't* send them for all cross-origin requests). And we specifically want to use this check to block requests originating from single-user servers, which are not cross-origin and thus usually lack an Origin header. Aside: I think it was a big mistake by browser folks to not send Origin for *all* requests.

Since we want to check origin, we should be getting the *protocol* as well and resolving the port if over-specified:

```python
proto = self.requests.protocol
host = self.request.get_header('Host')
hostname, _, port = host.partition(":")
# elide over-specified default ports
if proto == 'https' and port == '443':
    port = ''
if proto == 'http' and port == '80':
    port = ''
if port:
    port = ':' + port

origin = f"{proto}://{hostname}{port}"
```

Comparison could be done with parsed URL objects, in which case we want to check protocol, host, port, path prefix:

```python
host_url = urlparse(f"{origin}{self.hub.base_url}")
referer_url = urlparse(referer)
if referer_url.scheme != host_url.scheme:
    # fail
if referer_url.hostname != host_url.hostname:
    # fail
# resolve default ports (no need for elide block above if we do this)
referer_port = referer_url.port or (443 if referer_url.scheme == 'https' else 80)
host_port = host_url.port or (443 if host_url.scheme == 'https' else 80)
if referer_port != host_port:
    # fail
if not referer_url.path.startswith(host_url.path):
    # fail
```

Issue #2834, opened 2019-11-25
disable spawn button after press


fmgvalente:
Hello,
I have a custom spawner (a variation of the batchspawner) that takes some time to arrange an environment. After pressing the spawn button, on the interface created from options_form, there is a delay where a user can click that button a few more times, spawning multiple jobs. Is there a way to disable that button after spawning?
Barring that, is it possible to reduce the delay between the button press and the change to the spawning page?

possiblyMikeB:
For what it's worth, we recently ran into a similar issue with users repeatedly clicking the 'sign in' button and then doing the same with the 'start server' button exposed by the options_form related template.. 

The solution we came up with was to add a couple of lines of javascript to 'login.html' and 'spawn.html' which replace the buttons with spinner gifs, and run when the forms are submitted.

betatim:
Sounds like a good improvement. Maybe instead of using a GIF we can find a SVG icon (in fontawesome?) that we can include directly. Do you want to open a PR with a suggestion?

PR #3028, opened 2020-04-20
UI Feedback on Submit

possiblyMikeB:
This pull request is meant to add the UI enhancement requested in issue #2834; it adds a simple mechanism enabling the UI to provide feedback to the user when a form is submitted. 

This is accomplished by adding a pair of css classes (named feedback-container & feedback-widget), as well as, a generic 'onsubmit' handler which interacts with these classes. When a form is submitted, the handler toggles the visibility of the children of an element with class "feedback-container", then adds the class "fa-pulse" to all children of elements with class "feedback-widget".

ui-feedback-onsubmit

### DIFF
--- a/share/jupyterhub/static/less/login.less
+++ b/share/jupyterhub/static/less/login.less
@@ -6,7 +6,7 @@
         .bg-warning();
         padding:10px;
     }
-
+    
     .service-login {
       text-align: center;
       display: table-cell;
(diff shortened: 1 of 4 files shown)


betatim:
What do people think of the idea and looks of this?

What is UX best practice, (for screen readers, mobile, touch displays and the like) in terms of replacing a button with a spinner, replacing the label of a button with a spinner, disabling the button? I'd need to do a bit of research to know the answer.

fmgvalente:
That looks awesome and if it prevents multiple spawn events it fixes an itch of mine I never got to scratch. Disabling the button seems a logical choice to me, since the ''start'' page is then replaced by the 'spawn-pending' page after a few seconds.

minrk:
Looks great to me, if folks want to go ahead. @betatim were you able to find accessibility resources on the spinner?

minrk:
I think we should merge this as a nice improvement, and work on possible refinements in future PRs. Thanks for the contribution, @possiblyMikeB!

PR #3041, opened 2020-05-08
Sudo Environment access to jupyterhub.service

utkarshsaraf19:
Sudo Environment Access to Jupyter Hub as only giving root in user was not working for PAM authentication in ubuntu 18.04

### DIFF
--- a/docs/source/installation-guide-hard.md
+++ b/docs/source/installation-guide-hard.md
@@ -143,7 +143,7 @@ After=syslog.target network.target
 
 [Service]
 User=root
-Environment="PATH=/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/opt/jupyterhub/bin"
+Environment="PATH=/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/usr/bin/sudo:/opt/jupyterhub/bin"
 ExecStart=/opt/jupyterhub/bin/jupyterhub -f /opt/jupyterhub/etc/jupyterhub/jupyterhub_config.py
 
 [Install]


arshanvit:
I made entry of sudo in Environment section of jupyterhub.service as only user=root was not working in PAM authentication for me

docs/source/installation-guide-hard.md
```diff
@@ -143,7 +143,7 @@ After=syslog.target network.target
 
 [Service]
 User=root
-Environment="PATH=/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/opt/jupyterhub/bin"
+Environment="PATH=/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/usr/bin/sudo:/opt/jupyterhub/bin"
```
minrk, line 146: /usr/bin/sudo is a directory containing executables on your system?

arshanvit: Actually I faced peculiar issue.when I was running jupyter hub service from root user.it was not creating session for users.so I provided /usr/bin/sudo to run as "sudo jupyterhub" and it works in my ubuntu environment 18.04

consideRatio:
@utkarshsaraf19 for me to review this PR, I need some more information clarifying the issue this change is meant to solve, and clarifying why this change is the right way to go, and preferably an example of a binary file that is accessible from the added path together with a practical example.

This is a bit outside my comfort zone to review, but that would help this PR go towards a decision as I can then contribute to the review process better.

manics:
This doc was moved to a [separate repo](https://github.com/jupyterhub/jupyterhub-the-hard-way) as it's been lacking in maintenance.

Issue #2925, opened 2020-02-08
"Start my Server" button fails to start spawner automatically


ryogesh:
JupyterHub Version: 1.1.0

Steps to reproduce: Using LocalSpawner and PAMAuthenticator
1) Login using local account
2) Click on control panel and "Stop My Server"
3) We get a button with "Start My Server". On click, we get a notice "Server not running" and another button to "Launch Server"

Expected Behavior:
On step 3, click of "Start My Server", spawner should automatically launch.

Alternate observation:
Instead of clicking "Start my Server" if i click "jupterhub" logo on the top left, spawner launches automatically. i.e. the behavior of /hub/ and /user/<userid>/ is different, hence the differences in UX.

Potential Solution: Change the url for "Start My Server" to /hub/

I am not sure of the implications, and not familiar with the UI code of Jupyterhub hence not submitting the solution.

rohit2618:
@ryogesh I am also facing this issue often. Also, "Start my server" button does not work. But the behaviour is not always the same. Sometimes, the button works as expected.

rohit2618:
On debugging, I found that the "Stop my server" button does not work until home page is completely loaded. Once the home page is loaded, the stop button works as expected.

Also, I updated my templates and data files to latest and now I am getting "Start My Server" button. But still I get "Server not running" error.

manics:
Is this still an issue?

rabsr:
@manics It is not an issue anymore.

PR #3042, opened 2020-05-08
Changing start my server button link to spawn url once server is stopped

rabsr:
When single user server is stopped from hub/home, the text for the button "My server" is changed to "Start My server" but it does not change the url link to the button. It remains referenced to "/user/<username". So, on clicking "Start My server", it tries to redirect to the single user server. 
As, server is already stopped, so it redirects user to server not running page. Changing the url to spawn up the single user sever.
#2925


rabsr:
@minrk Can you please review PR? Also, I am not able to figure out reason for build failure. Please guide me for its fix.

minrk:
Thanks for the PR! I believe this should work AOK.

PR #3046, opened 2020-05-14
docs: add proxy_http_version 1.1

ceocoder:
add proxy_http_version 1.1 as it is required for KeepAlive connections


minrk:
Nice, thanks!

PR #3053, opened 2020-05-25
Add documentation for c.JupyterHub.extra_handlers

jtpio:
This change adds a new `Adding extra handlers to the hub` page to the documentation, to show how to use `c.JupyterHub.extra_handlers` to define custom handlers and have them displayed in the navigation bar:

image

The main motivation is to make this more visible, as for now `c.JupyterHub.extra_handlers` seems to be mentioned in the api docs and changelog only.


jtpio:
Additionally there could also be an example to show how to extend from `jupyterhub.apihandlers.APIHandler`.

docs/source/reference/extra-handlers.md
```diff
@@ -0,0 +1,54 @@
...
+c.JupyterHub.extra_handlers.extend([
+    (r"custom", CustomHandler),
+])
+```
+
+Issuing a GET request to `/hub/custom` will now render the default JupyterHub page with an empty content.
+
+## Adding a new link to the default navigation bar
+
+The new handler can be added to the navigation bar by creating a new `page.html` file template with the following content:
```
minrk, line 32: `page.html` is perhaps a confusing name, since there's an existing `page.html` in the base templates. Maybe use `custompage.html`?

jtpio: Shouldn't it be called `page.html` to override the default navbar and include the extra item?

GeorgianaElena: To override the default navbar of the home page, we need to override `page.html` indeed :+1: 

What was a bit confusing to me was that the CustomHandler is rendering `page.html`. I believe it would be nice to have CustomHandler render a custom page instead (maybe named `custompage.html`) that extends `page.html`.
That way, in addition to the instructions on how to add an extra handler and how to override a default template to include that handler in the home page, there will also be info on how to extend a base template.

What do you think?

docs/source/reference/extra-handlers.md
```diff
@@ -0,0 +1,54 @@
...
+{{ super() }}
+<li><a href="{{base_url}}custom">Custom Handler</a></li>
+{% endblock %}
+```
+
+The `jupyterhub_config.py` file can then be updated to include the custom template:
+
+```python
+extra_template_path = "/path/to/extra/templates"
+c.JupyterHub.template_paths.insert(0, extra_template_path)
```
GeorgianaElena, line 47:
```suggestion
c.JupyterHub.template_paths = [extra_template_path]
```

JupyterHub will fall back on the default templates if it doesn't find a custom template that overwrites the defaults in the `JupyterHub.template_paths` list ([docs ref](https://jupyterhub.readthedocs.io/en/stable/reference/templates.html?highlight=default%20template#custom-templates)). But overwriting the `template_paths` entirely might look more confusing than just inserting in the list. @jtpio, what do you think?

yuvipanda:
@jtpio thanks for opening this! Do you think you'll have time anytime soon to address the comments here?

docs/source/reference/extra-handlers.md
consideRatio: @GeorgianaElena I pick up mixed messages between this comment and the suggestion. It seems to me like you suggest A instead of B, but the code suggestion is to make B instead of A.

I know I would be worried if I had a replace operation that I would need to provide all default templates along those I've replaced. In a situation of mine, I have updated a single template, spawn.html, but nothing else, so I just want to add a paths location for that.

GeorgianaElena: > I know I would be worried if I had a replace operation that I would need to provide all default templates along those I've replaced. In a situation of mine, I have updated a single template, spawn.html, but nothing else, so I just want to add a paths location for that.

When you do `c.JupyterHub.template_paths = [extra_template_path]`, (from my understanding), it's not a replacement of default templates: https://github.com/jupyterhub/jupyterhub/blob/a4390a1f4fd0da9f8844339920428f4c0b3649f5/jupyterhub/app.py#L2218-L2219

> @GeorgianaElena I pick up mixed messages between this comment and the suggestion. It seems to me like you suggest A instead of B, but the code suggestion is to make B instead of A.

I was just trying to let @jtpio know that there's no need to accept my code change suggestion if they consider it to be confusing. Sorry if I ended up posting a confusing sentence instead :confused:

consideRatio: @GeorgianaElena ah so the config option is really be more like "extra_template_paths"?

GeorgianaElena: yeah, this is how I understand this.

consideRatio: :+1: !!

manics:
I've marked this as a draft for now

PR #3056, opened 2020-05-26
Use the issue templates from the central repo

GeorgianaElena:
Now that we have a central repo (https://github.com/jupyterhub/.github) where we can maintain the templates for the repos within the org, I believe we can remove them from here  :sun_with_face:


manics:
👍 
image

PR #3057, opened 2020-05-27
Warn if both bind_url and ip/port/base_url are set

GeorgianaElena:
This PR adds a warning if a user sets both `bind_url` and `ip`/`port`/`base_url` (which are deprecated as of JupyterHub 0.9 and replaced by `bind_url` according to the [config definitions](https://github.com/jupyterhub/jupyterhub/blob/81d423d6c674765400a6fe88064c1366b7070f94/jupyterhub/app.py#L516)).

This is a follow up on Min's comment [here](https://github.com/jupyterhub/jupyterhub/pull/2773#issuecomment-543119346) and Erik's [question](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/1680) about `JupyterHub.ip` and `.port`.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -568,7 +568,16 @@ class JupyterHub(Application):
         urlinfo = urlinfo._replace(netloc=fmt % (self.ip, self.port))
         urlinfo = urlinfo._replace(path=self.base_url)
         bind_url = urlunparse(urlinfo)
+
+        # Warn if both bind_url and ip/port/base_url are set
         if bind_url != self.bind_url:
+            if self.bind_url != "http://:8000" and self.bind_url != "https://:8000":
+                self.log.warning(
+                    "Both bind_url and ip/port/base_url have been configured. "
+                    "JupyterHub.ip, JupyterHub.port, JupyterHub.base_url are"
+                    " deprecated in JupyterHub 0.9,"
+                    " please use JupyterHub.bind_url instead."
+                )
             self.bind_url = bind_url
 
     bind_url = Unicode(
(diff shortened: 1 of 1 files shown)


consideRatio:
For this configuration, these hints are worth a lot! As far as I can tell, this LGTM, but I'm not confident in general about this which is why I appreciate the hints so much :) Thanks for working on this @GeorgianaElena!

jupyterhub/app.py
```diff
@@ -568,7 +568,16 @@ def _url_part_changed(self, change):
         urlinfo = urlinfo._replace(netloc=fmt % (self.ip, self.port))
         urlinfo = urlinfo._replace(path=self.base_url)
         bind_url = urlunparse(urlinfo)
+
+        # Warn if both bind_url and ip/port/base_url are set
         if bind_url != self.bind_url:
+            if self.bind_url != "http://:8000" and self.bind_url != "https://:8000":
```
consideRatio, line 574: I'm not sure why these values are picked out specifically to check against, are they default values?

GeorgianaElena: Yes, they are set [here](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/app.py#L594). Unfortunately, I didn't find a nicer way to check against the default values without hard-coding them :confused:

consideRatio: Ah! Will this work perhaps?

```suggestion
            if self.bind_url != self._bind_url_default():
```

GeorgianaElena: Yep, it works (I have no idea why I thought it wouldn't :laughing: )
Thanks @consideRatio :heart:

minrk:
Excellent, @GeorgianaElena! And thanks @consideRatio for the thoughtful review.

Issue #2503, opened 2019-04-04
Docker image should come with authentication


kevart:
It would be nice to include a default user+pw for the docker image (instructions here: https://hub.docker.com/r/jupyterhub/jupyterhub#configuration). Docker is an easy and quick way to test a software. Without credentials I can't log in, can't see anything, can't test and might just try something else.

minrk:
Now that dummy auth and simple spawner are available, we should probably use those by default in the image.

Issue #3007, opened 2020-03-31
Docker: "Spawn failed: Server didn't respond in 30 seconds", can be fixed by pip install notebook in container


JulianAssmann:
### Bug description
When using jupyterhub with docker, the error "Spawn failed: Server didn't respond in 30 seconds" occurs when logging in as a user. This can be fixed by running 
`pip install notebook`
inside the docker container.

#### Expected behaviour
Spawning the user server without problems, as is the case after pip installing notebook in the docker container

### How to reproduce
1. `docker run -d -p 8000:8000 jupyterhub/jupyterhub`
2. `docker exec -it [container_id] /bin/sh`
3. (in docker container) `adduser test`
4. Open `http://localhost:8000/` in browser of host machine
5. Login as the newly created user
6. Wait and see the error

### Your personal set up

 - OS:
Ubuntu 19.04, Docker Version 19.03.8 (build afacb8b7f0)
 - Version:
Docker image [jupyterhub/jupyterhub:1.2](https://hub.docker.com/layers/jupyterhub/jupyterhub/1.2/images/sha256-caeee1788998dcf8ab25ba5cd97313941075571c86445a843ec8bc9cc09f2b32?context=explore)
 - Configuration:
See steps above

### Logs
These logs were aquired by running `docker logs -f [container_id]`:
```
[I 2020-03-31 10:47:31.372 JupyterHub app:2291] Running JupyterHub version 1.2.0dev
[I 2020-03-31 10:47:31.374 JupyterHub app:2322] Using Authenticator: jupyterhub.auth.PAMAuthenticator-1.2.0dev
[I 2020-03-31 10:47:31.374 JupyterHub app:2322] Using Spawner: jupyterhub.spawner.LocalProcessSpawner-1.2.0dev
[I 2020-03-31 10:47:31.374 JupyterHub app:2322] Using Proxy: jupyterhub.proxy.ConfigurableHTTPProxy-1.2.0dev
[I 2020-03-31 10:47:31.388 JupyterHub app:1429] Writing cookie_secret to /srv/jupyterhub/jupyterhub_cookie_secret
[I 2020-03-31 10:47:31.422 alembic.runtime.migration migration:155] Context impl SQLiteImpl.
[I 2020-03-31 10:47:31.423 alembic.runtime.migration migration:162] Will assume non-transactional DDL.
[...]

PR #3059, opened 2020-05-28
Add JupyterHub Demo docker image

GeorgianaElena:
This adds a docker image example that includes the notebook package and sets up JupyterHub to use the dummy authenticator and the simple spawner. This can be used to quickly test out how JupyterHub works.
 
Closes https://github.com/jupyterhub/jupyterhub/issues/2503
Relevant reference: https://github.com/jupyterhub/jupyterhub/issues/3007#issuecomment-633984779


minrk:
Great! Next, we'll need to publish this image on DockerHub. I'll look at the hooks for that.

PR #3071, opened 2020-06-04
More consistent behavior for UserDict.get and `key in UserDict`

minrk:
- `in` supports all key types, checking for presence in cache (O(N) for str since we don't maintain a separate name map)
- `get` behaves the same as attempting to fetch (meaning it supports all key types); the inherited base dict implementation only checked presence in the Dict, not the db, and only worked with integers

### DIFF
--- a/jupyterhub/user.py
+++ b/jupyterhub/user.py
@@ -34,7 +34,23 @@ from .utils import url_path_join
 class UserDict(dict):
     """Like defaultdict, but for users
 
-    Getting by a user id OR an orm.User instance returns a User wrapper around the orm user.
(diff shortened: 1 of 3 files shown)


jupyterhub/user.py
```diff
@@ -34,7 +34,22 @@
...
+    - orm.User object
+    - username str
+
+    A User wrapper object is always returned.
+
+    This dict contains at least all active users,
+    but not necessarily all users in the database.
+
+    Checking `key in userdict` returns whether
+    an item
```
betatim, line 49: sentence ends mid sentence

PR #3076, opened 2020-06-05
Test with some master dependencies.

Carreau:
This does some of the test with the latest traitlets.
We are looking into making a 5.0 release and would like to have some
confidence that it does not break too many things.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,10 @@ before_install:
 install:
   - pip install --upgrade pip
   - pip install --upgrade --pre -r dev-requirements.txt .
+  - |
+    if [[ "$MASTER_DEPENDENCIES" == "True" ]]; then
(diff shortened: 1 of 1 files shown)


betatim:
Merged. Three questions:
* should we mark these tests as "allowed to fail"?
* should we keep this and use it for more than just traitlets, similar to what we do for python nightly?
* I'm not wild about "master dependencies". From just the PR title I thought maybe this was about testing with a subset of JH's dependency (the "important ones") or something like that. How about "Testing dependencies at HEAD"?

consideRatio:
I assumed this PR was intended to run tests with traitlets master through this PR and it worked.

If it is to be part of the actual test suite, i think it absolutely should be allowed to fail.

If we want long term runs of this as an indicator, a travis cronjob makes most sense, where notification on failure is configure to email the developers interested in the travis jobs outcome.

PR #3077, opened 2020-06-08
Add Configuration Reference section to docs

### LINKED ISSUE #3066: Include Default Config Files and Documented CLI Options in docs
### Proposed change
Many people deploy this into containerized environments, and as such generating a config file to modify or running `jupyterhub --help-all` to get the options when starting the server is a time-consuming and non-trivial task depending on your environment. It would be great if the repo (or some referenced location) could host a default `jupyterhub_config.py` users could modify without having to create an environment in which to install, generate, and extract the file. Similarly, it'd be great if the docs for configuration would just list the options for starting the process rather than saying "run --help".

### Alternative options
As mentioned above, in interactive environments this is easy to deal with and a non-issue, but when writing a dockerfile, working on a thin client before deploying to a cluster that costs you money to access, etc. these are inconvenient, time-consuming, and potentially have a cost associated. Since these aren't exactly tall asks, I think this is pretty reasonable.

Anyone creating a configuration for their deployment as code who isn't setting it up interactively, which I have to imagine is most people.

kinow:
Closes #3066 

Tried a naive approach to capture the output of `jupyterhub --generate-config` and `jupyterhub --help-all` commands, and display in the generated documentation.

It worked, but as I am still learning Sphinx and how to use/extend it, there may be other better alternatives? For example, I got similar behaviour with the extensions `sphinxcontrib-programoutput` and with `sphinx-autorun`.

At $work, there is a project using Sphinx extensions, and we found that some times extensions are not compatible with the latest version of sphinx, which may hold back the project from updating sphinx. That's why I decided on a directive :+1: 

Happy to move to another section, or update as necessary. One thing bothering me is trying to find the properties in the middle of the comments. Although I normally search the output for things like "ssl", or "proxy", etc.

Cheers
Bruno

### DIFF
--- a/docs/source/reference/config-reference.rst
+++ b/docs/source/reference/config-reference.rst
@@ -1,0 +1,30 @@
+==============================
+Configuration Reference
+==============================
+
+.. important::
+
+    Make sure the version of JupyterHub for this documentation matches your
+    installation version, as the output of this command may change between versions.
+
+JupyterHub configuration
+------------------------
+
+As explained in the `Configuration Basics <../getting-started/config-basics.html#generate-a-default-config-file>`_
+section, the ``jupyterhub-config.py`` can be automatically generated via
+
+    .. code-block:: bash
+
+        jupyterhub --generate-config
+
+
+The following contains the output of that command for reference.
+
+.. jupyterhub-generate-config::
+
+JupyterHub help command output
+------------------------------
+
+This section contains the output of the command ``jupyterhub --help-all``.
+
+.. jupyterhub-help-all::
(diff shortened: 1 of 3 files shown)


kinow:
Rendered page for review: https://kinow.github.io/jupyterhub/reference/config-reference.html

docs/source/reference/config-reference.rst
```diff
@@ -0,0 +1,30 @@
...
+.. important::
+
+    Make sure the version of JupyterHub for this documentation matches your
+    installation version, as the output of this command may change between versions.
+
+JupyterHub configuration
+------------------------
+
+As explained in the `Configuration Basics <../getting-started/config-basics.html#generate-a-default-config-file>`_
+section, the ``jupyterhub-config.py`` can be automatically generated via
```
minrk, line 14:
```suggestion
section, the ``jupyterhub_config.py`` can be automatically generated via
```

minrk:
This is really nice, thanks!

PR #3078, opened 2020-06-10
proxy settings might cause authentication errors

gatoniel:
This documentation lacks a fool proof fix, since unsetting the enironment variables might cause errors in other parts of the server setup, e.g. notebook extensions can´t connect to the internet.


minrk:
Great, thanks!

PR #3081, opened 2020-06-11
make Spawner.environment config highest priority

minrk:
so that it can override 'default' env variables like JUPYTERHUB_API_URL

use with caution, since this allows setting non-working values for required URLs, but is still a useful escape hatch when e.g. hub_connect_url might want to have different values across spawners

cf https://github.com/jupyterhub/dockerspawner/pull/361


yuvipanda:
tyvm, @minrk :)

PR #3086, opened 2020-06-13
Replace zonca/remotespawner with NERSC/sshspawner

manics:
https://github.com/zonca/remotespawner is archived, the repository readme recommends https://github.com/jupyterhub/batchspawner for the original use case.


consideRatio:
@manics ❤️

PR #3087, opened 2020-06-13
Make `delete_invalid_users` configurable

fcollonval:
The following warning may happen:

https://github.com/jupyterhub/jupyterhub/blob/b7f277147bad0855104e4b725254cefba784a9ed/jupyterhub/app.py#L1738-L1743

But the corresponding setting is not configurable.


minrk:
Great, thank you!

PR #3089, opened 2020-06-15
Include the query string parameters when redirecting to a new URL

### LINKED ISSUE #3061: query parameters dropped when redirecting from hub to login page
### Bug description
query parameters are dropped when redirecting from hub to login page.   Documentation which I list below says query parameters should be preserved.

From hub log you can see the parameters are dropped on redirect:
[I 2020-05-29 05:55:28.094 JupyterHub log:174] 302 GET /hub/?frog=water -> /hub/login (@10.0.1.5) 1.01ms
[I 2020-05-29 05:55:28.147 JupyterHub log:174] 302 GET /hub/login -> /hub/oauth_login (@10.0.1.5) 0.91ms

Documentation states:
https://jupyterhub.readthedocs.io/en/stable/reference/urls.html
All authenticated handlers redirect to /hub/login to login users prior to being redirected back to the originating page. The returned request should preserve all query parameters.

#### Expected behaviour
query parameters would be preserved on redirects

#### Actual behaviour
query parameters are dropped

### How to reproduce

1. Go to /hub?x=y
2. It will redirect to /login and your url parameters will be gone

### Your personal set up

 - OS:
linux
 - Version:
JupyterHub version 1.1.0
 - Configuration:

kinow:
Closes #3061

Tentative for #3061, will re-read the issue and my own changes, and then prepare tests if it's looking good.

I changed recently some code in another PR that was handling parameters in URL redirect, but with JavaScript. When I read the issue #3061, I thought I could have caused a regression, and decided to investigate it.

#3061 is not related to my change. But I thought the bug reported made sense, and keeping the parameters when redirecting could be useful (e.g. when passing some token/sessionId/flag/etc down to a spawner).

Cheers
Bruno

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -678,7 +678,7 @@ class BaseHandler(RequestHandler):
         a list with "next" (to avoid redirect-loops)
         :rtype (str)
         """
-        if not exclude:
+        if exclude is None:
             exclude = ['next']
         if self.request.query:
             query_string = [
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -654,7 +654,35 @@ class BaseHandler(RequestHandler):
                     next_url = url_path_join(self.hub.base_url, 'spawn')
             else:
                 next_url = url_path_join(self.hub.base_url, 'home')
+
+        next_url = self.append_query_parameters(next_url, exclude=['next'])
         return next_url
+
+    def append_query_parameters(self, url, exclude=None):
+        """Append the current request's query parameters to the given URL.
+
+        Supports an extra optional parameter ``exclude`` that when provided must
+        contain a list of parameters to be ignored, i.e. these parameters will
+        not be added to the URL.
+
+        This is important to avoid infinite loops with the next parameter being
+        added over and over, for instance.
+
+        :param str url: a URL
+        :param list exclude: optional list of parameters to be ignored
+        :rtype (str)
+        """
+        if not exclude:
+            exclude = []
+        if self.request.query:
+            query_string = [
+                param
+                for param in parse_qsl(self.request.query)
+                if param[0] not in exclude
+            ]
+            if query_string:
+                url = url_concat(url, query_string)
+        return url
 
     async def auth_to_user(self, authenticated, user=None):
         """Persist data from .authenticate() or .refresh_user() to the User database
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -668,12 +668,18 @@ class BaseHandler(RequestHandler):
         This is important to avoid infinite loops with the next parameter being
         added over and over, for instance.
 
+        The default value for ``exclude`` is an array with "next". This is useful
+        as most use cases in JupyterHub (all?) won't want to include the next
+        parameter twice (the next parameter is added elsewhere to the query
+        parameters).
+
         :param str url: a URL
-        :param list exclude: optional list of parameters to be ignored
+        :param list exclude: optional list of parameters to be ignored, defaults to
+        a list with "next" (to avoid redirect-loops)
         :rtype (str)
         """
         if not exclude:
-            exclude = []
+            exclude = ['next']
         if self.request.query:
             query_string = [
                 param
--- a/jupyterhub/handlers/login.py
+++ b/jupyterhub/handlers/login.py
@@ -90,20 +90,13 @@ class LoginHandler(BaseHandler):
     """Render the login page."""
 
     def _render(self, login_error=None, username=None):
-        # login url with any parameters specified by the user/application
-        login_url = self.append_query_parameters(
-            self.settings['login_url'], exclude=['next']
-        )
-        # the next url for after the login
-        next_url = url_escape(self.get_argument('next', default=''))
-        # now concat the login_url with the next_url
-        login_url = url_concat(login_url, {'next': next_url})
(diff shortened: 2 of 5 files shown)


kinow:
Tests passing. Just need to write some extra tests, and then ready for review.

jupyterhub/handlers/base.py
```diff
@@ -654,8 +654,36 @@ def get_next_url(self, user=None, default=None):
                     next_url = url_path_join(self.hub.base_url, 'spawn')
             else:
                 next_url = url_path_join(self.hub.base_url, 'home')
+
+        next_url = self.append_query_parameters(next_url, exclude=['next'])
         return next_url
 
+    def append_query_parameters(self, url, exclude=None):
```
minrk, line 661: Maybe the default should be to exclude `next` since it's called that way 100% of the time so far?

kinow: Hmmm, good point. That would prevent other developers using it and accidentally including `next`, which could trigger a redirect-loop. Let me update it.

minrk:
Nice! Makes sense. Left one comment in-line about defaults matching common (universal?) usage, but looks great to me.

kinow:
> Nice! Makes sense. Left one comment in-line about defaults matching common (universal?) usage, but looks great to me.

Thanks for the review @minrk . I've added a [new commit](https://github.com/jupyterhub/jupyterhub/pull/3089/commits/83eaa028dec74eaa15c03f1bab9c90a201ab864b) with a tentative pydocs text and default `exclude=["none"]` (happy to squash it later if necessary).

I did that within the function, instead of in the signature to avoid the mutable parameter, but happy to update it if there's another way that works better for this.

kinow:
One unit test failure on 3.5. Tried to reproduce but it passes for me (3.7-anaconda). It looks
like I wrote a fragile unit test. So instead of checking the exact string, will test simply that it
contains both parameters in the string, as from the log output in CI it appears the order
the parameters are added is not guaranteed.

```py
assert r.url.endswith('/hub/login?next=foo&param1=test')

assert 'next=foo' in r.url
assert 'param1=test' in r.url
```

Edited last commit with the changes above. Let's see if CI builds pass now.

jupyterhub/handlers/login.py
```diff
@@ -90,13 +90,20 @@ class LoginHandler(BaseHandler):
     """Render the login page."""
 
     def _render(self, login_error=None, username=None):
+        # login url with any parameters specified by the user/application
+        login_url = self.append_query_parameters(
+            self.settings['login_url'], exclude=['next']
+        )
+        # the next url for after the login
+        next_url = url_escape(self.get_argument('next', default=''))
```
minrk, line 98: I don't think it's quite right that the URL parameters from the login page are preserved on the destination. Instead, they should be *encoded* in the `next` URL. Adding them here might be doubly-encoding params.

jupyterhub/handlers/base.py
```diff
@@ -654,8 +654,42 @@ def get_next_url(self, user=None, default=None):
                     next_url = url_path_join(self.hub.base_url, 'spawn')
             else:
                 next_url = url_path_join(self.hub.base_url, 'home')
+
+        next_url = self.append_query_parameters(next_url, exclude=['next'])
```
minrk, line 658: I think this addition of params is the one that's definitely correct, adding params where `next` is already defined is probably not correct.

jupyterhub/handlers/login.py
kinow: Ah! I think that makes sense! Let me try that.

kinow: Had a try at adding the tests and adding the parameters escaped/encoded to the `next` parameter @minrk . Let me know if that looks better :+1:

jupyterhub/tests/test_pages.py
```diff
@@ -516,6 +517,58 @@ async def test_user_redirect_deprecated(app, username):
...
+        (
+            # login
+            # simplest case, accessing the login URL, gives an empty next url
+            'login',
+            [],
+            '/hub/login',
+            '/hub/login?next=',
+        ),
+    ],
+)
```
kinow, line 557: Realized today that this test is quite similar to `test_login_redirect`, with the difference that it follows redirects, matching with the final URL, and also checks the `<form action="">` value. I wonder whether that's OK, or if it needs to be merged/updated to fit into `test_login_redirect` or renamed, etc?

minrk: I think it's fine to add this to focus on slightly different things.

jupyterhub/handlers/base.py
```diff
@@ -654,8 +654,42 @@ def get_next_url(self, user=None, default=None):
...
+        as most use cases in JupyterHub (all?) won't want to include the next
+        parameter twice (the next parameter is added elsewhere to the query
+        parameters).
+
+        :param str url: a URL
+        :param list exclude: optional list of parameters to be ignored, defaults to
+        a list with "next" (to avoid redirect-loops)
+        :rtype (str)
+        """
+        if not exclude:
```
minrk, line 681: Checking `if not exclude` prevents callers from specifying `exclude=[]`. Explicit None check will more accurately capture the default:

```suggestion
        if exclude is None:
```

kinow: Oops! Fixing it!

minrk:
Thanks!

kinow:
Thank you for the review @minrk !

PR #3090, opened 2020-06-15
Rename Authenticator.white/blacklist to allowed/blocked

minrk:
cf https://github.com/jupyterhub/team-compass/issues/299

todo:

- [x] Authenticator.whitelist -> Authenticator.allowed_users
- [x] Authenticator.blacklist -> Authenticator.blocked_users
- [x] group_whitelist -> allowed_groups
- [x] Authenticator.check_whitelist -> Authenticator.check_allowed (this is the top-level method that checks *all* sources of allow—users, groups, etc.)
- [x] LocalAuthenticator.check_group_whitelist -> LocalAuthenticator.check_allowed_groups
- [x] deprecate traits, config
- [x] test deprecated names still work
- [x] test various combinations of deprecated config for proper priority resolution
- [x] subclass resolution in deprecated method names (subclass overriding check_whitelist should take priority over base class check_allowed, but warn about deprecation *unless* check_allowed is also defined on the subclass)
- [x] update docs

### DIFF
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -297,7 +297,7 @@ class Authenticator(LoggingConfigurable):
         self._init_deprecated_methods()
 
     def _init_deprecated_methods(self):
-        # TODO: properly handle deprecated signature *and* name
+        # handles deprecated signature *and* name
         # with correct subclass override priority!
         for old_name, new_name in (
             ('check_whitelist', 'check_allowed'),
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -101,9 +101,9 @@ class Authenticator(LoggingConfigurable):
         """
     ).tag(config=True)
 
-    whitelist = Set(help="Deprecated, use `Authenticator.allowed`", config=True,)
+    whitelist = Set(help="Deprecated, use `Authenticator.allowed_users`", config=True,)
 
-    allowed = Set(
+    allowed_users = Set(
         help="""
         Set of usernames that are allowed to log in.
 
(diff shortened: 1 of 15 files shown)


minrk:
I believe this is ready to go now

jupyterhub/auth.py
```diff
@@ -100,41 +101,74 @@ class Authenticator(LoggingConfigurable):
         """
     ).tag(config=True)
 
-    whitelist = Set(
+    whitelist = Set(help="Deprecated, use `Authenticator.allowed`", config=True,)
+
+    allowed = Set(
```
manics, line 106: Some authenticators have additional properties such as `allowed_groups`. Any thoughts on changing this to `allowed_users` for clarity?

minrk:
What do people think abut the `.check_allowed` method name? This is the main method for checking if a user is allowed by any mechanism and is overridden by subclasses to also call `.check_allowed_groups()` etc. (see the LocalAuthenticator here that does just that), so I feel it's appropriate to be generic. I thought about `.check_allowed_users` to match the config name, but found that confusing that it should be overridden to check other sources of information, not just the `.allowed_users` set.

jupyterhub/auth.py
minrk: Good idea. Switched to `allowed_users` pending resolution of [the discussion on team-compass](https://github.com/jupyterhub/team-compass/issues/299#issuecomment-648759244)

minrk:
Ready for review again

jupyterhub/auth.py
```diff
@@ -260,39 +294,74 @@ def my_hook(authenticator, handler, authentication):
...
         super().__init__(**kwargs)
-        for method_name in (
-            'check_whitelist',
-            'check_blacklist',
-            'check_group_whitelist',
+        self._init_deprecated_methods()
+
+    def _init_deprecated_methods(self):
+        # TODO: properly handle deprecated signature *and* name
+        # with correct subclass override priority!
```
GeorgianaElena, line 301:
```suggestion
        # handles deprecated signature *and* name
        # with correct subclass override priority!
```
not a todo anymore :tada:

minrk: Thanks!

PR #3100, opened 2020-06-22
Remove old context-less print statement

mriedem:
This was added in PR #2721 and by default results in just printing
out "10" without any context when starting the hub service. This
simply removes the orphan print statement.

I'm open to changing this to a debug log statement with context if
someone finds that useful, e.g.:

`self.log.debug('Effective init_spawners_timeout: %s', init_spawners_timeout)`


mriedem:
@minrk FYI, not sure if you wanted to keep this around and make it an actual log statement or just remove it.

minrk:
Nope, just a debug statement leftover. Thanks!

PR #3103, opened 2020-06-24
Hide hamburger button menu in mobile/responsive mode and fix other minor issues

### LINKED ISSUE #2182: hub/login - breadcrumb with no content
On mobile, while visiting the login page, the breadcrumb can be seen at the top right corner. When clicked, it shows no content.

breadcrumb

kinow:
Closes #2182 

I think for #2182 we can simply hide the hamburger menu when there's no `user` variable set? The menu that goes under that hamburger menu [appears to be filled only when there is a `user` set](https://github.com/jupyterhub/jupyterhub/blob/dbc6998375dd8251d196863425b26adc814624cd/share/jupyterhub/templates/page.html#L118), so that would match it I think.

So the login page would look like this on mobile:

image

---

Another issue that I noticed is that the JupyterHub icon doesn't have a left margin, as the hamburger button menu has on its right hand-side.

image

Added a commit that adds the same margin, using a Bootstrap variable, but only when in mobile mode. That's because the `container-fluid` class appears to be doing the trick when not in mobile, so simply adding it would add the margin twice.

image

GIFrecord_2020-06-24_155738

---

And clicking on the hamburger menu when logged in, displays the user name and the logout button without left padding/margin.

image

That's because Bootstrap adds it to `.nav > li > a`, but for the user name and button, we have both under `.nav > li > span`. Added the same values as Bootstrap for `.nav > li > a`, but again only in responsive mode, as in desktop mode it displays everything in horizontal/inline mode.

Final version with the changes in this PR:

GIFrecord_2020-06-24_171210

To review, it's possible to check out this branch/PR, then run `python setup.py css` to re-compile the less styles, then run `jupyterhub`.

Cheers
Bruno

### DIFF
--- a/share/jupyterhub/static/less/page.less
+++ b/share/jupyterhub/static/less/page.less
@@ -13,6 +13,20 @@
   }
 }
 
+.navbar-right {
+  li {
+    span {
+      // same as .nav > li > a from bootstrap, but applied to the span[id="login_widget"]
+      // or any other span that matches .nav > li > span, but only in responsive mode
+      @media (max-width: @grid-float-breakpoint) {
+        position: relative;
+        display: block;
+        padding: 10px 15px;
+      }
+    }
+  }
+}
+
 #header {
   border-bottom: 1px solid #e7e7e7;
 }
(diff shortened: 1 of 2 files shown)


kinow:
(Happy to split into multiple PR's if that's easier to review)

kinow:
Hmm, funny, the CI errors doesn't seem to be related to this change?

minrk:
awesome, thanks!

PR #3104, opened 2020-06-30
Update version in docs/rest-api.yaml

cmd-ntrf:
The version tag has not been updated since 0.9.0dev.


minrk:
Thanks, if we pick up bump2version in this repo, adding this file to the list should keep it in sync.

PR #3107, opened 2020-07-05
docs: unsqueeze logo, remove unused CSS and templates

consideRatio:
This is another PR like https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/1710, lets take the discussion there and get back to this PR after that.

Hmmm it seems this caused some changes in this repo while I could not note any in z2jh.

**Changes observed**:
- A broken "Edit this page" button to the right disappeared. 
- Font sizes changed etc
- Width of the page went from 100% to less

This page is using alabaster_jupyterhub, I'm not understanding the documentation setup fully but I think some of the changes I did assumed alabaster_jupyterhub was no longer used as in z2jh.


choldgraf:
this LGTM - I can't see the broken "edit this page" in the docs previews, could you link to a page w/ that problem?

My guess is that the style changes could be because the docs haven't been built in a while and the theme has updated since then.

Also, I wanted to activate the RTD preview on this repo as well, but I'm not an admin for the RTD page with jupyterhub (I think?)...are you?

consideRatio:
@choldgraf ah I compared with the /en/stable version on readthedocs, while latest didn't have any such button, which means that this change does not remove it after all.

Then I think this is possible to merge as it is.

choldgraf:
wohoo! although travis is unhappy now? I can't re-start it because apparently I don't have admin privileges there either. @consideRatio can you try restarting? If all tests pass I am +1 on merge!

consideRatio:
I'm not sure about the build failure, I don't understand it =/

choldgraf:
OK sounds good - looking at the diff I am almost positive that the failure is not related to this patch. And since it is fixing a nasty docs bug I'll give this a merge! However I want to ping @betatim before I do so, so that he knows to shame me the next time we interact with each other 😅

PR #3111, opened 2020-07-10
Log slow_stop_timeout when hit like slow_spawn_timeout

mriedem:
When `slow_spawn_timeout` is hit the configured timeout value
gets logged [1]. This does the same thing when `slow_stop_timeout`
is hit.

[1] https://github.com/jupyterhub/jupyterhub/blob/1.1.0/jupyterhub/handlers/base.py#L947

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -1120,7 +1120,10 @@ class BaseHandler(RequestHandler):
         except gen.TimeoutError:
             # hit timeout, but stop is still pending
             self.log.warning(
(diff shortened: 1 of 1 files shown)


consideRatio:
@mriedem this LGTM!

Like you said, this PR is mirroring the logging of slow_spawn_timeout!

https://github.com/jupyterhub/jupyterhub/blob/3d524f2092c8f1ca4c2953d78156b73107ef8c1c/jupyterhub/handlers/base.py#L972-L987

Issue #2559, opened 2019-05-13
build docker images from a different repo


minrk:
We currently build the `jupyterhub/jupyterhub` and `jupyterhub/singleuser` images from this repo. That's a little funky, because it means building stable versions (1.0.0) and master from the repo tracking master. It makes it hard to separate the code for building images from the source of the application.

I think we should make a dedicated repo for building the jupyterhub docker images. This would make it easier to have variants, e.g.

- default (what we have now)
- with notebook (for fully operational testing/demo purposes, though this is ~never used in production)
- light alpine build (no proxy, no nodejs, proxy must be in another container, as is ~always done in production)

PR #3118, opened 2020-07-20
only build tagged versions on docker tags

minrk:
instead of building 'stable' from master

manually tracking what 'stable' is here is resulting in out-of-date docker images. Instead, build pinned versions on tags like we do for the jupyterhub images.

Related to https://github.com/jupyterhub/dockerspawner/pull/379 where the 1.1 image was never uploaded

### DIFF
--- a/singleuser/hooks/build
+++ b/singleuser/hooks/build
@@ -1,11 +1,4 @@
 #!/bin/bash
 set -ex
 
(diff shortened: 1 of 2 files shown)


minrk:
In the long run, I think we should really be building our docker images from a separate repo and not this one: #2559

Issue #3120, opened 2020-07-21
Spawner state cleared before post_stop_hook


rkdarst:
### Bug description
I have a complex `post_stop_hook` that is preserved via the spawner state.  Currently, spawner state is cleared, and then `post_stop_hook` gets invoked.  This was a bit of a surprise to me, and required me to do some debugging to figure out why this is.  Luckily the relevant code is easy to understand:

https://github.com/jupyterhub/jupyterhub/blob/f3c3225124f1c5d9acb2503ec4d9c35a140f6a78/jupyterhub/user.py#L798-L829

#### Expected behaviour
I'd like to have that spawner state available at post_stop_hook time.  I guess for now I can hack around it in the hooks, but given how spawner objects are reused (I think that was the outcome of the last discussion...), not being able to rely on state or clean attributes make that harder.

### Your personal set up
Latest master on k8s.

### Questions
I know enough to be cautious when messing with state...
* I can think of one solution is to set `spawner.orm_spawner.state = {}` and move the clear state to after post_stop_hook.  But, post_stop_hook isn't in the try/except block, so that changes the behavior when `.clear_state()` raises an exception.
  * Currently this is `spawner.clear_state(); spawner.orm_spawner.state = spawner.get_state()` - but I wouldn't be surprised if someone relies on existing behavior!  
  * This also assumes that `.clear_state()` sets state to `{}`!  I don't know what kinds of things that would mess up...

I'd appreciate some expert advice before I go digging too deep into this... whatever my solution may be!

cc @murhum1 @Gehock (our post_stop_hook can't get course_slug, so doesn't delete our tokens...)

PR #3121, opened 2020-07-22
jupyterhub/user: clear spawner state after post_stop_hook

rkdarst:
I'm submitting this more for discussion than I think it's a good idea, because it's easier to have something concrete to talk about than discussing abstractly.

- Related issue: #3120.  Closes: #3120.

- I realized that spawner.clear_state() is called before
  spawner.post_stop_hook().  This caused was a bit surprising to me,
  and caused some issues.

- I tried the naive strategy of moving clear_state to later and
  setting the orm_state to `{}` at the point where it used to be
  clear.

- This tries to maintain the exception behavior of clear_state and
  post_stop_hook, but is exactly identical.

- To review:

  - Is this even a good idea!

  - Carefully consider the implications of this.  I am not at all sure
    about unintended side-effects or what intended semantics are.
    Should the statements be re-ordered some other way?

### DIFF
--- a/jupyterhub/user.py
+++ b/jupyterhub/user.py
@@ -795,8 +795,7 @@ class User:
             status = await spawner.poll()
             if status is None:
                 await spawner.stop()
-            spawner.clear_state()
-            spawner.orm_spawner.state = spawner.get_state()
+            spawner.orm_spawner.state = {}
             self.last_activity = spawner.orm_spawner.last_activity = datetime.utcnow()
             # remove server entry from db
             spawner.server = None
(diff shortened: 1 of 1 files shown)


minrk:
I think delaying the clear to a later point is fine, but the `= {}` should *not* be there, since it can result in lost data.

rkdarst:
Thanks for the review.  I'm glad it's not *too* bad.

I pushed a change for that, and we get some meaningful test failures, which I guess encode previous semantics more than something that must be the case:
```
E       AssertionError: assert 'pid' not in {'pid': 7394}
E        +  where {'pid': 7394} = <jupyterhub.orm.Spawner object at 0x7f43f39c6400>.state
```

I can't do more now; if anyone wants to look sooner by all means go ahead.  @minrk, if this is a sign that we shouldn't adjust the order, let me know.

minrk:
> if this is a sign that we shouldn't adjust the order, let me know.

I don't think so, it's just that the moved clear_state is incomplete because it doesn't persist the cleared state to the db. It needs to be:

```python
# tell the spawner to clear its state
spawner.clear_state()
# persist cleared state to db
spawner.orm_spawner.state = spawner.get_state()
self.db.commit()
```

minrk:
docs failure is fixed by #3151, not related to this PR.

Thanks, I'm happy to give this a try!

rkdarst:
Thanks!  I hope it works and doesn't end up causing problems down the line...

PR #3123, opened 2020-07-22
app.py: Work around incompatibility between Tornado 6 and asyncio proactor event loop in python 3.8 on Windows

alexweav:
In py38, asyncio made the `WindowsProactorEventLoopPolicy` the default setting for event loops on Windows. However, this is incompatible with Tornado. The result is that the Hub does not start on Windows on Python 3.8, with a `NotImplementedError`. See tornadoweb/tornado#2608 for additional context. 

The workaround suggested by Tornado is to use the older `WindowsSelectorEventLoopPolicy`, which is the default in py37 and before. This is done via:
```
asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
```

This workaround has already been applied in the single-user server: jupyter/notebook#5047. The content of this PR is almost identical to the single-user one, save for some formatting differences. We only alter the default setting if we're running in an environment where this issue might occur (Windows, py38, and a sufficiently advanced asyncio API).

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -2751,13 +2751,16 @@ class JupyterHub(Application):
             try:
                 from asyncio import (
                     WindowsProactorEventLoopPolicy,
-                    WindowsSelectorEventLoopPolicy
+                    WindowsSelectorEventLoopPolicy,
                 )
             except ImportError:
                 pass
                 # not affected
             else:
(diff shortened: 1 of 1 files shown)


alexweav:
Hello @minrk, sorry to bother, but is there a chance you or someone else might be able to review this PR? (I pinged you because the content of this PR was originally authored by you, but in the single-user server instead)

minrk:
Thanks for the PR! Just getting back from vacation today.

PR #3127, opened 2020-07-23
Don't log a warning when slow_spawn_timeout is disabled

### LINKED ISSUE #3126: slow_spawn_timeout warning when timeout is disabled (0)
### Bug description

We're using zero-to-jupyterhub-k8s which sets `slow_spawn_timeout` to 0 [1] and that results in a warning in the hub logs, e.g.:

>Jul 23 14:09:36 hub-55d978d7cb-75lwl hub WARNING WARNING 2020-07-23T19:09:36.897Z [JupyterHub base:950] User hub-stress-test-123 is slow to start (timeout=0) 

If the `slow_spawn_timeout` is disabled we should not log a warning for that situation since we're not waiting at all.

Related to #1222 

[1] https://github.com/jupyterhub/zero-to-jupyterhub-k8s/commit/b4738edc5dd5a3b7ecc117ca34c556afbf8981e6

#### Expected behaviour
We shouldn't log a warning for an expected situation where we're not waiting.

#### Actual behaviour
Don't log a warning when `slow_spawn_timeout` is 0 and we don't wait for the server to start.

### How to reproduce
Configure `slow_spawn_timeout: 0` in `c.JupyterHub.tornado_settings`, create a user and start their server then check the hub logs for the warning.

### Your personal set up

 - OS: linux
 - Version: commit 8a3790b01ff944c453ffcc0486149e2a58ffabea
 - Configuration: zero-to-jupyterhub-k8s default configuration for `slow_spawn_timeout` (0 in this case)

mriedem:
When using the `KubeSpawner` it is typical to disable the
`slow_spawn_timeout` by setting it to 0. `zero-to-jupyterhub-k8s`
does this by default [1]. However, this causes an immediate `TimeoutError`
which gets logged as a warning like this:

>User hub-stress-test-123 is slow to start (timeout=0)

This avoids the warning by checking the value and if disabled simply
returns without logging the warning.

[1] https://github.com/jupyterhub/zero-to-jupyterhub-k8s/commit/b4738edc5

Closes #3126

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -977,13 +977,16 @@ class BaseHandler(RequestHandler):
             # waiting_for_response indicates server process has started,
             # but is yet to become responsive.
             if spawner._spawn_pending and not spawner._waiting_for_response:
-                # still in Spawner.start, which is taking a long time
-                # we shouldn't poll while spawn is incomplete.
-                self.log.warning(
-                    "User %s is slow to start (timeout=%s)",
-                    user_server_name,
-                    self.slow_spawn_timeout,
-                )
+                # If slow_spawn_timeout is intentionally disabled then we
+                # don't need to log a warning, just return.
+                if self.slow_spawn_timeout > 0:
+                    # still in Spawner.start, which is taking a long time
+                    # we shouldn't poll while spawn is incomplete.
+                    self.log.warning(
+                        "User %s is slow to start (timeout=%s)",
+                        user_server_name,
+                        self.slow_spawn_timeout,
+                    )
                 return
 
             # start has finished, but the server hasn't come up


jupyterhub/handlers/base.py
```diff
@@ -977,13 +977,16 @@ def abort():
...
-                # still in Spawner.start, which is taking a long time
-                # we shouldn't poll while spawn is incomplete.
-                self.log.warning(
-                    "User %s is slow to start (timeout=%s)",
-                    user_server_name,
-                    self.slow_spawn_timeout,
-                )
+                # If slow_spawn_timeout is intentionally disabled then we
+                # don't need to log a warning, just return.
+                if self.slow_spawn_timeout > 0:
```
mriedem, line 982: Should we avoid even calling `with_timeout` if `slow_spawn_timeout` is <= 0? What do we do instead? Pass `finish_spawn_future` to `gen.maybe_future`? I'm new to the tornado library so I'm not sure what is best here.

minrk: I think it might be simpler to keep the control flow this way (too many branches and conditions can make it harder to ensure everything gets cleaned up in every case), but if you want to add the short-circuit before the wait, I think it would look like:

```python
if self.slow_spawn_timeout == 0 and not finish_spawn_future.done():
        # nothing to wait for if slow_spawn_timeout is disabled
        return

try:
    await ...
```

minrk:
Merging since this is a  straightforward improvement as-is (thanks!) We can look at restructuring a bit in another PR, if folks are interested in doing that.

PR #3128, opened 2020-07-24
Implement singleuser with mixins

minrk:
in order to be compatible with both nbclassic server and jupyter_server for jupyterlab 2.0+:

- [x] make `jupyterhub.singleuser` a package to break it up a little
- [x] adds `jupyterhub.singleuser.mixins` which define the mixins required
- [x] adds `jupyterhub.singleuser.make_singleuser_app` function to apply mixins to a base class
- [x] define default app in `jupyterhub.singleuser.notebookapp`
- <del>[ ] define default jupyter_server app (not sure if this makes sense?) </del> I don't think this makes sense. Setting JUPYTERHUB_SINGLEUSER_APP can work for any jupyter_server app anyway.
- [x] run tests with jupyter_server to validate assumptions about APIs used by the mixins

concludes #2601

cf https://github.com/jupyterlab/jupyterlab/issues/8704

/cc @blink1073 once this is tested and works, I believe you can ship a copy of `mixins.py` in jupyterlab_server to be compatible with existing jupyterhub releases, so you don't have to wait for a hub release, if that would help.


blink1073:
> I believe you can ship a copy of mixins.py in jupyterlab_server to be compatible with existing jupyterhub releases, so you don't have to wait for a hub release, if that would help.

Fantastic, thanks!  This is what @Zsailer and I had concluded yesterday as well.

Zsailer:
@minrk, thanks for taking this on. I've added myself as a reviewer to this PR once it's ready. Also, I'm happy to help this next week if you'd like some extra hands.

minrk:
OK, I think this is ready to go. The `make_singleuser_app(App)` function only takes a single argument, the `traitlets.config.Application` class to be wrapped. If it's a subclass of `notebook.notebookapp.NotebookApp` or `jupyter_server.serverapp.ServerApp` (including those classes themselves), it should Just Work®. Setting config:

```python
c.Spawner.environment = {
    "JUPYTERHUB_SINGLEUSER_APP": "jupyterlab_server.app.LabServerApp", # default to jupyterlab; requires jupyterhub >= 1.2
}
```

should result in using the lab server app wrapped in singleuser auth.

As it is, `mixins.py` should be using imports that work at least back to jupyterhub 1.0, so it work exactly as-is if you want to ship a copy in jupyterlab_server (or jupyterlab, I'm not sure!). That way, you can still support at least jupyterhub >= 1.0 with:

```python
c.Spawner.cmd = ["jupyter", "labhub"]
```


I did run into the hiccup in https://github.com/jupyter/jupyter_server/pull/257 causing the tests to fail, but the latest commit ought to work around that, I think.

PR #3131, opened 2020-08-03
Allow head requests for the health endpoint

rkevin-arch:
One of the uptime checkers we're using only sends HEAD requests, and tornado doesn't like the HEAD requests and return a 405. We currently have a workaround as follows in the helm chart for Z2JH, but a more permanent solution would be ideal.
```yaml
hub:
  extraConfig:
    enable-head-request-on-health: |
      from jupyterhub.handlers.base import BaseHandler
      class HealthCheckHandler(BaseHandler):
          def get(self, *args):
              self.finish()
          def head(self, *args):
              self.finish()
      c.JupyterHub.extra_handlers.append((r'/health2$', HealthCheckHandler))
```

---

Edit by @consideRatio, related PR in BinderHub: https://github.com/jupyterhub/binderhub/pull/1137

### DIFF
--- a/jupyterhub/handlers/pages.py
+++ b/jupyterhub/handlers/pages.py
@@ -658,6 +658,9 @@ class HealthCheckHandler(BaseHandler):
     def get(self, *args):
         self.finish()
 
+    def head(self, *args):
+        self.finish()
+
 
 default_handlers = [
     (r'/', RootHandler),


rkevin-arch:
What's the status of this PR? Let me know if there's anything I can do to get this merged.

rkevin-arch:
My guess is to just have something in the function, rather than just a `pass` statement, but that's just a guess.

minrk:
@betatim you're absolutely right that `self.finish()` is unnecessary. We also don't need the `*args` argument, since that's only for parsed URLs with variable segments, e.g. `/user/(.+)`, which isn't used for the health endpoint.

We could just as easily have the simpler:

```python
def get(self):
    # nothing to do, just respond as quickly as we can
    pass

head = get
```

if anyone wants to make that PR.

PR #3132, opened 2020-08-05
Fix the services dropdown on the admin page

### LINKED ISSUE #3003: No services dropdown on the admin page
#### Expected behaviour

On the admin page, clicking on `Services` in the nav bar should show the list of services.

#### Actual behaviour

services-admin-page

It might be because the click event on the `Services` item in the navigation bar is being intercepted.

### How to reproduce

- Have one JupyterHub managed service running
- Go to the `Admin` page
- Click on `Services`

### Your personal set up

- OS: Ubuntu 19.10
- Version: JupyterHub 1.1.0
- Browser: Latest Firefox and Chromium 
- Configuration: Hub running with one managed service locally

pabepadu:
Fixes #3003 opened by @jtpio 
Remove the unused bootstrap variable which canceled the dropdown events.
I tested the admin actions with different browsers, everything works fine.

### DIFF
--- a/share/jupyterhub/static/js/admin.js
+++ b/share/jupyterhub/static/js/admin.js
@@ -1,9 +1,8 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
 
-require(["jquery", "bootstrap", "moment", "jhapi", "utils"], function(
+require(["jquery", "moment", "jhapi", "utils"], function(
   $,
-  bs,
   moment,
   JHAPI,
   utils


pabepadu:
Hi @GeorgianaElena,

To be fully honest, it's not obvious for me too (I'm not a javascript expert).

I add a breakpoint inside the require function in admin.js, and the `bs` variable is `undefined`.
```javascript
require(["jquery", "bootstrap", "moment", "jhapi", "utils"], function(
  $,
  bs,
  moment,
  JHAPI,
  utils
) {
  "use strict";
...
```
I try to dig on issues between requiredjs and bootstrap 3, but it seems the `bootstrap` variable is correctly wrapped in page.html:
```html
...
    <script>
      require.config({
          {% if version_hash %}
          urlArgs: "v={{version_hash}}",
          {% endif %}
          baseUrl: '{{static_url("js", include_version=False)}}',
          paths: {
            components: '../components',
            jquery: '../components/jquery/dist/jquery.min',
            bootstrap: '../components/bootstrap/dist/js/bootstrap.min',
            moment: "../components/moment/moment",
          },
          shim: {
            bootstrap: {
              deps: ["jquery"],
              exports: "bootstrap"
            },
          }
      });
    </script>
...
```

Sorry I can't explain why bootstrap is corrupted by requiredjs.

GeorgianaElena:
Thanks for explaining how you discovered the solution @pabepadu :heart: 
I tested this myself and it indeed solves the problem :tada:

After digging into why this works (I'm not a javascript expert either :sweat_smile:), I *think* the reason why this was failing in the first place was that we were importing bootstrap twice (once in `page.html` and once in `admin.js`), hence there were two `
click` events associated with the dropdown, each trying to handle the event and apparently the result was that none of them did.

The dropdown events from the admin tab:
two-events

The dropdown events from the token tab:
one-event

Not sure if this is truly the reason of the bug, but I see no harm in removing `bootstrap` from `admin.js` since it's not used and it solves the issue.

Thanks a lot for fixing this @pabepadu and congratulations for your fist JupyterHub PR :rocket: :heart:

pabepadu:
Thank a lot for the explanation and for the merge @GeorgianaElena

PR #3133, opened 2020-08-05
Allow JupyterHub.default_url to be a callable

danlester:
This allows c.JupyterHub.default_url to be a Unicode as before (e.g. '/hub/home') but also allows a callable to be provided instead, calculating the initial URL for a logged-in user based on the user object.

For example in jupyterhub_config.py:

```
def default_url_fn(user):
    if user:
        if user.admin:
            return '/hub/admin'
    return '/hub/home'

c.JupyterHub.default_url = default_url_fn
```

This is useful when the JupyterHub is being used in large part for authentication, but users are often there for other services such as [ContainDS Dashboards](https://cdsdashboards.readthedocs.io/).

Regular users might be redirected to a specific 'intro' dashboard, whereas admin users (or users belonging to a particular group etc) would prefer to go to the main JupyterHub.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1322,12 +1322,12 @@ class JupyterHub(Application):
 
         By default, redirects users to their own server.
 
-        Can be a Unicode string (e.g. '/hub/home') or a callable based on the user object:
+        Can be a Unicode string (e.g. '/hub/home') or a callable based on the handler object:
 
-            def default_url_fn(user):
-                if user:
-                    if user.admin:
-                        return '/hub/admin'
+            def default_url_fn(handler):
+                user = handler.current_user
+                if user and user.admin:
+                    return '/hub/admin'
                 return '/hub/home'
 
             c.JupyterHub.default_url = default_url_fn
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1324,6 +1324,8 @@ class JupyterHub(Application):
 
         Can be a Unicode string (e.g. '/hub/home') or a callable based on the handler object:
 
+        ::
+        
             def default_url_fn(handler):
                 user = handler.current_user
                 if user and user.admin:
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -55,6 +55,7 @@ from traitlets import (
     Instance,
     Bytes,
     Float,
+    Union,
     observe,
     default,
     validate,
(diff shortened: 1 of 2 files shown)


manics:
Are others likely to want additional parameters passed into the callable in future? It's a bit of a pain to add them later in a backwards compatible manner so it's worth considering this now.

danlester:
@manics Good question. I think there is only a limited amount of context that is likely to be interesting to the callable given the current codebase - maybe some components of 'request', or site config. But of course we don't know how the JupyterHub codebase will evolve in the future.

Do you think it would be worth specifying in the documentation that the callable handler should have a **kwargs argument too, so that (as you suggest) at least if something does come up in the future, it shouldn't immediately break existing configs?

minrk:
> Do you think it would be worth specifying in the documentation that the callable handler should have a **kwargs

I'm not sure about the impact of documenting this requirement, since there can be no consequences of failing to do so until possibly breaking changes in the future.

We can use [backcall](https://pypi.org/project/backcall/) to make backward-compatible updates to hook arguments. But maybe to start we should pass the `handler` object, which gives access to url params, etc. (and ~everything else!). This has been used in spawners for custom launch parameters.

danlester:
So do people think the signature should _only_ take handler now? E.g.

```
def default_url_fn(handler):
   user = handler.current_user
    if user:
        if user.admin:
            return '/hub/admin'
    return '/hub/home'
```

In some ways I think this complicates things for the administrator, having to understand a Tornado-based class, whereas before it was only User that was available. But in practice these snippets are designed by experts and copy-and-pasted by the rest!

I will change to handler-only as above unless I hear further feedback. Thanks for your thoughts so far.

minrk:
> So do people think the signature should only take handler now?

Sorry, that wasn't my original intention. I had meant `default_url_fn(user, handler)` where we *expect* a custom default url function to be interested in the user, since that's what prompted the change, but other data is accessible on the handler, so it can still be accessed without needing a new jupyterhub release. That said, I think it's okay to just take handler as this PR stands now, given your point about who tends to write these hooks. What do you think, @danlester?

jupyterhub/app.py
```diff
@@ -1314,12 +1315,23 @@ def add_url_prefix(prefix, handlers):
...
+        Can be a Unicode string (e.g. '/hub/home') or a callable based on the handler object:
+
+            def default_url_fn(handler):
+                user = handler.current_user
+                if user and user.admin:
+                    return '/hub/admin'
+                return '/hub/home'
+
+            c.JupyterHub.default_url = default_url_fn
+        """,
```
consideRatio, line 1334: This section seem to be giving this warning (treated like an error). Perhaps it is because of the code example indentation?

```
Warning, treated as error:
/home/circleci/project/jupyterhub/app.py:docstring of jupyterhub.app.JupyterHub.default_url:10:Unexpected indentation.
Makefile:64: recipe for target 'html' failed
make: *** [html] Error 2
```

danlester:
@minrk Yes, let's just leave as default_url_fn(handler) to avoid duplication.

@consideRatio Thanks for spotting - it wasn't the indentation as such, but it needed a :: to mark it as a code block in rst. (I've just pushed a fix.)

danlester:
Not too sure why docs are still failing in circleci... It works for me locally now. Any ideas?

minrk:
The remaining doc build failure is not related to these changes, but because this PR is forked before #3151. Should be AOK after merge.

Thanks!

PR #3136, opened 2020-08-07
Add a footer block + wrap the admin footer in this block

pabepadu:
Add a footer block in the base page template to be able to customize it since the admin page have a footer.
In the admin page, wrap the existing footer part into the new footer block.


minrk:
Thanks!

PR #3137, opened 2020-08-07
admin page sorts on spawner last_activity instead of user last_activity

lydian:
Currently the admin page last_activity is sorted on User's last activity, which makes it work odd because the column is actually showing the last_activity spawners.   This PR changed the sort column to be spawner's last activity.


minrk:
Thanks!

PR #3140, opened 2020-08-10
jupyterhub/utils: Load system default CA certificates in make_ssl_context

chancez:
Fixes issues with OAuth flows when internal_ssl is enabled.
When internal_ssl was enabled requests to non-internal endpoints failed
because the system CAs were not being loaded.

This caused failures with public OAuth providers with public CAs since
they would fail to validate.

### DIFF
--- a/jupyterhub/utils.py
+++ b/jupyterhub/utils.py
@@ -85,6 +85,7 @@ def make_ssl_context(keyfile, certfile, cafile=None, verify=True, check_hostname
         return None
     purpose = ssl.Purpose.SERVER_AUTH if verify else ssl.Purpose.CLIENT_AUTH
     ssl_context = ssl.create_default_context(purpose, cafile=cafile)
+    ssl_context.load_default_certs()
     ssl_context.load_cert_chain(certfile, keyfile)
     ssl_context.check_hostname = check_hostname
     return ssl_context
@@ -580,7 +581,7 @@ def utcnow():
 def _parse_accept_header(accept):
     """
     Parse the Accept header *accept*
-    
+
     Return a list with 3-tuples of
     [(str(media_type), dict(params), float(q_value)),] ordered by q values.
     If the accept header includes vendor-specific types like::


jupyterhub/utils.py
```diff
@@ -85,6 +85,7 @@ def make_ssl_context(keyfile, certfile, cafile=None, verify=True, check_hostname
         return None
     purpose = ssl.Purpose.SERVER_AUTH if verify else ssl.Purpose.CLIENT_AUTH
     ssl_context = ssl.create_default_context(purpose, cafile=cafile)
+    ssl_context.load_default_certs()
```
manics, line 88: Is this dependent on the `purpose`? https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_default_certs says it can be `Purpose.SERVER_AUTH` (default) and `Purpose.CLIENT_AUTH`?

chancez: It is dependent, though in my case I only need it to load server certs.

That being said, `verify` isn't ever specified that I can tell when calling `make_ssl_context` in the the code base. so it's always `SERVER_AUTH` I think. Additionally, when I reviewed the history of changes related to internal_ssl, it seems that it was added without ever being used.

I'd say we could probably remove `purpose` and `verify` entirely from the function. What do you think?

jupyterhub/spawner.py
```diff
@@ -944,15 +944,13 @@ async def move_certs(self, paths):
...
 
         Returns:
-            dict: a list (potentially altered) of paths for key, cert,
-            and CA.
-            These paths should be resolvable and readable
-            by the notebook server to be launched.
-
+            dict: a list (potentially altered) of paths for key, cert, and CA.
+            These paths should be resolvable and readable by the notebook
+            server to be launched.
```
minrk, line 953: There's a conflict on these changes, but it looks like there aren't any intentional changes here. Rebasing without the changes to spawner.py should fix the conflict

jupyterhub/utils.py
minrk: The simpler fix is to pass the purpose arg to this function:

```suggestion
        ssl_context.load_default_certs(purpose)
```

that way we can still use this function for creating client_auth contexts.

chancez:
> The simpler fix is to pass the purpose arg to this function:

As I asked @manics in another comment, do we even need to keep the purpose/verify argument? At this point, nothing calling this function passes an argument to `verify` from what I can tell, so it's always being used to construct a server auth context.

jupyterhub/utils.py
chancez: Do we even need the `verify`/`purpose` anymore? It doesn't like `verify` is ever passed as an argument, so it looks like it's just dead code at this point.

manics: @chancez Apologies for dropping this. I'll let @minrk confirm as he knows this better, but I think this is considered a public function that could be used by external components, so removing a parameter is an API change.

chancez:
> FAILED jupyterhub/tests/test_version.py::test_check_version[0.8.0--30-Single-user server has no version header]

Seems unrelated, but not sure.

manics:
I restarted the failing test and it's now passing 😀

jupyterhub/utils.py
chancez: I updated it to take purpose. I think removal could be a separate change from this PR anyways.

chancez:
Woot. Can we get https://github.com/jupyterhub/jupyterhub-idle-culler/pull/11 reviewed too? It's about enabling SSL support for the idle culler + includes similar changes as this PR.

PR #3142, opened 2020-08-13
Document external service api_tokens better

snickell:
We're implementing an externally managed service (implemented in javascript/node), following the [implementing your own auth](https://jupyterhub.readthedocs.io/en/stable/reference/services.html#implementing-your-own-authentication-with-jupyterhub) docs.

Getting this to work was difficult (we've spent a couple days debugging, just getting 403 errors, playing guessing games, and reading lots of source code), and two additions to the docs would have saved us a couple days:

- Explicitly mention the min-8-char constraint on (at least `api_token`) secrets, if the api_token is less than 8 characters, which we used in dev to prevent typing mistakes, it will silently not be registered, and trying to use it will always return 403s.
- Connect the api_token in the configuration with the one mentioned in auth requests. This probably wouldn't have been as important, but when debugging "all bets are off", and we started to question which token was which, including confusing this with the proxy API token at one point. I think explicitly mentioning that the tokens mentioned in the two sections are one and the same will help other people.

Co-authored-by: Mike Situ <msitu@ceresimaging.net>

### DIFF
--- a/docs/source/reference/services.md
+++ b/docs/source/reference/services.md
@@ -151,6 +151,8 @@ c.JupyterHub.services = [
     {
         'name': 'my-web-service',
         'url': 'https://10.0.1.1:1984',
+        # any secret >8 characters, you'll use api_token to
+        # authenticate api requests to the hub from your service
         'api_token': 'super-secret',
     }
 ]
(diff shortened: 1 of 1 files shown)


manics (approved):
This looks good to me. The 8 character requirement orignates in https://github.com/jupyterhub/jupyterhub/blob/e2ee18fa868b67e792e1b5627ec4487ff9a701a8/jupyterhub/orm.py#L321

PR #3147, opened 2020-08-17
Get error description from error key vs error_description key

### LINKED ISSUE #3146: 500 error does not output error description
### Bug description

Launching a single user notebook leads to a 500 error (as expected) if the user's server fails to authenticate with the JupyterHub. However, the error's description is not displayed due to an invalid key.

#### Expected behavior

Display the user error in the 500 error template page.

500_error_with_message

#### Actual behaviour

Generic 500 error.

500_error_without_message

Additionally, the logs display `TypeError: can only concatenate str (not "NoneType") to str` since the key/value does not exist (None).

### How to reproduce

Launch a single user notebook that has an unrecognized (invalid) client or token to get a `401` error.

jgwerner:
Fetch description from `error` key if available or `error_description` key as a fallback.

Closes #3146

### DIFF
--- a/jupyterhub/services/auth.py
+++ b/jupyterhub/services/auth.py
@@ -371,9 +371,9 @@ class HubAuth(SingletonConfigurable):
             )
             app_log.warning(r.text)
             msg = "Failed to check authorization"
-            # pass on error_description from oauth failure
+            # pass on error from oauth failure
             try:
-                description = r.json().get("error_description")
+                description = r.json().get("error")
             except Exception:
                 pass
             else:
--- a/jupyterhub/services/auth.py
+++ b/jupyterhub/services/auth.py
@@ -373,10 +373,10 @@ class HubAuth(SingletonConfigurable):
             msg = "Failed to check authorization"
             # pass on error from oauth failure
             try:
-                description = (
-                    r.json().get("error")
-                    if r.json().get("error")
-                    else r.json().get("error_description")
+                response = r.json()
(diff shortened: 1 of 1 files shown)


consideRatio:
Closing / Reopening to try if I can get the docs build failure to go away after a fix in master.

jgwerner:
Thanks @consideRatio if you need me to pull/rebase let me know 👍

consideRatio:
Try that @jgwerner, i failed with close / reopen! :heart:

jupyterhub/services/auth.py
```diff
@@ -371,9 +371,13 @@ def _api_request(self, method, url, **kwargs):
...
             msg = "Failed to check authorization"
-            # pass on error_description from oauth failure
+            # pass on error from oauth failure
             try:
-                description = r.json().get("error_description")
+                description = (
+                    r.json().get("error")
+                    if r.json().get("error")
+                    else r.json().get("error_description")
+                )
```
minrk, line 380: I think error_description should be the more specific explanation and preferred, if defined, in which case the priority should be reversed:

```suggestion
                response = r.json()
                # prefer more specific 'error_description', fallback to 'error'
                description = response.get("error_description", response.get("error", "Unknown error"))
```

jgwerner: Ah great point, I'll push an update based on your suggestion. Thanks!

minrk:
Thanks!

minrk:
Thanks!

PR #3150, opened 2020-08-19
update prometheus metrics for server spawn when it fails with exception

yhal-nesi:
When the spawner start() fails with exception, server_spawn_duration_seconds_bucket should be updated.


jupyterhub/handlers/base.py
```diff
@@ -936,6 +936,17 @@ def _clear_spawn_future(f):
...
+
+        # when spawn fails with an exception
+        # update prometheus metrics
+        def _metrics_server_spawn_failed(f):
+            if f.exception():
+                SERVER_SPAWN_DURATION_SECONDS.labels(
+                    status=ServerSpawnStatus.failure
+                ).observe(time.perf_counter() - spawn_start_time)
+
+        finish_spawn_future.add_done_callback(_metrics_server_spawn_failed)
```
GeorgianaElena, line 948: Instead of registering a new callback, lets update the metric in ` _track_failure_count`  (around [this line](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/handlers/base.py#L947)).
This way, we also handle the case when the spawn gets cancelled.

What do you think?

GeorgianaElena:
Thanks a lot for opening the PR @yhal-nesi! :tada:
I just left a comment inline :sun_with_face:

jupyterhub/handlers/base.py
yhal-nesi: Yep, this is much cleaner! Updated the change.

PR #3151, opened 2020-08-19
docs: please docs linter (move_cert docstring)

consideRatio:
This fixes [a CircleCI build failure](https://app.circleci.com/pipelines/github/jupyterhub/jupyterhub/160/workflows/025ff595-c247-43b8-9492-008e765d2e74/jobs/2390) noticed in another PR failing with the following error.

```
Warning, treated as error:
/home/circleci/project/jupyterhub/spawner.py:docstring of jupyterhub.spawner.Spawner.move_certs:6:Field list ends without a blank line; unexpected unindent.
Makefile:68: recipe for target 'html' failed
make: *** [html] Error 2
```

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -943,14 +943,14 @@ class Spawner(LoggingConfigurable):
         """Takes certificate paths and makes them available to the notebook server
 
         Arguments:
-            paths (dict): a list of paths for key, cert, and CA. These paths
-            will be resolvable and readable by the Hub process, but not
-            necessarily by the notebook server.
+            paths (dict): a list of paths for key, cert, and CA.
+                These paths will be resolvable and readable by the Hub process,
+                but not necessarily by the notebook server.
 
         Returns:
             dict: a list (potentially altered) of paths for key, cert, and CA.
-            These paths should be resolvable and readable by the notebook server
-            to be launched.
+                These paths should be resolvable and readable by the notebook
+                server to be launched.
 
 
         `.move_certs` is called after certs for the singleuser notebook have
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -943,15 +943,14 @@ class Spawner(LoggingConfigurable):
         """Takes certificate paths and makes them available to the notebook server
 
         Arguments:
-            paths (dict): a list of paths for key, cert, and CA.
-            These paths will be resolvable and readable by the Hub process,
-            but not necessarily by the notebook server.
+            paths (dict): a list of paths for key, cert, and CA. These paths
+            will be resolvable and readable by the Hub process, but not
+            necessarily by the notebook server.
 
         Returns:
-            dict: a list (potentially altered) of paths for key, cert,
-            and CA.
-            These paths should be resolvable and readable
-            by the notebook server to be launched.
+            dict: a list (potentially altered) of paths for key, cert, and CA.
+            These paths should be resolvable and readable by the notebook server
+            to be launched.
 
 
         `.move_certs` is called after certs for the singleuser notebook have
@@ -1618,5 +1617,5 @@ class SimpleLocalProcessSpawner(LocalProcessSpawner):
         return env
 
     def move_certs(self, paths):
-        """No-op for installing certs"""
+        """No-op for installing certs."""
         return paths


betatim:
There are a few places in `jupyterhub/spawner.py` where a `move_certs` method is defined. I don't know how to tell from the error message which one it is that needs fixing :-/

jupyterhub/spawner.py
```diff
@@ -1618,5 +1618,7 @@ def user_env(self, env):
         return env
 
     def move_certs(self, paths):
-        """No-op for installing certs"""
+        """
+        No-op for installing certs
+        """
         return paths
```
GeorgianaElena, line 1624: I think it also wants for the file to end with an empty line.
```suggestion
        return paths

```

consideRatio:
I could only reproduce this locally with sphinx 3. With sphinx 2, it works.

betatim:
Sounds reasonable that having only one line as description per argument is what the linter wants. I don't know if I agree with who ever made the linter rule but it seems like a good candidate for why it is failing. Scrolling through `spawner.py` I can't find any other examples that use this way of formatting the docstring except for L887/`create_certs`. Confusingly enough that looks like it has a multiline string for the argument :-/

consideRatio:
I'm not sure what went wrong, but apparently by making sure that the lines under `Arguments:` and `Returns:` wrapped according to a max 80 chars rule the warning (turned into an error by a flag) went away.

consideRatio:
I isolated the issue to the following diff, this fixes the error if there is a trailing comma on line1 or line2, but the line0 doesn't care if it has a comma on it or not.

```diff
        Arguments:
            dummy (dict): line0
-           line1,
+           line1
            line2
```

Due to this, my rewrapping fixed the issue, as a trailing comma on a line shifted so there now was a word behind it on that line. In other words, this bug can happen again.

jupyterhub/spawner.py
consideRatio: No worries @GeorgianaElena :) Thanks for thinking about this with me!

jupyterhub/spawner.py
```diff
@@ -943,15 +943,14 @@ async def move_certs(self, paths):
         """Takes certificate paths and makes them available to the notebook server
 
         Arguments:
-            paths (dict): a list of paths for key, cert, and CA.
-            These paths will be resolvable and readable by the Hub process,
-            but not necessarily by the notebook server.
+            paths (dict): a list of paths for key, cert, and CA. These paths
+            will be resolvable and readable by the Hub process, but not
+            necessarily by the notebook server.
```
minrk:
```suggestion
            paths (dict): a list of paths for key, cert, and CA.
              These paths will be resolvable and readable by the Hub process,
              but not necessarily by the notebook server.
```

jupyterhub/spawner.py
```diff
@@ -943,15 +943,14 @@ async def move_certs(self, paths):
...
 
         Returns:
-            dict: a list (potentially altered) of paths for key, cert,
-            and CA.
-            These paths should be resolvable and readable
-            by the notebook server to be launched.
+            dict: a list (potentially altered) of paths for key, cert, and CA.
+            These paths should be resolvable and readable by the notebook server
+            to be launched.
```
minrk:
```suggestion
            dict: a list (potentially altered) of paths for key, cert, and CA.
              These paths should be resolvable and readable by the notebook server
              to be launched.
```

minrk (commented):
The ` unexpected unindent` part of the error message points vaguely in the right direction: these extended descriptions should be indented relative to the argument list, not on the same column. Re-wrapping to avoid ending the lines on punctuation may avoid errors but might also result in incorrect output. I've made a couple of suggestions that fix the error for me.

I generally try to use [semantic line breaks](https://rhodesmill.org/brandon/2012/one-sentence-per-line/) in plaintext prose which means lots of lines ending in punctuation. I would like to keep that where possible, since it definitely makes diffs better, and I think improves source readability a bit.

minrk (approved):

minrk:
Thanks for digging into the weird error! Latest change looks great to me, feel free to merge once CI agrees that it fixed the error.

jupyterhub/spawner.py
consideRatio: Resolved in 0679586 (but with four spaces instead of two, hmm...)

jupyterhub/spawner.py
consideRatio: Resolved in 0679586 (but with four spaces instead of two, hmm...)

consideRatio:
> I generally try to use semantic line breaks in plaintext prose which means lots of lines ending in punctuation. I would like to keep that where possible, since it definitely makes diffs better, and I think improves source readability a bit.

Excellent pointer! It's good to put words to ideas like this. I agree but I didn't have a way to express this sentiment until now.

consideRatio:
Having concluded that Google Style Python Docstrings allow for 2 as well as 4 spaces to indent the additional lines, I'll settle and merge this with 4 space-indentation just because that was what I ended up doing without much thinking. It does look quite nice with two spaces, but it would also be conflicting with the norm of always using 4 in python so I'll just go with this now.

https://google.github.io/styleguide/pyguide.html#doc-function-args

consideRatio:
Thanks @betatim @GeorgianaElena @minrk for your help! :heart:

PR #3160, opened 2020-08-27
Control service display

rcthomas:
The Hub UI now includes a dropdown that lists links to services.  Some services may not have a particularly useful UI that makes sense to expose.  This PR:

-  adds a configurable boolean `display` traitlet to the `Service` class (default `True`),
-  changes the base handler to read the `display` value and omit services we don't want to provide access to, and
-  includes the `display` variable in the API model

The traitlet itself is documented but I haven't updated the `Service` class doc-string itself to include it.  I notice that some other fields (e.g. `info`) are not included in the doc-string either.  I interpret it to mean there are some fields that are much less important to call out, and I could see `display` being one of those (though I find it useful!).

### DIFF
--- a/jupyterhub/apihandlers/services.py
+++ b/jupyterhub/apihandlers/services.py
@@ -23,6 +23,7 @@ def service_model(service):
         'command': service.command,
         'pid': service.proc.pid if service.proc else 0,
         'info': service.info,
+        'display': service.display,
     }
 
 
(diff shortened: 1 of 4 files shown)


rcthomas:
Failing test appears to be a timeout, but everything else looks good I think

minrk:
Makes sense, thanks!

PR #3167, opened 2020-09-10
[WIP] Remove @ from list of safe chars when escaping username

danlester:
Reading up a little bit on URLs I think JupyterHub should not allow @ in the server's URL - it should be URL encoding it I believe. See [here](https://perishablepress.com/stop-using-unsafe-characters-in-urls/).

While @ is technically valid in URLs, it's not really valid in the path and should be URL encoded there.

IMO we shouldn't be seeing URLs like https://myjupyterhub.net/user/dan@test.com/
They should be https://myjupyterhub.net/user/dan%40test.com/
which admittedly doesn't look so nice!

Please note I haven't tested my code change - other aspects might need amending too - but wanted to see if anyone had any important opinions on this first.

This issue has caused confusion [here](https://github.com/ideonate/jhsingle-native-proxy/pull/6). (Although there is also arguably another fix needed in the component in question...)

### DIFF
--- a/jupyterhub/user.py
+++ b/jupyterhub/user.py
@@ -396,7 +396,7 @@ class User:
     @property
     def escaped_name(self):
         """My name, escaped for use in URLs, cookies, etc."""
-        return quote(self.name, safe='@~')
+        return quote(self.name, safe='~')
 
     @property
     def json_escaped_name(self):


manics:
Thanks for bringing this up. I'm not in a position to say whether we should do this or not, though it's potentially a breaking change depending on how other components decode (or not!) the character. For instance, the username is used to label some persistent resources. Might require some testing or a clear warning.

danlester:
Thank you for your thoughts @manics.

At first glance it seems to work for me and happens to solve the obscure issue we were seeing... but I think you're right that it would need a lot of testing to understand its interaction with other components. How does jupyter/notebook deal with these characters in file names for example.

Since I've brought this up, it might be a chance to document the current decision...

@anderbubble tried to make the change I was suggesting in a [commit](https://github.com/jupyterhub/jupyterhub/commit/fc49aac02be91aafcb73b2f3bfab7e3ee133b1bc) on 24 Feb 2016.

> Do not consider `@' character url-safe
> 
> Usernames that have an `@`-separated domain component
> break JupyterHub when the server expects to see query
> strings that contain an `@`, when browsers and other
> clients send `%40`.

The next day, @minrk [reverted the change](https://github.com/jupyterhub/jupyterhub/commit/450281a90abfdfd454850cbc0c2d5fdf2342c57e).

I'm not sure if things are working just by chance, and wonder if there is something else that needs fixing rather than allowing these technically-illegal URLs (if I've understood correctly). Things might have changed in the codebase and in browsers in 4 years anyway.

manics:
Travis CI check failed due to this change:
<details>

```
=================================== FAILURES ===================================

______________________________ test_user_escaping ______________________________

db = <sqlalchemy.orm.session.Session object at 0x7f7d2432ae80>

    def test_user_escaping(db):

        orm_user = orm.User(name='company\\user@company.com,\"quoted\"')

        db.add(orm_user)

        db.commit()

        user = User(orm_user)

        assert user.name == 'company\\user@company.com,\"quoted\"'

>       assert user.escaped_name == 'company%5Cuser@company.com%2C%22quoted%22'

E       AssertionError: assert 'company%5Cus...C%22quoted%22' == 'company%5Cus...C%22quoted%22'

E         - company%5Cuser@company.com%2C%22quoted%22

E         ?               ^

E         + company%5Cuser%40company.com%2C%22quoted%22

E         ?               ^^^

/home/travis/build/jupyterhub/jupyterhub/jupyterhub/tests/test_orm.py:83: AssertionError

___________________________ test_check_routes[has@] ____________________________

app = <jupyterhub.tests.mocking.MockHub object at 0x7f7d255b5400>

username = 'has@', disable_check_routes = None

    @pytest.mark.parametrize("username", ['zoe', '50fia', '秀樹', '~TestJH', 'has@'])

    async def test_check_routes(app, username, disable_check_routes):

        proxy = app.proxy

        test_user = add_user(app.db, app, name=username)

        r = await api_request(app, 'users/%s/server' % username, method='post')

        r.raise_for_status()

    

        # check a valid route exists for user

        routes = await app.proxy.get_all_routes()

        before = sorted(routes)

>       assert test_user.proxy_spec in before

E       AssertionError: assert '/@/space%20word/user/has%40/' in ['/@/space%20word/', '/@/space%20word/user/has@/']

E        +  where '/@/space%20word/user/has%40/' = <User(has@ 1/1 running)>.proxy_spec

/home/travis/build/jupyterhub/jupyterhub/jupyterhub/tests/test_proxy.py:179: AssertionError

----------------------------- Captured stdout call -----------------------------

13:05:34.799 [ConfigProxy] info: Adding route /@/space word/user/has@ -> http://127.0.0.1:46201

13:05:34.799 [ConfigProxy] info: Route added /@/space word/user/has@ -> http://127.0.0.1:46201

13:05:34.800 [ConfigProxy] info: 201 POST /api/routes/@/space%20word/user/has%40 

13:05:34.806 [ConfigProxy] info: 200 GET /api/routes 

----------------------------- Captured stderr call -----------------------------

[D 05:34.278 MockHub base:285] Recording first activity for <APIToken('525b...', user='admin')>

[D 05:34.289 MockHub user:289] Creating <class 'jupyterhub.tests.mocking.MockSpawner'> for has@:

[D 05:34.292 MockHub base:867] Initiating spawn for has@

[D 05:34.292 MockHub base:874] 0/100 concurrent spawns

[D 05:34.292 MockHub base:879] 0 active servers

[D 05:34.329 MockHub user:603] Calling Spawner.start for has@

[I 05:34.331 MockHub spawner:1457] Spawning /home/travis/virtualenv/python3.5.6/bin/python -m jupyterhub.tests.mocksu --port=46201

[D 05:34.356 MockHub spawner:1153] Polling subprocess every 30s

[I 200910 13:05:34 mocksu:55] Starting mock singleuser server at 127.0.0.1:46201

[D 200910 13:05:34 selector_events:65] Using selector: EpollSelector

[I 200910 13:05:34 web:2250] 200 GET /@/space%20word/user/has%40/ (127.0.0.1) 2.39ms

[W 05:34.796 MockHub _version:43] Single-user server has no version header, which means it is likely < 0.8. Expected 1.2.0dev

[I 05:34.796 MockHub base:897] User has@ took 0.505 seconds to start

[I 05:34.797 MockHub proxy:262] Adding user has@ to proxy /@/space%20word/user/has%40/ => http://127.0.0.1:46201

[D 05:34.797 MockHub proxy:766] Proxy: Fetching POST http://127.0.0.1:8001/api/routes/@/space%20word/user/has%40

[I 05:34.801 MockHub log:181] 201 POST /@/space%20word/hub/api/users/has@/server (admin@127.0.0.1) 527.62ms

[D 05:34.804 MockHub proxy:766] Proxy: Fetching GET http://127.0.0.1:8001/api/routes

--------------------------- Captured stdout teardown ---------------------------

13:05:34.827 [ConfigProxy] info: Removing route /@/space word/user/has@

13:05:34.827 [ConfigProxy] info: 204 DELETE /api/routes/@/space%20word/user/has%40 
```

</details>

Looking at https://pretty-rfc.herokuapp.com/RFC3986#reserved it sounds ambiguous whether the path component of the URL needs to be encoded:
>  If data for a URI component would conflict with a reserved character's purpose as a delimiter, then the conflicting data must be percent-encoded before the URI is formed.

Also worth noting we had problems with escaping/not escaping characters in https://github.com/jupyterhub/jupyter-server-proxy/pull/186

danlester:
@manics thank you especially for the links to jupyter-server-proxy. That is a very similar issue to the one I was seeing, although I think the reason is actually very different.

Yes, the tests would definitely need to be changed.

danlester:
Yeah, if [npmjs.com](https://www.npmjs.com/package/@ideonate/jupyter-containds) is allowed to use @ in their URLs maybe I should let you do so too :) Sorry if I misinterpreted the RFC there.

Feel free just to close if I'm either completely wrong or this subject isn't deterministic enough, although if anyone wants to document more thoroughly while we're on the subject please do.

manics:
Let's keep this open for now, it'd be good to have definitive reference we can refer to in future.

betatim:
Out of curiosity: what was/is the issue you are seeing that made you want to encode `@`s?

danlester:
Hi @betatim! This was the issue that came up:

https://github.com/ideonate/jhsingle-native-proxy/pull/6

It was very similar to the problems seen in using VSCode within jupyter-server-proxy (linked by Simon above) and in fact using the same kind of components (VSCode / code-server) within jhsingle-native-proxy. But their problems were using `@` further down the path in the URL, mine were at the CHP level (/user/user@name/). We had a simple fix in jhsingle-native-proxy that has solved the problem but it feels like a fudge around the way that CHP and/or JupyterHub happen to work.

minrk:
This is fiddly! I started working on symmetrical equivalence of reserved charactrs a long time ago in https://github.com/jupyterhub/configurable-http-proxy/pull/163 but ran out of time to think it through thoroughly enough. If I remember correctly, we should have equivalence in one direction: as long as we use `@` internally everywhere, strictly encoded requests with e.g. `@->%40`, routing should work, but if JupyterHub *internally* does the escapes and we internally store the routing table as the encoded URLs then the cases where non-encoded URLs are sent will break. Now, arguably that's not a big problem, as long as JupyterHub is consistent in producing links.

I opened #3168 to test that this is the case.

It does sound like from https://github.com/ideonate/jhsingle-native-proxy/pull/6 that there are at least some cases where this is not the case. I don't believe extensions like yours should need to be worrying too much about this, so maybe there is something in JupyterHub that does need to be fixed, such as perhaps what we pass in the JUPYTERHUB_SERVICE_URL, which populates the routing prefix. I think maybe that node handles it safely in both cases, while tornado handles it *as long as it's not in the prefix*.

For https://github.com/ideonate/jhsingle-native-proxy/pull/6 btw, it should probably not be the *fully* escaped prefix, but rather re-escaped with a new safe list:

   `quote(unquote(prefix))` to avoid double-encoding `%`. This is equivalent to `quote(prefix)` if `@` is the *only* character needing encoding, but will be wrong if there are other characters requiring escapes that are actually passed in their escaped form.

danlester:
Thanks @minrk for explaining! Good point in https://github.com/ideonate/jhsingle-native-proxy/pull/6 - it will be best only to think about `@` for now.

Also worth pointing to your commit https://github.com/jupyterhub/jupyterhub/commit/0c5d56483093a0f9315def6ee464aacc8f5f4dbc on 28 May 2018:

> tilde is a safe character in user URLs
> 
> Chrome unconditionally reverts any not-strictly-necessary escaping in URLs (this seems wrong?)

Which could also cause problems if different browsers have different ideas!

danlester:
Actually ~ is always considered safe in [urllib.parse.quote](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote) from Python 3.7 onwards... which references an RFC change!

Pugio:
This is affecting me when using `jupyter-server-proxy` for `code-server`. Our hub usernames are emails, and `code-server` sometimes makes requests of the form `/user/name%40domain/...` which the `jupyter-server-proxy` fails to resolve (ie. 404). It would be very helpful if JupyterHub mapped both "@" and "%40" to a single canonical path.

PR #3169, opened 2020-09-11
First step for implementing oauth scopes - update to rest-api.yml

IvanaH8:
Ref: scopes issue #1057 

The changes to the yml file aim to reflect oauth scopes definition for different permission levels that JupyterHub could use to manage token-based access. The definitions follow [OpenAPI 2.0 Specification ](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md).

Scopes will provide more specific access (e.g. read-only, access only users' servers, etc.) instead of "everything as the user/admin can do" (current state). Users, services and tokens will be associated with one or more roles where each role contains a list of permission scopes (ref.: user roles #2245). 

Users/services and tokens will be checked if they carry roles with necessary scopes for the requested action (in rest-api.yml the scopes listed for each API endpoint; if no scope defined, global security applies) through API handlers.

The full list of proposed scopes is available [here](https://hackmd.io/@ZBwjUKuOTz650XPDXRgrjQ/Bycgiah7v).

The Rest API documentation will show the added scopes as:

image

Please note comments with few additions:

- authorizationUrl and tokenUrl (line 19 and 20): the security definition requires absolute paths

- global security (lines 46-49): applies to all API endpoints unless overwritten on the operational level for an API endpoint (with a 'security' keyword); for the current state keeping both apiKey and oauth (either one can be used to access API endpoints with no scopes defined)

- scopes for /authorize/... endpoints (lines 713, 740 and 771)

### DIFF
--- a/docs/rest-api.yml
+++ b/docs/rest-api.yml
@@ -737,7 +737,7 @@ paths:
       summary: Identify a user or service from an API token
       security:
         - oauth2:
-          - read:all # minrk: is it really necessary to have a scope for this, or use self handler for token whoami?
+          - read:users:tokens # minrk: is it really necessary to have a scope for this, or use self handler for token whoami?
       parameters:
         - name: token
           in: path
--- a/docs/rest-api.yml
+++ b/docs/rest-api.yml
@@ -16,34 +16,34 @@ securityDefinitions:
   oauth2:
     type: oauth2
     flow: accessCode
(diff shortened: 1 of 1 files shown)


consideRatio:
@IvanaH8 wieee thank you for your work on this!!! :heart: :tada:!!!

PR #3176, opened 2020-09-17
Enable async support in jinja2 templates

yuvipanda:
Follows https://jinja.palletsprojects.com/en/2.11.x/api/#async-support

- This blocks the main thread fewer times
- We can use async methods inside templates too
- Provide sync versions of render_template for `write_error`

   write_error is a synchronous method called by an async
   method from inside the event loop. This means we can't just
   schedule an async render_templates in the same loop and wait
   for it - that would deadlock.

   jinja2 compiled your code differently based on wether you
   enable async support or not. Templates compiled with async
   support can't be used in cases like ours, where we already
   have an event loop running and calling a sync function. So
   we maintain two almost identical jinja2 environments

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -1264,7 +1264,7 @@ class BaseHandler(RequestHandler):
         # so we run it sync here, instead of making a sync version of render_template
 
         try:
-            html = self.render_template(f'{status_code}.html', sync=True, **ns)
+            html = self.render_template('%.html' % status_code, sync=True, **ns)
             print(html, flush=True)
         except TemplateNotFound:
             self.log.debug("No template for %d", status_code)
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -1476,14 +1476,10 @@ class UserUrlHandler(BaseHandler):
 
         # if request is expecting JSON, assume it's an API request and fail with 503
         # because it won't like the redirect to the pending page
-        if (
-            get_accepted_mimetype(
-                self.request.headers.get('Accept', ''),
-                choices=['application/json', 'text/html'],
-            )
-            == 'application/json'
-            or 'api' in user_path.split('/')
-        ):
+        if get_accepted_mimetype(
+            self.request.headers.get('Accept', ''),
+            choices=['application/json', 'text/html'],
+        ) == 'application/json' or 'api' in user_path.split('/'):
             self._fail_api_request(user_name, server_name)
             return
 
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -1156,16 +1156,36 @@ class BaseHandler(RequestHandler):
             "<a href='{home}'>home page</a>.".format(home=home)
         )
 
-    def get_template(self, name):
-        """Return the jinja template object for a given name"""
-        return self.settings['jinja2_env'].get_template(name)
+    def get_template(self, name, sync=False):
+        """
+        Return the jinja template object for a given name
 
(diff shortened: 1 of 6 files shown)


jupyterhub/handlers/base.py
```diff
@@ -1159,16 +1159,36 @@ def spawn_home_error(self):
...
         )
 
-    def get_template(self, name):
-        """Return the jinja template object for a given name"""
-        return self.settings['jinja2_env'].get_template(name)
+    def get_template(self, name, sync=False):
+        """
+        Return the jinja template object for a given name
+
+        If sync is True, we return a Template that is compiled without async suppor.
```
betatim, line 1166:
```suggestion
        If sync is True, we return a Template that is compiled without async support.
```

yuvipanda:
Whelp, it looks like async support in jinja2 requires Python 3.6.

I could shim it to do it synchronously in 3.5.

consideRatio:
@jupyterhub/jupyterhubteam I'd vote to stop supporting Python 3.5, the security support ended 13th september. Note that @yuvipanda in this PR also ran into a situation where dropping 3.5 support would also save some work. What do you think?

[image](https://devguide.python.org/#status-of-python-branches)

minrk:
This is awesome! I didn't know we could do this.

I agree that we should drop Python 3.5, maybe right after cutting 1.2?

jupyterhub/handlers/base.py
```diff
@@ -1159,16 +1159,36 @@ def spawn_home_error(self):
...
+
+        If sync is False, we return a Template that is compiled with async support
+        """
+        if sync:
+            key = 'jinja2_env_sync'
+        else:
+            key = 'jinja2_env'
+        return self.settings[key].get_template(name)
+
+    def render_template(self, name, sync=False, **ns):
```
minrk, line 1177: If we want to preserve backward-compatibility, we can default to `sync=True`. However, this backward-compatibility concern only applies to third-party template-based pages. These are *technically* possible, but I wouldn't call them "supported" since I specifically *do not* want JupyterHub customizations to add extra pages like notebook extensions do. Such extensions really do belong outside the Hub as services.

jupyterhub/handlers/base.py
```diff
@@ -1243,17 +1263,20 @@ def write_error(self, status_code, **kwargs):
...
             self.clear_header('Content-Length')
 
-        # render the template
+        # render_template is async, but write_error can't be!
+        # so we run it sync here, instead of making a sync version of render_template
+
         try:
-            html = self.render_template('%s.html' % status_code, **ns)
+            html = self.render_template('%s.html' % status_code, sync=True, **ns)
+            print(html, flush=True)
```
minrk, line 1271: debug

jupyterhub/handlers/login.py
```diff
@@ -89,8 +89,8 @@ async def get(self):
 class LoginHandler(BaseHandler):
     """Render the login page."""
 
-    def _render(self, login_error=None, username=None):
-        return self.render_template(
+    async def _render(self, login_error=None, username=None):
```
minrk, line 92: FWIW, you don't need to add this `async def` and `await` to a method that only wraps a function already returning an awaitable. The existing method will work fine and pass through the awaitable without needing two stacked awaits, one for the wrapper and another for the caller

rgbkrk:
This is super cool to see, thanks @yuvipanda

jupyterhub/handlers/login.py
yuvipanda: Done!

minrk:
Awesome!

PR #3184, opened 2020-09-23
Mention the PAM pitfall on fedora.

rainwoodman:
The Linux guide here is only applicable to Debian based systems.
On Fedora the shallow files are supposed to be only read by root, with PAM handling the rest (I am not sure about the magic behind this). 

This is just one of the subtle differences between the distributions, but at least we shouldn't confuse the Fedora users by asking them to change the permissions of shadow files. There isn't even a wheel group on Fedora since probably the Fedora Core days.

I settled using sshauthenticator. There may be a way to properly set up PAM authentication for Fedora, but I wasn't able to find out. I recall the furthest I got was authenticating using rhea's password for any user on the sudoer list, if I remember it accurately.

### DIFF
--- a/docs/source/reference/config-sudo.md
+++ b/docs/source/reference/config-sudo.md
@@ -120,6 +120,11 @@ the shadow password database.
 
 ### Shadow group (Linux)
 
+**Note:** On Fedora based distributions there is no clear way to configure
+the PAM database to allow sufficient access for autenticating with the target user's password
+from jupyterhub.As a workaround we recommend use an alternative authentication method, e.g.
+[sshauthenticator](https://github.com/andreas-h/sshauthenticator).
+
 ```bash
 $ ls -l /etc/shadow
 -rw-r-----  1 root shadow   2197 Jul 21 13:41 shadow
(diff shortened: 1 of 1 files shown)


docs/source/reference/config-sudo.md
```diff
@@ -120,6 +120,11 @@ the shadow password database.
 
 ### Shadow group (Linux)
 
+**Note:** On Fedora based distributions there is no clear way to configure
+the PAM database to allow sufficient access for autenticating with the target user's password
+from jupyterhub.As a workaround we recommend use an alternative authentication method, e.g.
+[sshauthenticator](https://github.com/andreas-h/sshauthenticator).
+
```
consideRatio, line 127:
```suggestion
**Note:** On Fedora based distributions there is no clear way to configure
the PAM database to allow sufficient access for autenticating with the target user's password
from JupyterHub. As a workaround we recommend use an alternative authentication method, e.g.
[sshauthenticator](https://github.com/andreas-h/sshauthenticator).

```

Formatting detail. I'm don't understand the details about this enough to review the actual content though =/

manics: Also a typo in `autenticating`.

I'm not sure about recommending ssh as an alternative though, maybe just link to https://github.com/jupyterhub/jupyterhub/wiki/Authenticators instead?

PR #3194, opened 2020-10-06
Add more server event testing

kiendang:
Add testing to make sure the correct events are emitted for server start/stop.


kiendang:
close in favor of #3218

PR #3195, opened 2020-10-06
add option to use a different Host header for referer checks

kylewm:
Hi folks 👋 , 

Here's the problem I'm trying to solve: I'm running JupyterHub on Kubernetes cluster with Istio. Our Istio is configured to internally route requests based on the `Host` header, so I need to actually overwrite the Host header that comes into node-proxy (e.g., `jupyterhub.example.com`) with the internal hostname (e.g., `jupyterhub.production.svc.cluster.local`). Otherwise any requests that the proxy sends just loop back to itself. 

node-proxy's `--change-origin` flag accomplishes this, and stores the original host in the conventional `X-Forwarded-Host` header.

This all works nicely until the check_referer function in JupyterHub tries to confirm the `Referer` and `Host` match. At this point I have headers like 

```
Referer: https://jupyterhub.example.com/hub/etc
Host: jupyterhub.production.svc.cluster.local
X-Forwarded-Host: jupyterhub.example.com
```

Proposed solution: Allow jupyterhub_config to define an alternative host header to use for the check_referer call. I'm not married to this solution, very open to thoughts and suggestions. Thanks!

### DIFF
--- a/jupyterhub/apihandlers/base.py
+++ b/jupyterhub/apihandlers/base.py
@@ -58,7 +58,7 @@ class APIHandler(BaseHandler):
 
         - allow unspecified host/referer (e.g. scripts)
         """
-        host = self.request.headers.get("Host")
+        host = self.request.headers.get(self.app.forwarded_host_header or "Host")
         referer = self.request.headers.get("Referer")
 
         # If no header is provided, assume it comes from a script/curl.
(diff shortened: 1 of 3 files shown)


minrk:
Sorry for leaving this for over a year!

Merged. Thanks for the PR!

PR #3203, opened 2020-10-11
Create config requires file to exist

dsblank:
Suggestion:

```python
jupyterhub --generate-config -f /etc/jupyterhub/jupyterhub_config.py
```

to:

```python
touch /etc/jupyterhub/jupyterhub_config.py
jupyterhub --generate-config -f /etc/jupyterhub/jupyterhub_config.py
```

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -255,7 +255,8 @@ class JupyterHub(Application):
 
     generate default config file:
 
-        jupyterhub --generate-config /etc/jupyterhub/jupyterhub_config.py
+        touch /etc/jupyterhub/jupyterhub_config.py
+        jupyterhub --generate-config -f /etc/jupyterhub/jupyterhub_config.py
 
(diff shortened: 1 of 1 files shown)


manics:
This shouldn't be needed, `jupyterhub --generate-config -f /tmp/jupyterhub_config.py` works fine for me.

Could you please:
- show us the full output of your error message
- provide full details of how you installed JupyterHub, ideally step by step instructions so we can attempt to replicate your problem

Thanks!

consideRatio:
I could not reproduce this need in the current JupyterHub master.

image

I also concluded that the permission errors could be avoided by using `sudo jupyter ...` which made sense as well. So I think the key to consider here, is that `sudo` is likely required for jupyterhub to write within /etc, rather than the need to create a file ahead of time.

Would you like to update this PR to describe this in some way @dsblank ?

manics:
No response, closing

PR #3204, opened 2020-10-12
Prevent OverflowErrors in exponential_backoff()

kreuzert:
We start JupyterLabs on a slurm based batch system. Therefore, JupyterLabs may take a while to start. This lead to OverflowErrors in exponential_backoff(). 

This small patch will prevent crashes while waiting for JupyterLabs.

### DIFF
--- a/jupyterhub/utils.py
+++ b/jupyterhub/utils.py
@@ -171,16 +171,9 @@ async def exponential_backoff(
         # add some random jitter to improve performance
         # this prevents overloading any single tornado loop iteration with
         # too many things
-        # Except Overflow Error. Otherwise start_wait * scale will be too big at some point.
-        try:
-            if scale == 0:
-                dt = min(max_wait, remaining)
-            else:
-                dt = min(max_wait, remaining, random.uniform(0, start_wait * scale))
-        except OverflowError:
-            scale = 0
-            dt = min(max_wait, remaining)
-        scale *= scale_factor
+        dt = min(max_wait, remaining, random.uniform(0, start_wait * scale))
+        if dt < max_wait:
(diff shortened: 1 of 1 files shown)


manics:
Do you know roughly what value triggers OverflowError in practice? I'm wondering if it'd be simpler to bound `dt` by a fixed max-value instead.

jupyterhub/utils.py
```diff
@@ -171,7 +171,15 @@ async def exponential_backoff(
         # add some random jitter to improve performance
         # this prevents overloading any single tornado loop iteration with
         # too many things
-        dt = min(max_wait, remaining, random.uniform(0, start_wait * scale))
+        # Except Overflow Error. Otherwise start_wait * scale will be too big at some point.
```
manics, line 174: Ignore my previous comment. Could this be simplified to something like:
```py
if dt < max_wait:
    dt = min(max_wait, remaining, random.uniform(0, start_wait * scale))
    scale *= scale_factor
```
else leave`dt` unchanged?

jupyterhub/utils.py
```diff
@@ -171,7 +171,15 @@ async def exponential_backoff(
...
+        try:
+            if scale == 0:
+                dt = min(max_wait, remaining))
```
consideRatio, line 177:
```suggestion
                dt = min(max_wait, remaining)
```

syntax error

jupyterhub/utils.py
kreuzert: Thanks for your response. I think any code snippet that prevents `scale *= scale_factor` from growing at some point will be fine.

minrk:
Thanks!

PR #3205, opened 2020-10-13
Logout on all devices and strict session ids

kreuzert:
Hi everyone,

This patch tries to improve security for JupyterHub, regarding the logout mechanism and running services.

As a User, I have no control over my previously used sessions on different computers or browsers. So I wanted to add a mechanism, which ensures that I'm the only person who is able to browse to my Service ( https://<jupyter_hub>/user/<user_name>/<server_name> ). 

For this, I added a simple "logout_on_all_devices" mechanism. 
You can use `c.JupyterHub.logout_on_all_devices` to set a default behavior and `c.JupyterHub.logout_on_all_devices_argname` to set an argument, which is checked when logging out. So you can send a GET request to `/hub/logout?onalldevices=false` and you will not be logged out from all devices even if `c.JupyterHub.logout_on_all_devices` is `true`.

As a User, I have now at least control if someone else has access to my JupyterHub control panel or other JupyterHub pages. But old browsers are still able to browse directly to my running Services. To avoid this behavior we added the following:

For each login, JupyterHub adds the session_id of the login_cookie to a list in the user's auth_state. If the user logs out, the session_id of the cookie will be removed from this list. 
Services may use the new `SessionIDAPIHandler` to check if the browser, which wants to access this service, has an "active" JupyterHub login cookie. The `session_id_cache_time` parameter will reduce traffic between service and JupyterHub. If I log out from all devices, all session ids are removed.

Test this patch:
```
$ virtualenv -p python3 venv
$ source venv/bin/activate
$ git clone --single-branch --branch strict_session_ids https://github.com/kreuzert/jupyterhub.git strict_session_ids
$ pip3 install -e ./strict_session_ids
$ pip3 install notebook
$ echo -e "c.JupyterHub.logout_on_all_devices = True\nc.JupyterHub.strict_session_ids = True\nc.Spawner.environment = {'JUPYTERHUB_SESSION_ID_REQUIRED': 'true', 'JUPYTERHUB_SESSION_ID_CACHE_TIME': '5'}\nc.Authenticator.enable_auth_state = True" >> jupyterhub_config.py
$ export JUPYTERHUB_CRYPT_KEY="3ded12ca7931404eac114e288cc9b89c"
$ jupyterhub -f jupyterhub_config.py

### Open private window in browser
###### Browse to http://localhost:8000 
###### Log in, start jupyterhub-singleuser
### Open normal window in browser
###### Browse to http://localhost:8000, log in
###### Logout in normal window
### now the user in private window can NOT:
#### 1. Use JupyterHub sites, which require an authentication
#### 2. Open a new browser tab with my running Service ( https://<jupyterhub>/user/<user_name>/<server_name> will lead to an 401 HTTP Error )
#### 3. Open a terminal/notebook/kernel in an active BrowserTab

### Login again in normal browser window
##### Normal browser window can use the Jupyter service
##### the private window cannot use this service
```

----

What are your thoughts about this patch? Is it useful, can it be improved?
Important: If you don't set `c.JupyterHub.strict_session_ids` or `c.JupyterHub.logout_on_all_devices` nothing will change with this patch.

Interesting logging outputs while testing (commented out):
jupyterhub/services/auth.py:494
jupyterhub/apihandlers/auth.py:97
jupyterhub/handlers/login.py:71, 82

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.vscode
 node_modules
 *.py[co]
 *~
--- a/jupyterhub/apihandlers/auth.py
+++ b/jupyterhub/apihandlers/auth.py
@@ -88,6 +88,7 @@ class TokenAPIHandler(APIHandler):
             )
         )
 
+
 class SessionIDAPIHandler(APIHandler):
     @token_authenticated
     async def get(self, username, session_id):
@@ -103,6 +104,7 @@ class SessionIDAPIHandler(APIHandler):
             self.write(json.dumps(model))
         else:
             raise web.HTTPError(404)
+
 
 class CookieAPIHandler(APIHandler):
     @token_authenticated
--- a/jupyterhub/apihandlers/auth.py
+++ b/jupyterhub/apihandlers/auth.py
@@ -98,9 +98,8 @@ class SessionIDAPIHandler(APIHandler):
         if db_user:
             if self.app.strict_session_ids:
                 session_ids = await self.app.load_session_ids(username)
-                _session_id = self.request.query_arguments.get('session_id', [b''])
-                session_id = [v.decode("utf-8") for v in _session_id]
-                if len(session_id) == 0 or session_id[0] not in session_ids:
+                session_id = self.request.headers.get("sessionid", "")
+                if (not session_id) or (session_id not in session_ids):
                     raise web.HTTPError(404)
             model = self.user_model(self.users[db_user])
             self.write(json.dumps(model))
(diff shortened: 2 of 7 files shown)


minrk:
Sorry for leaving you hanging on this one. We discussed this a bit more in #3342

To summarize, I think we should:

1. add an API endpoint to logout all existing sessions (i.e. reset cookie id, clear session ids)
2. add UI to the tokens page to "logout all existing sessions"
3. no configuration to do this on regular session logout

minrk:
Closing some stale PRs. This feature request is tracked as #3688

PR #3212, opened 2020-10-19
Implementing RBAC scope checking in API handlers

0mar:
Linked to #1057, #3169 and #2245 

Adjusting the API handlers and ORM to allow for more fine-grained, role-access control.

Concretely, this means
- [x] Identify which actions require which scopes
- [x] Implement scope-based access checking in API handlers
- [x] Write unit tests for scope resolving and scope filtering
- [ ] Implementing roles with corresponding scopes in ORM (#3215)
- [ ] ...

Available scopes: https://hackmd.io/@ZBwjUKuOTz650XPDXRgrjQ/Bycgiah7v

### DIFF
--- a/jupyterhub/apihandlers/hub.py
+++ b/jupyterhub/apihandlers/hub.py
@@ -9,11 +9,12 @@ from tornado.ioloop import IOLoop
 
 from .._version import __version__
 from ..utils import admin_only
+from ..utils import needs_scope
 from .base import APIHandler
 
 
 class ShutdownAPIHandler(APIHandler):
-    @admin_only
+    @needs_scope('shutdown')
     def post(self):
         """POST /api/shutdown triggers a clean shutdown
         
@@ -65,7 +66,7 @@ class RootAPIHandler(APIHandler):
 
 
 class InfoAPIHandler(APIHandler):
-    @admin_only
+    @needs_scope('admin')  # Todo: Probably too strict
     def get(self):
         """GET /api/info returns detailed info about the Hub and its API.
 
--- a/jupyterhub/apihandlers/hub.py
+++ b/jupyterhub/apihandlers/hub.py
@@ -66,7 +66,6 @@ class RootAPIHandler(APIHandler):
 
 
 class InfoAPIHandler(APIHandler):
-    @needs_scope('admin')  # Todo: Probably too strict
     def get(self):
         """GET /api/info returns detailed info about the Hub and its API.
 
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -28,7 +28,7 @@ class SelfAPIHandler(APIHandler):
     Based on the authentication info. Acts as a 'whoami' for auth tokens.
     """
 
-    @needs_scope('all')
+    # @needs_scope('read:users') # Should be read:users:user=username
     async def get(self):
         user = self.current_user
         if user is None:
@@ -41,12 +41,8 @@ class SelfAPIHandler(APIHandler):
 
 class UserListAPIHandler(APIHandler):
     @needs_scope('read:users')
-    def get(self, subset=None):
+    def get(self):
         users = self.db.query(orm.User)
-        if subset is not None:
-            users = users.filter(
-                User.name.in_(subset)
-            )  # Should result in only one db query
         data = [
             self.user_model(u, include_servers=True, include_state=True) for u in users
         ]
@@ -131,7 +127,7 @@ class UserAPIHandler(APIHandler):
     async def get(self, name, subset=None):
         user = self.find_user(name)
         if subset is not None:
-            if user not in subset:
+            if name not in subset:
                 raise web.HTTPError(403, "No access to users")
 
         model = self.user_model(
@@ -224,12 +220,14 @@ class UserAPIHandler(APIHandler):
 class UserTokenListAPIHandler(APIHandler):
     """API endpoint for listing/creating tokens"""
 
-    @needs_scope('users:tokens')
-    def get(self, name):
+    @needs_scope('read:users:tokens')
+    def get(self, name, subset=None):
         """Get tokens for a given user"""
         user = self.find_user(name)
         if not user:
             raise web.HTTPError(404, "No such user: %s" % name)
+        if subset is not None and name not in subset:
+            raise web.HTTPError(403, "No access to tokens")
 
         now = datetime.utcnow()
 
@@ -258,7 +256,6 @@ class UserTokenListAPIHandler(APIHandler):
             oauth_tokens.append(self.token_model(token))
         self.write(json.dumps({'api_tokens': api_tokens, 'oauth_tokens': oauth_tokens}))
 
-    @needs_scope('users:tokens')
     async def post(self, name):
         body = self.get_json_body() or {}
         if not isinstance(body, dict):
@@ -323,7 +320,6 @@ class UserTokenListAPIHandler(APIHandler):
 class UserTokenAPIHandler(APIHandler):
     """API endpoint for retrieving/deleting individual tokens"""
 
-    @needs_scope('read:users:tokens')
     def find_token_by_id(self, user, token_id):
         """Find a token object by token-id key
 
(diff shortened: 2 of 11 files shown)


jupyterhub/utils.py
```diff
@@ -296,6 +299,30 @@ def metrics_authentication(self):
         raise web.HTTPError(403)
 
 
+def needs_scope(scope):
+    """Decorator to restrict access to users or services with the required scope"""
+
+    def scope_decorator(func):
+        @functools.wraps(func)
+        def _auth_func(self, *args, **kwargs):
+            allows_subset = 'subset' in func.__code__.co_varnames
```
minrk, line 308: we can use [inspect.signature](https://docs.python.org/3/library/inspect.html#inspect.signature) here

jupyterhub/utils.py
```diff
@@ -296,6 +299,30 @@ def metrics_authentication(self):
...
+            if scope in self.scopes:
+                return func(self, *args, **kwargs)
+            elif allows_subset:
+                # Check if access is not restricted to user/server/service
+                match_string = re.compile("^" + re.escape(scope) + r"!.+=.+$")
+                subscopes = filter(lambda s: re.search(match_string, s), self.scopes)
+                subset = [subscope.split('=')[1] for subscope in subscopes]
```
minrk, line 315: maybe parse these to a dict for example `!user=name` would give `{"user": "name"}`

minrk: We also need to be careful because the filter applies on a per-scope basis.


Let's add lots of unittests for just this decorator. We will have lots of combinations of multiple filters, etc.

Maybe add a utility for checking later on, maybe parse scopes to objects with methods (or utility functions) something like:


```python
check_scope(have_scopes, needed_scope, user=x, server=y, group=z) -> bool
```

minrk: Maybe another utility for extract limited scope

```
extract_filtered_scope(scopes, needed_scope) -> subset dict
```

jupyterhub/apihandlers/users.py
```diff
@@ -371,12 +382,17 @@ def delete(self, name, token_id):
 class UserServerAPIHandler(APIHandler):
     """Start and stop single-user servers"""
 
-    @admin_or_self
-    async def post(self, name, server_name=''):
+    @needs_scope('users:servers')
+    async def post(self, name, server_name='', subset=None):
```
minrk, line 386: apply resource filtering in decorator, e.g. explicitly:

```python
@needs_scope("users:servers", user_pos=0, server_pos=1)
```


or automatic (at least default) by inspection, i.e. `name` == username, `server_name` == server name, etc.

jupyterhub/utils.py
```diff
@@ -296,6 +299,30 @@ def metrics_authentication(self):
         raise web.HTTPError(403)
 
 
+def needs_scope(scope):
```
minrk, line 302: Lots of direct unittests for this decorator, e.g.


```

class Test:

    @needs_scope("read:user")
    def foo(self, name):
        pass

obj = Test()
obj.scopes = []
with pytest.raises(web.HTTPError):
    obj.foo()
```

in a new `test_scopes.py`

jupyterhub/handlers/base.py
```diff
@@ -426,6 +427,10 @@ async def get_current_user(self):
                 # don't let errors here raise more than once
                 self._jupyterhub_user = None
                 self.log.exception("Error getting current user")
+        if self._jupyterhub_user is not None or self.get_current_user_oauth_token():
+            self.scopes = self.settings.get("mock_scopes", [])
+        else:
+            self.scopes = []
```
minrk, line 433: since this is already in prepare, no need to set here

jupyterhub/handlers/base.py
```diff
@@ -83,6 +83,7 @@ async def prepare(self):
         except Exception:
             self.log.exception("Failed to get current user")
             self._jupyterhub_user = None
+            self.scopes = []
```
minrk, line 86: Let's move this up to an unconditional initial value, so we only override it when we have some scopes to set

jupyterhub/utils.py
```diff
@@ -296,6 +299,133 @@ def metrics_authentication(self):
...
+    """
+    if not (filter_ == 'user' and 'group' in sub_scope):
+        return False
+    if 'user' in sub_scope:
+        return filter_value not in sub_scope['user']
+    else:
+        return True
+
+
+# def expand_groups_to_users(db, filter_scope):
```
minrk, line 322: Looks like this can be removed?

jupyterhub/utils.py
```diff
@@ -296,6 +299,133 @@ def metrics_authentication(self):
...
+            for resource in {'user', 'server', 'group', 'service'}:
+                resource_name = resource + '_name'
+                if resource_name in bound_sig.arguments:
+                    resource_value = bound_sig.arguments[resource_name]
+                    s_kwargs[resource] = resource_value
+            parsed_scopes = parse_scopes(self.scopes)
+            if check_scope(self, scope, parsed_scopes, **s_kwargs):
+                return func(self, *args, **kwargs)
+            else:
+                raise web.HTTPError(403, "Action is not authorized with current scopes")
```
minrk, line 422: Prior to raising, it would be a good idea to add a more detailed `self.log.warning` that includes:

- the path being requested (self.request.path)
- the required scope(s)
- the authorized scope(s)

The error message should also include the required scope (not the current scopes, though, *I think*)

jupyterhub/tests/test_scopes.py
```diff
@@ -0,0 +1,195 @@
...
+    [
+        ('martha', 200),
+        ('michael', 200),
+        ('gob', 200),
+        ('tobias', 403),
+        ('ann', 404),  # this leaks a bit of information, what do we think about that?
+    ],
+)
+async def test_expand_groups(app, name, status_code):
+    app.init_db()
```
minrk, line 179: `app.init_db()` shouldn't be run in individual tests

jupyterhub/tests/test_scopes.py
```diff
@@ -0,0 +1,195 @@
...
+    group = orm.Group(name='bluth')
+    app.db.add(group)
+    for user_name in ['gob', 'michael']:
+        user = orm.User(name=user_name)
```
minrk, line 183: Try using `utils.add_user` to create the users in this (and other) tests. It might be related to the db failures, since this could end up creating multiple users with the same name (I would expect that to fail a uniqueness constraint, but still best to use the utility).

jupyterhub/tests/test_scopes.py
```diff
@@ -0,0 +1,195 @@
...
+        ('martha', 200),
+        ('michael', 200),
+        ('gob', 200),
+        ('tobias', 403),
+        ('ann', 404),  # this leaks a bit of information, what do we think about that?
+    ],
+)
+async def test_expand_groups(app, name, status_code):
+    app.init_db()
+    group = orm.Group(name='bluth')
```
minrk, line 180: Check for the existence of the group first, it may already exist (it definitely will, given that this test is run multiple times)

jupyterhub/apihandlers/hub.py
```diff
@@ -65,7 +66,6 @@ def get(self):
 
 
 class InfoAPIHandler(APIHandler):
-    @admin_only
```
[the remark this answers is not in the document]
0mar: The description said it was not an authenticated endpoint and it does not seem to contain any sensitive information so I removed the scope. What should it be?

jupyterhub/tests/test_scopes.py
```diff
@@ -0,0 +1,195 @@
...
+    scope_list = ['users:servers!server=maeby/server1', 'read:users!user=maeby']
+    parsed_scopes = parse_scopes(scope_list)
+    assert check_scope(
+        handler, 'users:servers', parsed_scopes, user='maeby', server='server1'
+    )
+
+
+class MockAPIHandler:
+    def __init__(self):
+        self.scopes = ['users']
```
minrk, line 83: something like:

```python
self.request = Mock(spec=tornado.httputil.HTTPServerRequest)
```

or build the mock object yourself:

```python
self.request = Mock()
self.request.path = "/thing"
```
[ref](https://www.tornadoweb.org/en/stable/httputil.html#tornado.httputil.HTTPServerRequest)

minrk:
32Great! Now we have the handler-side, we can reconcile with #3215 to connect them to the underlying scopes

PR #3215, opened 2020-10-20
[RBAC] Implementing roles as collections of permission scopes

IvanaH8:
Ref: roles issue #2245 , 
Building on: implementing oauth scopes PR #3169 (ref: scope issue #1057 )  

Roles represent a bundle of permission scopes and are stored in the database (table with id, name, description, scopes, users and services columns).

There are 3 default roles - user, admin and server - defined in roles module to preserve the current state. These can be overwritten and custom roles can be added through config file as a list of role dictionaries:

`c.JupyterHub.load_roles = [ { "name": rolename, "description": roledescription, "scopes": [scopenames], "users": [usernames],                                               "services": [servicenames], "tokens": [tokenstrings] } ]`

All keys apart from "name" are optional. Only existing/predefined users, services and tokens can be added to roles through config file. 

A user/service without any role is automatically assigned a default role based on their admin status. If a user's admin status is changed, their default role updates.

Tokens without specified role are assigned the default 'user' role. Specific roles can be assigned to tokens requested via API in the request body. 

The implementation is included in tests and can be run separately with `role` mark.

Tasks:

- [x] User roles

- [x] Service roles

- [x] Token roles


jupyterhub/roles.py
```diff
@@ -0,0 +1,263 @@
...
+            'read:users:servers',
+        ],
+        'users:tokens': ['read:users:tokens'],
+        'admin:users': None,
+        'admin:users:servers': None,
+        'groups': ['read:groups'],
+        'admin:groups': None,
+        'read:services': None,
+        'proxy': None,
+        'shutdown': None,
```
0mar, line 68: I think we need one more scope for info on the Hub (see `RootAPIHandler` and `InfoAPIHandler` in `jupyterhub/apihandlers/hub.py`. I named it `read:hub` but this negotiable.

jupyterhub/roles.py
```diff
@@ -0,0 +1,263 @@
...
+    scope_list = []
+    if role:
+        scope_list = role.scopes
+    elif roles:
+        for role in roles:
+            scope_list.extend(role.scopes)
+    else:
+        raise ValueError('Function get_subscopes is missing an argument')
+
+    scopes = list(chain.from_iterable(list(map(expand_scope, scope_list))))
```
0mar, line 116: I think the outer `list` could be changed to `set` and that would save a call in the next line

jupyterhub/roles.py
```diff
@@ -0,0 +1,263 @@
...
+        expand_subscopes(index=1)
+        # check for "subscopes of sub-subscopes"
+        expand_subscopes(index=index_for_sssc)
+
+    expanded_scope = set(subscopes)
+
+    return expanded_scope
+
+
+def get_subscopes(role=None, roles=None):
```
0mar, line 103: Perhaps change this function to only take an iterable `roles`. Then any single-role arguments could be packed into a list.
Alternatively, something like
```python
def get_subscopes(*args):
    for role in args:
(...)

subscopes = get_subscopes(*roles)
```

0mar (commented):
Looks very good! I think we can start merging

jupyterhub/roles.py
IvanaH8: You are right, thank you. I agree with the scope name `read:hub`, added to the scopes and admin role in roles.py.

jupyterhub/roles.py
IvanaH8: Yes, thank you for spotting this, changed.

jupyterhub/roles.py
IvanaH8: Nice suggestion, thank you, implemented with *args now.

minrk:
Excellent!

PR #3219, opened 2020-10-21
Fix #2284 must be sent from authorization page

### LINKED ISSUE #2284: Authorization form must be sent from authorization page

elgalu:
Fix #2284 Authorization form must be sent from authorization page

### DIFF
--- a/jupyterhub/apihandlers/auth.py
+++ b/jupyterhub/apihandlers/auth.py
@@ -274,9 +274,12 @@ class OAuthAuthorizeHandler(OAuthHandler, BaseHandler):
         uri, http_method, body, headers = self.extract_oauth_params()
         referer = self.request.headers.get('Referer', 'no referer')
         full_url = self.request.full_url()
-        if referer != full_url:
+        stripped_referer = referer.strip('https:').strip('http:').strip('\n')
+        stripped_full_url = full_url.strip('https:').strip('http:').strip('\n')
+        if stripped_referer != stripped_full_url:
             # OAuth post must be made to the URL it came from
-            self.log.error("OAuth POST from %s != %s", referer, full_url)
+            self.log.error("Original OAuth POST from %s != %s", referer, full_url)
+            self.log.error("Stripped OAuth POST from %s != %s", stripped_referer, stripped_full_url)
             raise web.HTTPError(
                 403, "Authorization form must be sent from authorization page"
             )
--- a/jupyterhub/apihandlers/auth.py
+++ b/jupyterhub/apihandlers/auth.py
@@ -277,19 +277,23 @@ class OAuthAuthorizeHandler(OAuthHandler, BaseHandler):
         # trim protocol, which cannot be trusted with multiple layers of proxies anyway
         # Referer is set by browser, but full_url can be modified by proxy layers to appear as http
         # when it is actually https
-        referer_proto, _, stripped_referer = referer.partition("://")[2]
+        referer_proto, _, stripped_referer = referer.partition("://")
         referer_proto = referer_proto.lower()
-        req_proto, _, stripped_full_url = full_url.partition("://")[2]
+        req_proto, _, stripped_full_url = full_url.partition("://")
         req_proto = req_proto.lower()
         if referer_proto != req_proto:
             self.log.warning("Protocol mismatch: %s != %s", referer, full_url)
             if req_proto == "https":
                 # insecure origin to secure target is not allowed
-                raise web.HTTPError(403, "Not allowing authorization form submitted from insecure page")
+                raise web.HTTPError(
+                    403, "Not allowing authorization form submitted from insecure page"
+                )
         if stripped_referer != stripped_full_url:
             # OAuth post must be made to the URL it came from
             self.log.error("Original OAuth POST from %s != %s", referer, full_url)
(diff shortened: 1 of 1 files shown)


manics:
Is it possible for this to open up a security hole, e.g. by allowing an attacker to force a user to use http instead of https?

elgalu:
Hi, I opened this PR to help with https://github.com/jupyterhub/jupyterhub/issues/2284#issuecomment-713670155 however I'm not a security expert to answer your question.

manics:
I opened an experimental PR at https://github.com/jupyterhub/configurable-http-proxy/pull/267. That needs to be reviewed for security implications, but might also work? Are you able to give it a go? Most (or all) of the current maintainers don't have a platform where we can reproduce the error.

chicocvenancio:
> Is it possible for this to open up a security hole, e.g. by allowing an attacker to force a user to use http instead of https?

I don't think so (?). 

@minrk should take a look at this, but I am not opposed to @elgalu's approach here. 

Tornado is checking origin before it gets to this, and the endpoint is authenticated, I do not see any holes added by stripping the protocol here.
In particular, the deployment should be enforcing https in front of jupyterhub and this really would not be impacted.


#2284 is about JupyterHub seeing the request as http while the browser is using https. In theory we could allow only that combination, but I do think it will make the code significantly less readable and will impact security either way.

elgalu:
@manics even if your fix at configurable-http-proxy works, I think most enterprise users do not use configurable-http-proxy so that alone won't solve this issue (for everyone)

minrk:
This *does* open a security issue, in that what appears to be a cross-origin request (two websites that share a domain, one of which could be compromised, especially since it is by definition insecure) would no longer be blocked.

The problem in #2284 is in failure to *identify* that an HTTPS request is actually HTTPS because of the multiple proxied hops, and some layers not correctly implementing proxy-protocols (I get lost in terms of who is actually doing something wrong!).

I think we can combine the approach here with what is in https://github.com/jupyterhub/configurable-http-proxy/pull/267 and allow the http request *if and only if* https is found in the proxy headers (e.g. x-forwarded-proto, Forwarded).

However, because we do things like set cookies with `secure=true`, if we got this far, then we actually do know that the browser (if it implements cookies correctly and defaults are being used) sent the request over https.

While it's not ideal, getting proxies to agree on what's being forwarded and how to determine the actual URL being requested has turned out to be between hard and impossible. So I think with the other measures we have in place, only considering the host is perhaps enough here.

minrk:
I also think that we can benefit from the fact that when this fails and things are working right, it's because the request is http and the origin is https. (proxies messed up request url, but Referer was left alone and is set by the browser, which is whose perspective we care about)

jupyterhub/apihandlers/auth.py
```diff
@@ -274,9 +274,12 @@ def post(self):
         uri, http_method, body, headers = self.extract_oauth_params()
         referer = self.request.headers.get('Referer', 'no referer')
         full_url = self.request.full_url()
-        if referer != full_url:
+        stripped_referer = referer.strip('https:').strip('http:').strip('\n')
+        stripped_full_url = full_url.strip('https:').strip('http:').strip('\n')
```
minrk, line 278:
```suggestion
        # trim protocol, which cannot be trusted with multiple layers of proxies anyway
        # Referer is set by browser, but full_url can be modified by proxy layers to appear as http
        # when it is actually https
        referer_proto, _, stripped_referer = referer.partition("://")[2]
        referer_proto = referer_proto.lower()
        req_proto, _, stripped_full_url = full_url.partition("://")[2]
        req_proto = req_proto.lower()
        if referer_proto != req_proto:
            self.log.warning("Protocol mismatch: %s != %s", referer, full_url)
            if req_proto == "https":
                # insecure origin to secure target is not allowed
                raise web.HTTPError(403, "Not allowing authorization form submitted from insecure page")
```

minrk (commented):
@elgalu I added a suggestion to restrict the error to http->https requests and allow https->http, which, if I understand correctly, is ~always how these proxy issues appear. Can you verify that this will work for you?

minrk (approved):

minrk:
@elgalu if you can verify that this change still works for you, I'll go ahead and merge it.

elgalu:
Hi, I'll verify this tonight.

elgalu:
Confirmed! this works in our custom environment that uses https://github.com/zalando-incubator/kubernetes-on-aws and https://github.com/zalando/skipper proxy. Thanks!!!

minrk:
@elgalu Thank you for your testing and patience! :)

elgalu:
my pleasure!!

meeseeksmachine:

https://discourse.jupyter.org/t/set-up-multiple-jupyterhub-on-same-domain-with-different-ports-using-nginx/8706/6

PR #3223, opened 2020-10-25
Make api_request to CHP's REST API more reliable

### LINKED ISSUE #3222: Unreliable attempt to add a route for a new user to an external CHP proxy - HTTP 599 Broken pipe (JupyterHub 1.2.0b1 issue)
## Issue background
This issue is migrated from Z2JH where the CI system started having intermittent failures to spawn user pods all the way. The failure wasn't the creation of pods, but when the `hub` pod was speaking with the externally running `proxy` pod (CHP server) and getting `tornado.curlhttpclient.CurlError: HTTP 599: Send failure: Broken pipe` as part of its POST request to add a route for the newly spawned user.

## Logs inspection
The hub errors when speaking with the external CHP proxy pod after the a user pod has successfully been spawned.

```
[I 2020-10-20 16:48:25.540 JupyterHub base:893] User testuser-aa6aa5ea-64bb-4719-93c7-04557f25c3d5 took 2.179 seconds to start
[I 2020-10-20 16:48:25.540 JupyterHub proxy:257] Adding user testuser-aa6aa5ea-64bb-4719-93c7-04557f25c3d5 to proxy /user/testuser-aa6aa5ea-64bb-4719-93c7-04557f25c3d5/ => http://192.168.79.75:8888
[D 2020-10-20 16:48:25.540 JupyterHub proxy:765] Proxy: Fetching POST http://proxy-api:8001/api/routes/user/testuser-aa6aa5ea-64bb-4719-93c7-04557f25c3d5
[I 2020-10-20 16:48:25.541 JupyterHub log:181] 200 GET /hub/api (@10.20.0.120) 1.70ms
[E 2020-10-20 16:48:25.542 JupyterHub base:915] Failed to add testuser-aa6aa5ea-64bb-4719-93c7-04557f25c3d5 to proxy!
      File "/usr/local/lib/python3.8/dist-packages/jupyterhub/handlers/base.py", line 908, in finish_user_spawn
        await self.proxy.add_user(user, server_name)
      File "/usr/local/lib/python3.8/dist-packages/jupyterhub/proxy.py", line 270, in add_user
        await self.add_route(
      File "/usr/local/lib/python3.8/dist-packages/jupyterhub/proxy.py", line 781, in add_route
        await self.api_request(path, method='POST', body=body)
      File "/usr/local/lib/python3.8/dist-packages/jupyterhub/proxy.py", line 773, in api_request
        result = await client.fetch(req)
    tornado.curl_httpclient.CurlError: HTTP 599: Send failure: Broken pipe
```

If everything were as it should, the proxy should add a network route towards the newly added pod - it is a configurable proxy that JupyterHub controls after all. This is the kind of entry we should see twice because we have two tests adding a pod.

```
16:48:18.515 [ConfigProxy] info: Adding route /user/testuser-c2521e10-e843-40c3-8cb6-dac422b1433a -> http://192.168.79.74:8888
```

It seems like this never happens, and that the proxy pod / CHP doesn't notice any incoming request to add a route, because it doesn't log any incoming POST request either.

#### Build failures
[Build failure with logs of hub and proxy pod](https://travis-ci.org/github/jupyterhub/zero-to-jupyterhub-k8s/jobs/737443531) but not for the autohttps pod.
[Build failure with logs of the hub, proxy, and autohttps pod](https://travis-ci.org/github/jupyterhub/zero-to-jupyterhub-k8s/jobs/738037421), this is one of two failures from [the same travis build](https://travis-ci.org/github/jupyterhub/zero-to-jupyterhub-k8s/builds/738037416)!

#### Code that fail to reach return statement
https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/2f7135cfeed631553eb3452f6b3f47f27e47c9f9/tests/test_spawn.py#L217-L224

Example user_model data in code linked above
```python
# example user_model that is seen during failures
user_model = {'kind': 'user', 'name': 'testuser-f8ff861a-538a-4c93-a864-ae3345cea907', 'admin': False, 'groups': [], 'server': None, 'pending': None, 'created': '2020-10-20T14:28:03.524748Z', 'last_activity': '2020-10-20T14:28:11.990951Z', 'servers': {}, 'auth_state': None}
```

consideRatio:
This PR addresses instabilities such as HTTP 599: Broken pipe, and general timeout errors trying to connect with the CHP proxy's REST API as done when for example adding a route for a newly spawned user.

This closes #3222.

### DIFF
--- a/jupyterhub/proxy.py
+++ b/jupyterhub/proxy.py
@@ -774,23 +774,39 @@ class ConfigurableHTTPProxy(Proxy):
         )
 
         async def _wait_for_api_request():
+            most_recent_error = None
             try:
                 async with self.semaphore:
                     return await client.fetch(req)
             except HTTPError as e:
-                if e.code == 599:
-                    # Warn about 599 failures to communicate with the proxy
-                    self.log.warning("api_request to the proxy failed with status code 599, retrying...")
-                    # A falsy return value triggers exponential_backoff to retry
-                    return False
+                most_recent_error = e
+                # Retry on potentially transient errors as for example also
+                # suggested for communication witht he k8s api-server
+                #
+                # ref: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#error-codes
+                if e.code in [
+                    429,  # Too many requests
+                    500,  # Internal server error
+                    503,  # Service unavailable
+                    504,  # Gateway timeout
+                    599,  # Network connect timeout error
+                ]:
+                    self.log.warning(
+                        "api_request to the proxy failed with status code {}, retrying...".format(
+                            e.code
+                        )
+                    )
+                    return False  # a falsy return value make exponential_backoff retry
                 else:
-                    self.log.error("api_request to proxy failed: {0}".format(error))
+                    self.log.error("api_request to proxy failed: {0}".format(e))
                     # An unhandled error here will help the hub invoke cleanup logic
                     raise
 
         result = await exponential_backoff(
             _wait_for_api_request,
-            "api_request to proxy failed over a 30 second duration of attempts for path: {}".format(path),
+            'Repeated api_request to proxy path "{}" failed, most recently with error "{}".'.format(
+                path, most_recent_error
+            ),
             timeout=30,
         )
         return result
--- a/jupyterhub/proxy.py
+++ b/jupyterhub/proxy.py
@@ -39,11 +39,6 @@ from traitlets import observe
 from traitlets import Unicode
 from traitlets.config import LoggingConfigurable
 
-try:
-    from tornado.curl_httpclient import CurlError
-except ImportError:
-    CurlError = NotImplementedError
-
 from . import utils
 from .metrics import CHECK_ROUTES_DURATION_SECONDS
 from .metrics import PROXY_POLL_DURATION_SECONDS
@@ -782,15 +777,21 @@ class ConfigurableHTTPProxy(Proxy):
             try:
                 async with self.semaphore:
                     return await client.fetch(req)
-            except (TimeoutError, CurlError) as error:
-                self.log.debug("api_request to proxy failed: {0}".format(error))
-                # A falsy return value triggers exponential_backoff to retry
-                return False
+            except HTTPError as e:
+                if e.code == 599:
+                    # Warn about 599 failures to communicate with the proxy
+                    self.log.warning("api_request to the proxy failed with status code 599, retrying...")
+                    # A falsy return value triggers exponential_backoff to retry
+                    return False
+                else:
+                    self.log.error("api_request to proxy failed: {0}".format(error))
+                    # An unhandled error here will help the hub invoke cleanup logic
+                    raise
 
         result = await exponential_backoff(
             _wait_for_api_request,
-            "api_request to proxy failed to complete in 20 seconds",
-            timeout=20,
+            "api_request to proxy failed over a 30 second duration of attempts for path: {}".format(path),
+            timeout=30,
         )
         return result
 
--- a/jupyterhub/proxy.py
+++ b/jupyterhub/proxy.py
@@ -39,10 +39,16 @@ from traitlets import observe
 from traitlets import Unicode
 from traitlets.config import LoggingConfigurable
 
+try:
+    from tornado.curl_httpclient import CurlError
+except ImportError:
+    CurlError = NotImplementedError
+
 from . import utils
 from .metrics import CHECK_ROUTES_DURATION_SECONDS
 from .metrics import PROXY_POLL_DURATION_SECONDS
 from .objects import Server
+from .utils import exponential_backoff
 from .utils import make_ssl_context
 from .utils import url_path_join
 from jupyterhub.traitlets import Command
@@ -768,10 +774,25 @@ class ConfigurableHTTPProxy(Proxy):
             method=method,
             headers={'Authorization': 'token {}'.format(self.auth_token)},
             body=body,
+            connect_timeout=1,  # default: 20s
+            request_timeout=5,  # default: 20s
         )
-        async with self.semaphore:
-            result = await client.fetch(req)
-            return result
+
+        async def _wait_for_api_request():
+            try:
+                async with self.semaphore:
+                    return await client.fetch(req)
+            except (TimeoutError, CurlError) as error:
+                self.log.debug("api_request to proxy failed: {0}".format(error))
+                # A falsy return value triggers exponential_backoff to retry
+                return False
+
+        result = await exponential_backoff(
+            _wait_for_api_request,
+            "api_request to proxy failed to complete in 20 seconds",
+            timeout=20,
+        )
+        return result
 
     async def add_route(self, routespec, target, data):
         body = data or {}
--- a/jupyterhub/proxy.py
+++ b/jupyterhub/proxy.py
@@ -773,8 +773,8 @@ class ConfigurableHTTPProxy(Proxy):
             request_timeout=5,  # default: 20s
         )
 
+        most_recent_error = None
         async def _wait_for_api_request():
-            most_recent_error = None
             try:
                 async with self.semaphore:
                     return await client.fetch(req)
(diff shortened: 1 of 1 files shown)


jupyterhub/proxy.py
```diff
@@ -39,10 +39,16 @@
 from traitlets import Unicode
 from traitlets.config import LoggingConfigurable
 
+try:
+    from tornado.curl_httpclient import CurlError
+except ImportError:
+    CurlError = NotImplementedError
+
```
consideRatio, line 46: This will only work if pycurl is installed, and is only a relevant error to catch if it is installed. By letting CurlError become a reference to NotImplementedError I can just do `except (TimeoutError, CurlError) as error:` later without dynamically creating a tuple etc.

betatim: Why do you want to attempt to catch a `CurlError` instead of `HTTPClientError` which is used by all HTTP clients?

consideRatio: I want to catch specifically Broken Pipe errors which are low level network error (CurlError), while I don't want to catch for example 401 Forbidden which is just a plain failure to authenticate and doesn't make sense to try again.

Does HTTPClientError include 401 errors for example? I wonder exactly what errors we should aim to retry for, it is not obvious in my mind. Some 4XX errors like 429 about "too many requests atm, retry again" would make sense to retry, but I'm not sure if CHP emits those ever.

I'm open to suggestions on what to catch, but I think its sensible to aim for the errors that represents network connection related errors to begin with, how do we single them out?

betatim: I think the recommendation is to use the client independent classes when catching exceptions. Somewhere Min even commented on this but I can't find the issue/PR :-/

What I'd do is catch `HTTPClientError` and then check the status code. 599 is a special status code used by Tornado to signal "something went wrong and it isn't a normal HTTP status code". For example network request timeouts, broken pipes, etc. That way we can decide for which status codes or range of status codes we do want to retry.

consideRatio: Excellent push in the right direction I think @betatim! I'll do this!

599 is after all...
image

consideRatio:
```python
class CurlError(HTTPError):
    def __init__(self, errno: int, message: str) -> None:
        HTTPError.__init__(self, 599, message)
        self.errno = errno
```
- [Source code for CurlError in tornado here](https://github.com/tornadoweb/tornado/blob/bee08e25b1b990edd7ebfb7cd463ec18a98a3fde/tornado/curl_httpclient.py#L575-L578)

```python
class HTTPError(Exception):
    """An exception that will turn into an HTTP error response.
    ...
    """
```
- [Source code for HTTPError in tornado here](https://github.com/tornadoweb/tornado/blob/bee08e25b1b990edd7ebfb7cd463ec18a98a3fde/tornado/web.py#L2333-L2338)

---

It seems like HTTPClientError is the same as HTTPError.

image

---

Conclusion, I should catch HTTPError / HTTPClientError, but raise it again if it doesn't seem appropriate to handle, and in this case I'm currently only confident about 599 errors.

consideRatio: I found this example from the existing code base.

```python
        except HTTPError as e:
            if e.code == 404:
                # Warn about 404s because something might be wrong
                # but don't raise because the route is gone,
                # which is the goal.
                self.log.warning("Route %s already deleted", routespec)
            else:
                raise
```

Thanks @betatim!

consideRatio: Resolved in b91c5a4

consideRatio: I also made 429,500,503,504 retry as well, which is in line with recommendations for the k8s api-server, so it probably can make some general sense I think.

See this related discussion: https://github.com/jupyterhub/kubespawner/issues/436

minrk (commented):
Sorry, wrote a review yesterday but forgot to submit it.

Main comments:

- retrying 429 without handling a requested delay from retry-after header can make things *worse* by sending more requests when have just been told they will fail (though CHP doesn't do 429, so handling 429 is moot for now).
- I think retrying anything >= 500 makes sense since that means it's not our fault. We should log every failure at at least warning-level, though.

Looks great in general!

jupyterhub/proxy.py
```diff
@@ -768,10 +769,47 @@ async def api_request(self, path, method='GET', body=None, client=None):
...
+                    return False  # a falsy return value make exponential_backoff retry
+                else:
+                    self.log.error("api_request to proxy failed: {0}".format(e))
+                    # An unhandled error here will help the hub invoke cleanup logic
+                    raise
+
+        result = await exponential_backoff(
+            _wait_for_api_request,
+            'Repeated api_request to proxy path "{}" failed, most recently with error "{}".'.format(
+                path, most_recent_error
```
minrk, line 808: This string-formatting is evaluated immediately, so the attempt to use most_recent_error here isn't going to work since it will always be None.

I don't think we need it, though, because the most recent error is already logged when it happens, before reaching this failure message.

jupyterhub/proxy.py
```diff
@@ -768,10 +769,47 @@ async def api_request(self, path, method='GET', body=None, client=None):
             method=method,
             headers={'Authorization': 'token {}'.format(self.auth_token)},
             body=body,
+            connect_timeout=1,  # default: 20s
+            request_timeout=5,  # default: 20s
```
consideRatio, line 773:
```suggestion
            connect_timeout=3,  # default: 20s
            request_timeout=10,  # default: 20s
```

betatim:
Looks good to me. Min had a comment about handling 429 only if we also implement the requested "wait before you retry" behaviour. Should we remove 429 from the list of retried status codes for now as CHP does not (ever) return it right now? Feels like if we leave it in and CHP starts serving 429s we will not remember that we already added it here but didn't implement all of the handling.

jupyterhub/proxy.py
consideRatio: Deal - removed it!

consideRatio:
@minrk @betatim I simplified the logic to be >=500 and made a note about CHP not implementing 429.

minrk (approved):

minrk:
> I have assumed the risk of getting blocked by waiting to get the semaphore quite low, but currently, it would be possible for a REST API request to the proxy to never be attempt because exponential_backoff got stuck waiting for the semaphore the entire 30 seconds. Is this acceptable?

Yes. This is already the case, and related to the concurrent-spawn limit, so unchanged wrt this PR (there is already an outer timeout). Concurrency parameters need to be tuned for a particular deployment, and both the concurrent-spawn limit and CHP semaphore can be modified, but I think the defaults are reasonable.

In practice, CHP (when it's running) responds super quickly and is ~never the bottleneck. The semaphore is only likely to get clogged up if CHP is unavailable and everything is stuck retrying and bound to fail.

consideRatio:
Wieee! I'm really excited to try this out, I hope this can resolve many long standing issues, such as the hub being forced to restart because it makes one call to get all routes from the proxy which typically fails after 20 seconds in case the proxy hasn't started when this call is sent out.

consideRatio:
Wieee it works wonderful now! It also closes #2745 thoroughly!

```
[I 2020-10-29 00:19:42.599 JupyterHub app:2398] Initialized 0 spawners in 0.007 seconds
[I 2020-10-29 00:19:42.604 JupyterHub app:2610] Not starting proxy
[D 2020-10-29 00:19:42.606 JupyterHub proxy:766] Proxy: Fetching GET http://proxy-api:8001/api/routes
[W 2020-10-29 00:19:45.607 JupyterHub proxy:785] api_request to the proxy failed with status code 599, retrying...
[W 2020-10-29 00:19:48.642 JupyterHub proxy:785] api_request to the proxy failed with status code 599, retrying...
[W 2020-10-29 00:19:51.792 JupyterHub proxy:785] api_request to the proxy failed with status code 599, retrying...
[D 2020-10-29 00:19:52.459 JupyterHub proxy:839] Omitting non-jupyterhub route '/'
[I 2020-10-29 00:19:52.461 JupyterHub app:2646] Hub API listening on http://:8081/hub/
[I 2020-10-29 00:19:52.461 JupyterHub app:2648] Private Hub API connect url http://hub:8081/hub/
```

PR #3225, opened 2020-10-27
Allow options_from_form to be configurable

cbanek:
From the documentation, it is easy to override the options_form to set either a string or a callable to provide a custom options form.  But there doesn't seem to be the same way of configuring the options_from_form, which receives the form data from the options form.

This is a bit tricky, since I want to be able to provide a function to make an options_form that has a few different form elements, and I need to be able to manipulate the output.  But I can't configure the options_from_form callback the same way.  Right now, this all works on subclassing, but you have to derive from the spawner, rather than using traitlets.

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -354,8 +354,9 @@ class Spawner(LoggingConfigurable):
 
         return options_form
 
-    def options_from_form(self, form_data):
-        """Interpret HTTP form data
+    options_from_form = Callable(
+        help="""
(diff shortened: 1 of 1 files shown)


cbanek:
Apologies if this is way off base or not done correctly, this is my first traitlets sort of PR, and I'm not sure if this is a breaking change (I've tried to make it not be).

Related conversation: https://github.com/jupyterhub/kubespawner/pull/452

minrk:
Nice! This should work just fine.

PR #3229, opened 2020-10-28
make pagination configurable

### LINKED ISSUE #3228: Allow configuring max number of users per page in admin
### Proposed change

Right now, it's impossible to have more than 250 users per-page in the JupyterHub admin page. However, this makes searching for a single user basically impossible - you have to manually paginate through them.

Search functionality is the right way to do this, but until then, I think we should allow using a traitlet to configure how many users can be displayed on a single page.

### Alternative options

Add a user search bar! But that's more work.

Anyone with more than 250 users on their hub.

Add a traitlet to the `JupyterHub` object that gets passed down to the Pagination class, where it can use those numbers instead of the hardcoded limits there.

minrk:
add some unittests for pagination

reorganize pagination a bit to make it easier to configure

still need to do a bit more testing

closes #3228

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -77,6 +77,7 @@ from .user import UserDict
 from .oauth.provider import make_provider
 from ._data import DATA_FILES_PATH
 from .log import CoroutineLogFormatter, log_request
+from .pagination import Pagination
 from .proxy import Proxy, ConfigurableHTTPProxy
 from .traitlets import URLPrefix, Command, EntryPointType, Callable
 from .utils import (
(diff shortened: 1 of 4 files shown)


yuvipanda:
Awesome, @minrk!!!!

consideRatio:
@minrk I added the enhancement label because it make github-activity automatically categorize the PR for the changelog.

yuvipanda:
Hmm, this seems to not work. 

I've the following in my extra_config:

```yaml
      07-no-pagination: |
        c.Pagination.default_per_page = 10000
        c.Pagination.max_per_page = 10000
```

But pagination still seems to be at 250.

minrk:
Found and fixed it in #3294

PR #3234, opened 2020-10-30
Make external JupyterHub services' oauth_no_confirm configuration work as intentend

bitnik:
fixup for regression introduced in https://github.com/jupyterhub/jupyterhub/pull/3090

This was the typo causing the regression, the equivalent variable for oauth_no_confirm_whitelist was oauth_no_confirm_list rather than oauth_no_confirm which is a boolean.

```diff
-            # or it's in the global whitelist
-            or oauth_client.identifier
-            in self.settings.get('oauth_no_confirm_whitelist', set())
+            # or it's in the global no-confirm list
+            or oauth_client.identifier in self.settings.get('oauth_no_confirm', set())
```

### DIFF
--- a/jupyterhub/apihandlers/auth.py
+++ b/jupyterhub/apihandlers/auth.py
@@ -215,7 +215,8 @@ class OAuthAuthorizeHandler(OAuthHandler, BaseHandler):
             # it's the user's own server
             oauth_client.identifier in own_oauth_client_ids
             # or it's in the global no-confirm list
-            or oauth_client.identifier in self.settings.get('oauth_no_confirm', set())
+            or oauth_client.identifier
+            in self.settings.get('oauth_no_confirm_list', set())
         ):
             return False
         # default: require confirmation


jupyterhub/apihandlers/auth.py
```diff
@@ -215,7 +215,8 @@ def needs_oauth_confirm(self, user, oauth_client):
             # it's the user's own server
             oauth_client.identifier in own_oauth_client_ids
             # or it's in the global no-confirm list
-            or oauth_client.identifier in self.settings.get('oauth_no_confirm', set())
+            or oauth_client.identifier
+            in self.settings.get('oauth_no_confirm_list', set())
```
consideRatio, line 219: This code segment makes it clear to me that your suggested change make sense!

```python
        oauth_no_confirm_list = set()
        for service in self._service_map.values():
            if service.oauth_no_confirm:
                self.log.warning(
                    "Allowing service %s to complete OAuth without confirmation on an authorization web page",
                    service.name,
                )
                oauth_no_confirm_list.add(service.oauth_client_id)
```

consideRatio:
@bitnik what do you mean fix #3090? It is a closed PR. Was it a regression in #3090 that this fixes?

It will probably fix the issue observed when bumping to JupyterHub 1.2.0 observed by BinderHub, which is registered with JupyterHub as an external service. See https://github.com/jupyterhub/binderhub/pull/1184#issuecomment-719120887.

consideRatio (approved):
This fixes a problem that will influence BinderHub deployments I think.

This was caught this in the BinderHub CI to test authentication by noticing that JupyterHub redirected the user for a confirmation. I figure this should be quite easy to test in JupyterHub as well.

jupyterhub/apihandlers/auth.py
```diff
@@ -215,7 +215,8 @@ def needs_oauth_confirm(self, user, oauth_client):
             # it's the user's own server
             oauth_client.identifier in own_oauth_client_ids
             # or it's in the global no-confirm list
```
consideRatio, line 217:
```suggestion
            # DUMMY TESTING GITHUB STUFF
```

bitnik:
>  Was it a regression in #3090 that this fixes?

yes, that's what I mean

PR #3242, opened 2020-11-04
Assume py36 and remove @gen.coroutine etc.

### LINKED ISSUE #3241: Remove py35 support for async functions and async generators
- Python 3.4 has reached end of life, and we can now cleanup the code base from @gen.coroutine / yeild with async / await instead.
- Python 3.5 has reached end of life, so we can cleanup the code base from @async_generator / yeild_ with async / yield instead.

By doing this, we make the code base easier to understand and contribute to.

consideRatio:
This PR is meant to close #3241 by removing patches for py34 / py35 which are now both at their end of life. This PR removes py35 support. We will still need the [async_generator](https://pypi.org/project/async_generator/) until python 3.7 for the @asynccontextmanager decorator that is useful to wrap async generator functions that is to be used in a `with` statement.

## Things done
- [x] Replace @gen.coroutine/yield with async/await
- [x] Replace gen.sleep with asyncio.sleep
- [x] (REVERTED) Replace gen.with_timeout with async_timeout.timeout & asyncio.shield
  Initially I tried what was described in https://github.com/dask/distributed/pull/3372, but as it failed I tried with advice in https://hynek.me/articles/waiting-in-asyncio/ and in https://pypi.org/project/async-timeout/ which made it work.
- [x] Replace gen.multi(futures) with asyncio.gather(*futures)
- [x] Replace @async_generator/yeild_ with async/yeild

## Review checks
- [x] (REVERTED) In the commit about replacing gen.with_timeout with async_timeout.timeout & asyncio.shield, I wrap the future in a "shield" that will absorb a CancelledError that comes from async_timeout.timeout that doesn't come from gen.with_timeout. Does this make sense? It is my understanding that we may end up with stranded execution threads that may continue forever if we are not careful. I do think this is something we did in the past though, so I believe what I've done only mirrored current behavior.

## Related
- I learned a lot from this blog post: https://hynek.me/articles/waiting-in-asyncio/
- I made us depend on the async-timeout package: https://pypi.org/project/async-timeout/
- I learned about asyncio.shield from here: https://stackoverflow.com/a/52511210
- I may have learned about the distinction between a future and a coroutine, this is my current understanding that I'd love to get corrected if it seems off.
   ```python
   async def my_async_func():
       asyncio.sleep(1)
       return "slow response"
   
   # Example 1 - what is a coroutine?
   my_coroutine = my_async_func()  # the function isn't scheduled for execution here
   my_return_value = await my_coroutine  # this is where it is scheduled!
   
   # My understanding:
   # a coroutine is an async function + arguments for it not yet running
   
   # Example 2 - what is a future?
   import asyncio
   my_future = asyncio.ensure_future(my_async_func())  # ensure_future took a coroutine and started it!
   asyncio.sleep(1)
   my_return_value = await my_future  # this line completes instantly!
   
   # My understanding:
   # a future is a placeholder for the return value of a coroutine that has started running
   ```

### DIFF
--- a/docs/source/reference/authenticators.md
+++ b/docs/source/reference/authenticators.md
@@ -235,10 +235,9 @@ to Spawner environment:
 
 ```python
 class MyAuthenticator(Authenticator):
-    @gen.coroutine
-    def authenticate(self, handler, data=None):
-        username = yield identify_user(handler, data)
-        upstream_token = yield token_for_user(username)
+    async def authenticate(self, handler, data=None):
+        username = await identify_user(handler, data)
+        upstream_token = await token_for_user(username)
         return {
             'name': username,
             'auth_state': {
@@ -246,10 +245,9 @@ class MyAuthenticator(Authenticator):
             },
         }
 
-    @gen.coroutine
-    def pre_spawn_start(self, user, spawner):
+    async def pre_spawn_start(self, user, spawner):
         """Pass upstream_token to spawner via environment variable"""
-        auth_state = yield user.get_auth_state()
+        auth_state = await user.get_auth_state()
         if not auth_state:
             # auth_state not enabled
             return
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -930,7 +930,7 @@ class JupyterHub(Application):
 
         with an :meth:`authenticate` method that:
 
-        - is a coroutine (asyncio)
+        - is a coroutine (asyncio or tornado)
         - returns username on success, None on failure
         - takes two arguments: (handler, data),
           where `handler` is the calling web.RequestHandler,
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -930,7 +930,7 @@ class JupyterHub(Application):
 
         with an :meth:`authenticate` method that:
 
-        - is a coroutine (asyncio or tornado)
+        - is a coroutine (asyncio)
         - returns username on success, None on failure
         - takes two arguments: (handler, data),
           where `handler` is the calling web.RequestHandler,
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -2120,7 +2120,7 @@ class JupyterHub(Application):
             self.log.debug(
                 "Awaiting checks for %i possibly-running spawners", len(check_futures)
             )
-            await gen.multi(check_futures)
+            await asyncio.gather(*check_futures)
         db.commit()
 
         # only perform this query if we are going to log it
--- a/jupyterhub/tests/conftest.py
+++ b/jupyterhub/tests/conftest.py
@@ -54,6 +54,13 @@ from .utils import ssl_setup
 _db = None
 
 
+def pytest_collection_modifyitems(items):
+    """add asyncio marker to all async tests"""
+    for item in items:
+        if inspect.iscoroutinefunction(item.obj):
+            item.add_marker('asyncio')
+
+
 @fixture(scope='module')
 def ssl_tmpdir(tmpdir_factory):
     return tmpdir_factory.mktemp('ssl')
--- a/jupyterhub/tests/conftest.py
+++ b/jupyterhub/tests/conftest.py
@@ -36,7 +36,6 @@ from unittest import mock
 
 from pytest import fixture
 from pytest import raises
-from tornado import gen
 from tornado import ioloop
 from tornado.httpclient import HTTPError
 from tornado.platform.asyncio import AsyncIOMainLoop
@@ -53,16 +52,6 @@ from .utils import ssl_setup
 
 # global db session object
 _db = None
-
-
-def pytest_collection_modifyitems(items):
-    """add asyncio marker to all async tests"""
-    for item in items:
-        if inspect.iscoroutinefunction(item.obj):
-            item.add_marker('asyncio')
-        if hasattr(inspect, 'isasyncgenfunction'):
-            # double-check that we aren't mixing yield and async def
-            assert not inspect.isasyncgenfunction(item.obj)
 
 
 @fixture(scope='module')
@@ -244,17 +233,14 @@ def _mockservice(request, app, url=False):
         assert name in app._service_map
         service = app._service_map[name]
 
-        @gen.coroutine
-        def start():
+        async def start():
             # wait for proxy to be updated before starting the service
-            yield app.proxy.add_all_services(app._service_map)
+            await app.proxy.add_all_services(app._service_map)
             service.start()
 
         io_loop.run_sync(start)
 
         def cleanup():
-            import asyncio
-
             asyncio.get_event_loop().run_until_complete(service.stop())
             app.services[:] = []
             app._service_map.clear()
--- a/jupyterhub/tests/conftest.py
+++ b/jupyterhub/tests/conftest.py
@@ -55,10 +55,16 @@ _db = None
 
 
 def pytest_collection_modifyitems(items):
-    """add asyncio marker to all async tests"""
+    """This function is automatically run by pytest passing all collected test
+    functions.
+
+    We use it to add asyncio marker to all async tests and assert we don't use
+    test functions that are async generators which wouldn't make sense.
+    """
     for item in items:
         if inspect.iscoroutinefunction(item.obj):
             item.add_marker('asyncio')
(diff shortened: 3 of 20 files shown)


GeorgianaElena (approved):
I went through the changes and I think they look really good.
I really like how you documented every decision and I think they make a lot of sense (mirroring the behavior we had before for timeouts by wrapping them in a `shield` and keeping the `async_generator` dependency to take advantage of  `@asynccontextmanager` decorator and the [`aclosing`](https://github.com/python-trio/async_generator/blob/22eddc191c2ae3fc152ca13cf2d6fa55ac3f1568/async_generator/_util.py#L6) class).
I learned a lot from reviewing this PR! Thanks @consideRatio :tada: :heart:

minrk:
Awesome! Is there a benefit to switching from `@gen.with_timeout` to adding a new dependency for functionality we already have? Unlike dask, we are not going to remove tornado as a dependency, and tornado is fully native on top of asyncio. So if we need a new dependency to preserve existing behavior, why not keep what we have?

jupyterhub/app.py
```diff
@@ -930,7 +930,7 @@ def _deprecate_api_tokens(self, change):
 
         with an :meth:`authenticate` method that:
 
-        - is a coroutine (asyncio or tornado)
+        - is a coroutine (asyncio)
```
minrk, line 933: This is a major backward-incompatible change to no longer allow Spawners and Authenticators to use tornado `@gen.coroutine` or return any Future.

The biggest difference between tornado's implementation and asyncio's is that tornado supports strictly more cases. With tornado 6, at least, a tornado coroutine and an asyncio coroutine are the same thing, so this makes no difference, but implicit here is probably a new minimum tornado version. We should test this.

docs/source/reference/authenticators.md
```diff
@@ -235,21 +235,19 @@ to Spawner environment:
 
 ```python
 class MyAuthenticator(Authenticator):
-    @gen.coroutine
```
minrk, line 238: Let's make sure we still have tests with `@gen.coroutine` to see if this is a breakage forcing chagnes on all Spawners and Authenticators. If it is, I'd argue it is not worth it to drop support.

jupyterhub/tests/conftest.py
```diff
@@ -60,9 +59,6 @@ def pytest_collection_modifyitems(items):
     for item in items:
         if inspect.iscoroutinefunction(item.obj):
             item.add_marker('asyncio')
-        if hasattr(inspect, 'isasyncgenfunction'):
-            # double-check that we aren't mixing yield and async def
-            assert not inspect.isasyncgenfunction(item.obj)
```
minrk, line 65: I think this assert wants to stay, no? Just the condition is always True?

minrk (commented):
Nice! 👍 to adopting `async def` everywhere in our code and native async generators. The main things I think we need to dig into are:

1. what are the backward-compatibility consequences for existing Authenticators or Spawners that may still be using `@gen.coroutine`? Checking with latest tornado, there appear to be no issues, but this may mean we are requiring tornado 6. This should be checked by making sure we still have tests where Authenticators and Spawners use `@gen.coroutine`, and also make sure we have a test matrix entry with the oldest-supported tornado.
2. tornado's `gen.with_timeout` is *not* deprecated in any way, so replacing it with a new dependency doesn't seem to come with a benefit or improvement. What do you think?

minrk:
Re: CancelledError, I think it's context-dependent. It's possible we have some cases where we *should* be cancelling and there are definitely others (e.g. slow_spawn_timeout) where we definitely should *not* be cancelling. Preserving existing behavior is definitely the right place to start, which is never cancel.

jupyterhub/tests/conftest.py
consideRatio: I added it back but removed the `hasattr(inspect, 'isasyncgenfunction')` part as it is added in py36 and we no longer support py36.

I'd like to enhance the comment to make me understand it better. I now better understand this as a test to ensure our test functions run by pytest, all prefixed with `test_`, not are async generator functions. I guess it may not make sense for them to be that, so this assertion is unrelated to the support for py36 support for async generator functions in general, right?

docs/source/reference/authenticators.md
consideRatio: Could you help me with this step? I figure we need to add variations a set of tests to use a mocked spawner and mocked authenticator with @gen.coroutine in various functions.

jupyterhub/app.py
consideRatio: Thanks for helping me understand this better, I have added back the comment for now.

consideRatio:
> Awesome! Is there a benefit to switching from @gen.with_timeout to adding a new dependency for functionality we already have? Unlike dask, we are not going to remove tornado as a dependency, and tornado is fully native on top of asyncio. So if we need a new dependency to preserve existing behavior, why not keep what we have?

- [x] I did it because I saw it be done in a PR to migrate to asyncio. I removed the commit with this change!

> Re: CancelledError, I think it's context-dependent. It's possible we have some cases where we should be cancelling and there are definitely others (e.g. slow_spawn_timeout) where we definitely should not be cancelling. Preserving existing behavior is definitely the right place to start, which is never cancel.

- [x] The CancelledError was a question related to the transition away from gen.with_timeout which was reverted.

> What are the backward-compatibility consequences for existing Authenticators or Spawners that may still be using @gen.coroutine? Checking with latest tornado, there appear to be no issues, but this may mean we are requiring tornado 6. This should be checked by making sure we still have tests where Authenticators and Spawners use @gen.coroutine, and also make sure we have a test matrix entry with the oldest-supported tornado.

- [ ] Add tests for gen.coroutine usage in Spawner, Authenticator, and perhaps Proxy classes that can be plugged in.
- [ ] Run tests with tornado 5 and not only tornado 6 since we still support tornado 5

I think I'll need help with these parts to land this PR or parts of it in some way. I tried adding back a @gen.coroutine to a `start` call of the `_mockservice`, and I got the following error for example.

image

consideRatio:
Test failure is a flake and I have diagnosed the reason and reported it in https://github.com/jupyterhub/jupyterhub/issues/3251

docs/source/reference/authenticators.md
minrk: Can do! This is part of why I didn't update these, to make sure they are covered in the tests without duplicating everything. But we can duplicate them and add a small number of dedicated tests.

jupyterhub/tests/conftest.py
minrk: Correct! The condition is always True for py36 so to leave the behavior unchanges would be to *always* execute the block without a condition instead of never executing it (which would be the py35 behavior).

The assertion was added specifically to catch folks used to writing tornado coroutines which use `yield` switching to `async def` functions. In particular, during the bulk migration in #2358. The reason I added it was that I saw that several tests I had not updated were *passing* without being run.

For instance, this test:

```python
async def test_fail():
    yield gen.sleep(1)
    assert False
```

would *pass*.

**However**, when I run with the latest versions of pytest, pytest-asyncio, this pattern does now fail:

```
E           TypeError: An asyncio.Future, a coroutine or an awaitable is required
```

so we can safely remove the check, as recent-enough pytest-asyncio handles this check itself.

minrk:
Added a matrix entry to turn all the `>=` requirements into `==` so we can run a test with oldest supported Python + dependencies. This revealed that we've been requiring async_generator 1.9 and tornado 5.1, not the declared 1.8 and 5.0. But after bumping those pins, everything seems to pass.

minrk:
Instead of going through the full process of separate mocks for the `@gen.coroutine` style on Spawner and friends, I went for a much simpler unittest verifying that `@gen.coroutine` and `run_on_executor` produce 'awaitables', which is really our requirement for public async methods, not that they strictly be coroutines. This isn't as rigorous as real integration tests, but I think it catches our primary assumption, which is that our use of `await` will still work with methods defined with tornado's async decorators.

consideRatio:
@minrk I'm I don't overview this well enough to make a call on merge and not sure how to acquire such overview. If you would like help of any kind let me know.

minrk:
This is great, thanks!

consideRatio:
Thank you for your help getting this PR all the way @minrk! :heart:

PR #3243, opened 2020-11-04
[Metrics] Add prefix to prometheus metrics to group all jupyterhub metrics 

agp8x:
To integrate nicely into an existing Prometheus based monitoring system, having a clear prefix for all metrics exported by a service improves maintainability a lot.
This changes only the name of the exported metrics, no internal names or references.

Of course, this breaks existing monitoring installations, which need to adapt the metric names.

### DIFF
--- a/jupyterhub/metrics.py
+++ b/jupyterhub/metrics.py
@@ -5,7 +5,7 @@ Read https://prometheus.io/docs/practices/naming/ for naming
 conventions for metrics & labels. We generally prefer naming them
 `jupyterhub_<noun>_<verb>_<type_suffix>`. So a histogram that's tracking
 the duration (in seconds) of servers spawning would be called
-jupyterhub_SERVER_SPAWN_DURATION_SECONDS.
+jupyterhub_server_spawn_duration_seconds.
 
 We also create an Enum for each 'status' type label in every metric
 we collect. This is to make sure that the metrics exist regardless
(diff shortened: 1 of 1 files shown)


jupyterhub/metrics.py
```diff
@@ -21,13 +21,13 @@
 from prometheus_client import Histogram
 
 REQUEST_DURATION_SECONDS = Histogram(
-    'request_duration_seconds',
+    'jupyter_request_duration_seconds',
```
yuvipanda, line 24: jupyterhub_ as a prefix instead of jupyter_ maybe? Jupyter server also will have metrics

yuvipanda:
I think this is actually a great thing to do! Will break a little bit of backwards compatibility, but that should be easily fixable in people's metrics visualizations.

jupyterhub/metrics.py
```diff
@@ -3,9 +3,9 @@
 
 Read https://prometheus.io/docs/practices/naming/ for naming
 conventions for metrics & labels. We generally prefer naming them
-`<noun>_<verb>_<type_suffix>`. So a histogram that's tracking
+`jupyterhub_<noun>_<verb>_<type_suffix>`. So a histogram that's tracking
 the duration (in seconds) of servers spawning would be called
-SERVER_SPAWN_DURATION_SECONDS.
+jupyterhub_SERVER_SPAWN_DURATION_SECONDS.
```
minrk, line 8: the all caps names are the object names used internally, but I think this is about the metric names, so:

```suggestion
jupyterhub_server_spawn_duration_seconds.
```

minrk: Can we also add in the module docstring:

```rst
.. versionchanged:: 1.3

    added ``jupyterhub_`` prefix to all metric names
```

minrk (commented):
I think this makes sense!

PR #3246, opened 2020-11-08
Migrate from travis to GitHub actions

consideRatio:
Migration of Travis CI tests into a GitHub workflow, motivated by https://github.com/jupyterhub/team-compass/issues/353.

## Noteworthy
I have put in effort to document things that were relevant for me to learn about GitHub actions within this workflow so it can help others not familiar with them as a reference.

## Example how it will look
image

## Review points
- [ ] I have removed a test that only worked on py36 when subdomain tests were used. See 1ae039d for the change, [the error in full](https://github.com/consideRatio/jupyterhub/runs/1372389563?check_suite_focus=true), official python docs [about a function that gives the error](https://docs.python.org/3.7/library/socket.html#socket.getaddrinfo), and a related [stackoverflow post](https://stackoverflow.com/questions/15246088/what-does-this-socket-gaierror-mean).


.github/workflows/test.yml
```diff
@@ -0,0 +1,146 @@
+# This is a GitHub workflow defining a set of jobs with a set of steps.
+# ref: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions
+#
+name: Test
+
+on:
+  pull_request:
+    branches: ["**"]
+  push:
+    branches: [main, master]
```
manics, line 10: Is this restriction needed? Are there branches that are created but don't need CI testing?

.github/workflows/test.yml
```diff
@@ -0,0 +1,146 @@
+# This is a GitHub workflow defining a set of jobs with a set of steps.
+# ref: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions
+#
+name: Test
+
+on:
+  pull_request:
+    branches: ["**"]
```
manics, line 8: Isn't this the default?

.github/workflows/test.yml
```diff
@@ -0,0 +1,146 @@
...
+      # NOTE: UTF-8 content may be interpreted as ascii and causes errors
+      #       without this.
+      LANG: C.UTF-8
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - python: "3.6"
+            pre_commit: true
```
manics, line 32: I think it's clearer to move pre-commit to a separate lint job instead of including it in the matrix. The required steps are fairly trivial so I don't think the added complexity of the if conditionals is worth it, and logically it's different to the oher conditionals which affect the test scenario.
```
    steps:
    - uses: actions/checkout@v2
    - uses: actions/setup-python@v2
    - uses: pre-commit/action@v2.0.0
```
In addition it's easier for developers to see the lint error show up immediately in it's own job- it'll often show up in a few seconds since there's not much setup involved.

.github/workflows/test.yml
consideRatio: If someone forks this repo, it will create jobs in the forks when pushing to the PR branches which isn't needed. I think it's quite nice to clearly signal the intent.

.github/workflows/test.yml
consideRatio: Mainly to achieve clear intent signalling.

consideRatio: I'll remove it and add a comment on the `on` to help clarify intent

.github/workflows/test.yml
consideRatio: I retained the restriction and added a comment, let me know if you think we should remove it. I think it makes sense though as it conserves compute resources and avoids potential clutter of notifications of failed CI jobs that can show up twice etc otherwise.

.github/workflows/test.yml
consideRatio: Excellent point, added as a separate step, there was very little duplication as we didn't need to install anything and what you listed above works to run pre-commit!

.github/workflows/test.yml
minrk: Personally, I really *like* being able to run tests on my forks before opening PRs. I'm not asking for a change, just noting that I personally benefit from lacking this config rather often.

.github/workflows/test.yml
```diff
@@ -0,0 +1,187 @@
...
+      - name: Help message if pre-commit fail
+        if: ${{ failure() }}
+        run: |
+          echo "You can install pre-commit hooks to automatically run formatting"
+          echo "on each commit with:"
+          echo "    pre-commit install"
+          echo "or you can run by hand on staged files with"
+          echo "    pre-commit run"
+          echo "or after-the-fact on already committed files with"
+          echo "    pre-commit run --all-files"
```
minrk, line 53: ❤️

.github/workflows/test.yml
```diff
@@ -0,0 +1,187 @@
...
+              sudo apt-get install -y postgresql-client
+              DB=postgres bash ci/docker-db.sh
+              DB=postgres bash ci/init-db.sh
+          fi
+
+      - name: Run pytest
+        # FIXME: --color=yes explicitly set because:
+        #        https://github.com/actions/runner/issues/241
+        run: |
+          pytest -vx --color=yes --cov=jupyterhub jupyterhub/tests
```
minrk, line 184: Stopping on *one* failure is often frustrating, since you don't get to see if there are going to be more problems to fix until you push again. We had `--maxfail=2` because more than a couple failures often meant there would be a hang due to possible cleanup issues. We can tune `--maxfail` to a good balance, but I do think it should be more than one.

```suggestion
          pytest -v --maxfail=2 --color=yes --cov=jupyterhub jupyterhub/tests
```

ci/docker-db.sh
```diff
@@ -18,7 +18,7 @@ case "$DB" in
   CHECK="mysql --host $MYSQL_HOST --port $MYSQL_TCP_PORT --user root -e \q"
   ;;
 "postgres")
-  RUN_ARGS="-p 5432:5432 postgres:9.5"
+  RUN_ARGS="-e POSTGRES_HOST_AUTH_METHOD=trust -p 5432:5432 postgres:9.5"
```
minrk, line 21: This environment variable explicitly disables auth, meaning that we are not testing that we have postgres auth set up correctly. If this is required, it means that we *don't* have postgres auth set up correctly.

```suggestion
  RUN_ARGS="-p 5432:5432 postgres:9.5"
```

jupyterhub/tests/test_spawner.py
```diff
@@ -382,7 +382,7 @@ async def test_spawner_delete_server(app):
     assert spawner.server is None
 
 
-@pytest.mark.parametrize("name", ["has@x", "has~x", "has%x", "has%40x"])
+@pytest.mark.parametrize("name", ["has@x", "has%x", "has%40x"])
```
minrk, line 385: This test still needs to pass. If the test is failing, JupyterHub's *behavior* needs to change, not just skipping a test.

However, [this run](https://github.com/minrk/jupyterhub/runs/1378629601) suggests that no change is needed.

```suggestion
@pytest.mark.parametrize("name", ["has@x", "has~x", "has%x", "has%40x"])
```

If there's a change in behavior across Python versions, then we are likely missing some important test matrix entries.

minrk: I dug into it, and you're right, this test fails on Python 3.7+ with subdomains enabled. This is a bug in JupyterHub! We need to make sure that we are testing with subdomains on more versions of Python. The test matrix is too bare if it misses this, as it lacks some key combination (e.g. 3.9 + subdomain).

The issue is not in the socket library, but rather [urllib.parse.quote](https://docs.python.org/3/library/urllib.parse.html?highlight=urllib%20quote#urllib.parse.quote) which added `~` to the list of safe characters in 3.7. We need to handle this in jupyterhub, where we are using url-quoting for a domain name, which isn't quite right.

Short term answer: leave this test alone (this PR will not fail) and we can look at a new PR to add a matrix entry to cover it, and fix the bug

minrk (commented):
Awesome, thanks for this one!

I left a few review suggestions, but overall, seems great!

ci/docker-db.sh
```diff
@@ -1,59 +1,76 @@
...
+# These environment variables influence the mysql and psql clients
+export MYSQL_HOST=${MYSQL_HOST:-127.0.0.1}
 export MYSQL_TCP_PORT=${MYSQL_TCP_PORT:-13306}
-export PGHOST=127.0.0.1
-NAME="hub-test-$DB"
-DOCKER_RUN="docker run -d --name $NAME"
-
-docker rm -f "$NAME" 2>/dev/null || true
-
-case "$DB" in
```
minrk, line 15: Is there a reason we can't use a switch statement anymore?

ci/init-db.sh
```diff
@@ -1,27 +1,30 @@
...
 
 set -eu
 
-MYSQL="mysql --user root --host $MYSQL_HOST --port $MYSQL_TCP_PORT -e "
-PSQL="psql --user postgres -c "
+# These environment variables influence the mysql and psql clients
+export MYSQL_HOST=${MYSQL_HOST:-127.0.0.1}
+export MYSQL_TCP_PORT=${MYSQL_TCP_PORT:-13306}
+export PGHOST=${PGHOST:-127.0.0.1}
+export PGPORT=${PGPORT:-5432}
```
minrk, line 12: These lines don't do anything, do they? Those arguments are not used in this script anymore, they are used by the clients. This effectively overrides the default implementation in the clients themselves, hardcoding new defaults (that are presumably the same), but I think these `exports` can all be removed.

ci/docker-db.sh
```diff
@@ -1,59 +1,76 @@
 #!/usr/bin/env bash
-# source this file to setup postgres and mysql
-# for local testing (as similar as possible to docker)
+# This scripts starts a postgres or mysql database server, waits for it, and
+# setup setups users to be used for local testing.
 
 set -eu
 
-export MYSQL_HOST=127.0.0.1
+# These environment variables influence the mysql and psql clients
```
minrk, line 7: In this script, unlike init-db, this comment is not quite accurate. The ports influence the server, and must be set. Like `init-db`, the HOST args appear to be unused and can be omitted.

ci/docker-db.sh
```diff
@@ -1,59 +1,76 @@
...
+    echo "A dedicated user for mysql isn't created."
+elif [[ "$DB" == "postgres" ]]; then
+    psql --user postgres --command "CREATE USER $PGUSER WITH PASSWORD '$PGPASSWORD';"
+fi
+
+# Print message for humans running the script
+if [[ "$DB" == "mysql" ]]; then
+    echo -e "
+Set these environment variables:
+    export MYSQL_HOST=$MYSQL_HOST
```
minrk, line 68: as above, echoing HOST doesn't provide useful info, since it's either

1. already set and doesn't need to be set again, or
2. not set, and doesn't need to be set because it's the default

.github/workflows/test.yml
manics: I'd also like to run tests on my own branches. See related discussions:
- https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/323/files#r155242906
- https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/388
- https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/500

minrk (commented):
Should be ready to go once the skipped test is reverted. But now there seems to be a total rewrite of the db scripts, so more review to do.

ci/docker-db.sh
consideRatio: It was the readability for myself with some motivation that I believed it could be improved also for others.

.github/workflows/test.yml
consideRatio: This section is your work retained from the .travis.yml file but with a git loss of information through git blame. Do you think a git rename of the .travis.yml file and morphing it to a GitHub workflow file would have retained git history like this? I think so if there are a sequence of git commits rather than one blob.

.github/workflows/test.yml
consideRatio: I think it is a bit wasteful of GitHub's free compute to make this the default behavior even though it would be nice sometimes to have, I would prefer it myself as the default behavior but it feels like a compromise worth making to not be wasteful of the free compute resources given to us.

Create a suggestion and merge it if you want though, I'm not strongly opined about this.

consideRatio:
@minrk pweh! Okay I got it working properly now.

.github/workflows/test.yml
minrk: I'm not worried about git history at that level of detail, thanks for preserving this! I didn't know how to do this kind of thing on actions yet.

minrk:
Awesome, thank you!

minrk:
@meeseeksdev backport to 1.2.x

PR #3250, opened 2020-11-11
remove push-branch conditions for CI

minrk:
testing branches by default is useful, and there's little cost to removing the conditions:

- we don't issue PRs from our repo, so test runs aren't duplicated on this repo for PRs
- testing on a fork without opening a PR is still useful (I use this often)
- if we push a branch, it should probably be tested (e.g. a backport branch), and filters make this extra work
- the cost of running a few extra tests is low, especially given actions' current quotas and parallelism. If this changes (like Travis!) we can reconsider. Very slow repos like z2jh might want to choose differently.

For instance, I've just started an `rbac` branch so we can start landing PRs related to RBAC without going all the way to master. I don't think we should need to modify our workflow configuration to enable tests for this.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,8 +9,6 @@ name: Run tests
 on:
   pull_request:
   push:
-    branches: [main, master, pr/migrate-to-gh-actions-from-travis]
-    tags:
 
 defaults:
   run:


consideRatio (approved):
There was a test flaking, what does spawner.poll status of -2 mean? It is unrelated though so merging.

Thanks for describing your findings about this Min!

minrk:
> what does spawner.poll status of -2 mean?


Just saw this as I was looking for a 1.2.2 changelog. Exit with `-N` means the process was terminated by signal N, in this case 2 == SIGINT (the process was terminated by an interrupt).

In this particular test where we are asserting an exit code of 0 (clean exit), it means that the process was killed less gracefully than it should have been. Most likely, due to not being fully initialized before we tore it down.

consideRatio:
Ah, good to know, thanks for helping me understand that part @minrk!

minrk:
@MeeeeksDev backport to 1.2.x

minrk:
@MeeseeksDev backport to 1.2.x

PR #3252, opened 2020-11-11
Standardize "Sign in" capitalization on the login page

cmd-ntrf:
In the login page form header, the `in` of `Sign in` is not capitalized, while the `In` of the `Sign In` button is capitalized. I would suggest we choose one form or another, but not both in the same contex. This PR implements the non-capitalized `in` version.

Sorry for making a PR for such a small change, but I just saw it and now I cannot unsee it.

### DIFF
--- a/share/jupyterhub/templates/login.html
+++ b/share/jupyterhub/templates/login.html
@@ -61,7 +61,7 @@
         id="login_submit"
         type="submit"
         class='btn btn-jupyter'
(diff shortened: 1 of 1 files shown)


minrk:
Thanks! No need to apologize for improving things :)

minrk:
@meeseeksdev backport to 1.2.x

PR #3254, opened 2020-11-12
Exclude "code" and "state" query params from next_url

### LINKED ISSUE #3247: Jupyter skips spawn form after authentication
### Bug description
We have just noticed a change in behaviour, probably coming from https://github.com/jupyterhub/jupyterhub/commit/6f2e409fb9bb4959310f26f2316bc4d4d228acc0

JupyterHub skips spawner now directly after authentication as it seems some params are passed along from auth and callback:

```
302 GET /hub/oauth_callback?code=[secret]&state=[secret] -> /hub/spawn?code=[secret]&state=[secret] (@::ffff:10.128.2.1) 261.56ms
302 GET /hub/spawn?code=[secret]&state=[secret] -> /hub/spawn-pending/vpavlin?code=[secret]&state=[secret] 
```

The `code` seems to trigger the code in the commit linked above.

This is our oauth config: https://github.com/opendatahub-io/jupyterhub-odh/blob/master/.jupyter/jupyterhub_config.py#L86-L143

Is it possible that it is a misconfiguration on our side? Could it be an issue with authenticator?

#### Expected behaviour
Spawner is shown

#### Actual behaviour
Spawner is skipped

### How to reproduce
I don't have a simple reproducer at the moment, but I can try to get one if this turns out to me more complicated

### Your personal set up

 - OpenShift 4.5
- Kubepawner, oauthenticator/openshift

vpavlin:
This fixes our issue with JH skipping spawner UI after authentization although no query parameters were provided - `code` and `state` stay in query after auth and cause Spawner to kick off. All the details are in the issue #3247.

Fixes #3247 

cc @kinow - since the `append_query_parameters` is your contribution, can you take a look?

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -659,7 +659,9 @@ class BaseHandler(RequestHandler):
             else:
                 next_url = url_path_join(self.hub.base_url, 'home')
 
-        next_url = self.append_query_parameters(next_url, exclude=['next'])
+        next_url = self.append_query_parameters(
+            next_url, exclude=['next', 'code', 'state']
+        )
         return next_url
 
     def append_query_parameters(self, url, exclude=None):


kinow (commented):
Great detective work @vpavlin 

The function you mentioned is used just to strip parameters from the URL. My memory tells me we wanted to strip only `next`, as otherwise any parameter given would have a use further down the request execution lifecycle.

Your change is really small, but it's done in the `handlers.base`. So I suspect that would affect other parts of the code, and possibly other spawners.

If other spawners are working with latest `master`, but openshift fails, then I think the openshift spawner could be doing something wrong with those parameters.

But if other spawners (besides LocalProcessSpawner maybe, that's really simple) are also failing, then someone with more understanding of the code of jupyterhub would have to confirm that we can remove those parameters, as you suggested in this PR :+1: 

Just my 0.02 cents :+1:

vpavlin:
HI @kinow thank you for your thoughts!

This is not a problem of spawner per se. This seems to be the combination of OAuth and the changes in spawner (appending query params and picking them up and doing "auto-spawn").

OAuth actually uses the [BaseHandler.get_next_url](https://github.com/jupyterhub/oauthenticator/blob/master/oauthenticator/oauth2.py#L190), so there are potentially 2 places where to implement my change - in the BaseHandler or in the OAuth.

My preference is BaseHandler since if we only remove `code` and `state` for OAuth, we could get inconsistent behaviour.

Also probably worth mentioning - `code` (a temporary code for client to get access token) and `state` (a unique/random id to identify a request is coiming from a user who initiated the authentication) are used in OAuth flow, so they should not be relied upon after the auth is done.

I would also consider reviewing the whole approach with the query params, since it seems to me that it does not actually work with OAuth. What happened when I tried:

1. Access `/hub?xyz=test`
2. Redirect to `/hub/login` (the query param is lost)
3. Perform login
4. Redirect to `/hub/spawn?code=...&state=...` (no `xyz` param and spawner wants to start based on `code` and `state`)

So my preference woudl be to merge this PR to fix the OAuth flow and then we can dig deeper into how we can preserve the query params for OAuth.

WDYT?

PR #3261, opened 2020-11-17
Only preserve params when ?next= is unspecified

### LINKED ISSUE #3247: Jupyter skips spawn form after authentication

minrk:
Adding params for the current request when `?next=...` has been specified can result in extra parameters being added that should not be.

Additionally, fix the case where parameters on the root `/` handler were not preserved for initial, unauthenticated requests, resulting in different destinations based on initial authenticated state

I believe this fixes #3247

closes #3254

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -634,6 +634,12 @@ class BaseHandler(RequestHandler):
                 next_url,
             )
 
+        # this is where we know if next_url is coming from ?next= param or we are using a default url
+        if next_url:
+            next_url_from_param = True
+        else:
+            next_url_from_param = False
+
         if not next_url:
             # custom default URL, usually passed because user landed on that page but was not logged in
             if default:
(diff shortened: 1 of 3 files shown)


minrk:
@MeeseeksDev hello

minrk:
@MeeseeksDev backport to 1.2.x

PR #3264, opened 2020-11-18
Add user agreement to login screen

tlvu:
This is my first PR.  Please let me know if I have forgotten something.

If enabled (by setting `login_term_url`), user will have to check the checkbox to accept the terms and conditions in order to login.

Example how to enable: https://github.com/bird-house/birdhouse-deploy/blob/c6066482942940bf11661d9ea5e69593b93fbd9d/birdhouse/config/jupyterhub/custom_templates/login.html.template#L38

The feature is disabled by default.

recorded

See PR https://github.com/bird-house/birdhouse-deploy/pull/104 for how this feature is being used.


share/jupyterhub/templates/login.html
```diff
@@ -68,6 +68,18 @@
         <i class="fa fa-spinner"></i>
       </div>
     </div>
+
+    {% block login_terms %}
+    {% if login_term_url %}
+    <div id="login_terms" class="login_terms">
+    <input type="checkbox" id="login_terms_checkbox" name="login_terms_checkbox"/>
```
minrk, line 75: I believe you can use the [required attribute](https://www.w3schools.com/tags/att_input_required.asp) to avoid the need for the javascript enable/disable toggle.

minrk:
Seems okay to me, thanks for making it opt-in. One implementation suggestion made above, but it should work.

share/jupyterhub/templates/login.html
tlvu: Oh great !  Didn't know about the `required` attribute.  Will test it out if it does disable/enable the submit button as we toogle the checkbox then it can completely replace the javascript, which is much better.

Thinking more about this, is there any docs I should update to expose this optional feature?

tlvu: @minrk I've implemented your suggestion.

Overall nicer since no dependency on javascript and user also get a pop-up reminder they need to check the checkbox.  

Similar pop-up behavior for both Mozilla Firefox and Google Chrome.

See animated gif:

recorded

minrk: Nice, thanks!

minrk:
Thank you!

tlvu:
@minrk Thanks !  Are you planning a release in a week or two?  I've patched my deployment with this feature but I'd rather use the real release.  Not urgent if there are something else on your mind.

PR #3265, opened 2020-11-18
Fix RootHandler when default_url is a callable

danlester:
default_url can now be a callable, and this fixes a possible internal server error if the user is already logged in.


minrk:
Thanks!

minrk:
@MeeSeeksDev backport to 1.2.x

PR #3266, opened 2020-11-19
Test internal_ssl separately

### LINKED ISSUE #3262: Fewer internal_ssl tests
Currently, we run our internal_ssl tests by re-running several test modules with `ssl_enabled = True`. There's a lot of redundant testing going on here, since only a small subset of these tests are covering additional functionality the second time they run.

I see two ways to improve things:

1. (probably simplest, easiest to keep good coverage) remove the duplicate `internal_ssl_*` test files and switch the ssl_enabled logic to be based on an environment variable, adding ssl testing to the CI matrix. This will significantly reduce the number of tests each run goes through.
2. keep the current always-test-ssl behavior, but opt-in to a much smaller subset of tests known to cover ssl-related functionality:
  - starting, connecting to spawners
  - registering activity
  - services
  - proxy api

I don't have a strong preference for which one, but I think we can speed up the tests quite a bit (times on my laptop):

- with internal_ssl tests: 3:53
- without internal_ssl tests (keeping test_internal_ssl_connections): 2:22

so we are roughly doubling the runtime of all of our test runs to cover internal_ssl configurations.

0mar:
PR to close #3262.

Moves internal SSL testing to a configuration in the CI matrix based on an environment variable.
This environment variable is only read in the test cases that prior had a separate internal ssl testing module

### DIFF
--- a/jupyterhub/tests/test_api.py
+++ b/jupyterhub/tests/test_api.py
@@ -1,5 +1,6 @@
 """Tests for the REST API."""
 import json
+import os
 import re
 import sys
 import uuid
@@ -27,7 +28,7 @@ from .utils import async_requests
 from .utils import auth_header
 from .utils import find_user
 
-
+ssl_enabled = os.environ.get('SSL_ENABLED', False)
 # --------------------
 # Authentication tests
 # --------------------
--- a/jupyterhub/tests/test_api.py
+++ b/jupyterhub/tests/test_api.py
@@ -1,6 +1,5 @@
 """Tests for the REST API."""
 import json
-import os
 import re
 import sys
 import uuid
@@ -28,7 +27,6 @@ from .utils import async_requests
 from .utils import auth_header
 from .utils import find_user
 
-ssl_enabled = os.environ.get('SSL_ENABLED', False)
 # --------------------
 # Authentication tests
 # --------------------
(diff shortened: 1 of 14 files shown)


jupyterhub/tests/test_api.py
```diff
@@ -27,7 +28,7 @@
 from .utils import auth_header
 from .utils import find_user
 
-
+ssl_enabled = os.environ.get('SSL_ENABLED', False)
```
minrk, line 31: Instead of defining this on each module, can we load it centrally (i.e. *replace* the request.module.ssl_enabled) check with this env check? Or if we still want it sometimes, use something like `getattr(request.module, ssl_enabled, os.environ.get...)`

0mar: That was my first attempt, but I ran into trouble since the tests in e.g. `test_proxy.py` error out due to failing SSL certificates. I thought the `ssl_enabled` module variable was best left in to be able to differentiate between tests that allow for internal SSL and tests that don't.

Alternatively, I can fix the tests that fail under internal SSL?

minrk: That is worthwhile to investigate, I think! The other option would be to make it globally on by default and then skip the tests that require not-ssl if ssl is enabled.

minrk:
Great, thanks! Failure is unrelated, caused by #3289

PR #3288, opened 2020-11-30
add Authenticator.strict_config

minrk:
and better document the relationship of Authenticator user sets and db 

Some folks may expect/want config modification to be able to *remove* access, which is not currently

this is a draft because it's incompatible with secondary allowed-users implementations like group_whitelist, etc. I'm not certain that's a bad thing (it doesn't make much sense to use both), but at the very least the conflict should be clearer before merge.

I'm going to open a separate PR to allow better init logging for Authenticators to declare their configuration, which ought to allow better debugging.


minrk:
Roles and scopes obviate this need - if a deployment doesn't want to allow user creation via the API, it simply doesn't need to give anyone permission to do so. Specifying `user` role membership also works to strictly define the user list.

PR #3289, opened 2020-11-30
fix and test TOTAL_USERS count

### LINKED ISSUE #3280: /hub/metrics total_users double the actual value
As @ianabc reported in https://github.com/jupyterhub/jupyterhub/issues/2685#issuecomment-732310172, there is a discrepancy for the `total_users` metric reported in `/hub/metrics` with the total users reported in `/hub/admin`.

I'm using a JupyterHub 1.2.1. I would guess that total_users is incremented twice per user by mistake.

## /hub/metrics

```
# HELP total_users total number of users
# TYPE total_users gauge
total_users 960.0
```

## /hub/admin

image

image

minrk:
Don't assume UserDict contains all users

which assumption led to double-counting when a user in the db was loaded into the dict cache

closes #3280

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -40,6 +40,7 @@ from ..metrics import SERVER_STOP_DURATION_SECONDS
 from ..metrics import ServerPollStatus
 from ..metrics import ServerSpawnStatus
 from ..metrics import ServerStopStatus
+from ..metrics import TOTAL_USERS
 from ..objects import Server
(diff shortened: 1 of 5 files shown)


consideRatio (approved):
LGTM nice! Test failures are unrelated intermittent failures.

minrk:
There is actually a real failure with the oldest_dependencies test, where the return type is different. Missing these is the hazard of having intermittent failures!

consideRatio:
Oooh woops =/

PR #3295, opened 2020-12-02
begin changelog for 1.3

minrk:
I'd like to do a small release, so we don't accumulate too much backlog into big releases.
There are still a few PRs I'd like to merge before publishing,
but I think we're mostly ready for 1.3.


minrk:
Going to publish 1.3 now!

PR #3303, opened 2020-12-10
Remove the extra parenthesis in service.md

Sangarshanan:


docs/source/reference/services.md
```diff
@@ -346,7 +346,7 @@ and taking note of the following process:
         headers = {
             'Authorization' : 'token %s' % api_token,
         },
-    )
+    ))
```
consideRatio, line 349: I think this code snippet is wrong, but I think the issue the surplus parenthesis after `.join(`, so instead of adding one, lets remove one!

Sangarshanan: Ah yes !! 😅

consideRatio:
Thanks for your help maintaining this project @Sangarshanan ! ❤️🥳

PR #3307, opened 2020-12-16
get groups from authenticator; improve group mgmt on admin page

eliaswimmer:
* Authenticator can supply a group field with group memberships for user
  Groups will be created if no already existing

* Manage groups via admin page
  * Add and delete groups
  * assign and revoke groups to/from users
 * Self clearing fixtures for user(s) and group(s)


jupyterhub/apihandlers/groups.py
```diff
@@ -69,6 +69,28 @@ async def post(self):
...
+            group = orm.Group.find(self.db, name=name)
+            if group is None:
+                raise web.HTTPError(409, "Group %s does not exist" % name)
+            # delete the group
+            self.log.info("Delete group %s", name)
+            self.db.delete(group)
+            self.db.commit()
+            deleted.append(group)
+        self.write(json.dumps([self.group_model(group) for group in deleted]))
+        self.set_status(201)
```
eliaswimmer, line 93: What is the correct return code for bulk delete

jupyterhub/apihandlers/users.py
```diff
@@ -195,7 +203,7 @@ async def get(self, name):
     @admin_only
     async def post(self, name):
         data = self.get_json_body()
-        user = self.find_user(name)
```
eliaswimmer, line 198: No need for the full User Object?!

jupyterhub/apihandlers/users.py
```diff
@@ -244,7 +259,7 @@ async def delete(self, name):
 
     @admin_only
     async def patch(self, name):
-        user = self.find_user(name)
```
eliaswimmer, line 247: Do we really need the full User object here?

eliaswimmer (commented):
Tests still missing

jupyterhub/apihandlers/groups.py
```diff
@@ -69,6 +69,28 @@ async def post(self):
         self.write(json.dumps([self.group_model(group) for group in created]))
         self.set_status(201)
 
+    @admin_only
+    async def delete(self):
```
eliaswimmer, line 73: DELETE should not be used for batch delete.
Better to use PATCH

jupyterhub/apihandlers/groups.py
minrk: 204, generally

jupyterhub/apihandlers/users.py
minrk: Is there a downside to consistency? In nearly all cases, the user is in memory already, and this adds only a single dict lookup to return the object.

jupyterhub/apihandlers/users.py
```diff
@@ -259,11 +274,24 @@ async def patch(self, name):
...
+                # delete user from groups not in new groups
+                for group in user.groups:
+                    if group.name not in groups:
+                        user.groups.remove(group)
             else:
                 setattr(user, key, value)
         self.db.commit()
         user_ = self.user_model(user)
-        user_['auth_state'] = await user.get_auth_state()
+        user_['auth_state'] = await self._user_from_orm(user).get_auth_state()
```
minrk, line 294: I see by this call to `_user_from_orm` that the above `find_user` really is required, so it should probably be put back.

jupyterhub/handlers/base.py
```diff
@@ -734,6 +735,15 @@ async def auth_to_user(self, authenticated, user=None):
         if admin is not None and admin != user.admin:
             user.admin = admin
             self.db.commit()
+        # add user to the given groups
+        if groups is not None:
```
minrk, line 739: 👍

minrk (commented):
This is great! I left a little feedback int he Python code, but everything looks nice. I will take the UI updates for a spin a bit later.

jupyterhub/apihandlers/groups.py
eliaswimmer: hast to be 200, as 204 does not allow to have a body

eliaswimmer:
As #3548 is merged, I close this PR.

PR #3308, opened 2020-12-16
fixed default roles for mocked services

IvanaH8:
Ref.: User roles (RBAC) #2245 

Fixes default role assignment for mocked services used in tests.

As @0mar pointed out (thank you!), the mocked services in API tests were not being assigned default roles. 

This is now fixed with calling `mock_roles` function (added to roles.py) in conftest.py, which assigns a default role to the mocked service.


minrk:
Great!

PR #3312, opened 2020-12-21
Separate metrics port

dtaniwaki:
I'd like to get the JupyterHub metrics from Prometheus which is not mapped as any user in the system. However, the metrics handler is protected by user authentication. As I can protect the metrics port w/o authentication by Kubernetes NetworkPolicy if it's separated.

This implementation may not be enough. Please tell me what else I need to change.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -75,7 +75,7 @@ from jupyter_telemetry.eventlog import EventLog
 here = os.path.dirname(__file__)
 
 import jupyterhub
-from . import handlers, apihandlers
+from . import handlers, apihandlers, metricshandler
 from .handlers.static import CacheControlStaticFilesHandler, LogoHandler
 from .services.service import Service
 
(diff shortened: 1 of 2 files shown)


dtaniwaki:
Furthermore, using `internal-ssl` requires a client cert which is hard to set in Prometheus 😢

jupyterhub/metricshandler.py
```diff
@@ -0,0 +1,18 @@
+from prometheus_client import CONTENT_TYPE_LATEST
+from prometheus_client import generate_latest
```
minrk, line 2: Can we get rid of this extra file and use the existing `jupyterhub/handlers/metrics`?

jupyterhub/app.py
```diff
@@ -2668,6 +2675,24 @@ async def start(self):
...
+                metricshandler.default_handlers, **self.tornado_settings
+            )
+            self.metrics_server = tornado.httpserver.HTTPServer(
+                metrics_tornado_application,
+                ssl_options=ssl_context,
+                xheaders=True,
+                trusted_downstream=self.trusted_downstream_ips,
+            )
+            try:
+                self.metrics_server.listen(self.metrics_port, address="::")
```
minrk, line 2690: I don't think we want to unconditionally bind on all interfaces. Instead, let's use the hub ip already. If that's not always appropriate, then it probably makes sense to use `metrics_url` instead of `metrics_port` so we can configure both port and interface.

```suggestion
                self.metrics_server.listen(self.metrics_port, address=bind_url.hostname)
```

minrk (commented):
Sorry for leaving you hanging. I think it makes sense to do this, but I left a few questions about how exactly to configure it.

PR #3315, opened 2020-12-23
Make Authenticator Custom HTML Flexible

dtaniwaki:
I modified `_render` method of the login handler so I can make the following custom HTML work. The class is based on the idea of https://github.com/jupyterhub/oauthenticator/issues/136#issuecomment-568950338.

```python
class MultiOAuthenticator(Authenticator):
    ...

    def get_custom_html(self, base_url):
        html = []
        for authenticator in self._authenticators:
            login_service = authenticator["instance"].login_service

            url = url_path_join(base_url, authenticator["url_scope"], "oauth_login")

            html.append(
                f"""
                <div class="service-login">
                  <a role="button" class='btn btn-jupyter btn-lg' href='{url}'>
                    Sign in with {login_service}
                  </a>
                </div>
                """
            )
        return "\n".join(html)
```

### DIFF
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -1102,6 +1102,7 @@ class PAMAuthenticator(LocalAuthenticator):
             return super().normalize_username(username)
 
     def get_custom_html(self, base_url):
+        """Get custom HTML for the authenticator."""
         return self.custom_html
 
 
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -1101,6 +1101,9 @@ class PAMAuthenticator(LocalAuthenticator):
         else:
             return super().normalize_username(username)
 
+    def get_custom_html(self, base_url):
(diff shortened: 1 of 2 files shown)


jupyterhub/auth.py
```diff
@@ -1101,6 +1101,10 @@ def normalize_username(self, username):
         else:
             return super().normalize_username(username)
 
+    def get_custom_html(self, base_url):
+        """Get custom HTML for the authenticator."""
```
minrk, line 1105: Can you add:

```python
.. versionadded: 1.4
```

to the docstring?

otherwise, 👍

dtaniwaki:
@minrk I updated the code, so would you check it?

minrk:
Yeah, thanks! I think we may need to think a bit about what's made available (e.g. is it true that base_url is needed, but nothing else?), but this definitely works.

PR #3316, opened 2020-12-23
Fix subdomain issue in google oauth

dtaniwaki:
Using Google OAuth with subdomains doesn't work properly because it uses an email address as a user name.  For example, the email of `foo@example.com` and domain of `jupyterhub.example.com` becomes `foo_example.com.jupyterhub.example.com`. However, we typically have a wild card domain only 1 rank like `*.jupyterhub.example.com` so `foo_example.com.jupyterhub.example.com` is not a valid domain in terms of the SSL certification.


minrk:
Thanks for the PR! Need to think about this a bit, since it's technically a breaking change and the multi-level subdomains have been working in the past. I need to figure out whether this is something we need to allow configuration or if the change is okay to just make.

PR #3323, opened 2021-01-05
Reconciliating API with scopes from database

0mar:
Progress on #2245, ready for review

# List of major changes
 - APIHandlers now allow/refuse request based on scopes from database roles.
 - Refactored `scope` module
    - Made some methods private
    - Moved all scope-related methods from `utils.py` to `scopes.py`
 - Implemented scope-based filter for `*ListApiHandler`
    - Modified `test_api` unit tests to conform with new behaviour
    - Added new unit tests
 - Scope decorators can now be stacked
 - Error messages for users/groups/services requested that do not exist or that do exist but are not accessible for requester now return the same error code (404) and error message (`No access to resources or resources not found`).
## Scope-based filter

In API classes `UserListAPIHandler`, `GroupAPIListHandler` and `ServiceListAPIHandler` a `scope_filter` parameter is added to the `get` methods. Empty by default, this parameter is used by the scope authentication decorator.
When the requester only has access to a subset of the requested resources, the `scope_filter` provides a list of allowed names of resources. The inner body of the methods either adapt their query based on this set or filter the query results.

In case of `UserListAPIHandler`, the decorator also expands groups of users to include in the filter.

# Issues/questions

 - Resource filter implementation is still a bit shoddy
 - Is `self.current_user` always available in an API call? (resolved)
 - Admin scope on `pages.py` is currently set to `admin:users`, does that suffice? (resolved)
 - Setup and teardown of test roles in `test_scopes.py` is a bit inconvenient at the moment. suggestions?
 - External services don't read roles yet, so a test in `test_services` fails. (resolved)

@IvanaH8, do you have any suggestions for the last two points?

### DIFF
--- a/jupyterhub/apihandlers/services.py
+++ b/jupyterhub/apihandlers/services.py
@@ -39,6 +39,7 @@ class ServiceListAPIHandler(APIHandler):
 
 def admin_or_self(method):
     """Decorator for restricting access to either the target service or admin"""
+    """***Deprecated in favor of RBAC, use scope-based decorator***"""
 
     def decorated_method(self, name):
         current = self.current_user
--- a/jupyterhub/apihandlers/services.py
+++ b/jupyterhub/apihandlers/services.py
@@ -9,8 +9,7 @@ import json
 from tornado import web
 
 from .. import orm
-from ..utils import admin_only
-from ..utils import needs_scope
+from ..scopes import needs_scope
 from .base import APIHandler
 
 
--- a/jupyterhub/apihandlers/services.py
+++ b/jupyterhub/apihandlers/services.py
@@ -31,8 +31,10 @@ def service_model(service):
 
 class ServiceListAPIHandler(APIHandler):
     @needs_scope('read:services')
-    def get(self):
+    def get(self, scope_filter=None):
         data = {name: service_model(service) for name, service in self.services.items()}
+        if scope_filter is not None:
+            data = dict(filter(lambda tup: tup[0] in scope_filter))
         self.write(json.dumps(data))
 
 
--- a/jupyterhub/apihandlers/services.py
+++ b/jupyterhub/apihandlers/services.py
@@ -33,7 +33,7 @@ class ServiceListAPIHandler(APIHandler):
     def get(self, scope_filter=None):
         data = {name: service_model(service) for name, service in self.services.items()}
         if scope_filter is not None:
-            data = dict(filter(lambda tup: tup[0] in scope_filter))
+            data = dict(filter(lambda tup: tup[0] in scope_filter, data.items()))
         self.write(json.dumps(data))
 
 
--- a/jupyterhub/tests/test_scopes.py
+++ b/jupyterhub/tests/test_scopes.py
@@ -217,7 +217,7 @@ async def test_expand_groups(app, user_name, in_group, status_code):
 
 
 async def test_user_filter(app):
-    user_name = 'rollerblade'
+    user_name = 'rita'
     test_role = {
         'name': 'test',
         'description': '',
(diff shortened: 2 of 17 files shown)


IvanaH8:
Thanks @0mar, initial comments to the last 3 questions: 

> - Admin scope on `pages.py` is currently set to `admin:users`, does that suffice?

This scope allows only to create/delete users while the admin page allows for starting/stopping users' servers, changing admin status, etc. so I think it should have multiple scopes of `users`, `admin:users` and `admin:users:servers`.

> - Setup and teardown of test roles in `test_scopes.py` is a bit inconvenient at the moment. suggestions?

I will have a look, we could probably merge some role tests in `test_scopes.py` and `test_roles.py`.

> - External services don't read roles yet, so a test in `test_services` fails.

I think the problem in this particular test is that neither the hub itself nor the service here is mocked so the `init_roles()` does not get called. The service is simply 'added' to the app (like it would through the config file that's read in when the app is initiated and all `init_...()` functions are executed one by one, including the `init_roles()`) so we need to manually add the `init_roles()` func there after `init_services()`.

jupyterhub/tests/test_scopes.py
```diff
@@ -189,11 +228,105 @@ async def test_expand_groups(app, user_name, in_group, status_code):
...
+    roles.add_role(app.db, test_role)
+    user = add_user(app.db, name=user_name)
+    name_set = {'sally', 'stan'}
+    group_name = 'sitwell'
+    group = orm.Group.find(app.db, name=group_name)
+    if not group:
+        group = orm.Group(name=group_name)
+        app.db.add(group)
+    for name in name_set:
+        user = add_user(app.db, name=name)
```
minrk, line 293: `user` variable is redefined here and will sometimes be `stan` and sometimes `sally` after this loop, meaning the `update_roles` call below is sometimes called for sally and sometimes for stan. Python's forced iteration randomization of unordered collections means that iterating through `name_set` is *explicitly* nondeterministic. The result should be expected failures ~50% of the time. Renaming this variable to avoid colliding with `user` should solve the issue.

jupyterhub/apihandlers/services.py
```diff
@@ -31,8 +30,10 @@ def service_model(service):
 
 class ServiceListAPIHandler(APIHandler):
     @needs_scope('read:services')
-    def get(self):
+    def get(self, scope_filter=None):
         data = {name: service_model(service) for name, service in self.services.items()}
+        if scope_filter is not None:
+            data = dict(filter(lambda tup: tup[0] in scope_filter))
```
minrk, line 36: I don't think this branch is tested, because this filter isn't valid (filter requires two arguments, the filter function and the thing being filtered).

What structure does  scope_filter have, exactly?

You can check for coverage of your PR (after tests finish) at https://app.codecov.io/gh/jupyterhub/jupyterhub/compare/3323/diff

minrk (commented):
Looking really good! I left some comments, but I think this is about ready.

jupyterhub/tests/test_scopes.py
```diff
@@ -189,11 +228,105 @@ async def test_expand_groups(app, user_name, in_group, status_code):
...
+    roles.update_roles(app.db, user, kind, roles=['test'])
+    roles.remove_obj(app.db, user_name, kind, 'user')
+    app.db.commit()
+    r = await api_request(app, 'users', headers=auth_header(app.db, user_name))
+    assert r.status_code == 200
+    result_names = {user['name'] for user in r.json()}
+    assert result_names == name_in_scope
+
+
+async def test_user_filter_with_group(app):  # todo: Move role setup to setup method
```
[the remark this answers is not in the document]
0mar: Good idea, I'll add that

jupyterhub/apihandlers/services.py
0mar: Thanks! The code coverage tool helps. And indeed, a scope_filter should be an iterable (set) of resource names. I'll write a test for it and fix it

PR #3326, opened 2021-01-08
Allow to set spawner-specific hub connect URL

dtaniwaki:
I'd like to use JupyterHub w/ `DockerSpawner` but don't want to set up any domain which points to the host IP. Docker for Mac supports `host.docker.internal` which indicates the host IP in Docker containers, so I can use it as hub connect URL. However, the hub config of hub_connect_url is also for chp. This fix allow seperately set the URL.

Currently, I created a custom spawner inheriting DockerSpawner and replace the hostname as below.

```python
import re
from urllib.parse import urlsplit, urlunsplit

import dockerspawner

_docker_hostname = "host.docker.internal"


def _replace_hostname(url: str, hostname: str) -> str:
    p = list(urlsplit(url))
    p[1] = re.sub("^[^:]*", hostname, p[1])
    return urlunsplit(p)


class CustomDockerSpawner(dockerspawner.DockerSpawner):
    def get_env(self):
        env = super(CustomDockerSpawner, self).get_env()
        for target in ("JUPYTERHUB_API_URL", "JUPYTERHUB_ACTIVITY_URL"):
            env[target] = _replace_hostname(env[target], _docker_hostname)
        return env
```

After this fix, it will be,

```python
c.DockerSpawner.hub_connect_url = "host.docker.internal"
```

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -699,8 +699,8 @@ class Spawner(LoggingConfigurable):
         If the Hub URL set in your JupyterHub config is not reachable
         from spawned notebooks, you can set differnt URL by this config.
 
-        Is None if you want to use the same API URL.
-        """
+        Is None if you don't need to change the URL.
+        """,
     ).tag(config=True)
 
     def load_state(self, state):
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -783,7 +783,9 @@ class Spawner(LoggingConfigurable):
         env['JUPYTERHUB_USER'] = self.user.name
         env['JUPYTERHUB_SERVER_NAME'] = self.name
         if self.hub_connect_url is not None:
-            hub_api_url = url_path_join(self.hub_connect_url, urlparse(self.hub.api_url).path)
+            hub_api_url = url_path_join(
+                self.hub_connect_url, urlparse(self.hub.api_url).path
+            )
         else:
             hub_api_url = self.hub.api_url
         env['JUPYTERHUB_API_URL'] = hub_api_url
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -13,6 +13,7 @@ import sys
 import warnings
 from subprocess import Popen
 from tempfile import mkdtemp
+from urllib.parse import urlparse
 
 if os.name == 'nt':
     import psutil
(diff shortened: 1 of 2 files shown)


minrk (commented):
Makes sense! A little more clarification about when to use this instead of the existing config and we can be ready to go.

dtaniwaki:
@minrk I updated the code, thanks!

jupyterhub/tests/test_spawner.py
```diff
@@ -415,3 +415,9 @@ async def test_spawner_env(db):
     for key, value in env_overrides.items():
         assert key in env
         assert env[key] == value
+
+
+async def test_hub_connect_url(db):
+    spawner = new_spawner(db, hub_connect_url="example.com")
+    env = spawner.get_env()
+    env["JUPYTERHUB_API_URL"] = "https://example.com/hub/api"
```
minrk, line 423:
```suggestion
    assert env["JUPYTERHUB_API_URL"] == "https://example.com/hub/api"
```

jupyterhub/spawner.py
```diff
@@ -768,9 +783,9 @@ def get_env(self):
         # Info previously passed on args
         env['JUPYTERHUB_USER'] = self.user.name
         env['JUPYTERHUB_SERVER_NAME'] = self.name
-        env['JUPYTERHUB_API_URL'] = self.hub.api_url
+        env['JUPYTERHUB_API_URL'] = self.hub_connect_url
```
minrk, line 786: This is not treated the same as JuptyerHub.hub_connect_url, and it probably should be based on the description. In which case, this should be `url_path_join(self.hub_connect_url, 'hub/api')` and ACTIVITY_URL should match.

dtaniwaki: I'm a little bit confused. Do you think asking a user to set a URL with `/hub/api` or just URL w/o path on `hub_connect_url`?

minrk: To match `hub_connect_url` it should not include the `hub/api` path

dtaniwaki:
@minrk I think the CI failure came from other changes. Could you help me to fix the failure?

dtaniwaki:
Hmm... Maybe, `hub_connect_url` is not set in the test environment.

dtaniwaki:
I think this is it 😄

minrk:
Great, thank you!

PR #3329, opened 2021-01-12
Mention Jupyter Server as optional single-user backend in documentation

Zsailer:
Jupyter Server 1.0 was released mid-2020 and JupyterLab 3.0 uses it by default (though JLab 3.0 is still compatible with the classic notebook server).

#3128 added some logic to allow users to switch to Jupyter Server as the single-user server backend by setting the `JUPYTERHUB_SINGLEUSER_APP` environment variable. 

Here is some documentation to point users to this environment variable.

### DIFF
--- a/docs/source/reference/config-user-env.md
+++ b/docs/source/reference/config-user-env.md
@@ -179,3 +179,13 @@ The number of named servers per user can be limited by setting
 ```python
 c.JupyterHub.named_server_limit_per_user = 5
 ```
+
+## Switching to Jupyter Server
+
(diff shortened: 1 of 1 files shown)


Zsailer:
Failing test appears to be unrelated.

minrk:
Nice! Note that once https://github.com/jupyterlab/jupyterlab/issues/8807 is addressed, setting this directly should not be necessary. Changing the entrypoint to e.g. `jupyter-labhub` will have the same effect and be a more appropriate official way to enable jupyterlab 3.0.

PR #3333, opened 2021-01-13
:memo: Fix telemetry section

### LINKED ISSUE #3239: Docs for event logging and telemetry have two errors
### Bug description
The docs about event logging and telemetry have two errors.

### How to reproduce error 1
1. Go to the docs' [raw source](https://raw.githubusercontent.com/jupyterhub/jupyterhub/ca00c0eab065c18284d0da3ca7b57de14041d3df/docs/source/events/index.rst).
2. Notice that the third paragraph seems to be extraneous, or accidentally left in place when it shouldn't have been.

### How to reproduce error 1
1. Notice in that same raw source that there is a code block about 60% of the way down the page...
2. Go to the docs [online](https://jupyterhub.readthedocs.io/en/stable/events/index.html).
3. Notice that the code block simply doesn't appear, making the sentence "Here is a basic example:" quite confusing, because it is not actually followed by the promised basic example.

trallard:
Fixes #3239 

I also went ahead and rephrased a chunk of text to remove ambiguity. (replace below..)

### DIFF
--- a/docs/source/events/index.rst
+++ b/docs/source/events/index.rst
@@ -1,10 +1,7 @@
 Eventlogging and Telemetry
 ==========================
 
-JupyterHub can be configured to record structured events from a running server using Jupyter's `Telemetry System`_. The types of events that JupyterHub emits are defined by `JSON schemas`_ listed below_
-
- emitted as JSON data, defined and validated by the JSON schemas listed below.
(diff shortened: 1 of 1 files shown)


minrk:
Nice, thanks!

Issue #3334, opened 2021-01-15
Admin pagination issues with named servers


rcthomas:
### Bug description

Pagination's great, but I think it gets confused by named servers.  After named servers are registered they count as "users" (entries?) when being rendered, but the `pagination.total` is only counting users. 

#### Expected behaviour

If you have named servers going on, maybe the admin page should set the total to be from the total number of `Spawner` entries or take that into account.  The behavior for no named servers could be just left alone.  The admin page could say it was listing spawners in the bottom too.

#### Actual behaviour

Assuming default settings, you might have 789 users, but if they have some named servers you won't be able to see them all listed unless you override the `page` argument. If each one has a named server, you'll only see like half. Interestingly though you'll see at least the number of default entries per page (like if a user has 105 named servers you'll see their user row and 105 named server rows, but no other users).

### How to reproduce

1. Set up a hub with an admin and test users, say 12 of them
2. For `jupyterhub_config.py`:

```
c.JupyterHub.admin_users = set(["admin"])
c.JupyterHub.allow_named_servers = True
c.JupyterHub.default_url = "/hub/home"
c.Authenticator.allowed_users = set([
    "user01", "user02", "user03",
    "user04", "user05", "user06",
    "user07", "user08", "user09",
    "user10", "user11", "user12",
    "admin"
])
c.Pagination.default_per_page = 3
```

3. Login as the `admin` user and go to the admin page. You should see 3 users listed and pages for 5 pages worth of users which seems right.
4. Go to the home page and add say 4 servers for yourself.
5. When you're done, go back to the admin page.  What you see is the admin user, and then the 4 servers you created.  The pagination info still says it's showing a subset of users for 13 total users and there are 5 pages worth
6. Move to the last page; some of the higher-numbered users will be missing.
7. They are still there but you have to set `?page=` off the end of the list of expected pages.

### Work

I'm going to see about whether the proposed solution works and submit a PR but I'm a little confused how the template knows to stop showing users past the per_page setting.  If anyone has a hint let me know but I'll probably eventually track it down.

rcthomas:
Looks like you just need to make the user query return only distinct users then apply the limit and offset.  PR shortly.

rcthomas:
Closed by #3335

PR #3335, opened 2021-01-15
Fix pagination with named servers

rcthomas:
Just need to consider distinct users for pagination to work right with named servers.  I am not a sqlalchemy expert (sqlalchemist?) so I am not sure if the "optimization" I did actually does what I think it should or if it actually makes 3 queries happen under the hood.

* Add distinct user id clause
* Separate join and distinct into one query
* Apply limit and offset to the result
* Apply count to the result for pagination total

### DIFF
--- a/jupyterhub/handlers/pages.py
+++ b/jupyterhub/handlers/pages.py
@@ -501,20 +501,16 @@ class AdminHandler(BaseHandler):
         # get User.col.desc() order objects
         ordered = [getattr(c, o)() for c, o in zip(cols, orders)]
 
-        users = (
-            self.db.query(orm.User)
-            .outerjoin(orm.Spawner)
-            .order_by(*ordered)
-            .limit(per_page)
-            .offset(offset)
-        )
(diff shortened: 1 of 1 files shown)


rcthomas:
Addresses #3334

rcthomas:
Not quite, looks like the mysql and postgresql failures have to do with distinct and ordering in the `self._user_from_orm(u)` mapping.

rcthomas:
Worked when I was using the default sqlite backend but I tried postgresql myself and yeah, looks like if you want unique `user.id` then that also needs to be first in the `order by` clause, I don't think we want that.  Switching to `user.name` might be OK but we don't want that to be first in `order by` either if admin-ness should come first.  Suggestions?

rcthomas:
Haha, fixed the specific problem I had but now makes more tests fail than before! 🙄

rcthomas:
Yay!  Not sure this is the most optimal query, if someone else can think of something better please do!

minrk:
Thanks!

PR #3337, opened 2021-01-19
Add Spawner.delete_forever

nsshah1288:
related to this PR: https://github.com/jupyterhub/kubespawner/pull/475

tested with sqlite pvc

### DIFF
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -237,6 +237,10 @@ class UserAPIHandler(APIHandler):
                 )
 
         await maybe_future(self.authenticator.delete_user(user))
+
+        # delete the user's PVC
+        await user.spawner.delete_namespaced_persistent_volume_claim(user.spawner.pvc_name, user.spawner.namespace)
+
         # remove from registry
         self.users.delete(user)
 
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -241,7 +241,7 @@ class UserAPIHandler(APIHandler):
         # Delete the user's storage, if it exists.
         # In kubespawner spawner.py, we have delete_forever
         # method to delete the user's PVC.
-        await user.spawner.delete_forever()
+        user.spawner.delete_forever()
 
         # remove from registry
         self.users.delete(user)
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -239,7 +239,9 @@ class UserAPIHandler(APIHandler):
         await maybe_future(self.authenticator.delete_user(user))
 
         # delete the user's PVC
-        await user.spawner.delete_namespaced_persistent_volume_claim(user.spawner.pvc_name, user.spawner.namespace)
+        await user.spawner.delete_namespaced_persistent_volume_claim(
+            user.spawner.pvc_name, user.spawner.namespace
+        )
 
         # remove from registry
         self.users.delete(user)
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -238,10 +238,10 @@ class UserAPIHandler(APIHandler):
 
         await maybe_future(self.authenticator.delete_user(user))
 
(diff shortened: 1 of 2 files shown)


minrk:
Thanks for the PR! Unfortunately, it's not appropriate to have any kubernetes-specific methods in this repo like references to PVCs. Spawners having a hook to cleanup resources on user-delete is an interesting challenge because there may not be any existing Spawners at the time the user is deleted, but we can explore it.

Maybe a generic `delete_forever()` hook method to be called when the user is to be deleted that should be called by jupyterhub?

nsshah1288:
Hi @minrk , thanks for the reply!

I have changed the code a bit, what do you think? I moved the kubernetes-specific part to spawner.py here: https://github.com/jupyterhub/kubespawner/pull/475/files

Now would it be like all other Spawners would need this method, `delete_forever` as part of their spawner base class?

Also, could you please explain more on this part? "there may not be any existing Spawners at the time the user is deleted" -- not sure I understand this

I tested this change again with sqlite pvc, by using the admin console.
Thanks!

jupyterhub/apihandlers/users.py
```diff
@@ -237,6 +237,12 @@ async def delete(self, name):
                 )
 
         await maybe_future(self.authenticator.delete_user(user))
+
+        # Delete the user's storage, if it exists.
+        # In kubespawner spawner.py, we have delete_forever
+        # method to delete the user's PVC.
+        user.spawner.delete_forever()
```
minrk, line 244: This should match the `await maybe_future()` above to allow async operations. No need for kubespawner discussion in the comments. If there's a comment, use "allow the spawner to cleanup any persistent resources associated with the user.

Probably even want try/except here because I suspect we don't want a spawner operation error to prevent user deletion.

jupyterhub/tests/mocking.py
```diff
@@ -93,6 +93,9 @@ def user_env(self, env):
     def _cmd_default(self):
         return [sys.executable, '-m', 'jupyterhub.tests.mocksu']
 
+    def delete_forever(self):
```
minrk, line 96: Add a docstring explaining what this is an when it is called:

> Called when a user is deleted. This can do things like request removal of resources such as persistent storage. Only called on stopped spawners, and is likely the last action ever taken for the user. Will only be called once on the user's default Spawner.

jupyterhub/tests/mocking.py
```diff
@@ -93,6 +93,9 @@ def user_env(self, env):
...
+        return "mocking_this_method"
```
minrk, line 97: Use `pass` for empty base methods:

```suggestion
        pass
```

jupyterhub/tests/mocking.py
nsshah1288: added in https://github.com/jupyterhub/kubespawner/pull/475/files. Hope that's where you wanted it

jupyterhub/tests/mocking.py
nsshah1288: done

jupyterhub/apihandlers/users.py
nsshah1288: added 'await' and try/except, also changed comment

jupyterhub/tests/mocking.py
nsshah1288: also added a docstring here in mocking.py

nsshah1288:
Hi @minrk , I have tried to resolve all your comments, can you please let me know if anything else is required?

Note: This solution only works for dynamic singleuser storage, not for static. I tested with static storage as well (EFS) and I was still able to delete a user thru the admin console fine, but the underlying storage was not deleted.

minrk:
Great, thanks! I think this is enough fromt eh juptyerhub side.

nsshah1288:
Awesome, thanks @minrk ! If this kubespawner PR also looks good, could you approve it as well? https://github.com/jupyterhub/kubespawner/pull/475/files

meeseeksmachine:

https://discourse.jupyter.org/t/should-remove-option-be-documented-with-delete-users-name-server/8457/2

PR #3338, opened 2021-01-20
always log slow requests at least at info-level

minrk:
if health or static responses are taking longer than 1s, it's useful to know

### DIFF
--- a/jupyterhub/log.py
+++ b/jupyterhub/log.py
@@ -2,7 +2,9 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 import json
+import logging
 import traceback
+from functools import partial
 from http.cookies import SimpleCookie
 from urllib.parse import urlparse
 from urllib.parse import urlunparse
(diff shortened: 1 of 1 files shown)


jupyterhub/log.py
```diff
@@ -132,19 +134,25 @@ def log_request(handler):
...
     uri = _scrub_uri(request.uri)
     headers = _scrub_headers(request.headers)
 
     request_time = 1000.0 * handler.request.request_time()
 
+    # always log slow responses (longer than 1s) at least info-level
+    if request_time >= 1000 and log_level < logging.INFO:
+        log_level = logging.INFO
+
+    log_method = partial(access_log.log, log_level)
```
consideRatio, line 154: I could not verify that this function takes a log_level as its first argument, but this LGTM if it does.

consideRatio (approved):

jupyterhub/log.py
minrk: This is how all the log level methods are implemented - `logging.info(...) == logging.log(logging.INFO, ...)`. Method signature is documented [here](https://docs.python.org/3/library/logging.html#logging.Logger.log).

PR #3341, opened 2021-01-24
Clear tornado xsrf cookie on logout

dtaniwaki:
We should remove `_xsrf` cookie [set by tornado](https://github.com/tornadoweb/tornado/blob/cef029abd244cd796ea36487050f722a57bfab60/tornado/web.py#L1399-L1420) just in case for security.


minrk:
Sure!

PR #3342, opened 2021-01-24
Clear cookie ID on logout

dtaniwaki:
Your cookie ID doesn't get cleared when you logout from JupyterHub. Therefore, when your cookie ID is stolen in some way, you cannot immediately expire the cookie ID even if you logout from JupyterHub.

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -483,7 +483,11 @@ class BaseHandler(RequestHandler):
                     count += 1
                 if count:
                     self.log.debug("Deleted %s access tokens for %s", count, user.name)
-                    self.db.commit()
+
+                # clear cookie from DB
+                user.set_new_cookie_id()
+                self.db.add(user)
+                self.db.commit()
 
         # clear hub cookie
         self.clear_cookie(self.hub.cookie_name, path=self.hub.base_url, **kwargs)
--- a/jupyterhub/orm.py
+++ b/jupyterhub/orm.py
@@ -221,6 +221,9 @@ class User(Base):
         """
         return db.query(cls).filter(cls.name == name).first()
 
+    def set_new_cookie_id(self):
+        self.cookie_id = new_token()
+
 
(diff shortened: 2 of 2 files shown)


minrk:
The cookie id is itself not a useful piece of information, even if stolen, unless the encryption keys for the deployment are also stolen. What issue is this addressing?

Clearing this on logout necessarily logs out all other sessions for the same user on all browsers, which shouldn't happen.

dtaniwaki:
A bad guy who steals the cookie can impersonate the person. Currently, the person doesn't have any method to stop it.

minrk:
Good point. Adding a separate endpoint to logout all other sessions is a sensible feature to add. That should be a new endpoint, though. Logging out all other sessions on every logout isn't the right thing to do, though.

dtaniwaki:
Or, how about making it configurable? Having 2 buttons of different endpoints may confuse users.

minrk:
> how about making it configurable?

I think they are totally different functionalities:

- logout this session (what we have now)
- logout other sessions (a new feature)

I don't think it is ever the right thing to do for a regular logout action of one session to break other login sessions unconditionally.

I'm not quite sure where it should go UI-wise, since we don't have an advanced/user-preferences page at the moment, which I think is the logical place for it. It definitely should not be next to the existing, standard 'logout' button.

For now, I might put it on the 'tokens' page, since that's where we do have existing controls for revoking logins of other sessions, even though they are of a different sort internally, and it's an implementation detail that hub cookies are (currently) distinct from OAuth tokens. That may change, in fact, with the RBAC work. A "logout all sessions" button on that page would make sense.

dtaniwaki:
I see. I didn’t think about the tokens because my JupyterHub doesn’t provide multi-servers nor tokens. We actually removed the page and handler entirely.
I’m trying to implement it like your suggestion when I have time.

minrk:
Sounds great! Feel free to ping if you have any questions about the implementation.

PR #3343, opened 2021-01-25
Allow cookie_secret to be set to a hexadecimal string

consideRatio:
As YAML/JSON configuration cannot represent bytes directly, it is helpful to support a string representation alongside a Bytes representation.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -848,14 +848,29 @@ class JupyterHub(Application):
         to reduce the cost of checking authentication tokens.
         """,
     ).tag(config=True)
-    cookie_secret = Bytes(
+    cookie_secret = Union(
+        [Bytes(), Unicode()],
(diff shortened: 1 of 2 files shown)


jupyterhub/app.py
```diff
@@ -848,15 +848,30 @@ def _update_hub_prefix(self, change):
...
         Should be exactly 256 bits (32 bytes).
-        """
+        """,
     ).tag(config=True, env='JPY_COOKIE_SECRET')
 
+    @validate('cookie_secret')
+    def _validate_secret_key(self, proposal):
+        """Coerces strings with even number of hexadecimal characters to bytes."""
+        r = proposal['value']
+        if type(r) == str:
```
minrk, line 865:
```suggestion
        if isinstance(r, str):
```

minrk (commented):
Makes sense, given constraints.

minrk:
Thanks!

PR #3344, opened 2021-01-26
[TST] Do not implicitly create users in auth_header

minrk:
implicit user creation results in surprising behavior when the user shouldn't exist


minrk:
@0mar this surprising behavior is what caused your latest unexpected test failure. With this change you would have gotten a more meaningful error for requesting an auth header for a user that doesn't exist anymore.

PR #3345, opened 2021-01-26
Allow customization of service menu via templates

stv0g:
This is a relative minor PR to allow the customization of the service dropdown menu


minrk:
Nice!

PR #3349, opened 2021-01-29
publish release outputs as artifacts

minrk:
makes local testing of a PR even easier since we build an sdist and wheel for every PR and push. We are already building this, this change only makes them available via the `artifacts` link of a workflow run.

since artifacts are double-archived, it's not quite as simple as giving a URL to install from, but this at least makes it available. To use:

- download and unpack zip
- `pip install path/to/whl`

I've been using this extensively to test recent issues in pyzmq releases, and it's awesome.

Artifacts expire by default after 30 days, I believe.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -51,6 +51,7 @@ jobs:
           print("OK")
           EOF
 
+      # ref: https://github.com/actions/upload-artifact#readme
       - uses: actions/upload-artifact@v2
         with:
           name: jupyterhub-${{ github.sha }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -51,6 +51,12 @@ jobs:
           print("OK")
           EOF
 
+      - uses: actions/upload-artifact@v2
+        with:
+          name: jupyterhub-${{ github.sha }}
+          path: "dist/*"
+          if-no-files-found: error
+
       - name: Publish to PyPI
         if: startsWith(github.ref, 'refs/tags/')
         env:


consideRatio:
@minrk I believe 30 days are the default on project level, but it is configurable up to 400 days in the org level, but it seems like the cap within a project is 90 days. Anyhow, it is 90 days for jupyterhub/jupyterhub now.

image

.github/workflows/release.yml
```diff
@@ -51,6 +51,12 @@ jobs:
           print("OK")
           EOF
 
+      - uses: actions/upload-artifact@v2
+        with:
+          name: jupyterhub-${{ github.sha }}
+          path: "dist/*"
+          if-no-files-found: error
+
```
consideRatio, line 59:
```suggestion
      # ref: https://github.com/actions/upload-artifact#readme
      - uses: actions/upload-artifact@v2
        with:
          name: jupyterhub-${{ github.sha }}
          path: "dist/*"
          if-no-files-found: error

```

consideRatio: I find it generally useful to have a quick reference to external actions to find changelogs, inputs / outputs, etc.

consideRatio (approved):
LGTM! Good to know about this action!

.github/workflows/release.yml
consideRatio: Woops the preview wasn't so nice now that I commented on multiple lines, won't do that again. I only added the reference link though.

PR #3357, opened 2021-02-10
Do not pass self twice to a options_form callable

dolfinus:
If someone wants to create custom Spawner class with `options_form` as a function, it should accept 2 arguments:
```python
def options_form(self, _self):
  ...
```
This second argument is just the same as `self`, and there is no reason to use it.

Another way to define this method to avoid adding the second argument is:
```python
@staticmethod
def options_form(self)
```
Which smells really awful.

Here I've removed this unnecessary argument and also added few more tests for different cases.

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -347,7 +347,7 @@ class Spawner(LoggingConfigurable):
         .. versionadded:: 0.9
         """
         if callable(self.options_form):
-            options_form = await maybe_future(self.options_form(self))
+            options_form = await maybe_future(self.options_form())
         else:
             options_form = self.options_form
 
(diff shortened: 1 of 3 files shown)


minrk:
Thanks for the PR! Unfortunately, I don't think we should make this change.

`options_form` comes from *configuration*, and it can be a string or callable to return a string. It is not a method to be overridden in subclasses, so it will not get `self` automatically. For example:

```python
async def render_options_form(spawner):
    ....
c.Spawner.options_form = render_options_form
```

So it will be called with the spawner argument only once. This change would make spawners unavailable to options_form from config, and require subclassing to do that.

If you want to override the options form in a subclass to *not* be configurable, override the `get_options_form()` method, which has the signature you are expecting.

dolfinus:
Fair enough, I haven't thought about such an option.

PR #3363, opened 2021-02-15
Vertical filtering

0mar:
Implemented filters to restrict access to components in user, group and service models.
These changes affect the following API endpoints:

 - UserListAPIHandler
 - GroupListAPIHandler
 - ServiceListAPIHandler

### Changes/highlights:
 - If client has no access to `servers` in the user model, this component is not included in the model, rather than sending `{'servers':None}`.
 - Scopes `read:users:activity`, `read:users:groups`, and `read:users:servers` return a user model with components `kind` and `name`, in addition to the resource present in the scope.

## Additional changes 
 - Token scope limitation also done at API request level
 - `all` scope is now a metascope only valid for tokens that expands to all the scopes of the token owner
 - `self` scope is now a metascope only valid for users that expands to all the scopes used to read their own data
 - Default token role is `token` with scope `all`
 - Default user role is `user` with scope `self`
 
## Errors
 - Test `test_singleuser.py::test_singleuser_auth` fails. I think it is related to OAuth vs API tokens so I didn't spend too much time on debugging it.
 - Test `test_auth_expiry.py::test_auth_expired_api` fails too. In the test, cookies get invalidated by letting the user authorization expire. But the API token is still active, which is how authorization succeeds anyway. Not sure if this is an error in test assumptions or in the code?

## Future work (not included in this PR)
 - Any token should have access to the name of its owner
 - Create an `identify` scope that sees the name/groups/servers of an owner
 - Create an `auth_state` scope that allows obtaining users' `auth_state`
 - Ensure that horizontal filtering and vertical filtering is resolved correctly when an token and its owner have non-overlapping scope models

### DIFF
--- a/jupyterhub/apihandlers/base.py
+++ b/jupyterhub/apihandlers/base.py
@@ -10,6 +10,7 @@ from tornado import web
 
 from .. import orm
 from ..handlers import BaseHandler
+from ..handlers import scopes
 from ..utils import isoformat
 from ..utils import url_path_join
 
@@ -61,6 +62,38 @@ class APIHandler(BaseHandler):
             )
             return False
         return True
+
+    def get_scope_filter(self, req_scope):
+        """Produce a filter for `*ListAPIHandlers* so that GET method knows which models to return"""
+        scope_translator = {
+            'read:users': 'users',
+            'read:services': 'services',
+            'read:groups': 'groups',
+        }
+        if req_scope not in scope_translator:
+            raise AttributeError("Internal error: inconsistent scope situation")
+        kind = scope_translator[req_scope]
+        Resource = orm.get_class(kind)
+        try:
+            sub_scope = self.parsed_scopes[req_scope]
+        except AttributeError:
+            raise web.HTTPError(
+                403,
+                "Resource scope %s (that was just accessed) not found in scopes anymore"
+                % req_scope,
+            )
+        if sub_scope == scopes.Scope.ALL:
+            return None  # Full access
+        sub_scope_values = next(iter(sub_scope.values()))
+        query = self.db.query(Resource).filter(Resource.name.in_(sub_scope_values))
+        scope_filter = {entry.name for entry in query.all()}
+        if 'group' in sub_scope and kind == 'users':
+            groups = orm.Group.name.in_(sub_scope['group'])
+            users_in_groups = (
+                self.db.query(orm.User).join(orm.Group.users).filter(groups)
+            )
+            scope_filter |= {user.name for user in users_in_groups}
+        return scope_filter
 
     def get_current_user_cookie(self):
         """Override get_user_cookie to check Referer header"""
@@ -189,46 +222,61 @@ class APIHandler(BaseHandler):
         """Get the JSON model for a User object"""
         if isinstance(user, orm.User):
             user = self.users[user.id]
-        model = {'kind': 'user'}
+        model = {
+            'kind': 'user',
+            'name': user.name,
+            'admin': user.admin,
+            'roles': [r.name for r in user.roles],
+            'groups': [g.name for g in user.groups],
+            'server': user.url if user.running else None,
+            'pending': None,
+            'created': isoformat(user.created),
+            'last_activity': isoformat(user.last_activity),
+        }
+        access_map = {
+            'read:users': {
+                'kind',
+                'name',
+                'admin',
+                'roles',
+                'groups',
+                'server',
+                'servers',
+                'pending',
+                'created',
+                'last_activity',
+            },
+            'read:users:name': {'kind', 'name'},
+            'read:users:groups': {'kind', 'name', 'groups'},
+            'read:users:activity': {'kind', 'name', 'last_activity'},
+            'read:users:servers': {'kind', 'name', 'servers'},
+        }
         # Todo: Should 'name' be included in all access?
         self.log.debug(
-            "Asking for models with scopes [%s]" % ", ".join(self.parsed_scopes)
-        )  # debug
-        if 'read:users' in self.parsed_scopes:
-            model.update(
-                {
-                    'name': user.name,
-                    'admin': user.admin,
-                    'roles': [r.name for r in user.roles],
-                    'groups': [g.name for g in user.groups],
-                    'server': user.url if user.running else None,
-                    'pending': None,
-                    'created': isoformat(user.created),
-                    'last_activity': isoformat(user.last_activity),
-                }
-            )
-            server_permission = True
-        else:
-            if 'read:users:name' in self.parsed_scopes:
-                model['name'] = user.name
-            if 'read:users:groups' in self.parsed_scopes:
-                model['groups'] = [g.name for g in user.groups]
-            if 'read:users:activity' in self.parsed_scopes:
-                model['last_activity'] = isoformat(user.last_activity)
-            server_permission = 'read:users:servers' in self.parsed_scopes
-
-        if '' in user.spawners:
+            "Asking for user models with scopes [%s]" % ", ".join(self.raw_scopes)
+        )
+        allowed_keys = set()
+        for scope in access_map:
+            if scope in self.parsed_scopes:
+                scope_filter = self.get_scope_filter(scope)
+                if scope_filter is None or user.name in scope_filter:
+                    allowed_keys |= access_map[scope]
+        model = {key: model[key] for key in allowed_keys if key in model}
+        if not model:
+            return model  # No access to this user
+        if '' in user.spawners and 'pending' in allowed_keys:
             model['pending'] = user.spawners[''].pending
-        if not (include_servers and server_permission):
+        if not (include_servers and 'servers' in allowed_keys):
             model['servers'] = None
-            return model
-
-        servers = model['servers'] = {}
-        for name, spawner in user.spawners.items():
-            # include 'active' servers, not just ready
-            # (this includes pending events)
-            if spawner.active:
-                servers[name] = self.server_model(spawner, include_state=include_state)
+        else:
+            servers = model['servers'] = {}
+            for name, spawner in user.spawners.items():
+                # include 'active' servers, not just ready
+                # (this includes pending events)
+                if spawner.active:
+                    servers[name] = self.server_model(
+                        spawner, include_state=include_state
+                    )
         return model
 
     def group_model(self, group):
--- a/jupyterhub/apihandlers/base.py
+++ b/jupyterhub/apihandlers/base.py
@@ -65,7 +65,7 @@ class APIHandler(BaseHandler):
             return False
         return True
 
-    @functools.lru_cache
+    @functools.lru_cache()
     def get_scope_filter(self, req_scope):
         """Produce a filter for `*ListAPIHandlers* so that GET method knows which models to return.
         Filter is a callable that takes a resource name and outputs true or false"""
--- a/jupyterhub/apihandlers/base.py
+++ b/jupyterhub/apihandlers/base.py
@@ -34,6 +34,7 @@ class APIHandler(BaseHandler):
         self.raw_scopes = set()
         self.parsed_scopes = {}
         self._parse_scopes()
+        # todo: Check if okay
 
     def _parse_scopes(self):
         """Parse raw scope collection into a dict with filters that can be used to resolve API access"""
--- a/jupyterhub/apihandlers/base.py
+++ b/jupyterhub/apihandlers/base.py
@@ -2,6 +2,7 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 import json
+import re
 from datetime import datetime
 from http.client import responses
 
@@ -9,6 +10,7 @@ from sqlalchemy.exc import SQLAlchemyError
 from tornado import web
 
 from .. import orm
+from .. import roles
 from ..handlers import BaseHandler
 from ..handlers import scopes
 from ..utils import isoformat
@@ -24,7 +26,26 @@ class APIHandler(BaseHandler):
     - strict referer checking for Cookie-authenticated requests
     - strict content-security-policy
     - methods for REST API models
+    - scope loading
     """
+
+    async def prepare(self):
+        await super().prepare()
+        self.raw_scopes = set()
+        self.parsed_scopes = {}
+        self._parse_scopes()
+
+    def _parse_scopes(self):
+        """Parse raw scope collection into a dict with filters that can be used to resolve API access"""
+        self.log.debug("Parsing scopes")
+        if self.current_user is not None:
+            self.raw_scopes = roles.get_subscopes(*self.current_user.roles)
+        oauth_token = self.get_current_user_oauth_token()
+        if oauth_token:
+            self.raw_scopes |= scopes.get_user_scopes(oauth_token.name)
+        if 'all' in self.raw_scopes:
+            self.raw_scopes |= scopes.get_user_scopes(self.current_user.name)
+        self.parsed_scopes = scopes.parse_scopes(self.raw_scopes)
 
     @property
     def content_security_policy(self):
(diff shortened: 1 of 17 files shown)


jupyterhub/apihandlers/base.py
```diff
@@ -201,19 +238,39 @@ def user_model(self, user, include_servers=False, include_state=False):
...
-            # (this includes pending events)
-            if spawner.active:
-                servers[name] = self.server_model(spawner, include_state=include_state)
+
+        access_map = {
+            'read:users': set(model.keys()),  # All available components
+            'read:users:names': {'kind', 'name'},
+            'read:users:groups': {'kind', 'name', 'groups'},
+            'read:users:activity': {'kind', 'name', 'last_activity'},
+            'read:users:servers': {'kind', 'name', 'servers'},
```
minrk, line 247: do we have a scope for auth state yet?

0mar:
Updated: Now queries the database only at the initial stage. The results are cached to construct the relevant scope filters later on

jupyterhub/handlers/base.py
```diff
@@ -425,14 +427,35 @@ async def get_current_user(self):
...
+        self._parse_scopes()
+        self._cache_db()
         return self._jupyterhub_user
 
+    def _cache_db(self):
+        self.db_cache = {}
+        for kind in {'users', 'groups', 'services'}:
+            Resource = orm.get_class(kind)
+            resources_names = {
+                resource.name for resource in self.db.query(Resource).all()
```
minrk, line 444: This preemptively fetches *all* resources from the database into memory on every request. I think that will be too expensive.

I think the idea is, within a given request i.e. in `Handler.get()`, not in the base handler preparation), fetch the resources that will be relevant for that request. That should be, e.g. Users (possibly filtered) and fully resolved scopes for the token. Then, when it comes time to do filtering, we can do that with the in-memory objects we have already retrieved. These should all already be available in the resolved_scopes and orm.User object, for instance.


I think the gist should be something like:

```python
def check_scope(resource, need_scopes, have_scopes):
    if it's a user and I have !user=user.name:
        return True
    if it's a group and I have !group=group.name:
       return True
    if it's a user and I have any(!group=group.name for group in user.groups):
       return True
```

None of these construct new queries, and we shouldn't need to fetch users or groups as part of constructing the filter - the filter should only be fetching roles -> scopes, and *applying* the filter, should work on local attributes only. No additional lookups, needed.

jupyterhub/apihandlers/base.py
```diff
@@ -62,6 +64,43 @@ def check_referer(self):
             return False
         return True
 
+    def get_scope_filter(self, req_scope):
```
minrk, line 67: I think scope_filter probably wants to produce a callable instead of a set. That way, it doesn't need to resolve all possible answers before checking values.

jupyterhub/apihandlers/base.py
```diff
@@ -65,10 +86,9 @@ def check_referer(self):
         return True
 
     def get_scope_filter(self, req_scope):
-        """Produce a filter for `*ListAPIHandlers* so that GET method knows which models to return
-        If the APIHandler has unrestricted access
-        """
-        kind_regex = re.compile(r':?(users|services|groups):?')
+        """Produce a filter for `*ListAPIHandlers* so that GET method knows which models to return.
+        Filter is a callable that takes a resource name and outputs true or false"""
+        kind_regex = re.compile(r':?(user|service|group)s:?')
```
minrk, line 91: Minor: to avoid compiling a pattern on every call, we can define the pattern at a higher level - a class attribute or module constant

jupyterhub/apihandlers/base.py
```diff
@@ -62,6 +85,43 @@ def check_referer(self):
             return False
         return True
 
+    def get_scope_filter(self, req_scope):
```
minrk, line 88: We can use [lru_cache()](https://docs.python.org/3/library/functools.html#functools.lru_cache) to memoize this since we call it so many times (once each for a few scopes for each user in user_model) with the same few arguments:

```python
@lru_cache()
def get_scope_filter(self, req_scope):
```

jupyterhub/apihandlers/base.py
```diff
@@ -62,6 +85,43 @@ def check_referer(self):
...
+        try:
+            sub_scope = self.parsed_scopes[req_scope]
+        except AttributeError:
+            raise web.HTTPError(
+                403,
+                "Resource scope %s (that was just accessed) not found in parsed scope model"
+                % req_scope,
+            )
+
+        def has_access(resource_name):
```
minrk, line 110: I think it will be better to pass the actual orm object as `resource` rather than resource_name. That will make things easier. For example, `kind` can be based on the type of `resource` instead of parsing the scope with a regex.

jupyterhub/apihandlers/base.py
```diff
@@ -62,6 +85,43 @@ def check_referer(self):
...
+            if sub_scope == scopes.Scope.ALL:
+                found_resource = True
+            else:
+                found_resource = resource_name in sub_scope[kind]
+            if not found_resource:  # Try group-based access
+                if 'groups' in sub_scope and kind == 'users':
+                    user = self.current_user()
```
minrk, line 117: I think this lookup is incorrect (current_user is the requesting user, not the one we are checking against), but passing `resource` to has_access should fix this.

jupyterhub/apihandlers/base.py
```diff
@@ -201,19 +260,37 @@ def user_model(self, user, include_servers=False, include_state=False):
...
+            if not (include_servers and 'servers' in allowed_keys):
+                model['servers'] = None
+            else:
+                servers = model['servers'] = {}
+                for name, spawner in user.spawners.items():
+                    # include 'active' servers, not just ready
+                    # (this includes pending events)
+                    if spawner.active:
+                        servers[name] = self.server_model(
+                            spawner, include_state=include_state
```
minrk, line 292: For future, this `include_state` is the argument that I think should be resolved by a scope like the rest. Such as `read:user:auth_state` or perhaps `admin:user:auth_state` since it's a highly sensitive field that users shouldn't generally have access to themselves by default. That doesn't need to be in this PR, though. This is one of the things where we are using `user.admin` as a blunt proxy for what should be handled with scopes.

jupyterhub/apihandlers/groups.py
```diff
@@ -35,12 +35,11 @@ def find_group(self, group_name):
...
 class GroupListAPIHandler(_GroupAPIHandler):
     @needs_scope('read:groups')
-    def get(self, scope_filter=None):
+    def get(self):
         """List groups"""
         groups = self.db.query(orm.Group)
-        if scope_filter is not None:
-            groups = groups.filter(orm.Group.name.in_(scope_filter))
-        data = [self.group_model(g) for g in groups]
+        scope_filter = self.get_scope_filter('read:groups')
```
minrk, line 41: 👍

minrk (commented):
Great! I left a few comments that I think should help finish this one, but it's looking very close to ready to me.

jupyterhub/apihandlers/base.py
```diff
@@ -62,6 +65,42 @@ def check_referer(self):
             return False
         return True
 
+    @functools.lru_cache
+    def get_scope_filter(self, req_scope):
+        """Produce a filter for `*ListAPIHandlers* so that GET method knows which models to return.
+        Filter is a callable that takes a resource name and outputs true or false"""
+
+        try:
+            kind = re.search(scopes.kind_regex, req_scope).group(1)
```
minrk, line 74: I thought we were going to move `kind` to inside `has_access` based on `isinstance(orm_resource, ...)`? Then the regex can be dropped.

jupyterhub/apihandlers/base.py
```diff
@@ -201,19 +239,38 @@ def user_model(self, user, include_servers=False, include_state=False):
...
+            'read:users:servers': {'kind', 'name', 'servers'},
+        }
+        self.log.debug(
+            "Asking for user models with scopes [%s]" % ", ".join(self.raw_scopes)
+        )
+        allowed_keys = set()
+        for scope in access_map:
+            if scope in self.parsed_scopes:
+                scope_filter = self.get_scope_filter(scope)
+                if scope_filter(user):
```
minrk, line 256: 👍

jupyterhub/apihandlers/users.py
```diff
@@ -95,14 +101,14 @@ def get(self, scope_filter=None):
...
-            query = query.filter(orm.User.name.in_(scope_filter))
-
-        data = [
-            self.user_model(u, include_servers=True, include_state=True)
-            for u in query
-            if (post_filter is None or post_filter(u))
-        ]
+        data = []
+        for u in query:
+            if post_filter is None or post_filter(u):
```
minrk, line 106: I love that scope_filter no longer shows up here!

jupyterhub/handlers/base.py
```diff
@@ -80,9 +81,11 @@ async def prepare(self):
         The current user (None if not logged in) may be accessed
         via the `self.current_user` property during the handling of any request.
         """
-        self.scopes = set()
+        self.raw_scopes = set()
+        self.parsed_scopes = {}
         try:
             await self.get_current_user()
+            scopes.build_scope_schema(self)
```
minrk, line 88: It feels a tiny bit off to me to pass the handler itself to the module utility function. My general feeling is that the structure should be:

- Handler retrieves 'requester' i.e. the APIToken if token-authenticated, or just the User itself if cookie-authenticated
- that object is passed to the utility, which *returns*
- Handler deals with assigning the results to attributes

If a standalone function takes an Instance as a single argument and performs several operations on that instance, *especially attribute assignment*, that's a good sign to me that it should be a method instead. If you think separating it like I described above would not be an improvement, keeping it as-is and moving it back to a method is AOK, but I like the do like the idea of making it standalone as you have it.

What do you think?

jupyterhub/scopes.py
```diff
@@ -1,18 +1,36 @@
...
+
+
+def build_scope_schema(handler):
+    """Parse raw scope collection into a dict with filters that can be used to resolve API access"""
+    app_log.debug("Parsing scopes")
+    if handler.current_user is not None:
+        handler.raw_scopes = roles.get_subscopes(*handler.current_user.roles)
+    oauth_token = handler.get_current_user_oauth_token()
+    if oauth_token:
+        handler.raw_scopes |= get_user_scopes(oauth_token.name)
```
minrk, line 27: This is performing *unions*, which means that the token can *increase* a user's permissions, which I think we don't want.

I believe the design we have is that a user's tokens can only *limit* permissions, never expand them.

Resolving a request's authorized scopes should look roughly like this:

1. get scopes for the request (APIToken if used, otherwise user); (token -> role -> raw scopes)
2. If requester is a token:
    1. compute *intersection* of token scopes and user scopes (user -> roles -> raw scopes)
    2. Discard any scopes the token has and owner does not
    3. Warn if there were any scopes discarded

If tests are passing, I think we need more tests to ensure that tokens limit scopes, and don't expand them. For example:

1. test that a token with a scope the owner does not have will not actually be able to use the expanded scope
2. test that a token with a limited scope will not be able to use a scope that its owner has that t
3. test that services with the same names as users don't get that user's scope(s) (I see a couple pure string comparisons)

jupyterhub/apihandlers/base.py
0mar: Agree. I'll remove the regex and add kind as a function argument

jupyterhub/handlers/base.py
0mar: I think that makes more sense indeed. I have refactored the logic a bit, now these functions are 'pure' and the handler takes care of the assignments

jupyterhub/scopes.py
0mar: Ah, I didn't take the restrictions into account yet. I'll change the behaviour and add some unit tests, since the ones I have now don't test this behaviour as far as I know.

0mar: This is taken care of now!

jupyterhub/apihandlers/base.py
```diff
@@ -62,6 +65,38 @@ def check_referer(self):
...
+                % req_scope,
+            )
+
+        def has_access(orm_resource, kind):
+            """
+            param orm_resource: User or Service or Group
+            param kind: 'users' or 'services' or 'groups'
+            """
+            if sub_scope == scopes.Scope.ALL:
+                found_resource = True
```
minrk, line 88: I think we can `return True` when we get to any match, rather than continuing to `|=` until the end. There's no need to keep checking once we've met a single positive match.

jupyterhub/apihandlers/base.py
```diff
@@ -201,30 +235,49 @@ def user_model(self, user, include_servers=False, include_state=False):
...
+    def group_model(self, group):  # Todo: make consistent to do scope checking here
         """Get the JSON model for a Group object"""
         return {
             'kind': 'group',
             'name': group.name,
             'users': [u.name for u in group.users],
         }
 
-    def service_model(self, service):
+    def service_model(self, service):  # Todo: make consistent to do scope checking here
```
minrk, line 280: 👍

jupyterhub/handlers/base.py
```diff
@@ -429,10 +434,22 @@ async def get_current_user(self):
...
+        self.raw_scopes = set()
+        app_log.debug("Loading and parsing scopes")
+        api_token = self.get_token()
+        if api_token:
+            self.raw_scopes = scopes.get_scopes_for(api_token)
+        elif self.current_user:
+            self.raw_scopes = scopes.get_scopes_for(self.current_user)
+        else:  # deprecated oauth tokens
+            user_from_oauth = self.get_current_user_oauth_token()
+            self.raw_scopes = scopes.get_scopes_for(user_from_oauth)
```
minrk, line 449: This is where we should hardcode *different* scopes for oauth to preserve current behavior (read:user for just the owner), e.g. `self.raw_scopes = {'read:user': {'user': user_from_oauth.name}`. This here would expand oauth token permissions to match default api token permissions.

After #3380, this will go away, of course.

jupyterhub/roles.py
```diff
@@ -37,19 +37,51 @@ def get_default_roles():
             'description': 'Post activity only',
             'scopes': ['users:activity!user=username'],
         },
+        {
+            'name': 'token',
+            'description': 'Token with same rights as token owner',
+            'scopes': ['all'],
```
minrk, line 43: 👍

jupyterhub/roles.py
```diff
@@ -99,6 +131,15 @@ def expand_subscopes(index):
     return expanded_scope
 
 
+def get_scopes_for(orm_object):
+    """Get the scopes for User/Service/Group/Token"""
+    scopes = get_subscopes(*orm_object.roles)
+    if 'self' in scopes:
```
minrk, line 137: Should probably add a check that `self` is only valid for orm.User

```suggestion
    if 'self' in scopes:
        if not isinstance(orm_object, orm.User):
            raise ValueError(f"'self' scope only valid for Users, got {orm_object}")
```

jupyterhub/tests/utils.py
```diff
@@ -121,7 +121,9 @@ def add_user(db, app=None, **kwargs):
 
 
 def auth_header(db, name):
-    """Return header with user's API authorization token."""
+    """Return header with user's API authorization token.
+    If inherit is True, copies the roles of the invoking user
```
minrk, line 125: inherit?

jupyterhub/scopes.py
```diff
@@ -6,33 +6,36 @@
...
-    scope_list.extend(['read:' + scope for scope in scope_list])
-    return {"{}!user={}".format(scope, name) for scope in scope_list}
+def get_scopes_for(orm_object):
+    """Find scopes for a given user or token"""
+    scopes = set()
+    if orm_object is None:
+        return scopes
+    elif isinstance(orm_object, orm.APIToken):
+        owner = orm_object.user or orm_object.service
+        token_scopes = roles.get_scopes_for(orm_object)
```
minrk, line 23: I think it's a little confusing to have two `get_scopes_for` that do different things. What do the do differently, and can we perhaps get a clearer name for one or the other?

jupyterhub/scopes.py
```diff
@@ -6,33 +6,36 @@
...
+        if 'all' in token_scopes:
+            token_scopes.remove('all')
+            token_scopes |= owner_scopes
+        scopes = token_scopes & owner_scopes
+        discarded_token_scopes = token_scopes - scopes
+        # Not taking symmetric difference here because token owner can naturally have more scopes than token
+        if discarded_token_scopes:
+            app_log.warning(
+                "Token-based access, discarding scopes [%s]"
+                % ", ".join(discarded_token_scopes)
```
minrk, line 34: It's probably important to include the token object's repr itself in the logs, otherwise it will be hard to identify which token is responsible for the warning.

jupyterhub/tests/test_auth_expiry.py
```diff
@@ -101,6 +101,7 @@ async def test_auth_expired_page(app, user, disable_refresh):
     assert user._auth_refreshed == before
 
 
+# Fixme: Why does this text fail?
```
minrk, line 104: This is failing because `BaseHandler.refresh_user()` called in `get_current_user()` refreshes auth and can take a resolved user back to None.

I think the fix is that `_resolve_scopes` should return an empty set if `get_current_user()` is None, even if `get_token()` returns a token.

minrk (commented):
Looking great! I left some comments, but this is very close to ready. I haven't done a thorough read of the tests themselves yet, but I've looked at all the rest of the code.

jupyterhub/apihandlers/base.py
0mar: I think the code does not execute additional branches after getting a match, although the unnecessary `|=` makes it a bit unclear, thanks for pointing that out. I cleaned it up

jupyterhub/scopes.py
0mar: Good point. I renamed one of them

jupyterhub/handlers/base.py
```diff
@@ -439,16 +439,21 @@ async def get_current_user(self):
...
+                user_from_oauth = self.get_current_user_oauth_token()
+                self.raw_scopes = {f'read:users!user={user_from_oauth.name}'}
+            except:
+                pass
+        else:
+            api_token = self.get_token()
+            if api_token:
+                self.raw_scopes = scopes.get_scopes_for(api_token)
+            elif self.current_user:
+                self.raw_scopes = scopes.get_scopes_for(self.current_user)
```
minrk, line 454: Minor revision:

```suggestion
        if not self.current_user:
            # check for oauth tokens as long as #3380 not merged
            user_from_oauth = self.get_current_user_oauth_token()
            if user_from_oauth is not None:
                self.raw_scopes = {f'read:users!user={user_from_oauth.name}'}
            else:
                app_log.debug("No user found, no scopes loaded")
        else:
            api_token = self.get_token()
            if api_token:
                self.raw_scopes = scopes.get_scopes_for(api_token)
            else:
                self.raw_scopes = scopes.get_scopes_for(self.current_user)
```

jupyterhub/scopes.py
```diff
@@ -30,11 +31,11 @@ def get_scopes_for(orm_object):
         # Not taking symmetric difference here because token owner can naturally have more scopes than token
         if discarded_token_scopes:
             app_log.warning(
-                "Token-based access, discarding scopes [%s]"
+                "discarding scopes [%s], not present in owner roles"
                 % ", ".join(discarded_token_scopes)
             )
     else:
-        scopes = roles.get_scopes_for(orm_object)
+        scopes = roles.expand_roles_to_scopes(orm_object)
```
minrk, line 38: Very clear new name!

jupyterhub/apihandlers/base.py
```diff
@@ -62,6 +65,38 @@ def check_referer(self):
             return False
         return True
 
+    @functools.lru_cache
```
minrk, line 68: This is one of those weird decorators where you have to call the function to return the decorator itself:

```suggestion
    @functools.lru_cache()
```

jupyterhub/handlers/base.py
0mar: That makes more sense, thanks.

jupyterhub/apihandlers/base.py
0mar: Oh I missed that. I took the example from here: https://docs.python.org/3/library/functools.html where they don't use parentheses. What is the difference between that situation and this one?

minrk:
Looks great! One real test failure in singleuser_auth, but I think we can hit that one with another PR. Ignore the db upgrade failure, which is already fixed elsewhere.

great job!

PR #3366, opened 2021-02-19
[rbac] Add RBAC documentation with myst-parser

IvanaH8:
Ref: #2245 

In collaboration with @0mar 

1. Replaces `recommonmark`  with `myst-parser` for docs
2. Adds documentation (`docs/source/rbac`) explaining the JupyterHub's RBAC framework (the raw markdown of the documentation can be found [here](https://hackmd.io/QzxDMUUnQqm-YFTdcCGdCA?view) ):

- the motivation behind
- definitions of roles and scopes within JupyterHub
- user guide for working with roles and scopes in JupyterHub
- use case examples
- technical implementation with details where the roles and scopes are resolved
- upgrade guide


docs/source/rbac/index.md
```diff
@@ -0,0 +1,29 @@
+# JupyterHub RBAC
+
+Role Based Access Control (RBAC) in JupyterHub serves to provide finer grained access to perform actions by users or services.
+
+## Motivation
+The JupyterHub API requires authentication before allowing changes to the administration system. For instance, currently the default behaviour is that creating or deleting users requires *admin rights*. This ensures that an arbitrary user, or even an unauthenticated third party, cannot disrupt the status of the Hub.
+
+This system is functional, but lacks flexibility. If your Hub serves a number of users in different departments, you might want to delegate permissions to other users or automate certain processes. With this framework, appointing a 'group-only admin', or a bot that culls idle servers, requires granting full rights to all actions. This can be error-prone and violates the [principle of least privilige](https://en.wikipedia.org/wiki/Principle_of_least_privilege).
```
manics, line 8:
```suggestion
This system is functional, but lacks flexibility. If your Hub serves a number of users in different groups, you might want to delegate permissions to other users or automate certain processes. Appointing a 'group-only admin', or a bot that culls idle servers, requires granting full rights to all actions. This can be error-prone and violates the [principle of least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege).
```

docs/source/rbac/roles.md
```diff
@@ -0,0 +1,60 @@
+# Roles
+JupyterHub provides three **default roles** which are automatically loaded to the database at the startup:
+
+```{admonition} **Default roles**
+- **_user_** role carries single `all` scope that grants _least user access_ to perform only default user actions.
```
manics, line 5: > role carries single `all` scope

I think there's a word missing here, or there's some other grammatical error.
Also I was puzzled by the meaning of `all` until I got to the scopes page, would be nice to cross-reference the relevant section

docs/source/rbac/roles.md
```diff
@@ -0,0 +1,60 @@
...
+- **_admin_** role contains all available scopes and grants full rights to all actions similarly to the current admin status.
+- **_server_** role allows for posting activity only through the single `users:activity` scope.
+```
+
+Roles can also be customly defined and assigned to users, services, groups and tokens. 
+
+**_Users_** and **_services_** are assigned a default role if no custom role is requested based on their admin status.
```
manics, line 12:
```suggestion
**_Users_** and **_services_** are assigned a default role ( **_user_** or **_admin_**)  if no custom role is requested based on their admin status.
```
This suggestion is based on my understanding, if it's wrong then obviously reword as you think best!

docs/source/rbac/roles.md
```diff
@@ -0,0 +1,60 @@
...
+
+**_Tokens_**’ roles cannot grant the token higher permissions than their owner’s roles. If no specific role is requested, tokens are assigned the default _user_ role.
```
manics, line 14: Minor formatting point, is `_user_` intentionally not bold?

docs/source/rbac/roles.md
```diff
@@ -0,0 +1,60 @@
...
+- **_server_** role allows for posting activity only through the single `users:activity` scope.
+```
+
+Roles can also be customly defined and assigned to users, services, groups and tokens. 
+
+**_Users_** and **_services_** are assigned a default role if no custom role is requested based on their admin status.
+
+**_Tokens_**’ roles cannot grant the token higher permissions than their owner’s roles. If no specific role is requested, tokens are assigned the default _user_ role. 
+
+**_Groups_** do not require any role and are not assigned any roles by default. Once group roles are defined and assigned, the privileges of each group member are extended with the group roles in the background during the API request permission check. This is useful for requesting the same permissions for several users.
```
manics, line 16: > in the background

Could you explain what this means? Are a user's privileges the combination of their own role and their group's role, or does `background` imply something more complicated?

docs/source/rbac/roles.md
```diff
@@ -0,0 +1,60 @@
...
+(define_role_target)=
+## Defining Roles
+
+### In `jupyterhub_config.py`
+
+Roles can be defined or modified in the configuration file as a list of dictionaries. An example:
+```python
+c.JupyterHub.load_roles = [
+ {
+   'name': 'Server rights',
```
manics, line 27:
```suggestion
   'name': 'server-rights,
```
The following paragraph says "The role `server-rights` ...", so presumably it's referring to this name?

docs/source/rbac/roles.md
```diff
@@ -0,0 +1,60 @@
...
+   'description': 'Allows parties to start and stop user servers',
+   'scopes': ['users:servers', 'read:users:servers'],
+   'users': ['alice', 'bob'],
+   'services': ['idle-culler'],
+   'groups': ['admin-group'],
+   'tokens': ['foo-6f6e65','bar-74776f']
+ }
+]
+```
+The role `server-rights` now allows the starting and stopping of servers by users `alice` and `bob` and the service `idle-culler`, and any member of the `admin-group` or requests using the tokens `foo-6f6e65`/`bar-74776f` (providing the tokens owner has at least the same permissions).
```
manics, line 37:
```suggestion
The role `server-rights` now allows the starting and stopping of servers by _any_ of the following:
- users `alice` and `bob`
- the service `idle-culler`
- any member of the `admin-group`
- requests using the tokens `foo-6f6e65` or `bar-74776f` (providing the tokens owner has at least the same permissions).
```

I think this makes it a bit clearer than a long sentence with lots of conjunctions, feel free to rewrite/reformat

docs/source/rbac/roles.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+The role `reader` allows users `maria` and `joe` and service `external` to read (but not modify) any user’s model.
+
+```{admonition} Requirements
+:class: warning
+In a role definition, the `name` field is required, while all other fields are optional.
+Moreover, `users`, `services`, `groups` and `tokens` only accept objects that already exist or are defined previously in the file.
+It is not possible to implicitly add a new user to the database by defining a new role.
+```
```
manics, line 58: What happens if you create a role without `scopes`?

docs/source/rbac/scopes.md
```diff
@@ -0,0 +1,37 @@
+# Scopes
+
+A scope has a syntax-based design that reveals which resources it provides access to:
```
manics, line 3: In this section I think it'd be helpful to use different syntax to separate strings which are valid in scopes (e.g. `users`, `admin`) from placeholders like `resource`. E.g. `admin:<resource>`?

docs/source/rbac/scopes.md
```diff
@@ -0,0 +1,37 @@
...
+- The base`resource` scopes, such as `users` or `groups`, provides non-elevated 'default' read or write permissions to the resource itself and all sub-resources. For instance, the scope`users:servers` is included within the scope`users`.
+- The elevated `admin:resource` scopes extend permissions beyond default. For instance, where the scope `users` provides read and write access to user information, the scope `admin:users` allows creating and deleting users.
+- The scope structure `resource:subresource` vertical filtering: it provides access to a subset of the information granted by the `resource` scope. For example, the scope `users:names` provides access to user names only.
+- The scope structure `read:resource` (or `read:resource:subresource`) limits permissions to read-only operations on `resource` (or `subresource`).
+- The scope structure `resource!user=charlie` allows for horizontal filtering: it limits access to only return resources of user `charlie`. Only one filter per scope is allowed, but filters for the same scope have an additive effect; a larger filter can be used by supplying the scope multiple times with different filters.
```
manics, line 8: When I first skimmed though the docs I missed this explanation of what `!` means as the text is quite dense:
image

Since it's introducing new syntax compared to the previous page is there a way to highlight this? A use-case I'm thinking of is someone reading a `jupyterhub_config.py` sees some scopes for the first time. The `aaa:bbb` format is similar to that used in many other projects, but the `!` is less familiar.

Might also be nice to cross-reference "horizontal filtering" or provide a brief in-line explanation since it's new terminology that (at least to me) doesn't have an intuitive meaning.

docs/source/rbac/use-cases.md
```diff
@@ -0,0 +1,98 @@
...
+ },
+ {
+   'name': 'teacher',
+   'description': 'Allows for accessing information about teacher group members and starting/stopping their servers',
+   'scopes': [ 'read:users!group=class-B', 'users:servers!group=class-B'],
+   'users': ['johan']
+ }
+]
+``` 
+In the above example, `johan` has privileges inherited from class-A role and the `teacher` role on top of those.
```
manics, line 96:
```suggestion
In the above example, `johan` has privileges inherited from `class-A role` and the `teacher` role on top of those. 
```

docs/source/rbac/use-cases.md
```diff
@@ -0,0 +1,98 @@
...
+   'name': 'teacher',
+   'description': 'Allows for accessing information about teacher group members and starting/stopping their servers',
+   'scopes': [ 'read:users!group=class-B', 'users:servers!group=class-B'],
+   'users': ['johan']
+ }
+]
+``` 
+In the above example, `johan` has privileges inherited from class-A role and the `teacher` role on top of those. 
+
+Note the filters (`!group=`) limiting the priviliges only to the particular groups. `johan` can access the servers and information of `class-B` members only.
```
manics, line 98:
```suggestion
Note the filters (`!group=`) limiting the privileges only to the particular groups. `johan` can access the servers and information of `class-B` members only.
```

manics (commented):
I'm excited to see this :smile_cat:

consideRatio (commented):
# Review notes

Thank you for your amazing work on this @IvanaH8 and @0mar!!! :heart: :tada:! WOW!!!

I read through the RBAC section at https://jupyterhub--3366.org.readthedocs.build/en/3366/rbac/index.html and made comments while doing it. Below are those comments.

## rbac/index

__JupyterHub RBAC__

- [ ] "finer grained access" should perhaps be "fine grained control of access"?
- [ ] "perform actions" could be clarified in index someplace, will the scopes/roles only be used for a JupyterHub API on /api or are there also /hub/spawn separately or similar? Hmm....

__Motivation__

- [ ] "authentication" should perhaps be "authorization"
- [ ] "cannot disrupt the status of the Hub" was a big unclear, I think the message is that admin actions should be allowed to be taken.
- [ ] "error-prone" wasn't clear to me what it meant

__Definitions__

- Perfect! Excellent flow of the text!

__Technical overview__

- Perfect! Excellent having an index here!

### Feedback summary
- Excellent length of index, not too much text and not too little!
- Excellent red thread in the narrative!
- Excellent starting point and finish!

## rbac/roles

__Roles__

- Nice use of a MyST admonition!
- [ ] Default roles: the user role description confused me - "all scope", "least user access", "default user actions"
- [ ] Default roles: the server role is like "least server access", right?
- [ ] "tokens are assigned the default _user_ role", this means to say that the token are assigned the `user` role right? Strip default from the sentence I think.
- [ ] "Once group roles are defined and assigned", could perhaps be "Once roles are defined and assigned to a group
- [ ] "in the background during the API request permission check", could perhaps be removed as a too technical note?
- [ ] "is useful for requesting the same permissions for several users.", could perhaps become "is useful in order to assign a common set of permissions to several users."

__Defining roles__
- [ ] "(providing the tokens owner has at least the same permissions)" - I don't understand this part
- [ ] There is only one space indentation in the second example.
- [ ] The header `In jupyterhub_config.py` didn't feel so suitable, it could perhaps be a comment part of the code examples instead?

### Technical considerations
- "Moreover, users, services, groups and tokens only accept objects that already exist or are defined previously in the file."

  "only accept" indicates to me that I can't specify a username that have access but isn't yet created, but in practice that would be okay right?
- Does setting `c.JupyterHub.load_roles` imply save_to_database_and_load_roles, which mean that if load_roles is a blank list on next startup, it would start with the previously saved roles?

## rbac/scopes

__Scopes__

- [ ] "The base resource scopes", perhaps could be "the top level scopes"?
- [ ] "provides non-elevated ‘default’ read or write permissions" - I don't understand this part

Perhaps this section could provide some more overview before the bullet list starts listing points. I understand this section as describing the meaning of scope names, where all scopes relates to an _endpoint_ but there are various common _modifiers_ that when combined with an endpoint can compose another valid scope.
UPDATE: Ah scopes doesn't center around endpoints, I was confused. What is a "resource" or "object" by the way? Hmmm... Those kinds of questions could perhaps be overviewed?

__Available scopes__

- [ ] A bit complicated sentence, some revision?

__Standard user scope__

- [ ] "default user scope" vs "standard user scope", there have been some inconsistent language between these.

__Horizontal filtering__ + __Vertical filtering__

The examples provided was very useful!

## rbac/use-cases

__Use cases__

NICE! Excellent information to have!

__User scripting their own access__

- [ ] I was unsure what "scripting" and "marking" refers to and/or means.

__Service to cull idle servers__

- [ ] The code example isn't indented sufficiently to be nested under the second bullet point.
- [ ] `["read:users:servers", "admin:users:servers"]` was passed, is the read scope needed if you also pass admin version?

__Users as group admins/Group admin roles__
- [ ] `c.JupyterHub.roles.append()` and `c.JupyterHub.load_roles()` - what is the difference?
- [ ] Indentation in example

## rbac/tech-implementation

__Requesting API token with specific roles__
- [ ] `config.py` was used instead of the commonly referenced `jupyterhub_config.py`

__Figure 2__

- I didn't understand the "Resolve the scopes" box, isn't the token + api scopes resolved?


### Feedback summary

- I love the flow charts!!!!!!!!!!
- WOW!!! NICE WORK!!!

docs/source/rbac/tech-implementation.md
```diff
@@ -0,0 +1,49 @@
...
+API tokens grant access to JupyterHub's APIs. The RBAC framework allows for requesting tokens with specific existing roles. To date, it is only possible to add roles to a token in two ways: 
+1. through the `config.py` file as described in the {ref}`define_role_target` section 
+2. through the POST /users/{name}/tokens API where the roles can be specified in the token parameters body (see [](../reference/rest-api.rst)).
+
+The RBAC framework adds several steps into the token issue flow.
+
+If no roles are requested, the token is issued with a default role (providing the requester is allowed to create the token).
+
+If the token is requested with a specific role or multiple roles, the permissions of the would-be token's owner are checked against the requested permissions to ensure the token will not grant its owner additional privileges. In practice, this corresponds to resolving all the token owner's roles (including the roles associated with their groups) and the token's requested roles into a set of scopes. The two sets are compared and if the token's scopes are a subset of the token owner's scopes, the token is issued with the requested roles.
+
```
eliaswimmer, line 18: I think this is the wrong place for my comment, but I did not find the origin pull request for this feature.

Granting a token access to arbitrary roles matching the users and groups scopes is dangerous, as user roles and thus scopes might change, therefore I think it will be wise to add a relation to users to tokens, and compare the scopes at runtime.

eliaswimmer (commented):
I am really looking forward to RBAC in JupyterHub!

docs/source/rbac/tech-implementation.md
IvanaH8: Very good point, thank you. 

We realized that the token owner's roles/scopes can change after issuing tokens. Additional permissions check is implemented when the token is passed to requests. Instead of assessing only the token's scopes, the owner's scopes are checked again to make sure that the owner can still use the token (ref: #3363). 

The docs will provide more info about that as well.

minrk: Yes, token scopes are always evaluated to the *intersection* of scopes on the token and those on the owner at request time, to ensure that a token never has more permissions than its owner if the owner's permissions are ever reduced in the future.

This paragraph only notes that any *attempt* to expand permissions at token creation time will be met with an error, rather than a token that silently doesn't actually have the requested permissions due to a limitation on the owner.

A clarification of that fact would be appropriate.

An `all` metascope preserves the current default behavior of "this token has all its owner's permissions" evaluated at request time.

docs/source/rbac/roles.md
IvanaH8: The user role has been rephrased and cross-referenced with the default user scope in the Scopes section.

docs/source/rbac/roles.md
IvanaH8: No it wasn't, thank you for noticing. All the role names are now rather in code syntax to distinguish them better from objects.

docs/source/rbac/roles.md
IvanaH8: The  text was rephrased to "..., the privileges of each group member are extended when needed (see ref for more details)" and reference to the relevant section added.

docs/source/rbac/scopes.md
IvanaH8: The whole section is now reorganized to highlight the different scope syntax with added cross-referencing of filtering sections.

IvanaH8:
@consideRatio thank you for the review!
 
We have implemented your suggestions/addressed your comments as follows:
> **JupyterHub RBAC**
> 
> * [x]  "finer grained access" should perhaps be "fine grained control of access"?

Changed.

> * [x]  "perform actions" could be clarified in index someplace, will the scopes/roles only be used for a JupyterHub API on /api or are there also /hub/spawn separately or similar? Hmm....

The roles and scopes will be used for APIs, and the text have been changed to "... access to JupyterHub's API resources."

> **Motivation**
> 
> * [x]  "authentication" should perhaps be "authorization"

Thank you for spotting this, changed to authorization.

> * [x]  "cannot disrupt the status of the Hub" was a big unclear, I think the message is that admin actions should be allowed to be taken.

Rephrased to " ... are not allowed to perform such actions."

> * [x]  "error-prone" wasn't clear to me what it meant

Rephrased to "This poses a risk of the user or service intentionally or unintentionally accessing and modifying any data within the Hub ..."

> **Roles**
>
> * [x]  Default roles: the user role description confused me - "all scope", "least user access", "default user actions"

The user role has been rephrased to: `user` role provides a default user scope {ref} `all` that grants access to only the user's own resources.

> * [x]  Default roles: the server role is like "least server access", right?

Yes.

> * [x]  "tokens are assigned the default _user_ role", this means to say that the token are assigned the `user` role right? Strip default from the sentence I think.

Stripped.

> * [x]  "Once group roles are defined and assigned", could perhaps be "Once roles are defined and assigned to a group

Rephrased.

> * [x]  "in the background during the API request permission check", could perhaps be removed as a too technical note?

This has been rephrased to: "..., the privileges of each group member are extended when needed (see {ref} for more details)."

> * [x]  "is useful for requesting the same permissions for several users.", could perhaps become "is useful in order to assign a common set of permissions to several users."

Rephrased.

> **Defining roles**
> 
> * [x]  "(providing the tokens owner has at least the same permissions)" - I don't understand this part

This has been removed and instead a note has been added with an example and cross-reference to Figure 1.

> * [x]  There is only one space indentation in the second example.

Fixed.

> * [x]  The header `In jupyterhub_config.py` didn't feel so suitable, it could perhaps be a comment part of the code examples instead?

We are thinking of having the option to define roles through API as well in the future so the idea was to have the two ways of definition separate. But for now we moved it to the code block.

> ### Technical considerations
> * "Moreover, users, services, groups and tokens only accept objects that already exist or are defined previously in the file."
>   "only accept" indicates to me that I can't specify a username that have access but isn't yet created, but in practice that would be okay right?

When assigning the roles, the code checks if the username exists in the database. If not, it throws an error. This is to prevent for example misspelled usernames being used to create "new" non-existing user.

> * Does setting `c.JupyterHub.load_roles` imply save_to_database_and_load_roles, which mean that if load_roles is a blank list on next startup, it would start with the previously saved roles?

Yes, it has now been added to the text:
"Once a role is cd loaded, it will remain in the database until explicitly deleting it through `remove_role` function in `roles.py`. I.e., omitting the `c.JupyterHub.load_roles` or specifying different roles in `jupyterhub_config.py` file on the next startup will not erase previously defined roles."

> **Scopes**
> 
> * [ ]  "The base resource scopes", perhaps could be "the top level scopes"?
> * [ ]  "provides non-elevated ‘default’ read or write permissions" - I don't understand this part
> 
> Perhaps this section could provide some more overview before the bullet list starts listing points. I understand this section as describing the meaning of scope names, where all scopes relates to an _endpoint_ but there are various common _modifiers_ that when combined with an endpoint can compose another valid scope.
> UPDATE: Ah scopes doesn't center around endpoints, I was confused. What is a "resource" or "object" by the way? Hmmm... Those kinds of questions could perhaps be overviewed?

The Scope section now starts with a little intro explaining what resource refers to. The bullet point list has been redesigned and rephrased to highlight the syntax differences.

> **Available scopes**
> 
> * [x]  A bit complicated sentence, some revision?

Rephrased. 

> **Standard user scope**
> 
> * [x]  "default user scope" vs "standard user scope", there have been some inconsistent language between these.

Fixed to remain consistent "default user scope" across the docs.

> **User scripting their own access**
> 
> * [x]  I was unsure what "scripting" and "marking" refers to and/or means.

The title has been changed to 'User access' and the example to "(e.g., when a teacher wants to check their notebooks)".

> **Service to cull idle servers**
> 
> * [x]  The code example isn't indented sufficiently to be nested under the second bullet point.

The code block has been realigned.

> * [x]  `["read:users:servers", "admin:users:servers"]` was passed, is the read scope needed if you also pass admin version?

The scopes have been corrected for the idle-culler service. Note that the admin:<resource> scopes grants only permission to create/delete the resources and do not include read and write permissions. This has been now highligted in the Scopes section.

> **Users as group admins/Group admin roles**
> 
> * [x]  `c.JupyterHub.roles.append()` and `c.JupyterHub.load_roles()` - what is the difference?

The `.append()` appends more roles to roles previously defined by the ` .load_roles()`.  As the previously defined roles do not get erased on the next startup, the ` .load_roles()` can be also used to append. To be consistent we will use only `c.JupyterHub.load_roles()` in docs.

> * [x]  Indentation in example

Fixed.

> ## rbac/tech-implementation
> **Requesting API token with specific roles**
> 
> * [x]  `config.py` was used instead of the commonly referenced `jupyterhub_config.py`

Fixed.

> **Figure 2**
> 
> * I didn't understand the "Resolve the scopes" box, isn't the token + api scopes resolved?

The definition of Resolving roles and Resolving scopes has been added to the beginning of the section.

consideRatio:
Really excited about your work on this @IvanaH8 and @0mar!! Thank you for the care and effort you put into this documentation it looks great!

IvanaH8:
The docs have been updated substantially as we added metascopes and few other scopes, default role for tokens as well and updated how and where resolving the roles/scopes occurs.

Upgrading JupyterHub with RBAC section has been added, which needs the most review (@0mar can you also have a look?).

docs/rest-api.yml
```diff
@@ -173,9 +183,12 @@ paths:
...
       security:
         - oauth2:
-            - users
             - read:users
-            - read:users!user=username
+            - read:users:name
+            - read:users:groups
+            - read:users:activity
+            - read:users:servers
+            #TODO: add admin:users:auth_state/server_state?
```
0mar, line 191: Yes, I think those belong here as well

docs/rest-api.yml
```diff
@@ -208,7 +221,8 @@ paths:
       description: Change a user's name or admin status
       security:
         - oauth2:
-            - users
+            - admin:users
+            #TODO: confirm, this scope allows to also create/delete users but users scope would allow any user to change their own name and admin status
```
0mar, line 225: I agree with `admin:users` here

docs/rest-api.yml
```diff
@@ -513,8 +522,12 @@ paths:
...
       security:
         - oauth2:
-            - all
-            - read:all
+            - read:users
+            - read:users:name
+            - read:users:groups
+            - read:users:activity
+            - read:users:servers
+            #TODO: add admin:users:auth_state/server_state?
```
0mar, line 530: Same here I think

docs/source/rbac/upgrade.md
```diff
@@ -0,0 +1,54 @@
...
+
+1. All running **servers must be stopped** before proceeding with the upgrade.
+2. To upgrade the Hub, follow the [Upgrading JupyterHub](../admin/upgrading.rst) instructions.
+   ```{attention}
+   We advise against defining any new roles in the `jupyterhub.config.py` file right after the upgrade is completed and JupyterHub restarted for the first time. This preserves the 'current' state of the Hub. You can define and assign new roles on any other following startup.
+   ```
+3. After restarting the Hub **re-issue all tokens that were previously issued manually** (i.e., not through the `jupyterhub_config.py` file).
+
+When the JupyterHub is restarted for the first time after the upgrade, all users, services and tokens stored in the database or re-loaded through the configuration file will be assigned their default role. Any newly added entities after that will be assigned their default role only if no other specific role is requested for them.
+
```
0mar, line 21: Do we assign the admin role to all users that have an admin flag?

0mar (commented):
Looks great! Left some comments but overall I think it's pretty descriptive

docs/rest-api.yml
IvanaH8: Ok, thanks, will add them in.

docs/rest-api.yml
IvanaH8: Great, thanks!

docs/source/rbac/upgrade.md
IvanaH8: Yes, same for services.

minrk:
Great job, @IvanaH8 and @0mar! I added some minor edits and todos, and now we can merge and make finer refinements.

PR #3370, opened 2021-03-02
Always raise on failed token creation

### LINKED ISSUE #3367: c.JupyterHub.services.*.api_token has a minimum length required?
### Bug description

When configuring a JupyterHub service with an api_token that is 7 characters (ddd4444) instead of 8 characters (dddd4444) or more, the JupyterHub responds with 403 when it otherwise don't.

I expected to get some warning or early error if it had a minimum length, rather than a runtime failure 403 response to REST API requests.

### How to reproduce

This was discovered in the JupyterHub Helm chart's CI system where making the dev-config.yaml specify `hub.services.test.apiToken=ddd4444` led to 403 errors accessing the REST api while it would work fine if set to an eight characters long token.

The JupyterHub configuration to reproduce without z2jh may be...

```python
c.JupyterHub.services = [
    {
        "name": test,
        "admin": true,
        "api_token": "ddd4444",
    }
]
```

And then sending a Rest API request with the oken.

### Setup

JupyterHub 1.3.0 was used in z2jh at this point in time.

minrk:
the logic was there but at the wrong indentation level causing it to only raise sometimes, when it should always raise.

closes #3367, which creates a token that fails [check_token](https://github.com/jupyterhub/jupyterhub/blob/1.3.0/jupyterhub/orm.py#L394).

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1876,7 +1876,7 @@ class JupyterHub(Application):
                         # don't allow bad tokens to create users
                         db.delete(obj)
                         db.commit()
-                        raise
+                    raise
             else:
                 self.log.debug("Not duplicating token %s", orm_token)
         db.commit()


consideRatio (approved):
Nice LGTM!

PR #3373, opened 2021-03-04
allow the hub to not be the default route

minrk:
Rare, but can make sense for api-only deployments where the JupyterHub errors can be confusing (e.g. users of mybinder.org can't login and restart their server, for instance)

Should we go so far as to have 'api-only' as an option,
only serving API Handlers? Not sure.

### DIFF
--- a/jupyterhub/handlers/pages.py
+++ b/jupyterhub/handlers/pages.py
@@ -678,4 +678,5 @@ default_handlers = [
     (r'/token', TokenPageHandler),
     (r'/error/(\d+)', ProxyErrorHandler),
     (r'/health$', HealthCheckHandler),
+    (r'/api/health$', HealthCheckHandler),
 ]
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -858,6 +858,67 @@ class JupyterHub(Application):
     def _hub_prefix_default(self):
         return url_path_join(self.base_url, '/hub/')
 
+    hub_routespec = Unicode(
+        "/",
+        help="""
+        The routing prefix for the Hub itself.
+        
+        Override to send only a subset of traffic to the Hub.
+        Default is to use the Hub as the default route for all requests.
+
+        This is necessary for normal jupyterhub operation,
+        as the Hub must receive requests for e.g. `/user/:name`
+        when the user's server is not running.
+
+        However, some deployments using only the JupyterHub API
+        may want to handle these events themselves,
+        in which case they can register their own default target with the proxy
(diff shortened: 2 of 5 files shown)


yuvipanda:
No reason to not have this, and makes the hub easier to be used 'headless' - like with Thebe, for example.

yuvipanda:
Is there a way for us to add a test for this?

minrk:
I'll try to think of what things could be messed up by this, and add a test or two.

minrk:
What do you think about the 'api-only' version, skipping `/hub` and only having `/hub/api`?

yuvipanda:
So what happens if they hit /hub? 404? Redirect to /hub/api?

minrk:
> So what happens if they hit /hub? 

There would be no route for it in the proxy *registered by the Hub*, so if no other route is added, it would be the default 404 page served by CHP. But the goal would be to *allow* that override, which is hard right now.

An example application would be mybinder.org, where we have a [custom 404 page](https://hub.gke.mybinder.org/hub/fourohfoured) served by nginx. Any time a user of mybinder.org sees an actual jupyterhub page, something's not right, and jupyterhub's own pages are not at all informative due to the fact that the users seeing those pages have no reason to know or care that jupyterhub has anything to do with mybinder.org. We have various configurations and patches to try to make that not happen, but it would be easier to avoid routing those requests to the hub in the first place.

In mybinder.org's case, I would like to serve:

- `/hub/api` only from the hub (need to deal with `/hub/status` not being on `/hub/api/...`)
- `/user/...` as usual, mapping to running users (nothing to do)
- everything else goes to custom 404 page (via CHP's default route)

In this case, that would mean only changing the routespec for the hub to be `/hub/api` from the currently hardcoded `/` and that's it.

jupyterhub/handlers/pages.py
```diff
@@ -678,4 +678,5 @@ def get(self):
     (r'/token', TokenPageHandler),
     (r'/error/(\d+)', ProxyErrorHandler),
     (r'/health$', HealthCheckHandler),
+    (r'/api/health$', HealthCheckHandler),
```
minrk, line 681: Adding the `/api/` prefix to health and metrics is required for api-only operation. These should probably have been on `/api/` all along, but it never really had a consequence until now.

minrk:
@yuvipanda I realized this hasn't landed and I think it might be nice in 1.4

yuvipanda:
Done! Thanks, @minrk!

PR #3377, opened 2021-03-09
always start redirect count at 1 when redirecting /hub/user/:name -> /user/:name

### LINKED ISSUE #3378: Infinite redirect without backoff after accessing /spawn-pending/:user page
### Bug description

The user redirect counter is not being set if redirected from the `/spawn-pending/:user` page, which causes a redirect loop without exponential backoff, which causes Chrome to show the "too many redirects" error page

The bug seems to happen here: https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/handlers/base.py#L1590
When directed from the the `/spawn-pending/:user` page, the Referer header is set to this URL. It never changes.
Is there a reason why the `elif` clause exists? I think a simple fix would be to just replace it with an `else` clause.

I did a quick search it seems like the behavior of the Referer header for 302s is not defined:
https://stackoverflow.com/questions/2158283/will-a-302-redirect-maintain-the-referer-string
This behavior could differ between browsers.

#### Expected behaviour

The browser should redirect with exponential backoff to prevent "too many redirects"

#### Actual behaviour

The browser (Chrome) gets caught in a redirect loop with backoff, which causes it to show the "too many redirects" error page

### How to reproduce

This error is tricky to reproduce because it depends on timing. The reason I am able to see this problem is because I use KubeSpawner to create notebook instances. KubeSpawner creates Kubernetes resources that are set up asynchronously by controllers, and the ingress resource for the notebook is set up by NGINX ingress controller. This means that the spawner could "finish" creating the notebook but the notebook's ingress may not be ready yet, which will cause /hub/spawn-pending/:user to redirect to /hub/user/:user instead of. /user/:user.

1. Log in to JupyterHub and spawn a notebook
2. After the progress bar fills up, you will be redirected to /user/:user which will redirect to /hub/user/:user
3. These redirects will happen rapidly until either your browser gives up, or the notebook's route finishes setting up
4. Either way, if you check Chrome's network tab, you will notice that you went through a redirect loop

### Your personal set up

Latest zero-to-jupyterhub release

 - OS:
 
ubuntu 20.04
 - Version(s):
 
Jupyterhub 1.3.0 Python 3.8.5

- <details><summary>Full environment</summary>

```
# paste output of `pip freeze` or `conda list` here
alembic==1.4.3
async-generator==1.10
attrs==20.3.0
bcrypt==3.2.0
cachetools==4.1.1
certifi==2020.12.5
certipy==0.1.3
cffi==1.14.4
chardet==3.0.4
cryptography==3.2.1
entrypoints==0.3
escapism==1.0.1
google-auth==1.23.0
idna==2.10
ipython-genutils==0.2.0
Jinja2==2.11.2
jsonschema==3.2.0
jupyter-telemetry==0.1.0
jupyterhub==1.3.0
jupyterhub-firstuseauthenticator==0.14.1
jupyterhub-hmacauthenticator==1.0
jupyterhub-idle-culler==1.0
jupyterhub-kubespawner==0.15.0
jupyterhub-ldapauthenticator==1.3.2
jupyterhub-ltiauthenticator==1.0.0
jupyterhub-nativeauthenticator==0.0.7
jupyterhub-tmpauthenticator==0.6
kubernetes==12.0.1
ldap3==2.8.1
Mako==1.1.3
MarkupSafe==1.1.1
mwoauth==0.3.7
nullauthenticator==1.0.0
oauthenticator==0.12.3
oauthlib==3.1.0
onetimepass==1.0.1
pamela==1.0.0
prometheus-client==0.9.0
psycopg2-binary==2.8.6
py-spy==0.3.3
pyasn1==0.4.8
pyasn1-modules==0.2.8
pycparser==2.20
pycurl==7.43.0.6
PyJWT==1.7.1
PyMySQL==0.10.1
pyOpenSSL==20.0.0
pyrsistent==0.17.3
python-dateutil==2.8.1
python-editor==1.0.4
python-json-logger==2.0.1
python-slugify==4.0.1
PyYAML==5.3.1
requests==2.25.0
requests-oauthlib==1.3.0
rsa==4.6
ruamel.yaml==0.16.12
ruamel.yaml.clib==0.2.2
six==1.15.0
SQLAlchemy==1.3.20
statsd==3.3.0
text-unidecode==1.3
tornado==6.1
traitlets==5.0.5
urllib3==1.26.2
websocket-client==0.57.0
```
</details>

- <details><summary>Configuration</summary>

kubespawner 0.15.0
</details>

- <details><summary>Logs</summary>

N/A
</details>

minrk:
`/hub/user/:name` is now only reasonably visited as a result of redirect from `/user/:name`

The multi-redirect spawn-pending behavior means that the referer check was unreliable, so the exponential backoff counter would sometimes not start.

closes #3378

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -1575,7 +1575,6 @@ class UserUrlHandler(BaseHandler):
         if self.subdomain_host:
             target = user.host + target
 
-        referer = self.request.headers.get('Referer', '')
         # record redirect count in query parameter
         if redirects:
             self.log.warning("Redirect loop detected on %s", self.request.uri)
@@ -1587,8 +1586,12 @@ class UserUrlHandler(BaseHandler):
             query_parts['redirects'] = redirects + 1
             url_parts = url_parts._replace(query=urlencode(query_parts, doseq=True))
             target = urlunparse(url_parts)
-        elif '/user/{}'.format(user.name) in referer or not referer:
-            # add first counter only if it's a redirect from /user/:name -> /hub/user/:name
+        else:
+            # Start redirect counter.
+            # This should only occur for redirects from /user/:name -> /hub/user/:name
+            # when the corresponding server is already ready.
+            # We don't check this explicitly (direct visits to /hub/user are technically possible),
+            # but that's now the only normal way to get here.
             target = url_concat(target, {'redirects': 1})
 
         self.redirect(target)


consideRatio:
## Rubber duck review

### My understanding of the change

1. `/hub/user` have a registered handler in JupyterHub called the UserHandler
2. `_redirect_to_user_server` is called from the UserHandler when a user is found with a running server.
3. `_redirect_to_user_server` redirects from `/hub/user(/:user)` to `/user/:user`.
4. `_redirect_to_user_server` now always adds a `redirect=1` instead of doing it only if the `Referer` header was set with the request, as indicating it was arriving from a redirect from the JupyterHub controlled proxy.

Okay I think the change is fine.

### My understanding of the bugfix

The bug was that using KubeIngressProxy with some unknown ingress controller, the check to set `redirects=1` was too rigorous and could fail.

### Conclusion

Okay! Sounds good, merging!

PR #3380, opened 2021-03-10
Merge OAuth and API tokens

minrk:
With RBAC attaching scopes to tokens, we no longer need a distinction between OAuth tokens and other API tokens. Instead, the goal of separate OAuth tokens is achieved by a

- [x] remove OAuthAccessToken
- [x] merge oauth-related columns from OAuthAccessToken into APIToken
- [x] create oauth client 'jupyterhub' which owns tokens issued by the hub itself
- [ ] set appropriate scopes on oauth-issued tokens to match current behavior of limited permissions (read:user:name, read:user:groups)
- [ ] make access token scopes configurable?

db upgrade is currently achieved by dropping both token tables, forcing issuing of fresh tokens on next hub start. Migration will be very complicated if we want to do it (filling out default scopes, etc.), but without it, this will be our most disruptive upgrade yet.

### DIFF
--- a/ci/init-db.sh
+++ b/ci/init-db.sh
@@ -20,7 +20,7 @@ fi
 
 # Configure a set of databases in the database server for upgrade tests
 set -x
-for SUFFIX in '' _upgrade_072 _upgrade_081 _upgrade_094; do
+for SUFFIX in '' _upgrade_100 _upgrade_122 _upgrade_130; do
     $SQL_CLIENT "DROP DATABASE jupyterhub${SUFFIX};" 2>/dev/null || true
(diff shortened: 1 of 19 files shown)


minrk:
@0mar @IvanaH8 I think this is now ready if you could do some review. The main thing this changes is the removal of OAuthAccessTokens, in favor of APITokens with *no roles*.

The next step, in a subsequent PR, should be to implement a 'requested_roles' property on OAuth clients so that the OAuth process can issue tokens with one (or more?) roles.

Either that, *or* follow oauth more conventionally and allow oauth to request scopes directly (this would mean

0mar:
Looks great! I only have a minor comment: the identify scope `read:users:admin`, where would that be used?

PR #3381, opened 2021-03-11
Stop specifying `--ip` and `--port` on the command-line

minrk:
JUPYTERHUB_SERVICE_URL env is already enough and has been around for some time

Specifying CLI args can cause some issues for custom commands

closes jupyterhub/dockerspawner#319

### DIFF
--- a/jupyterhub/singleuser/mixins.py
+++ b/jupyterhub/singleuser/mixins.py
@@ -402,7 +402,7 @@ class SingleUserNotebookAppMixin(Configurable):
         return value
 
     @default('log_level')
-    def _log_level_defaul(self):
+    def _log_level_default(self):
         if _bool_env("JUPYTERHUB_DEBUG"):
             return logging.DEBUG
         else:
--- a/jupyterhub/singleuser/mixins.py
+++ b/jupyterhub/singleuser/mixins.py
@@ -10,6 +10,7 @@ with JupyterHub authentication mixins enabled.
 # Distributed under the terms of the Modified BSD License.
 import asyncio
 import json
+import logging
 import os
 import random
 import warnings
@@ -46,6 +47,17 @@ from ..utils import exponential_backoff
 from ..utils import isoformat
 from ..utils import make_ssl_context
 from ..utils import url_path_join
+
+
+def _bool_env(key):
+    """Cast an environment variable to bool
+
+    0, empty, or unset is False; All other values are True.
+    """
+    if os.environ.get(key, "") in {"", "0"}:
+        return False
+    else:
+        return True
 
 
 # Authenticate requests with the Hub
@@ -268,6 +280,10 @@ class SingleUserNotebookAppMixin(Configurable):
     def _user_changed(self, change):
         self.log.name = change.new
 
+    @default("default_url")
+    def _default_url(self):
+        return os.environ.get("JUPYTERHUB_DEFAULT_URL", "/tree/")
+
     hub_host = Unicode().tag(config=True)
 
     hub_prefix = Unicode('/hub/').tag(config=True)
@@ -350,7 +366,26 @@ class SingleUserNotebookAppMixin(Configurable):
         """,
     ).tag(config=True)
 
-    @validate('notebook_dir')
+    @default("disable_user_config")
+    def _default_disable_user_config(self):
+        return _bool_env("JUPYTERHUB_DISABLE_USER_CONFIG")
+
+    @default("root_dir")
+    def _default_root_dir(self):
+        if os.environ.get("JUPYTERHUB_ROOT_DIR"):
+            proposal = {"value": os.environ["JUPYTERHUB_ROOT_DIR"]}
+            # explicitly call validator, not called on default values
+            return self._notebook_dir_validate(proposal)
+        else:
+            return os.getcwd()
+
+    # notebook_dir is used by the classic notebook server
+    # root_dir is the future in jupyter server
+    @default("notebook_dir")
+    def _default_notebook_dir(self):
+        return self._default_root_dir()
+
+    @validate("notebook_dir", "root_dir")
     def _notebook_dir_validate(self, proposal):
         value = os.path.expanduser(proposal['value'])
         # Strip any trailing slashes
@@ -365,6 +400,13 @@ class SingleUserNotebookAppMixin(Configurable):
         if not os.path.isdir(value):
             raise TraitError("No such notebook dir: %r" % value)
         return value
+
+    @default('log_level')
+    def _log_level_defaul(self):
+        if _bool_env("JUPYTERHUB_DEBUG"):
+            return logging.DEBUG
+        else:
+            return logging.INFO
 
     @default('log_datefmt')
     def _log_datefmt_default(self):
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -999,17 +999,6 @@ class Spawner(LoggingConfigurable):
         """
         args = []
 
-        if self.ip:
-            args.append('--ip=%s' % _quote_safe(self.ip))
-
-        if self.port:
-            args.append('--port=%i' % self.port)
-        elif self.server and self.server.port:
-            self.log.warning(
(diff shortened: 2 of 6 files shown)


consideRatio:
The test failure seem related to

An assumption of ...|Broken by ...
-|-
image|image

minrk:
Nice! I think it's the other way around, though. I suspect it's a real failures, as 202 Accepted means spawning is taking a long time (probably because it didn't actually work), while 201 Created means it's finished and ready.

Now that I think of this, some custom deployments like mybinder.org launching classic notebook server instead of jupyterhub-singleuser will probalby be broken by this, since they won't handle `$JUPYTERHUB_SERVICE_URL` env.

Maybe safer to leave these in...

minrk:
This is a breaking change, since anything that *assumes* `--ip`, `--port` will be passed would need to handle $JUPYTERHUB_SERVICE_URL instead, *or* specify `--ip` and friends in custom `get_args()`, but it is a small one, so it shouldn't be too hard to work around, and it would be a big help for things like https://github.com/jupyterhub/kubespawner/issues/493 so maybe it will be good to aim for this in JupyterHub 2.0: **JupyterHub 2.0 will not add anything to the CLI except for user-configured Spawner.cmd and user-configured Spawner.args**.

I also assumed Binder would be one of the broken cases because it launches `jupyter notebook` instead of `jupyterhub-singleuser`, but it isn't because BinderHub itself [already overrides the default CLI args](https://github.com/jupyterhub/binderhub/blob/0b4462c0b1fb15e355b32b38b99c60f440666e94/helm-chart/binderhub/values.yaml#L86-L92) so changing the defaults doesn't change what happens there.

wdyt @manics @consideRatio?

jupyterhub/spawner.py
```diff
@@ -797,8 +797,27 @@ def get_env(self):
...
+        if self.ip or self.port:
+            # specify JUPYTERHUB_SERVICE_URL *if* ip or port is specified
+            # TODO: set this always?
+            # Prior to 2.0, leaving 'ip' or 'port' unset meant letting the subprocess default be used
+            # setting via a URL means we cannot specify explicit ip or port without specifying *both*
+            # this results in a changed default behavior of specifying only report
+            # specifying ip='' *explicitly*, which is the same as all interfaces, instead of localhost
+            s = 's' if self.internal_ssl else ''
+            bind_url = f"http{s}://{self.ip}:{self.port}{base_url}"
+            env["JUPYTERHUB_SERVICE_URL"] = bind_url
```
minrk, line 820: This is the remaining question mark, I think: Current behavior is to specify ip or port only *if defined*, leaving it to the subprocess to pick if they are not defined. That means a default of `127.0.0.1:8888` in most circumstances, whereas the defaults on the Spawner are `:0`, which is random port on all interfaces, if passed explicitly. Local spawners explicitly have `if self.port == 0: self.port = random_port()` in `start`, but the *base* Spawner class does not.

However, specifying via URL means it's not possible to specify only ip or port and leave the other as default: both must be specified. Further, `ip=''` has a meaning, which is all interfaces, not 'use the default'. Same for port: `0` means random port, not some default. That means specifying only the port (as was done in LocalProcessSpawner) would result in changing the default behavior from listening on localhost only to all interfaces.

Broad categories of ip/port config patterns:

1. local spawners (localprocess, sudo, systemd) where it makes sense to use `127.0.0.1:0` with `:0` handled in the Spawner, and transformed to an [explicit port](https://github.com/jupyterhub/jupyterhub/blob/574d34388135cceb6c3a633038982529c1b18309/jupyterhub/spawner.py#L1459) prior to launch
2. isolated spawners (docker, k8s, remote VM), where it makes sense to use fixed, explicit `0.0.0.0:8888` (docker + host networking is a wrinkle here, where it should actually do the same as LocalProcessSpawner)
3. subprocess actually does decide (batchspawner, #2727), where it really is reasonable to give the subprocess `:0` and use some special Spawner machinery to report back what host/port were selected and how to connect to it

Possible solutions:

- keep current behavior, where JUPYTERHUB_SERVICE_URL is only defined if one or both of ip and port is specified. The weird behavior: **setting port changes ip from default localhost to all interfaces** which makes me think we should do something else
- make default ip 127.0.0.1 by default, matching current *effective* defaults. This would also mean JUPYTERHUB_SERVICE_URL is always set (by default, at least), which means the singleuser server's internally determined defaults would ~never be used
- always pass JUPYTERHUB_SERVICE_URL, remove case where JUPYTERHUB_SERVICE_URL can be skipped, changing the interpretaion of `ip=''` from "unset" to "specified all interfaces". The default can still be either localhost or all interfaces.


My inclination is to set the default ip, port on the base Spawner class to match the local spawners `127.0.0.1:0` and always pass this as JUPYTERHUB_SERVICE_URL. Then:

- local spawners need to map `if self.port == 0: self.port = random_port()`, which they already do. They do not tend to override default ip, as the existing localhost default works for them.
- container spawners need to specify defaults for both ip, port, which they already do
- if ip is ever specified as `''` this will be explicitly passed and interpreted as 'all interfaces'. This never happens, currently

I think this is the most consistent, backward-compatible approach, where most existing Spawners would need no changes at all, but I'm open to other ideas.

jupyterhub/singleuser/mixins.py
```diff
@@ -366,6 +401,13 @@ def _notebook_dir_validate(self, proposal):
             raise TraitError("No such notebook dir: %r" % value)
         return value
 
+    @default('log_level')
+    def _log_level_defaul(self):
```
manics, line 405:
```suggestion
    def _default_log_level(self):
```
Might as well match the naming convention unless there's areason to make it different?

jupyterhub/spawner.py
manics: I don't have enough knowledge of this at the moment, but I think always setting `JUPYTERHUB_SERVICE_URL` makes sense.

manics (commented):
👍 for getting rid of mandatory command line arguments. I think it'll help with future additions since new environment variables will be ignored by older applications, whereas adding unknown command line arguments will break an old application.

Do you think it's worth documenting these environment variables and the expected behaviour of a singleuser application more formally, so that people can implement their own `jupyter-singleuser` compatible applications, or is it better to keep things as they are?

minrk:
> Do you think it's worth documenting these environment variables and the expected behaviour of a singleuser application more formally

great idea! I added some docs on CLI, env for spawners. I don't want to say that we have a fully defined spec for swapping out single-user implementations beyond what jupyterhub-singleuser happens to do (yet), but documenting what we do is always a good thing.

jupyterhub/singleuser/mixins.py
minrk: Yeah, fixed to match the convention. `_:trait_name:_default` is the convention from when traitlets only had magic method names, no `@default` decorators. The name no longer has any significance, but might as well use the established one. I think we could use `def _:(self)` for all of these if we wanted (not sure how it would like using the same name for all of them...)

yuvipanda:
BTW, this breaks running JupyterHub 1.5 server with JupyterHub 1.4 singleuser.

minrk:
This alone shouldn't break JupyterHub cross versions, since it should get it from $JUPYTERHUB_SERVICE_URL, but I may have gotten something confused. Some *images* that make assumptions might break, but that's different. But there probably won't be a 1.5, and 2.0 definitely can't work with 1.x because of the RBAC stuff, so we can ensure this change doesn't get backported to 1.x if it's breaking.

yuvipanda:
@minrk right, so far it has broken `Spawner.default_url`, and am validating to see if the idle culler is also affected. Agree that it's ok if we don't do it for 1.x

minrk:
Ah! Default url not working is probably a bug. That can only be done by CLI so shouldn't have changed, I think. Can you open an issue?

minrk:
Oops, never mind. Just re-read the PR to see that that was changed to an env in a 2.0-assuming way. I shouldn't comment from my phone. So yes, this is a breaking change for 2.0-only.

PR #3386, opened 2021-03-22
alpine dockerfile: avoid compilation by getting some deps from apk

minrk:
cryptography is the big one, which needs rust and is a huge pain

The latest release of cryptography is causing our dockerfile tests to fail on CI, due to the added requirement of rust

this simplifies the dockerfile and makes everything faster and smaller by getting a few compiled dependencies from apk


minrk:
Merging so CI returns to normal

Issue #3189, opened 2020-09-28
auth state visible to self


diocas:
### Proposed change
Currently a normal/non admin user is unable to see its own auth state and this is on purpose, according to the comment in the code. I would like to ask if there is a possible security problem or any other particular reason for this decision and, if not, wether it will be possible to change it and make the auth state visible to the corresponding user as well.

### Who would use this feature?
In our JH deployment, we need to forward the oAuth token we obtain from the authenticator to the user container/environment (so that other processes/extensions can use it). Of course this can already be done (by injecting it via the spawner), but then the user container (or some sidecar container/process/etc) needs to be responsible for the renewal (which seems to be the way others are doing it). But, if for some reason a re-authentication is required, then that process will get stuck and the session will have to be restarted completely.

JH already has a renewal functionality baked inside and tightly coupled with everything that needs authentication. If the authenticator implements `refresh_user`, all of this will be handled automatically.  If the auth state was visible to the logged-in user, then Jupyter extensions would be able to fetch the valid token just by querying the user API.
This has the added advantage of keeping all the auth code in a single place, which would allow switching authenticators (which would have to implement the refresh) without having to worry about all of these extra sidecar containers/processes.

### Alternative options
If this information is considered sensitive and you don't think it should be public by default (especially since this field can be used by any authenticator), it would be beneficial to have an extra field in user model that allows us to inject extra public information from the authenticator or other place.

Please let me know what you think.

manics:
`auth_state` may contain tokens and other secrets that shouldn't be visible to a user. They're available to the Spawner though, so you can inject them into a user's environment, e.g. as environment variables. Will this solve your problem?

diocas:
Not really. As I mentioned in the ticket, I already do that. But then I need to get the **refreshed token** to continue to be able to use it (I believe that since JH already handles everything related to auth, it should be able to provide some of this info to Jupyter).
Allowing extra info to be available in some other (public) field, or making the the visibility of `auth_state` configurable, would help.

octavd:
Hello,

I am having the same issue/question like @diocas.
I have this scenario: Each user should have an jupyterhub api token and with that i want to retreive the auth_state -> access_token / refresh_token / id_token for him - **not for other users** and update afterwards the env variables of the containers with the new tokens.
Currently this doesn't work because the user needs to be an admin and not all of them should be admins.

chainlink:
[...]

PR #3393, opened 2021-03-26
Added `auth_state` and `server_state` and fixed `self`

0mar:
# Changes

 - Access to a user's`auth_state`, previously covered by the `admin` flag, now requires the scope `read:users:auth_state`, which is a subscope of `admin:users`
 -  Idem for a user's `server_state` with scope `read:users:server_state`
 -  Metascope `self` now valid for services, but resolves to empty scope set
 - Group model and service model composition now have consistent behavior with respect to user model composition


IvanaH8 (commented):
Nicely done! 

One suggestion: 
As the `self` scope for services has been fixed, we can remove the `service` role in `get_default_roles()` and the temporary assignment of it to services in `switch_default_role()` in lines 227-228. That way non-admin services will get `user` role with `self` scope as for now they are still being assigned `service` role with no scopes by default.

manics:
Does this effectively resolve this feature request? https://github.com/jupyterhub/jupyterhub/issues/3189 ?

IvanaH8:
@manics thanks for pointing that out. 
I think yes eventually, a user will not have to be admin but only be assigned a role that includes the `read:users:auth_state`. BUT for now this scope allows access to any user's auth_state so we would need to think of how to make it user specific generically. Maybe something like `read:self:auth_state` scope that then translates into `read:users:auth_state!user=<username>`.
 
@0mar correct me if I'm wrong.

0mar:
@IvanaH8 That sounds like the best option. It would require a reinterpretation of `self` as we currently have it, but if there exist more use cases for giving admin permissions to user-owned resources, it might be the simplest way to go.

jupyterhub/apihandlers/base.py
```diff
@@ -234,13 +240,26 @@ def user_model(self, user, include_servers=False, include_state=False):
...
+                'server',
+                'pending',
+                'created',
+                'last_activity',
+            },
             'read:users:name': {'kind', 'name'},
             'read:users:groups': {'kind', 'name', 'groups'},
             'read:users:activity': {'kind', 'name', 'last_activity'},
             'read:users:servers': {'kind', 'name', 'servers'},
+            'read:users:auth_state': {'kind', 'name', 'auth_state'},
```
minrk, line 261: `admin:users:auth_state` for these so they aren't under `read:users`?

jupyterhub/apihandlers/base.py
```diff
@@ -220,7 +220,13 @@ def token_model(self, token):
         model.update(extra)
         return model
 
-    def user_model(self, user, include_servers=False, include_state=False):
+    def user_model(
+        self,
+        user,
+        include_servers=False,
+        include_server_state=False,
+        include_auth_state=False,
```
minrk, line 228: Let's try removing these args and rely purely on scopes for what's included.

jupyterhub/apihandlers/base.py
```diff
@@ -79,23 +79,32 @@ def get_scope_filter(self, req_scope):
...
-                if 'group' in sub_scope and kind == 'user':
+                if kind == 'server':
+                    # First check if we have access to user info
+                    user_name = orm_resource.user.name
+                    found_resource = user_name in sub_scope['user']
+                    if not found_resource:
+                        # Now check for specific servers:
+                        server_format = f"{orm_resource.user / orm_resource.name}"
+                        found_resource = server_format in sub_scope[kind]
+                elif 'group' in sub_scope and kind == 'user':
```
minrk, line 101: What happens for `kind='server'` when access is granted through a group? Can we make sure to have a test for that?

jupyterhub/apihandlers/base.py
```diff
@@ -168,17 +177,24 @@ def write_error(self, status_code, **kwargs):
 
     def server_model(self, spawner, include_state=False):
```
minrk, line 178: Let's remove this now unused `include_state` argument

jupyterhub/apihandlers/base.py
```diff
@@ -234,13 +250,26 @@ def user_model(self, user, include_servers=False, include_state=False):
             'pending': None,
             'created': isoformat(user.created),
             'last_activity': isoformat(user.last_activity),
+            'auth_state': '',  # placeholder, filled in later
```
minrk, line 253: Use `None` here instead of empty string, just so it keeps the right type (None or dict, never str).

jupyterhub/apihandlers/base.py
```diff
@@ -257,34 +286,36 @@ def user_model(self, user, include_servers=False, include_state=False):
...
-            'kind': 'service',
-            'name': service.name,
-            'admin': service.admin,
-            'roles': [r.name for r in service.roles],
-        }
+        model = {'kind': 'service', 'name': service.name}
+        req_scope = 'read:services'
+        if req_scope in self.parsed_scopes:
+            scope_filter = self.get_scope_filter(req_scope)
+            if scope_filter(service, kind='service'):
```
minrk, line 314: If this is False, the kind and name are both still returned. Should the service model be entirely empty if there is no read access to this service?

jupyterhub/tests/test_api.py
```diff
@@ -159,6 +159,7 @@ def fill_user(model):
     model.setdefault('created', TIMESTAMP)
     model.setdefault('last_activity', TIMESTAMP)
     model.setdefault('servers', {})
+    model.setdefault('auth_state', '')
```
minrk, line 162: I think this probably shouldn't change? If anything, it should be None instead of a str.

minrk:
Great!

PR #3394, opened 2021-03-29
Use 'secrets' module to generate secrets

yuvipanda:
Python 3.6+ has this


betatim (approved):

yuvipanda:
`secrets` is just a wrapper around `os.urandom`, so there's no need to actively replace. I think the `secrets` module is more ergonomic, so I change it when / if I touch the code for other reasons.

PR #3397, opened 2021-03-30
Refactor scopes tests

0mar:
- Note: based on branch from #3393
# Changes
 - Split `update_roles` in `update_roles` and `assign_default_roles` to reflect intent.
 - Removed call to `assign_default_roles` at hub startup
 - Renamed `add_role` to `create_role`, `add_obj` to `grant_role` and `remove_obj` to `strip_role`
 - Eliminated `kind` as argument in `roles.py` methods

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1890,7 +1890,7 @@ class JupyterHub(Application):
         # make sure that on no admin situation, all roles are reset
         admin_role = orm.Role.find(db, name='admin')
         if not admin_role.users:
-            app_log.info(
+            app_log.warning(
                 "No admin users found; assuming hub upgrade. Initializing default roles for all entities"
             )
             # make sure all users, services and tokens have at least one role (update with default)
--- a/jupyterhub/tests/test_scopes.py
+++ b/jupyterhub/tests/test_scopes.py
@@ -180,7 +180,6 @@ def mock_handler():
     ],
 )
 def test_scope_method_access(mock_handler, scopes, method, arguments, is_allowed):
-    mock_handler = MockAPIHandler()
     mock_handler.current_user = mock.Mock(name=arguments[0])
     mock_handler.set_scopes(*scopes)
     api_call = getattr(mock_handler, method)
@@ -235,7 +234,6 @@ async def test_expand_groups(app, user_name, in_group, status_code):
         app.db.add(group)
     if in_group and user not in group.users:
         group.users.append(user)
-    kind = 'users'
     roles.update_roles(app.db, user, roles=['test'])
     roles.strip_role(app.db, user, 'user')
     app.db.commit()
@@ -261,21 +259,21 @@ err_message = "No access to resources or resources not found"
 
 
 @pytest.fixture
-def temp_role(app):
+def create_temp_role(app):
     """Generate a temporary role with certain scopes.
-    Convenience function that takes care of db stuff and teardown"""
+    Convenience function that provides setup, database handling and teardown"""
     temp_roles = []
     index = [1]
 
     def temp_role_creator(scopes, role_name=None):
         if not role_name:
-            role_name = f'test_role_{index[0]}'
+            role_name = f'temp_role_{index[0]}'
             index[0] += 1
-        role = orm.Role(name=role_name, scopes=list(scopes))
-        temp_roles.append(role)
-        app.db.add(role)
+        temp_role = orm.Role(name=role_name, scopes=list(scopes))
+        temp_roles.append(temp_role)
+        app.db.add(temp_role)
         app.db.commit()
-        return role
+        return temp_role
 
     yield temp_role_creator
     for role in temp_roles:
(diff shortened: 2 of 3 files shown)


jupyterhub/app.py
```diff
@@ -1877,16 +1878,23 @@ async def init_roles(self):
...
+        # make sure that on no admin situation, all roles are reset
+        admin_role = orm.Role.find(db, name='admin')
+        if not admin_role.users:
+            app_log.info(
+                "No admin users found; assuming hub upgrade. Initializing default roles for all entities"
+            )
+            for bearer in role_bearers:
+                Class = orm.get_class(bearer)
+                for obj in db.query(Class):
+                    # if len(obj.roles) < 1: # todo: Should I check if some roles are already assigned?
```
IvanaH8, line 1896: I don't think that anyone will be adding roles to the config file yet on upgrade. We can also mention that in docs as well to make sure. Unless we want to allow for new role creation and assignment during upgrade already which I don't think is a good idea. @0mar what do you think?

IvanaH8 (commented):
I like the split of `update_roles()` and `assign_default_roles()`, much easier to understand what's happening where. Also the renamed role functions.

Looks good to me, left one comment regarding the default role assignment on upgrade. Otherwise I think it's all good to proceed with #3393 update :)

jupyterhub/app.py
0mar: I agree, thanks!

minrk:
Ready for rebase now that #3393 is merged

jupyterhub/tests/test_scopes.py
```diff
@@ -169,44 +179,30 @@ def secret_thing(self):
...
     ],
 )
-def test_scope_method_access(scopes, method, arguments, is_allowed):
-    obj = MockAPIHandler()
-    obj.current_user = mock.Mock(name=arguments[0])
-    obj.raw_scopes = set(scopes)
-    obj.parsed_scopes = parse_scopes(obj.raw_scopes)
-    api_call = getattr(obj, method)
+def test_scope_method_access(mock_handler, scopes, method, arguments, is_allowed):
+    mock_handler = MockAPIHandler()
```
minrk, line 183: This line is now redundant with the `mock_handler` fixture

jupyterhub/tests/test_scopes.py
```diff
@@ -262,79 +260,121 @@ async def test_by_fake_user(app):
...
+def temp_role(app):
+    """Generate a temporary role with certain scopes.
+    Convenience function that takes care of db stuff and teardown"""
+    temp_roles = []
+    index = [1]
+
+    def temp_role_creator(scopes, role_name=None):
+        if not role_name:
+            role_name = f'test_role_{index[0]}'
+            index[0] += 1
```
minrk, line 273: You can use `nonlocal counter` to declare a shared variable from the inherited scope

rather than relying on a list for the closure:

```suggestion
            nonlocal counter
            role_name = f'test_role_{counter}'
            counter += 1
```
```

File: jupyterhub/tests/test_scopes.py [hunk 4]
```diff
@@ -262,79 +260,121 @@ async def test_by_fake_user(app):
...
+        app.db.delete(role)
     app.db.commit()
+
+
+@pytest.fixture
+def user_with_scopes(app, temp_role):
+    """Generate a temporary user with specific scopes.
+    Takes care of db stuff and teardown"""
+    temp_users = []
+    index = [1]
```
minrk, line 291: same here for the nonlocal counter

jupyterhub/tests/test_scopes.py
```diff
@@ -262,79 +260,121 @@ async def test_by_fake_user(app):
...
 
-async def test_request_fake_user(app):
-    user_name = 'buster'
-    fake_user = 'annyong'
-    user = add_user(app.db, name=user_name)
-    test_role = generate_test_role(user_name, ['read:users!group=stuff'])
-    roles.create_role(app.db, test_role)
-    roles.grant_role(app.db, entity=user, rolename='test')
+@pytest.fixture
+def temp_role(app):
```
minrk, line 264: Maybe call this `make_temp_role`? It's often nice for functions to have a verb name, while items have noun names. Looking at the tests, it's reasonable to expect `temp_role` fixture to *be* a temporary role, when it's really a function to make one.

minrk (commented):
This is great! I left some comments about implementation details, but overall a very nice improvement

jupyterhub/tests/test_scopes.py
0mar: Much nicer, thanks!

jupyterhub/tests/test_scopes.py
0mar: Agree! Taken care of

0mar:
Ready for merge!

minrk:
Nice!

PR #3398, opened 2021-04-05
Make JupyterHub Admin page into a React app

### LINKED ISSUE #3039: Make admin page a full-on react app
### Proposed change

JupyterHub pages started out as *extremely* basic: just a start and stop button. They are getting more <del>complex</del> feature-rich with things like named servers, profiles, options forms, etc.

I think it's probably time that we move to a real UI framework, and I think react is a really good fit since we are mostly displaying tables of simple data types (list of users and running servers on admin, list of spawners on named-servers, etc.).

The admin page is currently our most complex page, with pagination, users, sorting, etc. I think this is probably a great place to start with rewriting as a react app with data-flow of users to display. This means, among other things, that things like pagination, etc. need to move to javascript. The HTML template can be greatly simplified and the only relevant Python code is the query for the user list (likely needs pagination in the REST API user list).

### Alternative options

- Use another js framework
- Leave it alone

JupyterHub admins, especially those with lots of users

### How much effort will adding it take?

A moderate amount of effort to:

1. add react as a js dependency
2. rework the admin Jinja template to do much less
3. create react app from scratch to regenerate ~the same HTML we have now

### Who can do this work?

Javascript/React/frontend developers, with some REST API experience.

naatebarber:
closes #3039 

This PR includes a built out stateless React app meant to replace the server-rendered Jinja templates used for the admin console. The app is pulled into the template as a static bundle, and carries out admin requests through the JuptyerHub REST API. This change (and future ones like it) will likely need to be included in the JupyterHub build process. This will mainly just include React code testing, linting, and transpilation.
 
This is currently a work in progress, please feel free to reach out with suggestions / edits you'd like to see as well as further steps needed to integrate 👍 

TODO:

- [x] [Run frontend tests in CI](https://github.com/jupyterhub/jupyterhub/pull/3398#issuecomment-814690486)
- [x] [Explore](https://github.com/jupyterhub/jupyterhub/pull/3398#issuecomment-814290195) moving from class Components to functional components + hook usage
- [x] [UX parity fixes](https://github.com/jupyterhub/jupyterhub/pull/3398#issuecomment-814177817)
- [x] [Pagination](https://github.com/jupyterhub/jupyterhub/pull/3398#issuecomment-814689689)


yuvipanda:
Excited :)

Few questions from a very quick look:

1. What do the `.pre.jsx` files do? Not entirely sure :)
2. What are your thoughts on class based `Component`s vs functional components with hooks?
3. Can you help me understand what redux is used for here?

Excited to get this going :)

/cc @minrk @consideRatio

naatebarber:
@yuvipanda - Quick answer to the questions:

1) I'm pretty much splitting each component into presentational (`.pre.js`), and higher order components (`.js`). Separating them out makes focusing on either the presentational aspect or the functional aspect a bit easier. It also helps out with unit testing the DOM elements rendered by the front end without deep rendering Redux providers, routers, etc per component.
2) (And 3) I chose Redux / class components over functional components / hooks mainly because it acts as a single source of truth, and updates every component's props immediately after any new data is introduced. ie - you delete a user with the `ServerDashboard` component -> a reducer is called to update the centrally managed `user_data` state variable -> now every component that references that variable with the Redux `connect(...)(Component)` function has the updated `user_data` var. Basically helps to keep app data accessible and out of the weeds.

Hooks can't fully replace those features of Redux yet, that said, Hooks + Redux is also a totally valid combination. 
Thoughts? If the app could be improved using a different approach I'd love to check it out

yuvipanda:
That all makes sense, @naatebarber! I'm hoping someone with more frontend chops can look through the code too.

JupyterLab also seems to use yarn, so I guess it's ok to do that over npm.

jsx/README.md
```diff
@@ -0,0 +1,25 @@
...
+This will integrate with Jupyterhub, speeding up client interactions while simplifying the 
+admin dashboard codebase.  
+
+### Build Commands  
+- `yarn build`: Installs all dependencies and bundles the application  
+- `yarn hot`: Bundles the application and runs a mock (serverless) version on port 8000  
+
+### Directory Tree 
+```
+jhadmin/
```
yuvipanda, line 13: I'm guessing this is out of date now? :)

jsx/package.json
```diff
@@ -0,0 +1,67 @@
+{
+  "name": "jupyterhub-admin-react",
+  "version": "1.0.0",
+  "description": "React application for the Jupyter Hub admin dashboard service",
+  "main": "index.js",
+  "author": "nabarber",
+  "license": "MIT",
```
yuvipanda, line 7: I get this error:

```
warning ../package.json: License should be a valid SPDX license expression
```.

Most code in the JupyterHub ecosystem is BSD-3. Would you consider using that instead of MIT?

File: jsx/README.md (hunk 1 above)
Comment (naatebarber) [reply jsx/README.md:13]: Haha yes, leftovers from the extension - updating

File: jsx/package.json (hunk 2 above)
Comment (naatebarber) [reply jsx/package.json:7]: Fixing in next commit 👍

Comment (naatebarber) [reply jsx/package.json:7]: Setting them to `BSD-3-Clause` seems to fix the warning

**Comment by bollwyvl** [comment]
Hey folks: I don't have a lot of skin in the game in Hub-land, so I don't pull a lot of weight around here, but:

I feel like entirely throwing in with a specific, opinionated frontend The One Stack such as react + redux is not a great play for Hub. Unlike the average CRUD SPA, where the goal is to ship _one_ application that works _one_ way, most of the Jupyter frontends end up needing to work many different ways, depending on their deployment goals, and Hub hardly seems like it will be an exception.

Hub has the opportunity to enable a _tremendous_ swathe of collaboration, some of which I've written about [on discourse](https://discourse.jupyter.org/t/feature-idea-jupyter-books-as-a-landing-page-for-hub-log-ins/7831/5) and on a number of other tickets/PRs. 

To that end, I propose taking a breather and considering some larger questions:

- How might we achieve hub-administrator-configurable UI components that meet a broad range of needs?
- How might we achieve harmony with existing communities of frontend practice within the Jupyter ecosystem?
- How might we achieve reuse and composability of actual code between jupyter clients and a future hub client?

Some thoughts on the above:

- I have yet to see a Hub deployment that didn't need to extend the existing jinja templates
  - putting customization behind learning JSX/webpack/babel/npmjs will basically be endgame for keeping stuff running
    - What might we be able to move towards data-driven components, e.g. JSON Schem/RJSF, GraphQL, XState, etc. rather than requiring JS/TS skills (and nodejs) to customize views?
- The heavy first-party organizations are nteract and jupyterlab...
  - both publish npm-based packages with robust TypeScript support
  - both offer existing encapsulation techniques for talking to various jupyter endpoints
  - both offer some approaches for working with JSON schema and other "best effort" ways to deal with _probably_ well-typed data on the wire, full well knowing they (or someone using it) might break the contract
- The _federated module_ distribution approach for JupyterLab 3 (enabled by WebPack 5) likely represents the best balance between ease-of-development and ease-of-ownership
  - a core team manages a toolchain that enables...
    - a distributed community of developers to use `<npm-like> install && <npm-like> build && python setup.py bdist_wheel` and ship packages to...
    - any number of Hub administrators to `pip install`
      - again, I feel it is _not_ reasonable to expect hub administrators to have deal with npmjs.com/webpack-based errors
- Going further, the lumino/lab stack really does offer some very good value for _application_-level features such as keyboard shortcuts, themes/icons, commands, settings, i18n, and is making strides on a11y
  - [jupyterlab-classic](https://github.com/jtpio/jupyterlab-classic) and [voila](https://github.com/voila-dashboards/voila/pull/846) are demonstrating reuse between frontends is possible 
  - being able to reuse existing testing, development, documentation, etc. techniques, not to mention _code_ will have a tremendous network effect between these frontends

**Comment by minrk** [comment]
@naatebarber thanks for the PR! This is awesome. I'll try to have a look over it when I can.

@bollwyvl I don't think jupyterhub's own UI should be particularly customizable. That's a huge increase in maintenance scope that I think isn't manageable for us right now. I think instead the design goal for custom UI should be using the API and building your own UI against it. Shifting UI to be more pure API consumers like this is part of that, I think. JupyterLab extensions that add jupyterhub admin UI without leaving your server would be cool! But I also don't think they need to be developed in jupyterhub itself.

I think the target for what's in jupyterhub itself should be:

- *minimal* UI to get basic tasks done, and
- a fully-functional REST API to enable development of custom and alternative UI, as seen in mybinder.org or things like @yuvipanda's [configurator](https://github.com/yuvipanda/jupyterhub-configurator).

I think custom UI should be served by making it easier for "API-only" deployments of jupyterhub, rather than customization supported by jupyterhub itself.

If folks do want to build client applications against the jupyterhub API using whatever stack they like, that's great! But I think for the Hub itself, it makes the most sense to go with the simplest stack we can. That's currently why the client API now is pure js+html (no build steps at all, which I *love*). The admin page is the main place this is stressing the current approach because the per-user per-server templating is hard in jinja where react-style (whatever stack) fits more naturally, I think.

**Comment by yuvipanda** [comment]
I got it working locally. I'm extremely impressed and excited, @naatebarber :)

Here's a few things I spotted:

1. When there are no groups present, we should show something helpful.
image

2. Content area is full-width, rather than inside a div with a `container`class that restricts it to a  max width. Some 

3. I'm not sure, but does the react app paginate for max users? It probably should, and it should also respect the config we have that controls that. 

4. In the current interface, the set of actions available for the whole hub are visually separated from the list of users:
image

The table in the new interface is a little more obvious:

image


I'm sure there are more. This PR is also already pretty large. @minrk what do you think about us merging this into a branch here, so we can keep improving this incrementally?

**Comment by yuvipanda** [comment]
@bollwyvl https://github.com/yuvipanda/jupyterhub-configurator uses a schema driven approach with pluggy providing support for customization, would love for you to look at that :) I'm with @minrk for the most part - JupyterHub should have enough APIs for anyone to work with. I'm actually half tempted to remove the admin interface and ship it as a separate sergvice...

I do agree we need to make sure customization continues to be possible without requiring react / webpack use. Most of the jinja2 customization I've seen is for home pages - do you see changes for the admin interface as well?

**Comment by yuvipanda** [comment]
> Hooks can't fully replace those features of Redux yet, that said, Hooks + Redux is also a totally valid combination.

@naatebarber I don't work with JS everyday, so I definitely find code with hooks easier to read, write & reason about than class based code. What do you think about exploring a hooks + redux approach?

**Comment by naatebarber** [comment]
@bollwyvl Thanks for your input, I totally agree that switching parts of the JupyterHub front end to React/JSX will call for more of a niche in available developers to make edits. However I do agree with @minrk that the benefits outweigh the costs in taking some of the load off the Jinja templating and at the end of the day the extensibility is limitless via the API.

@yuvipanda Glad to see you got it up! - and good point on those UI inconsistencies, will push up a change to make them a bit more clean. As for pagination, would you recommend creating a separate handler to carry out pagination operations over REST instead of referencing admin page args? I'm sure we can use the same Pagination class & config with some modifications to suit the new UI. Definitely want your guys's opinion on that 👍 

> @naatebarber I don't work with JS everyday, so I definitely find code with hooks easier to read, write & reason about than class based code. What do you think about exploring a hooks + redux approach?

Yep for sure, I'll spin up a branch of this using functional/hooks syntax and we can test it out. Will let you guys know when it's in.

**Comment by bollwyvl** [comment]
> @minrk "API-only" deployments of jupyterhub
>
> @yuvipanda  I'm actually half tempted to remove the admin interface and ship it as a separate service...

Again, folks: I'm just weighing in from the side of having gotten a complex nodejs build chain _out_ of a lot of _product_ builds on Lab 3, and not wanting to see Hub deployments I care about not being able to upgrade because they don't have on-site redux engineers.

Pushing harder on API-only is a very attractive statement indeed, as `jupyterhub-simple-admin` would make it clear: _hey, this is a way, but not The Way_. 

That aside, I'm all for keeping the _shipped_ UI as simple as possible, and no-js-build is indeed the simplest. But if something _is_ getting complicated, "blessing" a particular full front-of-house stack makes it very hard to get it _out_ of there at some point and/or create an alternative to it. In particular, my experience with redux is that getting it out (or even significant upgrades) can mean... _rewrite everything_. 

Alternatively, if a single component on a page needs to be gracefully enhanced, often 200mb of webpack/babel and an app framework are not needed: typescript's (~4mb) `tsc` can happily build hella-backwards-compatible `.js` from `.jsx`, and a `script` tag can wire them together, no pesky, helpful typings required. Keep emitting "dumb" HTML, click to enhance.

>  JupyterLab extensions that add jupyterhub admin UI without leaving your server would be cool! 
>
> do you see changes for the admin interface as well?

Yeah, we've done some of the above. My point is, i guess, things like `jupyter-videochat`: it's a very hub-focused feature, as delivered, and one doesn't even really _need_ an environment capable of launching kernels to use it... but it would taste great next to _Start My Server_.

The Hub UI customizations I've seen do impact the admin side of the house: per user spawners (and spawner config), authn/authz junk, message of the day, etc. where even _one more click_ out to the "source of truth" is too far when folk start to think of The Hub as JupyterHub+the ten layers of stuff under it. 

It's hard to encourage folk to package/test these things, of course, as _hacking it in site-packages keeps the ship afloat_, so nothing much improves over time, or updates become impossible if upstream changes too much.

If things were more modular (or had more, potentially dynamic, schema) it might be easier to get non-firedrill work to update these things.

> @yuvipanda  schema driven approach with pluggy providing support for customization, would love for you to look at that

The configurator looks interesting. Schema Rules Everything Around Me. Whenever possible, unless totally dynamic, i like seeing JSON schema... _in a JSON file_. This allows leveraging stuff in any number of languages/tools without a transpilation step... generating typings, test harnesses, language servers, etc. 

But yeah, more broad application of schema vs special in-house things, and ways to use them, are probably the way we can keep maintenance scope _down_: if a well-written schema can deliver a customizable frontend admin UI, last-word backend validation, interactive documentation, an exhaustive testing harness, autocompletion in the editor, typing stubs... that's definitely a win.

**Comment by minrk** [comment]
@bollwyvl thanks for your points. 

As it is now, due to the extensive jinja macros, lots of the admin page is really implemented server-side in a way that's difficult to replicate. There's also at least one feature that's required for the admin interface and not yet available in the rest api: pagination.

My perspective on rewriting this with react/whatever is that part of the goal is becoming be a pure consumer of the rest API. Then it can serve as an *example* of an admin UI that anyone can ship separately as @yuvipanda suggested, and if we happen to ship one in our default app that's a relatively minor detail.

> not wanting to see Hub deployments I care about not being able to upgrade because they don't have on-site redux engineers.

I wouldn't consider any react/redux rewrite here to be an API anyone would consume except contributors to the default UI itself. I wouldn't expect or require any custom deployment to use react, or even (be able to) reuse these components. Making reusable components that are both right for us and sufficiently customizable for *someone* is out of scope, I think. I think of it as how the default UI *happens to be implemented*. I don't really see how existing deployments would be affected by this, because the admin table should already be opaque to custom templates. From a custom template standpoint, this should only changing the color of the black box.

> things like jupyter-videochat: it's a very hub-focused feature, as delivered, and one doesn't even really need an environment capable of launching kernels to use it... but it would taste great next to Start My Server.

It sounds like this would be sensible as a JupyterHub *service* - where the only interaction with jupyterhub is oauth, and not relying on anything but the rest api. I don't want JupyterHub itself to ever have in-process extensions or plugins, or arbitrary custom pages (even though it's technically possible now...I don't want to advertise or encourage that).

> The Hub UI customizations I've seen do impact the admin side of the house: per user spawners (and spawner config), authn/authz junk, message of the day, etc. where even one more click out to the "source of truth" is too far when folk start to think of The Hub as JupyterHub+the ten layers of stuff under it.

For me, the problem I would like to solve here is making it easier to make a *fully controlled* landing page, completely independent of the Hub, an easier proposition, rather than making JupyterHub pages themselves more customizable. If anything, I'd like them to be *less* customizable.

> If things were more modular (or had more, potentially dynamic, schema) it might be easier to get non-firedrill work to update these things.

I really think the rest API [*is*](https://jupyterhub.readthedocs.io/en/stable/_static/rest-api/index.html) this schema. If they were instead using their own pages and interfacing only via rest/oauth, these are very stable APIs and I think they would have a much better experience. I'd really like to stress that nontrivial custom pages *should not be served by the hub*, and whatever we can do to make that easier for deployers should be the priority here, rather than making our components more robustly customizable.

**Comment by yuvipanda** [comment]
> As for pagination, would you recommend creating a separate handler to carry out pagination operations over REST instead of referencing admin page args? I'm sure we can use the same Pagination class & config with some modifications to suit the new UI.

Until @minrk pointed this out, I didn't realize our REST API didn't support pagination at all! So there are two tasks here:

1. Support pagination in the REST API (opened https://github.com/jupyterhub/jupyterhub/issues/3405)
2. Allow hub admins to configure pagination UX

(2) raises a question of how to let hub admins set configuration (via traitlets) that passes down to the UX. Not entirely sure how, but we need to figure that out too.

**Comment by yuvipanda** [comment]
@naatebarber I'm also very grateful you've added tests! Can you add a github action that runs these in CI as well?

**Comment by yuvipanda** [comment]
I've also updated the PR description with a TODO list from the comments in the PR, so they don't get lost.

**Comment by naatebarber** [comment]
> @naatebarber I'm also very grateful you've added tests! Can you add a github action that runs these in CI as well?
For sure, will implement in this next branch 👍 

Also, on the topic of Redux - following what @bollwyvl said - if we switch to functional React with hooks we have two options in connecting our Redux Store, `useSelector` and `connect`. The former is more hooks-based, but would more tightly couple us with the library and remove some of the separation. Personally I think we should stick with the `connect` approach to keep Redux abstracted. 
_(Helpful article on this: https://www.samdawson.dev/article/react-redux-use-selector-vs-connect)_

Hooks will definitely be used for component isolated state, just wanted to hear your guys's opinion on this.

**Comment by yuvipanda** [comment]
`useSelector` seems the way to go simply because of https://redux.js.org/style-guide/style-guide#use-the-react-redux-hooks-api. I think the lack of boilerplate and conceptual simplicity is worth it, especially since we don't want these components to be reusable outside of the admin interface.

**Comment by naatebarber** [comment]
> `useSelector` seems the way to go simply because of https://redux.js.org/style-guide/style-guide#use-the-react-redux-hooks-api. I think the lack of boilerplate and conceptual simplicity is worth it, especially since we don't want these components to be reusable outside of the admin interface.

Sounds good, gonna try out a branch with those changes and compare 👍

**Comment by naatebarber** [comment]
@minrk @yuvipanda Finished my first pass at restructuring the React app using functional components / hooks - thoughts?
Also, since it's related, should we try tackling pagination (#3405) in this PR as well?

**Comment by minrk** [comment]
Great! I'll find some time to have a look.

> should we try tackling pagination (#3405) in this PR as well?

No, that should be its own PR. It's a big api-only change, so should probably be kept to its own.

**Comment by echarles** [comment]
> I'm actually half tempted to remove the admin interface and ship it as a separate sergvice...

I would love to see this happen. This would allow to replace the admin UI based on configuration.

**Comment by echarles** [comment]
>> useSelector seems the way to go simply because of https://redux.js.org/style-guide/style-guide#use-the-react-redux-hooks-api. I think the lack of boilerplate and conceptual simplicity is worth it, especially since we don't want these components to be reusable outside of the admin interface.
> 
> Sounds good, gonna try out a branch with those changes and compare 👍

`useSelector` makes the code so simple to read that I could not live anymore without it.

**Comment by echarles** [comment]
Probably to be discussed afterwards, but have you considered having typescript instead of plain javascript?

**Comment by minrk** [comment]
> have you considered having typescript instead of plain javascript?

I have, and I've only ever had very negative experiences with typescript. So if it were me doing the work, it wouldn't be typescript, but I won't begrudge whoever ends up doing it from making this choice if it suits them. I know other folks are happy with it.

**Comment by echarles** [comment]
> I have, and I've only ever had very negative experiences with typescript. So if it were me doing the work, it wouldn't be typescript, but I won't begrudge whoever ends up doing it from making this choice if it suits them. I know other folks are happy with it.

Yeah, I am one of the happy guy. I have seen less and less friction those last years with my work, but also more and more well-known project moving to that (or from pure javascript, or from flux). If you want to give it another one day, just start with the typescript option with a simple create-react app https://create-react-app.dev/docs/adding-typescript to setup your env easilly.

**Comment by naatebarber** [comment]
Adding pagination to the React front end, here's the branch if anybody wants to follow. Should be merging it into this PR soon.

https://github.com/naatebarber/jupyterhub/tree/ui-pagination

**Comment by naatebarber** [comment]
@minrk @yuvipanda Just added pagination into the React UI, as well as some cool UI updates to the group management feature. Feel free to take a look and let me know if anything could be improved 👍

**Comment by echarles** [comment]
Here are a few feedbacks after testing this branch

For the code:

- a few unused import remain in some jsx files (AddUser.jsx, GroupEdit.jsx...) - a improvement for next iteration would be to tune the linter,
- Great to have already so many tests - `yarn test` worked great for me.
- The generated `./jsx/build/admin-react.js` is not optimised for prod, especially welcome when copied to `./share/jupyterhub/static/js/admin-react.js`.
- I usually favor material-ui, but in the jupyterhub case, I find react-bootstrap to integrate very nicely in the existing UI.

At runtime:

- When user creation fails (eg system user is not available), no error message is show in the UI.
- After group creation, the group page list is not populated with the effective group list.

Great job @naatebarber ! If the group creation bug is fixed, I would be +1 to get this merged.

**Comment by naatebarber** [comment]
@echarles Thanks for the feedback!
Just tuned up the linter to remove unused imports and variables, fixed the bug with group creation, and added some alert boxes when requests fail. Going to keep searching for bugs and probably add some unit tests that go deeper into testing state change and promise handling.

**Comment by echarles** [comment]
@naatebarber I have tested your latest changes and the reported issues are fixed. The error messages are shown, the group members are shown after group creation and the generated javascript is prod ready. Thx a lot for these.

I will be my LGTM on this, I think it brings much value, especially in terms of group management and bring a strong foundation to enrich the management UI features.

Screenshot 2021-05-18 at 07 22 20

**Review by echarles — APPROVED** [review]
LGTM Kudos @naatebarber

**Comment by yuvipanda** [comment]
This is awesome :) What else do you think needs to happen before the [WIP] tag can be removed?

File: jsx/src/Store.js [hunk 3]
```diff
@@ -0,0 +1,28 @@
+export const initialState = {
+  user_data: undefined,
+  user_page: 0,
+  groups_data: undefined,
+  groups_page: 0,
+  limit: 50,
```
yuvipanda, line 6: Is this configurable? We have about 15k users in our JupyterHub sometimes, and paginating through them is impossible since search doesn't really exist. Right now, the only way to do this is to have all 15k show in the page at once and use the browser's search feature.

echarles: Configuration of the react app is a bigger question. I typically inject the setting in a json in the HTML header and let them read by the react app.

Regarding the many users case, would a typeahead component like http://ericgio.github.io/react-bootstrap-typeahead/ be a good answer, at least better than a  15k lines table?

echarles: This could do the job. @yuvipanda do you need that feature (the typeahead, or the injected settings) before merging?

Screenshot 2021-05-18 at 09 35 34

yuvipanda: @echarles yeah, I think some way to search for users would be important I think. Right now this can be accomplished by setting a huge limit and using browser's search functionality. Some way to accomplish that would be required I think.

We can also just set a limit in the URL. But not sure how the app will react to 15k elements rendered :D

echarles: So to recap, there are 2 options:

1. Add a json in the html header (in the jinja template) populated from a new setting like `c.JupyterHub.max_row_per_page``
2. Add that 3rd party component that allows to search

I let @naatebarber elaborate on more options and give feedback to see if he can pull that requirement in this PR.

> We can also just set a limit in the URL. But not sure how the app will react to 15k elements rendered :D

Quite badly I expect.

naatebarber: @yuvipanda @echarles This is a good point. Currently on the API side we have a pagination default limit (initially 50) and a pagination maximum limit (200). The default limit can be changed with the jupyterhub_config, but if we're dealing with 15k+ users and 200 item-per-page maximums I feel like we'd need some back-end search functionality. 

The best thing we have right now is the `/users/{username}` endpoint, which at least lets us know if a user exists. However I don't think it'd be too difficult to add another endpoint, or at least expand on the one above, to provide some basic text-matching of usernames. Not sure if it's out of scope for this PR, but if it's done it'll definitely be a more efficient way to handle a bunch of users like you guys are saying.

Even if it is out of scope, I think it'd be cool to add search endpoints later on to pull back users and groups by their various attributes. Thoughts?

naatebarber: @echarles 
 > Add a json in the html header (in the jinja template) populated from a new setting like `c.JupyterHub.max_row_per_page``
Good call btw, we should be passing the default limit / custom limit set in the config to the front end

naatebarber:
> This is awesome :) What else do you think needs to happen before the [WIP] tag can be removed?

I'm thinking we just need a couple more improvements to the unit testing, and possibly addressing the search 👍

echarles:
> The default limit can be changed with the jupyterhub_config

@naatebarber Can you point to the config you refer to, and how you read that config?

naatebarber:
@echarles This line here is where the default limit is initialized:
https://github.com/jupyterhub/jupyterhub/blob/6be3160d7487bb362df7095f24454ef2814951fc/jupyterhub/app.py#L999
... and it's referenced by the API handlers through this helper function:
https://github.com/jupyterhub/jupyterhub/blob/6be3160d7487bb362df7095f24454ef2814951fc/jupyterhub/apihandlers/base.py#L268

echarles:
Thx @naatebarber for the links. Are you using that setting in the UI? It looks like the limit of the initialState is not updated based on that server setting.

naatebarber:
> Thx @naatebarber for the links. Are you using that setting in the UI? It looks like the limit of the initialState is not updated based on that server setting.

Nope, currently it's completely detached. I feel like the best way to go about this is to discard the client side limit spec completely and just rely on what's set in the config.

echarles:
> Render the limit as a variable into the front end with tornado

You could use jinja to add in the head of the html something like this

```
    <script id="jupyterhub-admin-config" type="application/json">
      {
        "api_page_default_limit": "...",
      }
    </script>
```

and read it with react

```js
let config = Object.create({});
const htmlConfig = document.getElementById('jupyterhub-admin-config');
if (htmlConfig) {
  config = JSON.parse(htmlConfig.textContent || '') as {
    [key: string]: string;
  }
...
```

cslocum:
> I got it working locally. I'm extremely impressed and excited, @naatebarber :)

@yuvipanda what did you do to get it working locally?  I tried simply installing from @naatebarber's master and I see no changes.  I'm sure this is a dumb question...

echarles:
@cslocum the following should work to install the branch

```
git clone https://github.com/naatebarber/jupyterhub
cd jupyterhub
pip install -e .
```

cslocum:
Thanks @echarles.  That is what I did, but perhaps not in the correct environment.  I did just try installing it directly into the base Ubuntu image, before we do any of our customizations, but that didn't work either.  I will keep poking.  Really excited to see this merged!!

naatebarber:
@yuvipanda Is there anything you think needs to be addressed before this is ready for merge?

yuvipanda:
This LGTM, we can iterate later if we want to. Thoughts, @minrk? Good to hit merge button?

minrk:
:shipit:

minrk:
Thanks @naatebarber!

consideRatio:
Thank you @naatebarber for your heart and effort into this :heart: :tada:

PR #3401, opened 2021-04-06
Added Azure AD as a supported authenticator.

maxshowarth:
Azure AD is listed as a supported service in OAuthenticators README.md, but it is not listed in the JupyterHub documents.


yuvipanda:
Thanks, @maxshowarth!

PR #3403, opened 2021-04-07
Fastapi example

kafonek:
Example of a FastAPI / uvicorn JupyterHub managed service.  

 * The .gif is ~3 MB, I can try to host it somewhere else if you don't want that file size in the repo
 * I didn't try to feature match the `httpx.AsyncClient` (`client.py`) with `HubAuth`, although we could go that route
 * "Works on my machine" levels of testing
 * I'm not a FastAPI guru, although I think the code and patterns are acceptable from what I've read


examples/service-fastapi/app/security.py
```diff
@@ -0,0 +1,86 @@
...
+
+from .client import get_client
+
+### For authenticated endpoints, we want to get auth from one of three ways
+### 1. "token" in the url params
+### 2. "jupyterhub-services" cookie (or config via env)
+### 3. Authorization Bearer header (with oauth to Hub support)
+auth_by_param = APIKeyQuery(name="token", auto_error=False)
+
+COOKIE_NAME = os.getenv("JUPYTERHUB_COOKIE_NAME", "jupyterhub-services")
```
minrk, line 18: I think you can skip the shared-cookie auth. This is on its way out in favor of oauth all the time. It will work, but it's never required.

examples/service-fastapi/app/security.py
```diff
@@ -0,0 +1,86 @@
...
+auth_by_cookie = APIKeyCookie(name=COOKIE_NAME, auto_error=False)
+
+if "PUBLIC_HOST" in os.environ:
+    ### When running in Docker or maybe other infrastructure,
+    ### JUPYTERHUB_API_URL is "http://jupyterhub" but we need to give
+    ### clients (swagger webpage) a link to the public url
+    auth_url = os.environ["PUBLIC_HOST"] + "/hub/api/oauth2/authorize"
+else:
+    auth_url = os.environ["JUPYTERHUB_API_URL"] + "/oauth2/authorize"
```
minrk, line 27: jupyterhub_api_url is not generally correct. It should be calculated based on the *request* host, while the API url is only for direct service->hub communication, not user requests from the outside.

If FastAPI doesn't allow calculation of path-only authorization targets (not unusual, since 'internal' oauth is weird!), then requiring public host to always be set is probably best. Or defaulting to the default public url: `http://127.0.0.1:8000/hub/api` with a warning.

examples/service-fastapi/app/security.py
```diff
@@ -0,0 +1,86 @@
...
+)
+### ^^ For Oauth in the Swagger webpage, we set the authorizationUrl
+### to the Hub /oauth2/authorize endpoint, so browser does a GET there and
+### receives a 'code' in return.  Then the browser does a POST to our
+### internal /get_token endpoint with that code, and our server subsequently
+### POSTs to Hub /oauth2/token to get/return an acecss_token.
+### The reason for the double POST is that the client (swagger ui) doesn't have
+### the client_secret (JUPYTERHUB_API_TOKEN), but our server does.
+
+### For consideration: build a pydantic User model
```
minrk, line 39: I think that would be cool, just because I'm interested to see what it looks like! Is it easy to allow extra fields with pydantic? Then you can make a subset at least.

minrk (commented):
Awesome! Thanks for submitting. I'll try to have a more thorough look in a bit.

bollwyvl:
Here's the one @yuvipanda and i hacked together... not even _works on my machine_ level of testing, but may have some stuff that can be stolen:

https://gist.github.com/bollwyvl/1609d764317a439dfd95f6427fbd38f8

bollwyvl:
Some notes: I lurve some FastAPI, pydantic, etc. but, as mentioned on #3398, having a complex class (or worse, a tree of classes) be the source of the JSON schema is kind of a bummer.

examples/service-fastapi/app/security.py
kafonek: I get turned in circles on this oauth2 stuff, it's good to talk through it.  There are four places that I deal with the auth url right now.  The problem child in this flow is Swagger, not necessarily FastAPI.

1. In jupyterhub_config.py, starting the service.
2. In security.py, setting the url that users get redirected to when they click "authorize" in the OAuth2 option in Swagger
3. In app.py, configuring Swagger UI
4. In service.py get_token route, which handles the callback from 3 and then posts to Hub to get the token

2 determines what url you get redirected to when you click on "Authorize -> OAuth2" in the swagger page.  After some testing it looks like this can just be set to a relative path `/hub/api/oauth2/authorize`.

When the redirect to the Hub login page happens, the `redirect_uri` url parameter is created by Swagger from the request host + config in 3.  Without setting `oauth_redirect_uri` in 1, you'll always get a `redirect uri mismatch` error `'<your-host>/services/fastapi/oauth_callback != '/services/fastapi/oauth_callback'`.  4 needs to match this as well.

So in summary, Swagger adds the request host to the `redirect_uri` and we need to account for that when starting the service (`jupyterhub_config.py`) and on the `/oauth2/token` callback (`service.py`) or we'll hit `redirect uri mismatch` errors.  

Would you put that warning and set the default env variable is in `jupyterhub_config.py`?

kafonek:
Thanks @bollwyvl I've added User, Group, and Server models based on https://jupyterhub.readthedocs.io/en/stable/_static/rest-api/index.html.  I always get `null` for groups and servers in testing.  Maybe it's just my setup.

Gif is updated to show how Swagger renders the Model and response types

examples/service-fastapi/app/security.py
kafonek: auth by cookie removed

examples/service-fastapi/app/security.py
kafonek: Added a warning when `PUBLIC_HOST` env wasn't defined in `jupyterhub_config.py` and set it to `http://127.0.0.1:8000` as a default.

examples/service-fastapi/app/security.py
kafonek: Pydantic models added and used in `get_current_user` / endpoint routes

examples/service-fastapi/app/client.py
```diff
@@ -0,0 +1,13 @@
+import os
+
+import httpx
+
+
+### For consideration: turn this into a class with
+### feature parity for HubAuth, or even subclass HubAuth?
```
minrk, line 7: I think both are useful - using HubAuth as in https://gist.github.com/bollwyvl/1609d764317a439dfd95f6427fbd38f8, or the no-imports implementation. We could have both.

It would be better, though, to use HubOAuth with a callback handler than HubAuth, which doesn't do the oauth flow, it only works for token-authenticated API requests or old shared-cookie auth.

kafonek: comment and README updated to say HubOAuth instead of HubAuth.  If you're okay with it, I'll leave the minimal implementation as is, and not try to create a HubOAuth with an httpx (async requests) plug-in.

kafonek:
Happy to add anything else you want @minrk @yuvipanda @bollwyvl .  Otherwise think I'm done on my side.  Thanks for all the feedback and help.

minrk:
Awesome, thanks!

PR #3407, opened 2021-04-08
Don't delete all oauth clients on startup

yuvipanda:
When an oauth client changes, we delete all the tokens
associated with that client. This invalidates all user sessions
for that oauth client, and the oauth client's users will need to
go through the OAuth workflow again after the cache period (specified
by cache_max_age in HubAuth, 5min by default). This is fine in theory,
since oauth client information doesn't change frequently.

However, we were deleting and re-adding all oauth clients each time
the hub started! This was unnecessary, since the data was going to
be the same 99% of the time. Rest of the time, we should just update,
preventing unnecessary churn.

This PR does that.

Ref https://github.com/yuvipanda/jupyterhub-configurator/issues/2
Ref https://github.com/berkeley-dsep-infra/datahub/issues/2284
Chat Log starting at https://gitter.im/jupyterhub/jupyterhub?at=606eda5446a93d4a19b30e46

### DIFF
--- a/jupyterhub/oauth/provider.py
+++ b/jupyterhub/oauth/provider.py
@@ -563,7 +563,6 @@ class JupyterHubOAuthServer(WebApplicationServer):
         # so we do this manually. It's protected inside a
         # transaction, so should fail if there are multiple
         # rows with the same identifier.
-        # TODO: Is this safe?
         orm_client = (
             self.db.query(orm.OAuthClient).filter_by(identifier=client_id).one_or_none()
         )
@@ -571,11 +570,13 @@ class JupyterHubOAuthServer(WebApplicationServer):
             orm_client = orm.OAuthClient(
                 identifier=client_id,
             )
-
+            self.db.add(orm_client)
+            app_log.info(f'Creating oauth client {client_id}')
+        else:
+            app_log.info(f'Updating oauth client {client_id}')
         orm_client.secret = hash_token(client_secret)
         orm_client.redirect_uri = redirect_uri
         orm_client.description = description
-        self.db.add(orm_client)
         self.db.commit()
 
     def fetch_by_client_id(self, client_id):
--- a/jupyterhub/oauth/provider.py
+++ b/jupyterhub/oauth/provider.py
@@ -558,19 +558,23 @@ class JupyterHubOAuthServer(WebApplicationServer):
 
         hash its client_secret before putting it in the database.
         """
-        # clear existing clients with same ID
-        for orm_client in self.db.query(orm.OAuthClient).filter_by(
-            identifier=client_id
-        ):
-            self.db.delete(orm_client)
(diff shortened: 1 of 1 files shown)


jupyterhub/oauth/provider.py
```diff
@@ -558,19 +558,23 @@ def add_client(self, client_id, client_secret, redirect_uri, description=''):
...
+        # Sqlalchemy doesn't have a good db agnostic UPSERT,
+        # so we do this manually. It's protected inside a
+        # transaction, so should fail if there are multiple
+        # rows with the same identifier.
+        # TODO: Is this safe?
+        orm_client = (
+            self.db.query(orm.OAuthClient).filter_by(identifier=client_id).one_or_none()
         )
+        if orm_client is None:
+            orm_client = orm.OAuthClient(
```
minrk, line 571: Can we get an info-level message about "Updating existing oauth client {client_id}" vs "Creating new" based on whether it was found?

jupyterhub/oauth/provider.py
```diff
@@ -558,19 +558,23 @@ def add_client(self, client_id, client_secret, redirect_uri, description=''):
...
-            identifier=client_id,
-            secret=hash_token(client_secret),
-            redirect_uri=redirect_uri,
-            description=description,
+        # Update client if it already exists, else create it
+        # Sqlalchemy doesn't have a good db agnostic UPSERT,
+        # so we do this manually. It's protected inside a
+        # transaction, so should fail if there are multiple
+        # rows with the same identifier.
+        # TODO: Is this safe?
```
minrk, line 566: Yes, this is safe. This is the pattern we use to create ~everything (Users, groups, etc.).

We may need to come back to recovering more gracefully from failures due to races if and when we get to developing horizontally scalable Hubs, but it can't fail right now since the failure addressed by UPSERT can only occur with multiple db writers, which is impossible with jupyterhub today.

jupyterhub/oauth/provider.py
```diff
@@ -558,19 +558,23 @@ def add_client(self, client_id, client_secret, redirect_uri, description=''):
...
         )
+        if orm_client is None:
+            orm_client = orm.OAuthClient(
+                identifier=client_id,
+            )
+
+        orm_client.secret = hash_token(client_secret)
+        orm_client.redirect_uri = redirect_uri
+        orm_client.description = description
         self.db.add(orm_client)
```
minrk, line 578: This `db.add` can move to the `if orm_client is None` branch, since we don't need to add an *existing* client to the session, only new ones.

minrk (commented):
Great! Minor comments about logging, but 👍

yuvipanda:
Thanks for the review, @minrk! Done :)

minrk (approved):

minrk:
Thanks!

PR #3411, opened 2021-04-09
make oauth token expiry configurable

minrk:
and default to cookie_max_age_days instead of 1 hour, since it makes sense for the token in the cookie to expire when the cookie itself is set to expire.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -358,8 +358,10 @@ class JupyterHub(Application):
     ).tag(config=True)
 
     oauth_token_expires_in = Integer(
-        24 * 3600,
         help="""Expiry (in seconds) of OAuth access tokens.
+
+        The default is to expire when the cookie storing them expires,
+        according to `cookie_max_age_days` config.
 
         These are the tokens stored in cookies when you visit
         a single-user server or service.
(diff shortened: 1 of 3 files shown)


jupyterhub/app.py
```diff
@@ -356,6 +356,27 @@ def _validate_config_file(self, proposal):
...
         """,
     ).tag(config=True)
+
+    oauth_token_expires_in = Integer(
+        24 * 3600,
+        help="""Expiry (in seconds) of OAuth access tokens.
+
+        These are the tokens stored in cookies when you visit
+        a single-user server or service.
+        When they expire, you must re-authenticate with the Hub,
```
yuvipanda, line 366: Can you add a comment here perhaps about API tokens that never need refreshing? Or should that be elsewhere?

minrk: API tokens *may* expire, depending on how they are issued, but the default is no expiry.

minrk: Added a note that this doesn't apply to API tokens

yuvipanda:
To recap my understanding:

1. API Tokens are given to python process in the user server (or service). They never expire, and can be used to auth against the user server API or hub API
2. OAuth2 tokens are kept in the client side as cookies, and are there when the oauth flow (with the hub & service or hub & user server) is completed.

Until now, the oauth token itself expired before the cookie did, requiring users refresh to re-run the OAuth flow between their service / server and the hub again. But JupyterLab (and maybe notebook?) put the *API Token* in their html and used it for requests, so they weren't affected by this. Services, and user servers running via jupyter-server-proxy (like RStudio) ran into this, and would affect their API requests - they would return 403s, even though the local cookie hasn't expired.

With this, the oauth token expires when the cookie expires. This is a longer time duration - probably longer than a single server is expected to live. It also makes debugging easier.

yuvipanda:
Debug session talking through this with @minrk from https://matrix.to/#/!IrRJzNeYwTiocwdjaf:gitter.im/$NufgCo65X6ddFc-K5aWfIA7Q-HD-6VEvFhPfbuxWUJg?via=gitter.im&via=matrix.org&via=matrix.wasserbrand.org.

Thanks a lot for helping me understand this, and for the fix, @minrk!

Issue #3405, opened 2021-04-07
Support pagination in the REST API


yuvipanda:
### Proposed change

Currently, our REST API does not support pagination. This blocks a number of features:

- Replacing our admin interface with a React app (https://github.com/jupyterhub/jupyterhub/pull/3398, https://github.com/jupyterhub/jupyterhub/issues/3039)
- Performance with large number of users (https://github.com/jupyterhub/jupyterhub/issues/2954)

To fix these, we should support pagination in the REST API

### Who would use this feature?

Anyone using the API, including core JupyterHub itself (with the React admin interface)

### (Optional): Suggest a solution

The following handlers seem the only ones needing pagination support:

1. Listing all users - https://github.com/jupyterhub/jupyterhub/blob/c5bfd2800542f977528720c0cc8c56684f0da380/jupyterhub/apihandlers/users.py#L56
2. List all groups - https://github.com/jupyterhub/jupyterhub/blob/c5bfd2800542f977528720c0cc8c56684f0da380/jupyterhub/apihandlers/groups.py#L38
3. List users in group - https://github.com/jupyterhub/jupyterhub/blob/c5bfd2800542f977528720c0cc8c56684f0da380/jupyterhub/apihandlers/groups.py#L77
4. List routes in proxy - https://github.com/jupyterhub/jupyterhub/blob/c5bfd2800542f977528720c0cc8c56684f0da380/jupyterhub/apihandlers/proxy.py#L15

SQLAlchemy should be fairly straightforward to paginate for 1-3, and 4 is just returning an in-memory datastructure, so should be easy to paginate as well.

minrk:
Thanks for opening this! I thought we had an Issue already, but I realize after searching it was only discussed in related issues like #2954. It's been on my mental todo list for some time...

minrk:
oops, closed by #3413

PR #3413, opened 2021-04-09
Support Pagination in the REST API

naatebarber:
As mentioned in #3405, add pagination for:
 - Listing all users - `/hub/api/users`
 - Listing all groups - `/hub/api/groups`
 - Listing users in a group -`/hub/api/groups/{group_name}`
 - Listing routes in proxy - `/hub/api/proxy`

Trying to conform with JupyterHub's design as much as possible so please nitpick on best practices 👍


jupyterhub/apihandlers/groups.py
```diff
@@ -37,7 +37,18 @@ class GroupListAPIHandler(_GroupAPIHandler):
...
     def get(self):
         """List groups"""
-        data = [self.group_model(g) for g in self.db.query(orm.Group)]
+        offset = self.get_argument("offset", None)
+        limit = self.get_argument("limit", None)
+
+        query = self.db.query(orm.Group)
+
+        if offset is not None:
+            query = query.offset(int(offset))
```
minrk, line 46: Need some argument validation to avoid 500 error on invalid offset, limit. Generally `raise web.HTTPError(400, "informative message")`

minrk (commented):
1. this is rad, thanks!
2. I think we want to add a default limit config at JupyterHub.api_page_default_limit (and probably also separate upper limit to truncate limit if more than the max is requested). For backward-compatibility, this can be None for now.
3. we'll need tests to exercise these
4. the `docs/rest-api.yml` spec probably needs updating to mention pagination, and probably a prose example in `docs/source/reference/rest.md`

Truncating user lists in group models seems tricky - the group list API has two-dimensional pagination (group list itself, user list in each group model). Any great ideas for that one? Maybe just truncate the user lists in each group model and require requesting each group individually to get full lists?

naatebarber:
@minrk Thanks for taking a look!

I'll start adding the config, documentation mods, and tests in with the upcoming commits 👍 

Also, I've got a couple questions on the double pagination topic, as well as applying pagination to child collections.
1) For truncation of the users list in `/hub/api/groups`, should we return a smaller list of users (max 5 or so), or return an int representing number of users in the group? Just wondering if it'd be misleading to show only a portion of the data.
2) I haven't found a way to paginate the user lists within a group model through the first SQLAlchemy query, and instead have been using a roundabout method to split off the user list, paginate it, and return the paginated data with the JSON model. Is there a method you're familiar with that could apply limit and offset to the `users` child collection within the first query?

jupyterhub/apihandlers/groups.py
```diff
@@ -75,8 +104,37 @@ class GroupAPIHandler(_GroupAPIHandler):
...
+
+        if offset is not None:
+            try:
+                offset = int(offset)
+            except Exception as e:
+                raise web.HTTPError(
+                    400, "Invalid argument type, offset must be an integer"
+                )
+
+            group.users = group.users[offset:]
```
minrk, line 121: Editing this list attribute is actually changing group membership, which we don't want to do. The magic of ORMs!

jupyterhub/apihandlers/groups.py
```diff
@@ -75,8 +104,37 @@ class GroupAPIHandler(_GroupAPIHandler):
...
+
+        if limit != self.settings['app'].api_page_default_limit:
```
minrk, line 123: Since these validation checks happen several times, maybe add a new `def get_api_pagination(self):` method to the base APIHandler that returns `offset, limit` and does this validation?

minrk (commented):
Awesome!

I think it does make sense to add an integer for the number of users in the group to the group model.

I think we should consider how many users it takes for the group.users sub-list to become costly, and maybe defer paginating that one for now, starting with just the list APIs, since they are simpler.

jupyterhub/apihandlers/groups.py
naatebarber: Good point, will revert this one back for now 👍 Wasn't very happy with the after-the-fact slicing either. You think we should add the number of users int in or save that for later?

minrk: Yeah, since it's both more complex and less used than the lists themselves, I think we can save it for later.

jupyterhub/apihandlers/groups.py
naatebarber: Just pushed up a commit, looks much cleaner

jupyterhub/app.py
```diff
@@ -996,6 +996,15 @@ def _cookie_secret_check(self, change):
         """,
     ).tag(config=True)
 
+    api_page_default_limit = Integer(
```
minrk, line 999: For backward-compatibility, let's set both of these to 0 (no limit). That lets us make pagination opt-in for at least one release cycle.

minrk (commented):
Awesome! I think we can land  this

Future improvements:

- next page link in headers, if there is a next page

minrk:
FWIW, for further discussion (doesn't need to be addressed her): we now have 'Pagination' config used in the admin page, and this offset and limit that are not connected to that. The main difference is that config used `page=1&per_page=100`, while this uses `offset=10&limit=100`. I don't have a strong preference - page number and per_page are easier for building UI, and that's why we used them there. Offset and limit are easier for implementing in the API handler, since they are exactly the arguments to pass to SQL.

jupyterhub/app.py
naatebarber: Can do 👍 Should that be done in this PR now that it's merged or somewhere else?

PR #3421, opened 2021-04-13
Docker arm64 builds

manics:
This is a copy of https://github.com/jupyterhub/configurable-http-proxy/blob/fb2bca9e038281398358d6f66fcc72e9ecc2d536/.github/workflows/publish.yml#L26-L75
What do you reckon?

As with https://github.com/jupyterhub/configurable-http-proxy/pull/304 this would require creating GitHub secrets and removing any existin Docker Hub automated builds.


manics:
Can be tested using https://github.com/users/manics/packages/container/package/jupyterhub-jupyterhub
(pushed using a copy of this branch with an additional commit to push to the GitHub registry https://github.com/manics/jupyterhub/commit/60da423c4c5f37cc0aa4e1ddd5d824a3517c58d9)

consideRatio:
I think this is sensible! Thanks for pushing these kinds of changes across the org @manics!

Is there a sensible approach to retaining a checksum / digest when we use amd64 + arm64 base images? I guess the problem is that we can't use a checksum / digest for the image because it varies depending on the architecture or something like that?

## Action points
- [x] Setup DOCKERHUB_USERNAME and DOCKERHUB_TOKEN secrets
- [x] Setup access rights for our jupyterhubbot account
- [ ] Deactivate automated builds by at https://hub.docker.com/repository/docker/jupyterhub/jupyterhub/builds/edit

manics:
That's right, the digest is for the individual amd64 and arm64 images so they're different as you can see: https://hub.docker.com/_/ubuntu?tab=tags&name=focal-20200729

Do we need a hash though? `ubuntu:focal-20200729` contains a date which seems pretty specific. Having a hash guards against malicious replacement of a tagged image but I don't think we're requiring this anywhere else.

consideRatio:
@manics I feel fine about it especially because it is the ubuntu image and it is date-tagged rather than major-minor-version tagged or similar. I recall that someone may have felt it was needed in the past though so I figured we consider it a bit.

minrk:
👍 the docker hub builds seem to be hard for folks to inspect/debug due to permission issues.

The hash is about trust - the date tag can easily be rewritten, whereas the hash is a cryptographic signature of the contents, and harder to change out from under you. Using just the tag is trusting the maintainers of the upstream image and any local registry or cache to not redefine the tag (maliciously or otherwise). The likelihood of that is *very* small, so I don't feel bad removing it.

@manics can you also remove the now-unused `hooks` directory which did the tags stuff on docker hub?

minrk (approved):

minrk:
Oh, I realized one last thing: this should also build and push the `jupyterhub/jupyterhub-onbuild` image, as done in `hooks/post_build` and `post_push`.

We've also been pushing `master` and `.dev` tags from the master branch. What do you think about continuing that, or maybe stopping?

manics:
I've:
- Moved the circle-ci workflow into test.yml
- Deleted `hooks/`
- Added the `-onbuild` and `-demo` images to release.yml
- Enabled image pushes for `master`

I'm testing the release/push workflow with this additional commit https://github.com/manics/jupyterhub/commit/32e5013adbb55ca66132b6c746a5362f24b68339 on my https://github.com/manics/jupyterhub/tree/docker-arm64-push branch, if it works it should push to my GitHub registry and tag images `:docker-arm64-push`.

I haven't added the `.dev` tags, does it offer any benefits over `master`? I think the latest dev tag will always be the same as `master`, and older dev tags will be frozen at the commit before the tag, which will be identical to the tagged code apart from the version bump.

minrk:
> I haven't added the .dev tags, does it offer any benefits over master?

Only when we bump versions (i.e. 1.4.dev will stop getting updated when we mark master as 2.0) and if/when we have backport branches. I don't think it's important, though, so fine to stop.

> causing the circle-ci test to fail 

I disabled circleci, so this should be fine from now on.

Great work!

manics:
I spotted a problem with the demo arm build, I'll open a follow-up PR with details (and a registry test)

minrk:
ok, great. jupyterhub/jupyterhub pushed just fine. onbuild failed with permissions, but I checked the repo permissions and bots should have read & write. Any idea there?

minrk:
Maybe a smoke test `docker run $image jupyterhub --help` during the docker build workflow?

manics:
See https://github.com/jupyterhub/jupyterhub/pull/3425 though that needs another fix to run in PRs ....

The problem isto do with the build workflow (Python wheels aren't available for all the jupyterhub-demo dependencies), the `jupyterhub/jupyterhub:master` image is fine- I've run it on an arm system. I don't think we can test `docker run` for arm64 in this workflow as `buildx` is only for building.

I've no idea about the Docker repo permissions for `jupyterhub-onbuild` and `jupyterhub-demo`... @consideRatio any ideas?

PR #3422, opened 2021-04-14
login-template: Add a "login_container" block inside the div-container.

### LINKED ISSUE #3414: [RFC] Login page customization requires full HTML replacement
Currently, the login page can not be easily customized, since the full `main` or `login` block needs to be replaced. 
A common use case (in my opinion) is to keep the login box and add additional information, e.g. some news, logos, changelogs, usage details etc. 

Keeping the login box and addings things "around" it currently is not possible in a good way, since the`container` is part of the `login` block:
https://github.com/jupyterhub/jupyterhub/blob/2ff6d2b36c855e86dfab188dc62b31c8738b07a5/share/jupyterhub/templates/login.html#L9-L12
Since `container` can not be put inside `container` in `bootstrap`, this means one has to replace the full HTML. 

I'm currently using a template customization such as this to work around the issue:
```HTML
{% extends "templates/login.html" %} {% block main %} 
<div id="login-main" class="container">
        <div class="row">
                {{ super() }}
                <div class="col-sm-9">
                       <!-- my custom content here, to be shown right of the login box -->
                </div>
        </div>
</div>
{% endblock %}

{% block login %}
<div class="col-sm-3">
{% set lines=super().split('\n') %}
{% for line in lines %}
{% if line.find("login-main") > 0 and line.find("container") > 0 %}
<div>
{% else %}
{{ line }}
{% endif %}
{% endfor %}
</div>
{% endblock %}
```
This takes the base template, and replaces the line containing `login-main` and `container` with a simple `<div>` to allow defining your own `container` for the page by overriding the `main` block. But that feels like a dirty hack. 

Side-note: I somehow could not get:
```HTML
{{ super()|replace(' id="login-main" class="container"', "") }}
```
to work, which would be slightly less of a dirty hack, somehow quotation mark matching fails, even when escaping them. 

My idea for cleaning this up would be to pull out the:
```HTML
<div id="login-main" class="container">
```
into the `main` block to allow for easier customization, reusing as much as possible from the base template, so the customization is more stable upon upgrades. 

Any comments / further ideas?

olifre:
This allows for more flexible customization of the login page,
since it allows to re-use the login form in an extending template
by reusing the new block.

This was not cleanly possible before since the main container
was part of the very same block as the form code.

fixes #3414

### DIFF
--- a/share/jupyterhub/templates/login.html
+++ b/share/jupyterhub/templates/login.html
@@ -10,6 +10,7 @@
 
 {% block login %}
 <div id="login-main" class="container">
+{% block login_container %}
 {% if custom_html %}
 {{ custom_html | safe }}
 {% elif login_service %}
@@ -83,6 +84,7 @@
   </div>
 </form>
 {% endif %}
+{% endblock login_container %}
 </div>
 {% endblock login %}
 


minrk:
Thanks!

PR #3424, opened 2021-04-15
more changelog for 1.4

minrk:
I think we can release 1.4 once the last tweaks for the new docker builds are in (@manics)


yuvipanda:
what do you think of getting https://github.com/jupyterhub/jupyterhub/pull/3430 as well? Will help a lot with trying out api only hub

consideRatio (approved):

yuvipanda:
\o/ YAY!

PR #3426, opened 2021-04-15
[rbac] Fix log message for modifying existing roles

IvanaH8:
- Fixes the log message about modifying existing role (in `_overwrite_role()`) appearing on next startup even if the role definition in config file not modified.
- Adds the possibility to tests in `create_roles()` and simplifies the test.
- Removes duplicate `scopes` assignment in `expand_roles_to_scopes()`.

@0mar, @minrk could you have a quick look?


minrk:
Great!

PR #3430, opened 2021-04-19
Support Proxy.extra_routes

yuvipanda:
When the hub is running in API-only mode, it's
very useful to have the proxy know where to send
URLs that would normally be serviced by the hub.
For example, / might go to a service that renders
a home page, while `/user` might go to a service that
tells the user their server is dead.

Right now, this happens 'out of band', with a process
that has to talk to the proxy directly. This is a
bit messy - the routes need to be re-added when the
proxy restarts, the hub might try to remove them, etc.
By adding support for this in the hub itself, all
this complexity is now removed and the hub continues
to own all the routes in the proxy

### DIFF
--- a/jupyterhub/proxy.py
+++ b/jupyterhub/proxy.py
@@ -123,6 +123,12 @@ class Proxy(LoggingConfigurable):
         a URL as target. The hub will ensure this route is present
         in the proxy.
 
+        If the hub is running in host based mode (with 
+        JupyterHub.subdomain_host set), the routespec *must*
+        have a domain component (example.com/my-url/). If the
+        hub is not running in host based mode, the routespec
+        *must not* have a domain component (/my-url/).
+
         Helpful when the hub is running in API-only mode.
         """,
     )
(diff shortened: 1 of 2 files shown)


jupyterhub/proxy.py
```diff
@@ -384,6 +399,11 @@ async def check_routes(self, user_dict, service_dict, routes=None):
                     )
                     futures.append(self.add_service(service))
 
+        # Add extra routes we've been configured for
+        for routespec, url in self.extra_routes.items():
+            good_routes.add(routespec)
+            futures.append(self.add_route(routespec, url, {'hub': True}))
```
minrk, line 405: Not 'hub'? These aren't routes for the hub itself.

Maybe:

```suggestion
            futures.append(self.add_route(routespec, url, {'extra': True})
```

We mostly don't use this info for things other than user servers.

yuvipanda: Done

JamesSun-LR: @yuvipanda, the extra_routes will be auto-deleted after a while, what's the reason?

PR #3431, opened 2021-04-19
Persist roles through OAuth process

minrk:
- [x] Attach role limit to OAuthClient.allowed_roles
- [x] Attach authorized roles to OAuthCode.roles
- [x] pass OAuthCode.roles to APIToken.roles on completion
- [x] test requesting subset of roles
    
standard 'scopes' in oauth process are matched against our 'roles' instead of our actual low-level scopes.

Not in this PR:

- setting OAuthClient.allowed_roles via config:
  - for services (presumably Service.allowed_roles)
  - for single-user servers (different for different users?)

It doesn't feel 100% great that this maps the standard scopes in the oauth flow onto roles. However, we have to do this because tokens currently only have roles, not direct assignment of scopes. It means that the scopes part of an oauth flow for a given service may vary depending on the configuration of the Hub deployment.

In practice, an oauth service will usually define its own role and only be allowed to request that one and only role. This is functionally equivalent, but a bit more verbose than putting scopes directly on the client itself.

We *could* use actual scopes here if we made one of these changes instead:

- attach scopes directly to API Tokens, in addition to (or instead of?) roles
- generate roles on demand for scope bundles, e.g. `oauth-role-$hashOfScopes`
- generate single roles for oauth clients, and disallow requesting specific scopes in oauth flow

### DIFF
--- a/jupyterhub/orm.py
+++ b/jupyterhub/orm.py
@@ -137,43 +137,34 @@ class Server(Base):
         return "<Server(%s:%s)>" % (self.ip, self.port)
 
 
-# user:role many:many mapping table
-user_role_map = Table(
-    'user_role_map',
-    Base.metadata,
-    Column('user_id', ForeignKey('users.id', ondelete='CASCADE'), primary_key=True),
-    Column('role_id', ForeignKey('roles.id', ondelete='CASCADE'), primary_key=True),
-)
+# lots of things have roles
+# mapping tables are the same for all of them
 
-# service:role many:many mapping table
-service_role_map = Table(
-    'service_role_map',
-    Base.metadata,
-    Column(
-        'service_id', ForeignKey('services.id', ondelete='CASCADE'), primary_key=True
-    ),
-    Column('role_id', ForeignKey('roles.id', ondelete='CASCADE'), primary_key=True),
-)
+_role_map_tables = []
 
-# token:role many:many mapping table
-api_token_role_map = Table(
-    'api_token_role_map',
-    Base.metadata,
-    Column(
-        'api_token_id',
-        ForeignKey('api_tokens.id', ondelete='CASCADE'),
-        primary_key=True,
-    ),
-    Column('role_id', ForeignKey('roles.id', ondelete='CASCADE'), primary_key=True),
-)
-
-# group:role many:many mapping table
-group_role_map = Table(
-    'group_role_map',
-    Base.metadata,
-    Column('group_id', ForeignKey('groups.id', ondelete='CASCADE'), primary_key=True),
-    Column('role_id', ForeignKey('roles.id', ondelete='CASCADE'), primary_key=True),
-)
+for has_role in (
+    'user',
+    'group',
+    'service',
+    'api_token',
+    'oauth_client',
+    'oauth_code',
+):
+    role_map = Table(
+        f'{has_role}_role_map',
+        Base.metadata,
+        Column(
+            f'{has_role}_id',
+            ForeignKey(f'{has_role}s.id', ondelete='CASCADE'),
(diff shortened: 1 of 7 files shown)


minrk:
@0mar @IvanaH8 I made some changes to `_get_subscopes` to ensure that `self` is resolved there instead of `expand_roles_to_scopes`. I encountered test failures when requesting a token with a role that includes 'self', claiming that it exceeded the permissions of the owner - which had the same role! The reason was `expand_roles_to_scopes` removed and resolved `self` while `_token_allowed_role` also called `_get_subscopes` but didn't handle `self`, leading to an inaccurate comparison of scopes.

I also added a log statement to show exactly *what* extra but disallowed scopes were being requested.

jupyterhub/roles.py
```diff
@@ -163,22 +163,22 @@ def _expand_subscopes(index):
 def expand_roles_to_scopes(orm_object):
     """Get the scopes listed in the roles of the User/Service/Group/Token
     If User, take into account the user's groups roles as well"""
+    if not isinstance(orm_object, orm.Base):
```
minrk, line 166: This type check allows us to ignore the User wrapper

jupyterhub/roles.py
```diff
@@ -380,32 +385,36 @@ def _token_allowed_role(db, token, role):
...
+    raw_owner_scopes = {
+        scope.split('!', 1)[0] if '!' in scope else scope for scope in owner_scopes
+    }
+    disallowed_scopes = raw_scopes.difference(raw_owner_scopes)
+    if not disallowed_scopes:
+        # no scopes requested outside owner's own scopes
+        return True
     else:
-        raise ValueError('Owner the token %r not found', token)
+        app_log.warning(
```
minrk, line 414: Logging exactly *what* is disallowed is really helpful for debugging

jupyterhub/scopes.py
```diff
@@ -18,7 +18,18 @@ def get_scopes_for(orm_object):
     scopes = set()
     if orm_object is None:
         return scopes
-    elif isinstance(orm_object, orm.APIToken):
+
+    if not isinstance(orm_object, orm.Base):
+        from .user import User
+
+        if isinstance(orm_object, User):
+            orm_object = orm_object.orm_user
```
minrk, line 26: This is the only place we need to handle the User wrapper, and we immediately cast it to the orm object so everything else gets the lower-level objects

0mar (approved):

jupyterhub/orm.py
```diff
@@ -137,43 +137,34 @@ def __repr__(self):
...
+            ForeignKey(f'{has_role}s.id', ondelete='CASCADE'),
+            primary_key=True,
+        ),
+        Column(
+            'role_id',
+            ForeignKey('roles.id', ondelete='CASCADE'),
+            primary_key=True,
+        ),
+    )
+    _role_map_tables.append(role_map)
```
IvanaH8, line 167: This is great!

jupyterhub/roles.py
```diff
@@ -380,32 +385,36 @@ def _token_allowed_role(db, token, role):
     """Returns True if token allowed to have requested role through
     comparing the requested scopes with the set of token's owner scopes"""
 
-    standard_permissions = {'all', 'read:all'}
+    owner = token.user
+    if owner is None:
+        owner = token.service
```
IvanaH8, line 390: I love this, thanks! I thought we had to query db to find out who's the token owner.

IvanaH8 (approved):
Perfect, thanks for the fixes as well!

minrk:
Thanks for the reviews!

PR #3432, opened 2021-04-19
be strict about role names

minrk:
Role names must be:

- 3-255 characters
- ascii lowercase, numbers, 'unreserved' URL punctuation '-_.~'
- must start with a letter
- must end with letter or number

this lets us avoid url escaping issues, etc. in e.g. oauth params

### DIFF
--- a/jupyterhub/roles.py
+++ b/jupyterhub/roles.py
@@ -227,7 +227,7 @@ def _overwrite_role(role, role_dict):
                     )
 
 
-_role_name_pattern = re.compile(r'^[a-z][a-z0-9\-]{1,253}[a-z0-9]$')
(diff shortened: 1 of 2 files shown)


minrk:
@IvanaH8 @0mar if you want to have a look. This came up in #3431 where it would be good to not have to worry about certain kinds of issues if possible fields require escaping.

consideRatio (approved):
So nice to review a very tightly scoped PR =)

IvanaH8 (approved):
Nice! I'll add this to docs as well.

0mar:
Looks good!

0mar (approved):

PR #3433, opened 2021-04-19
Remove handling of jupyterhub 0.8 oauth client ids

minrk:
These only affected Hub deployments upgrading directly from 0.8 with still-running servers

0.8 was a long time ago, it's okay to require restarting servers for an upgrade that long


manics (approved):
This is way before my time, but the reasoning makes sense to me :smile:

consideRatio (approved):
This LGTM!

PR #3435, opened 2021-04-20
Fixed scope checking in UserTokenListAPIHandler

0mar:
Since the `UserTokenListAPIHandler` allows authentication with username and password, it bypasses some standard `APIHandler` setup. The handler has been patched so it checks for the relevant scopes. Also, we no longer distinguish between 'user not present' and 'user present but no valid scopes' for security purposes.

### DIFF
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -301,7 +301,6 @@ class UserTokenListAPIHandler(APIHandler):
 
         self.write(json.dumps({'api_tokens': api_tokens}))
 
-    # @needs_scope('users:tokens') #Todo: needs internal scope checking
     async def post(self, user_name):
         body = self.get_json_body() or {}
         if not isinstance(body, dict):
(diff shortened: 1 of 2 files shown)


jupyterhub/apihandlers/users.py
```diff
@@ -330,13 +329,16 @@ async def post(self, user_name):
...
-            raise web.HTTPError(403, "Only admins can request tokens for other users")
-        if not user:
-            raise web.HTTPError(404, "No such user: %s" % user_name)
-        if requester is not user:
-            kind = 'user' if isinstance(requester, User) else 'service'
+        kind = 'user' if isinstance(requester, User) else 'service'
+        scope_filter = self.get_scope_filter('users:tokens')
+        if user is None or not scope_filter(user, kind):
+            raise web.HTTPError(
+                404,
```
minrk, line 339: Can this still be 403?

The main reason we often have 404 is to avoid leaking user info in failed authentication requests, but as long as the error code is the same for both missing and insufficient permissions, I think 403 is better, since it more naturally fits the more *likely* case of knowing the user is there, but not having the right permissions.

0mar: Good point, thanks. I'll change it back

minrk:
Nice!

PR #3436, opened 2021-04-21
ci: github workflow security, pin action to sha etc

consideRatio:
Pin references to github actions we rely on in workflows with jobs that reference GitHub secrets that could get exposed.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -89,27 +89,32 @@ jobs:
 
       - uses: actions/checkout@v2
 
-      # Setup docker to build for multiple platforms (requires qemu).
-      # See:
-      # https://github.com/docker/build-push-action/tree/v2.3.0#usage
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -89,27 +89,32 @@ jobs:
...
+        #    permissions to push to the relevant images:
+        #    https://hub.docker.com/orgs/jupyterhub/teams/bots/permissions
+        # 4. Registering the username and token as a secret for this repo:
+        #    https://github.com/jupyterhub/jupyterhub/settings/secrets/actions
         if: env.REGISTRY != 'localhost:5000/'
-        with:
-          username: ${{ secrets.DOCKERHUB_USERNAME }}
-          password: ${{ secrets.DOCKERHUB_TOKEN }}
+        run: |
+          docker login -u "${{ secrets.DOCKER_USERNAME }}" -p "${{ secrets.DOCKERHUB_TOKEN }}"
```
consideRatio, line 117: I removed use of the docker-login action as it felt trivial.

PR #3438, opened 2021-04-22
Roles and Scopes (RBAC)

### LINKED ISSUE #1057: Auth token scopes?

### LINKED ISSUE #2245: User roles (RBAC)

### LINKED ISSUE #3189: auth state visible to self

minrk:
Opening the big PR to merge RBAC for JupyterHub 2.0. It's not quite ready yet, but getting pretty close!

**[Preview docs](https://jupyterhub.readthedocs.io/en/rbac/rbac)**

todo items:

- [x] define scopes for all API endpoints
- [x] define roles in database
- [x] assign roles to users, services, tokens
- [x] resolve request->roles->scopes for auth on API requests
- [x] request roles in oauth flow
- [x] remove OAuthToken as a separate entity
- [x] document using RBAC for limiting permissions of services/users that previously had to be admin
- [x] configure requested roles for oauth clients
- [x] resolve strict `!user` filter for "owner-only" scopes
- [x] confirmation page for user confirmation of requested scopes on `oauth2/authorize` #3504
- [x] more examples and exercises to make sure we have the right scopes
- [x] generate (some) scopes docs from code
- [x] be very clear/precise about what happens in `load_roles` when role assignments may change, e.g. from config and from API requests
- [ ] <del>API endpoints for roles</del> (not for 2.0, I think roles coming only from config is appropriate for 2.0, and we can think about using the API to define roles at a later point.

This PR can cover general discussion of landing/finishing the RBAC work

closes #2245 
closes #1057 
closes #3189


minrk:
I added the rbac branch to readthedocs, so folks should be able to watch the latest rbac buidls of the docs [here](https://jupyterhub.readthedocs.io/en/rbac/rbac/index.html)

docs/source/rbac/generate-scope-table.py
```diff
@@ -0,0 +1,99 @@
...
+    @classmethod
+    def create_writer(cls, table_name, headers, values):
+        writer = MarkdownTableWriter()
+        writer.table_name = table_name
+        writer.headers = headers
+        writer.value_matrix = values
+        writer.margin = 1
+        return writer
+
+    def _get_scope_relationships(self):
```
davidbrochart, line 24: I'm not sure this is an issue, but this functions generates duplicates, e.g. `('read:users', 'read:users:activity')` and `('users:activity', 'read:users:activity')`, which can be seen in the generated table:

image

IvanaH8: @davidbrochart the duplication here is intentional, it is to show that the `read:users:activity` is a sub-scope of both `read:users` and `users:activity`, which are not themselves related. Is it confusing?

davidbrochart: > Is it confusing?

Maybe it could help to mention that this special scope can be specified using two different syntaxes, since I believe it is the only one?

IvanaH8: We will definitely mention this, there is another similar duplicate with `read:users:name` which is a sub-scope of both `read:users` and `read:users:servers` (again, not related). We will highlight the duplications and the reasons behind them in the docs, they are still being updated as we speak. Thank you for the inquiry and suggestion!

davidbrochart:
I don't understand why in the [available scopes' description](https://jupyterhub.readthedocs.io/en/rbac/rbac/scopes.html#available-scopes) we have:
- `users`: Grants read and write permissions to users' models apart from servers, tokens and authentication state.
- `users:tokens`: Grants read, write, create and delete permissions to users' tokens.

According to the definition of [vertical filtering](https://jupyterhub.readthedocs.io/en/rbac/rbac/scopes.html#vertical-filtering), `users:tokens` should have less permissions that `users`, so why do `users` permissions exclude tokens?

minrk:
@davidbrochart the vertical filtering does not define the relationships as being based on their strings, however many relationships do logically follow that way, so I can see how one might make that assumption. The relationships are defined in the hierarchy [below](https://jupyterhub--3504.org.readthedocs.build/en/3504/rbac/scopes.html#available-scopes). It is a *convention* that most sub-scopes happen to look this way, but it is neither a requirement that the sub-scopes be in the form `parent:filed` or `read:parent`, nor that `parent` imply `parent:field`.

That said, it may help clarify some of these relationships if the 'elevated' scopes, such as `read:user:roles` don't use the `user:subfield` form. There is no technical reason for this, but it may help better *communicate* the relationships. With anything permission related, clarity is a security issue.

Examples that currently don't match the hierarchy:

- `users:servers` could be `servers`
- `users:tokens` could be `tokens`
- `users:roles` could be `roles:users`, `roles:services`, etc.

davidbrochart:
Thanks for clarifying @minrk !

minrk:
I believe this is now ready to land in main. Great work @0mar and @IvanaH8!

It's a massive change, so I don't expect anyone to review all the lines of code in this PR. That's already been done in the many smaller PRs that make up this one (no unreviewed commits are in this PR other than conflict resolution with main).

What I would encourage is a look at the [docs for the new features](https://jupyterhub--3438.org.readthedocs.build/en/3438/rbac/) to ask any questions and land this promptly

0mar:
> (...)
> That said, it may help clarify some of these relationships if the 'elevated' scopes, such as `read:user:roles` don't use the `user:subfield` form. There is no technical reason for this, but it may help better _communicate_ the relationships. With anything permission related, clarity is a security issue.
> 
> Examples that currently don't match the hierarchy:
> 
>     * `users:servers` could be `servers`
> 
>     * `users:tokens` could be `tokens`
> 
>     * `users:roles` could be `roles:users`, `roles:services`, etc.

@davidbrochart This particular issue has been addressed in https://github.com/jupyterhub/jupyterhub/pull/3513

minrk:
Given that this has been reviewed piecemeal by various folks, I'll plan to land this end of next week unless I hear otherwise from folks.

docs/source/rbac/generate-scope-table.py
```diff
@@ -0,0 +1,125 @@
...
+
+        used for creating hierarchical scope table in _parse_scopes()
+        """
+        pairs = []
+        for scope in self.scopes.keys():
+            if self.scopes[scope].get('subscopes'):
+                for subscope in self.scopes[scope]['subscopes']:
+                    pairs.append((scope, subscope))
+            else:
+                pairs.append((scope, None))
```
sgaist, line 40: This might be nitpicking but wouldn't it be simpler to use something like:
```python
for scope, data in self.scopes.items():
    subscopes = data.get('subscopes')
    if subscopes is not None:
        for subscope in subscopes:
            pairs.append((scope, subscope))
    else:
        pairs.append((scopes, None))
```

This would avoid multiple accesses to the various elements of the dictionnaries.

minrk: Yeah, that's a fine improvement. Want to make a PR?

sgaist: Sure thing

minrk: PR merged. Thanks!

ci/init-db.sh
```diff
@@ -20,7 +20,7 @@ fi
 
 # Configure a set of databases in the database server for upgrade tests
 set -x
-for SUFFIX in '' _upgrade_072 _upgrade_081 _upgrade_094; do
+for SUFFIX in '' _upgrade_100 _upgrade_122 _upgrade_130; do
```
consideRatio, line 23: Is it correct that this implies that you are required to upgrade to 1.* before upgrading to jupyterhub 2.* (assuming 2.0 is the next released version?)

minrk: Yes, that's exactly right.

consideRatio: Thanks!

minrk:
With #3520 merged, I'm going to land this. Great work, @IvanaH8 and @0mar!

consideRatio:
Wieeeeeeee amazing work @minrk @ivanah8 @0mar!!! ♥️🥳🌟

PR #3439, opened 2021-04-23
Add service.oauth_roles configuration

0mar:
PR that adds a list of allowed roles for API tokens from a service when service acts as OAuth client (see #3431)

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -2127,7 +2127,6 @@ class JupyterHub(Application):
             name = spec['name']
             # get/create orm
             orm_service = orm.Service.find(self.db, name=name)
-            allowed_roles = spec.get('allowed_roles', [])
             if orm_service is None:
                 # not found, create a new one
                 orm_service = orm.Service(name=name)
@@ -2192,7 +2191,7 @@ class JupyterHub(Application):
                     client_id=service.oauth_client_id,
                     client_secret=service.api_token,
                     redirect_uri=service.oauth_redirect_uri,
-                    allowed_roles=allowed_roles,
+                    allowed_roles=service.oauth_roles,
                     description="JupyterHub service %s" % service.name,
                 )
 
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -111,14 +111,12 @@ from .objects import Hub, Server
 # For faking stats
 from .emptyclass import EmptyClass
 
-
 common_aliases = {
     'log-level': 'Application.log_level',
     'f': 'JupyterHub.config_file',
     'config': 'JupyterHub.config_file',
     'db': 'JupyterHub.db_url',
 }
-
 
 aliases = {
     'base-url': 'JupyterHub.base_url',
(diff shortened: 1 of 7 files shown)


jupyterhub/app.py
```diff
@@ -2129,6 +2127,7 @@ def init_services(self):
             name = spec['name']
             # get/create orm
             orm_service = orm.Service.find(self.db, name=name)
+            allowed_roles = spec.get('allowed_roles', [])
```
minrk, line 2130: This isn't necessary, as the trait you added loads it onto the service object automatically.

jupyterhub/oauth/provider.py
```diff
@@ -586,7 +586,9 @@ def __init__(self, db, validator, *args, **kwargs):
         self.db = db
         super().__init__(validator, *args, **kwargs)
 
-    def add_client(self, client_id, client_secret, redirect_uri, description=''):
+    def add_client(
+        self, client_id, client_secret, redirect_uri, allowed_roles=None, description=''
```
minrk, line 590: 👍

jupyterhub/app.py
```diff
@@ -2193,6 +2192,7 @@ def init_services(self):
                     client_id=service.oauth_client_id,
                     client_secret=service.api_token,
                     redirect_uri=service.oauth_redirect_uri,
+                    allowed_roles=allowed_roles,
```
minrk, line 2195: This is already taken care of by the `input=True` trait added to services:

```suggestion
                    allowed_roles=service.allowed_roles,
```

jupyterhub/user.py
```diff
@@ -568,7 +568,7 @@ async def spawn(self, server_name='', options=None, handler=None):
                 client_id,
                 api_token,
                 url_path_join(self.url, server_name, 'oauth_callback'),
-                description="Server at %s"
+                description="Server at %s"  # todo: own server: 'users:servers!user={username} as allowed role?
```
minrk, line 571: I think the todo here is make server allowed_roles configurable.

Logical question: should this be a Spawner configuration or an Authenticator configuration? This is an open question!

jupyterhub/services/service.py
```diff
@@ -189,6 +190,13 @@ class Service(LoggingConfigurable):
         """
     ).tag(input=True)
 
+    allowed_roles = List(
```
minrk, line 193: Maybe let's call this `oauth_roles`? I think it might be a little clearer to have oauth in the name to communicate that these are the allowed (and requested by default) roles for this service *when it functions as an oauth client*.

Then it'll be `add_oauth_client(allowed_roles=service.oauth_roles)`

jupyterhub/app.py
0mar: Thanks, didn't catch that!

jupyterhub/services/service.py
0mar: Good idea, I'll fix that

jupyterhub/user.py
0mar: For now, as discussed, I'll go for adding the roles on a Spawner configuration, then we can evaluate the setup later

minrk:
Great!

PR #3444, opened 2021-04-23
Oauth details docs

### LINKED ISSUE #3155: Documentation challenge: explain JupyterHub's double-oauth
A relatively common source of confusion when deploying and configuring oauth for jupyterhub authentication, is that jupyterhub *also* uses oauth to implement authentication to services. The outer identity provider is configurable via `JupyterHub.authenticator_class`, but the internal oauth is just how jupyterhub works, and not optional or configurable.

I think there are two main sources of confusion, but there certainly could be more:

1. one of these oauths is configurable, the other is not
2. since there are two oauths happening, there's two of everything (two redirects, two callback urls, etc.) so one can get lost

It looks a bit like:

service -> internal oauth (always) -> hub -> external oauth (configuable! Maybe not oauth!) -> GitHub, Google, etc. -> hub oauth callback (if using oauth) -> service oauth callback (always) -> authenticated!

I don't know where folks could get lost in there :)

### Proposed change

Add documentation clearly illustrating (with pictures!) and explaining how auth works for a service when using an OAuth provider via OAuthenticator. This is proposed here, not on oauthenticator docs, because the internal oauth is always on and understanding can still be a challenge when the Authenticator doesn't itself use oauth ("I'm not using OAuthenticator, why is oauth happening?").

Readers looking to understand jupyterhub authentication, configure oauth with services, etc. Also contributors to authentication getting an overview of how things work

minrk:
I finally decided to put in some time documenting the whole oauth flow. It's a lot!

This is a detailed sequence of (almost) every request involved in authenticating with a single-user server when the Hub itself is using OAuthenticator.

I still need to add some discussion of the distinction between the tokens issued to the browser (user token from oauth flow) and the server (api token issued at launch).

Still some to fill out, but I need to pause here so opening a WIP.

closes #3155

### DIFF
--- a/docs/source/reference/oauth.md
+++ b/docs/source/reference/oauth.md
@@ -1,0 +1,214 @@
+# JupyterHub and OAuth
+
+JupyterHub uses OAuth 2 internally as a mechanism for authenticating users.
+As such, JupyterHub itself always functions as an OAuth **provider**.
+More on what that means below.
+
+Additionally, JupyterHub is _often_ deployed with [oauthenticator](https://oauthenticator.readthedocs.io),
+where an external identity provider, such as GitHub or KeyCloak, is used to authenticate users.
+When this is the case, there are \*two nested
+
+This means that when you are using JupyterHub, there is always _at least one_ and often two layers of OAuth involved in a user logging in and accessing their server.
+
+Some relevant points:
+
+- Single-user servers _never_ need to communicate with or be aware of the upstream provider.
+  As far as they are concerned, only JupyterHub is an OAuth provider,
+  and how users authenticate with the Hub itself is irrelevant.
+- When talking to a single-user server,
+  there are ~always two tokens:
+  a token issued to the server itself to communicate with the Hub API,
+  and a second per-user token in the browser to represent the completed login process and authorized permissions.
+  More on this later.
+
+### Key OAuth terms
+
+- **provider** the entity responsible for managing.
+  JupyterHub is _always_ an oauth provider for its components.
+  When OAuthenticator is used, an external service, such as GitHub or KeyCloak, is also an oauth provider.
+- **client** An entity that requests OAuth tokens on a user's behalf.
+  JupyterHub _services_ or single-user _servers_ are OAuth clients of the JupyterHub _provider_.
+  When OAuthenticator is used, JupyterHub is itself also an OAuth _client_ for the external oauth _provider_, e.g. GitHub.
+- **browser** A user's web browser, which makes requests and stores things like cookies to
+- **token** The secret value used to represent a user's authorization. This is the final product of the OAuth process
+
+### The oauth flow
+
+OAuth flow is what we call the sequence of HTTP requests involved in authenticating a user and issuing a token.
+
+It generally goes like this:
+
+1. A _browser_ makes an HTTP request to an oauth _client_.
+2. There are no credentials, so the client _redirects_ the browser to an "authorize" page on the oauth _provider_ with some extra information:
+   - the oauth **client id** of the client itself
+   - the **redirect uri** to be redirected back to after completion
+   - the **scopes** requested, which the user should be presented with to confirm.
+     This is the "X would like to be able to Y on your behalf. Allow this?" page you see on all the "Login with ..." pages around the Internet.
+3. During this authorize step,
+   the browser must be _authenticated_ with the provider.
+   This is often already stored in a cookie,
+   but if not the provider webapp must begin its _own_ authentication process before serving the authorization page.
+4. After the user tells the provider that they want to proceed with the authorization,
+   the provider records this authorization in a short-lived record called an **oauth code**.
+5. Finally,
+   the oauth provider redirects the browser _back_ to the oauth client's "redirect uri"
+   (or "oauth callback uri"),
+   with the oauth code in a url parameter.
+
+At this point:
+
+- The browser is authenticated with the _provider_
+- The user's authorized permissions are recorded in an _oauth code_
+- The _provider_ knows that the given oauth client's requested permissions have been granted, but the client doesn't know this yet.
+- All requests so far have been made directly by the browser.
+  No requests have originated at the client or provider.
+
+Now we get to finish the OAuth process.
+Let's dig into what the oauth client does when it handles
+the oauth callback request with the
+
+- The OAuth client receives the _code_ and makes an API request to the _provider_ to exchange the code for a real _token_.
+  This is the first direct request between the OAuth _client_ and the _provider_.
+- Once the token is retrieved, the client _usually_
+  makes a second API request to the _provider_
+  to retrieve information about the owner of the token (the user)
+- Finally, the oauth client stores its own record that the user is authorized in a cookie.
+  This could be the token itself, or any other appropriate representation of successful authentication.
+
+_phew_
+
+So that's _one_ OAuth process.
+
+## Full sequence of OAuth in JupyterHub
+
+Let's go through the above oauth process in Jupyter,
+with specific examples of each HTTP request and what information is contained.
+
+Our starting point:
+
+- a user's single-user server is running. Let's call them `danez`
+- jupyterhub is running with GitHub as an oauth provider,
+- Danez has a fresh browser session with no cookies yet
+
+First request:
+
+- browser->single-user server
+- `GET /user/danez/notebooks/mynotebook.ipynb`
+- no credentials, so client starts oauth process with JupyterHub
+- response: 302 redirect -> `/hub/api/oauth2/authorize`
+  with:
+  - client-id=`jupyterhub-user-danez`
+  - redirect-uri=`/user/danez/oauth_callback` (we'll come back later!)
+
+Second request, following redirect:
+
+- browser->jupyterhub
+- `GET /hub/api/oauth2/authorize`
+- no credentials, so jupyterhub starts oauth process _with GitHub_
+- response: 302 redirect -> `/hub/api/oauth2/authorize`
+  with:
+  - client-id=`jupyterhub-client-uuid`
+  - redirect-uri=`/hub/oauth_callback` (we'll come back later!)
+
+Third request, following redirect:
+
+- browser->GitHub
+- `GET https://github.com/login/oauth/authorize`
+
+Prompts for login and asks for confirmation of authorization.
+
+After successful authorization
+(either by looking up a pre-existing authorization,
+or recording it via form submission)
+GitHub issues oauth code and redirects to `/hub/oauth_callback?code=github-code`
+
+Next request:
+
+- browser->JupyterHub
+- `GET /hub/oauth_callback?code=github-code`
+
+Inside the callback handler, JupyterHub makes two API requests:
+
+The first:
+
+- JupyterHub->GitHub
+- `POST https://github.com/login/oauth/access_token`
+- request made with oauth code from url parameter
+- response includes an Access token
+
+The second:
+
+- JupyterHub->GitHub
+- `GET https://api.github.com/user`
+- request made with access token in the `Authorization` header
+- response is the user model, including username, email, etc.
+
+Now the oauth callback request completes with:
+
+- set cookie on `/hub/` recording jupyterhub authentication so we don't need to do oauth with github again for a while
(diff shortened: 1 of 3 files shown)


minrk:
Some notes from gitter chat with @rajeshK88 to add coverage for here: discuss what happens when auth *expires* at one or both levels, and how *either one* can expire first:

- [oauth_token_expires_in](https://jupyterhub.readthedocs.io/en/stable/api/app.html?highlight=oauth_expires_in#jupyterhub.app.JupyterHub.oauth_token_expires_in) governs expiry of internal oauth
- [Authenticator.refresh_user](https://jupyterhub.readthedocs.io/en/stable/api/auth.html#jupyterhub.auth.Authenticator.refresh_user) (and [refresh_pre_spawn](https://jupyterhub.readthedocs.io/en/stable/api/auth.html#jupyterhub.auth.Authenticator.refresh_pre_spawn) and [auth_refresh_age](https://jupyterhub.readthedocs.io/en/stable/api/auth.html#jupyterhub.auth.Authenticator.auth_refresh_age)) govern expiry of *upstream* auth (e.g. OAuthenticator) and when the external provider needs to be checked with again.

docs/source/reference/oauth.md
```diff
@@ -0,0 +1,214 @@
+# JupyterHub and OAuth
+
+JupyterHub uses OAuth 2 internally as a mechanism for authenticating users.
+As such, JupyterHub itself always functions as an OAuth **provider**.
+More on what that means below.
```
willingc, line 5: Let's link to the section where it is discussed if possible.

docs/source/reference/oauth.md
```diff
@@ -0,0 +1,214 @@
...
+
+Additionally, JupyterHub is _often_ deployed with [oauthenticator](https://oauthenticator.readthedocs.io),
+where an external identity provider, such as GitHub or KeyCloak, is used to authenticate users.
+When this is the case, there are \*two nested
```
willingc, line 9: Sentence fragment

docs/source/reference/oauth.md
```diff
@@ -0,0 +1,214 @@
...
+Some relevant points:
+
+- Single-user servers _never_ need to communicate with or be aware of the upstream provider.
+  As far as they are concerned, only JupyterHub is an OAuth provider,
+  and how users authenticate with the Hub itself is irrelevant.
+- When talking to a single-user server,
+  there are ~always two tokens:
+  a token issued to the server itself to communicate with the Hub API,
+  and a second per-user token in the browser to represent the completed login process and authorized permissions.
+  More on this later.
```
willingc, line 22: Perhaps link to later.

docs/source/reference/oauth.md
```diff
@@ -0,0 +1,214 @@
...
+
+### Key OAuth terms
+
+- **provider** the entity responsible for managing.
+  JupyterHub is _always_ an oauth provider for its components.
```
willingc, line 27:
```suggestion
  JupyterHub is _always_ an oauth provider for JupyterHub's components.
```

docs/source/reference/oauth.md
```diff
@@ -0,0 +1,214 @@
...
+  When OAuthenticator is used, an external service, such as GitHub or KeyCloak, is also an oauth provider.
+- **client** An entity that requests OAuth tokens on a user's behalf.
+  JupyterHub _services_ or single-user _servers_ are OAuth clients of the JupyterHub _provider_.
+  When OAuthenticator is used, JupyterHub is itself also an OAuth _client_ for the external oauth _provider_, e.g. GitHub.
+- **browser** A user's web browser, which makes requests and stores things like cookies to
```
willingc, line 32:
```suggestion
- **browser** A user's web browser, which makes requests and stores things like cookies
```

docs/source/reference/oauth.md
```diff
@@ -0,0 +1,214 @@
...
+- **token** The secret value used to represent a user's authorization. This is the final product of the OAuth process
```
willingc, line 33:
```suggestion
- **token** The secret value used to represent a user's authorization. This is the final product of the OAuth process.
```

docs/source/reference/oauth.md
```diff
@@ -0,0 +1,214 @@
...
+
+### The oauth flow
+
+OAuth flow is what we call the sequence of HTTP requests involved in authenticating a user and issuing a token.
```
willingc, line 37:
```suggestion
OAuth flow is what we call the sequence of HTTP requests involved in authenticating a user (or service) and issuing a token.
```

docs/source/reference/oauth.md
```diff
@@ -0,0 +1,214 @@
...
+
+It generally goes like this:
+
```
willingc, line 40:
```suggestion

#### Oauth request and redirect
```

docs/source/reference/oauth.md
```diff
@@ -0,0 +1,214 @@
...
+   This is often already stored in a cookie,
+   but if not the provider webapp must begin its _own_ authentication process before serving the authorization page.
+4. After the user tells the provider that they want to proceed with the authorization,
+   the provider records this authorization in a short-lived record called an **oauth code**.
+5. Finally,
+   the oauth provider redirects the browser _back_ to the oauth client's "redirect uri"
+   (or "oauth callback uri"),
+   with the oauth code in a url parameter.
+
+At this point:
```
willingc, line 58:
```suggestion
#### State after redirect
At this point:
```

docs/source/reference/oauth.md
```diff
@@ -0,0 +1,214 @@
...
+
+- The browser is authenticated with the _provider_
+- The user's authorized permissions are recorded in an _oauth code_
+- The _provider_ knows that the given oauth client's requested permissions have been granted, but the client doesn't know this yet.
+- All requests so far have been made directly by the browser.
+  No requests have originated at the client or provider.
+
+Now we get to finish the OAuth process.
```
willingc, line 66:
```suggestion
#### OAuth Client Handles Callback Request
Now we get to finish the OAuth process.
```

docs/source/reference/oauth.md
```diff
@@ -0,0 +1,214 @@
...
+
+Our starting point:
+
+- a user's single-user server is running. Let's call them `danez`
+- jupyterhub is running with GitHub as an oauth provider,
+- Danez has a fresh browser session with no cookies yet
+
+First request:
+
+- browser->single-user server
```
willingc, line 95:
```suggestion
- browser with JupyterLab or Jupyter Classic->single-user server
```

docs/source/reference/oauth.md
```diff
@@ -0,0 +1,214 @@
...
+- `GET /hub/oauth_callback?code=github-code`
+
+Inside the callback handler, JupyterHub makes two API requests:
+
+The first:
+
+- JupyterHub->GitHub
+- `POST https://github.com/login/oauth/access_token`
+- request made with oauth code from url parameter
+- response includes an Access token
```
willingc, line 137:
```suggestion
- response includes an access token
```

willingc (commented):
Great start on this @minrk. Thanks!

docs/source/reference/oauth.md
```diff
@@ -0,0 +1,217 @@
...
+  More on this later.
+
+### Key OAuth terms
+
+- **provider** the entity responsible for managing.
+  JupyterHub is _always_ an oauth provider for JupyterHub's components.
+  When OAuthenticator is used, an external service, such as GitHub or KeyCloak, is also an oauth provider.
+- **client** An entity that requests OAuth tokens on a user's behalf.
+  JupyterHub _services_ or single-user _servers_ are OAuth clients of the JupyterHub _provider_.
+  When OAuthenticator is used, JupyterHub is itself also an OAuth _client_ for the external oauth _provider_, e.g. GitHub.
```
echarles, line 31: `When OAuthenticator is used, JupyterHub is itself also an OAuth _client_ for the external oauth _provider_, e.g. GitHub.`: Is this really the case? For e.g. GitHub, the oauth client is the effective user (eg. @minrk), not JupyterHub.

docs/source/reference/oauth.md
echarles: I guess it will be something like `..., there are two nested oauth flows. For example, for a user launching a server with GitHub authentication enabled, the first oauth flow to start will be the user authenticating against jupyterhub. Before completion, of that first flow, the GitHub oauth flow needs to be performed (the second flow). The detailed nested steps are detailed later on in this page`.

docs/source/reference/oauth.md
```diff
@@ -0,0 +1,217 @@
...
+- a user's single-user server is running. Let's call them `danez`
+- jupyterhub is running with GitHub as an oauth provider,
+- Danez has a fresh browser session with no cookies yet
+
+First request:
+
+- browser->single-user server running JupyterLab or Jupyter Classic
+- `GET /user/danez/notebooks/mynotebook.ipynb`
+- no credentials, so client starts oauth process with JupyterHub
+- response: 302 redirect -> `/hub/api/oauth2/authorize`
```
echarles, line 101: `no credentials, so client starts oauth process with JupyterHub` : In this sentence (and in the other of this explanation), the word `client` may be ambiguous. Does it mean the effective end-user, or is it the oauth client ?

echarles: `client` is defined above: - **client** An entity that requests OAuth tokens on a user's behalf.
  JupyterHub _services_ or single-user _servers_ are OAuth clients of the JupyterHub _provider_.
  When OAuthenticator is used, JupyterHub is itself also an OAuth _client_ for the external oauth _provider_, e.g. GitHub.

Sticking to that definition, some of my above comments are answered and the current text is OK. It is just a bit hard to convert a physical user to those OAuth clients. Maybe an introductory text that stresses on that aspect would help.

docs/source/reference/oauth.md
echarles: mmh ok, rereading makes more sense. For GitHub example, JupyterHub (the oauth client) asks GitHub 'me, JupyterHub, I ask you to give me a token so I can grab @minrk identity`. Therefor, @minrk needs to authenticate on GitHub and agree that JupyterHub has a token to query @minrk identity later on. In that sense, the client in that specific OAuth flow is indeed JupyterHub, and not @minrk

docs/source/reference/oauth.md
```diff
@@ -0,0 +1,217 @@
...
+  JupyterHub _services_ or single-user _servers_ are OAuth clients of the JupyterHub _provider_.
+  When OAuthenticator is used, JupyterHub is itself also an OAuth _client_ for the external oauth _provider_, e.g. GitHub.
+- **browser** A user's web browser, which makes requests and stores things like cookies
+- **token** The secret value used to represent a user's authorization. This is the final product of the OAuth process.
+
+### The oauth flow
+
+OAuth flow is what we call the sequence of HTTP requests involved in authenticating a user and issuing a token, ultimately used for authorized access to a service or single-user server.
+
+It generally goes like this:
```
echarles, line 39: Maybe restate that we have here 2 nested oauth flows.

docs/source/reference/oauth.md
minrk: Right. Probably worth linking to and/or cribbing from [this oauth terminology reference](https://www.oauth.com/oauth2-servers/definitions/). Users are never Clients in this language.

docs/source/reference/oauth.md
minrk: Since it's an oauth keyword, maybe we should keep it bold and/or capitalized ot make it clear that client is a specific term meaning "the application using jupyterhub for authentication, registered with a client id" - i.e. the single-user server in this case.

I'll give the terminology intro another pass with this in mind.

docs/source/reference/oauth.md
minrk: This is a general description of what one oauth flow is, so I don't think it makes sense to mention that there are two at this point in the doc. That comes later.

minrk:
Thanks for the review @willingc and @echarles ! I've given it another pass, and I think it's got all of what I want in there for now.

docs/source/reference/oauth.md
echarles: Makes sense.

echarles (approved):
Great doc @minrk Thx a lot.

PR #3446, opened 2021-04-24
[rbac] Add !user filter for "owner-only" scopes

IvanaH8:
Ref.: #3438 

- Implements `!user` horizontal filter for scopes which resolves into `!user=<username>` limiting the associated scope permissions to the user's own resources. Any scope can be customized with this filter.

- Default `server` role now has `users:activity!user` scope and is assigned to server tokens.

_Additional improvements:_

- Adds horizontal filters comparison when resolving token scope permissions. This avoids discarding valid token scopes with horizontal filters when the token owner scopes have no filter. 
For example, previously if user `admin` had scope `users:activity` and the token `users:activity!user=admin` (or vice versa), this was discarded when resolving token permissions as it was considering only equality. The resolving now includes the horizontal filter validity which retains the `users:activity!user=admin` scope for the token in the above example.

Includes tests.


0mar (approved):
Hey, thanks for taking care of this, it looks good!

jupyterhub/tests/test_roles.py
```diff
@@ -898,3 +941,52 @@ async def test_valid_names(name, valid):
...
+    ],
+)
+async def test_server_posting_activity(app, token_role, response):
+    user = add_user(app.db, app, name='test_user')
+    if token_role == 'no_role':
+        api_token = user.new_api_token(roles=[])
+    else:
+        api_token = user.new_api_token(roles=[token_role])
+
+    r = await api_request(
```
minrk, line 985: Can we also add a couple more tests to validate that the server scope doesn't grant more permissions than it should:

1. reading the owner's model from `GET /users` (should be able to read activity, but not other fields)
2. posting activity to another *existing* user should fail with 404, I think (maybe 403)

jupyterhub/user.py
```diff
@@ -531,7 +531,7 @@ async def spawn(self, server_name='', options=None, handler=None):
         orm_server = orm.Server(base_url=base_url)
         db.add(orm_server)
         note = "Server at %s" % base_url
-        api_token = self.new_api_token(note=note)
+        api_token = self.new_api_token(note=note, roles=['server'])
```
minrk, line 534: 👍

jupyterhub/roles.py
```diff
@@ -200,7 +216,7 @@ def _check_scopes(*args):
     """Check if provided scopes exist"""
 
     allowed_scopes = _get_scope_hierarchy()
-    allowed_filters = ['!user=', '!service=', '!group=', '!server=']
+    allowed_filters = ['!user=', '!service=', '!group=', '!server=', '!user']
```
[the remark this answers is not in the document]
IvanaH8: It is checking for the exact string including the equal sign in the first element and no equal sign in the last.

jupyterhub/tests/test_roles.py
IvanaH8: Yes, good point, thanks.

jupyterhub/scopes.py
```diff
@@ -13,6 +13,42 @@ class Scope(Enum):
...
+    for base in common_bases:
+        if owner_parsed_scopes[base] == Scope.ALL:
+            common_filters[base] = token_parsed_scopes[base]
+        elif token_parsed_scopes[base] == Scope.ALL:
+            common_filters[base] = owner_parsed_scopes[base]
+        else:
+            common_entities = (
+                owner_parsed_scopes[base].keys() & token_parsed_scopes[base].keys()
+            )
+            common_filters[base] = {
```
minrk, line 34: From in-person review, add notes here about what's not handled yet:

- filters intersect correctly with ALL and exact filter matches (i.e. `users!user=x & read:users:name` -> `read:users:name!user=x`)
- filters *do not* intersect with containing filters, i.e. `users!group=x` does not intersect with `users!user=y` even if user `y` is in group `x`. Same for user->server.

We should probably log a warning when this happens (group, user filters both present, result may be lower than intended permissions).

jupyterhub/tests/test_scopes.py
```diff
@@ -648,3 +648,62 @@ async def test_server_state_access(
             assert r.status_code == 403
         app.db.delete(group)
         app.db.commit()
+
+
+@mark.parametrize(
+    "name, user_scope, token_scope, discarded_own_scope, retained_owner_scope",
+    [
+        ('no_filter', 'users:activity', 'users:activity', False, False),
```
minrk, line 656: More combinations to test:

- subscopes cross-filter users!user=x, read:users:name
- multiple user filters ([users!user=x, users!user=y] vs [read:users:name!user=y])
- users!group=y vs users!user=x (won't do what we *want*, but good to have a test to ensure it does what we *expect*)
- users *and* groups, with user match: `users!group=x, users!user=y` vs `read:users!user=y, read:users:name!group=x`

jupyterhub/scopes.py
IvanaH8: Note and warning added.

jupyterhub/tests/test_scopes.py
IvanaH8: Thank you, the combinations were added.  

There is one more with `!user=` and `!server=` treated in the same way as `!group=` and `!user=` (no intersection with warning). As now any scope containing the `!server=` filter gets 'expanded' with `read:users:name!user=` scope, I think the `read:users:servers` should then have `read:users:name` as a subscope. It has been added in this PR to make sure that a user with `users:servers!user=x` scope is allowed to have token with `users:servers!server=y` (which checks for all subscopes ignoring horizontal filters) but their intersection (for now) returns an empty set.

The scope hierarchy needs to be updated according to latest docs (e.g., `admin:users` scopes should contain `users` scope), this will be added in tiny separate PR.

minrk:
Nice!

PR #3448, opened 2021-04-29
[rbac] Update scope hierarchy

IvanaH8:
1. Updates scope hierarchy:
- `admin:` scopes to be top level
- `read:users:servers` to include `read:users:name`
- `read:users` to include `read:users:roles` subscope
- `read:services` to have `read:services:name` and `read:services:roles` subscopes
2. Updates `admin` role scopes following the updated hierarchy
3. Changes `read:users:admin` to `read:users:roles` (same for services) in `identify_scopes()` and adds them to scope hierarchy


0mar (commented):
Looks good! I think this requires some changes to the user model and service model, so that the scopes return the correct values (and that for instance `users:roles` also allows access to the User(List)ApiHandler

IvanaH8:
Reverting the `read:users(services):roles` scopes for now, these will be added together with the required changes to the handlers in separate PR.

PR #3452, opened 2021-05-04
Fix documentation

davidbrochart:


docs/source/getting-started/authenticators-users-basics.md
```diff
@@ -107,8 +107,8 @@ with any provider, is also available.
 
 ## Use DummyAuthenticator for testing
 
-The :class:`~jupyterhub.auth.DummyAuthenticator` is a simple authenticator that
-allows for any username/password unless if a global password has been set. If
+The `DummyAuthenticator` is a simple authenticator that
+allows for any username/password unless a global password has been set. If
```
consideRatio, line 111: This is an improvement from the current state and we can improve it further at a later point if we want to make this a reference to documentation about the class or similar I think.

image

consideRatio (approved):
Nice this LGTM, thanks @davidbrochart for taking the time to contribute these fixes! :heart:

PR #3454, opened 2021-05-05
define Spawner.delete_forever on base Spawner

### LINKED ISSUE #3451: Implementation details about the delete_forever hook optionally on a Spawner object
#3337 introduced a `delete_forever` hook for the Spawner base class and I want to deliberate around some details.

The only change besides a test case is the following lines of code.

https://github.com/jupyterhub/jupyterhub/blob/3fec19d191119e4203c243f44c43e74c8fe62f04/jupyterhub/apihandlers/users.py#L241-L245

## Question 1

Do we want the Spawner base class to implement a placeholder function that only have `pass` in it? It would become an entrypoint to write documentation

## Question 2

If not a yes on Question 1, do we want this check added? Now we will write an error message for all Spawners no matter if they have this functionality.

```diff
# allow the spawner to cleanup any persistent resources associated with the user 
try: 
+   if user.spawner.delete_forever:
        await user.spawner.delete_forever() 
    except Exception as e: 
        self.log.error("Error cleaning up persistent resources: %s" % e) 
```

minrk:
...where I thought it already was! Instead of on the test class.

and fix the logic for when it is called a bit:

- call on *all* Spawners, not just the default
- call on named server deletion when remove=True

closes #3451, finishes #3337

### DIFF
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -242,11 +242,7 @@ class UserAPIHandler(APIHandler):
 
         await maybe_future(self.authenticator.delete_user(user))
 
-        # allow the spawner to cleanup any persistent resources associated with the user
-        try:
-            await user.spawner.delete_forever()
-        except Exception as e:
-            self.log.error("Error cleaning up persistent resources: %s" % e)
+        await user.delete_spawners()
 
         # remove from registry
         self.users.delete(user)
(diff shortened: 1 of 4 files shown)


consideRatio (approved):
Excellent docstring on the Spawner delete_forever method, this LGTM!

minrk:
Should probably do a 1.4.1 with this, since delete_forever was added there and doesn't quite work as designed. Can do that next week if there are any other bugs we've fixed since release.

PR #3456, opened 2021-05-05
avoid re-using asyncio.Locks across event loops

minrk:
should never occur in real applications where only one loop is run, but may occur in tests if the Proxy object lives longer than the loop that is running when it's created (imported?).

I *suspect* this is the source of our intermittent test failures with:

> got Future <Future pending> attached to a different loop

But since they are intermittent, it's hard to be sure, even if this PR passes.

The issue: we were allocating an asyncio.Lock(), which in turn grabs a handle on the current event loop, at *method definition time* in the decorator, instead of *call time*.

The solution: allocate the method at call time *and* double-check to ensure we never use a lock across event loops by storing the locks per-loop.

This should change nothing for 'real' hub instances, where only one loop is ever running, only tests where we start and stop loops a bunch.

### DIFF
--- a/jupyterhub/proxy.py
+++ b/jupyterhub/proxy.py
@@ -24,6 +24,7 @@ import time
 from functools import wraps
 from subprocess import Popen
 from urllib.parse import quote
+from weakref import WeakKeyDictionary
 
 from tornado.httpclient import AsyncHTTPClient
 from tornado.httpclient import HTTPError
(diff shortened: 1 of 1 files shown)


consideRatio (approved):
Thank you for explaining this @minrk, the event loops and such is a domain I'm not very confident about yet - this made sense though!

consideRatio:
The release workflow failed when logging into dockerhub.

image

The username wasn't set?

PR #3460, opened 2021-05-07
Fix role assignment on startup

### LINKED ISSUE #3441: [rbac] separate role creation from assignment
Right now, `init_roles` does two things:

1. define the roles, and
2. assign users, services, groups, tokens to those roles

The second part has to run after users and services exist, while the first part has to run before e.g. services with `admin: true` are created.

I just encountered this because jupyterhub with an admin service will fail to start because the admin role doesn't exist yet because `init_roles()` is called after `init_services()`.

I think we should rearrange this a bit to:

1. create roles before init_users or other entities
2. have a separate `init_role_assignments` stage that assigns entities to roles

And also probably *remove* support for assigning roles directly to tokens. Looking at it, I think that's maybe more complicated and niche than we need. Let's keep role assignment from config only to the other encapsulating entities.

Question: **Do we still have a single `load_roles` config with both role definitions and assignments, or two separate `load_roles` and `assign_roles` config inputs for the two stages?**

We should also make it *very* clear what happens between two jupyterhub starts with:

1. launch with load_roles with `{name: "myrole", users: ['alice', 'bob']}`
2. launch with load_roles with `{name: "myrole", users: ['alice'] }`

Does bob still have role `myrole` or not? If he does, how should I update bob's role assignment to remove `myrole`?

0mar:
Closes  #3441

Role creation is now separated from role assignment in hub initialization.
Any users and services in the config definition of a role are exactly the ones that the hub on startup assigns the role to.
This means that if first, we launch with hub config `load_roles` containing `{name: "myrole", users: ['alice', 'bob']}`
`alice` and `bob` are granted `myrole`.

If we restart the hub after adjusting the configuration file to launch with `load_roles` containing `{name: "myrole", users: ['alice'] }` `bob` no longer has role `myrole`.
In addition, assigning roles to tokens in the configuration file is no longer possible.

## `user` and `admin` roles
 - Any user present in the config list `allowed_users` gets the default `user` role
 - Any user gets a `user` and an `admin` role if either
   * listed in config as `admin_user` 
   * listed in `users` section of config `admin` role description 
   * has an admin flag in the database
 - Any  user with an admin role gets an admin flag in the database.
 -


jupyterhub/tests/test_roles.py
```diff
@@ -571,7 +571,8 @@ async def test_load_roles_groups(tmpdir, request):
     hub.init_db()
     db = hub.db
     await hub.init_groups()
-    await hub.init_roles()
+    await hub.init_role_creation()
```
minrk, line 574: Let's try to preserve the order - creation should happen before groups

jupyterhub/app.py
```diff
@@ -1966,25 +1965,27 @@ async def init_groups(self):
...
         default_roles = roles.get_default_roles()
         for role in default_roles:
-            roles.create_role(db, role)
+            roles.create_role(self.db, role)
 
+    async def init_role_assignment(self):
+        # tokens are added separately
+        role_bearers = ['users', 'services', 'groups']
+
+        db = self.db
```
minrk, line 1979: Let's also implement the admin role assignment here, instead of in `init_users`: query users and services with admin=True, then assign them to the admin role.

jupyterhub/app.py
```diff
@@ -2000,9 +2001,11 @@ async def init_roles(self):
...
-                            db, entity=orm_obj, rolename=predef_role['name']
-                        )
+                        orm_role_bearers.append(orm_obj)
+                        # roles.grant_role(
+                        #     db, entity=orm_obj, rolename=predef_role['name']
+                        # )
+                setattr(predef_role_obj, bearer, orm_role_bearers)
         # make sure that on no admin situation, all roles are reset
         admin_role = orm.Role.find(db, name='admin')
         if not admin_role.users:
```
minrk, line 2011: I don't think this is the correct check for upgrading default roles. The check should be whether the admin role existed *at all* in `init_role_creation`. No admin users is itself a valid state, so it doesn't indicate an upgrade is taking place. Then we can set a private flag to consume here.

jupyterhub/app.py
```diff
@@ -1966,25 +1965,27 @@ async def init_groups(self):
...
 
+    async def init_role_assignment(self):
+        # tokens are added separately
+        role_bearers = ['users', 'services', 'groups']
+
+        db = self.db
         # load predefined roles from config file
         self.log.debug('Loading predefined roles from config file to database')
         for predef_role in self.load_roles:
             roles.create_role(db, predef_role)
```
minrk, line 1983: `create_role` should be in role_creation, right?

jupyterhub/tests/test_roles.py
```diff
@@ -1046,3 +884,34 @@ async def test_oauth_allowed_roles(app, create_temp_role):
...
+            'scopes': ['users', 'groups'],
+            'users': other_users,
+        },
+    ]
+    for user_in_config in [False, True]:
+        if user_in_config:
+            roles_to_load[0]['users'].append(user_name)
+        kwargs = {'load_roles': roles_to_load}
+        hub = MockHub(**kwargs, clean_db=False)
+        hub.init_db()
```
minrk, line 906: I think init_db always starts with an empty database in MockHub, so I'm not sure if it's testing multiple initializations from one database. Try:

1. initialize and load roles
2. check presence
3. change `.load_roles` and re-run only `init_role_assignment()`
4. check membership again (it should change)

jupyterhub/app.py
```diff
@@ -329,11 +329,10 @@ def _load_classes(self):
             load_roles = [
                             {
                                 'name': 'teacher',
-                                'description': 'Access users information, servers and groups without create/delete privileges',
+                                'description': 'Access to users' information, servers and groups',
```
IvanaH8, line 332: The `users` scope doesn't include servers anymore, so we should either change the description or add the `users:servers` scope here (I know it's only example for config but better make it consistent I think).

jupyterhub/app.py
IvanaH8: Yes, also agree, role_creation should create the custom roles as well, not just default.

jupyterhub/tests/test_roles.py
```diff
@@ -1046,3 +884,34 @@ async def test_oauth_allowed_roles(app, create_temp_role):
...
+        hub.authenticator.admin_users = ['admin']
+        hub.authenticator.allowed_users = other_users + [user_name]
+        await hub.init_users()
+        await hub.init_role_creation()
```
IvanaH8, line 910: Will be changed following Min's comment above but just to point out that `init_role_creation` should always be before initializing users, services, groups, etc.

jupyterhub/app.py
0mar: If I'm not mistaken, that's covered a bit further down, in `roles.check_for_default_roles(db, bearer)`, which calls `roles.assign_default_roles` which queries for the admin flag

jupyterhub/tests/test_roles.py
0mar: Good idea, thanks

jupyterhub/app.py
0mar: Didn't consider that no admin users could be a valid situation but of course we can have other roles with elevated privileges as well, thanks

minrk:
Let's also make sure that we test that roles removed from load_roles are removed from the database:

1. add role to load_roles
2. init roles
3. remove role from load_roles
4. init roles again
5. assert role does not exist


(basically, assert on every axis that load_roles == roles in db after loading, no more, no less)

jupyterhub/app.py
0mar: Good point, moved it!

jupyterhub/tests/test_roles.py
0mar: Thanks!

jupyterhub/app.py
0mar: Good point, updated the information

jupyterhub/tests/test_roles.py
0mar: Thanks, rewrote the test

minrk:
Let's also double check that customizing the default roles (user, server) *doesn't* do something unexpected with membership. Setting only the user scopes shouldn't also change who has the user role.

minrk:
Another test case to check: restoring default roles:

1. load custom `user` role via config
2. remove custom user role from config, so it doesn't get loaded
3. start juptyerhub without custom user role
4. verify *original* user role definition is restored, without custom scopes (no effect on membership)

jupyterhub/app.py
minrk: It should be unconditional, though. Always assigning admin users to the admin role. Right now, that's only done in the `if not admin_role.users` case, which I think doesn't behave correctly (no users with admin role doesn't mean anything is wrong or that a db upgrade has occurred, because having no admins is valid.)

i.e. `user.admin` should always exactly match the admin role assignment, in both directions - no user should have `.admin = True` without being in the admin group, and no user should have the admin role without `.admin = True`.

0mar: Good point, I'll address it in the next commit

jupyterhub/app.py
```diff
@@ -1975,25 +1974,42 @@ async def init_groups(self):
...
         default_roles = roles.get_default_roles()
-        for role in default_roles:
-            roles.create_role(db, role)
+        default_role_names = [r['name'] for r in default_roles]
+        init_roles = default_roles + self.load_roles
+        if not orm.Role.find(self.db, name='admin'):
+            self._rbac_upgrade = True
+        for role in self.db.query(orm.Role):
+            if role.name not in default_role_names:
+                self.db.delete(role)
```
minrk, line 1987: It's a good idea to log any significant action like this:

```suggestion
                self.log.info(f"Deleting role not in configuration: {role.name}")
                self.db.delete(role)
```

jupyterhub/app.py
```diff
@@ -1975,25 +1974,42 @@ async def init_groups(self):
...
+            roles.create_role(self.db, role)
+
+    async def init_role_assignment(self):
+        # tokens are added separately
+        role_bearers = ['users', 'services', 'groups']
 
+        db = self.db
         # load predefined roles from config file
+        for role_spec in self.load_roles:
+            if role_spec['name'] == 'admin' and self.admin_users:
```
minrk, line 1999:
```python
if self.admin_users:
        if role_spec['name'] == 'admin':
            ...
```

jupyterhub/app.py
```diff
@@ -1975,25 +1974,42 @@ async def init_groups(self):
...
+                app_log.info(
+                    "Extending admin role assignment with config admin users: %s",
+                    str(self.admin_users),
+                )
+                role_spec['users'].extend(self.admin_users)
```
minrk, line 2004: Handle `'users'` being undefined in the original role spec (it could e.g. have only services):


```suggestion
                role_spec.setdefault('users', [])
                role_spec['users'].extend(self.admin_users)
```

jupyterhub/app.py
```diff
@@ -2139,7 +2156,9 @@ def init_services(self):
             if orm_service is None:
                 # not found, create a new one
                 orm_service = orm.Service(name=name)
-                if spec.get('admin', False):
+                if spec.get(
+                    'admin', False
+                ):  # Todo: fix double assignment of admin roles
```
minrk, line 2161: I think we can leave this with a deprecation warning:
```suggestion
                ):
                    self.log.warning(
                        f"Service {name} sets `admin: True`, which is deprecated in JupyterHub 2.0."
                        " You can assign now assign roles via `JupyterHub.load_roles` configuration."
                    )
```

jupyterhub/tests/test_services.py
```diff
@@ -89,10 +89,10 @@ async def test_external_service(app):
             }
         ]
         await maybe_future(app.init_services())
-        await maybe_future(app.init_roles())
+        await maybe_future(app.init_role_creation())
         await app.init_api_tokens()
         await app.proxy.add_all_services(app._service_map)
-        await app.init_roles()
+        await app.init_role_creation()
```
minrk, line 95:
```suggestion
        await app.init_role_assignment()
```

minrk (commented):
Looking great! I see a few test cases left to write with todos, but well on its way.

jupyterhub/app.py
```diff
@@ -1975,25 +1974,42 @@ async def init_groups(self):
...
         self.log.debug('Loading default roles to database')
         default_roles = roles.get_default_roles()
-        for role in default_roles:
-            roles.create_role(db, role)
+        default_role_names = [r['name'] for r in default_roles]
+        init_roles = default_roles + self.load_roles
+        if not orm.Role.find(self.db, name='admin'):
+            self._rbac_upgrade = True
+        for role in self.db.query(orm.Role):
+            if role.name not in default_role_names:
```
IvanaH8, line 1986: This will delete all roles apart from default roles from the database on every startup and then recreate the ones present in the config file again (in the code below). Shouldn't we only delete the roles not present in the config anymore ?

Then again, deleting the roles automatically strips the roles from all previous bearers so we can start with clean plate, reassigning from scratch. Is that the intention?

jupyterhub/tests/test_roles.py
```diff
@@ -1114,3 +952,78 @@ async def test_user_group_roles(app, create_temp_role):
...
+        },
+    ]
+    kwargs = {'load_roles': roles_to_load}
+    hub = MockHub(**kwargs)
+    hub.init_db()
+    hub.authenticator.admin_users = ['admin']
+    await hub.init_role_creation()
+    for role_conf in roles_to_load:
+        assert orm.Role.find(hub.db, name=role_conf['name'])
+    # Now reload and see if user is removed from role list
```
IvanaH8, line 977:
```suggestion
    # Now reload without 'elephant' role and see if it was removed from the db
```

jupyterhub/tests/test_roles.py
```diff
@@ -1114,3 +952,78 @@ async def test_user_group_roles(app, create_temp_role):
...
+    hub.authenticator.admin_users = ['admin']
+    hub.authenticator.allowed_users = user_names
+    await hub.init_role_creation()
+    await hub.init_users()
+    await hub.init_role_assignment()
+    user = orm.User.find(hub.db, name=user_name)
+    role = orm.Role.find(hub.db, name=role_name)
+    assert role in user.roles
+    # Now reload and see if user is removed from role list
+    roles_to_load[0]['users'].remove(user_name)
```
IvanaH8, line 1016: The 2nd definition of the same role will still have the `user_name` in `'users'` list but it gets ignored, correct?

jupyterhub/app.py
0mar: Agree, I'll take care of it

jupyterhub/app.py
0mar: The latter was indeed my intention, the reason being that I don't see a way to (efficiently) check if a defined role was previously differently defined in the configuration file. Since this actually also holds for the default roles, I went on to delete all roles prior to creating them

jupyterhub/tests/test_roles.py
0mar: Thanks!

jupyterhub/tests/test_services.py
0mar: Thanks for catching that

jupyterhub/app.py
0mar: Taken care of

jupyterhub/app.py
0mar: More efficient indeed, thanks

jupyterhub/app.py
0mar: Good to know, thanks. I forgot to take the `service` spec into account at all, actually. I'll push a fix for that as well

jupyterhub/tests/test_roles.py
0mar: Yes indeed. I suppose that this is the way it should be, or should we implement duplicate roles to be prohibited?

jupyterhub/app.py
IvanaH8: The `_overwrite_role()` called within `create_role()` in `roles.py` is checking if the 'description' and 'scopes' keys are different in the config than what is stored in db and overwrites them if yes. Maybe something like this? But starting from a clean plate completely reduces the chance of mismatch anywhere along the way so I am actually more inclined towards your solution here.

jupyterhub/tests/test_roles.py
0mar: Good idea, I'll implement it to be prohibited then

jupyterhub/app.py
0mar: Great, then I'll start from an empty role db on startup, thanks!

jupyterhub/tests/test_roles.py
```diff
@@ -1023,7 +1020,119 @@ async def test_config_role_users():
...
+    for user_name in user_names:
+        user = orm.User.find(hub.db, user_name)
+        assert user in user_role.users
+
+
+async def test_admin_role_respects_config():
+    role_spec = [
+        {
+            'name': 'admin',
+            'scopes': ['self', 'shutdown'],
```
IvanaH8, line 1125: Does this rewrite `admin` role's scopes? It shouldn't, the 'admin' role's definition and scopes should not be rewritable, the `create_role()` in `roles.py` should call `_overwrite_role()` which should raise error. Do the scopes get overwritten in the default `admin` role?

jupyterhub/app.py
```diff
@@ -1975,25 +1975,63 @@ async def init_groups(self):
...
-        # load default roles
         self.log.debug('Loading default roles to database')
         default_roles = roles.get_default_roles()
-        for role in default_roles:
-            roles.create_role(db, role)
+        config_role_names = [r['name'] for r in self.load_roles]
+        init_roles = self.load_roles.copy()
+        for role_name in config_role_names:
+            if config_role_names.count(role_name) > 1:
+                raise AttributeError(
```
minrk, line 1986:
```suggestion
                raise ValueError(
```

ValueError is the more appropriate error for when the value of something is a problem. AttributeError is generally reserved for attempting to access an undefined attribute.

More information on the [Python exceptions list](https://docs.python.org/3/library/exceptions.html).

jupyterhub/app.py
```diff
@@ -1975,25 +1975,63 @@ async def init_groups(self):
...
+                    f"Role {role_name} multiply defined. Please check the `load_roles` configuration"
+                )
+        for role_spec in default_roles:
+            if role_spec['name'] not in config_role_names:
+                init_roles.append(role_spec)
+        if not orm.Role.find(self.db, name='admin'):
+            self._rbac_upgrade = True
+        self.db.query(orm.Role).delete()
```
minrk, line 1994: I don't think we want to delete all the roles on every startup. This will make startup very expensive with lots of users, as we'll need to reassign potentially thousands of users to the `user` role on each launch.

Instead, just delete the roles whose name is not in config_role_names. Then it will be up to the role assignment step to ensure everybody gets the right assignments (and unassignments).

Deleting here will also unassign token roles, which won't get reassigned later (another case to test!).

jupyterhub/app.py
```diff
@@ -1975,25 +1975,63 @@ async def init_groups(self):
...
+        app_log.info("Deleting all roles in database")
+        self.db.commit()
+        for role in init_roles:
+            roles.create_role(self.db, role)
 
+    async def init_role_assignment(self):
+        # tokens are added separately
+        role_bearers = ['users', 'services', 'groups']
+        admin_role_objects = ['users', 'services']
+        config_admin_users = set(self.admin_users) | self.authenticator.admin_users
```
minrk, line 2004: Don't need a reference to the long-deprecated `self.admin_users` here. It's already copied with a warning into `authenticator.admin_users` by this point.

jupyterhub/app.py
```diff
@@ -1975,25 +1975,63 @@ async def init_groups(self):
...
+                    role_spec['users'] = set(role_spec.get('users', []))
+                    role_spec['users'] |= config_admin_users
         self.log.debug('Loading predefined roles from config file to database')
+        has_admin_role_spec = {role_bearer: False for role_bearer in admin_role_objects}
         for predef_role in self.load_roles:
-            roles.create_role(db, predef_role)
+            predef_role_obj = orm.Role.find(db, name=predef_role['name'])
+            for bearer in admin_role_objects:
+                if predef_role['name'] == 'admin':
+                    has_admin_role_spec[bearer] = bool(predef_role.get(bearer, False))
```
minrk, line 2022: Make sure to add a test for empty user list, which should result in clearing assignments, which is *not* the same as leaving it undefined.

```suggestion
                    has_admin_role_spec[bearer] = bearer in predef_role
```

jupyterhub/app.py
```diff
@@ -1975,25 +1975,63 @@ async def init_groups(self):
...
+                    )
+                    role_spec['users'] = set(role_spec.get('users', []))
+                    role_spec['users'] |= config_admin_users
         self.log.debug('Loading predefined roles from config file to database')
+        has_admin_role_spec = {role_bearer: False for role_bearer in admin_role_objects}
         for predef_role in self.load_roles:
-            roles.create_role(db, predef_role)
+            predef_role_obj = orm.Role.find(db, name=predef_role['name'])
+            for bearer in admin_role_objects:
+                if predef_role['name'] == 'admin':
```
minrk, line 2021: Reverse this for/if to avoid iterating for every role when only admin has something to do
```suggestion
            if predef_role['name'] == 'admin':
                for bearer in admin_role_objects:
```

jupyterhub/app.py
```diff
@@ -2009,22 +2047,35 @@ async def init_roles(self):
...
+        for bearer in admin_role_objects:
+            Class = orm.get_class(bearer)
+            for admin_obj in db.query(Class).filter_by(admin=True):
+                if has_admin_role_spec[bearer]:
+                    admin_obj.admin = admin_role in admin_obj.roles
+                else:
+                    roles.grant_role(db, admin_obj, 'admin')
+        db.commit()
+        # make sure that on hub upgrade, all roles are reset
+        if not getattr(self, '_rbac_upgrade', False):
```
minrk, line 2071: This is backwards, right?

```suggestion
        if getattr(self, '_rbac_upgrade', False):
```

If this is backwards, it also means it's untested, which should probably be fixed.

jupyterhub/app.py
```diff
@@ -1975,25 +1975,63 @@ async def init_groups(self):
...
+        role_bearers = ['users', 'services', 'groups']
+        admin_role_objects = ['users', 'services']
+        config_admin_users = set(self.admin_users) | self.authenticator.admin_users
+        db = self.db
         # load predefined roles from config file
+        if config_admin_users:
+            for role_spec in self.load_roles:
+                if role_spec['name'] == 'admin':
+                    app_log.info(
+                        "Extending admin role assignment with config admin users: %s",
```
minrk, line 2011: Let's add a warning here if both `admin_users` and admin *role* users are defined. That doesn't make a huge amount of sense, I think. Let's add an informative message for all three cases when we get here:

- only Authenticator.admin_users is defined: admins in db are preserved, admin_users granted admin permissions (e.g. "admin role does not specify users, preserving admin membership in database")
- only admin role.users is defined: admin list is exactly as specified ("admin role specifies static admin user list")
- both defined: matches admin role.users behavior after merging sets, with a warning saying that only admin role.users should be used.

jupyterhub/app.py
```diff
@@ -2009,22 +2047,35 @@ async def init_roles(self):
                                 )
                         Class = orm.get_class(bearer)
                         orm_obj = Class.find(db, bname)
```
minrk, line 2049: Maybe not for this PR, but add a todo item: if a username appears *only* in a role list, this will return None and raise an unhandled error. role-defined users should probably be added to `init_users`.

jupyterhub/app.py
```diff
@@ -2009,22 +2047,35 @@ async def init_roles(self):
...
-                        )
-        # make sure that on no admin situation, all roles are reset
-        admin_role = orm.Role.find(db, name='admin')
-        if not admin_role.users:
+                        orm_role_bearers.append(orm_obj)
+                        # Ensure all with admin role have admin flag
+                        if predef_role['name'] == 'admin':
+                            orm_obj.admin = True
+                setattr(predef_role_obj, bearer, orm_role_bearers)
+                db.commit()
```
minrk, line 2055: Let's move the commit outside the loop to avoid too many transactions:

```suggestion
        db.commit()
```

jupyterhub/tests/test_services.py
minrk: Did you commit a change? I still see creation here.

minrk (commented):
great! A few reviews inline, and some test cases to add to make sure the things I think I saw are caught by test:

1. tokens are assigned a role do not lose their roles after restart (caused currently by role deletion)
2. Setting an explicit empty admin.users (or services) should revoke assignment and not behave the same as admin.users being *left undefined*.
3. verify that the `_rbac_upgrade` initialization step is taken when it should be (and not when it shouldn't!)

jupyterhub/app.py
0mar: Actually, I just ran into that because some tests were failing. But I will add an explicit test for it, thanks

jupyterhub/tests/test_roles.py
0mar: Good catch, thanks! Fixed and added a test for it

jupyterhub/app.py
0mar: Done and took tare of the `None` case

jupyterhub/app.py
0mar: Did not consider that case but indeed, I'll only overwrite the roles that are not present in current config

jupyterhub/tests/test_services.py
0mar: Slipped through, fixed in the next commit, thanks!

jupyterhub/tests/test_roles.py
```diff
@@ -1114,3 +952,205 @@ async def test_user_group_roles(app, create_temp_role):
...
+        {
+            'name': role_name,
+            'description': 'painting with colors',
+            'scopes': ['users', 'groups'],
+            'users': user_names,
+        },
+    ]
+    hub = MockHub(load_roles=roles_to_load)
+    hub.init_db()
+    with pytest.raises(AttributeError):
```
IvanaH8, line 1037:
```suggestion
    with pytest.raises(ValueError):
```

IvanaH8: Should be raising `ValueError` from `app.py`, does it?

jupyterhub/app.py
0mar: Agreed, implemented it as such

jupyterhub/tests/test_roles.py
0mar: Indeed, thanks for catching that

minrk:
Great!

PR #3461, opened 2021-05-08
Add setting user_options in the DELETE logic of UserServerAPIHandler

sky7th:
I wanted to selectively remove containers created with spawner using jupyterhub api.

I used dockerspawner, but the remove option provided by dockerspawner is set to True or False when running jupyterhub and cannot be changed.

In the POST logic of UserServerAPIHandler, the property values received as the body are put into the user_options property of the user's spawner object and used in start() of the dockerspawner.

However, in the DELETE logic of UserServerAPIHandler, only the remove property received as the body is assigned to the variable and is used only within the controller.

I wanted to pass an additional body property to remove the container, put it in the user_options property of the user's spawner object, and use the user_options property in the condition clause of the if statement about removing the container in stop() of dockerspawner.

In the DELETE logic of UserServerAPIHandler, if you put it in the user_options property of the user's spawner object, it will be convenient to customize in the stop() part of the Spawner implementation.

### DIFF
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -531,7 +531,9 @@ class UserServerAPIHandler(APIHandler):
             # include notify, so that a server that died is noticed immediately
             status = await spawner.poll_and_notify()
             if status is None:
-                stop_future = await self.stop_single_user(user, server_name, options=options)
+                stop_future = await self.stop_single_user(
+                    user, server_name, options=options
+                )
 
         if remove:
             if stop_future:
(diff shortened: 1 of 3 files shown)


manics:
I've marked this as draft since there's been no response. @minrk Do you think this is best solved in another way, perhaps by moving the discussion to Discourse?

minrk:
It's a valid feature request for an equivalent to `user_options` applied to stop instead of start. I just want to be sure there's an actual use case not already served by `user_options` first (i.e. choices that cannot be made at start time).

minrk:
Closing with no response, feel free to reopen the discussion!

PR #3462, opened 2021-05-11
prepare to rename default branch to main

minrk:
- update references to default branch name in docs, workflows
- use HEAD in github urls, which always works regardless of default branch name
- fix petstore URLs since the old petstore links seem to have stopped working

to merge, in order:

- [x] approve this PR
- [x] rename the default branch to main in settings
- [x] merge this PR

Related tangent: I've been using [this git default-branch](https://github.com/minrk/git-stuff/blob/main/bin/git-default-branch) to help with my aliases and friends working with repos with different branch names.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -80,7 +80,7 @@ jobs:
     steps:
       - name: Should we push this image to a public registry?
         run: |
-          if [ "${{ startsWith(github.ref, 'refs/tags/') || (github.ref == 'refs/heads/master') }}" = "true" ]; then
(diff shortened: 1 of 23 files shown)


betatim:
The diff looks good to me. I think the "risk" of merging this is that there are occurrences of `master` left that aren't shown in the diff (because we missed them). I did a `grep --exclude-dir=node_modules --exclude-dir=share/jupyterhub/static -R "master" *` on this PR and didn't find any more occurrences.

There will be stuff we miss but the quickest way to find it will be to make the switch and then clean up as we go along (I think). 

Took a quick look at https://github.com/minrk/git-stuff/blob/main/bin/git-default-branch but not quite sure what it does. Is it a helper script to switch my local checkout (and my fork) to `main`?

PR #3464, opened 2021-05-11
add scopes.unparse_scopes, refine intersect_scopes

minrk:
and fix warning condition for intersection overlap

- only warn when there's a group only on one side and a user only on the other,
  otherwise there is no lost information to warn about (group and/or defined on both sides)
- correctly resolve servers as sub-scopes of user

<del>I made scopes.intersect_scopes public because I think it might be used in the single-user auth (separate PR), but that might change by the time I finish.</del> No need to make intersect public, I think. Reverted that part.

### DIFF
--- a/jupyterhub/scopes.py
+++ b/jupyterhub/scopes.py
@@ -1,5 +1,7 @@
+"""General scope definitions and utilities"""
 import functools
 import inspect
+import warnings
 from enum import Enum
 
 from tornado import web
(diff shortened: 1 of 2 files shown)


jupyterhub/scopes.py
```diff
@@ -208,6 +227,19 @@ def parse_scopes(scope_list):
     return parsed_scopes
 
 
+def unparse_scopes(parsed_scopes):
```
IvanaH8, line 230: Nice!

jupyterhub/scopes.py
```diff
@@ -13,60 +15,77 @@ class Scope(Enum):
...
+                entity: set(parsed_scopes_a[base][entity])
+                & set(parsed_scopes_b[base][entity])
                 for entity in common_entities
             }
 
-    if warn:
-        app_log.warning(
-            "[!user=, !group=] or [!user=, !server=] combinations of filters present, intersection between not considered. May result in lower than intended permissions."
-        )
+            if 'server' in all_entities and 'user' in all_entities:
```
IvanaH8, line 70: What if we have a `[read:users:servers!group=A, 'read:users:servers!server=abc/xyz]` where abc is member of group A? Technically that should be allowed, shouldn't it? Is it reasonable to raise a warning here as well for such situation?

IvanaH8:
Great, thank you for jumping on this! Just one comment, otherwise perfect.

jupyterhub/tests/test_scopes.py
```diff
@@ -760,3 +760,66 @@ async def test_resolve_token_permissions(
...
+            True,
+        ),
+        # and the reverse, just in case
+        (
+            set(["users!user=x"]),
+            set(["users!group=y"]),
+            set([]),
+            True,
+        ),
+        # resolves server under user, without warning
```
IvanaH8, line 807: Maybe adding another one where it resolves to an empty set? Probably not needed as the code does check properly for the username match.

jupyterhub/scopes.py
minrk: Yes, the group->server relationship is the same as group->user. Good catch!

jupyterhub/tests/test_scopes.py
minrk: Added, thanks!

jupyterhub/scopes.py
minrk: added tests and warning conditions for this

IvanaH8 (approved):
Great!

minrk:
Thanks for the review!

PR #3466, opened 2021-05-12
[rbac] Access scopes

minrk:
adds `access:services` and `access:users:servers` scopes, which can be used in auth

- [x] stores relationship between service/spawner and oauth client for easier lookup both directions
- [x] adds scopes to `/api/user` model
- [x] adds $JUPYTERHUB_OAUTH_SCOPES env for what scopes to check
- [x] adds utilities for checking scopes and uses them 
- [x] should users have access to services by default? **answer: no** Pro: this is mostly the current state of things. Con: some admin-only services could end up granting permission to regular users if they are not careful. I initially went with 'yes', but the prospect of accidentally granting permissions changed my mind for a more secure default.

The question right now: oauth tokens by default don't get any scopes, which means they don't have the `access` scope needed to make requests of their owner. The absence of the ability to assign scopes directly means we need a role for this, which in turn means we need a meta-scope.

This has come up a few times, so I have a question:

- do we allow attaching scopes directly to tokens, independent of roles?
- do we add another `access:owner` scope for oauth tokens that resolves to something like `access:oauth_client!client_id=...`?
- do we 'automatically' grant the relevant `access:services!service=...` scope for the owning oauth client in the `/api/user` identify handler. This could work because it means that any token owned by an oauth client necessarily has access to that client, assuming the owner still has that permission. The downside: this is getting more complicated and edge-casey, and also means that it's not possible to revoke access permission from a token to its issuing client.


minrk:
This is the main underlying mechanism for integrating scopes for access in e.g. RTC discussed in https://github.com/jupyterlab/jupyterlab/issues/10119#issuecomment-824799688 .

The missing piece will be a config API for *custom* scopes, but that's a fairly small change when all the underlying bits are there.

echarles:
Looking a bit more at the changes of this PR, I guess the tokens are just API tokens (cfr Merge OAuth and API tokens https://github.com/jupyterhub/jupyterhub/pull/3380) , but oauth_clients are still needed and are instantiated with those tokens. Is my understanding correct?

examples/service-notebook/external/jupyterhub_config.py
```diff
@@ -1,15 +1,33 @@
 # our user list
```
echarles, line 1: `examples/service-notebook/managed` content needs to be updated the same way. I have changed that content on my local env and it was working fine.

minrk:
>  I guess the tokens are just API tokens

Yes, API token and oauth token no longer mean different things. APIToken is the class / sql table. *All* tokens are now associated with an oauth client (jupyterhub itself is an oauth client now), so in that sense all tokens are "oauth tokens" but not OAuthTokens (class distinct from APIToken).

Without custom scopes, the way to grant a user access is to assign them a role with one of the scopes:

- `access:users:servers!server=user/servername` - one specific server
- `access:users:servers!user=$user` - any server owned by $user
- `access:users:servers` - any server

A class instructor example would be a good idea to have (and make sure it works!). Something like:

```python
c.JupyterHub.load_roles = [
  {"name": "my-class", "scopes"= ["access:users:servers!group=my-class-students"], "groups": ["my-class-instructors"]},
]
c.JupyterHub.load_groups = [
  {"name": "my-class-instructors", "users"= ["teacher-a"]},
  {"name": "my-class-students", "users"= ["student-1", "student-2"]},
]
```

then teacher-a would have access to servers owned by student-1 and student-2, etc. and could participate in RTC.

echarles:
Thx for those very useful inputs @minrk.

So in jhub 2.0 with RBAC, the mental model would be "every connection need a (API) token which is always associated with a oauth client". 

I think I get your initial setup for the roles/scopes that would desserve a RTC case (I will try to build an example for that), but I expect this is going to fail because of https://github.com/jupyterhub/jupyterhub/blob/1cd37a139698a67062f269d8dd7c5069132e2bc3/jupyterhub/services/auth.py#L874-L889

That block code has not been changed in the `rbac` branch, and just like before, if the authenticated UserB types in his browser UserA URL (http://localhost:8000/user/UserA), this is going to refused by the Hub auth.py as this does not take into account the roles. Or am I missing something?

minrk:
The [if self.hub_scopes here](https://github.com/jupyterhub/jupyterhub/pull/3466/files#diff-514e695392b67987ea4b144e522365c14062bb806eaa0bfe8f1d24175cabe072R928-R945) means that if scopes are defined (always for single-user servers in jupyterhub 2.0), the groups/users checks you've highlighted won't ever be reached. Only scopes are considered. Other configurations of HubAuth can always make different choices, but only considering scopes will be the default behavior. I'll add some notes to that effect by the time this is done.

echarles:
> The if self.hub_scopes here means that if scopes are defined (always for single-user servers in jupyterhub 2.0), the groups/users checks you've highlighted won't ever be reached.

Cool. So this PR provides a solution for UserA looking at UserB. I will try out that case and report any question.

> Only scopes are considered. Other configurations of HubAuth can always make different choices, but only considering scopes will be the default behavior. I'll add some notes to that effect by the time this is done.

The more notes, the better 👍

minrk:
> So in jhub 2.0 with RBAC, the mental model would be "every connection need a (API) token which is always associated with a oauth client".

Yes, exactly. And the standard way to get this token is by attempting to visit the page, which will initiate the oauth process, ultimately resulting in a token issued by the single-user server's oauth client, stored in the browser's cookies and used for all subsequent requests.

echarles:
> Yes, exactly. And the standard way to get this token is by attempting to visit the page, which will initiate the oauth process, ultimately resulting in a token issued by the single-user server's oauth client, stored in the browser's cookies and used for all subsequent requests.

So far, so good. I am curious about the client_id/client_secret of those oauth clients. I can look at the code and the database content, but notes are always welcome.

I guess the answer (which id/secret, who defines/generates them) depends on the various cases:

- a server connection the hub
- a user being authenticated
-...

minrk:
> I am curious about the client_id/client_secret of those oauth clients.

You might be interested in [this pr](https://github.com/jupyterhub/jupyterhub/pull/3444) adding some detailed oauth docs. In short, anything that uses jupyterhub for auth is an oauth client. That means every service with a web endpoint and every single-user server, which get a client id and secret issued and passed via environment variables when they start.

minrk:
> I guess the answer (which id/secret, who defines/generates them) depends on the various cases:

When it's a regular browser request, it's always the client associated with the server being visited. OAuth means the token in the cookie is always issued by the server you are visiting. Only API tokens used for relatively advanced cases (e.g. cross-site requests from another user's server, or API requests from a service or script) would be associated with another oauth client.

It shouldn't generally be relevant what oauth client any token is associated with though.

echarles:
> You might be interested in this pr 

Thx for the very useful and didactic conten. I have added comments there.

> When it's a regular browser request, it's always the client associated with the server being visited. OAuth means the token in the cookie is always issued by the server you are visiting. 

Yep, getting that model in mind now. The documentation helped the understanding.

> cross-site requests from another user's server

In that case, JupyterHub will receive a HTTP request with a cookie, will deduce the identity of the user based on the cookie, will get the roles and associated scopes, and will check if the requested resource is in allowed.

>  API requests from a service or script
 
For that case, the service/script will need to be equipped with a generated jupyterhub token like before. The difference with  RBAC is that the token would need to be associated with a role which contains the needed scopes to achieve his tasks (e.g. is the service need to list the users, the adhoc scope would need to be in the assigned role).

minrk:
@IvanaH8 @0mar I think this is ready to go if you want to have a look.

Having come up with an example where group access is important, I implemented support for`group=` filters in intersect_scopes. It does the necessary db lookups when resolving the scopes for a token.

docs/source/reference/services.md
```diff
@@ -86,10 +86,18 @@ Hub-Managed Service would include:
 This example would be configured as follows in `jupyterhub_config.py`:
 
 ```python
+c.JupyterHub.load_roles = [
+    {
+        "name": "idle-culler",
+        "scopes": [
+            "users:servers",
```
IvanaH8, line 93: `idle-culler` also needs access to users' last activity which is currently not covered by the `users:servers` scope (it does include `read:users:name` scope though). Shall we make the `read:users:activity` a subscope of `users:servers`? Or just add it to the role definition?

docs/source/reference/services.md
```diff
@@ -315,25 +282,38 @@ undefined, then any user will be allowed.
 If you don't want to use the reference implementation
 (e.g. you find the implementation a poor fit for your Flask app),
 you can implement authentication via the Hub yourself.
-We recommend looking at the [`HubAuth`][hubauth] class implementation for reference,
+JupyterHub is a standard OAuth2 provider,
+so you can any OAuth 2 implementation appropriate for
```
IvanaH8, line 286: Some words missing here?

examples/service-announcement/README.md
```diff
@@ -41,10 +56,11 @@ Anyone can read the announcement:
 The time the announcement was posted is recorded in the `timestamp` field and
 the user who posted the announcement is recorded in the `user` field.
 
-To clear the announcement text, just DELETE. Only admin users can do this.
+To clear the announcement text, send a DELETE request.
+This has the same permissionOnly admin users can do this.
```
IvanaH8, line 60:
```suggestion
This has the same permission requirement.
```

jupyterhub/apihandlers/users.py
```diff
@@ -35,20 +35,31 @@ async def get(self):
...
-        # otherwise our filters did something wrong
-        for key in ("kind", "name"):
-            if key not in model:
-                raise ValueError(f"Missing identify model for {user}: {model}")
+
+        model = get_model(user)
+
+        # add scopes to identify model,
+        # but not the scopes we added to ensure we could read our own model
+        model["scopes"] = sorted(self.raw_scopes.difference(_added_scopes))
```
IvanaH8, line 62: :+1:

jupyterhub/handlers/base.py
```diff
@@ -428,6 +429,24 @@ def _resolve_scopes(self):
...
+
+        def no_access(orm_resource, kind):
+            return False
+
+        if req_scope not in self.parsed_scopes:
+            return no_access
+
+        sub_scope = self.parsed_scopes[req_scope]
+
+        return functools.partial(scopes.check_scope_filter, sub_scope)
```
IvanaH8, line 448: :+1:

jupyterhub/orm.py
```diff
@@ -361,7 +386,7 @@ class Service(Base):
 
     # common user interface:
     name = Column(Unicode(255), unique=True)
-    admin = Column(Boolean, default=False)
+    admin = Column(Boolean(create_constraint=False), default=False)
```
IvanaH8, line 389: Just for my benefit, why is the `(create_constraint=False)` needed here? Isn't it False by default?

minrk: `create_constraint=False` is the default in sqlalchemy 1.4. This changed from 1.3, so being explicit here means we get consistent behavior on sqlalchemy before and after 1.4.

docs/source/reference/services.md
minrk: I'll add the role definition. Good catch!

docs/source/reference/services.md
minrk: Yes, fixed

jupyterhub/services/auth.py
```diff
@@ -33,13 +33,50 @@
...
 
 
+def check_scopes(required_scopes, scopes):
+    """Check that required_scope(s) are in scopes
+
+    Returns the subset of scopes matching required_scopes,
+    which is truthy if any scopes match any required scopes.
+
+    Correctly resolves scope filters *except* for groups -> user,
+    e.g. require: access:server!user=x, have: access:server!group=y
```
IvanaH8, line 52:
```suggestion
    e.g. require: access:users:servers!user=x, have: access:users:servers!group=y
```

0mar (commented):
Looks great! It's a huge PR so I can't see all the implications of the changes, but I think overall it's excellent and covers the access-scopes use case. I left one question on user models, otherwise I have no comments

docs/source/reference/services.md
```diff
@@ -342,13 +323,27 @@ and taking note of the following process:
...
 
-   ```json
+   ```python
    {
      "name": "inara",
-     "groups": ["serenity", "guild"]
+     # groups  may be omitted, depending on permissions
+     "groups": ["serenity", "guild"],
+     # scopes is new in JupyterHub 2.0
+     "scopes": [
```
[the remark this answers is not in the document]
minrk: This is only for the 'whoami' endpoint (`/api/user`), where it resolves the scopes of the current request credentials. It is not in the user model in general.

PR #3470, opened 2021-05-17
(docs) Fix DummyAuthenticator class reference

kinow:
Hi, not sure if there is a macro or some customization that renders that text with a link to the class docs perhaps?

It doesn't appear to be rendering correctly at the moment, at least for me.

image

This PR simply uses backticks instead :+1: happy to amend if there's another way to fix it.

Cheers
Bruno

### DIFF
--- a/docs/source/reference/authenticators.md
+++ b/docs/source/reference/authenticators.md
@@ -37,7 +37,7 @@ with any provider, is also available.
 ## The Dummy Authenticator
 
 When testing, it may be helpful to use the
-:class:`~jupyterhub.auth.DummyAuthenticator`. This allows for any username and
+{class}`jupyterhub.auth.DummyAuthenticator`. This allows for any username and
 password unless if a global password has been set. Once set, any username will
 still be accepted but the correct password will need to be provided.
 


docs/source/reference/authenticators.md
```diff
@@ -37,7 +37,7 @@ with any provider, is also available.
 ## The Dummy Authenticator
 
 When testing, it may be helpful to use the
-:class:`~jupyterhub.auth.DummyAuthenticator`. This allows for any username and
+`jupyterhub.auth.DummyAuthenticator`. This allows for any username and
```
minrk, line 40: I think the myst syntax for this is 

    {class}`~jupyterhub.auth.DummyAuthenticator`

minrk:
Retargeted base to the rbac branch, which has picked up the myst parser for markdown, which supports directives like this.

Hopefully will land rbac in main in the next couple of weeks

docs/source/reference/authenticators.md
kinow: Thanks @minrk, I've changed the PR to use this syntax. I wasn't aware of myst. I always assumed the jupyterhub docs were rst & markdown, but #TodayILearned about MyST. Thanks for that too!

minrk:
thanks!

PR #3475, opened 2021-05-20
handle async functions in check_db_locks

minrk:
check_db_locks checks for db lock state after the end of a function, but wasn't properly waiting when it wrapped an async function, meaning it would run the check while the async function was still outstanding, causing possible spurious failures

### DIFF
--- a/jupyterhub/orm.py
+++ b/jupyterhub/orm.py
@@ -681,7 +681,7 @@ class APIToken(Hashed, Base):
         if not token_role:
             # FIXME: remove this.
             # Creating a token before the db has roles defined should raise an error.
-            # PR #3460 should let us fix it by ensuring default roles are define
+            # PR #3460 should let us fix it by ensuring default roles are defined
 
             warnings.warn(
                 "Token created before default roles!", RuntimeWarning, stacklevel=2
--- a/jupyterhub/orm.py
+++ b/jupyterhub/orm.py
@@ -3,6 +3,7 @@
 # Distributed under the terms of the Modified BSD License.
 import enum
 import json
+import warnings
 from base64 import decodebytes
 from base64 import encodebytes
 from datetime import datetime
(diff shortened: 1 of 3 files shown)


jupyterhub/orm.py
```diff
@@ -674,18 +675,29 @@ def new(
...
             orm_token.expires_at = cls.now() + timedelta(seconds=expires_in)
+
         db.add(orm_token)
-        # load default roles if they haven't been initiated
-        # correct to have this here? otherwise some tests fail
         token_role = Role.find(db, 'token')
         if not token_role:
+            # FIXME: remove this.
+            # Creating a token before the db has roles defined should raise an error.
+            # PR #3460 should let us fix it by ensuring default roles are define
```
minrk, line 684: #3460 should allow us to remove this by ensuring we run `init_role_creation` as part of intializing a db

jupyterhub/orm.py
```diff
@@ -674,18 +675,29 @@ def new(
...
-            update_roles(db, entity=orm_token, roles=roles)
-        else:
-            assign_default_roles(db, entity=orm_token)
+        try:
+            if roles is not None:
+                update_roles(db, entity=orm_token, roles=roles)
+            else:
+                assign_default_roles(db, entity=orm_token)
+        except Exception:
+            db.delete(orm_token)
```
minrk, line 698: Fixing the db check to handle async functions revealed one of the kinds of bugs it was meant to catch! When we create tokens, we need to make sure that we don't put partially-created tokens into the database, which means deleting them explicitly if they don't finish getting created.

jupyterhub/orm.py
IvanaH8: Yes!

IvanaH8, line 684:
```suggestion
            # PR #3460 should let us fix it by ensuring default roles are defined
```

minrk:
Thanks for reviewing @IvanaH8!

PR #3476, opened 2021-05-20
[rbac] Generate scope table for docs

IvanaH8:
Ref.: #3438 

Generates scope table in markdown format for RBAC documentation from a dictionary of scopes. Eliminates the need to manually create/update the table.

- Adds scope dictionary
_Question:_ The dict is currently in a separate file in docs/source/rbac to make sure that any changes to it will be picked up when bulding the docs through makefile. Could be more appropriate in the jupyterhub source code in the future?
- Adds code to generate the table in a separate md file with appropriate indents to illustrate the scope hierarchy
- Includes the table in the documentation via myst parser admonition
- Calls the table generation code via Makefile 
_Issue_: Although changing the scope dictionary will trigger rewriting the scope table in scope-table.md with `make html`, it won't trigger rebuilding the scopes.html (as the scopes.md does not change itself). This is only achieved with running `make clean` before `make html`. For now, a message has been added to inform about this every time the scope table is regenerated.


minrk:
> Could be more appropriate in the jupyterhub source code in the future?

Yes, the file should be in the source code, probably added to `jupyterhub.scopes`. That can be imported from juptyerhub in the generate-scope-table.py. We need the same information to build the access confirmation page, so it needs to be importable at runtime.

docs/source/rbac/scopedict.py
```diff
@@ -0,0 +1,97 @@
...
+def get_scope_dict():
+    """Returns a nested dictionary of all available scopes:
+
+    {scopename: {'description': description,
+                 'subscopes': [immediate subscopes]},
+    }
+
+    without 'subscopes' key if the scope has no subscopes.
+    """
+    scope_dict = {
```
minrk, line 13: Once added to jupyterhub.scopes.py, this should probably be a module-level dict, rather than a new dict created on each call, e.g. `jupyterhub.scopes.scope_definitions = { scopename: ...}`

IvanaH8: Good idea, thanks.

IvanaH8:
> My guess is this is because the makefile isn't run on readthedocs, it needs to be triggered by conf.py, as we do [here](https://github.com/jupyterhub/jupyterhub/blob/main/docs/source/conf.py#L222).

Good guess!

minrk:
[It works!](https://jupyterhub--3476.org.readthedocs.build/en/3476/rbac/scopes.html#available-scopes). Great job

PR #3480, opened 2021-05-26
[rbac] Synchronize variable nomenclature across rbac utils

IvanaH8:
1. Updates the RBAC utils in `roles.py` and `scopes.py` with consistent scope variable names:

- `scopes`: list of scopes with abbreviations (e.g., in role definition)
- `expanded scopes`: set of expanded scopes without abbreviations (i.e., resolved metascopes, filters and subscopes)
- `parsed scopes`: dictionary JSON like format of expanded scopes
- `intersection` : set of expanded scopes as intersection of 2 expanded scope sets
- `identify scopes`: set of expanded scopes needed for identify (whoami) endpoints

2. Updates functions' docstrings accordingly 
3. Updates some function names for clarity:
- `_intersect_scopes()` to `intersect_expanded_scopes()` (defined in `scopes.py)` to highlight that `expanded scopes` need to be passed as arguments
- `_check_scope()` to `_check_scope_access()` (defined in `scopes.py`) to distinguish it from `_check_scopes()` function in `roles.py` which checks if provided scope(s) exist 
- `_resolve_scopes()` to `_resolve_roles_and_scopes()` (defined in `handlers/base.py`) to reflect the whole process from expanding roles to scopes, including intersection for token, and parsing scopes

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -81,7 +81,7 @@ class BaseHandler(RequestHandler):
         The current user (None if not logged in) may be accessed
         via the `self.current_user` property during the handling of any request.
         """
-        self.raw_scopes = set()
+        self.expanded_scopes = set()
         try:
             await self.get_current_user()
         except Exception:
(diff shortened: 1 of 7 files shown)


jupyterhub/handlers/base.py
```diff
@@ -416,7 +416,7 @@ async def get_current_user(self):
                 self.log.exception("Error getting current user")
         return self._jupyterhub_user
 
-    def _resolve_scopes(self):
+    def _resolve_roles_and_scopes(self):
         self.raw_scopes = set()
```
IvanaH8, line 420: Maybe we should rename the `raw_scopes` to `expanded_scopes` here too as this is the format the scopes are in here before being parsed.

minrk: Yeah, that's a good idea to be consistent

minrk:
This is great! I think this makes it easier to see which functions can be used when, and which transformations are needed

jupyterhub/handlers/base.py
0mar: Agree; there are also some references called `raw_scopes` left in `jupyterhub/roles.py`.

0mar (commented):
Looks good! Much clearer to get into with consistent naming indeed, thanks!

jupyterhub/handlers/base.py
IvanaH8: Thank you both, coming up.

minrk:
Conflicts resolved. Merging!

PR #3481, opened 2021-05-26
[rbac] Use scopes.scope_definitions to expand scopes

IvanaH8:
~`roles._get_scope_hierarchy()` now returns scope hierarchy dictionary created from the `scopes.scope_definitions` dictionary.~

~Eliminates the need for manual update of the _get_scope_hierarchy() used for expanding scopes and checking scope names.~

**Update**:
With the `scopes.scope_definitions` dictionary there is no need to create separate dictionary of scope hierarchy (`_get_scope_hierarchy()`). The `scopes.scope_definitions` can be used directly to obtain all subscopes for a scope. Checking scope names in `_check_scopes()` can also use this dictionary directly.

### DIFF
--- a/jupyterhub/roles.py
+++ b/jupyterhub/roles.py
@@ -74,34 +74,6 @@ def expand_self_scope(name):
     return {"{}!user={}".format(scope, name) for scope in scope_list}
 
 
-def _get_scope_hierarchy():
-    """
-    Returns:
-      scope hierarchy (dict): dictionary of available scopes and their hierarchy where
-        scopes.keys() = top level scopes and scopes that have their own subscopes
-        scopes.values() = list of immediate subscopes or None
-    """
-    subscope_lists = [
-        value['subscopes']
-        for value in scopes.scope_definitions.values()
-        if 'subscopes' in value
-    ]
-
-    scope_hierarchy = {}
-    for scope, definition in scopes.scope_definitions.items():
-
-        has_subscopes = scopes.scope_definitions[scope].get('subscopes')
-        is_subscope = any(scope in subscope_list for subscope_list in subscope_lists)
-
-        if has_subscopes:
-            scope_hierarchy[scope] = scopes.scope_definitions[scope]['subscopes']
-        else:
-            if not is_subscope:
-                scope_hierarchy[scope] = None
-
-    return scope_hierarchy
-
-
 def horizontal_filter(func):
     """Decorator to account for horizontal filtering in scope syntax"""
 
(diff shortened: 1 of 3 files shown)


jupyterhub/roles.py
```diff
@@ -75,35 +76,30 @@ def expand_self_scope(name):
...
-        'read:users:servers': ['read:users:name'],
-        'admin:groups': ['groups'],
-        'groups': ['read:groups'],
-        'read:services': None,
-        'read:hub': None,
-        'proxy': None,
-        'shutdown': None,
-    }
+    scope_hierarchy = {}
+    for scope in scopes.scope_definitions.keys():
```
minrk, line 91: To avoid repeated key access, you can use `.items()`
```suggestion
    for scope, definition in scopes.scope_definitions.items():
```

jupyterhub/roles.py
```diff
@@ -75,35 +76,30 @@ def expand_self_scope(name):
...
 
-    return scopes
+        has_subscopes = scopes.scope_definitions[scope].get('subscopes')
+        is_subscope = any(scope in subscope_list for subscope_list in subscope_lists)
+
+        if has_subscopes:
+            scope_hierarchy[scope] = scopes.scope_definitions[scope]['subscopes']
+        else:
+            if not is_subscope:
```
minrk, line 99: What's the case for `not has_scopes and is_subscope`? I'm not sure why we need to know if a scope *is* a subscope or not.

IvanaH8: The scope hierarchy dict does not have the lowest level scopes as keys (which do not have subscopes themselves but are subscopes of other scopes), e.g., `read:users:name`, so it won't include those as keys. On the other hand, it still needs to include top level scopes that do not have any subscopes, e.g., `shutdown`. 

This is the format that is used to get subscopes onwards, which thinking of now is probably confusing. We could change it to include all scopes and adjust the `_expand_scope()` function.

jupyterhub/roles.py
IvanaH8: Thanks!

jupyterhub/roles.py
minrk: `_expand_scope` already needs to handle proxy/read:hub/etc. with `: None` right? So would it work to have None there for all leaf scopes?

jupyterhub/roles.py
```diff
@@ -75,35 +76,30 @@ def expand_self_scope(name):
 
 def _get_scope_hierarchy():
```
minrk, line 77: Let's put a `@lru_cache()` on this one, since it doesn't change and is getting a bit more expensive.

jupyterhub/scopes.py
```diff
@@ -36,13 +36,13 @@
...
             'read:users:activity',
-            'read:users:roles',
         ],
+        # TODO: add read:users:admin and read:users:roles as subscopes here once implemented
     },
     'read:users:name': {'description': 'Read-only access to users’ names.'},
     'read:users:groups': {'description': 'Read-only access to users’ group names.'},
-    'read:users:roles': {'description': 'Read-only access to users’ role names.'},
     'read:users:activity': {'description': 'Read-only access to users’ last activity.'},
+    # TODO: add read:users:admin and read:users:roles once implemented
```
0mar, line 45: Is `read:users:admin` still part of the scopes we use?

IvanaH8: Ah, I think it's only used within the identify scopes and nothing else so we probably don't need to add it to the scope definitions, thanks!

jupyterhub/roles.py
IvanaH8: We do not actually need this at all! `scopes.scope_definitions` can be used directly (vis last commit).

jupyterhub/roles.py
IvanaH8: Thank you for the suggestion, which prompted the idea of just scrapping the whole hierarchy and it works!

jupyterhub/roles.py
```diff
@@ -133,29 +101,17 @@ def ignore(scopename):
...
-        index_for_sssc = len(subscopes)
-        # check for "subscopes of subscopes"
-        _expand_subscopes(index=1)
-        # check for "subscopes of sub-subscopes"
-        _expand_subscopes(index=index_for_sssc)
+    def _add_subscopes(scopename):
+        expanded_scope.append(scopename)
+        if scopes.scope_definitions[scopename].get('subscopes'):
+            for subscope in scopes.scope_definitions[scopename].get('subscopes'):
+                _add_subscopes(subscope)
```
0mar, line 111: Smart to reuse the scope definitions this way! Why do you use a nested method here?

IvanaH8: It's a recursive function that keeps calling itself until the subscopes have no subscopes of their own based on the dictionary (each scope either has a list of its immediate subscopes or it doesn't). It basically walks through every scope in a list of subscopes for the given scope.

minrk:
Great!

PR #3488, opened 2021-06-03
Support auto login when used as a OAuth2 provider

### LINKED ISSUE #3487: Allow setting `auto_login` only for OAuth2 authorization requests
### Proposed change

If you set `auto_login` to `True`, anyone landing on the hub home page is automatically redirected to the login page - the homepage isn't shown. I want to have this behavior but *only* if the endpoint being redirected to is `/hub/api/oauth2/authorize`. The use case is when JupyterHub is being used as an authentication *provider* for other services (including other JupyterHubs!). The home page of the hub provides no context to the user that it is being used as a pass through login service, and creates quite a bit of confusion.

### Alternative options

Modify the hub home page template to provide different information based on wether the user is trying to OAuth to a different application, or just logging in to JupyterHub. This can get complicated quickly though.

People using JupyterHub as a *source* of auth for other oauth2 enabled applications.

Add a `auto_login_oauth2_authorization` traitlet to `JupyterHub`?

yuvipanda:
Fixes #3487

### DIFF
--- a/jupyterhub/apihandlers/auth.py
+++ b/jupyterhub/apihandlers/auth.py
@@ -222,6 +222,14 @@ class OAuthAuthorizeHandler(OAuthHandler, BaseHandler):
         # default: require confirmation
         return True
 
+    def get_login_url(self):
+        """
+        Support automatically logging in when JupyterHub is used as auth provider
+        """
+        if self.authenticator.auto_login_oauth2_authorize:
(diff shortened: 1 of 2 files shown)


yuvipanda:
I just tested this, it works :)

yuvipanda:
@MeeSeeksDev backport to 1.4.x

PR #3492, opened 2021-06-08
Read scopes

0mar:
# Scope for reading roles

## New scopes

 - `read:users:roles`: read user roles (subscope of`admin:users`)
 - `read:services:roles`: read service roles (added to `admin` role)
 - `read:group:roles`: read group roles (subscope of `admin:groups`)
 - `read:groups:name` read group name (subscope of `read:groups`)

## Model updates

 - `read:users:name` and `read:services:name` now  also include the admin attribute in the respective models
 - the two service models have been merged
 - the (new) service model and group model now have filtering akin to user model


jupyterhub/scopes.py
```diff
@@ -48,13 +47,13 @@
             'read:users:name',
             'read:users:groups',
             'read:users:activity',
+            'read:users:roles',
```
minrk, line 50: Let's make read access to roles *not* a child of read:users. It's auth-related, so I think should be opt-in rather than part of a default role.

minrk (commented):
Great! Let's make groups also match, adding `read:groups:roles` and `read:groups:name` (read:groups gets membership, not roles)

jupyterhub/apihandlers/groups.py
```diff
@@ -34,7 +34,7 @@ def find_group(self, group_name):
 
 
 class GroupListAPIHandler(_GroupAPIHandler):
-    @needs_scope('read:groups')
+    @needs_scope('read:groups', 'read:groups:names', 'read:groups:roles')
```
IvanaH8, line 37:
```suggestion
    @needs_scope('read:groups', 'read:groups:name', 'read:groups:roles')
```

jupyterhub/apihandlers/groups.py
```diff
@@ -77,7 +77,7 @@ async def post(self):
 class GroupAPIHandler(_GroupAPIHandler):
     """View and modify groups by name"""
 
-    @needs_scope('read:groups')
+    @needs_scope('read:groups', 'read:groups:names', 'read:groups:roles')
```
IvanaH8, line 80:
```suggestion
    @needs_scope('read:groups', 'read:groups:name', 'read:groups:roles')
```

jupyterhub/roles.py
```diff
@@ -42,6 +42,7 @@ def get_default_roles():
                 'shutdown',
                 'access:services',
                 'access:users:servers',
+                'read:services:roles',
```
IvanaH8, line 45: Isn't this already included within `read:services` scope?

jupyterhub/apihandlers/groups.py
minrk: Good catch, @IvanaH8!

@0mar the fact that tests pass with this bug suggests we have some missing test coverage for the `read:group:name` scope.

0mar: Thanks! I'll fix it and add some tests.

PR #3504, opened 2021-06-11
[rbac] Authorization page for OAuth services

0mar:
Continuing from @minrk's [fork](https://github.com/minrk/jupyterhub/tree/authorization-page)

Updates the authorization page for OAuth services to show their required permissions based on their scopes.
Also updates the scope permission descriptions to be usable in both documentations and requesting authorization.

### DIFF
--- a/jupyterhub/apihandlers/auth.py
+++ b/jupyterhub/apihandlers/auth.py
@@ -216,10 +216,7 @@ class OAuthAuthorizeHandler(OAuthHandler, BaseHandler):
                 self.log.debug(
                     f"User {user.name} has already authorized {oauth_client.identifier} for roles {roles}"
                 )
-                # FIXME: uncomment after debugging
-                # it's easier to run manual tests
-                # if we always show the confirmation page
-                # return False
+                return False
             else:
                 self.log.debug(
                     f"User {user.name} has authorized {oauth_client.identifier}"
@@ -249,6 +246,31 @@ class OAuthAuthorizeHandler(OAuthHandler, BaseHandler):
             )
             credentials = self.add_credentials(credentials)
             client = self.oauth_provider.fetch_by_client_id(credentials['client_id'])
+            allowed = False
+
+            # check for access to target resource
+            if client.spawner:
+                scope_filter = self.get_scope_filter("access:users:servers")
+                allowed = scope_filter(client.spawner, kind='server')
+            elif client.service:
+                scope_filter = self.get_scope_filter("access:services")
+                allowed = scope_filter(client.service, kind='service')
+            else:
+                # client is not associated with a service or spawner.
+                # This shouldn't happen, but it might if this is a stale or forged request
+                # from a service or spawner that's since been deleted
+                self.log.error(
+                    f"OAuth client {client} has no service or spawner, cannot resolve scopes."
+                )
+                raise web.HTTPError(500, "OAuth configuration error")
+
+            if not allowed:
+                self.log.error(
+                    f"User {self.current_user} not allowed to access {client.description}"
+                )
+                raise web.HTTPError(
+                    403, f"You do not have permission to access {client.description}"
+                )
             if not self.needs_oauth_confirm(self.current_user, client, role_names):
                 self.log.debug(
                     "Skipping oauth confirmation for %s accessing %s",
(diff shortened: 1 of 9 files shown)


jupyterhub/apihandlers/auth.py
```diff
@@ -211,53 +241,66 @@ async def get(self):
...
+                ]
+            else:
+                # expanded_scopes = roles._get_subscopes(*role_objects, owner=self.current_user.orm_user)
+                # print("expanded scopes", expanded_scopes, self.current_user)
+                # parsed_scopes = scopes.parse_scopes(expanded_scopes)
+                scope_descriptions = scopes.describe_raw_scopes(
+                    raw_scopes,
+                    username=self.current_user.name,
+                )
+            print("scope definitions", raw_scopes, scope_descriptions)
```
minrk, line 295: debug

jupyterhub/scopes.py
```diff
@@ -23,92 +23,89 @@
...
-    '(no_scope)': {'description': 'Allows for only identifying the owning entity.'},
+    '(no_scope)': {'description': 'Identify the owner of this entity.'},
     'self': {
-        'description': 'Metascope, grants access to user’s own resources only; resolves to (no_scope) for services.'
+        'description': 'The user’s own resources.',
+        'metadescription': 'metascope for users, resolves to (no_scope) for services',
     },
     'all': {
-        'description': 'Metascope, valid for tokens only. Grants access to everything that the token-owning entity can access.'
+        'description': 'Everything that the token-owning entity can access.',
```
minrk, line 32: For the authorization page, I think it would really be best to have the text "you" instead of "the token-owning entity". I think the ideal user text here is probably something like "The application will be able to do anything and everything you can do." There are some tricky differences between the authorization page where you're addressing the prospective token owner and the docs, where you are talking in general terms.

What do you think?

jupyterhub/scopes.py
```diff
@@ -23,92 +23,89 @@
...
             'read:users:activity',
         ],
     },
-    'read:users:name': {'description': 'Read-only access to users’ names.'},
-    'read:users:groups': {'description': 'Read-only access to users’ group names.'},
-    'read:users:activity': {'description': 'Read-only access to users’ last activity.'},
-    # todo: describe that it only specifies timestamp of activity
-    'read:users:roles': {'description': 'Read-only access to user roles.'},
+    'read:users:name': {'description': 'Read names of users.'},
+    'read:users:groups': {'description': 'Read names of users’ groups.'},
```
minrk, line 53:
```suggestion
    'read:users:groups': {'description': 'Read users’ group membership.'},
```

jupyterhub/scopes.py
```diff
@@ -23,92 +23,89 @@
...
+    'read:users:activity': {'description': 'Read time of last user activity.'},
+    'read:users:roles': {'description': 'Read names of users’ roles.'},
```
minrk, line 55:
```suggestion
    'read:users:roles': {'description': 'Read users’ role assignments.'},
```

jupyterhub/apihandlers/auth.py
```diff
@@ -195,6 +201,30 @@ def needs_oauth_confirm(self, user, oauth_client):
...
+
+        if authorized_roles:
+            if set(roles).issubset(authorized_roles):
+                self.log.debug(
+                    f"User {user.name} has already authorized {oauth_client.identifier} for roles {roles}"
+                )
+                # FIXME: uncomment after debugging
+                # it's easier to run manual tests
+                # if we always show the confirmation page
+                # return False
```
minrk, line 222: This is a debug bit to make it easier to run manual tests. It should be put back to `return False` when we are done.

jupyterhub/apihandlers/auth.py
```diff
@@ -211,53 +241,66 @@ async def get(self):
...
-                )
-                raise web.HTTPError(500, "OAuth configuration error")
-
-            if not allowed:
-                self.log.error(
-                    f"User {self.current_user} not allowed to access {client.description}"
-                )
-                raise web.HTTPError(
-                    403, f"You do not have permission to access {client.description}"
-                )
```
minrk, line 243: The tests are failing because I lost this block in my rebase. It should be put back as it was.

minrk (commented):
Great! Definitely an improvement.

I think we'll want some BeautifulSoup tests to verify the authorization form presentation, and also some tests to exercise the prior authorization visits. Check `test_pages.py` for some examples of those.

jupyterhub/scopes.py
0mar: Agreed. I'll refactor and use a separate description for the docs, if the auth-page one does not suffice

minrk:
Great!

PR #3505, opened 2021-06-11
exclude dependabot push events from release workflow

minrk:
it's failing in the tag calculation action

maybe the action itself should do something special to avoid dependabot branches?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,6 +5,8 @@ name: Release
 # but only publish to PyPI on tags
 on:
   push:
+    branches:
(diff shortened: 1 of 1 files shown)


consideRatio:
@manics developed a feature flag for the tag calculator action called `branchRegex`, documented [here](https://github.com/jupyterhub/action-major-minor-tag-calculator#optional-input-parameters). It is my understanding that if we use that to detect specific branches we can avoid the error.

I think this change is good by itself also though as dependabot will also create a PR and otherwise trigger duplicate workflow runs.

PR #3508, opened 2021-06-11
[rbac] Fix self scope list

minrk:
adding `read:` to everything isn't right because not everything has a `read:` counterpart and not every `read:` has a write counterpart

includes a test verifying that every scope has a definition


0mar (approved):
Looks good!

PR #3509, opened 2021-06-13
release docker workflow: 'branchRegex: ^\w[\w-.]*$'

manics:
If a branch-name is not a valid docker tag use a default instead. See https://github.com/jupyterhub/jupyterhub/pull/3505

Question: revert #3505 or not?


consideRatio:
Dependabot will trigger workflows to run via PR trigger anyhow, so let's not revert that PR.

PR #3513, opened 2021-06-15
[rbac] Refactored scope names

0mar:
# Refactored scope names
Most of the RBAC scopes are constructed in such a way that `<resource>:<subresource>` requires a subset of the permissions of `<resource>`. This PR fixes this for all scopes, renaming potentially confusing scope names such as `users:servers` (that is not actually a subscope of `users`) to `servers`. The documentation has been updated to reflect this. 

In addition, the generation of the scope description in the REST API is now generated from the source code.


minrk:
We seem to have introduced a couple of flaky tests, but I don't think it's related to this PR, so going to go ahead and merge.

docs/rest-api.yml
```diff
@@ -17,38 +17,53 @@ securityDefinitions:
...
+      all: Everything that the token-owning entity can access _(metascope for tokens)_
+      admin:users:
+        Read, write, create and delete users and their authentication state,
+        not including their servers or tokens.
+      admin:auth_state: Read a user’s authentication state.
+      users:
+        Read and write permissions to user models (excluding servers, tokens
+        and authentication state).
+      read:users:
+        Read user models (excluding including servers, tokens and authentication
```
manics, line 34: `excluding` or  `including`?

docs/rest-api.yml
```diff
@@ -700,7 +745,7 @@ paths:
       summary: Get a service by name
       security:
         - oauth2:
-            - read:services
+            - read:services - read:services:name - read:roles:services
```
manics, line 748: Is this meant to be a list- if so it should be split over multiple lines.

manics (commented):
@0mar @minrk I've spotted a couple of errors in the `rest-api.yml` doc

docs/rest-api.yml
0mar: Good catch, thanks!

docs/rest-api.yml
0mar: Excluding; access to servers, tokens and auth_state info is not granted with the `read:users` scope

PR #3515, opened 2021-06-15
[rbac] Revoke tokens for OAuth services if roles expand permissions

0mar:
Users explicitly opt-in to all permissions of OAuth services. Those are stored as roles on tokens, and upon hub startup, it is possible to change the scope set of a role. To ensure that that does not lead to services having permissions the user does not grant them, this PR ensures all tokens associated with OAuth services are revoked if any of their roles obtain more permissions than originally authorized.

## Minor changes
Improve DB queries and attempt to fix a test


minrk:
Terrific!

Failure is only the hang in SSL, fixed by #3514

PR #3517, opened 2021-06-16
[rbac] Resolve small issues

0mar:
This PR resolves a number of `todo` items:

 - Test services do not get any default roles on startup
 - Raise an error when entities defined in role configurations do not exist
 - Check scope guards on hub startup
 - Ensure default roles are in db on startup (and adjust tests accordingly)
 - Test the scope guards for the AdminPageHandler are correct and function

## Open
 - Add scope filter for 'users:activity!user' in`handlers/base.py:get_current_user_token()`

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1960,7 +1960,7 @@ class JupyterHub(Application):
         user = orm.User.find(self.db, name=username)
         if user is None:
             if not self.authenticator.validate_username(username):
-                raise ValueError("Group username %r is not valid" % username)
+                raise ValueError("Username %r is not valid" % username)
             self.log.info(f"Creating user {username}")
             user = orm.User(name=username)
             self.db.add(user)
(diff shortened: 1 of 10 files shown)


jupyterhub/app.py
```diff
@@ -1950,6 +1951,22 @@ async def init_users(self):
...
+    async def _get_or_create_user(self, username):
+        """Create user if username is found in config but user does not exist"""
+        if not (await maybe_future(self.authenticator.check_allowed(username, None))):
+            raise ValueError(
+                "Username %r is not in Authenticator.allowed_users" % username
+            )
+        user = orm.User.find(self.db, name=username)
+        if user is None:
+            if not self.authenticator.validate_username(username):
+                raise ValueError("Group username %r is not valid" % username)
```
minrk, line 1963: 'Group' in message?

jupyterhub/orm.py
```diff
@@ -716,18 +716,8 @@ def new(
...
-            # PR #3460 should let us fix it by ensuring default roles are defined
-
-            warnings.warn(
-                "Token created before default roles!", RuntimeWarning, stacklevel=2
-            )
-            default_roles = get_default_roles()
-            for role in default_roles:
-                create_role(db, role)
+        if not Role.find(db, 'token'):
+            raise AttributeError("Default token role has not been created")
```
minrk, line 720: Let's call this RuntimeError since it means the state of the process is generally wrong. AttributeError is only for failure to get a specific attribute of an object.

jupyterhub/app.py
0mar: Artifact from refactoring, thanks!

jupyterhub/orm.py
0mar: :+1:

PR #3519, opened 2021-06-17
Refactor scope relationships creation

sgaist:
This version reduces the number of access to dictionaries data.

The test suite has run successfully locally.


IvanaH8 (approved):
Thanks @sgaist!

PR #3524, opened 2021-06-22
Remove pre-commit from GHA

minrk:
per https://github.com/jupyterhub/team-compass/issues/379 , trying out pre-commit.ci, which makes the GHA step redundant


minrk:
It [ran](https://results.pre-commit.ci/run/github/20785986/1624368548.8HwTn_i4T9-Ff4KTSbV01g)!

PR #3535, opened 2021-07-09
add opt-in model for paginated list results

minrk:
*(reissue #3523 because it targeted the rbac branch which has been merged)*

With a paginated API, we need to return pagination info (next page arguments, whether a next page exists, etc.),
but a simple list response doesn't give a good way to do that. Some APIs use special response headers, but that can be tedious to work with as an API consumer, because you need to keep track of the http response object, not just the response body, which is all you get from nice API wrappers.

This PR: use `Accept: application/jupyterhub-pagination+json`  to opt-in to a new response format, following what GitHub does for beta changes to their REST API.

We can follow precedents in GitHub, kubernetes, etc. and use a dict with an `items` field for the actual items, and a `_pagination` field for info about pagination, including offset, limit, url for the next request.

So what was:

```python
[ {user1}, {user2}, ... ]
```

becomes

```python
{
  "items": [{user1}, {user2}, ...],
  "_pagination": {
    "offset": 5,
    "limit": 5,
    "total": 501,
    "next": {
      "offset": 10,
      "limit": 5,
      "url": "https://hub.tld/hub/api/users?offset=10&limit=5",
    }
  }
}
```

`next` is null if there is no next page.

An alternative would be to put the new response format at a different URL instead of using the Accept header as a switch.


minrk:
Because of outside-the-db filtering by scopes, offset and limit directly translate to the db rows, but *may not* directly map to response item counts, depending on permissions. This means a user with limited permissions can infer user counts and where rows are that they don't have permission to read.

I'm thinking we should add an explicit `list:users` scope with a simple filter to identify which users can be listed. This can be implemented purely in the query without any post-processing.

jupyterhub/apihandlers/groups.py
```diff
@@ -37,11 +37,20 @@ class GroupListAPIHandler(_GroupAPIHandler):
     @needs_scope('read:groups', 'read:groups:name', 'read:roles:groups')
     def get(self):
         """List groups"""
-        query = self.db.query(orm.Group)
+        query = full_query = self.db.query(orm.Group)
         offset, limit = self.get_api_pagination()
+        full_query
```
consideRatio, line 42: Is this a typo?

jupyterhub/apihandlers/proxy.py
```diff
@@ -19,13 +19,26 @@ async def get(self):
...
+                routes[key] = all_routes[key]
+
+        if self.accepts_pagination:
+            data = self.paginated_model(routes, offset, limit, len(all_routes))
+        else:
+            data = routes
+
+        self.write(json.dumps(data))
 
         self.write(json.dumps(routes))
```
consideRatio, line 43:
```suggestion
```

consideRatio (commented):
__Regarding introducing a breaking change__

I think the approach to introduce a breaking change like this to the REST API, by using a specific Accept header, sounds fine! I appreciate that you mention that GitHub does so for beta changes to their REST API, it sounds reasonable to me as I've had experience explicitly requesting JSON responses instead of some silly XML or similar from REST APIs.

We don't have a /v1 /v2 etc of our API (right?), do you have an idea on introducing that or not and letting that be the time when this can switch on to be the response type by default?

__Data structure of response?__
Thanks for your research on how GitHub and Kubernetes have opted to format their responses, :+1: for following their practice!

__What remain?__

This PR is marked as a draft, what action points do you see remaining?

minrk:
> We don't have a /v1 /v2 etc of our API (right?), do you have an idea on introducing that or not and letting that be the time when this can switch on to be the response type by default?

That's right. I don't think the `vX` prefix approach is a good fit unless you plan to do large, sweeping changes to your API on a semi-regular basis. It would be much more complicated to define a `v2` API where only these three handlers are different, while all others are the same. I don't expect JupyterHub to make backward-incompatible API changes very often. We mostly add fields here and there in fully backward-compatible ways. 2.0 is our *first* backward-incompatible REST API change, I believe.  Most APIs I've seen (like GitHub) that use a `vX` prefix, stop updating the number once things are relatively stable (GitHub for instance, has ~always been v3, and they use v4 to refer to the totally different graphQL API, not REST at all), and opt for a different mechanism like this one to opt-in to upcoming changes.

> This PR is marked as a draft, what action points do you see remaining?

I haven't updated the tests to use the new header, so I need to do that first. Then I think it should be set, if it seems like a good approach to folks.

The other open question that I think I didn't ask is what to do for requests that don't opt-in to the new structure, where we have some choices: should they still get pagination without hints, or always request the full page size, or another option? I see these main choices:

1. **no hints** pagination is applied, but there are no hints in the response (current state, results in legacy code possibly missing out on users in list without updates)
2. **no change** ignore pagination arguments, always request all resources (1.x behavior, nothing should break, allows legacy clients to DOS large-user-count deployments with known-expensive requests)
3. **compromise** pagination is applied, but if the new `Accept` isn't present, the default page size is the *max* page size. This keeps the new upper-bound on the cost of list requests, but increases the number of users before legacy API consumers start missing out on items. It also lets deployments set the max size to infinite, in which case the legacy behavior is preserved exactly, even if it's not default.

It's not obvious to me which is the best choice, but my slight inclination is number 3. Number 2 is the one that doesn't break anything.

consideRatio:
I don't understand the difference between option 1 and 3 clearly. The options consider the situation when a user hasn't opted in to have JSON object responses I assume, and all responses will be JSON arrays still.
- In option 1, what does it mean that pagination is applied with no hints? Is it that the list response is truncated based on pagination params etc but that the response as a JSON array doesn't provide the user to get the next response properly?
- In option 3, is the difference between to option 1 that the max-size of pagination is used, rather than a default size?

minrk:
Yes, the only difference between 1 and 3 is the default page size when the request doesn't specify one (which I expect will be ~all requests, since pagination is new). So the question is: should these requests, where pagination might not be handled by the client, be "as big as allowed" or share the same default page size as the fully-supported pagination requests.

manics:
I think (3) makes sense. (2) is my least favourite, Ideally I don't think we should ever knowingly return unlimited results in one response.

minrk:
I'll go for 3, then. I added the `list:` scopes I mentioned in #3557, which I think are important for pagination to make sense. I'll rebase this on that one and implement 3 and update the tests to use the new model.

minrk:
I believe this is ready to go now, and tested with cull-idle in https://github.com/jupyterhub/jupyterhub-idle-culler/pull/30

minrk:
In some ways, pagination *increases* the load when there are lots of users - it's certainly more work to do 10 requests and 10 db queries for 50 users each than 1 request and query for 500 users. But this spreads the load into chunks, so that the hub can field other requests in the meantime.

kinow (approved):
Learned about the 2.0 beta yesterday while looking for the RBAC to see if it had landed in JupyterHub, just so we can prepare to update our jupyter server app.

Tested this PR to test the scopes with tokens, and everything worked with no issues.

Tried a few invalid values (1.0, 1o0, 0b1, -1, etc) and everything was validated correctly. The underscore numeric literals are accepted but shouldn't matter (e.g. limit=1_0 is the same as limit=10).

Also limit=0 is accepted and executes a SQL with `LIMIT 0`, but not related to this pagination PR.

After I created a token, added the scope role, set the authenticator to dummy and then created 9 users, tried:

```
curl -X GET -H "Accept: application/jupyterhub-pagination+json" -H "Authorization: token bla" "http://127.0.0.1:8081/hub/api/users?offset=0&limit=3" | python -m json.tool | jq '.items[] | {user: .name}'
{
  "user": "kinow"
}
{
  "user": "abc"
}
{
  "user": "cylc"
}
```

That brought the 3 first users. Then paginated individually and got the right users.

```
curl -X GET -H "Accept: application/jupyterhub-pagination+json" -H "Authorization: token bla" "http://127.0.0.1:8081/hub/api/users?offset=0&limit=1" | python -m json.tool | jq '.items[] | {user: .name}'
{
  "user": "kinow"
}

curl -X GET -H "Accept: application/jupyterhub-pagination+json" -H "Authorization: token bla" "http://127.0.0.1:8081/hub/api/users?offset=1&limit=1" | python -m json.tool | jq '.items[] | {user: .name}'
{
  "user": "abc"
}

curl -X GET -H "Accept: application/jupyterhub-pagination+json" -H "Authorization: token bla" "http://127.0.0.1:8081/hub/api/users?offset=2&limit=1" | python -m json.tool | jq '.items[] | {user: .name}'
{
  "user": "cylc"
}
```

Removing filters got all values correctly too. Tested the URL returned in the _pagination.next and it also pointed to the right location. Looking forward to 2.0 beta! We will probably use both RBAC and pagination (for groups if we find listing all users doesn't work for all our users) in Cylc :tada:  Thanks!

PR #3543, opened 2021-07-19
Avoid zombie processes in case of using LocalProcessSpawner

dolfinus:
Hi.

I've faced with an issue using custom spawner class based on LocalSpawner - after child process has been stopped it become zombie:
```
dolfinus 32746  0.0  0.0      0     0 ?        Z    Jul16   0:00 [ssh] <defunct>
```

This is because child is sending SIGCHILD to the parent process, but it is necessary to call `proc.wait()` method to handle it properly. Otherwise it will become a zombie which cannot be killed.

So I've wrapped the `.poll()` method call using `with` block. In Python 3+ Popen class does have `__exit__` method which is calling `.wait()` along with other useful cleanup stages, like closing `stdout` and `stderr` file descriptors.

### DIFF
--- a/jupyterhub/proxy.py
+++ b/jupyterhub/proxy.py
@@ -722,8 +722,9 @@ class ConfigurableHTTPProxy(Proxy):
         def _check_process():
             status = self.proxy_process.poll()
             if status is not None:
-                e = RuntimeError("Proxy failed to start with exit code %i" % status)
-                raise e from None
+                with self.proxy_process:
+                    e = RuntimeError("Proxy failed to start with exit code %i" % status)
+                    raise e from None
 
         for server in (public_server, api_server):
             for i in range(10):
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -1542,8 +1542,11 @@ class LocalProcessSpawner(Spawner):
         if self.proc is not None:
             status = self.proc.poll()
             if status is not None:
-                # clear state if the process is done
-                self.clear_state()
+                # handle SIGCHILD to avoid zombie processes
+                # and also close stdout/stderr file descriptors
+                with self.proc:
+                    # clear state if the process is done
+                    self.clear_state()
             return status
 
         # if we resumed from stored state,


minrk:
Interesting! Can you provide more info on your environment (OS, distro, Python version)? I haven't seen any zombie processes with lots and lots of LocalProcessSpawner launches on mac and linux, so I wonder what could be the cause.

Can you share your custom spawner, or some part of it?  Do you perhaps override [start_new_session here](https://github.com/jupyterhub/jupyterhub/blob/4ad9af583285e8cc9c8c1e455aca09ae3c21828c/jupyterhub/spawner.py#L1530)?

dolfinus:
Docker image based on CentOS 7.6 with Python 3.8.6 installed.
I've already tried to run docker container with `--init` option to use tiny which will handle zombie processes, but it handles only signals from direct child, e.g. Jupyterhub process, not the nested ones like ssh tunnel.

My custom spawner class was something like this:
```
import os
miport sys
import pexpect
import signal
import shlex
from subprocess import Popen, run, spawn


def try_setcwd(path):
    while path != "/":
        try:
            os.chdir(path)
        except OSError as e:
            exc = e  # break exception instance out of except scope
            print(f"Couldn't set CWD to {path} ({e})", file=sys.stderr)
            path, _ = os.path.split(path)
        else:
            return
    print(f"Couldn't set CWD at all ({exc}), using temp dir", file=sys.stderr)
    tempdir = mkdtemp()
    os.chdir(tempdir)

def make_preexec_fn(username, chdir=True):
    uid, gid, gids, home, *_ = get_user_info(username)

    def preexec():
        os.setgid(gid)
        os.environ["HOME"] = home
        try:
            os.setgroups(gids)
        except Exception as e:  # pylint: disable=broad-except
            print(f"Failed to set groups {e}", file=sys.stderr)

        os.setuid(uid)

        # start in the user's home dir
        if chdir:
            try_setcwd(home)

    return preexec


def spawn_as_user(cmd, user, log, timeout=10, safe_run=False):
    popen_kwargs = dict(timeout=timeout, encoding="utf-8", preexec_fn=make_preexec_fn(user))

    if not safe_run:
        log.debug(f"Running: {cmd} as {user}")
    return spawn(cmd, **popen_kwargs)

class SSHSpawner(LocalProcessSpawner):
    @property
    def ssh_socket(self):
        return shlex.quote(f"/tmp/ssh-{self.user.name}@{self.ssh_target}:{self.port}")

    @property
    def ssh_opts(self):
        return f"-o ControlMaster=auto -o ControlPath={self.ssh_socket} -o ControlPersist=180"

    @run_on_executor
    def open_tunnel(self):
        self.notify("Opening new SSH tunnel")

        # Open SSH tunnel with port forwarding
        open_tunnel_cmd = f"ssh -q {self.ssh_opts} -L 0.0.0.0:{self.port}:127.0.0.1:{self.remote_port} -N {self.ssh_target}"
        open_tunnel_child = spawn_as_user(open_tunnel_cmd, self.user.name, self.log, timeout=None)

        try:
            open_tunnel_child.expect(pexpect.TIMEOUT, timeout=1)
        except pexpect.EOF:
            self.log.error("Cannot open SSH tunnel. Failing spawner...")
            return proc.wait()

        self.proc = open_tunnel_child
        self.pid = self.proc.pid

        return None


    @run_on_executor
    def close_tunnel(self):
        self.notify("Closing SSH tunnel")

        # Open SSH tunnel with port forwarding
        close_tunnel_command = f"ssh -q {self.ssh_opts} -O exit {self.ssh_target}"
        proc = spawn_as_user(close_tunnel_command, self.user.name, self.log, timeout=None)
        proc.expect([pexpect.EOF, pexpect.TIMEOUT], timeout=1)


    @run_on_executor
    def start_notebook(self):
        self.notify("Starting remote notebook")

        opts = self.ssh_opts()
        command = os.path.join(self.remote_resource_path, self.start_notebook_script_name)
        start_notebook_cmd = f"ssh -q {self.ssh_opts} {self.ssh_target} 'nohup bash -c {shlex.quote(command)} > /dev/null &'"
        proc = spawn_as_user(start_notebook_cmd, self.user.name, self.log)
        proc.expect([pexpect.EOF, pexpect.TIMEOUT], timeout=1)


    async def start(self):
        # remove password for security reasons
        password, self._password = self._password, None
        self.progress_events = []

        host = self.get_target_host()
        if not await maybe_future(self.check_host_available(host)):
            raise OSError(f"Host {host} is unavailable")

        self.ssh_target = await maybe_future(self.ip_for_host(host))

        if not await maybe_future(self.check_ssh_connection(host, fail=False)):
            # First run, ssh key hasn't been copied yet to a remote host
            if not password:
                raise SSHCopyKeyError(host)

            await maybe_future(self.mkdir(self.user))
            await maybe_future(self.gen_key(self.user))
            await maybe_future(self.copy_key(self.user, host, password))

            await maybe_future(self.check_ssh_connection(host))

        await maybe_future(self.open_tunnel())
        await maybe_future(self.start_notebook())

        return self.ip or "127.0.0.1", self.port


    @run_on_executor
    def stop_notebook(self):
        if not self.remote_resource_path:
            return True

        self.notify("Stopping user notebook")

        command = os.path.join(self.remote_resource_path, self.stop_notebook_script_name)
        stop_command = f"ssh -q {self.ssh_opts} {self.ssh_target} {shlex.quote(command)}"
        proc = spawn_as_user(stop_command, self.user.name, self.log)
        proc.expect(pexpect.EOF)

        return proc.wait() == 0


    async def stop(self, now=False):
        await maybe_future(self.stop_notebook())
        await maybe_future(self.close_tunnel())

        if self.pid > 0:
            self.log.debug(f"Killing pid {self.pid}")
            await self._signal(signal.SIGKILL)
        else:
            self.log.warning(f"Wrong pid {self.pid}. Skip killing process.")

        # close the tunnel
        if self.ssh_socket and os.path.exists(self.ssh_socket):
            os.remove(self.ssh_socket)


    async def poll(self):
        if not self._stop_pending and not os.path.exists(self.ssh_socket):
            # tunnel is closed or non-existent
            tunnel_status = await maybe_future(self.open_tunnel())
            if tunnel_status is not None:
                return tunnel_status

        host = self.ip or "127.0.0.1"
        protocol = self.get_protocol()
        ctx = self.get_ssl_context()

        self.log.debug("Polling status...")
        url = f"{protocol}://{host}:{self.port}"
        await self.check_http_availability(url, ctx)

        if self.failed_heartbeat_checks > 5:
            self.log.error("Too many failed heartbeat checks in a row. Failing spawner...")
            return 1
```

After I found issues with zombies I've wrapped all the process-related blocks using `with` keyword to properly handle `SIGCHILD` signals:
```
...
    @run_on_executor
    def start_notebook(self):
        self.notify("Starting remote notebook")

        opts = self.ssh_opts()
        command = os.path.join(self.remote_resource_path, self.start_notebook_script_name)
        start_notebook_cmd = f"ssh -q {opts} {self.ssh_target} 'nohup bash -c {shlex.quote(command)} > /dev/null &'"
        with spawn_as_user(start_notebook_cmd, self.user.name, self.log) as proc:
            proc.expect([pexpect.EOF, pexpect.TIMEOUT], timeout=1)

...

    async def stop(self, now=False):
        # proc can be None if SSH tunnel has been reattached
        # so create fake closeable object
        with self.proc or io.StringIO(""):
            await maybe_future(self.stop_notebook())
            await maybe_future(self.close_tunnel())

        if self.pid > 0:
...
```

Now I don't see any zombie processes at all.

minrk:
Were you able to produce zombies with the base class? Looking at the code, `Popen.poll()` *does* [call waitpid](https://github.com/python/cpython/blob/v3.8.6/Lib/subprocess.py#L1741), which is the call that is followed by changes in this PR. So it seems to me that this should not change any behavior in LocalProcessSpawner. Your custom spawner is substantially different in that it does not wait for processes in the same way, so it makes sense that you might need different things to ensure `waitpid` is called.

This test also confirms that `process.poll()` reaps processes:

```python
import time
from subprocess import Popen

import psutil

popen = Popen(['sleep', '0.1'])
p = psutil.Process(popen.pid)
print(p.status())  # running
time.sleep(1)
print(p.status())  # zombie
popen.poll()  # reaps process
try:
    print(p.status())  # this would be 'zombie' if poll fails to reap
except psutil.NoSuchProcess:
    print("process reaped")  # but we end up here
```

So there may well be places we should add something like this (I don't *think* so), but after `poll()` returns non-None doesn't appear to be one of them.

dolfinus:
Ok, I've got your point.

But there is another reason to use `with` block here - it is closing stdin, stdout and stderr channels:
https://github.com/python/cpython/blob/83ca46b7784b7357d82ec47b33295e09ed7380cb/Lib/subprocess.py#L1032-L1038

Without it I'm getting warnings about unclosed files like:
```
/usr/local/lib/python3.8/concurrent/futures/thread.py:57: ResourceWarning: unclosed file <_io.FileIO name=35 mode='wb' closefd=True>
  result = self.fn(*self.args, **self.kwargs)
Object allocated at (most recent call last):
  File "/usr/local/lib/python3.8/subprocess.py", lineno 838
    self.stdin = io.open(p2cwrite, 'wb', bufsize)
/usr/local/lib/python3.8/concurrent/futures/thread.py:57: ResourceWarning: unclosed file <_io.FileIO name=36 mode='rb' closefd=True>
  result = self.fn(*self.args, **self.kwargs)
Object allocated at (most recent call last):
  File "/usr/local/lib/python3.8/subprocess.py", lineno 844
    self.stdout = io.open(c2pread, 'rb', bufsize)
```

minrk:
Good point, though this is another difference between your custom launcher and the base class. The base class does not use pipes, so there are no FDs to close. Or do you see these resource warnings when you use the base class?

dolfinus:
I don't see such errors with the default spawner class. But it is possible to pass custom [popen_kwargs](https://github.com/jupyterhub/jupyterhub/blob/71e976730768ddef0a700ef9cef17b6b0c1826a1/jupyterhub/spawner.py#L1365) with pipes redirection, and this could cause the same effect.

dolfinus:
I found a way to reproduce starting a zombie process.
Add a service to jupyterhub_config.py:
```
c.JupyterHub.services = [{
    'name': 'example-failing',
    'command': ['false']
}, {
    'name': 'example-success',
    'command': ['true']
}]
```

These processes are immediately stopping just after starting:
```
[I 2021-08-03 15:23:28.480 JupyterHub app:2679] Starting managed service example-failing
[I 2021-08-03 15:23:28.481 JupyterHub service:339] Starting service 'example-failing': ['false']
[I 2021-08-03 15:23:28.482 JupyterHub service:121] Spawning false
[D 2021-08-03 15:23:28.486 JupyterHub spawner:1147] Polling subprocess every 30s
[I 2021-08-03 15:23:28.486 JupyterHub app:2679] Starting managed service example-success
[I 2021-08-03 15:23:28.487 JupyterHub service:339] Starting service 'example-success': ['true']
[I 2021-08-03 15:23:28.488 JupyterHub service:121] Spawning true
...
[E 2021-08-03 15:23:58.488 JupyterHub service:379] Service example-failing exited with status 1
[I 2021-08-03 15:23:58.489 JupyterHub service:339] Starting service 'example-failing': ['false']
[I 2021-08-03 15:23:58.493 JupyterHub service:121] Spawning false
[D 2021-08-03 15:23:58.505 JupyterHub spawner:1147] Polling subprocess every 30s
[E 2021-08-03 15:23:58.507 JupyterHub service:379] Service example-success exited with status 0
[I 2021-08-03 15:23:58.507 JupyterHub service:339] Starting service 'example-success': ['true']
[I 2021-08-03 15:23:58.512 JupyterHub service:121] Spawning true
[D 2021-08-03 15:23:58.522 JupyterHub spawner:1147] Polling subprocess every 30s
[E 2021-08-03 15:24:28.507 JupyterHub service:379] Service example-failing exited with status 1
[I 2021-08-03 15:24:28.507 JupyterHub service:339] Starting service 'example-failing': ['false']
[I 2021-08-03 15:24:28.512 JupyterHub service:121] Spawning false
[D 2021-08-03 15:24:28.521 JupyterHub spawner:1147] Polling subprocess every 30s
[E 2021-08-03 15:24:28.525 JupyterHub service:379] Service example-success exited with status 0
[I 2021-08-03 15:24:28.525 JupyterHub service:339] Starting service 'example-success': ['true']
[I 2021-08-03 15:24:28.530 JupyterHub service:121] Spawning true
[D 2021-08-03 15:24:28.538 JupyterHub spawner:1147] Polling subprocess every 30s
```

Now let's check for zombie processes in the system:
# ps auxwwww | grep defun
root      580931  0.0  0.0      0     0 ?        Zs   15:23   0:00 [false] <defunct>
root      580932  0.0  0.0      0     0 ?        Zs   15:23   0:00 [true] <defunct>
```
They disapper after Jupyterhub is being stopped because their parent process is now `init` process which handles SIGCHILD signals properly.

dolfinus:
Hello.

Any updates here?

minrk:
Thanks for your patience! I see no harm in having it, though I don't think this is an indication of any bugs in the base class, and since it is up to downstream classes to introduce the bug, it is also up to them to clean up after themselves.

Issue #3528, opened 2021-06-25
Pure python alternative to configurable-http-proxy


AbdealiJK:
### Proposed change

I'm looking to understand if there is a pure python alternative for configurable-http-proxy ?
I have found deploying jupyterhub difficult in some of my setups because while I will always have python installed, I many a times don't have nodejs installed.
I'm just wondering if it makes sense for there to be a pure-python alternative for configurable-http-proxy which could be used/provided.

### Alternative options

If there is some way for me to just disable the http-proxy and just use Jupyterhub's routing system to manage what the proxy does as described in: https://github.com/jupyterhub/jupyterhub/blob/6be3160d7487bb362df7095f24454ef2814951fc/docs/source/reference/separate-proxy.md
That would be great too!

### Who would use this feature?

I think it would be awesome to have this for users who are just looking to get up and running quickly.
I'm not sure about any performance issues - but atleast for me, even if the performance is 70-80% - I think I would be happy with the ease of deployment it could provide

### (Optional): Suggest a solution

I guess this could be a parallel to configurable-http-proxy that could be community maintained or jupyterhub maintained (Sadly I don't have the time to do this though :( )

mapsacosta:
I've been on this same boat for a while, there's a discourse topic [*] I started about something similar.. marking myself as interested in case I can help.

[*] https://discourse.jupyter.org/t/disable-proxy-pod/9348

manics:
JupyterHub _requires_ a proxy for managing routing to the singleruser servers. By default it manages a local `configurable-http-proxy` server but that's not mandatory, you can disable it and run the proxy service yourself.

You can also use alternative proxies such as Traefik with the help of https://github.com/jupyterhub/traefik-proxy

It's definitely possible to write a pure Python implementation, but that would be done outside this repository. I'm going to close this since there's no action to be taken here, but please feel free to start a [Discourse](https://discourse.jupyter.org/) topic to see if anyone else is interested in helping, and please open a docs PR on this repo if you or someone else writes a new Python proxy!

AbdealiJK:
For anyone looking for something similar or wanting to help on this:

I spent some time this week on it, and have a working proxy created at:
https://github.com/AbdealiJK/configurable-http-proxy
It is using tornado and has atleast the basic capabilities available

AbdealiJK:
@manics The python implementation I mentioned above is now being used in our application servers, and we haven't seen any issues. It still doesn't support everything like client-ssl, etc.
[...]

PR #3546, opened 2021-07-22
doc: Mention a list of known proxies available

AbdealiJK:
Following up on the issue: https://github.com/jupyterhub/jupyterhub/issues/3528
Where we were discussing a pure python implementation of the proxy.
Since then, I have worked on a pure python alternative to configurable-http-proxy - https://pypi.org/project/configurable-http-proxy/
Thought I'd document this in the docs in case others are interested in using it

I tried to keep it similar to https://www.pyfilesystem.org/page/index-of-filesystems/

### DIFF
--- a/docs/source/reference/proxy.md
+++ b/docs/source/reference/proxy.md
@@ -220,3 +220,11 @@ previously required.
 Additionally, configurable attributes for your proxy will
 appear in jupyterhub help output and auto-generated configuration files
 via `jupyterhub --generate-config`.
+
+### Index of proxies
+
+A list of the proxies that are currently available for JupyterHub (that we know about).
+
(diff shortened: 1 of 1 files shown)


docs/source/reference/proxy.md
```diff
@@ -220,3 +220,11 @@ previously required.
 Additionally, configurable attributes for your proxy will
 appear in jupyterhub help output and auto-generated configuration files
 via `jupyterhub --generate-config`.
+
+### Index of proxies
+
+A list of the proxies that are currently available for jupyterhub (that we know about).
```
manics, line 226:
```suggestion
A list of the proxies that are currently available for JupyterHub (that we know about).
```

AbdealiJK: Done - amended my commit

PR #3548, opened 2021-07-23
Authenticator user group management

thomafred:
Added authenticator hook for synchronizing user groups

- Added hook function stub to authenticator base class
- Added new argument `authenticator_managed_group` to base `Authenticator` class
- Call authenticator hook from `refresh_auth`-function in `Base` handler class
- Added example

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -2001,6 +2001,9 @@ class JupyterHub(Application):
     async def init_groups(self):
         """Load predefined groups into the database"""
         db = self.db
+
+        if self.authenticator.manage_groups and self.load_groups:
+            raise ValueError("Group management has been offloaded to the authenticator")
         for name, usernames in self.load_groups.items():
             group = orm.Group.find(db, name)
             if group is None:
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -635,6 +635,30 @@ class Authenticator(LoggingConfigurable):
         """
         self.allowed_users.discard(user.name)
 
+    manage_groups = Bool(
+        False,
+        config=True,
+        help="""Let authenticator manage user groups
+        
+        Authenticator must implement get_user_groups for this to be useful.
+        """,
+    )
+
+    def load_user_groups(self, user, auth_state):
+        """Hook called allowing authenticator to read user groups
+
+        Updates user group memberships
+
+        Args:
+            auth_state (dict): Proprietary dict returned by authenticator
+            user(User): the User object associated with the auth-state
+
+        Returns:
+            groups (list):
+                List of user group memberships
+        """
+        return None
+
     auto_login = Bool(
         False,
         config=True,
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -582,9 +582,13 @@ class Authenticator(LoggingConfigurable):
                 or None if Authentication failed.
 
                 The Authenticator may return a dict instead, which MUST have a
-                key `name` holding the username, and MAY have two optional keys
-                set: `auth_state`, a dictionary of of auth state that will be
-                persisted; and `admin`, the admin setting value for the user.
+                key `name` holding the username, and MAY have additional keys:
+
+                - `auth_state`, a dictionary of of auth state that will be
+                  persisted;
+                - `admin`, the admin setting value for the user
+                - `groups`, the list of group names the user should be a member of,
+                  if Authenticator.manage_groups is True.
         """
 
     def pre_spawn_start(self, user, spawner):
@@ -639,25 +643,14 @@ class Authenticator(LoggingConfigurable):
         False,
         config=True,
         help="""Let authenticator manage user groups
-        
-        Authenticator must implement get_user_groups for this to be useful.
+
+        If True, Authenticator.authenticate and/or .refresh_user
+        may return a list of group names in the 'groups' field,
+        which will be assigned to the user.
+
+        All group-assignment APIs are disabled if this is True.
         """,
     )
-
-    def load_user_groups(self, user, auth_state):
-        """Hook called allowing authenticator to read user groups
-
-        Updates user group memberships
-
-        Args:
-            auth_state (dict): Proprietary dict returned by authenticator
-            user(User): the User object associated with the auth-state
-
-        Returns:
-            groups (list):
-                List of user group memberships
-        """
-        return None
 
     auto_login = Bool(
         False,
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -622,6 +622,9 @@ class BaseHandler(RequestHandler):
     def authenticate(self, data):
         return maybe_future(self.authenticator.get_authenticated_user(self, data))
 
+    def load_user_groups(self, user, auth_info):
+        return maybe_future(self.authenticator.load_user_groups(user, auth_info))
+
     def get_next_url(self, user=None, default=None):
         """Get the next_url for login redirect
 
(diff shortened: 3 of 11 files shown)


jupyterhub/user.py
```diff
@@ -223,6 +223,18 @@ def authenticator(self):
     def spawner_class(self):
         return self.settings.get('spawner_class', LocalProcessSpawner)
 
+    async def sync_groups(self, user_groups):
```
minrk, line 226:
```suggestion
   def sync_groups(self, user_groups):
```

this isn't async. Add a docstring describing what it does?

minrk (commented):
Great! I think this makes sense in principle. Let's add some tests and make sure we iron out some details.

A few things to cover:

- How is group *creation* handled? (i.e. what happens if a group is not already present in the db?)
- Presumably when this flag is set, other group assignment mechanisms should fail (via config, via the API)

Is there any issue with users who have not logged in yet (i.e. have no auth state)?

minrk:
@thomafred now that 2.0 is out, I'd like to make authenticator-managed groups my focus for 2.1.

Do you have time to address some of the review to get this ready to merge, or would you prefer us to finish it up?

thomafred:
Awesome!

I will make the changes suggested quickly

minrk:
Great! I never want to assume availability when a PR's been waiting for months.

thomafred:
@minrk - Responding to your questions:

> How is group creation handled? (i.e. what happens if a group is not already present in the db?)

The group is automatically created in the database. There is no need for the admin to manually create new groups provided by the authenticator the way things work now. However, I can see this being a bit controversial, so maybe this feature could be optional. What do you think?

> Presumably when this flag is set, other group assignment mechanisms should fail (via config, via the API)

I am not sure what would be best here. Ideally, this PR should provide the features needed to offload group management to the authenticator. In this case, having other mechanisms fail would make sense.

minrk:
> What do you think?

I think that's fine. If the authenticator is in charge, it can be its responsibility.

>  this PR should provide the features needed to offload group management to the authenticator.

I think the right thing to do here is to add a check to all endpoints that can modify group membership (includes, optionally, user creation):

```python
if self.authenticator.manage_groups:
    raise web.HTTPError(400, "Group management via API is disabled")
```

and do the same during startup if `JupyterHub.load_groups` is defined.

thomafred:
@minrk  - I have implemented the changes you suggested. However, the unit-tests `test_auth.py` appear to be failing stating `AttributeError: 'MockPAMAuthenticator' object has no attribute 'load_groups'`.

jupyterhub/app.py
```diff
@@ -1999,6 +1999,11 @@ async def _get_or_create_user(self, username):
     async def init_groups(self):
         """Load predefined groups into the database"""
         db = self.db
+
+        if self.authenticator.load_groups and self.load_groups:
```
minrk, line 2003:
```suggestion
        if self.authenticator.mange_groups and self.load_groups:
```

jupyterhub/auth.py
```diff
@@ -635,6 +635,24 @@ def delete_user(self, user):
...
         self.allowed_users.discard(user.name)
 
+    manage_groups = Bool(
+        config=True, help="Let authenticator manage user groups"
+    )
+
+    @default('manage_groups')
+    def _manage_groups_default(self):
+        return False
+
```
minrk, line 645: Don't need a dynamic default method for a static value:

```suggestion
    manage_groups = Bool(
        False,
        config=True,
        help="""Let authenticator manage user groups
        
        Authenticator must implement get_user_groups for this to be useful.
        """
    )
```

jupyterhub/auth.py
```diff
@@ -635,6 +635,24 @@ def delete_user(self, user):
...
+    def load_user_groups(self, auth_state):
```
minrk, line 646: Should probably include the user itself?

jupyterhub/auth.py
```diff
@@ -635,6 +635,24 @@ def delete_user(self, user):
...
+        """Hook called allowing authenticator to read user groups
+
+        Updates user group memberships
+
+        Args:
+            auth_state (dict): Propriatary dict returned by authenticator
```
minrk, line 652: We should specify: this should _return_ a list of group names. Maybe even call it `get_user_groups(user, auth_state)`.

jupyterhub/orm.py
```diff
@@ -222,7 +222,7 @@ class Group(Base):
     __tablename__ = 'groups'
     id = Column(Integer, primary_key=True, autoincrement=True)
     name = Column(Unicode(255), unique=True)
-    users = relationship('User', secondary='user_group_map', backref='groups')
+    users = relationship('User', secondary='user_group_map', back_populates='groups')
```
minrk, line 225: Can you explain this change? `backref` is supposed to be equivalent to two instances of `back_populates`, so I'm not sure why this is here.

thomafred: Changed to `backref`

jupyterhub/handlers/base.py
```diff
@@ -621,8 +621,8 @@ def set_login_cookie(self, user):
     def authenticate(self, data):
         return maybe_future(self.authenticator.get_authenticated_user(self, data))
 
-    def load_user_groups(self, auth_state):
-        return maybe_future(self.authenticator.load_user_groups(auth_state))
+    def load_user_groups(self, auth_state, user):
+        return maybe_future(self.authenticator.load_user_groups(user, auth_state))
```
minrk, line 625: Let's be consistent with the argument order - this call has `user, auth_state`, whereas the definition has `auth_state, user`. I think user first makes the most sense.

Let's also be consistent about variable names - `auth_state` means something specific (the persisted value from `await user.get_auth_state()`, whereas this is being called with the return value of `authenticate()`, which is a dict that may _contain_ auth state, but is not auth_state itself. We've called this dict `auth_info` elsewhere `auth_info (dict): the return value of Authenticator.authenticate() or Authenticator.refresh_user()`.

jupyterhub/handlers/base.py
```diff
@@ -779,7 +782,14 @@ async def auth_to_user(self, authenticated, user=None):
         if not self.authenticator.enable_auth_state:
             # auth_state is not enabled. Force None.
             auth_state = None
+
+        if self.authenticator.manage_groups:
+            # Run authenticator user-group reload hook
+            user_groups = await self.load_user_groups(authenticated, user)
+            user.sync_groups(user_groups)
```
minrk, line 789: As in other refresh fields, I think if the return value is `None`, it should mean no change, rather than empty the list:

```suggestion
            if user_groups is not None:
                user.sync_groups(user_groups)
```

minrk:
@thomafred thanks! I think this is just about ready to go - I rebased the PR and added some tests and docs, but wasn't allowed to update the branch. Is it possible for you to either allow maintainers to update the branch (usually a checkbox on the right), or reset your branch to [minrk:authenticator_user_group_management](https://github.com/minrk/jupyterhub/tree/authenticator_user_group_management)? Then I think this can be merged. If not, I can make a new PR from my own branch.

I mostly added docs and tests, but fixed a few things revealed by the tests. I also updated the implementation to use the `groups` field returned by authenticate, which I think better matches how other fields are already populated (auth_state, admin), and what's in #3307.

thomafred:
Was not able to find the "allow maintainers"-toggle, so I added as you a maintainer in the repo instead.

minrk (approved):

minrk:
> Was not able to find the "allow maintainers"-toggle,

Ah, I didn't realize it was an org-to-org PR. I think the checkbox is only available from personal forks, for some reason. 

I think this is all set, I'll merge as soon as I finish making 2.1.1.

minrk:
Woo! Landed. Thanks so much.

PR #3552, opened 2021-07-28
Add expiration date dropdown to Token page

dolfinus:
Hi.

It is possible to pass `expires_in` field into `POST /api/users/me/tokens` request to create token which will be later deleted from the database. But it is not possible to pass that field through Tokens page in the user interface.

I've added added a customization with Token expiration dropdown field and Expires at column in the token table:
image

By default, there are 4 default values op this dropdown field:
* 1 hour
* 1 day
* 1 month
* Never

It is possible to rewrite this block with Jinja templating, so these options could be changed, or field type could be replaced with radiobutton or string input.

### DIFF
--- a/share/jupyterhub/static/js/token.js
+++ b/share/jupyterhub/static/js/token.js
@@ -20,9 +20,14 @@ require(["jquery", "jhapi", "moment"], function ($, JHAPI, moment) {
     if (!note.length) {
       note = "Requested via token page";
     }
+    var expiration_seconds =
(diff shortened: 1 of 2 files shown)


minrk:
Wonderful!

yuvipanda:
@MeeSeeksDev backport to 1.4.x

yuvipanda:
@MeeSeeksDev backport to 1.4.x

PR #3557, opened 2021-08-02
add list:users|groups scopes

minrk:
and govern GET /users | /groups with these

Greatly simplifies filtering and pagination, because these filters can be expressed in db filters, unlike the potentially complex `read:users`.

Now the query itself will never return a model that should be excluded.

as mentioned [here](https://github.com/jupyterhub/jupyterhub/pull/3535#issuecomment-876986197), pagination + post-query filters makes things weird because counts and offsets don't line up.


minrk:
To help with parametrized tests, I added some cleanup code to our default `cleanup_after` test fixture, so all users and groups are deleted after each test. The group tests relied on this leaking state, so needed some updates to start clean for each test. This should improve reliability for running individual tests in isolation.

PR #3558, opened 2021-08-02
remove very old backward-compat for LocalProcess subclasses

minrk:
0.6 introduced start returning connection info instead of relying on db state

this avoids what is almost certainly a redundant db transaction for subclasses


yuvipanda:
Thanks, @minrk!

PR #3559, opened 2021-08-03
support inherited `--show-config` flags from base Application

minrk:
inherits flags & aliases from base classes

adds an async wrapper around the `start_show_config` entrypoint.

I think @yuvipanda was encountering issues with this feature recently. This is how it's supposed to work!

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -118,6 +118,8 @@ common_aliases = {
     'config': 'JupyterHub.config_file',
     'db': 'JupyterHub.db_url',
 }
+if isinstance(Application.aliases, dict):
+    common_aliases.update(Application.aliases)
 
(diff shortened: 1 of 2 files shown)


jupyterhub/app.py
```diff
@@ -3191,6 +3197,11 @@ def stop(self):
             self.http_server.stop()
         self.io_loop.add_callback(self.io_loop.stop)
 
+    async def start_show_config(self):
+        """Async wrapper around base start_show_config method"""
```
yuvipanda, line 3201: Can you add a comment on why this is needed?

minrk: 👍 comment added

yuvipanda:
Thanks, @minrk!

PR #3560, opened 2021-08-03
try out codeql

minrk:


minrk:
Didn't mean to open this on the real repo

consideRatio:
@minrk from my experience seeing it in BinderHub, I was dissappointed. It took a long time to execute and never signaled clear issues that i saw.

I suggest that if its tried again within the jupyterhub org, it should run as a cronjob and not directly associated with PRs as it introduces complexity for contributors and slow things down without having to my knowledge detected a clear action point in any PR to my knowledge.

By using it as a cronjob, it can be running over and over on a daily basis and alert us on failure via a badge in the repo that turn red.

minrk:
Thanks for the report! Yeah, I was mostly curious about what it wound find. I wasn't expecting to actually add it here.

minrk:
FWIW, it didn't find anything in https://github.com/minrk/jupyterhub/pull/6 . I think the tools we already run are working well enough.

PR #3562, opened 2021-08-03
remove use of deprecated distutils

minrk:
distutils is slated for deprecation in the stdlib, in favor of setuptools, etc.

we can use packaging for version parsing and setuptools in setup.py

packaging is technically an extra dependency, but rarely missing because it's so widespread


consideRatio (approved):
LGTM!

PR #3563, opened 2021-08-03
ensure admin requests for missing users 404

minrk:
we were calling `user_model(None)` which doesn't work.


consideRatio (approved):

0mar (commented):
Looks good!

PR #3564, opened 2021-08-03
don't omit server model if it's empty

minrk:
if request has access to read servers, leave it present and empty

only omit it if there's no access to read server models.

The `servers` field was meant to be omitted when a request is made that doesn't have access to this field. However, we were removing it any time it's empty, which is true for users not running any servers, even if the request has access to read the field.


0mar (commented):
Good call, makes response handling a bit more consistent

Issue #3567, opened 2021-08-09
Jupyterhub crashes when statsd metrics are defined


paccorsi:
### Bug description
Releases 4.4 and above of the configurable HTTP proxy
(https://github.com/jupyterhub/configurable-http-proxy/blob/main/CHANGELOG.md) removed support for statsd.
However if Jupyterhub is configured with a statsd client, the statsd host and port will be injected into
the proxy command, thus failing the application at startup:
https://github.com/jupyterhub/jupyterhub/blob/6be3160d7487bb362df7095f24454ef2814951fc/jupyterhub/proxy.py#L686-L696

#### Expected behaviour
Jupyterhub should not inject the statsd host and port to the configurable HTTP proxy command.

#### Actual behaviour
Since support for statsd in the proxy was removed, the command line argument are not acceptable and the proxy fails at startup with the following error:

```
[I 2021-08-09 14:26:39.541 JupyterHub app:2526] Initialized 0 spawners in 0.003 seconds
[W 2021-08-09 14:26:39.544 JupyterHub proxy:699] Running JupyterHub without SSL.  I hope there is SSL termination happening somewhere else...
[I 2021-08-09 14:26:39.544 JupyterHub proxy:703] Starting proxy @ http://0.0.0.0:8000/
[D 2021-08-09 14:26:39.545 JupyterHub proxy:704] Proxy cmd: ['configurable-http-proxy', '--ip', '0.0.0.0', '--port', '8000', '--api-ip', '127.0.0.1', '--api-port', '8001', '--error-target', 'http://localhost:10/hub/error', '--log-level', 'debug', '--statsd-host', 'localhost', '--statsd-port', '8125', '--statsd-prefix', 'jupyterhub.chp']
[D 2021-08-09 14:26:39.550 JupyterHub proxy:611] Writing proxy pid file: jupyterhub-proxy.pid
error: unknown option '--statsd-host'
[C 2021-08-09 14:26:40.505 JupyterHub app:2735] Failed to start proxy
    Traceback (most recent call last):
      File "/usr/local/lib/python3.8/dist-packages/jupyterhub/app.py", line 2733, in start
        await self.proxy.start()
      File "/usr/local/lib/python3.8/dist-packages/jupyterhub/proxy.py", line 730, in start
        _check_process()
      File "/usr/local/lib/python3.8/dist-packages/jupyterhub/proxy.py", line 726, in _check_process
        raise e from None
[...]

PR #3568, opened 2021-08-09
Stop injecting statsd parameters into the configurable HTTP proxy

paccorsi:
Addresses https://github.com/jupyterhub/jupyterhub/issues/3567
A more detailed explanation can be found in the body of the issue.

The latest release of the configurable HTTP proxy removed support for statsd. However if the statsd host and port are defined in the Jupyterhub config, they will be added to the command line arguments of the configurable HTTP proxy. Since these arguments are not recognized by the proxy, it fails to start with the following error message:
```
error: unknown option '--statsd-host'
```

This pull request removes the piece of code responsible for adding the statsd command line arguments (host, port and prefix) to the configurable HTTP proxy.
With this change users will be able to define the parameters for their statsd clients in the Jupyterhub config and the configurable HTTP proxy will run normally without failing at the start.

### DIFF
--- a/jupyterhub/proxy.py
+++ b/jupyterhub/proxy.py
@@ -683,17 +683,6 @@ class ConfigurableHTTPProxy(Proxy):
             cmd.extend(['--ssl-cert', self.ssl_cert])
         if self.app.internal_ssl:
             cmd.extend(self._get_ssl_options())
-        if self.app.statsd_host:
-            cmd.extend(
-                [
-                    '--statsd-host',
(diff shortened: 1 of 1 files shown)


consideRatio (approved):
LGTM I think. @minrk, what do you think?

minrk:
Yes! We should finish removing the statsd metrics which hav been deprecated in favor of prometheus for quite some time. 2.0 is probably a good time to finish removing them.

PR #3575, opened 2021-08-18
Validate Content-Type Header for api POST requests

VaishnaviHire:
The content-type of Hub API requests used for user management, specifically for creating a user
is not validated and so the `text/plain` type is accepted, where it must be `application/json`.
This commit adds validation for `Content-type` header for the /hub/api/users endpoint to only
allow requests with content-type as `application/json`

### DIFF
--- a/jupyterhub/apihandlers/base.py
+++ b/jupyterhub/apihandlers/base.py
@@ -61,14 +61,39 @@ class APIHandler(BaseHandler):
             return False
         return True
 
+    def check_post_content_type(self):
+        """Check request content-type, e.g. for cross-site POST requests
+
+        Cross-site POST via form will include content-type
+        """
+        content_type = self.request.headers.get("Content-Type")
+        if not content_type:
+            # not specified, e.g. from a script
+            return True
+
+        # parse content type for application/json
+        fields = content_type.lower().split(";")
+        if not any(f.lstrip().startswith("application/json") for f in fields):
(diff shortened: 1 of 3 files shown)


minrk:
Thanks for the PR! Can I ask what motivated this? This is a backward-incompatible change in that existing code that don't specify the request content-type will be rejected, which I think is not a positive change. If this were changed to reject based on content-type *if defined* and not application/json, I think that might be okay.

VaishnaviHire:
> Thanks for the PR! Can I ask what motivated this? This is a backward-incompatible change in that existing code that don't specify the request content-type will be rejected, which I think is not a positive change. If this were changed to reject based on content-type _if defined_ and not application/json, I think that might be okay.

That makes sense. I will update the PR to check if content-type is defined. The motivation for this change was that the server accepts requests without validating the `Content-type`. So a request with
content-type as `text/plain` is also accepted by the server making it vulnerable to a CSRF attack.

VaishnaviHire:
@minrk I have updated the PR as per your suggestion. Can you please take a look ?

vpavlin:
lgtm

minrk:
Thanks, that makes sense! There are other CSRF protections in place, so this should be redundant, though it's still a good idea. I moved the check to the base APIHandler so that it applies to all cookie-authenticated POST requests, not just these.

If you do find an exploitable vulnerability, please report it to security@ipython.org

PR #3583, opened 2021-08-25
Add and use pyupgrade in pre-commit

consideRatio:
- Add pyupgrade to pre-commit config
- Run pypgrade
- Run black
- Add pyupgrade --py36-plus to pre-commit config
- Run pypgrade --py36-plus
- Run black

PR #3586, opened 2021-08-26
Add pyupgrade --py36-plus to pre-commit config

consideRatio:
This is PR 3 out of 3 from splitting apart #3583.

Converted to draft because pre-commit.ci applied some auto-fixes to adding this config that is applied manually in #3584 and #3585 and should be merged first.


consideRatio:
Related issue: https://github.com/jupyterhub/team-compass/issues/453

PR #3601, opened 2021-09-11
Update/cleanup README

manics:
- circle CI no longer used
- ubuntu/debian nodejs may be too old (12.0+ required)
- remove mention of mailing list
- Python 3.6 required
- Emphasise JupyterLab over notebook


consideRatio:
I note that the mention of the mailing list was indeed removed, but we added discourse instead.

PR #3602, opened 2021-09-13
2.0 changelog

minrk:
I think we're ready to make a 2.0 beta with the RBAC stuff to get wider testing

This expands the pagination and API docs to cover some of the changes, and removes some now-unused code and docs I found along the way. I also removed any suggestions I found to use `admin: True` in a service, in favor of declaring roles.


docs/source/changelog.md
```diff
@@ -6,6 +6,52 @@ command line for details.
...
+
+The most significant change is the addition of [roles and scopes][rbac]
+to the JupyterHub permissions model,
+allowing more fine-grained access control.
+Read more about it [in the docs][rbac].
+
+In particular, the 'admin' level of permissions should not be needed anymore,
+and you can now grant users and services only the permissions they need, not more.
+We encourage you to review permissions, especially any service or user with `admin: true`
+and consider assigning only the necessary roles and scopes.
```
consideRatio, line 21: How does this relate to `admin_users`?

I recognize `admin: true` from `c.JupyterHub.services` but not from c.Authenticator.admin_users` for example.

docs/source/changelog.md
```diff
@@ -6,6 +6,52 @@ command line for details.
...
+
+[rbac]: ./rbac/index.md
+
+**Due to the change in permissions model,
+upgrading to 2.0 requires shutting down all user servers and re-issuing any tokens**.
```
consideRatio, line 26: Is it okay to do either of these options?
- shut down servers -> upgrade jh
- upgrade jh -> shut down servers

What do you think happens if a server is running during an upgrade?

minrk: I think it'll come out okay either way in the end, but shutting down first is better. The options are:

1. shut down (servers unavailable)
2. upgrade
3. start up working servers

or

1. upgrade (server authentication stops working, so *effectively* unavailable, but some caching may leave them accessible for a short time)
2. shut down
3. start up working servers

In either case, the results ought to be the same - servers are unavailable/non-functional until they are restarted. The difference is whether they are shutdown while they aren't working vs running and not working.

docs/source/changelog.md
minrk: Ah, forgot about that. I'll do one more pass on references to admin_users.

minrk: updated some notes on admin_users to point folks toward roles and mention roles as preferable for limiting permissions.

docs/source/changelog.md
consideRatio: I think this line would be relevant to convey somehow as an explanation of why servers needs to shut down. Can you try fit that into this text someplace?

> server authentication stops working, so effectively unavailable, but some caching may leave them accessible for a short time

minrk:
Publishing the beta!

PR #3604, opened 2021-09-14
Reduce logging verbosity of 'checking routes'

yuvipanda:
Of 18355 lines of logs in a 5day old hub instance,
8228 are just this message. That's 44% of the logs! We now
have prometheus metrics to monitor performance of this if
needed, and people can always turn on debug logging.


consideRatio (approved):
@minrk, agree?

PR #3610, opened 2021-09-20
Fix 1.3 level

mriedem:
The 1.3 section was a sub-section of 1.4
which makes it harder to find 1.3 release notes
in the changelog from the navigation.


manics (approved):
Thanks!

PR #3615, opened 2021-09-23
2.0: jupyterlab by default

minrk:
swaps from default nbclassic and opt-in to lab, to now default to lab and opt-in to nbclassic

defaults to jupyterlab *if* lab 3.1 is available, so should still work without configuration if lab is unavailable (or too old)

related: z2jh issue on lab by default: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/776

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,9 +38,9 @@ jobs:
         #   Tests everything when JupyterHub works against a dedicated mysql or
         #   postgresql server.
         #
-        # jupyter_server:
+        # nbclassic:
         #   Tests everything when the user instances are started with
-        #   jupyter_server instead of notebook.
+        #   notebook instead of jupyter_server.
         #
         # ssl:
         #   Tests everything using internal SSL connections instead of
@@ -48,7 +48,7 @@ jobs:
         #
         # main_dependencies:
         #   Tests everything when the we use the latest available dependencies
-        #   from: ipytraitlets.
+        #   from: traitlets.
         #
         # NOTE: Since only the value of these parameters are presented in the
         #       GitHub UI when the workflow run, we avoid using true/false as
@@ -56,6 +56,7 @@ jobs:
         include:
           - python: "3.6"
             oldest_dependencies: oldest_dependencies
+            nbclassic: nbclassic
           - python: "3.6"
             subdomain: subdomain
           - python: "3.7"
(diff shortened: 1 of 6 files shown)


consideRatio (approved):
The code changes LGTM.

I'm uncertain about if this would be something that calls for discussion as I perceive it was a big discussion regarding this in z2jh stretching years back. I'm personally in favor of transitions towards jupyter_server and a JupyterLab based UI.

minrk:
I feel like the discussion's already happened in https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/776 and  https://discourse.jupyter.org/t/poll-should-zero-to-jupyterhub-switch-to-jupyterlab-as-the-default-user-interface/8001/1 , with support well established. Should we raise the same questions again, given the experience on BinderHub? I do think the situation's easier on jupyterhub, which requires opt-in from users to a major version upgrade, and opting in to classic UI is also easy and documented (will need docs both here (✅ done) and in z2jh).

manics:
I don't think we need another discussion, it's clear that notebook is in maintenance mode and all focus is on JupyterLab. Unlike mybinder.org this isn't something that will catch people unawares, it's a deliberate upgrade for admins so we can make sure the consequences are clear.

However I think it'd be good not to rush into a final release of JupyterHub and Z2JH, instead release one or more betas of both, make some announcements, then leave a few weeks (is a month or two too long?) to allow time for unexpected issues to be reported.

meeseeksmachine:

https://discourse.jupyter.org/t/jupyterhub-singleuser-and-jupyter-server/10942/8

minrk:
@manics absolutely! With the RBAC stuff, I want a pretty long prerelease window to try to get real useful feedback. Getting the prereleases out there takes time and advertisement.

My possibly optimistic goal is to have these ready for academic spring term deployments, so maybe December.

minrk:
Maybe it's worth putting a (rough) calendar together. I'd like to put out prereleases pretty often (~weekly?) since that's what can get tested in z2jh. Maybe aim for an RC by the end of October and a release by the end of November, assuming things go smoothly. How does that sound?

manics:
Sounds good to me! We've got Hub team meetings on 21st Oct and 18th Nov, so we could use those to sign-off on an RC at the Oct meeting, and final release at the Nov meeting?

minrk:
@manics that sound like a good plan. I've added the 2.0 release candidate to the October meeting agenda.

PR #3616, opened 2021-09-23
add delete scopes for users, groups, servers

### LINKED ISSUE #3608: delete:users|servers|groups scope ?
Currently, add/delete permissions are represented as a single scope - `admin:users`, `admin:groups`, `servers`, `tokens`, etc. I.e. you have to have permission to *create* something in order to delete it.

The practical consequence: the cull-idle service must be granted permission to *start* servers in order to have permission to stop them, though it will only ever use the stop function. Same goes for culling idle users - it must have permission to *create* users if it wants permission to delete them.

More scopes is more complex, so we have to think of where the boundary is for when a scope makes sense to add. But this should probably be consistent across all resources, whether we add it or not.

### Proposed change

We could add a `delete:` subscope on users, groups, tokens, and servers to allow services only permission to delete resources, without needing permission to start servers, issue tokens, create users, etc.

### Alternative options

Keep it as-is, and bundle creation and deletion in one scope. It's always possible to add more finer-grained scopes, it's harder to remove them.

deployments of the cull-idle service (most deployments!) who want limited permissions.

minrk:
so e.g. cull-idle services do not need permission to start servers in order to be able to stop them, it seems we have a realistic need to distinguish scopes at this level

closes #3608

### DIFF
--- a/jupyterhub/scopes.py
+++ b/jupyterhub/scopes.py
@@ -36,12 +36,15 @@ scope_definitions = {
     },
     'admin:users': {
         'description': 'Read, write, create and delete users and their authentication state, not including their servers or tokens.',
-        'subscopes': ['admin:auth_state', 'users', 'read:roles:users'],
+        'subscopes': ['admin:auth_state', 'users', 'read:roles:users', 'delete:users'],
     },
     'admin:auth_state': {'description': 'Read a user’s authentication state.'},
     'users': {
         'description': 'Read and write permissions to user models (excluding servers, tokens and authentication state).',
         'subscopes': ['read:users', 'list:users', 'users:activity'],
+    },
+    'delete:users': {
+        'description': "Delete users.",
     },
     'list:users': {
         'description': 'List users, including at least their names.',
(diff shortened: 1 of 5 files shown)


jupyterhub/scopes.py
```diff
@@ -76,20 +79,21 @@
...
     'servers': {
         'description': 'Start and stop user servers.',
-        'subscopes': ['read:servers'],
+        'subscopes': ['read:servers', 'delete:servers'],
     },
     'read:servers': {
         'description': 'Read users’ names and their server models (excluding the server state).',
         'subscopes': ['read:users:name'],
     },
+    'delete:servers': {'description': "Stop users' servers."},
```
consideRatio, line 88: How does this relate to named servers? I believe a named server can be both stop/started, but also deleted.

minrk: They are not treated separately at the moment. While technically it's a separate operation, it's very rarely (if ever) treated separately, so I'd like to wait for someone to request this feature. I've updated the description to include "and delete"

consideRatio: > They are not treated separately at the moment. While technically it's a separate operation, it's very rarely (if ever) treated separately, so I'd like to wait for someone to request this feature.

:+1: 

> I've updated the description to include "and delete"

Great! Note that this change is not yet pushed though - right?

minrk: Good catch! Switching between too many branches today. Pushed now.

consideRatio (approved):
Nice, this LGTM!

Issue #3609, opened 2021-09-19
Should we bundle nullauthenticator in this repository?


manics:
### Proposed change
`nullauthenticator` is a very small JupyterHub authenticator used by BinderHub:
https://github.com/jupyterhub/nullauthenticator/blob/master/nullauthenticator.py

Is it worth bundling it in this repo? For example [`DummyAuthenticator` is already included](https://github.com/jupyterhub/jupyterhub/blob/4f6ef54b50ea577bb67f2726bc98a5ce909ebc92/jupyterhub/auth.py#L1150).

This is the only BinderHub dependency (aside from BinderHub) that's not on conda-forge.

### Alternative options
Keep https://github.com/jupyterhub/nullauthenticator in its own repo. Consider creating a conda-forge package for it.

### Who would use this feature?
Anyone wanting to install all BinderHub's dependencies from conda-forge.

minrk:
Seems reasonable, given that we've added other 'api-only' features. Disabling JupyterHub's own auth is part of such deployments.

PR #3619, opened 2021-09-23
Add NullAuthenticator to jupyterhub

manics:
Closes https://github.com/jupyterhub/jupyterhub/issues/3609

When JupyterHub is released we should archive https://github.com/jupyterhub/nullauthenticator


jupyterhub/nullauthenticator.py
```diff
@@ -0,0 +1,30 @@
...
+    """
+
+    # auto_login skips 'Login with...' page on Hub 0.8
+    auto_login = True
+
+    # for Hub 0.7, show 'login with...'
+    login_service = 'null'
+
+    def get_handlers(self, app):
+        return [('/nologin', NullLoginHandler)]
```
manics, line 30: What's the purpose of this `/nologin` handler? The original https://github.com/jupyterhub/nullauthenticator/blob/dedb18412a06da9e309c1b92c05dd305ec817d03/nullauthenticator.py#L18 uses the `web` module which isn't included in the imports, so `NullLoginHandler.get()` will return a 500 server error instead of the intended 403.

I could imagine it being used as the `login_url()` https://github.com/jupyterhub/jupyterhub/blob/b029d983f99878b365c28a6e8b606cfeb0e8d3d7/jupyterhub/auth.py#L673-L688
but that would require overriding it.

minrk: I think this is a leftover from Authenticators maybe *needing* to register a handler to be redirected in earlier versions of JupyterHUb.

I think `return []` is appropriate here. The Handler itself is not needed anymore.

*alternately*, this could be `/login` and forcefully override the default login page.

The result should ultimately be the same, though, so the simplest option of an empty list and no custom handler is probably correct.

manics:
Thanks for the explanation. Since `Nullauthenticator` is reduced to a single small class I've moved it into auth.py instead of a separate file, and I've added a test.

Issue #3624, opened 2021-09-26
_xsrf cookie for singleuser server not scoped to the base_url


albertmichaelj:
EDIT: The issue with the `jupyterhub-session-id` cookie has been fixed with pull request #3625, which has been merged in master. The issue with the `_xsrf` cookie remains, though there are configuration changes that can be made to fix it.

### Bug description

JupyterHub has the ability to be accessed at a subpath of a domain (such as `www.whatever.com/jupyterhub`) using the configurable `c.JupyterHub.base_url="/jupyterhub"`. When this is done, most of the cookies that JupyterHub uses are effectively scoped to the path "/jupyterhub/" (or a further sub-path). However, the cookies `jupyterhub-session-id` and `_xsrf` are scoped to the path `/`. This is maybe a problem, but I actually haven't figured out if it actually causes bugs. The potential issue is that if you run two instances of JupyterHub at two different subpaths of the same domain (i.e. `www.whatever.com/jupyterhub1` and `www.whatever.com/jupyterhub2`) the `jupyterhub-session-id` and `_xsrf` cookies are effectively shared between the two instances if you are logged into both instances at the same time. Whichever server wrote the cookie first gets to dictate the cookie value. Moreover, if you log out of one of the servers, it will delete the `jupyterhub-session-id` and `_xsrf` cookies even if you are still logged into the other server.

I actually do run multiple instances of Jupyterhub on sub-paths of a single domain, and that is how I noticed the cookie issue. However, I have been unable to determine if anything breaks because of this. As far as I can tell, even though `jupyterhub-session-id` and `_xsrf` occasionally gets mistakenly deleted due to above issue, the other server keeps working fine. However, this is likely unintended behavior, so I wanted to bring it up.

#### Expected behaviour

I expect all cookies to be scoped to the subpath indicated by `base_url`.

#### Actual behaviour

This doesn't happen for the cookies `jupyterhub-session-id` and `_xsrf`

### How to reproduce

Run a jupyterhub server with the following command:

```
jupyterhub --JupyterHub.base_url="/jupyterhub"
```

Log in at `http://localhost:8000/jupyterhub`. Then look at the cookies and the paths for the cookies (browser dependent as to how to do this).

### Proposed Solution
I have already forked JupyterHub and implemented a two line fix for `jupyterhub-session-id`. I will create a separate pull request and link it back to this issue after submitting. Fixing `jupyterhub-session-id` seems very straightforward and, I think, uncontroversial.

[...]

PR #3625, opened 2021-09-26
Added base_url to path for jupyterhub-session-id cookie

albertmichaelj:
This pull request fixes the `jupyterhub-session-id` cookie path issue as described in issue #3624. It does not fix the problem with the `_xsrf` cookie, so it does not completely close that issue. I'm happy for any feedback as to more elegant fixes. This seemed to be the fix with the minimum delta.


minrk:
Thanks!

PR #3636, opened 2021-10-04
Fail suspected API requests with 424, not 503

yuvipanda:
Non-running user servers making requests is a fairly
common occurance - user servers get culled while their
browser tabs are left open. So we now have a background level
of 503s responses on the hub *all* the time, making it
very difficult to detect *real* 503s, which should ideally
be closely monitored and alerted on.

I *think* 404 is a more appropriate response, as the resource
(API) being requested is no longer present.

Here is a snapshot of 503 responses on a bunch of hubs:

image

This is theoretically hubs under 'normal' usage, so we can't actually use 503
error rates for any monitoring or alerting.

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -1330,7 +1330,7 @@ class UserUrlHandler(BaseHandler):
 
     **Changed Behavior as of 1.0** This handler no longer triggers a spawn. Instead, it checks if:
 
-    1. server is not active, serve page prompting for spawn (status: 503)
+    1. server is not active, serve page prompting for spawn (status: 424)
     2. server is ready (This shouldn't happen! Proxy isn't updated yet. Wait a bit and redirect.)
     3. server is active, redirect to /hub/spawn-pending to monitor launch progress
        (will redirect back when finished)
@@ -1349,7 +1349,11 @@ class UserUrlHandler(BaseHandler):
         self.log.warning(
             "Failing suspected API request to not-running server: %s", self.request.path
         )
-        self.set_status(404)
+
+        # If we got here, the server is not running. To differentiate
+        # that the *server* itself is not running, rather than just the particular
+        # resource *in* the server is not found, we return a 424 instead of a 404.
+        self.set_status(424)
         self.set_header("Content-Type", "application/json")
 
         spawn_url = urlparse(self.request.full_url())._replace(query="")
@@ -1514,15 +1518,14 @@ class UserUrlHandler(BaseHandler):
             self.redirect(pending_url, status=303)
             return
 
-        # if we got here, the server is not running
-        # serve a page prompting for spawn and 503 error
-        # visiting /user/:name no longer triggers implicit spawn
-        # without explicit user action
+        # If we got here, the server is not running. To differentiate
+        # that the *server* itself is not running, rather than just the particular
+        # page *in* the server is not found, we return a 424 instead of a 404.
         spawn_url = url_concat(
             url_path_join(self.hub.base_url, "spawn", user.escaped_name, server_name),
             {"next": self.request.uri},
         )
-        self.set_status(503)
+        self.set_status(424)
 
         auth_state = await user.get_auth_state()
         html = await self.render_template(
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -1353,7 +1353,10 @@ class UserUrlHandler(BaseHandler):
         # If we got here, the server is not running. To differentiate
         # that the *server* itself is not running, rather than just the particular
         # resource *in* the server is not found, we return a 424 instead of a 404.
-        self.set_status(424)
+        # We allow retaining the old behavior to support older JupyterLab versions
+        self.set_status(
+            424 if not self.app.use_legacy_stopped_server_status_code else 503
+        )
         self.set_header("Content-Type", "application/json")
 
         spawn_url = urlparse(self.request.full_url())._replace(query="")
@@ -1521,11 +1524,14 @@ class UserUrlHandler(BaseHandler):
         # If we got here, the server is not running. To differentiate
         # that the *server* itself is not running, rather than just the particular
         # page *in* the server is not found, we return a 424 instead of a 404.
+        # We allow retaining the old behavior to support older JupyterLab versions
         spawn_url = url_concat(
             url_path_join(self.hub.base_url, "spawn", user.escaped_name, server_name),
             {"next": self.request.uri},
         )
-        self.set_status(424)
+        self.set_status(
+            424 if not self.app.use_legacy_stopped_server_status_code else 503
+        )
 
         auth_state = await user.get_auth_state()
         html = await self.render_template(
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -1349,7 +1349,7 @@ class UserUrlHandler(BaseHandler):
         self.log.warning(
             "Failing suspected API request to not-running server: %s", self.request.path
         )
-        self.set_status(503)
+        self.set_status(404)
         self.set_header("Content-Type", "application/json")
 
         spawn_url = urlparse(self.request.full_url())._replace(query="")
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -369,7 +369,7 @@ class JupyterHub(Application):
         even if your Hub authentication is still valid.
         If your Hub authentication is valid,
         logging in may be a transparent redirect as you refresh the page.
-        
+
         This does not affect JupyterHub API tokens in general,
         which do not expire by default.
         Only tokens issued during the oauth flow
@@ -862,7 +862,7 @@ class JupyterHub(Application):
         "/",
         help="""
         The routing prefix for the Hub itself.
-        
+
         Override to send only a subset of traffic to the Hub.
         Default is to use the Hub as the default route for all requests.
 
@@ -874,7 +874,7 @@ class JupyterHub(Application):
         may want to handle these events themselves,
         in which case they can register their own default target with the proxy
         and set e.g. `hub_routespec = /hub/` to serve only the hub's own pages, or even `/hub/api/` for api-only operation.
-        
+
         Note: hub_routespec must include the base_url, if any.
 
         .. versionadded:: 1.4
@@ -1457,7 +1457,7 @@ class JupyterHub(Application):
         Can be a Unicode string (e.g. '/hub/home') or a callable based on the handler object:
 
         ::
-        
+
             def default_url_fn(handler):
                 user = handler.current_user
                 if user and user.admin:
@@ -1484,6 +1484,25 @@ class JupyterHub(Application):
         current behavior.
         """,
     ).tag(config=True)
+
+    use_legacy_stopped_server_status_code = Bool(
+        False,
+        help="""
+        Return 503 rather than 424 when request comes in for a non-running server.
+
+        Prior to JupyterHub 2.0, we returned a 503 when any request came in for
+        a user server that was currently not running. By default, JupyterHub 2.0
+        will return a 424 - this makes operational metric dashboards more useful.
+
+        JupyterLab < 3.2 expected the 503 to know if the user server is no longer
+        running, and prompted the user to start their server. Set this config to
+        true to retain the old behavior, so JupyterLab < 3.2 can continue to show
+        the appropriate UI when the user server is stopped.
+
+        This option will be removed in a future release.
+        """,
+        config=True,
+    )
 
     def init_handlers(self):
         h = []
--- a/jupyterhub/tests/test_pages.py
+++ b/jupyterhub/tests/test_pages.py
@@ -1001,11 +1001,18 @@ async def test_token_page(app):
 async def test_server_not_running_api_request(app):
     cookies = await app.login_user("bees")
     r = await get_page("user/bees/api/status", app, hub=False, cookies=cookies)
-    assert r.status_code == 404
+    assert r.status_code == 424
     assert r.headers["content-type"] == "application/json"
     message = r.json()['message']
     assert ujoin(app.base_url, "hub/spawn/bees") in message
     assert " /user/bees" in message
+
+
+async def test_server_not_running_api_request_legacy_status(app):
+    app.use_legacy_stopped_server_status_code = True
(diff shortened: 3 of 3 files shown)


yuvipanda:
https://github.com/berkeley-dsep-infra/datahub/issues/2693#issuecomment-926950119 is the motivation.

yuvipanda:
I think we might need to modify the parts of jupyterlab / client that detect non-running servers as well?

yuvipanda:
We'll probably need to modify the code in JupyterLab too here: https://github.com/jupyterlab/jupyterlab/blob/b38e0847c33297e128e6993b7878d8c646a00040/packages/services/src/kernel/manager.ts#L242.

I think we should return a 404 with a special JSON payload.

yuvipanda:
@minrk I'd *love* to find a way to get this in to 2.0, as it's a bit of a breaking change (as it requires co-ordination with jupyterlab too)

yuvipanda:
Do we even need a JSON response? is the 404 alone not enough to say 'the server is gone!'?

yuvipanda:
I opened https://github.com/jupyterlab/jupyterlab/pull/11205 for the JupyterLab PR

minrk:
We used to return 404 on this, but [switched to 503][] in 1.0, in part because 503 is an error code for trying to talk to a server that's not available at the moment, which is what's happening here. I think the ambiguity of 404 for multiple unrelated reasons was confusing. 503 seems more appropriate than 404 for this situation, to me.

[switched to 503]: https://github.com/jupyterhub/jupyterhub/pull/2435

> is the 404 alone not enough to say 'the server is gone!'?

Not in general, at least. 404 on single-user endpoints would normally mean that the resource being requested is gone, e.g. missing resources in the single-user server's REST API (no such directory, kernel, or terminal extension not available, etc.). On *some* resources, e.g. roots like `/api/status` that have been accessed successfully before (first-request 404s are more likely base-url or jupyter-server plugin configuration issues), a 404 after the first successful request is enough to be *probably* a correct assessment, but checking the response content would make more sense to me.  But in general, I don't think this situation is accurately represented as the requested resource being gone, it's the *service* providing the requested resource being unavailable.

I think it's appropriate that the error code for trying to talk to a server that's not there is not the same as the error code you would expect from normal operation of that server. But I also understand how a using server error code could cause issues for coarse infrastructure tools assuming `5XX` is your problem to solve, even if I still think it's the most accurate description of the situation.

In skimming [httpstatus.es][], I don't see an obvious non-404 response indicating something else is wrong. Maybe [424 failed dependency][]? WDYT? Any change to this error code is a breaking change for JupyterLab, as reflected in your PR, so I'm not sure what the best course of action is here.

[httpstatus.es]: https://httpstatuses.com
[424 failed dependency]: https://httpstatuses.com/424

minrk:
[421 misdirected][] *sounds* like a good choice from the name and description, but it's part of HTTP/2 connection re-use, and I think intermediate layers could retry those requests on a new connection without propagating to the client.

[421 misdirected]: https://httpstatuses.com/421

[412 precondition failed][] also *sounds* right (precondition: a server is running!), but it's specific to preconditions in headers liked `If-Modified-Since`, not out-of-band conditions like ours. 

[412 precondition failed]: https://httpstatuses.com/412

I wish there was a special error code for "like 503, but the client has agency in resolving the problem" which is implicitly not the case for 5XX errors.

We could do a generic 400, but that doesn't seem better than re-using 404 to me. It really seems useful to me to distinguish the "tried to talk to a not-running server" event from other failing requests, and since monitoring tools only seem to preserve the status code (not even the upstream status code, apparently, only the outgoing one without the cause), status code seems to be the only place for it.

jupyterhub/handlers/base.py
```diff
@@ -1349,7 +1349,7 @@ def _fail_api_request(self, user_name='', server_name=''):
         self.log.warning(
             "Failing suspected API request to not-running server: %s", self.request.path
         )
-        self.set_status(503)
+        self.set_status(404)
```
minrk, line 1352: Probably should match the same status code below for the non-api page on L1525

minrk:
I tested this PR, and the failure mode for versions of jupyterlab that don't support it is pretty rough. You get this incorrect and worrying error message (rightly so, from JupyterLab's perspective, since the Hub is telling Lab the directory requested does not exist):

Screen Shot 2021-10-05 at 10 04 51

and a flood of retried requests:

```
[W 2021-10-05 10:04:16.853 JupyterHub base:1349] Failing suspected API request to not-running server: /hub/user/test-1/api/contents
[W 2021-10-05 10:04:16.854 JupyterHub log:189] 404 GET /hub/user/test-1/api/contents?content=1&1633421056831 (test-1@::1) 3.43ms
[I 2021-10-05 10:04:16.866 JupyterHub log:189] 302 GET /user/test-1/api/contents?content=1&1633421056854 -> /hub/user/test-1/api/contents?content=1&1633421056854 (@::1) 4.68ms
[W 2021-10-05 10:04:16.880 JupyterHub base:1349] Failing suspected API request to not-running server: /hub/user/test-1/api/contents
[W 2021-10-05 10:04:16.882 JupyterHub log:189] 404 GET /hub/user/test-1/api/contents?content=1&1633421056854 (test-1@::1) 5.77ms
[I 2021-10-05 10:04:16.892 JupyterHub log:189] 302 GET /user/test-1/api/contents?content=1&1633421056883 -> /hub/user/test-1/api/contents?content=1&1633421056883 (@::1) 2.30ms
[W 2021-10-05 10:04:16.909 JupyterHub base:1349] Failing suspected API request to not-running server: /hub/user/test-1/api/contents
[W 2021-10-05 10:04:16.911 JupyterHub log:189] 404 GET /hub/user/test-1/api/contents?content=1&1633421056883 (test-1@::1) 5.54ms
[I 2021-10-05 10:04:16.926 JupyterHub log:189] 302 GET /user/test-1/api/contents?content=1&1633421056914 -> /hub/user/test-1/api/contents?content=1&1633421056914 (@::1) 3.36ms
[W 2021-10-05 10:04:16.937 JupyterHub base:1349] Failing suspected API request to not-running server: /hub/user/test-1/api/contents
[W 2021-10-05 10:04:16.937 JupyterHub log:189] 404 GET /hub/user/test-1/api/contents?content=1&1633421056914 (test-1@::1) 3.10ms
[I 2021-10-05 10:04:16.952 JupyterHub log:189] 302 GET /user/test-1/api/contents?content=1&1633421056940 -> /hub/user/test-1/api/contents?content=1&1633421056940 (@::1) 5.22ms
[W 2021-10-05 10:04:16.970 JupyterHub base:1349] Failing suspected API request to not-running server: /hub/user/test-1/api/contents
[W 2021-10-05 10:04:16.971 JupyterHub log:189] 404 GET /hub/user/test-1/api/contents?content=1&1633421056940 (test-1@::1) 4.82ms
[I 2021-10-05 10:04:16.990 JupyterHub log:189] 302 GET /user/test-1/api/contents?content=1&1633421056973 -> /hub/user/test-1/api/contents?content=1&1633421056973 (@::1) 3.73ms
[W 2021-10-05 10:04:17.003 JupyterHub base:1349] Failing suspected API request to not-running server: /hub/user/test-1/api/contents
[W 2021-10-05 10:04:17.004 JupyterHub log:189] 404 GET /hub/user/test-1/api/contents?content=1&1633421056973 (test-1@::1) 3.47ms
[I 2021-10-05 10:04:17.018 JupyterHub log:189] 302 GET /user/test-1/api/contents?content=1&1633421057008 -> /hub/user/test-1/api/contents?content=1&1633421057008 (@::1) 4.26ms
[W 2021-10-05 10:04:17.036 JupyterHub base:1349] Failing suspected API request to not-running server: /hub/user/test-1/api/contents
[W 2021-10-05 10:04:17.037 JupyterHub log:189] 404 GET /hub/user/test-1/api/contents?content=1&1633421057008 (test-1@::1) 5.10ms
[I 2021-10-05 10:04:17.054 JupyterHub log:189] 302 GET /user/test-1/api/contents?content=1&1633421057039 -> /hub/user/test-1/api/contents?content=1&1633421057039 (@::1) 3.67ms
[W 2021-10-05 10:04:17.073 JupyterHub base:1349] Failing suspected API request to not-running server: /hub/user/test-1/api/contents
[W 2021-10-05 10:04:17.074 JupyterHub log:189] 404 GET /hub/user/test-1/api/contents?content=1&1633421057039 (test-1@::1) 3.48ms
[I 2021-10-05 10:04:17.087 JupyterHub log:189] 302 GET /user/test-1/api/contents?content=1&1633421057075 -> /hub/user/test-1/api/contents?content=1&1633421057075 (@::1) 3.49ms
```


So deploying this change for anyone who doesn't have a sufficiently recent jupyterlab (meaning newer than current stable) would mean a *massive* increase in load on the Hub and proxy, compared to the 503 behavior today. That seems a lot worse than current challenges with monitoring caused by not-admin-actionable 503s.

Using another error code like 424 results in *silent* failure, because JupyterLab doesn't show all errors to the user. Once you try certain actions like saving a notebook, you'll be shown JupyterHub's error message:

Screen Shot 2021-10-05 at 10 13 27

I think that behavior is still better than what we get from 404, which might be submitted as a separate bug to JupyterLab, because retrying at that rate probably isn't ever the right action.

manics:
When we reviewed switching mybinder to JupyterLab 3 we discussed having a more formal process for getting in breaking changes with wider implications for the community. Is this a case where it would be beneficial? Or do you consider it a minor JupyterLab bug that should be fixed without too much discussion?

minrk:
Hm, I'm not sure. Because of the user-facing consequences of this change, and the poor behavior for current latest JupyterLab, it effectively means that jupyterhub-singleuser requires at least jupyterlab >= 3.2, which is a consequence for the broader community.

But 503s being less useful for monitoring has also been a consequence for the community for the couple of years we've had the current behavior.

An advantage of handling 404 is that this is actually what we've been serving on mybinder.org for some time (mybinder.org doesn't trigger the behavior discussed here, because it has a custom backend to serve 404s when servers are gone). I think this 404 flood I just discovered is actually the reason a large majority of *all requests* to hub.mybinder.org [are 404s][].

[are 404s]: https://grafana.mybinder.org/d/ygtPwi7ik/network-activity?panelId=4&fullscreen&orgId=1&tab=visualization

Screen Shot 2021-10-05 at 11 12 43

 It should be easy to change this to match whatever we choose, though. If we switched this to 424 or 503, I think our CHP load would drop dramatically.


Another case for keeping 503: this is the error code you get if your server crashes, but is still routed from the proxy. So the error code is consistent for a server that's stopped, whether the proxy has been updated or not. However, I can see wanting to distinguish those cases as well! So from a client perspective, I still think we're doing the right thing, but I see how monitoring tools discarding the origin of the 503 makes using this common code hide real problems.

In all, it really seems to me that a distinguishable error code that doesn't have a clear, pre-existing meaning for listing files (404), and isn't conflated with deployment problems (503) is ideal, if we can find it.

jupyterhub/handlers/base.py
```diff
@@ -1522,7 +1522,7 @@ async def get(self, user_name, user_path):
             url_path_join(self.hub.base_url, "spawn", user.escaped_name, server_name),
             {"next": self.request.uri},
         )
-        self.set_status(503)
+        self.set_status(424)
```
minrk, line 1525: 👍 I think there are some surrounding comments still referencing 503. Maybe update those to indicate why we use 424, to distinguish a server being unavailable from regular bad links (404) and infrastructure problems (503) since it's been a valuable piece of information to see how many requests are being made to non-running servers, e.g. due to culling servers when users have still open (or reopened) tabs.

yuvipanda:
@minrk I've added a config option for it, but made it a boolean to just enable the older 503 behavior. I've also named it to make it clear what it's doing, although I'm open to tweaking it as needed.

minrk:
> So the issue seems to be that sometimes the 404 is from the notebook server, meaning the kernel / session is gone, and sometimes it's from JupyterHub saying that the entire server is gone. Differentiating between these two is important in the UX, as we can see in JupyterLab. Is that an accurate summary of the problem?

Yeah, we have two distinctions we seem to want to make in different contexts, because the same request can fail for 3 different reasons:

1. local server, resource not available (404)
2. server unavailable *but it should be* (deployment problem, usually 503)
3. server unavailable *and that's expected* (currently under consideration, 404 long ago, now 503, proposed 424)

JupyterLab mainly wants to distinguish between 1 and 3, because 3 requires user intervention to restart the server. Lab doesn't care overmuch *why* a server is unavailable, just that it is.

Monitoring tools want to distinguish between 2 and 3, because 2 is indicative of something actionable for a deployment administrator, whereas 3 is functioning 'as intended', though may indicate secondary issues, such as premature culling, overzealous reconnects in Lab, etc.

In the monitoring case, it seems that the fully outgoing status code is the only piece of information available (as opposed to including the upstream error code, which nginx logs, but does not record in prometheus as far as I can tell). For JupyterLab we *can* look at more information than just the error code, but the less Lab is sensitive to details of Hub behavior, the better, I think, so if we have a distinct error code for this event, it's simplest all around.

minrk:
@yuvipanda sounds great! I don't see that in the latest push, though (9cb19cc3426e7bfbb713e8a8166aa9c8e5d2af7e)

yuvipanda:
@minrk just pushed! I added a test too.

yuvipanda:
If we do this, I don't think we need to necessarily backport https://github.com/jupyterlab/jupyterlab/pull/11205 to lab 2.x. We can just add this to release notes for hub 2.0

minrk:
Yeah, and the lab PR looks to be backported to 3.1.x, which is the earliest supported lab for the lab-by-default behavior in singleuser, so I think we're in good shape.

With the given mitigations and problems solved, I think this is a good change.

yuvipanda:
Am so excited this landed! Thanks a lot, @minrk!

yuvipanda:
@meeseeksdev please backport to 1.4.x

minrk:
For folks looking for it, the jupyterlab handling of this behavior is [in 3.1.16][changelog] and later

[changelog]: https://github.com/jupyterlab/jupyterlab/blob/v3.1.16/CHANGELOG.md#maintenance-and-upkeep-improvements

yuvipanda:
image

Look at this beauty! Our background levels of 503 have dropped to 0, and now we can treat each 503 as an event we can investigate. THANK YOU!

PR #3639, opened 2021-10-05
Backport #3636 to 1.4.x

yuvipanda:
Same as #3636, but it sets the default to preserve current behavior.


yuvipanda:
Only 91c3d98  is new, so this should not create a lot of future conflicts if we backport more things in the future.

What do you think of this, @minrk?

minrk:
If it's causing problems, I think it's probably worth making a 1.5 with some off-by-default backports of new behavior, where feasible.

yuvipanda:
@minrk i agree, that would be great!

yuvipanda:
I can test this branch out once it's merged I think, and see what happens to the 503 rates.

PR #3644, opened 2021-10-07
/users/{name}/tokens should pass entire request body to authenticate

### LINKED ISSUE #3637: /users/{name}/tokens should pass entire request body to authenticate method
### Bug description
I want to pass a specific json key value to the authenticate method. The API docs does not mention I can only pass an `auth` property which is not what I expected to pass in my existing authenticate setup.

#### Expected behaviour
Pass the entire request body to the authenticate method

joe-gasewicz:
Fixes #3637

### DIFF
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -359,7 +359,7 @@ class UserTokenListAPIHandler(APIHandler):
             # defer to Authenticator for identifying the user
             # can be username+password or an upstream auth token
             try:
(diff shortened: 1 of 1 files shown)


minrk:
This isn't correct. My guess is there's some inaccurate documentation that should be fixed, not the behavior itself.

minrk:
Also a useful tip for future contributions: GitHub automatically produces links to text like "Fixes #3637" in the body, but not in the title. So it's better to put issue cross-references in the body instead.

PR #3651, opened 2021-10-14
Dynamic table for changing customizable properties of groups

vladfreeze:
Hi everyone,

for organizational reasons I have developed an additional feature in the control panel, where admins can now create and edit settings which are saved as a JSON object with key-value pairs in the jupyterhub database.

Combined with further code and hooks, this new "properties" column can simplify administrative work between groups and allow the admin to add additional settings.

As an example, I will use it to add specific amounts of RAM memory to a group, as well as a CPU count and required images. This would simplify the management of the Jupyterhub server running in our university, since each course will have its own group with customized settings.

The new feature is already fully working in frontend and backend and a backend test has been made and passed.

UI
image


minrk:
@vladfreeze getting back to this now that 2.0 is out. I rebased the PR to squash-out the reformat commits and re-sync with main, but there shouldn't be any other changes.

I think in principle, this looks great. As far as I can tell, group properties can only be set via the API. I imagine some folks (most?) would like to load fully defined groups from configuration. Can you add support in `jupyterhub/app.py` to load properties in the group definitions? It will need to change `load_groups` to allow loading a group from a dict that looks like `'groupname': {users: [usernames], properties: {}}` rather than just `'groupname': [usernames]`.

@naatebarber can I ask you to help review the react parts of it?

jupyterhub/alembic/versions/0eee8c825d24_added_properties_column.py
```diff
@@ -0,0 +1,42 @@
...
+import sqlalchemy as sa
+from jupyterhub.orm import JSONDict
+
+
+def upgrade():
+
+    try:
+        op.add_column("groups", sa.Column("properties", JSONDict))
+        print("Added column!")
+    except:
```
minrk, line 24: I think this upgrade shouldn't be allowed to fail. Let's also remove the debug-prints.

`4dc2d5a8c53c_user_options.py` is an example of what adding a column should look like.

jsx/src/components/GroupEdit/GroupEdit.jsx
```diff
@@ -67,38 +102,50 @@ const GroupEdit = (props) => {
...
+              if (changed) {
+                if (new_users.length > 0)
+                  promiseQueue.push(addToGroup(new_users, group_data.name));
+                if (removed_users.length > 0)
+                  promiseQueue.push(
+                    removeFromGroup(removed_users, group_data.name)
+                  );
+              }
 
+              if (hasDuplicates(propkeys) == true) {
```
naatebarber, line 122: We should be showing errors above the input using the `.alert.alert-danger` bootstrap class. A feature making this easier is coming soon, but for the time being try using conditional blocks of JSX instead of editing properties.

```jsx
{ 
  isError
  ? <div className="error"></div>
  : <></>
}
```

jsx/src/components/GroupEdit/GroupEdit.jsx
```diff
@@ -67,38 +102,50 @@ const GroupEdit = (props) => {
...
+                error.style.color = "red";
+              } else {
+                //error.textContent = "";
+                propkeys.forEach((key, i) => (propobject[key] = propvalues[i]));
+              }
+              if (
+                propobject != group_data.properties &&
+                hasDuplicates(propkeys) == false
+              ) {
+                promiseQueue.push(updateProp(propobject, group_data.name));
```
naatebarber, line 133: You could tie both the API call for group prop updates, and the API call for group user updates into one button click event that then returns the user to the `/groups` page. That might improve ease of use, but just an idea. @minrk thoughts?

jsx/src/util/withAPI.js
```diff
@@ -19,6 +19,8 @@ const withAPI = withProps(() => ({
     names.map((e) => jhapiRequest("/users/" + e + "/server", "DELETE")),
   addToGroup: (users, groupname) =>
     jhapiRequest("/groups/" + groupname + "/users", "POST", { users }),
+  updateProp: (propobject, groupname) =>
```
naatebarber, line 22: Update properties looks great. We'd also need a `GET` properties method so the user can see what's already stored. Right now it comes back blank when a user revisits the edit page.

jsx/src/components/DynamicTable/DynamicTable.jsx
```diff
@@ -0,0 +1,244 @@
...
+              <td>{this.renderKeyRows()}</td>
+              <td>{this.renderValueRows()}</td>
+              <td>{this.renderDelete()}</td>
+            </tr>
+          </tbody>
+        </table>
+        <form>
+          <tr>
+            <td>
+              <input
```
naatebarber, line 215: Could you try to align this input/button combination? I think there's a way to "join" them together. Check out the user input.

jsx/src/components/GroupEdit/GroupEdit.jsx
```diff
@@ -3,6 +3,19 @@ import { useSelector, useDispatch } from "react-redux";
 import { Link } from "react-router-dom";
 import PropTypes from "prop-types";
 import GroupSelect from "../GroupSelect/GroupSelect";
+import DynamicTable from "../DynamicTable/DynamicTable";
+
+function hasDuplicates(array) {
```
naatebarber, line 8: I'd simplify this and define it within the body of the GroupEdit component.

Something along the lines of
```js
const hasDuplicates = a => a.filter((e, i) => a.indexOf(e) != i).length > 0
```

naatebarber (changes requested):
Great work 👍 
I just made a couple comments on the React. With some changes and tests it LGTM.

vladfreeze:
Hello @minrk  and @naatebarber! Thank you very much for the nice and helpful feedback! I will check the code and make the changes until next week!

jsx/src/components/GroupEdit/GroupEdit.jsx
vladfreeze: Both API calls for group prop and user updates are applied on the same button (Apply). I have only changed the way the button works, so that it commits both changes to user and prop and no longer sends the user back to the /groups page, in case the user would like to make an additional change before leaving the page. At the moment this button is the only feature that is conflicting with the main branch.

jupyterhub/alembic/versions/0eee8c825d24_added_properties_column.py
vladfreeze: I structured the alembic revision so that it corresponds to the user_options one.

jsx/src/util/withAPI.js
vladfreeze: That is weird, for me the properties are displayed when leaving and entering the edit page again. I obtain the properties directly from the group_data state, because it also contains group_data.properties. I can also create a direct GET method for the properties only as well. 
If I recall correctly, this should happen here: [https://github.com/vladfreeze/jupyterhub/blob/group_property_feature/jsx/src/components/GroupEdit/GroupEdit.jsx#L53-L54](url)

vladfreeze:
> @vladfreeze getting back to this now that 2.0 is out. I rebased the PR to squash-out the reformat commits and re-sync with main, but there shouldn't be any other changes.
> 
> I think in principle, this looks great. As far as I can tell, group properties can only be set via the API. I imagine some folks (most?) would like to load fully defined groups from configuration. Can you add support in `jupyterhub/app.py` to load properties in the group definitions? It will need to change `load_groups` to allow loading a group from a dict that looks like `'groupname': {users: [usernames], properties: {}}` rather than just `'groupname': [usernames]`.
> 
> @naatebarber can I ask you to help review the react parts of it?

Hello @minrk! I have added support for adding properties to groups in app.py through load_groups and init_groups. Load_groups is now a dict and the config for loading the groups would now need to look like this {'groupname': {users: [usernames], properties: {}}}. This means that the config now requires the admins to enter the keys 'users' and 'properties' in the config when adding a new group, otherwise they can't be checked in the init_groups function.

jsx/src/components/DynamicTable/DynamicTable.jsx
```diff
@@ -0,0 +1,244 @@
+import React, { useState } from "react";
+import "./table-select.css";
+
+export default class DynamicTable extends React.Component {
```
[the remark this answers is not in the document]
vladfreeze: Hello @naatebarber, I am working on making DynamicTable functional, however I am not sure how to convert the following renderKeyRows class to a function: 

image

The problem here is that I can no longer use `var context = this;` in here because it can only be used in classes and without it, it leads to an endless loop when rendering. Is there any way to do something similar for functions? Many thanks!

jsx/src/components/GroupEdit/GroupEdit.jsx
vladfreeze: Done!

jsx/src/components/GroupEdit/GroupEdit.jsx
```diff
@@ -87,19 +110,12 @@ const GroupEdit = (props) => {
             data-testid="submit"
             className="btn btn-primary"
             onClick={() => {
-              // check for changes
```
minrk, line 90: Is this removal the result of a merge, or intentional?

minrk (commented):
Sorry for leaving you hanging for so long! This looks fine to me, with a few minor notes about backward-compatibility and some bits that look like they are from keeping up-to-date with merges.

Since you've been so patient, I'm happy to finish those myself if you'd like. It's only the `history.push("/groups")` changes that I'm not sure about, but I imagine I can figure it out.

vladfreeze:
> Sorry for leaving you hanging for so long! This looks fine to me, with a few minor notes about backward-compatibility and some bits that look like they are from keeping up-to-date with merges.
> 
> Since you've been so patient, I'm happy to finish those myself if you'd like. It's only the `history.push("/groups")` changes that I'm not sure about, but I imagine I can figure it out.

Hi @minrk ! Thank you, I just handled the issues caused by merging and made the commit you suggested. 

If I recall correctly, I made the  `history.push("/groups")` changes mostly so that the user does not get sent back to the manage groups page after clicking on apply. Now they might want to make additional changes to other entities, such as group memberships, without having to select the group page again after being sent back when applying the properties changes.

jupyterhub/app.py
```diff
@@ -2049,17 +2053,24 @@ async def init_groups(self):
...
             if group is None:
                 self.log.info(f"Creating group {name}")
                 group = orm.Group(name=name)
                 db.add(group)
-            for username in usernames:
-                username = self.authenticator.normalize_username(username)
-                user = await self._get_or_create_user(username)
-                self.log.debug(f"Adding user {username} to group {name}")
-                group.users.append(user)
+            if 'users' in contents:
```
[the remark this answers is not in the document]
minrk: Looks like this case still isn't handled. Is there a commit that hasn't been pushed, perhaps?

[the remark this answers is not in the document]
vladfreeze: Weird, I was sure I pushed it. I just made the commit again.

vladfreeze:
> OK! I think the only things missing now are coverage of the new handler:
> 
> * we always need test coverage of every API endpoint (simple exercise is enough for this one)
> * add the new endpoint to the REST API spec (`docs/source/_static/rest-api.yml`)
> 
> I can add these, if you like.

Yes sure, thank you!

minrk:
Thanks! There's a docs issue that isn't related to this PR.

vladfreeze:
Thank you very much @minrk !

manics:
I've added a https://github.com/jupyterhub/jupyterhub/labels/breaking label to this PR since it includes a DB upgrade (note it's not a breaking config change as https://github.com/jupyterhub/jupyterhub/pull/4299 fixes the backwards compatibility)

PR #3653, opened 2021-10-15
raise 404 on admin attempt to spawn nonexistent user

minrk:
in adding scope permission checks, this (and possibly other?) attempts by admins to do things on behalf of users may have lost the checks for nonexistent target entities, and instead raise 500.


minrk:
> Do we consider this a bugfix?

I'm not quite sure what you mean. It is definitely a bug to raise an unhandled error. But this is fixing a bug in the new 2.0 code, so it's not a bugfix that should go in the changelog, if that's what you mean.

consideRatio:
Do you want this to not go in the changelog at all? Then we should remove the label bug and not apply maintenance or similar.

PR #3672, opened 2021-10-30
use v2 of jupyterhub/action-major-minor-tag-calculator

minrk:
needed for handling of standard prerelease tags for Python packages


minrk:
Looks like we haven't actually pushed a v2 after https://github.com/jupyterhub/action-major-minor-tag-calculator/pull/127

manics:
Oops, I've just released v2.0.0

PR #3674, opened 2021-10-30
verify that successful login assigns default role

minrk:
and that repeated login after revoked role doesn't reassign role


meeseeksmachine:

https://discourse.jupyter.org/t/non-admin-users-cant-log-in-because-of-token-permission-errors/11174/8

consideRatio (approved):

PR #3677, opened 2021-11-02
1.4.x: update doc requirements

minrk:
pin down docutils, unpin autodoc-traits


minrk:
Noticed the doc build failed in #3676. This one fixes it.

PR #3681, opened 2021-11-08
Log single-user app versions at startup

minrk:
Reports jupyterlab, jupyter_server versions during startup

For easier debugging of single-user environment logs, where the first question is often "what version of jupyterhub and jupyterlab?"

Log output looks like:

```
[I 2021-11-08 09:17:37.223 SingleUserLabApp mixins:617] Starting jupyterhub single-user server version 2.0.0rc3
[I 2021-11-08 09:17:37.223 SingleUserLabApp mixins:629] Extending jupyterlab.labhubapp.SingleUserLabApp from jupyterlab 3.2.1
[I 2021-11-08 09:17:37.223 SingleUserLabApp mixins:629] Extending jupyter_server.serverapp.ServerApp from jupyter_server 1.9.0
```

### DIFF
--- a/jupyterhub/singleuser/mixins.py
+++ b/jupyterhub/singleuser/mixins.py
@@ -18,6 +18,7 @@ import sys
 import warnings
 from datetime import datetime
 from datetime import timezone
+from importlib import import_module
 from textwrap import dedent
 from urllib.parse import urlparse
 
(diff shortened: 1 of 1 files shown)


yuvipanda:
Thanks, @minrk

PR #3683, opened 2021-11-16
Strict session id checking

minrk:
Singleuser oauth works like this:

- a shared, opaque `jupyterhub-session-id` cookie is set on the browser
- oauth flow issues a token which is associated with the session id, which is stored by the singleusre server in a cookie for future requests

The session id is used exclusively for token revocation, by:

1. on logout, any tokens associated with the session id are revoked
2. session id is included in the cache key in singleuser auth, so that clearing / changing the session-id invalidates the singleuser token response cache, so that even if the expired token is in the cookie, it will notice immediately upon checking with the hub and requests will be rejected

However, it is not strict - if session_id is not set or artificially cleared, requests are still allowed if the token has not been revoked. The result is that if a token is stored in the cookie that is not associated with the current session id, it will not be revoked, and can still be used, resulting in 'sticky' singleuser logins even after logging out from the Hub.

This patch adds a `strict_session_id` behavior to singleuser auth and enables it by default:

1. if session id is not set, cookie auth is always rejected. This makes it _very_ unlikely for logout from jupyterhub to leave any valid sessions with single-user servers
2. The session_id of the token is added to the `/user` response model, and the session id is checked, which means that _changing_ session id also ensures tokens are effectively invalidated.

### DIFF
--- a/jupyterhub/services/auth.py
+++ b/jupyterhub/services/auth.py
@@ -28,6 +28,7 @@ from tornado.httputil import url_concat
 from tornado.log import app_log
 from tornado.web import HTTPError
 from tornado.web import RequestHandler
+from traitlets import Bool
 from traitlets import default
 from traitlets import Dict
 from traitlets import Instance
@@ -487,7 +488,6 @@ class HubAuth(SingletonConfigurable):
         - in URL parameters: ?token=<token>
         - in header: Authorization: token <token>
         """
-
         user_token = handler.get_argument('token', '')
         if not user_token:
             # get it from Authorization header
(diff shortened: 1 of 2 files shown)


minrk:
There is one test that's failing with this change, and it is the behavior of logging in a single-user session with a single visit to a URL like `https://hub.example/user/name/?token=xyz`. This works today, because the token is persisted into a cookie and the user is logged in with the single-user server, but they are never logged in with the Hub, which means there is never a session-id. That's exactly the circumstance this change is meant to prevent, but it's also something people really use.

Question: do we want to allow this pattern, and if so, is there an alternative? Perhaps `/hub/user-redirect/?token=...` should work? Maybe it does already?

meeseeksmachine:

https://discourse.jupyter.org/t/rfc-links-with-authentication-token-to-jupyterhub/11714/1

minrk:
>  Perhaps /hub/user-redirect/?token=... should work? Maybe it does already?

It does not. The Hub does not accept tokens in URLs, only in headers. 2.0 will not accept tokens at all for any human pages, only API endpoints.

This may have worked in the past, but not with RBAC - that would be a privilege escalation, as cookie-based auth always has full owner permissions, while a token may have significantly less.

I think this requires more thought, and we can make it opt-in at first. I extracted the model changes to #3685 so that the transition can be smoother (strict session id checking would be purely a singleuser-side enforcement check, with no associated API changes)

jupyterhub/services/auth.py
```diff
@@ -598,15 +598,54 @@ def _get_user_cookie(self, handler):
...
+        Requires session-id to match the token in a cookie.
+
+        If a session-id cookie is not set, no cookie-based auth will be accepted.
+        Additionally, the token in the cookie must be associated with the session id to be accepted.
+
+        Can be disabled via JUPYTERHUB_STRICT_SESSION_ID=0 to preserve JupyterHub 1.x behavior.
+
+        .. versionadded:: 2.0
+        """,
+    ).tag(config=True)
```
consideRatio, line 639: This help text could benefit from a clarification that this is configuration relevant on the singleuser-server rather than the jupyterhub server, if those are separate.

consideRatio, line 639: This help text could benefit from a clarification that this is configuration relevant on the singleuser-server rather than the jupyterhub server, if those are separate.

minrk: Such a note probably goes in the docstring for HubAuth and HubOAuth, to make it clear that these are set in the single-user and/or service environment.

minrk:
There are alternate implementations for how to implement the session-id check, e.g. doing the check Hub-side with a `?session_id=` GET param, so it doesn't return a model if session id is specified and it doesn't match. It's still on the oauth client to send the param, but not check it, so it doesn't really make a difference, I think. If we did the session-id check hub-side, we could do revocation on a mismatch, I suppose.

minrk:
With RBAC, one option is to define a `login` scope, such that you could visit `/hub/login?token=...&next=/hub/user-redirect/...` with a token, and only if it has that scope would it be allowed to complete login and redirect to a single-user server. Or we could explicitly check for the existing `inherit` metascope, since that means the token has exactly the same permissions a cookie would have.

The question is: for folks sharing URLs to single-user servers with tokens, do they _want_ to prevent them from logging into the Hub? Do folks want to also be able to login to the Hub with a token without going straight to a server?

Answering these will help us identify which set of new features / restrictions we shoudl take.

minrk:
Adding a note that one case we can't possibly cover with this: services on different hosts (i.e. using JupyterHub as a standard oauth provider for external services). In this case, services cannot access the session id, so cannot tell when it's changed or revoked.

So this mostly means we can only really do this for single-user servers. On one hand, that's not great because the auth really is the same for services and single-user servers. On the other, it would still cover 90% of cases _in practice_ to do this, even if it only covered same-host services/single-user servers. We just need to make sure it can be applied to same-host services but not external-host services at the same time.

I think maybe the best thing to do is to improve behavior of refreshing/expired tokens, and shorten default token expiry. That will prompt refresh more often, as is best practice with oauth. But right now, the behavior of a token expiring during a long-running jupyterlab session is not good.

PR #3686, opened 2021-11-18
Hub: only accept tokens in API requests

minrk:
do not allow token-based access to pages

Tokens are only accepted via Authorization header (not in the url via `?token=...` like the single-user server), which doesn't make sense to pass to pages, so disallow it explicitly to avoid surprises.

### DIFF
--- a/jupyterhub/apihandlers/base.py
+++ b/jupyterhub/apihandlers/base.py
@@ -31,6 +31,9 @@ class APIHandler(BaseHandler):
     - methods for REST API models
     """
 
+    # accept token-based authentication for API requests
+    _accept_token_auth = True
+
     @property
     def content_security_policy(self):
         return '; '.join([super().content_security_policy, "default-src 'none'"])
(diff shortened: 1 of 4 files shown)


consideRatio (commented):
Perhaps we can take a video chat about this PR? I have some questions that perhaps is easier to be answered via a chat, I've raised them below.

---

This seems okay but I can't really say. Could you clarify if my understanding of the PR is correct?

### The original PR description

>do not allow token-based access to pages
>
>Tokens are only accepted via Authorization header (not in the url via ?token=... like the single-user server), which doesn't make sense to pass to pages, so disallow it explicitly to avoid surprises.

### My PR understanding as a PR description

As JupyterHub api-tokens were designed to be used against JupyterHub's api-endpoints and as an option to cookie authentication to access a user server, it doesn't make sense for JupyterHub to accept them as credentials to typical JupyterHub pages that aren't api endpoints. Due to that, this PR will make these typical JupyterHub pages only accept cookie based authentication.

Note that while JupyterHub only accepts its api-tokens via an Authorization header, a user server can accept such JupyterHub token via a query parameter (`?token=...`).

### Specific question to verify my understanding

Is it correct that a the user servers accepts a jupyterhub token as can be generated via /hub/home, and when receiving one will verify its authorization with JupyterHub, and based on that set a cookie that it allows going onwards?

PR #3690, opened 2021-11-24
build jupyterhub/singleuser along with other images

minrk:
got lost in the migration to GHA docker builds


.github/workflows/release.yml
```diff
@@ -190,3 +190,31 @@ jobs:
           platforms: linux/amd64
           push: true
           tags: ${{ join(fromJson(steps.demotags.outputs.tags)) }}
+
+      # jupyterhub/singleuser
+
+      - name: Get list of just version tags
+        id: tags
+        uses: jupyterhub/action-major-minor-tag-calculator@v2
```
minrk, line 198: Is there a way we can run this only once without `$prefix` and add the prefix on push, rather than N + 1 times for each image?  e.g. `tags: tags: ${{ env.REGISTRY + 'juptyerhub/jupyterhub:' + join(fromJson(steps.tags.outputs.tags), ',' + env.REGISTRY + 'juptyerhub/jupyterhub:') }}`

consideRatio: Maybe, but instead of adding such complexity, can we instead make this job definition be simplified to run x steps N times with different parameterization (one per image), instead of being a very long and complicated job with N*x steps for N images?

manics: In general yes, though it's a bit more complicated.

If you set `prefix: ""` instead of `prefix: registry/org/name:` it should return just the tags, but since it's an array you'll need to loop through instead of just joining:
```
tags = []
for tag in tags:
  tags.push(env.REGISTRY + 'juptyerhub/jupyterhub:' + tag)
```

One alternative is to modify the action to optionally take multiple prefixes and return a dictionary:
```yaml
uses: jupyterhub/action-major-minor-tag-calculator@v2
with:
  prefixes:
    - "${{ env.REGISTRY }}jupyterhub/jupyterhub:"
    - "${{ env.REGISTRY }}jupyterhub/singleuser:"
```
Returning
```json
"tags": {
  "docker.io/jupyterhub/jupyterhub:": ["docker.io/jupyterhub/jupyterhub:1.5", "docker.io/jupyterhub/jupyterhub:1.5.0"],
  "docker.io/jupyterhub/singleuser:": ["docker.io/jupyterhub/singleuser:1.5", "docker.io/jupyterhub/singleuser:1.5.0"]
}
```

consideRatio: If you in this PR make this a clean copy-paste of the previous logic, without deviations from that structure and way of doing things, then if you want, I could refactor this later into a single job definition that takes parameters to run N times where N is the number of images we should build/publish.

minrk: Yup, that's exactly what I did. I'll leave it at that for now, and we can work out a simplification later, if it's worth it. The duplication happened to bite me in #3673 where a version bump of tag-calculator must be applied _5_ times after this PR, I think.

I hope to resolve https://github.com/jupyterhub/team-compass/issues/474 soon, in which case all docker building will be removed from this repo, anyway, and we'll probably want to change some of this logic when we do that (tag-calculator may not be appropriate anymore when we do that).

minrk:
Since the only way to test the image publishing is to make and publish a tag, I plan to land this and make a (final?) 2.0.0rc5 tomorrow.

manics (approved):

PR #3695, opened 2021-11-30
Service auth doc

minrk:
Update the services authentication docs with 2.0 in mind

A few things had gotten stale, either with the current sphinx theme (the in-body TOC is redundant with the sidebar TOC), or references to things that are "new in 0.7", or deprecated in 2.0.

Mostly, it's pointing things toward using HubOAuth instead of the base HubAuth class, since HubAuth cannot be used to authenticate a service with a browser, HubOAuth is what most services should use, so should be the main target of the docs.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -205,7 +205,10 @@ epub_exclude_files = ['search.html']
 
 # -- Intersphinx ----------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


consideRatio (approved):
Thanks @minrk! This LGTM!

Since this is a PR blocking release, while also being non-critical to get perfectly right away, I'm inclined to go for a merge right away.

PR #3696, opened 2021-11-30
Changelog for 2.0

minrk:
I think we're ready for 2.0. We got lots of useful feedback and fixed things during the betas, but there's nothing outstanding that I know of, and there's not much new feedback coming in. I'll plan to do the release tomorrow or Thursday, unless someone has something pressing.

Changelog includes #3695, which will be merged before I cut 2.0.


consideRatio (approved):
This LGTM!

I suggest you go for a self-merge of this right before you go for a release tomorrow if nothing popped up.

minrk:
I have a meeting shortly, then plan to cut this release in an hour or two.

manics:
Might be good to mention a database upgrade is required, and link to https://jupyterhub.readthedocs.io/en/stable/admin/upgrading.html in the pre-amble? Otherwise looks fine to me!

minrk:
@manics adding that now.

minrk:
Assuming docs pass, will merge and publish 2.0.0

consideRatio:
Hmmm, it sais database backup etc, but it sounds like the db is upgraded automatically even without --upgrade, is that so?

minrk:
> it sounds like the db is upgraded automatically even without --upgrade, is that so?

No, JupyterHub will exit if the schema version has changed, unless you pass the upgrade arg. Where do you see the suggestion that it will be upgraded without input?

consideRatio:
> JupyterHub 2.0 updates the database schema,
so **make sure to [backup your database](admin/upgrading) before upgrading**.

To me, this sounds like upgrading to v2 means JH2 updated thr DB. I'd appreciate having a clarification like...

"You need to upgrade the database schema to successfully start jupyterhub, so pass --upgrade, but take a backup first..." Or similar im thinking.

consideRatio (approved):

minrk:
2.0 is [published](http://pypi.org/project/jupyterhub/), [announced](https://blog.jupyter.org/jupyterhub-2-0-7a038715e96d), etc. 🎉 :shipit: 🚀

Issue #3606, opened 2021-09-16
error handling in react admin UI


minrk:
requests from the admin UI may fail, either being malformed (e.g. creating an invalid username), or permissions (while a user may have permission to access the admin page, they may not have permission to take every action, e.g. creating groups or shutting down the hub).

As it is now, errors are only logged to the console, there is no feedback to the user

naatebarber:
Good catch, will take a look at this.

naatebarber:
Addressed in https://github.com/jupyterhub/jupyterhub/pull/3697 🚀

PR #3697, opened 2021-12-01
React Error Handling

naatebarber:
@minrk This PR is addressing issue https://github.com/jupyterhub/jupyterhub/issues/3606

### What's been done:
 - Added in UI error dialogues for the client API requests, which should appear whenever anything out of the ordinary happens. Much less sparse than before.
 - Updated our testing library from enzyme (which is getting dated and was relying on a third party adapter to work with React 17) to React Testing Library. The latter is much more in tune with the testing of Hooks-based React code and is being actively maintained. 
 
I also plan on making the error messages more specific based on the response status code, just thought this was a good start.

Please let me know if there's anything else you guys see that could be improved on here 👍


minrk:
This is looking great! Anything more you want to do in the PR before merge?

naatebarber:
Thanks! I think it's good to merge.

yuvipanda:
\o/ Awesome, @naatebarber!

PR #3701, opened 2021-12-02
cors: handle mismatched implicit/explicit ports in host header

minrk:
http://host:80 should match http://host

parse URLs to make it easier to do correct comparisons.

cors tests are parametrized to make it easier to add more cases

closes #3027

### DIFF
--- a/jupyterhub/apihandlers/base.py
+++ b/jupyterhub/apihandlers/base.py
@@ -58,7 +58,8 @@ class APIHandler(BaseHandler):
 
         - allow unspecified host/referer (e.g. scripts)
         """
-        host = self.request.headers.get(self.app.forwarded_host_header or "Host")
+        host_header = self.app.forwarded_host_header or "Host"
+        host = self.request.headers.get(host_header)
         referer = self.request.headers.get("Referer")
 
         # If no header is provided, assume it comes from a script/curl.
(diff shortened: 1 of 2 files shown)


consideRatio (approved):
I tried reviewing this very carefully as it is security related logic. It LGTM though!

lvazquez:
These changes broke reverse proxy handling (Nginx, Apache mod_proxy) when forwarded protocol is https and the JupyterHub backend is working in plain http.

The proto variable in check_referer() is being taken directly from self.request.protocol , which is wrong in case of a proxied service, and this is combined with the app.forwarded_host_header or "Host" header (both can be configured to take the proxied address into account) to construct the (wrong) full_host URL string (http) which is in turn compared with the referer_url (https).

This yields to a "Blocking Cross Origin API request" even when the refererer and current url are the same (in the proxy) and there is no way to fix this with the traditional proxy header "X-Forwarded-Proto" or the more recent  "Forwarded" header.

In the patch bellow I have fixed this for using the "X-Forwarded-Proto" proxy header, but this should be fixed more generally to also handle the (now standard) "Forwarded" header format
```
--- base.py.2.0.1       2021-12-22 16:14:55.000000000 -0300
+++ base.py     2021-12-29 16:38:43.570041997 -0300
@@ -71,7 +71,16 @@
             self.log.warning("Blocking API request with no referer")
             return False

+        import re
+        # use local request protocol as default
         proto = self.request.protocol
+        # check proxy forwarded protocol header
+        xf_proto = self.request.headers.get('X-Forwarded-Proto', None)
+        if xf_proto:
+            forwarded_proto = xf_proto.split(',',1)[0]
+            if re.match(r'https?', forwarded_proto):
+                proto = forwarded_proto
+
         full_host = f"{proto}://{host}{self.hub.base_url}"
         host_url = urlparse(full_host)
         referer_url = urlparse(referer)
```
Best regards

manics:
@lvazquez `xheaders=True` in:
https://github.com/jupyterhub/jupyterhub/blob/2.0.1/jupyterhub/app.py#L3065
so according to https://www.tornadoweb.org/en/stable/httpserver.html#http-server the `X-*` headers should be taken into account.

Please could you add your JupyterHub deployment details and logs to https://github.com/jupyterhub/jupyterhub/issues/3737 ?

fritterhoff:
> These changes broke reverse proxy handling (Nginx, Apache mod_proxy) when forwarded protocol is https and the JupyterHub backend is working in plain http.
> 
> The proto variable in check_referer() is being taken directly from self.request.protocol , which is wrong in case of a proxied service, and this is combined with the app.forwarded_host_header or "Host" header (both can be configured to take the proxied address into account) to construct the (wrong) full_host URL string (http) which is in turn compared with the referer_url (https).
> 
> This yields to a "Blocking Cross Origin API request" even when the refererer and current url are the same (in the proxy) and there is no way to fix this with the traditional proxy header "X-Forwarded-Proto" or the more recent "Forwarded" header.

Similar issue in my case. I also have (major) problems caused by CORS. In my case I'm using caddy in front of jupyterhub and the logs are full of warnings.

```
jupyterhub | [W 2022-01-06 09:33:49.545 JupyterHub base:89] Blocking Cross Origin API request.  Referer: https://**redacted**/hub/spawn-pending/florian.ritterhoff, Host: **redacted**, Host URL: http://**redacted**/hub/
```

At the moment I manually patch the file so the setup keeps running ;)

meeseeksmachine:

https://discourse.jupyter.org/t/bug-in-2-0-1-pending-spawn-pending-after-login-due-to-missing-read-servers-scope/12377/7

PR #3708, opened 2021-12-07
Avoid clearing user role membership when defining custom user scopes

### LINKED ISSUE #3703: Problems overriding the default roles
### Bug description
It seems that overriding the role `user` is not functioning correctly. Specifically, on jh 2.0.0 if I set up the following role config (see [here](https://discourse.jupyter.org/t/plans-on-bringing-rtc-to-jupyterhub/9813/11?u=anton_akhmerov) for motivation)

```python
c.JupyterHub.load_roles = [
    {
        "name": "user",
        "description": "Allow users to access the shared server in addition to default perms",
        "scopes": ["self", "access:servers!user=shared"],
    }
]
```

Now after starting the hub, any non-admin user that existed on the hub before the start is unable to launch their server and sees a 500 error:

image

After deleting and recreating the user from the hub admin panel, everything seems to function normally.

minrk:
If the user role was defined but did not specify a user membership list, users granted access by the Authenticator would lose their status on hub startup.

Instead, do nothing on an undefined user membership list, leaving any users with their existing default role assignment

closes #3703

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -2015,14 +2015,25 @@ class JupyterHub(Application):
 
     async def init_role_creation(self):
         """Load default and predefined roles into the database"""
-        self.log.debug('Loading default roles to database')
+        self.log.debug('Loading roles into database')
         default_roles = roles.get_default_roles()
         config_role_names = [r['name'] for r in self.load_roles]
 
-        init_roles = default_roles
+        default_roles_dict = {role["name"]: role for role in default_roles}
+        init_roles = []
         roles_with_new_permissions = []
         for role_spec in self.load_roles:
             role_name = role_spec['name']
+            if role_name in default_roles_dict:
+                self.log.debug(f"Overriding default role {role_name}")
+                # merge custom role spec with default role spec when overriding
+                # so the new role can be partially defined
+                default_role_spec = default_roles_dict.pop(role_name)
+                merged_role_spec = {}
+                merged_role_spec.update(default_role_spec)
+                merged_role_spec.update(role_spec)
+                role_spec = merged_role_spec
+
             # Check for duplicates
             if config_role_names.count(role_name) > 1:
                 raise ValueError(
@@ -2037,6 +2048,9 @@ class JupyterHub(Application):
                         "Role %s has obtained extra permissions" % role_name
                     )
                     roles_with_new_permissions.append(role_name)
+
+        # make sure we load any default roles not overridden
+        init_roles = list(default_roles_dict.values()) + init_roles
         if roles_with_new_permissions:
             unauthorized_oauth_tokens = (
                 self.db.query(orm.APIToken)
@@ -2063,7 +2077,7 @@ class JupyterHub(Application):
         for role in self.db.query(orm.Role).filter(
             orm.Role.name.notin_(init_role_names)
         ):
-            app_log.info(f"Deleting role {role.name}")
+            app_log.warning(f"Deleting role {role.name}")
             self.db.delete(role)
         self.db.commit()
         for role in init_roles:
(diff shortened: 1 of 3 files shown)


jupyterhub/app.py
```diff
@@ -2014,14 +2014,25 @@ async def init_groups(self):
...
         default_roles = roles.get_default_roles()
         config_role_names = [r['name'] for r in self.load_roles]
 
-        init_roles = default_roles
+        default_roles_dict = {role["name"]: role for role in default_roles}
+        init_roles = []
         roles_with_new_permissions = []
         for role_spec in self.load_roles:
             role_name = role_spec['name']
+            if role_name in default_roles_dict:
```
minrk, line 2026: Reworked how default roles were overridden - previously, they were loaded twice - once with default spec, then again with any overrides. This merges the spec so that there aren't two logged changes to the role's permissions when there's no actual change in config (change from db to default; change from default to config, which matched db).

jupyterhub/app.py
```diff
@@ -2104,43 +2119,57 @@ async def init_role_assignment(self):
...
+                    setattr(role, kind, orm_role_bearers)
+                else:
+                    # no defined members
+                    # leaving 'users' undefined in overrides of the default 'user' role
+                    # should not clear membership on startup
+                    # since allowed users could be managed by the authenticator
+                    if kind == "users" and role_name == "user":
+                        # Default user lists can be managed by the Authenticator,
+                        # if unspecified in role config
+                        pass
```
minrk, line 2168: This is where the real fix is for #3703 - if `user` is defined, but `user.users` is not, do not clear membership (i.e. behave the same as if `user` itself was not defined).

This is _only_ the case for the default `user` role. Any other role losing its `users` member list is equivalent to setting an empty list, i.e. revoking membership from all users.

The main case where this happens is when `allowed_users` is empty, so any authenticated user should be allowed.

jupyterhub/roles.py
```diff
@@ -239,26 +239,6 @@ def _check_scopes(*args, rolename=None):
                 )
 
 
-def _overwrite_role(role, role_dict):
-    """Overwrites role's description and/or scopes with role_dict if role not 'admin'"""
-    for attr in role_dict.keys():
-        if attr == 'description' or attr == 'scopes':
-            if role.name == 'admin':
```
minrk, line 246: I moved the admin check to `create_role` so it would behave the same whether `admin` were in the db or not, which was not previously the case. Mostly a minor implementation detail.

jupyterhub/app.py
```diff
@@ -2104,43 +2119,57 @@ async def init_role_assignment(self):
             # tokens need to be checked for permissions
             for kind in kinds:
                 orm_role_bearers = []
-                if kind in predef_role.keys():
-                    for bname in predef_role[kind]:
+                if kind in role_spec:
+                    for name in role_spec[kind]:
                         if kind == 'users':
```
akhmerov, line 2124: According to the [docs](https://jupyterhub.readthedocs.io/en/stable/rbac/roles.html#roles), the default role name is `user` not `users`. Is this another bug?

minrk:
Added tests covering various combinations of users in the db and assignments coming from config, so I think this is good to go.

jupyterhub/app.py
```diff
@@ -2032,10 +2043,13 @@ async def init_role_creation(self):
             old_role = orm.Role.find(self.db, name=role_name)
             if old_role:
                 if not set(role_spec['scopes']).issubset(old_role.scopes):
-                    app_log.warning(
+                    self.log.warning(
```
minrk, line 2046: These don't change anything, but good to be consistent referencing `self.log`. They point to the same logger, ultimately.

PR #3714, opened 2021-12-13
Grant role after user creation during config load

a3626a:
I faced role related issue for my jupyterhub deployment:
- Using Zero To JupyterHub with Kubernetes
- Using GenericOAuthenticator
  - ```python
      from oauthenticator.generic import GenericOAuthenticator
      c.JupyterHub.authenticator_class = GenericOAuthenticator
      c.GenericOAuthenticator.oauth_callback_url = "http://<my-jupyter-hub-domain>/hub/oauth_callback"
      c.GenericOAuthenticator.client_id = "<client-id>"
      c.GenericOAuthenticator.client_secret = "<client-secret>"
      c.GenericOAuthenticator.login_service = "<my-service>"
      c.GenericOAuthenticator.authorize_url = "http://<my-oauth-provider-domain>/oauth/authorize"
      c.GenericOAuthenticator.userdata_url = "http://<my-oauth-provider-domain>/api/v1/me"
      c.GenericOAuthenticator.token_url = "http://<my-oauth-provider-domain>/oauth/token"
      c.GenericOAuthenticator.username_key = "email"
    ```
 - Loading Roles Like
   - ```
      c.JupyterHub.load_roles = [
        {
          "name": "role-for-rtc",
          "scopes": ["servers", "read:users", "access:servers"],
          "users": ["<rtc-privileged-user>@<my-domain>"]
        }
      ]
     ```
Under this condition:
  - User `<rtc-privileged-user>@<my-domain>` is not yet been created (signed in)

-> User `<rtc-privileged-user>@<my-domain>` cannot access to its own server.
- ```
  [W 2021-12-13 02:44:18.992 JupyterHub roles:436] Token requesting role server with scopes not held by owner <rtc-privileged-user>@<my-domain>: {<List of scopes which change depending on the scopes listed in load_roles>'}
  ```
- (I removed some part for my security)
![image](https://user-images.githubusercontent.com/4434752/145815166-c3594a12-1e64-4668-8726-eebd3024d346.png)

Under the other condition:
  - User `<rtc-privileged-user>@<my-domain>` is already created.
    - To reset everything, I deleted the user first.
    - I deleted my `load_roles` lines, too.
    - Then, I created the user by logging in.
    - Finally, I added my `load_roles` in my config, and restarted the hub.

-> Everything works fine.


In conclusion, I suggest these changes to fix this issue.
  - At app.py#L2131, `orm_obj` is always created as an new user. So it's fine to grant default role here.
  - In the cases mentioned above, this user is not granted for their default `user` role.
  - Without `roles.grant_role(db, orm_obj, 'user')`, default user role is not granted to the user, because it is not listed in `self.authenticator.allowed_users`

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1993,7 +1993,7 @@ class JupyterHub(Application):
             self.log.info(f"Creating user {username}")
             user = orm.User(name=username)
             self.db.add(user)
-            roles.grant_role(self.db, user, 'user')
+            roles.assign_default_roles(self.db, entity=user)
             self.db.commit()
         return user
 
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -2128,6 +2128,7 @@ class JupyterHub(Application):
                             if kind == 'users':
                                 orm_obj = await self._get_or_create_user(bname)
                                 orm_role_bearers.append(orm_obj)
+                                roles.grant_role(db, orm_obj, 'user')
                             elif kind == 'groups':
                                 group = orm.Group(name=bname)
                                 db.add(group)
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1993,6 +1993,7 @@ class JupyterHub(Application):
             self.log.info(f"Creating user {username}")
             user = orm.User(name=username)
             self.db.add(user)
+            roles.grant_role(self.db, user, 'user')
             self.db.commit()
         return user
 
@@ -2128,7 +2129,6 @@ class JupyterHub(Application):
                             if kind == 'users':
                                 orm_obj = await self._get_or_create_user(bname)
                                 orm_role_bearers.append(orm_obj)
-                                roles.grant_role(db, orm_obj, 'user')
                             elif kind == 'groups':
                                 group = orm.Group(name=bname)
                                 db.add(group)
--- a/jupyterhub/tests/test_app.py
+++ b/jupyterhub/tests/test_app.py
@@ -247,6 +247,7 @@ async def test_load_groups(tmpdir, request):
         kwargs['internal_certs_location'] = str(tmpdir)
     hub = MockHub(**kwargs)
     hub.init_db()
+    await hub.init_role_creation()
     await hub.init_users()
     await hub.init_groups()
     db = hub.db
--- a/jupyterhub/tests/test_roles.py
+++ b/jupyterhub/tests/test_roles.py
@@ -1203,6 +1203,27 @@ async def test_user_config_respects_memberships():
         assert user in user_role.users
 
 
+async def test_user_config_creates_default_role():
+    role_spec = [
+        {
+            'name': 'new-role',
+            'scopes': ['read:users'],
+            'users': ['not-yet-created-user'],
+        }
+    ]
+    user_names = []
+    hub = MockHub(load_roles=role_spec)
+    hub.init_db()
+    hub.authenticator.allowed_users = user_names
+    await hub.init_role_creation()
+    await hub.init_users()
+    await hub.init_role_assignment()
+    user_role = orm.Role.find(hub.db, 'user')
+    new_role = orm.Role.find(hub.db, 'new-role')
+    assert orm.User.find(hub.db, 'not-yet-created-user') in new_role.users
+    assert orm.User.find(hub.db, 'not-yet-created-user') in user_role.users
+
+
 async def test_admin_role_respects_config():
     role_spec = [
         {


jupyterhub/app.py
```diff
@@ -2128,6 +2128,7 @@ async def init_role_assignment(self):
                             if kind == 'users':
                                 orm_obj = await self._get_or_create_user(bname)
                                 orm_role_bearers.append(orm_obj)
+                                roles.grant_role(db, orm_obj, 'user')
```
minrk, line 2131: I think this makes sense, but can you move this assignment to within `_get_or_create_user` and only `if not username in self.authenticator.allowed_users:`?

minrk:
xref #3708 which is also related to users not having the `user` role at startup, who should, and also caused by the "any authenticated user should be allowed" configuration where 'user' role assignment comes programmatically from the authenticator, rather than role config.

Here's a question that could simplify things, depending on the answer:

Are there any users who, when `allowed_users` is empty, are present in the database and _shouldn't_ have the `user` role?

If not, we can simplify it to a check at the end of `init_role_assignment` like we do [here if allowed_users](https://github.com/jupyterhub/jupyterhub/blob/2.0.0/jupyterhub/app.py#L2135-L2140), to assign the role to all users in the db `if not allowed_users` and the `user.users` membership list is not explicitly defined. Then we'd have a simpler "ensure everything is consistent" final step that's easier to be confident in.

akhmerov:
> Are there any users who, when `allowed_users` is empty, are present in the database and shouldn't have the `user` role?

Wouldn't this be against the role definition in the docs, which says that all users should have `user` role?

jupyterhub/app.py
a3626a: > move this assignment to within _get_or_create_user

In some cases `_get_or_create_user` is called before default roles (ex. `'user'`) are created. So I cannot move it.


> only if not username in self.authenticator.allowed_users:

You seem to be worried that `role.grant_role` is called twice. `role.grant_role` is idempotent. I think that it is better to let the code simple.

a3626a:
> Are there any users who, when allowed_users is empty, are present in the database and shouldn't have the user role?

Test case `test_login_default_role` expects such kind of user to exist.

By the way, if someone want to remove `'user'` role from an user, they would better to create a service.

minrk:
One of the benefits of the RBAC switch is that a user can have their permissions reduced to zero without needing to _delete_ the user from the db. This is, in most cases, deleting their `user` role assignment (and all others).

However, I think things aren't quite right for the various cases of a User record being created for the first time, which can occur in a few places:

- explicit in allowed_users (assigns `user` role, this one is simple). Same for explicit `user.users` role membership.
- on first authentication (assigned user _if_ it doesn't already have a role assigned - we should probably always include 'user' here, rather than making it dependent on whether there are other roles)
- user declared for the first time in a custom role - this _prevents_ `user` role assignment on login, making an inconsistent state, depending on if this role assignment occurred before or after their first login
- user declared for the first time in a Group (mostly the same as in a role here, in that they will not have `user` on hub start, but they _will_ get it on first login because of the check for 0 role assignments there)

I think in all of these cases, it makes sense for them to get `user` assigned at creation time.

So I think at the very least, we should change the `assign_default_roles` to always assign `user`, no matter what, rather than only doing so if there are no other roles. We should make sure to call this function in the `_get_or_create_user` utility during startup. We also happen to _switch_ role assignments from user<-->admin, and I don't think there's a reason to do that. It's always fine for admin to be _in addition to_ user, and simpler that way.

Then I think the main question is: what do we do with users already in the db who have no role assignments, when not using `allowed_users`? They will get their role assignment the next time they login successfully if we fix the empty-roles check. But they won't get it until then, which means that e.g. an admin starting the server will fail the permission check. However, maybe this is the _right_ thing to do, because the role being revoked should mean they lost permission, and only if they successfully authenticate again, should they regain those permissions.  There isn't an obvious manual intervention for administrators to grant users the `user` role who may have the wrong assignment, other than `python -m jupyterhub.dbutil shell` which is a _very_ low-level, advanced option. They can temporarily set `allowed_users`, I suppose.

jupyterhub/app.py
minrk: > In some cases _get_or_create_user is called before default roles (ex. 'user') are created.

That [shouldn't be the case](https://github.com/jupyterhub/jupyterhub/blob/95649a3ece8532aa6d3e6bb47973434930fd61b8/jupyterhub/app.py#L2763-L2765), but it might be a missing call to `init_role_creation()` in some tests that use only a partially initialized Hub.

> You seem to be worried that role.grant_role is called twice

No, I'm just trying to get the code in the right place. This fixes one instance of the bug, but the same bug occurs in two places (here and in load_groups) - both places call `_get_or_create_user`, so that seems like the logical place - it shouldn't matter _why_ a User is getting created. Plus, this code involved db queries on a per-user basis, so there's a cost to calling it unnecessarily, even if it is idempotent.

a3626a: > That shouldn't be the case, but it might be a missing call to init_role_creation() in some tests that use only a partially initialized Hub.

I was too careful because of the test case. I don't have enough understanding of this project to modify existing test cases. I agree with your idea now.

minrk:
mysql failure is due to #3718, and not related to this PR.

Thanks @a3626a!

PR #3716, opened 2021-12-14
Fix error message about Authenticator.pre_spawn_start

minrk:
This isn't the only or even main thing likely to raise here, so don't blame it, which is confusing, especially in a message shown to users.

Has come up during spawn failures, e.g. #3703

Log the full exception, and show a more opaque message to the user to avoid confusion.

### DIFF
--- a/jupyterhub/handlers/pages.py
+++ b/jupyterhub/handlers/pages.py
@@ -308,10 +308,13 @@ class SpawnHandler(BaseHandler):
         # otherwise it may cause a redirect loop
         if f.done() and f.exception():
             exc = f.exception()
+            self.log.exception(f"Error starting server {spawner._log_name}: {exc}")
+            if isinstance(exc, web.HTTPError):
+                # allow custom HTTPErrors to pass through
+                raise exc
             raise web.HTTPError(
                 500,
-                "Error in Authenticator.pre_spawn_start: %s %s"
-                % (type(exc).__name__, str(exc)),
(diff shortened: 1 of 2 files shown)


jupyterhub/handlers/pages.py
```diff
@@ -308,10 +308,13 @@ async def _wrap_spawn_single_user(
         # otherwise it may cause a redirect loop
         if f.done() and f.exception():
             exc = f.exception()
+            self.log.exception(f"Error starting server {spawner._log_name}: {exc}")
+            if isinstance(exc, web.HTTPError):
+                # allow custom HTTPErrors to pass through
+                raise
```
manics, line 314: It took me a while to understand this Python syntax (https://stackoverflow.com/questions/25632147/raise-at-the-end-of-a-python-function-outside-try-or-except-block)

Is it worth being explicit and using `raise exc`, or does that behave differently?

minrk: Good catch, this is definitely wrong!

A naked `raise` is only appropriate inside an `except:`  block (`raise` is basically "cancel the `except` block," whereas `raise exc` modifies the traceback). It should be `raise exc` here.

manics:
Looks like the tests need updating with the new text https://github.com/jupyterhub/jupyterhub/blob/b2c5ad40c5aacbfd2cb040972e40dad0acc37209/jupyterhub/tests/test_pages.py#L1131-L1143

manics (approved):

PR #3717, opened 2021-12-14
fix Spawner.oauth_roles config

minrk:
missing cast to orm.Role from config when populating oauth client

encountered while testing #3713 

test included

### DIFF
--- a/jupyterhub/user.py
+++ b/jupyterhub/user.py
@@ -622,6 +622,19 @@ class User:
             if callable(allowed_roles):
                 allowed_roles = allowed_roles(spawner)
 
+            # allowed_roles config is a list of strings
+            # oauth provider.allowed_roles is a list of orm.Roles
+            if allowed_roles:
+                allowed_role_names = allowed_roles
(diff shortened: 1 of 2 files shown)


jupyterhub/user.py
```diff
@@ -622,6 +622,19 @@ async def spawn(self, server_name='', options=None, handler=None):
             if callable(allowed_roles):
                 allowed_roles = allowed_roles(spawner)
 
+            # allowed_roles config is a list of strings
+            # oauth provider.allowed_roles is a list or orm.Roles
```
manics, line 626:
```suggestion
            # oauth provider.allowed_roles is a list of orm.Roles
```

manics (commented):
One typo but otherwise sounds fine to me

jupyterhub/user.py
minrk: Thanks, fixed! Test should pass after rebasing after merge of #3719

manics (approved):

PR #3720, opened 2021-12-16
simplify default role assignment

minrk:
Related to #3714 

- always assign 'user' role, not just when no other roles are assigned
- 'admin' role is in addition to `user`, not instead of

Permissions are unchanged, because `admin` is always a superset, but this simplifies things. 

The main difference: when this is called, it always grants the `user` role, not just if there are no roles already present.

### DIFF
--- a/jupyterhub/roles.py
+++ b/jupyterhub/roles.py
@@ -381,24 +381,6 @@ def strip_role(db, entity, rolename):
         )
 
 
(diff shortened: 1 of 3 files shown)


minrk:
For taxonomy purposes, calling this a 'bug' because it fixes a problem where `assign_default_roles` would not assign the `user` role for a new user if they were assigned another role first.

PR #3726, opened 2021-12-20
update service-whoami example

minrk:
- update models with 2.0.0
- different scopes for oauth, api shows model depends on permissions
- update text with more details about scopes

inconsistency reported on gitter by @rcthomas

### DIFF
--- a/examples/service-whoami/README.md
+++ b/examples/service-whoami/README.md
@@ -8,58 +8,71 @@ There is an implementation each of api-token-based `HubAuthenticated` and OAuth-
 
 1.  Launch JupyterHub and the `whoami` services with
 
-        jupyterhub --ip=127.0.0.1
+        jupyterhub
 
 2.  Visit http://127.0.0.1:8000/services/whoami-oauth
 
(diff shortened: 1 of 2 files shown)


rcthomas:
Thanks for the updated and extended docs here @minrk!

examples/service-whoami/README.md
```diff
@@ -8,7 +8,7 @@ There is an implementation each of api-token-based `HubAuthenticated` and OAuth-
 
 1.  Launch JupyterHub and the `whoami` services with
 
-        jupyterhub --ip=127.0.0.1
+        jupyterhub
 
 2.  Visit http://127.0.0.1:8000/services/whoami-oauth
```
manics, line 14: The line below mentions `local-system credentials` but `jupyterhub_config.py` is using the dummy authenticator.

minrk: Good catch, fixed.

consideRatio (approved):
LGTM! Test failures are unrelated!

PR #3728, opened 2021-12-20
Changelog for 2.0.1

minrk:
I'd like to get 2.0.1 out now, because we've fixed various bugs that could cause frustration in the default user role and admin page.

I'd also like to include:

- #3722 (needs review, the only default user role fix so far that's _not_ fixing a clearly incorrect behavior, but instead making a decision about what _should_ happen)
- #3723 (approved, just waiting for a tiny fix)
-  #3731 (2.0.0 broke the nice error response when trying to talk to stopped servers via API)

### DIFF
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -6,7 +6,56 @@ command line for details.
 
 ## [Unreleased]
 
-## [2.0.0]
+## 2.0
+
+### [2.0.1]
+
+([full changelog](https://github.com/jupyterhub/jupyterhub/compare/2.0.0...2.0.1))
+
+2.0.1 is a bugfix release, with some additional small improvements,
+especially in the new RBAC handling and admin page.
+
+Several issues are fixed where users might not have the
+default 'user' role as expected.
+
+#### Enhancements made
+
+- Use URL from authenticator on default login form [#3723](https://github.com/jupyterhub/jupyterhub/pull/3723) ([@sgaist](https://github.com/sgaist))
+- always assign default roles on login [#3722](https://github.com/jupyterhub/jupyterhub/pull/3722) ([@minrk](https://github.com/minrk))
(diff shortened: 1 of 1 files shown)


consideRatio:
Do you think its right to release 2.0.1 instead of 2.1.0 here? I'm inclined to think this should be 2.1.0 as it matches semver2 specification given that we have enhancements in this PR alongside the bugfixes.

I'm thinking it's not important that we release this as 2.0.1 just because we just released 2.0.0, or that we backport changes to 2.0.1 if we release 2.1.0.

I realize I bring an assumption that we are looking to to follow SemVer2 because it and various calendar versioning schemes is all I really know.

minrk:
Yes, I would still call it 2.0.1. Enhancements aren't new features, so don't necessarily prompt minor bumps, even in strict semver. The technically new things are so tiny (only #3195 is new, a single low-level config option, the rest is refinement/fixes of existing behavior), it feels like a more accurate communication of the change.

consideRatio:
Resolving https://github.com/jupyterhub/jupyterhub/issues/3734 is probably a blocker.

minrk:
@consideRatio updated, thanks for reviewing! I don't have anything else I intend for 2.0.1

docs/source/changelog.md
```diff
@@ -6,7 +6,60 @@ command line for details.
...
+
+([full changelog](https://github.com/jupyterhub/jupyterhub/compare/2.0.0...2.0.1))
+
+2.0.1 is a bugfix release, with some additional small improvements,
+especially in the new RBAC handling and admin page.
+
+Several issues are fixed where users might not have the
+default 'user' role as expected.
+
+### 2.0.0...master@{2021-12-22}
```
consideRatio, line 21: Delete this line and this LGTM for merge and release!

minrk: Ooops, copypasta. Thanks for your vigilance!

minrk:
Pushed 2.0.1!

PR #3733, opened 2021-12-21
Fix missing f-string modifier

manics:
Noticed whilst trying to debug an admin page problem.


consideRatio:
@manics there are test failures which I'm confident are unrelated. They can be observed by running the tests on the main branch. Due to that, I'll go for a merge even though tests are failing.

See https://github.com/jupyterhub/jupyterhub/actions/runs/1609109514 for a test run on the main branch.

consideRatio (approved):

PR #3748, opened 2022-01-03
DOC: Add note about allowed_users not being set

choldgraf:
This adds a warning to the `allowed_users` section to clarify that any user will be able to log on if this configuration isn't set.

Correct me if this is wrong, but we ran into an issue with this on a 2i2c hub and I wanted to make sure we didn't forget to note this in the docs if it is indeed true.

### DIFF
--- a/docs/source/getting-started/authenticators-users-basics.md
+++ b/docs/source/getting-started/authenticators-users-basics.md
@@ -16,6 +16,10 @@ c.Authenticator.allowed_users = {'mal', 'zoe', 'inara', 'kaylee'}
 Users in the `allowed_users` set are added to the Hub database when the Hub is
 started.
 
+```{warning}
+If this configuration value is not set, then **all users will be allowed into your hub**.
+```
+
 ## Configure admins (`admin_users`)
 
 ```{note}
--- a/docs/source/getting-started/authenticators-users-basics.md
+++ b/docs/source/getting-started/authenticators-users-basics.md
@@ -17,7 +17,7 @@ Users in the `allowed_users` set are added to the Hub database when the Hub is
 started.
 
 ```{warning}
-If this configuration value is not set, then **all users will be allowed into your hub**.
+If this configuration value is not set, then **all authenticated users will be allowed into your hub**.
 ```
 
 ## Configure admins (`admin_users`)


consideRatio:
It seems that is correct, and there is a info note about it being logged also. To conclude, admin_users ensures users are granted admin rights, but without explicit allowed_users or other method of accepting/rejecting users in place all users will be allowed.

To have a note in the docs reflecting what the info message sais makes perfect sense in my mind (`"Not using allowed_users. Any authenticated user will be allowed."`).

https://github.com/jupyterhub/jupyterhub/blob/46e7a231fe3ae2a6081fb7fcc6282d1cdf1cbad7/jupyterhub/app.py#L1901-L1926

docs/source/getting-started/authenticators-users-basics.md
```diff
@@ -16,6 +16,10 @@ c.Authenticator.allowed_users = {'mal', 'zoe', 'inara', 'kaylee'}
 Users in the `allowed_users` set are added to the Hub database when the Hub is
 started.
 
+```{warning}
+If this configuration value is not set, then **all users will be allowed into your hub**.
```
consideRatio, line 20:
```suggestion
If this configuration value is not set, then **all authenticated users will be allowed into your hub**.
```

choldgraf:
What are people's opinions about changing this default behavior?

I think it is rare that a user would *want* to allow anybody to access their hub, and much more common that they'd want to restrict access to a subset of users. For that reason, it seems a more reasonable default to be more conservative here, and assume that if `allowed_users` is not set, then it should not allow *any* user rather than to allow *all* users.

mathematicalmichael:
@choldgraf commenting here as I don't know where that discussion is being picked up..

I think it's extremely rare that someone wants a wide-open hub... I for one have run across this particular issue after trying to debug why my `userlist` / `allowed_users` is not being honored and it was letting in anyone with a github account. This took place during a fresh installation as I was manually testing the deployment. Here's an extra caveat to consider as well regarding the two "types" of users:

I have a _single_ user in that file who happens _also_ to belong to the `admin` "group" in the `userlist` file (therefore it is assigned to `c.Authenticator.admin_user`) and I suspect that since it is getting placed there instead of in `c.Authenticator.allowed_users`, which is left empty as a result, my hub is wide-open as a result...

This was really hard to track down to being the source of my issue, and very unexpected behavior (wide open access).

I had to add an org I control as a dummy "user" to `userlist` so that I could finally (purposefully) block out the second github account I was using to test that authentication was being respected.

consider this my +1 to "block everyone by default", but check `admin_users` as well please.

PR #3750, opened 2022-01-03
ci: refactor to avoid triggering all tests on changes to docs

consideRatio:
I noted that all our documentation PRs triggered the full test suite, and since we had intermittent test failures it caused some trouble.

As I needed to learn a bit via code inspection to do this, I also updated a few docstrings in the process.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,14 +1,28 @@
 # This is a GitHub workflow defining a set of jobs with a set of steps.
-# ref: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions
+# ref: https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions
 #
 name: Test
 
-# Trigger the workflow's on all PRs but only on pushed tags or commits to
-# main/master branch to avoid PRs developed in a GitHub fork's dedicated branch
-# to trigger.
 on:
   pull_request:
+    paths-ignore:
+      - "docs/**"
+      - "**.md"
+      - "**.rst"
+      - ".github/workflows/*"
+      - "!.github/workflows/test.yml"
   push:
+    paths-ignore:
+      - "docs/**"
+      - "**.md"
+      - "**.rst"
+      - ".github/workflows/*"
+      - "!.github/workflows/test.yml"
+    branches-ignore:
+      - "dependabot/**"
+      - "pre-commit-ci-update-config"
+    tags:
+      - "**"
   workflow_dispatch:
 
 env:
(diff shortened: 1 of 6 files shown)


docs/test_docs.py
```diff
@@ -25,13 +27,11 @@ def test_rest_api_version():
...
+def test_rest_api_rbac_scope_descriptions_are_updated():
+    """Checks that the RBAC scope descriptions in JupyterHub's REST API
+    definition file (rest-api.yml) as can be updated by generate-scope-table.py
+    matches what is committed."""
     run([sys.executable, "source/rbac/generate-scope-table.py"], cwd=here, check=True)
-    run(
-        ['pre-commit', 'run', 'prettier', '--files', 'source/_static/rest-api.yml'],
-        cwd=here,
-        check=False,
-    )
```
consideRatio, line 34: This is version controlled and we have pre-commit run in other tests so no need to do it here as well.

.github/workflows/test.yml
```diff
@@ -17,25 +31,6 @@ env:
   PYTEST_ADDOPTS: "--verbose --color=yes"
 
 jobs:
-  rest-api:
```
consideRatio, line 20: Only by extracting this job to a dedicated workflow file could we avoid triggering the workflow with all tests on docs changes.

docs/test_docs.py
minrk: This pre-commit step is required for the test to pass. `generate-scope-table` above will generate a file with correct contents, but _not_ correct formatting.

The failure is being hidden because there's a bug a few lines below, where `--no-pager` must be passed _before_ `diff`:

```python
"git", "--no-pager", "diff", "--exit-code"...
```

(don't ask me why passing invalid options doesn't set an error code!)

minrk:
👍 in general, one note that a necessary step is removed from a test, and also revealing that the test isn't actually being run due to invalid arguments!

docs/test_docs.py
consideRatio: Ah hmmmm, since we have a dedicated pre-commit test, I'd argue that we let this test not run pre-commit also. What do you think?

- This test would check that the rest-api is updated
- The pre-commit test would check that all (updated) files are pre-commit verified

consideRatio: Oh nevermind I get it now! I kind of got it, then confused myself to think we didn't need pre-commit again, and now I'm back to thinking we need it.

Then: do we want it as part of the generate script or as part of the test?

minrk: We can also update the test to do a diff of the parsed content. Then formatting shouldn't matter. However, then _presentation_ of the diff is quite a bit harder. That shouldn't matter too much, as it's an auto-generated file.

consideRatio: I ended up making the pre-commit run as part of running the generate-scope-table.py script, I had to add pre-commit as a docs/requirements.txt for this. Acceptable?

minrk: Sure, I think that's fine. That'll prevent the pre-commit error on commit after run, too. Main disadvantage I see is that it might slow down iterating on the docs, since running pre-commit becomes part of the rebuilding the docs cycle. But maybe it won't run if the file hasn't changed.

We can switch to the structural comparison in the test if it turns out to be a problem.

minrk (approved):

Issue #3737, opened 2021-12-25
Empty Admin page when the hub is behind nginx proxy


ykazakov:
### Bug description

Merry Christmas!

The Admin settings of jupyterhub v. 2.0.1 are not shown if the hub is behind a reverse nginx proxy.
The symptoms appear to be similar to #3711 except that they are now not caused by `base_url`.

#### Expected behaviour

Admin page when the hub is accessed directly:

<img width="684" alt="Screenshot 2021-12-25 at 09 32 58" src="https://user-images.githubusercontent.com/2140361/147381008-67c6e49d-a583-47d0-aab2-bc1246976b1e.png">

#### Actual behaviour

Admin page when the hub is accessed through a proxy:

<img width="683" alt="Screenshot 2021-12-25 at 09 34 48" src="https://user-images.githubusercontent.com/2140361/147381033-be241b38-db3f-410c-a439-2d4eb3f71f33.png">

This message looks suspicious (see the full log below):
```
jupyterhub_1  | [W 2021-12-25 08:58:25.156 JupyterHub base:89] Blocking Cross Origin API request.  Referer: http://127.0.0.1:8080/hub/admin, Host: 127.0.0.1, Host URL: http://127.0.0.1/hub/
```

There is no problem with `jupyterhub/jupyterhub:1.5.0`

### How to reproduce

1. Ensure the following directory structure:
```
.
├── docker-compose.yml
├── jupyterhub
│   ├── Dockerfile
│   └── jupyterhub_config.py
└── nginx
    ├── Dockerfile
    └── nginx.conf
```
2. Ensure the following file contents:
  - `docker-compose.yml`:
```
version: '3'
services:
  nginx: 
    image: nginx
    build: nginx
    ports:
      - 8080:80
      
  jupyterhub:
    image: jupyterhub
    build: jupyterhub
    expose:
      - "8000"
```
  - `jupyterhub/Dockerfile`:
```
FROM jupyterhub/jupyterhub:2.0.1

COPY jupyterhub_config.py /etc/jupyterhub/

CMD jupyterhub -f /etc/jupyterhub/jupyterhub_config.py --debug
```
- `jupyterhub/jupyterhub_config.py`:
```
c.Authenticator.admin_users = {'user'}
c.JupyterHub.authenticator_class = 'dummy'
```
- `nginx/Dockerfile`:
```
FROM nginx:1.21.4

COPY nginx.conf /etc/nginx/
```
[...]

PR #3759, opened 2022-01-10
changelog for 2.0.2

minrk:
I'd like to get this out promptly due to the regression in 2.0.1 (#3737) for which there is not an easy workaround.


consideRatio (approved):
LGTM! Go for it!

minrk:
2.0.2 publishing...

PR #3762, opened 2022-01-11
Add the capability to inform a connection to Alembic Migration Script

DougTrajano:
Hi guys,

In this PR, I'm adding the capability to pass a connection in the config variable used by Alembic Migration Scripts.

The way that I did it uses the same approach that we can see in [Cookbook — Alembic 1.7.5 documentation](https://alembic.sqlalchemy.org/en/latest/cookbook.html#sharing-a-connection-with-a-series-of-migration-commands-and-environments)

This approach allows us to reuse an established connection in SQLAlchemy. Very useful in my use case that I replaced the default approach that uses the URL by IAM Role connection.

### DIFF
--- a/jupyterhub/alembic/env.py
+++ b/jupyterhub/alembic/env.py
@@ -55,9 +55,14 @@ def run_migrations_offline():
     script output.
 
     """
-    url = config.get_main_option("sqlalchemy.url")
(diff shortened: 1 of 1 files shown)


minrk:
Thanks!

PR #3763, opened 2022-01-12
apply scope checks to some admin-or-self situations

minrk:
Some non-api spawn and redirect checks still had `self or admin`, when they should have checked directly for the appropriate permissions. Explicit permissions checks added for each. The most direct bug here is that non-admin users with permission to launch servers of other users could only do so via the API, but not via the UI. This is now fixed (not a security issue because it was only users who should have permissions being denied, not users without permissions being allowed).

This removes the long-deprecated redirect from `/user/other` -> `/user/self` with the elaborate condition: `if the user is different and the other user's server is not running _and_ the user is not admin`. The result is a more consistent behavior whether the requested server is running or not, and whether the user has _access_ to the running server or not. This behavior has been problematic for a very long time and deprecated since the addition of `user-redirect` in 0.7, but it really makes no sense now that we have access scopes.

The new behavior is much simpler: attempt to access a not-running server is allowed if you have access to that server, and it is not if you do not. There is no longer an implicit redirect back to your own server, which was always just a guess, and has only ever worked when the requested server was not running.

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -45,6 +45,7 @@ from ..metrics import ServerSpawnStatus
 from ..metrics import ServerStopStatus
 from ..metrics import TOTAL_USERS
 from ..objects import Server
+from ..scopes import needs_scope
 from ..spawner import LocalProcessSpawner
 from ..user import User
 from ..utils import AnyTimeoutError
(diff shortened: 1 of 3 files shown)


jupyterhub/handlers/base.py
```diff
@@ -1448,24 +1449,18 @@ def non_get(self, user_name, user_path):
...
+        if user_name != current_user.name:
             user = self.find_user(user_name)
             if user is None:
                 # no such user
                 raise web.HTTPError(404, "No such user %s" % user_name)
             self.log.info(
-                "Admin %s requesting spawn on behalf of %s",
+                "User {current_user.name} requesting spawn on behalf of {user.name}",
                 current_user.name,
                 user.name,
```
consideRatio, line 1465: This is like a f-string, but isnt. Is it correct as it is / is this functional?

jupyterhub/handlers/base.py
```diff
@@ -1448,24 +1449,18 @@ def non_get(self, user_name, user_path):
...
         current_user = self.current_user
-
-        if (
-            current_user
-            and current_user.name != user_name
-            and current_user.admin
-            and self.settings.get('admin_access', False)
-        ):
-            # allow admins to spawn on behalf of users
+        if user_name != current_user.name:
```
consideRatio, line 1457: A check if current user is truthy is removed, is it safe?

jupyterhub/handlers/pages.py
```diff
@@ -106,22 +106,27 @@ async def _render_form(self, for_user, spawner_options_form, message=''):
...
         """GET renders form for spawning with user-specified options
 
         or triggers spawn via redirect if there is no form.
         """
+        # two-stage to get the right signature for @require_scopes filter on user_name
+        if user_name is None:
+            user_name = self.current_user.name
+        if server_name is None:
+            server_name = ""
+        return self._get(user_name=user_name, server_name=server_name)
```
consideRatio, line 120: Is the process of returning a coroutine here what makrs you remove the async keyword from this get handler - in practice "making it async" anyhow by returning a coroutine?

jupyterhub/handlers/base.py
minrk: Fixed!

jupyterhub/handlers/base.py
minrk: Yes, the `@web.authenticated` decorator ensures current_user is truthy, or the method won't be called.

jupyterhub/handlers/pages.py
minrk: Yes, a coroutine object is an object like anything else that can be returned from any function, async or not. I think this was clearer in the `@coroutine` decorator days, where it was more apparent that it really must be a regular Python function returning a regular Python object. The `async def` syntax can feel like something special is happening, when it's really not. The key, I think: calling a **coroutine function** returns a **coroutine object** which is **awaitable** (other awaitables include Tasks, Futures, any object implementing `__await__`, etc.). Any function can return any kind of object, so if you wrap a function and return its original return object, calling the wrapper will have the same behavior as the original wrapped function.

 So if you 'wrap' a coroutine, the wrapper does not itself need to be async unless it, too, has something to await. I could have made this _logically_ equivalent with `async def` if it ended with `return await self._get(...)`, but that would add an unnecessary callback layer to the event loop.

consideRatio: Thanks for helping me learn about the details Min!

jupyterhub/handlers/base.py
```diff
@@ -1448,24 +1449,18 @@ def non_get(self, user_name, user_path):
...
             user = self.find_user(user_name)
             if user is None:
                 # no such user
                 raise web.HTTPError(404, "No such user %s" % user_name)
             self.log.info(
-                "Admin %s requesting spawn on behalf of %s",
+                f"User {current_user.name} requesting spawn on behalf of {user.name}",
                 current_user.name,
                 user.name,
```
consideRatio, line 1465: You still pass log the objects part of the f-string!

minrk: Another good catch; fixed!

consideRatio (approved):

PR #3764, opened 2022-01-12
relay custom messages in exception.jupyterhub_message in progress API

### LINKED ISSUE #3761: Better handling of spawner failures
### Proposed change

Currently, when a Spawner fails to launch the Jupyter server, there is no good way in which it can inform the user of the cause.

I propose we define a "SpawnException" class that the spawner can throw, and which enables the spawner to pass an html_message that is used to display a nicely formatted exception (include some formatting, links to status pages of underlying systems, ...)

Advanced JupyterHub deployments can provide better error messages to inform the user about why their request failed: have they requested an invalid setup? Have they run out of quota? Is there a maintenance going on in the underlying system?

### Suggest a solution

Spawner class definition:

```
class SpawnException(Exception):
    def __init__(self, *args, html_message=None, **kwargs):
        super(SpawnException, self).__init__(*args, **kwargs)

        self.html_message = html_message
```

Add handling of this exception class in [SpawnProgressApiHandler](https://github.com/jupyterhub/jupyterhub/blob/e540d143bb8d219aca6023f104e3dbe76ca12bb8/jupyterhub/apihandlers/users.py#L716)

```
            if f and f.done() and f.exception():
                if isinstance(f.exception(), SpawnException) and f.exception().html_message is not None:
                   failed_event['html_message'] = f.exception().html_message
                failed_event['message'] = "Spawn failed: %s" % f.exception()
                await self.send_event(failed_event)

```

Log spawn failure as a warning/info instead of as an error in [user.py](https://github.com/jupyterhub/jupyterhub/blob/e540d143bb8d219aca6023f104e3dbe76ca12bb8/jupyterhub/user.py#L748), improve graceful handling of the SpawnException elsewhere.

minrk:
matches the message shown on the HTML spawn-failed page.

For consistency, also support `jupyterhub_html_message` to populate the `html_message` field. Both places now support both messages.

closes #3761

### DIFF
--- a/share/jupyterhub/templates/not_running.html
+++ b/share/jupyterhub/templates/not_running.html
@@ -18,8 +18,10 @@
       <p>
         {% if failed %}
         The latest attempt to start your server {{ server_name }} has failed.
-        {% if failed_message %}
-          {{ failed_message }}
+        {% if failed_html_message %}
+          </p><p>{{ failed_html_message | safe }}</p><p>
+        {% elif failed_message %}
+          </p><p>{{ failed_message }}</p><p>
         {% endif %}
         Would you like to retry starting it?
         {% else %}
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -714,7 +714,12 @@ class SpawnProgressAPIHandler(APIHandler):
             # check if spawner has just failed
             f = spawn_future
             if f and f.done() and f.exception():
-                failed_event['message'] = "Spawn failed: %s" % f.exception()
+                exc = f.exception()
+                message = getattr(exc, "jupyterhub_message", str(exc))
+                failed_event['message'] = f"Spawn failed: {message}"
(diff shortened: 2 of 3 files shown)


share/jupyterhub/templates/not_running.html
```diff
@@ -18,8 +18,10 @@ <h1>
       <p>
         {% if failed %}
         The latest attempt to start your server {{ server_name }} has failed.
-        {% if failed_message %}
-          {{ failed_message }}
+        {% if failed_html_message %}
+          <p>{{ failed_html_message | safe }}</p>
+        {% elif failed_message %}
+          <p>{{ failed_message }}</p>
```
consideRatio, line 24: I got thinking a bit about having a p tag within a p tag, but seems okay. Seems a bit off having parts of this section in dedicated p tags while others aren't etc but I don't really know the consequences.

If this feels OK then this LGTM.

minrk: Good catch, fixed by closing the previous p before opening the new one.

PR #3773, opened 2022-01-19
Using orm_spawner in server model user_options

### LINKED ISSUE #3772: user_options returning empty at users rest api after jupyterhub restart
### Bug description
I’ve noticed that after jupyterhub is restarted the user_options are empty i.e. {}
when submitting /hub/api/users request
(before restart I’m getting user_options which are persisted in db due implemented in persist user_options (https://github.com/jupyterhub/jupyterhub/pull/2446)

from looking at code it might be connected to the fact that in apihandler’s [server_model takes user_options from spawner and not spawner.orm_spawner] (https://github.com/jupyterhub/jupyterhub/blob/f5bb0a2622be51b02a64eaa23b10b5f66461ba2b/jupyterhub/apihandlers/base.py#L204)

I've patched as suggested above our private jupyterhub, then tested manually that it indeed solves the problem

#### Expected behaviour
my expectation that user_options will be properly returned after jupyter hub restart

#### Actual behaviour
user_options are empty {}

### How to reproduce
1. create pod with user_options
2. make sure it's up
3. access jupyterhub api of users
4. make sure user_options are returned as part of response with custom values inside
4. restart jupyter hub
4. access jupyterhub api of users
5. make sure user_options are returned but empty {}

### Your personal set up
zero-to-jupyterhub

IgorBerman:
closes #3772

### DIFF
--- a/jupyterhub/apihandlers/base.py
+++ b/jupyterhub/apihandlers/base.py
@@ -201,7 +201,7 @@ class APIHandler(BaseHandler):
             'pending': spawner.pending,
             'ready': spawner.ready,
             'url': url_path_join(spawner.user.url, spawner.name, '/'),
-            'user_options': spawner.user_options,
+            'user_options': spawner.orm_spawner.user_options,
             'progress_url': spawner._progress_url,
         }
         scope_filter = self.get_scope_filter('admin:server_state')
--- a/jupyterhub/apihandlers/base.py
+++ b/jupyterhub/apihandlers/base.py
@@ -201,7 +201,7 @@ class APIHandler(BaseHandler):
             'pending': spawner.pending,
             'ready': spawner.ready,
             'url': url_path_join(spawner.user.url, spawner.name, '/'),
-            'user_options': spawner.orm_spawner.user_options,
+            'user_options': spawner.user_options,
             'progress_url': spawner._progress_url,
         }
         scope_filter = self.get_scope_filter('admin:server_state')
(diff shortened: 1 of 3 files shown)


minrk:
Thanks for the PR! Based on your description in #3772, I think the bug is in a slightly different place, though, because at the time of calling `server_model` here, the two attributes ought to be in sync. I suspect there's a missing `spawner.user_options = orm_spawner.user_options` in `User._new_spawner` where we recreate a Spawner object from the db. That would cause `spawner.user_options` to revert to empty on restart.

IgorBerman:
thanks @minrk I'll try to change it as per your suggestion
will update later

IgorBerman:
Hi  @minrk 
I think I've found right place to put initialization of spawner from orm_spawner's user_options (PR updated)
my next question would be : where is the right place to implement test for this fix
found test_user.py, so is it the one?

minrk:
I think you can add a little test in `test_spawner.py` that quickly creates a spawner, deletes it, then tries to recreate it from the db:

```python
def test_spawner_from_db(app, user):
    spawner = user.spawners['name']
    user_options =  {"test": "value"}
    spawner.orm_spawner.user_options = user_options
    app.db.commit()
    # delete and recreate the spawner from the db
    user.spawners.pop('name')
    new_spawner = user.spawners['name']
    assert new_spawner.orm_spawner.user_options == user_options
    assert new_spawner.user_options == user_options
```

IgorBerman:
@minrk thanks for support and advice!

PR #3775, opened 2022-01-20
DOCS: Add github metadata for edit button

minrk:
after #3754 local builds don't work when this is True due to missing GitHub metadata. So make it true only on RTD.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -147,6 +147,13 @@ html_theme_options = {
     "navbar_align": "left",
 }
 
+html_context = {
+    "github_user": "jupyterhub",
+    "github_repo": "jupyterhub",
+    "github_version": "main",
+    "doc_path": "docs",
+}
+
 # -- Options for LaTeX output ---------------------------------------------
 
 latex_elements = {


choldgraf:
I think the other thing we could do is include our own GitHub metadata in `html_context`: https://pydata-sphinx-theme.readthedocs.io/en/latest/user_guide/configuring.html#github

We'd basically need to add:

```python
html_context = {
    "github_user": "jupyterhub",
    "github_repo": "jupyterhub",
    "github_version": "main",
    "doc_path": "docs",
}
```
that might be a bit less complex than special-casing readthedocs. what do you think?

minrk:
Better idea! That's actually what I tried to do first, but gave up because exactly what the config should be wasn't obvious after 30 seconds of looking :)

consideRatio (approved):
This LGTM, but the title of the PR should be updated. Since the git history is erased I'm not sure what makes sense. Go for a title update and self-merge!

choldgraf (approved):
Updated the title, LGTM

consideRatio:
Thanks @choldgraf that looks good to me, I went for a merge!

PR #3770, opened 2022-01-18
Add `read:metrics` scope for metrics endpoint

minrk:
and ensure token auth is accepted (missed this handler in #3686).

closes #3769 

I'm reticent to release this PR as 2.0.3 because it defines a new scope. However, authenticated prometheus simply doesn't work right now (#3769), so one could argue that the 2.0.x bug is that the _scope_ for authenticated metrics was missing.

we could also backport just the one-line `_accept_token_auth = True` to 2.0.x, to allow any authenticated token to access metrics, as was the pre-2.0 behavior, but then 2.1 will be adding the new scope. Arguably doing so would break _more_ things, because there would be a version that actually works without the scope.

PR #3776, opened 2022-01-21
Changelog for 2.1.0

minrk:
After a few more PRs, I think we're more solidly in 2.1 territory than 2.0.3. Due to #3770, I still think we should do a prompt release.


consideRatio (approved):
Wieee LGTM!

minrk:
release building...

PR #3778, opened 2022-01-24
ensure admins get all scopes

minrk:
adds missing `read:metrics` scope, and ensures we don't introduce similar bugs in the future by ensuring admins get all scopes, whatever they are, rather than needing a duplicate record.

I'll do a quick 2.1.1 with this fixed.

Reported on gitter by @katringoogoo


consideRatio (approved):

consideRatio (commented):
The changes lgtm, but there seems to be a relevant test failure about.

minrk:
@consideRatio switched to minimal patch. Will take another run at the more general fix after 2.1.1

consideRatio (approved):

PR #3781, opened 2022-01-27
Log proxy's public_url only when started by JupyterHub

cqzlxl:
When we run the proxy separately,  defaults of `hub.bind_url` may be different from proxy's public url. Actually, the hub has no ways to know about which address the proxy is serving at if we do not configure its `bind_url` explicitly.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -3150,7 +3150,8 @@ class JupyterHub(Application):
             self.last_activity_callback = pc
             pc.start()
 
-        self.log.info("JupyterHub is now running at %s", self.proxy.public_url)
+        if self.proxy.should_start:
+            self.log.info("JupyterHub is now running at %s", self.proxy.public_url)
         # Use atexit for Windows, it doesn't have signal handling support
         if _mswindows:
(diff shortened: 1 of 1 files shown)


jupyterhub/app.py
```diff
@@ -3150,7 +3150,8 @@ async def start(self):
             self.last_activity_callback = pc
             pc.start()
 
-        self.log.info("JupyterHub is now running at %s", self.proxy.public_url)
+        if self.proxy.should_start:
+            self.log.info("JupyterHub is now running at %s", self.proxy.public_url)
```
minrk, line 3154: I agree with @consideRatio that it's nice to still have the "JupyterHub is now running" marker to indicate that startup is done. How about:

```suggestion
            self.log.info("JupyterHub is now running at %s", self.proxy.public_url)
        else:
            self.log.info("JupyterHub is now running, internal Hub API at %s", self.hub.url)
```

or split "JupyterHub is now running" into its own log statement, and log the URL separately under the condition.

cqzlxl: Your solution looks better

PR #3785, opened 2022-02-02
Automatic group synchronization with groups obtained through authenticator

vladfreeze:
This new function added to base.py reads the details of the authenticator and checks if the authentication contains groups, which can then be automatically created in the JupyterHub database with the user as a member.
The group check happens on every authentication and offers a user membership synchronization. If a group is no longer showing up in the user's authentication, the user will be removed from the said group. As long as the groups remain in the user's authentication object, the user will still be part of these groups on JupyterHub.


websi96:
#3548

manics:
Closing based on the [last comment](https://github.com/jupyterhub/jupyterhub/pull/3785#issuecomment-1028449403)

Issue #2973, opened 2020-03-06
PAM sessions are not being initialized properly


gclen:
I have a configuration where mount points are being set up when a user authenticates with PAM. This worked on jupyterhub 0.8.1. However, this does not appear to work on newer versions of jupyterhub (I upgraded to 0.9.6). 

I think the issue can be traced back to [this commit](https://github.com/jupyterhub/jupyterhub/pull/1691/commits/4de6b39788aa1b7597a36e7c9e3b51d498b50900) where the interaction with PAM was moved to a thread. I think having authentication in a separate thread is _probably_ okay (though I haven't tested this). However, opening a PAM session should probably be done in the spawned process since they are typically setting process/thread attributes. Reverting the changes in jupyterhub/auth.py fixed the issue for me but I think a proper fix is needed (this fix may just be reverting the changes).

I think taking another look at the commit above may also help with issues #2056, #2754, and #2406. Let me know if you need me to do more testing or need any other information.

gatoniel:
Hi,

I am trying to use PAMAuthenticator with a non-root user JuypterHub controlled by systemd on CentOS 8. Since I am going to spawn singleuser servers with podman, I need PAM to open a session to set the correct context for each user. Right now this does not work, and I get several PAM related errors. This is the /etc/var/secure log

> Mar 31 16:04:51 picard python3[9768]: pam_loginuid(login:session): Error writing /proc/self/loginuid: Operation not permitted
> Mar 31 16:04:51 picard python3[9768]: pam_loginuid(login:session): set_loginuid failed
> Mar 31 16:04:51 picard python3[9768]: pam_selinux(login:session): Error sending audit message.
> Mar 31 16:04:51 picard python3[9768]: pam_systemd(login:session): Failed to create session: Access denied
> Mar 31 16:04:51 picard python3[9768]: pam_unix(login:session): session opened for user testuser by (uid=991)

And this is the service file for JupyterHub for systemd:

```
[Unit]
Description=JupyterHub
After=syslog.target network.target podman-chp.service

[Service]
User=jupyterhub
AmbientCapabilities=CAP_SETUID CAP_SETGID
Environment="PATH=/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/opt/jupyterhub/bin"
Environment="CONFIGPROXY_AUTH_TOKEN=754b5de95588a1c729d417809d3925cca09e61fc0255015d93b54d154ec6885b"
[...]

PR #3787, opened 2022-02-07
Stop opening PAM sessions by default

minrk:
We don't do it correctly, so don't try by default

It does work _sometimes_, but most of the times that it does work, it's because it's a no-op. Turning it off by default makes it more likely folks will see the caveat that it may not work.

xref #2973

### DIFF
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -975,16 +975,24 @@ class PAMAuthenticator(LocalAuthenticator):
     ).tag(config=True)
 
     open_sessions = Bool(
-        True,
+        False,
         help="""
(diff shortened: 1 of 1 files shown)


consideRatio (approved):
The CI failure seems unrelated, if agreed, this LGTM.

We should have an explicit entry about this in the changelog though. I listed it as "maintenance" but maybe create a new label called breaking or changed etc to track this as well?

minrk:
Added api-change label to [match github-activity](https://github.com/executablebooks/github-activity/blob/f994a6944fc50b167a7799b464af425f794ba1c8/github_activity/github_activity.py#L48)

Issue #3789, opened 2022-02-09
Unable to configure Spawner.options_from_form, "missing 1 required positional argument"


rcthomas:
#### Expected behaviour

Back in #3225 `Spawner.options_from_form` was made configurable, so I expected to be able to do something like this in a config file:

```
async def options_from_form(spawner, form_data):
    ...
    return user_options

c.Spawner.options_from_form = options_from_form
```

#### Actual behaviour

When processing the form though, you get the following error:

```
[E 2022-02-09 02:28:34.484 JupyterHub pages:264] Failed to spawn single-user server with form
    Traceback (most recent call last):
      File "/usr/local/lib/python3.8/dist-packages/jupyterhub/handlers/pages.py", line 258, in _post
        options = await maybe_future(spawner.options_from_form(form_options))
    TypeError: options_from_form() missing 1 required positional argument: 'form_data'
```

It looks like I'm really supposed to declare `async def options_from_form(form_data)` in my configuration and not be able to pass the spawner?  I realize that the function isn't supposed to have side effects and maybe (?) the idea behind not letting a user pass `spawner` along was to discourage side-effects, but it's somewhat inconvenient.

Note that if you create a new Spawner class and customize `options_from_form` you get access to the spawner/self.  That discrepancy makes me think this is actually a bug.

### How to reproduce

This `jupyterhub_config.py` seems to reproduce the problem:

```
async def options_form(spawner):
    return "hello"

async def options_from_form(spawner, form_data):
    return {}

c.Spawner.options_form = options_form
c.Spawner.options_from_form = options_from_form
```

The reproducer was confirmed using the `jupyterhub/jupyterhub:2` Docker container.

I find that I'm able to get the behavior I expect by creating a `run_options_from_form` wrapper in the Spawner class following the style of the `pre_spawn_hook` etc. but this breaks the situation where the default `options_from_form` is used, or (I think) where `options_from_form` is defined as part of a `Spawner` class, not as a configurable.

rcthomas:
This is kind of "neat" actually.

rcthomas:
This approach works with the `load_config_file` line commented out or not, and is kind of more in line with how the other hooks work:

```
from traitlets import Callable, default
from traitlets.config import Application

def _default_thing(obj, x):
    return x

class A(Application):

    thing = Callable().tag(config=True)

    @default("thing")
    def _thing(self):
        return _default_thing

    def initialize(self, argv=None):
[...]

PR #3791, opened 2022-02-10
Enable options_from_form(spawner, form_data) from configuration file

rcthomas:
In #3789 it was noted that attempting to customize `c.Spawner.options_from_form` would result in a syntax error if the callable signature was like `my_options_from_form(spawner, form_data)`.  Since the default `options_from_form` and subclass `options_from_form` implementations can take the spawner object and the form data as arguments, it appears to be a bug.

The question is how best to handle both cases, and here I document one that seems to work in my case and at least passes existing tests (though a new test with a configured `options_from_form` setting may make sense).  There might be a nicer way to do this than to check whether `options_from_form` is being a function (as opposed to a bound method).  If there are better ideas I'm happy to try them.

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -11,6 +11,7 @@ import shutil
 import signal
 import sys
 import warnings
+from inspect import signature
 from subprocess import Popen
 from tempfile import mkdtemp
 from urllib.parse import urlparse
@@ -423,6 +424,13 @@ class Spawner(LoggingConfigurable):
 
     def _default_options_from_form(self, form_data):
         return form_data
+
+    def run_options_from_form(self, form_data):
+        sig = signature(self.options_from_form)
+        if 'spawner' in sig.parameters:
+            return self.options_from_form(form_data, spawner=self)
+        else:
+            return self.options_from_form(form_data)
 
     def options_from_query(self, query_data):
         """Interpret query arguments passed to /spawn
--- a/jupyterhub/handlers/pages.py
+++ b/jupyterhub/handlers/pages.py
@@ -255,7 +255,7 @@ class SpawnHandler(BaseHandler):
             self.log.debug(
                 "Triggering spawn with supplied form options for %s", spawner._log_name
             )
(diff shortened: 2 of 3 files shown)


minrk:
It's not _exactly_ a bug - the spec is that it should be a function taking only the form data, but it could make use of the spawner if it happens to be a bound method (where `self` is passed implicitly). But it does seem reasonable to _allow_ passing the spawner.

Rather than inspecting the _type_, which would break any existing pure functions `options_from_form` like:

```python
def options_from_form(options_form):
    return {
        'number': int(options_form['number'][0])
    }

c.Spawner.options_from_form = options_from_form
```

I think it makes sense to inspect the signature with `inspect.signature`. If it takes an additional `spawner` keyword argument, pass that along:

```python
sig = inspect.signature(options_from_form)
if 'spawner' in sig.parameters:
    options_from_form(options_form, spawner=self)
else:
    options_from_form(options_form)
```

That ought to be a backward-compatible way to introduce the feature.

rcthomas:
Right, it's not _exactly_ a bug.  I'll rework to use `inspect` here.

jupyterhub/spawner.py
```diff
@@ -424,6 +425,13 @@ def _options_from_form(self):
     def _default_options_from_form(self, form_data):
         return form_data
 
+    def run_options_from_form(self, form_data):
+        sig = signature(self.options_from_form)
+        if 'spawner' in sig.parameters:
+            return self.options_from_form(form_data, spawner=self)
+        else:
+            return self.options_from_form(form_data)
```
minrk, line 433: Perfect! Can you add a test to test_spawner.py that exercises both forms of `options_from_form`? I was a little surprised the previous form of this PR didn't make anything fail, which means we didn't have coverage.

Sample form data is always a dict of lists of strings:

```python
from_data = {"key": ["value"]}
```

rcthomas: I added 2 tests, one for a case with a spawner argument and one without.  Does that look OK?  Happy to improve it if it needs to do more.

minrk:
Perfect, thanks!

PR #3793, opened 2022-02-12
show insecure-login-warning for all authenticators

satra:
the insecure login warning over http only appears when form-based authentication is in use. for example a github authenticator does not show this warning.

there is probably a more generic version of this change outside of the if-else construct.


minrk:
Makes sense, thanks!

PR #3801, opened 2022-02-22
Illustrate create() classmethod idea

athornton:
This is a draft PR meant to illustrate the "initialize async things at instance creation time" approach I am talking about with respect to https://github.com/jupyterhub/kubespawner/pull/563

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -97,6 +97,7 @@ from .utils import (
     print_stacks,
     print_ps_info,
     make_ssl_context,
+    create_instance,
 )
 from .metrics import HUB_STARTUP_DURATION_SECONDS
 from .metrics import INIT_SPAWNERS_DURATION_SECONDS
@@ -1119,7 +1120,8 @@ class JupyterHub(Application):
 
     @default('authenticator')
     def _authenticator_default(self):
-        return self.authenticator_class(parent=self, db=self.db)
+        return create_instance(self.authenticator_class,
+                               parent=self, db=self.db)
 
     implicit_spawn_seconds = Float(
         0,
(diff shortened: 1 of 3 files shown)


consideRatio:
I inspected how spawner objects are initialized, and it was tricky to switch to awaiting an async constructor. The reason is that they are lazily created objects - based on when the property `spawners` or `spawner` of a User object is used.

Each `User` object gets a _SpawnerDIct object initialized to the User objects `spawners` property.

https://github.com/jupyterhub/jupyterhub/blob/a3ea0f0449e9e19feff863b5f5cce4d8a4febf1f/jupyterhub/user.py#L210-L217

Then, this will create the instances when needed.

minrk:
Yeah, Spawners are accessed in a synchronous way in a lot of places, and indeed making _access_ of spawners async will lead to race conditions if we aren't careful (we must guarantee that we can't start another async creation of the same spawner name while awaiting the first one). We'll need to at the very least ensure we initialize things like:

- presence in the `.spawners` dict
- pending/etc. state

_before_ any awaits happen. So it will have to be two-stage somewhere, whether that's JupyterHub putting a placeholder Future in the dict

What might be easier is to officially support two-stage:

```python
obj = Class(**)
if hasattr(obj, 'initialize_async'):
    await obj.initialize_async()
```

Across Jupyter/traitlets we have a the _de facto_ pattern of an `.initialize()` method used to finish initialization of an instance, mostly in Application classes, which is async in JupyterHub. We could follow that pattern.

One way to support async init would be to replace the spawners dict with a dict of _futures_, so all the `user.spawners[name]` would become `await user.spawners[name]`. In this way, the item-access on the dict is actually synchronous, creating the Future (should resolve races), while the Future itself is awaiting the creation. The downside is that this adds a lot of awaits in a lot of places for objects that will usually be ready.

yuvipanda:
Given https://github.com/jupyterhub/kubespawner/pull/563 got merged, I'm going to close this one!

Issue #3780, opened 2022-01-25
Apache ReverseProxy not working with Hub 2.0.1 and later.


bviviano:
I am using Apache as a reverse proxy to frontend JupyterHub to my HPC cluster with wrapspawner and batchspawner.  Everything has worked fine, until I attempted to update the version of Jupyter*.  After some testing, it seems that between the 2.0.0 and 2.0.1 release of Hub is where I have problems and start getting messages like

```
[W 2022-01-25 11:26:32.887 JupyterHub base:90] Blocking Cross Origin API request.  Referer: https://luna1.mysite.com/jupyter/hub/spawn-pending/bdk, Host: luna1.mysite.com, Host URL: http://luna1.mysite.com/jupyter/hub/
[W 2022-01-25 11:26:32.887 JupyterHub scopes:498] Not authorizing access to /jupyter/hub/api/users/bdk/server/progress. Requires any of [read:servers], not derived from scopes []
[W 2022-01-25 11:26:32.887 JupyterHub web:1787] 403 GET /jupyter/hub/api/users/bdk/server/progress (::ffff:127.0.0.1): Action is not authorized with current scopes; requires any of [read:servers]
[W 2022-01-25 11:26:32.888 JupyterHub log:189] 403 GET /jupyter/hub/api/users/bdk/server/progress (@::ffff:127.0.0.1) 2.94ms

```
I'm using the following configuration file in Apache
```
<VirtualHost _default_:443>
        ServerName luna1.mysite.com
        ServerSignature Off
        SSLEngine on
        SSLProxyEngine on
        SSLCertificateFile /etc/pki/tls/certs/server.crt
        SSLCertificateKeyFile /etc/pki/tls/private/server.key
        SSLCertificateChainFile /etc/pki/tls/certs/epa_server-chain.crt
        SSLProtocol TLSv1.2
        SSLCompression Off
        SSLSessionTickets Off
        SSLHonorCipherOrder on
        SSLCipherSuite ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384

        # Enable HSTS
[...]

PR #3808, opened 2022-02-28
Apache: set X-Forwarded-Proto header

manics:
This should ensure JupyterHub correctly compares the protocol used for the request (e.g. https) with the protocol from the referrer when checking for cross-site requests in the admin pages.

Closes https://github.com/jupyterhub/jupyterhub/issues/3780

@rzo1 @tobi45 would you mind checking this matches your configuration?

Preview of the rendered docs is at https://jupyterhub--3808.org.readthedocs.build/en/3808/reference/config-proxy.html#apache

### DIFF
--- a/docs/source/reference/config-proxy.md
+++ b/docs/source/reference/config-proxy.md
@@ -208,6 +208,7 @@ Listen 443
     # proxy to JupyterHub
     ProxyPass         http://127.0.0.1:8000/
     ProxyPassReverse  http://127.0.0.1:8000/
+    RequestHeader     set "X-Forwarded-Proto" expr=%{REQUEST_SCHEME}
   </Location>
 </VirtualHost>
 ```


rzo1 (approved):
This matches our current configuration - so should work fine.

However, I would suggest to update the SSL/TLS configuration (to a more modern config) in an other PR.

tobi45 (approved):
Can confirm that adding that line solves #3780.

WebSocket-Config and Proxy-Configuration matches my apache config.

consideRatio:
> I would suggest to update the SSL/TLS configuration (to a more modern config)

@rzo1 are you referring to this section?

```
  # configure SSL
  SSLEngine on
  SSLCertificateFile /etc/letsencrypt/live/HUB.DOMAIN.TLD/fullchain.pem
  SSLCertificateKeyFile /etc/letsencrypt/live/HUB.DOMAIN.TLD/privkey.pem
  SSLProtocol All -SSLv2 -SSLv3
  SSLOpenSSLConfCmd DHParameters /etc/ssl/certs/dhparam.pem
  SSLCipherSuite EECDH+AESGCM:EDH+AESGCM:AES256+EECDH:AES256+EDH
```

I think indeed that would be important to keep updated, preferably with a comment next to it saying that it may needs to be updated over time and perhaps also with a link to some helpful resource on how to do it?

I've never worked with Apache and won't try this myself. If you want, please open a PR! For reference, this is some related configuration I did of a proxy server accepting HTTPS traffic: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/38c694b3c9a0a53f846379fd418a4a718eb073b7/jupyterhub/templates/proxy/autohttps/_configmap-dynamic.yaml#L89-L108

rzo1:
> @rzo1 are you referring to this section?

Yes. A good reference is offered by Mozilla: https://ssl-config.mozilla.org/#server=apache&version=2.4.41&config=intermediate&openssl=1.1.1d&guideline=5.6

We could link it in the documentation (or in a comment). I can provide a PR after this one is merged.

PR #3809, opened 2022-03-01
Add user token to JupyterLab PageConfig

minrk:
- `HubAuth.get_token()` can return the token authenticating the request:
  - in `?token=` param
  - in `Authorization: token ...` header
  - in encrypted cookie, stored after completion of oauth

This will be useful for frontend extensions that may want to make API requests to other services or the Hub itself.

Passing `in_cookie=False` to preserve prior behavior, when it matters to distinguish between cookie auth and token-in-header/param auth (has CORS effects).

- Use JupyterLab's [page_config_hook](https://github.com/jupyterlab/jupyterlab_server/pull/220) to restore PageConfig.token

The token was removed from PageConfig in JupyterHub 1.5 for GHSA-cw7p-q79f-m2v7 because it was not the right token (the server's own token) and couldn't be set on a per-request basis, but page_config_hook in jupyterlab-server 2.9 lets us put the token for the current request in page config each time.

### DIFF
--- a/docs/source/reference/services.md
+++ b/docs/source/reference/services.md
@@ -246,6 +246,19 @@ action.
 HubAuth also caches the Hub's response for a number of seconds,
 configurable by the `cookie_cache_max_age` setting (default: five minutes).
 
+If your service would like to make further requests _on behalf of users_,
+it should use the token issued by this OAuth process.
+If you are using tornado,
+you can access the token authenticating the current request with {meth}`.HubAuth.get_token`.
+
+:::{versionchanged} 2.2
+
+{meth}`.HubAuth.get_token` adds support for retrieving
+tokens stored in tornado cookies after completion of OAuth.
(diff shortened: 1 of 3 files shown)


consideRatio (approved):
This LGTM but I'm not feeling technically confident on evaluating this PR overall. @manics can you give it a glance as well?

manics:
I've tested `HubAuth.get_token` in a service and it works!

I couldn't find a quick way to test `PageConfig` in JupyterLab, is it guaranteed that this will always be called after the token is removed in
https://github.com/jupyterhub/jupyterhub/blob/bcad6e287d94fd7b7b8151a63b1a76fe0fce945d/jupyterhub/singleuser/mixins.py#L743-L753
?

minrk:
>  is it guaranteed that this will always be called after the token is removed

Yes, the removal happens one time at startup, the hook runs on every request that asks for page Config.

manics (approved):

PR #3810, opened 2022-03-01
Keep Spawner.server in sync with underlying orm_spawner.server

minrk:
Rather than one-time sets of a cached `._server` attribute, which could become out-of-sync with underlying orm_spawner.server in complex situations (wrapspawner).

Kept as a property instead of using traitlets, because this is mapping a _secondary_ attribute of an existing trait (orm_spawner.server), and we can't watch for changes of attributes on the underlying ORM object. Traitlets only allows actions on assignment or first access, not on every access like properties, which allows us to ensure it's in sync on every access.

related to https://github.com/jupyterhub/wrapspawner/pull/50

### DIFF
--- a/jupyterhub/tests/test_spawner.py
+++ b/jupyterhub/tests/test_spawner.py
@@ -483,3 +483,49 @@ async def test_spawner_options_from_form_with_spawner(db):
     for key, value in form_data.items():
         assert key in result
         assert result[key] == value
+
+
+def test_spawner_server(db):
+    spawner = new_spawner(db)
+    spawner.orm_spawner = None
+    orm_spawner = orm.Spawner()
+    orm_server = orm.Server(base_url="/1/")
+    orm_spawner.server = orm_server
+    db.add(orm_spawner)
+    db.add(orm_server)
+    db.commit()
+    # initial: no orm_spawner
+    assert spawner.server is None
+    # assigning spawner.orm_spawner updates spawner.server
+    spawner.orm_spawner = orm_spawner
+    assert spawner.server is not None
+    assert spawner.server.orm_server is orm_server
+    # update orm_spawner.server without direct access on Spawner
+    orm_spawner.server = new_server = orm.Server(base_url="/2/")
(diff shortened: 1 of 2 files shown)


jupyterhub/tests/test_spawner.py
```diff
@@ -483,3 +483,48 @@ def options_from_form(form_data, spawner):
...
+def test_spawner_server(db):
+    spawner = new_spawner(db)
+    orm_spawner = orm.Spawner()
+    orm_server = orm.Server(base_url="/1/")
+    orm_spawner.server = orm_server
+    db.add(orm_spawner)
+    db.add(orm_server)
+    db.commit()
+    # initial: no orm_spawner
+    assert spawner.server is None
```
consideRatio, line 497: Error on this line, ping me for review when ready! I didn't quickly figure out what was wrong with assumptions/logic etc.

jupyterhub/tests/test_spawner.py
```diff
@@ -483,3 +483,49 @@ def options_from_form(form_data, spawner):
...
+    spawner.orm_spawner = None
+    orm_spawner = orm.Spawner()
+    orm_server = orm.Server(base_url="/1/")
+    orm_spawner.server = orm_server
+    db.add(orm_spawner)
+    db.add(orm_server)
+    db.commit()
+    # initial: no orm_spawner
+    assert spawner.server is None
+    # assign orm_spawner sets server
```
GeorgianaElena, line 499:
```suggestion
    # assert orm_spawner sets server
```

minrk: I actually meant `assign` since it's assigning the spawner.orm_spawner attribute that updates the server value. I've updated the comments to hopefully be clearer.

GeorgianaElena: Oh, right! sorry for getting it wrong and thanks for updating the comments

manics (approved):
Merging based on the two reviews

PR #3813, opened 2022-03-03
Apache2 Documentation: Updates Reverse Proxy Configuration (TLS/SSL, Protocols, Headers)

rzo1:
# What does this PR do?

As discussed in #3808, this PR updates the apache proxy configuration to
- Use intermediate SSL config based on https://ssl-config.mozilla.org/#server=apache&version=2.4.41&config=intermediate&openssl=1.1.1k&ocsp=false&guideline=5.6 
- Adds reference to Mozilla SSL Config Page (so users can update)
- Sets HTTP Strict Transport Security Header
- Adds HTTP/2 as protocol


minrk:
Thanks!

PR #3819, opened 2022-03-10
allow Spawner.server to be mocked without underlying orm_spawner

minrk:
This can't happen in reality, but can happen in tests (https://github.com/jupyterhub/kubespawner/issues/582).

Instead of being strict about always reflecting underlying orm_spawner, allow setting Spawner.server to persist if Spawner.orm_spawner is never defined.


consideRatio (approved):
Thanks @minrk for quick followup, this LGTM!

I added the label `maintenance` for this PR for github-activity to consider.

yuvipanda:
Thanks, @minrk!

PR #3820, opened 2022-03-10
Add some docs on common log messages

yuvipanda:
When debugging errors and outages, looking at the logs emitted by
JupyterHub is very helpful. This document tries to document some common
log messages, and what they mean.

I currently added just one log message, but we can add more
over time.

Ref https://github.com/2i2c-org/infrastructure/issues/1081
where this would've been useful troubleshooting

### DIFF
--- a/docs/source/admin/log-messages.md
+++ b/docs/source/admin/log-messages.md
@@ -1,0 +1,32 @@
+# Common log messages emitted by JupyterHub
+
+When debugging errors and outages, looking at the logs emitted by
+JupyterHub is very helpful. This document tries to document some common
+log messages, and what they mean.
+
+## Failing suspected API request to not-running server
+
+Your logs might be littered with lines that might look slightly scary
+
+```
+[W 2022-03-10 17:25:19.774 JupyterHub base:1349] Failing suspected API request to not-running server: /hub/user/<user-name>/api/metrics/v1
+```
+
+However, what this most likely means is that the user's server has
+stopped running but they still have a browser tab open. This is
+extremely common - for example, you might have 3 tabs open, and shut
+your server down via one. Or you closed your laptop, your server was
+culled for inactivity, and then you reopen your laptop again! The
+client side code (JupyterLab, Classic Notebook, etc) does not know
+yet that the server is dead, and continues to make some API requests.
+JupyterHub's architecture means that the proxy routes all requests that
+don't go to a running user server to the hub process itself. The hub
+process then explicitly returns a failure response, so the client knows
+that the server is not running anymore. This is used by JupyterLab to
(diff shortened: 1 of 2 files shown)


consideRatio (approved):
I like the idea of committing to maintaining documentation like this centrally in the JupyterHub project. If it gets outdated, it isn't that problematic or will cause more confusion. Instead I find it could be a great hatch to document something once and for all to not have to answer it in multiple places.

This makes sense to me, but I figure this could use at least one more review as it means a commitment to something new to maintain.

choldgraf:
I think this page will be most-maintainable and easiest to quickly read if it is short and to-the-point with a specific structure. What do you think about structuring this page in a specific way for each message? Something like:

````
## <message snippet that a user would be likely to copy/paste from their logs to search for>

### Example ouptut

```
<paste sample log output>
```

### Most likely cause

Two to three sentences max about what this probably means.

### Steps to try

Bullet list of steps to try and resolve this log, if any.
````

**Comment by yuvipanda** [comment]
@choldgraf thanks for the suggestion. I've added that structure.

File: docs/source/admin/log-messages.md [hunk 1]
```diff
@@ -0,0 +1,38 @@
...
+your server down via one. Or you closed your laptop, your server was
+culled for inactivity, and then you reopen your laptop again! The
+client side code (JupyterLab, Classic Notebook, etc) does not know
+yet that the server is dead, and continues to make some API requests.
+JupyterHub's architecture means that the proxy routes all requests that
+don't go to a running user server to the hub process itself. The hub
+process then explicitly returns a failure response, so the client knows
+that the server is not running anymore. This is used by JupyterLab to
+tell you your server is not running anymore, and offer you the option
+to let you restart it.
```
choldgraf, line 31:
```suggestion
Tjos likely means is that the user's server has stopped running but they
still have a browser tab open. For example, you might have 3 tabs open, and shut
your server down via one. Or you closed your laptop, your server was
culled for inactivity, and then you reopen your laptop again! The
client side code (JupyterLab, Classic Notebook, etc) does not know
yet that the server is dead, and continues to make some API requests.
JupyterHub's architecture means that the proxy routes all requests that
don't go to a running user server to the hub process itself. The hub
process then explicitly returns a failure response, so the client knows
that the server is not running anymore. This is used by JupyterLab to
tell you your server is not running anymore, and offer you the option
to let you restart it.
```

choldgraf (approved):
Looks good after one minor edit suggestion

yuvipanda:
@choldgraf thanks :)

meeseeksmachine:

https://discourse.jupyter.org/t/documenting-common-jupyterhub-log-messages/13358/1

PR #3822, opened 2022-03-11
Fix admin dashboard table sorting

NarekA:
My previous changes to the admin dashboard seems to introduce errors when sorting the table.

Before:

https://user-images.githubusercontent.com/737367/157815221-62f28df7-c3c1-4b1b-8d3e-d68561839fe1.mov

After:

https://user-images.githubusercontent.com/737367/157815293-df0b4286-db2d-44f1-a9bd-bc44d51aa402.mov


jsx/src/components/ServerDashboard/ServerDashboard.jsx
```diff
@@ -339,87 +346,65 @@ const ServerDashboard = (props) => {
...
-                    {!server.name && (
-                      <td
-                        data-testid="user-row-admin"
-                        rowspan={userServers.length}
-                      >
-                        {e.admin ? "admin" : ""}
-                      </td>
-                    )}
+            {servers.map(([user, server], i) => {
+              server.name = server.name || "";
```
minrk, line 350: Test seems to be [failing here](https://github.com/jupyterhub/jupyterhub/runs/5507233231?check_suite_focus=true#step:6:650) with 

```
    TypeError: Cannot read property 'name' of null
      348 |             </tr>
      349 |             {servers.map(([user, server], i) => {
    > 350 |               server.name = server.name || "";
          |                                    ^
      351 |               return (
      352 |                 <tr key={i + "row"} className="user-row">
      353 |                   <td data-testid="user-row-name">{user.name}</td>
```

Presumably need to handle `if server` here.

consideRatio: It seems there are a lot of places below where `server.someProperty` is accessed, so addressing this requires several lines changes or to initialize some variable to a blank object.

I note that the `servers` list is created above, perhaps the sensible fix is to make sure it doesn't generate `null` entries for `server`. I've added a change suggestion.

jsx/src/components/ServerDashboard/ServerDashboard.jsx
```diff
@@ -176,6 +175,14 @@ const ServerDashboard = (props) => {
     );
   };
 
+  let servers = user_data.flatMap((user) => {
+    let userServers = Object.values({
+      "": user.server,
```
consideRatio, line 180: I think this could be the way to avoid issues below where it is assumed that you have a default server.

```suggestion
      "": user.server || {},
```

consideRatio: This code suggestion makes the tests pass, see https://github.com/consideRatio/jupyterhub/runs/5511339074?check_suite_focus=true.

consideRatio (approved):
This LGTM, thank you @NarekA!! :heart: :tada:

PR #3825, opened 2022-03-11
Update admin-react.js

NarekA:
Last admin PR someone fixed a CI error but did not update `admin-react.js`


NarekA:
@minrk @consideRatio the following [commit] in 
https://github.com/jupyterhub/jupyterhub/pull/3822 (https://github.com/jupyterhub/jupyterhub/pull/3822/commits/33ccfa7963505d6e759f72046e099e6d4ad5889d) did not update the compiled react file.

consideRatio:
@NarekA ooooh nice catch, hmmm! This also makes me think we should have a test to verify that the compiled output is the same as the source. If not, someone could easily sneak in a vulnerability to the compiled code as it is impossible to check in code review.

I'll merge this and develop such CI test.

minrk:
Perhaps we should remove admin-react.js from the repo and build it in setup.py. That's what we do with other things.

NarekA:
Yes, this is something that would be better if you did it as part of the CI/CD process.

manics:
> Perhaps we should remove admin-react.js from the repo and build it in setup.py

I'm thinking of trying Jupyter-packaging as used in some of our other repos

consideRatio:
@minrk @manics I've opened #3826 as a short term resolution. I'd like to cut 2.2.2 with this bugfix quickly as I figure 2.2.1 could be broken.

If you agree, I'll go for a quick release unless you do it even before me.

PR #3826, opened 2022-03-11
ci: standalone jsx workflow and verify compiled asset matches source code

consideRatio:
https://github.com/jupyterhub/jupyterhub/pull/3825 made us aware that we could fail to forget to compile the source code and ship an outdated compiled asset.

It also alerted me that an ill willed actor could submit a PR that was sound in the source code but wasn't sound in the compiled asset.

With a check like this, we should be safe from forgetting to compile a change as well as having someone introduce malicious code in the compiled asset.


NarekA:
This is a good idea, I might have to rebase https://github.com/jupyterhub/jupyterhub/pull/3827 on top of this.

PR #3827, opened 2022-03-11
[Admin Dash] Add search bar for user name

NarekA:
At our organization size, the admin dashboard is unusable without a search bar. Since I was already working in this module, I thought I would add a search bar

https://user-images.githubusercontent.com/737367/157988015-f6c48132-07d9-4166-8cac-1fa9598fed32.mov

With pagination (page size 3):

https://user-images.githubusercontent.com/737367/157988076-6279b94e-36ee-432f-9cc5-e36d3623dbf2.mov


minrk:
Nice! Can you add a test for the new features in both the REST API and jsx?

NarekA:
@minrk added tests for backend and jsx. Anything else you want before we merge this?

ykazakov:
@NarekA Nice work! I just noticed that the counter X in the label "Displaying X-Y" is a little off. In your animations it should be "Displaying 1-10" and "Displaying 4-5". Would it be possible to change the label how it was in Jupyterhub 1.5.x: "Displaying users X - Y of Z"? Note the spaces around `-`.

ykazakov (commented):
I believe, the `Spawn Page` button should also use `JupyterHub.base_url`.

https://github.com/jupyterhub/jupyterhub/blob/db7619fa7aca5cb4c529a7bb259f0d9b75194ccd/jsx/src/components/ServerDashboard/ServerDashboard.jsx#L414

ykazakov:
Another minor thing: would it be possible to add a tiny separator between the buttons `Start Server`/`Stop Sever` and `Spawn Page`/`Access Server`? Similar like between `Start All` and `Stop All` in the fourth column.

NarekA:
@ykazakov I fixed the spawn url, but I want to limit the scope of this PR, so I am going to leave the button styling as it is.

minrk:
@ykazakov feel free to open a separate issue/PR for the layout issue.

@NarekA thanks!

PR #3828, opened 2022-03-14
Add changelog for 2.2.2

consideRatio:
I suggest we cut a release with #3825, and figure we include #3826 as well that ensures we don't let this happen again.

### DIFF
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -8,7 +8,27 @@ command line for details.
 
 ## 2.2
 
-### 2.2.1 2021-03-11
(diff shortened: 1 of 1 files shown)


docs/source/changelog.md
```diff
@@ -81,7 +101,7 @@ The links to access user servers from the admin page have been restored.
 
 ## 2.1
 
-### 2.1.1 2021-01-25
+### 2.1.1 2022-01-25
```
minrk, line 104: Ha, thanks!

minrk (approved):

PR #3833, opened 2022-03-15
Tokens have scopes instead of roles

minrk:
Following #3713 (included here, will rebase after merge), it's proven to be a problem that API tokens can only be assigned _roles_ instead of permissions.

Among other things, this means that intersections of requested scopes and user-owned scopes cannot be assigned to tokens, making granular permissions difficult to manage.

The main change here is resolving roles to scopes when tokens are issued, rather than storing role relationships

- [x] merge #3713
- [x] Tokens have scopes, not roles
- [x] Creating tokens with roles still works, but scopes are resolved immediately when the token is issued instead of lazily, at each authorized request
- [x] On db upgrade, role assignments are turned into scopes before dropping token roles (needed some tweaks to migration config to ensure the detection of the 2.0 api_tokens drop would work)
- [x] Small reorganization of scope/role utilities to make it easier to work with scopes at the relevant levels here
- [x] handle migration of roles->scopes in oauth `?scope` parameter. Currently backward-compatible just by allowing both.
- [x] no longer need to check for permission-changes on tokens at startup, as we did with roles, because scopes are now fixed.
- [x] expand some test coverage for new branches

### DIFF
--- a/jupyterhub/alembic/env.py
+++ b/jupyterhub/alembic/env.py
@@ -65,8 +65,9 @@ def run_migrations_offline():
 
     """
     connectable = config.attributes.get('connection', None)
-    config_opts = dict(literal_binds=True)
+    config_opts = {}
     config_opts.update(common_config_opts)
+    config_opts["literal_binds"] = True
 
     if connectable is None:
         config_opts["url"] = config.get_main_option("sqlalchemy.url")
--- a/jupyterhub/alembic/env.py
+++ b/jupyterhub/alembic/env.py
@@ -9,7 +9,6 @@ from sqlalchemy import pool
 # this is the Alembic Config object, which provides
 # access to the values within the .ini file in use.
 config = context.config
-
 # Interpret the config file for Python logging.
 # This line sets up loggers basically.
 if 'jupyterhub' in sys.modules:
@@ -42,6 +41,16 @@ target_metadata = orm.Base.metadata
 # my_important_option = config.get_main_option("my_important_option")
 # ... etc.
 
+# pass these to context.configure(**config_opts)
+common_config_opts = dict(
+    # target_metadata for autogenerate
+    target_metadata=target_metadata,
+    # transaction per migration to ensure
+    # each migration is 'complete' before running the next one
+    # (e.g. dropped tables)
+    transaction_per_migration=True,
+)
+
 
 def run_migrations_offline():
     """Run migrations in 'offline' mode.
@@ -56,14 +65,14 @@ def run_migrations_offline():
 
     """
     connectable = config.attributes.get('connection', None)
+    config_opts = dict(literal_binds=True)
+    config_opts.update(common_config_opts)
 
     if connectable is None:
-        url = config.get_main_option("sqlalchemy.url")
-        context.configure(url=url, target_metadata=target_metadata, literal_binds=True)
+        config_opts["url"] = config.get_main_option("sqlalchemy.url")
     else:
-        context.configure(
-            connection=connectable, target_metadata=target_metadata, literal_binds=True
-        )
+        config_opts["connection"] = connectable
+    context.configure(**config_opts)
 
     with context.begin_transaction():
         context.run_migrations()
@@ -77,6 +86,8 @@ def run_migrations_online():
 
     """
     connectable = config.attributes.get('connection', None)
+    config_opts = {}
+    config_opts.update(common_config_opts)
 
     if connectable is None:
         connectable = engine_from_config(
@@ -86,7 +97,10 @@ def run_migrations_online():
         )
 
     with connectable.connect() as connection:
-        context.configure(connection=connection, target_metadata=target_metadata)
+        context.configure(
+            connection=connection,
+            **common_config_opts,
+        )
 
         with context.begin_transaction():
             context.run_migrations()
--- a/jupyterhub/oauth/provider.py
+++ b/jupyterhub/oauth/provider.py
@@ -10,6 +10,8 @@ from oauthlib.oauth2.rfc6749.grant_types import base
 from tornado.log import app_log
 
 from .. import orm
+from ..roles import roles_to_scopes
+from ..scopes import _check_scopes_exist
 from ..utils import compare_token
 from ..utils import hash_token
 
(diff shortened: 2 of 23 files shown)


minrk:
@consideRatio I think this is ready for a first read through. I think I can still improve some things and add test coverage for some of the changes.

The gist is that the underlying storage of permissions on tokens is scopes, not roles (effectively concatenating role.scopes for each assigned role at assignment time, rather than on each future request). This happens on tokens themselves, but also during the oauth flow because the user authorizes scopes which are persisted on the OAuthCode until they are finally assigned to the token.

The diff in scopes/roles is bigger than it looks, because I moved some scope utilities into scopes.py, which happened to be in roles.py.

jupyterhub/alembic/env.py
```diff
@@ -56,14 +65,14 @@ def run_migrations_offline():
 
     """
     connectable = config.attributes.get('connection', None)
+    config_opts = dict(literal_binds=True)
+    config_opts.update(common_config_opts)
```
consideRatio, line 69: Feel free to ignore. It doesn't matter, but it made me need to think "hmmm is common_config_ops including literal_binds? If so, that could lead to a bug.

```suggestion
    config_opts = {}
    config_opts.update(common_config_opts)
    config_opts["literal_binds"] = True
```

jupyterhub/oauth/provider.py
```diff
@@ -547,35 +549,51 @@ def validate_scopes(self, client_id, scopes, client, request, *args, **kwargs):
...
-        client_allowed_roles.setdefault('identify', None)
-        roles = []
-        requested_roles = set(scopes)
-        disallowed_roles = requested_roles.difference(client_allowed_roles)
-        if disallowed_roles:
+        scopes = set(scopes)
+        scopes.discard("identify")
+
+        # TODO: handle roles->scopes transition
+        # at this point, 'scopes' _may_ be roles
```
consideRatio, line 566: Can you make this comment more explicit? To clarify my confusion, "at this point" could refer to the upcoming Jupyterhub version or the current WIP state of this PR etc.

jupyterhub/orm.py
```diff
@@ -676,9 +676,11 @@ def find(cls, db, token, *, kind=None):
     def new(
         cls,
         token=None,
+        *,
```
consideRatio, line 679: Hmmm. I've forgot, what was this used to? To accept but ignore some arguments? Can you help me? It felt a bit complicated to google about.

minrk: `*` ensures all following arguments are keyword _only_. In Python, by default all keyword args can be passed positionally, but when you have more than a couple of arguments that's hard to read. The `*` stops positional parsing, to ensure all these args are passed as keywords. The effect here: `token` is the only argument accepted positionally, all others must be passed `arg=value`. This was already the case, but it is now explicit. It's part of what allows inserting a new arg (scopes) somewhere other than the end. If it could be passed positionally, that would cause weird results.

It's a way to say "Don't call me with a bunch of positional arguments, it'll be confusing."

For example:

```python
def foo(a, b=5):
    return a * b

def bar(a, *, b=5):
    return a * b

foo(1) # OK, intended
foo(1, b=3) # OK, intended
foo(1, 3) # b passed positionally, not by name!
bar(1, 3) # fails with TypeError: bar() takes 1 positional argument but 2 were given
```

minrk: [link for reference](https://peps.python.org/pep-3102/)

consideRatio: Thank you @minrk!!!!! You made it very clear and comprehensible to me!

jupyterhub/oauth/provider.py
minrk: For compatibility, I want to make a release that doesn't break OAuth as it is. Right now, our OAuth API says: "When requesting scopes, use role names as the requested scopes, not actual scopes." This is covered in the test `test_services_auth.py:test_oauth_service_roles`. After this change, it _expects_ scopes, but _allows_ roles. I think this role->scope conversion should be _deprecated_, though I'm not 100% sure.

Assigning permissions via roles as permission-bundles is still logical and convenient (e.g. the tokens REST API still accepts roles and I'm not sure it should be deprecated), so I don't have a firm idea where exactly in user-facing APIs it should be deprecated, vs just changing the underlying storage mechanism by resolving roles->scopes at assignment time.

minrk: (notably on my todo:  test_services_auth.py:test_oauth_service_roles doesn't test requesting scopes instead of roles, so the deprecated behavior is tested, but the new behavior is not, yet)

jupyterhub/scopes.py
```diff
@@ -346,7 +347,132 @@ def get_scopes_for(orm_object):
...
+            username, _, server = value.partition('/')
+            expanded_scopes.add(f'read:users:name!user={username}')
+    else:
+        expanded_scopes = expanded_scope_names
+
+    return expanded_scopes
+
+
+def expand_scopes(scopes, owner=None):
+    """Returns a set of fully expanded scopes for a collection of raw scopes
```
consideRatio, line 434: raw scopes, metascopes, and expanded scopes.

This is a bit confusing to me, and I think I could benefit from a clarification of the terminology.

Raw scopes is not yet expanded scopes I guess, which could result in the same scopes or not, depending on if they are metascopes?

consideRatio:
Terminology update in scopes.py docstring:
  - raw scopes, metascopes, expanded scopes
  - parsed scopes, and reduced scopes

Action point you mentioned:
- Memoization to increase performance

consideRatio:
@minrk since this PR touches large files, whenever you make `git commit --amend` followed by `git push --force` I'm unable to realize what changes in the files that was changed since last review. GitHub UI can help me avoid looking for changes in unchanged files since last review though thanks to the `viewed` checkboxes, but for any files changed it just sais that the file has changed since last review.

What do you think about going for a workflow where you instead push incremental commits, which you can name simply like `fixup` or similar, which you would then just in time before merge fixup and do `git push --force`?

minrk:
Reverted history so you can see the diff since we last looked at it.

minrk:
@consideRatio okay, I think this is complete, if you want to have another look.

PR #3790, opened 2022-02-09
Add Missing Features In Admin Console

NarekA:
The current admin console is missing 2 features from previous version:

1. The links to access the users notebooks (closes #3745)
2. Rows cooresponding to named servers (closes #3788)

Screen Shot 2022-02-09 at 11 52 36 AM

PR #3834, opened 2022-03-15
Admin Dashboard - Collapsible Details View

NarekA:
I made this PR to see if this feature would be useful for other people. Right now, you can't see all of a user or server's details in the admin page so I added a collapsible view which will let you see the entire server and user objects. I'm open to ideas about how the information is displayed. Will add more tests if this feature is accepted.

improved-collapse


NarekA:
@minrk @consideRatio How do you guys feel about merging this feature?

minrk:
It seems pretty nice, to me! I'd like feedback from @naatebarber or @ykazakov

ykazakov:
This is interesting. Particularly quickly seeing all groups of the user could be useful.
What happens if the user has started several (named) severs? 
I think the panel could be a bit too crowded for displaying many tables. What if to put this information in tooltips, separately for each user and server?

NarekA:
@ykazakov we currently have one row per server so it won't show more than one server at a time.

ykazakov:
@NarekA can you provide a screenshot with multiple servers like in #3790? I thought that the details view is per user, not per server.

NarekA:
@ykazakov I modified the layout a bit to make it scale better to servers with a large states
Screen Shot 2022-03-21 at 12 56 24 PM

ykazakov:
Looks good to me! My only concern is that someone would need to maintain this UI if hub API changes. ;-) 

I wonder if for the debugging scenario mentioned by @johnkpark, a better solution would be to use [REST API](https://jupyterhub.readthedocs.io/en/stable/reference/rest.html) to access the user and server details. This can be easily done using  a notebook [like this one](https://gist.github.com/minrk/1940060144e7ae2717f57018d4964769). One can even make some nice UI using pandas and javascript to display the data in tables and use buttons and layout similar to the Admin menu. We use similar notebooks to manage users via nbgrader api.

NarekA:
@minrk Can we merge this?

minrk:
Thanks @NarekA and @ykazakov for the review!

PR #3839, opened 2022-03-19
Document version mismatch log message

yuvipanda:


docs/source/admin/log-messages.md
```diff
@@ -1,4 +1,4 @@
-# Common log messages emitted by JupyterHub
+# Interpreting common log messages
```
yuvipanda, line 1: This is a better title I think

minrk (approved):

consideRatio (approved):
Thank you, I love knowing this can be documentation I can reference when this question pops up!

PR #3851, opened 2022-03-31
!service and !server filters

minrk:
Like `!user`, but these filters apply to the _issuing oauth client_. That way, a scope like `access:servers!server` only grant access to the _issuing_ server. This came up in prototyping read-only access to Jupyter Server via custom scopes in https://gist.github.com/minrk/e5f47d6f4253532635e9716076db0fb4. Each server should only issue tokens that can _access themselves_, but this isn't possible without these filters.

<del>includes #3850</del> rebased after #3850

### DIFF
--- a/jupyterhub/scopes.py
+++ b/jupyterhub/scopes.py
@@ -519,10 +519,10 @@ def expand_scopes(scopes, owner=None, oauth_client=None):
         "service": None,
         "server": None,
     }
-    owner_name = None
+    user_name = None
     if isinstance(owner, orm.User):
-        owner_name = owner.name
-        filter_replacements["user"] = f"user={owner_name}"
+        user_name = owner.name
+        filter_replacements["user"] = f"user={user_name}"
     elif isinstance(owner, orm.Service):
         filter_replacements["service"] = f"service={owner.name}"
 
@@ -552,11 +552,11 @@ def expand_scopes(scopes, owner=None, oauth_client=None):
 
     if 'self' in expanded_scopes:
         expanded_scopes.remove('self')
-        if owner_name:
-            expanded_scopes |= _expand_self_scope(owner_name)
+        if user_name:
+            expanded_scopes |= _expand_self_scope(user_name)
         else:
             warnings.warn(
-                "Not expanding 'self' scope without owner",
+                f"Not expanding 'self' scope for owner {owner} which is not a User",
                 stacklevel=2,
             )
 
--- a/jupyterhub/scopes.py
+++ b/jupyterhub/scopes.py
@@ -341,7 +341,13 @@ def get_scopes_for(orm_object):
             # only thing we miss by short-circuiting here: warning about excluded extra scopes
             return owner_scopes
 
-        token_scopes = set(expand_scopes(token_scopes, owner=owner))
+        token_scopes = set(
+            expand_scopes(
+                token_scopes,
+                owner=owner,
+                oauth_client=orm_object.oauth_client,
+            )
+        )
 
         if orm_object.client_id != "jupyterhub":
             # oauth tokens can be used to access the service issuing the token,
@@ -468,7 +474,7 @@ def _expand_scope(scope):
     return frozenset(expanded_scopes)
 
 
-def _expand_scopes_key(scopes, owner=None):
+def _expand_scopes_key(scopes, owner=None, oauth_client=None):
     """Cache key function for expand_scopes
 
     scopes is usually a mutable list or set,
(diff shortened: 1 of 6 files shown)


meeseeksmachine:

https://discourse.jupyter.org/t/refining-server-scopes-in-role-definitions/13760/2

jupyterhub/scopes.py
```diff
@@ -436,36 +442,79 @@ def _add_subscopes(scope_name):
...
+        "user": None,
+        "service": None,
+        "server": None,
+    }
+    owner_name = None
     if isinstance(owner, orm.User):
         owner_name = owner.name
-    else:
-        owner_name = None
+        filter_replacements["user"] = f"user={owner_name}"
```
consideRatio, line 493: I'm struggling to read this logic with `owner_name` as a variable set only for `owner` being an `orm.User` instance, but not otherwise.

Later there is code like `if owner_name` which doesn't really check if there is an owner name, but if there is a `owner` of instance `orm.User`, then it sets `expanded_scopes |= _expand_self_scope(owner_name)`.

I couldn't come up with a concrete suggested change. I'm thinking perhaps a rename of `owner_name` to `user_owner_name` or having an inline comment next to the `if owner_name` statement explaining what goes on a bit more.

jupyterhub/tests/conftest.py
```diff
@@ -285,7 +285,22 @@ class MockServiceSpawner(jupyterhub.services.service._ServiceSpawner):
...
+        app: MockHub application
+        external (bool):
+          If False (default), launch the service.
+          Otherwise, consider it 'external,
+          registering a service in the database,
+          but don't start it.
+        url (bool):
+          If True, register the service at a URL
+          (as opposed to headless, API-only).
+    """
```
consideRatio, line 303: :heart: :tada: :sunflower: :+1: :100:!

jupyterhub/tests/conftest.py
```diff
@@ -296,6 +311,10 @@ def _mockservice(request, app, url=False):
         else:
             spec['url'] = 'http://127.0.0.1:%i' % random_port()
 
+    if external:
+
+        spec['oauth_redirect_uri'] = 'http://127.0.0.1:%i' % random_port()
```
consideRatio, line 316: I always get a bit confused about the `%s`, `%i`, `%d`, `%r` etc.

I understand that this statement with `%i` is equivalent to `%d`, and it seems to be acting like `f{random_port():.0f}`.

I see

```python
def random_port():
    """Get a single random port."""
    sock = socket.socket()
    sock.bind(('', 0))
    port = sock.getsockname()[1]
    sock.close()
    return port
```

I assume random_port is always a integer, but is this protection for if it is a float? I note that both f strings with `:.0f` and `%i` will error if a string or `None` is given.

consideRatio: Is there relevant intent to use/not use an f-string like the one below where we don't coerce a float to an int, and is there relevant intent to use/not using an f-string with `.0f` formatting instead of `%i`?

```
        spec['oauth_redirect_uri'] = f'http://127.0.0.1:{random_port()}'
```

I'm asking to learn if I need to adopt certain patterns for various situations or if doesn't matter.

consideRatio (approved):
Looks great, nice!

jupyterhub/tests/conftest.py
minrk: I wouldn't call it protection, it's just the idiomatic way to put an integer in a string in the old percent-style (I've been writing it that way for a long time!). I might have used an f-string if I were typing from scratch, but you can see this is a copy/paste of the line just above. If I used an f-string, I wouldn't add a type formatter (If I did, it would be `f"{random_port():d}"` where 'd' is for [decimal integer](https://docs.python.org/3/library/string.html#format-specification-mini-language). But I'd trust the return type of the function in this case.

I wouldn't use a float format string (`.0f`) for something like a port that's an integer by definition.

jupyterhub/scopes.py
minrk: since it's only defined when the owner is a user, we can use `user_name`.

jupyterhub/tests/conftest.py
minrk: > I'm asking to learn if I need to adopt certain patterns for various situations or if doesn't matter.

It doesn't matter. It's just me copying pre-f-string code / habits. An f-string with no formatter would be AOK here, for me. I've just written `host:%i' % port` so many times in my life, it's hard to stop!

minrk:
@consideRatio thanks for the feedback! Rebased and took a pass at clarifying `user_name=owner_name`.

PR #3852, opened 2022-03-31
Use isort for import formatting

minrk:
Having used isort and reorder_python_imports in a few projects, I think isort produces substantially nicer output more often, especially when there are lots of imports. The main difference is that isort using a single import statement per module rather than one line per imported symbol. It also reformats imports in a few more places, where reorder_python_imports stops consideration the moment it sees a non-import statement.

reorder_python_imports trades more natual code and compactness for reducing the likelihood of merge conflicts for nearby imports _sometimes_. In my experience, it often doesn't really accomplish this goal, so the nicer output seems preferable to me.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,10 +18,10 @@ repos:
           - --py36-plus
 
   # Autoformat: Python code
-  - repo: https://github.com/asottile/reorder_python_imports
-    rev: v3.0.1
+  - repo: https://github.com/pycqa/isort
+    rev: 5.10.1
     hooks:
-      - id: reorder-python-imports
+      - id: isort
 
   # Autoformat: Python code
   - repo: https://github.com/psf/black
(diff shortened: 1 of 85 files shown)


consideRatio:
I agree, this LGTM!

consideRatio:
I'll go for a merge as this kind of PR is troublesome to have hanging. I also believe there is general agreement that grouping imports into a single line (unless too long) is preferred over letting them have one line per import.

Ping @yuvipanda for transparency about this topic. Do you agree that `isort --profile=black` is to be preferred over `reorder-python-imports`? I have assumed you do because it would save a lot of vertical space which I recall you raised as a key concern with these autoformatters.

yuvipanda:
I LIKE THIS SO MUCH MOREEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEE!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

Thank you @minrk and @consideRatio! This was my *only* issue with the autoformatters.

Issue #3821, opened 2022-03-10
Jupyterhub logout TypeError with xsrf_cookie_kwargs


jwclark:
Jupyterhub version 2.1.1

If xsrf_cookie_kwargs (example below) are configured on Jupyterhub tornado_settings, the request to clear_cookie() in base.py clear_login_cookie during logout can result in the following error:

TypeError: clear_cookie() got an unexpected keyword argument 'Secure'

Example configuration:

```python
c.JupyterHub.tornado_settings = {
    'xsrf_cookie_kwargs': {
        'Secure': True
    }
  }
```

It appears appears that the logout handler attempts to pass in all kwargs to web.py clear_cookies() method for the _xsrf cookie, which accepts only name, path and domain as parameters.

https://github.com/jupyterhub/jupyterhub/blob/c23cddeb515dbcc0bc59c063ddba32d191624dcb/jupyterhub/handlers/base.py#L532

minrk:
Should be fixed by #3823

PR #3853, opened 2022-04-01
Fix xsrf_cookie_kwargs ValueError

jwclark:
Fixes 

`ValueError: too many values to unpack (expected 2)`

Related to code added as a fix for https://github.com/jupyterhub/jupyterhub/issues/3821


minrk:
Thanks!

PR #3837, opened 2022-03-18
ensure _import_error is set when JUPYTERHUB_SINGLEUSER_APP is unavailable

minrk:
closes #3832

PR #3857, opened 2022-04-05
fixing Jupyter App import error

BSCrumpton:
Fixes: TypeError: exceptions must derive from BaseException whenever there is no jupyter client in your environment


minrk:
Thanks for the PR! I beleive this was fixed already by #3837

PR #3867, opened 2022-04-15
ci: update black configuration

consideRatio:


manics:
What does this do? Does it make sense to target more than one python version?

consideRatio:
Apparently its meant to list all versions one wants to ve compatible against. It is confusing indeed, forgot where i read it. This just updates some references not knowing if it matters, but it seems it didnt.

manics:
Oh well... Maybe add a comment saying it's a mystery (or something more sensible 😀) and merge if it's no harm?

consideRatio:
@manics done!

minrk:
I had no idea, thanks!

PR #3874, opened 2022-04-24
Missing `f` prefix on f-strings fix

### LINKED ISSUE #3873: Missing `f` prefix on f-strings
Some strings looks like they're meant to be f-strings but are missing the `f` prefix meaning variable interpolation won't happen.

https://github.com/jupyterhub/jupyterhub/blob/a588a0bfa39691793a6b8f788b112d8189179134/jupyterhub/handlers/pages.py#L499
https://github.com/jupyterhub/jupyterhub/blob/a588a0bfa39691793a6b8f788b112d8189179134/jupyterhub/spawner.py#L233
https://github.com/jupyterhub/jupyterhub/blob/a588a0bfa39691793a6b8f788b112d8189179134/jupyterhub/user.py#L272
https://github.com/jupyterhub/jupyterhub/blob/a588a0bfa39691793a6b8f788b112d8189179134/jupyterhub/user.py#L274
https://github.com/jupyterhub/jupyterhub/blob/a588a0bfa39691793a6b8f788b112d8189179134/jupyterhub/apihandlers/users.py#L127
https://github.com/jupyterhub/jupyterhub/blob/a588a0bfa39691793a6b8f788b112d8189179134/jupyterhub/apihandlers/groups.py#L52

I found this issue automatically. I'm a bot. Beep Boop 🦊. See other issues I found in your repo [here](https://codereview.doctor/jupyterhub/jupyterhub)

code-review-doctor:
Fixes #3873

### DIFF
--- a/jupyterhub/handlers/pages.py
+++ b/jupyterhub/handlers/pages.py
@@ -496,7 +496,7 @@ class TokenPageHandler(BaseHandler):
                 continue
             if not token.client_id:
                 # token should have been deleted when client was deleted
-                self.log.warning("Deleting stale oauth token {token}")
+                self.log.warning(f"Deleting stale oauth token {token}")
                 self.db.delete(token)
                 self.db.commit()
                 continue
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -230,7 +230,7 @@ class Spawner(LoggingConfigurable):
                 self.orm_spawner.server = server.orm_server
         elif server is not None:
             self.log.warning(
-                "Setting Spawner.server for {self._log_name} with no underlying orm_spawner"
+                f"Setting Spawner.server for {self._log_name} with no underlying orm_spawner"
             )
 
     @property
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -99,10 +99,15 @@ class Spawner(LoggingConfigurable):
 
         Used in logging for consistency with named servers.
         """
-        if self.name:
-            return f'{self.user.name}:{self.name}'
+        if self.user:
+            user_name = self.user.name
         else:
-            return self.user.name
+            # no user, only happens in mock tests
+            user_name = "(no user)"
+        if self.name:
+            return f"{user_name}:{self.name}"
+        else:
+            return user_name
 
     @property
     def _failed(self):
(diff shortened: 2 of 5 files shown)


jupyterhub/handlers/pages.py
```diff
@@ -496,7 +496,7 @@ def sort_key(token):
                 continue
             if not token.client_id:
                 # token should have been deleted when client was deleted
-                self.log.warning("Deleting stale oauth token {token}")
+                self.log.warning(f"Deleting stale oauth token {token}")
                 self.db.delete(token)
                 self.db.commit()
```
consideRatio, line 501: If we are to print a token, lets do it after we are sure we have deleted it.

```suggestion
                self.db.delete(token)
                self.db.commit()
                self.log.warning(f"Deleted stale oauth token {token}")
```

minrk: The repr of an ORM token object does not include the token value itself, so it's safe to log them. Plus, I think it's usually best to log what you are _about_ to do, because if there's an error you won't see the log message until it succeeds!

consideRatio: > The repr of an ORM token object does

Ah okay than this is perfectly fine!

jupyterhub/spawner.py
```diff
@@ -230,7 +230,7 @@ def server(self, server):
                 self.orm_spawner.server = server.orm_server
         elif server is not None:
             self.log.warning(
-                "Setting Spawner.server for {self._log_name} with no underlying orm_spawner"
+                f"Setting Spawner.server for {self._log_name} with no underlying orm_spawner"
```
code-review-doctor, line 233: the test fails with this error:
```py
      @property
      def _log_name(self):
          """Return username:servername or username
      
          Used in logging for consistency with named servers.
          """
          if self.name:
              return f'{self.user.name}:{self.name}'
          else:
  >           return self.user.name
  E           AttributeError: 'NoneType' object has no attribute 'name'
```

can someone give guidance of how to fix the test :)

minrk: SHould be fixed by 6c2c5e5

consideRatio:
Thanks for the help @code-review-doctor!!

PR #3876, opened 2022-04-28
don't confuse :// in next_url query params for a redirect hostname

### LINKED ISSUE #3014: nbgitpuller repo target triggers disallowed redirect
### Bug description

We've been seeing a problem where users following nbgitpuller links would sometimes have to go through a two step process the repository would clone/pull (I think this jupyterhub/nbgitpuller#118 is the same problem). It looks like [next_url check](https://github.com/jupyterhub/jupyterhub/blob/1f515464fe2edf5d9d8a7617c22ebe67002e0ba0/jupyterhub/handlers/base.py#L620-L626) notices when the `repo=` part of the query string contains unencoded special characters (specifically the `://`) and triggers this warning from the hub `Disallowing redirect outside hub` and redirects the user to `/hub/home`, skipping nbgitpuller.

**N.B. This only seems to happen when the user is not authenticated to start with, so visiting the nbgitpuller link a second time works**

#### Expected behaviour
Clicking an ngitpuller link should send the user through authentication then on to nbgitpuller as a single process without the user having to re-click the link.

#### Actual behaviour
If the user is not already logged in and follows an nbgitpuller link they are authenticated but nbgitpuller is not run. If they click the link a second time (now that they are logged in) nbgitpuller will do its thing and clone/pull.

### How to reproduce
1. Make sure you're not authenticated to the hub
2. Visit an nbgitpuller link where the repo target hasn't been encoded, e.g.
```
https://hub.example.com/hub/user-redirect/git-pull?repo=https://github.com/callysto/callysto-sample-notebooks
```
3. Authenticate to the hub
4. You'll be taken to `/hub/home` and nbgitpuller will not run.

### Workaround
Make sure the query string is properly encoded. The [nbgitpuller link generator](https://jupyterhub.github.io/nbgitpuller/link) does this for you automatically, but some URL shortening services (e.g. bit.ly) will undo the encoding and trigger the behaviour.

### Your personal set up
I've reproduced this on a couple of different hubs both running JupyterHub 1.1 & using different browsers (Chrome & Firefox) on the client side.

### Notes
* I'm not sure why this isn't triggered when the user is already authenticated
* This might be the intended behaviour and the solution might be to remind people that those characters should be encoded, but it comes up _a lot_ so I thought I should file an issue.

minrk:
closes #3014

These query params should be url-encoded (https://github.com/jupyterhub/nbgitpuller/issues/118), but we still shouldn't be making the wrong assumptions about when a hostname is specified

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -635,29 +635,32 @@ class BaseHandler(RequestHandler):
         next_url = next_url.replace('\\', '%5C')
         proto = get_browser_protocol(self.request)
         host = self.request.host
+        if next_url.startswith("///"):
+            # strip more than 2 leading // down to 2
+            # because urlparse treats that as empty netloc,
+            # whereas browsers treat more than two leading // the same as //,
+            # so netloc is the first non-/ bit
+            next_url = "//" + next_url.lstrip("/")
+        parsed_next_url = urlparse(next_url)
 
         if (next_url + '/').startswith((f'{proto}://{host}/', f'//{host}/',)) or (
             self.subdomain_host
-            and urlparse(next_url).netloc
-            and ("." + urlparse(next_url).netloc).endswith(
+            and parsed_next_url.netloc
+            and ("." + parsed_next_url.netloc).endswith(
                 "." + urlparse(self.subdomain_host).netloc
             )
         ):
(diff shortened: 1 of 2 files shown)


GeorgianaElena (approved):
Thanks for fixing this @minrk! This makes sense to me, so merging it 🚀

Issue #3797, opened 2022-02-15
Refine hub admin page based on individual access scopes


manics:
### Proposed change
This is a follow-up from @minrk's comment https://github.com/jupyterhub/jupyterhub/pull/3790#issuecomment-1040109321

The hub admin page is no longer restricted to "full" admins, as RBAC scopes allow users to have some elevated privileges. For example a user may have access to servers in a group.

Currently the admin page requires `admin:users` `admin:servers`
https://github.com/jupyterhub/jupyterhub/blob/71bbbe4a6772f5c80079c9b123c857c706ac8e89/jupyterhub/handlers/pages.py#L461-L471

### Alternative options
Don't allow users with limited admin privileges to access the admin page. This is easier, but means people will have to build their own UI if they want these features

### Who would use this feature?
Users and admins taking advantage of JupyterHub RBAC to only give out the required privileges for users and helpers.

### (Optional): Suggest a solution
We could do one or more of:
- Relax the requirement for `admin:users` `admin:servers`, e.g. a teaching assistant may have only `admin:servers!group=...` to give them access to the servers of users in a single group, but at the moment they can't reach the admin page as they won't have `admin:users`.
- Control whether elements of the admin page are displayed or not based on scopes, e.g. the shutdown button could be disabled or hidden if a user doesn't have the `shutdown` scope
- Get rid of `c.JupyterHub.admin_access` since it [no longer has a use](https://github.com/jupyterhub/jupyterhub/pull/3790#issuecomment-1040109321), and remove checks for `if user.admin`

minrk:
I _believe_ the admin page already mostly works if you have the necessary admin rights for subsets of users (the `needs_scope` decorator passes if you have access to _any_ resources, even subsets). This is thanks to the fully REST-api-based react admin page, so listing users will only return those you have read access to. It could get weird, though, if you have e.g. admin access to a subset, but read access to more (the list would show you all the users you can read, but operations would only _work_ for those operations you have permissions to take).

One thing that might be tricky: there isn't currently a UI-accessible indication of what permissions the client has access to, and certainly not on a per-user basis, so it's not currently possible to:

1. restrict the list of users to only those for which the viewer has admin permissions, or
2. hide buttons based on permissions (which may vary per-user)

Technically, this info is available via the identity API (`/api/user` returns scopes), but would require a full client-side scope-resolution implementation, including user/server/group filtering. Simple scopes, like shutdown, shouldn't be a problem, though.

[...]

PR #3878, opened 2022-04-29
add 'admin-ui' scope for access to the admin ui

minrk:
related to: https://github.com/jupyterhub/jupyterhub/issues/3797

Grants dedicated access to the admin UI, so that you can separate access to the UI itself from access to the actions on the page.

### DIFF
--- a/jupyterhub/tests/test_pages.py
+++ b/jupyterhub/tests/test_pages.py
@@ -1105,17 +1105,27 @@ async def test_bad_oauth_get(app, params):
     [
         (["users"], False),
         (["admin:users"], False),
-        (["users", "admin:users", "admin:servers"], True),
+        (["users", "admin:users", "admin:servers"], False),
+        (["admin-ui"], True),
     ],
 )
 async def test_admin_page_access(app, scopes, has_access, create_user_with_scopes):
     user = create_user_with_scopes(*scopes)
     cookies = await app.login_user(user.name)
-    r = await get_page("/admin", app, cookies=cookies)
+    home_resp = await get_page("/home", app, cookies=cookies)
+    admin_resp = await get_page("/admin", app, cookies=cookies)
+    assert home_resp.status_code == 200
+    soup = BeautifulSoup(home_resp.text, "html.parser")
+    nav = soup.find("div", id="thenavbar")
+    links = [a["href"] for a in nav.find_all("a")]
+
+    admin_url = app.base_url + "hub/admin"
     if has_access:
-        assert r.status_code == 200
+        assert admin_resp.status_code == 200
+        assert admin_url in links
     else:
-        assert r.status_code == 403
+        assert admin_resp.status_code == 403
+        assert admin_url not in links
 
 
 async def test_oauth_page_scope_appearance(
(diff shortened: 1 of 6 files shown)


jupyterhub/tests/test_pages.py
```diff
@@ -1105,17 +1105,27 @@ async def test_bad_oauth_get(app, params):
     [
         (["users"], False),
         (["admin:users"], False),
-        (["users", "admin:users", "admin:servers"], True),
+        (["users", "admin:users", "admin:servers"], False),
```
GeorgianaElena, line 1108: I have a question that might be stupid:

I understand why access to the admin ui through `admin-ui` shouldn't require `admin:users`, `admin:servers` scopes.
But shouldn't `admin:users`, `admin:servers` (which are the highest access permissions to servers and users, right?) give you access to the admin ui?

I guess what I'm trying to say, is should `"admin:users"`, `"admin:servers"` automatically enable the `admin-ui` scope?

manics: I think it's fine to keep the two concerns separate. One is about API/backend permissions (which is complicated since RBAC means there's no clear delineation of what an admin is), the other is about whether to activate the UI/frontend regardless of permissions. It also provides an easy way to disable the admin UI, perhaps if someone builds their own.

minrk: Good question! @manics answer is what I'm thinking - permission to take actions via the API and permission to view the admin page can be kept separate. This makes it easy for a deployment to totally hide the admin ui, for example, even if some users technically have very high permissions.

GeorgianaElena: I think I understand the motivation of keeping this separate now, thank you both for taking the time to explain this.
Basically, the UI it's treated as its own resource, with its own set of permissions that shouldn't be implied from access to other resources, esp since access to those other resources makes sense regardless of the access to the UI.
Also, this new RBAC setup is super cool but also, super 🤯

manics: I just had a thought, is this considered a breaking change since people will need to add `admin-ui` to their existing scopes instead of relying on `[admin:users, admin:servers]`? I'm OK with releasing this in a minor release with instructions on adding the new scope, since it improves the usability of the admin/RBAC setup and it help with long term maintenance, but let's make it an explicit decision!

minrk: Good point! I'm okay with this in a minor release as well, though I'm increasingly thinking the custom scopes / tokens-on-scopes should perhaps be called 3.0 anyway. Even though I'm trying hard not to break anything, it's a substantial enough change that a major release might communicate it better.

manics (approved):
I'll let you decide whether to merge now, or decide on the minor vs major release first.

minrk:
Thanks @manics!

danilopeixoto:
Hi @minrk,

Will be trivial a contribution from us to hide some UI components from Admin page that the user may not have permission to access?
Eg. "Shutdown Hub" and "Add Users" buttons.

minrk:
> Will be trivial a contribution from us to hide some UI components from Admin page that the user may not have permission to access?

It may not be trivial, but I'm sure some will be easier than others.

> Does the /hub/api/user endpoint return expanded scopes, including from groups and roles assigned to the user?

Yes.

Issue #3700, opened 2021-12-02
Revoke db access from Authenticator, Spawner


minrk:
### Proposed change

We need to change the db session lifecycle (#1932), and one impediment to that is assumptions of a persistent db session on various objects. Authenticators and Spawners have this, but shouldn't. The first step is to turn the existing `db` attribute into a property that warns on access. Authenticators and Spawners should never need direct db access, but since they have it they might use it. Any current use case relying on this should be addressed in a different way. If it can't be, that  should result in a separate feature request, not keeping a db connection.

### Alternative options

Provide a `session_factory` callable instead, so that they can manage their own db connections. I'd rather they never have direct access to the db though - this was never intended and should never be used. The ORM and db schema are absolutely not supported APIs in JupyterHub, they are purely private implementation details.

### Who would use this feature?

JupyterHub maintainers, working toward HA (#1932), and eventually anyone

meeseeksmachine:
This issue has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/jupyterhub-multi-instance-node/11831/5

meeseeksmachine:
This issue has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/is-there-anything-wrong-with-using-orm-user-instead-of-get-auth-state-to-customize-user-options/13199/2

PR #3885, opened 2022-05-05
Deprecate Authenticator.db, Spawner.db

minrk:
These objects (and indeed, all extension points) should not access the shared db session; add a warning pointing to Issue about their removal if it is accessed.

Doesn't break anything, but warns on access and solicits feedback for https://github.com/jupyterhub/jupyterhub/issues/3700 .

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1129,7 +1129,7 @@ class JupyterHub(Application):
 
     @default('authenticator')
     def _authenticator_default(self):
-        return self.authenticator_class(parent=self, db=self.db)
+        return self.authenticator_class(parent=self, _deprecated_db_session=self.db)
 
     implicit_spawn_seconds = Float(
         0,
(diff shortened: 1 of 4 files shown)


manics (approved):
Tested with a custom spawner that accesses `self.db`:
```
[W 2022-05-05 19:04:34.297 JupyterHub spawner:166] The shared database session at Spawner.db is deprecated, and will be removed.
                Please manage your own database and connections.
    
                Contact JupyterHub at https://github.com/jupyterhub/jupyterhub/issues/3700
                if you have questions or ideas about direct database needs for your Spawner.
```
The subsequent lines are overly indented, though if that's the intention go ahead and merge!

minrk:
> The subsequent lines are overly indented

Good catch. Should have a `textwrap.dedent`, I think.

minrk:
@manics dedented

manics (approved):

PR #3886, opened 2022-05-05
Cleanup everything on API shutdown

minrk:
`app.stop` triggers full cleanup and stopping of the event loop

closes #3881

PR #3887, opened 2022-05-05
2.3 backports

minrk:
There's still more to do before custom scopes and scopes-on-tokens are ready for release (esp. #3877), but I think we've accumulated enough fixes that we should be making a release.

Could also include #3886 

This is a collection of backported PRs for 2.3 (basically everything, excluding the custom scopes / token scopes PRs).

Because this is a backport release, there are two tasks post-merge:

- [ ] forward-port changelog to main branch
- [ ] update some 'versionadded' notes from 2.3 to 2.4 for the not-backported features


minrk:
@consideRatio yup, making 2.3 now that #3886 is merged

minrk:
Oops, should have included #3837. Will do that along with #3918 and a couple other bugfixes for 2.3.1 pretty soon

PR #3889, opened 2022-05-05
admin: make user-info table selectable

johnkpark:
Screen Shot 2022-05-05 at 9 48 12 AM

`admin-table-head` class sets `user-select` to `none`, which makes text inside the table un-selectable. This is inconvenient when you need to copy some properties while debugging.


minrk:
@naatebarber @NarekA review?

naatebarber:
LGTM :+1:

jsx/src/components/ServerDashboard/ServerDashboard.jsx
```diff
@@ -1,5 +1,4 @@
 import React, { useState } from "react";
-import regeneratorRuntime from "regenerator-runtime";
```
NarekA, line 2: Are you sure this is safe to remove? I remember I had to add this import to fix an error, but I forget what it was.

jsx/src/components/ServerDashboard/ServerDashboard.jsx
```diff
@@ -200,6 +199,25 @@ const ServerDashboard = (props) => {
...
+        keyStyle={{
+          padding: "4px",
+        }}
+        valueStyle={{
+          padding: "4px",
+        }}
+        data={data}
+      />
+    );
+  };
```
NarekA, line 219: I see you have continued the effort to un-nest this module over time.

NarekA (approved):
Good job tracking down the cause!

jsx/src/components/ServerDashboard/ServerDashboard.jsx
johnkpark: The linter removed this for me. I think it's safe to remove as long as it's not used within this file.

naatebarber: Good catch, There's a chance it'll throw an error if that file uses async await syntax. I'd leave it in

naatebarber: Meant to whitelist that import in the .eslintrc

johnkpark: Interesting. I've added this back in.

johnkpark:
@minrk can we merge this?

minrk:
Thanks all!

yuvipanda:
@MeeseeksDev backport 2.x

consideRatio:
This PR is part of the 2.3.1 release

Issue #3890, opened 2022-05-09
Documentation: Example with custom /jhub/ location leads to 400 Bad Request with files/folders containing whitespace characters


bbrauns:
### Bug description

The apache reverse proxy documentation [1] has an example for running jupyterhub under the /jhub/ location. Unfortunately the line `RewriteRule /jhub/(.*) ws://127.0.0.1:8000/jhub/$1 [NE,P,L]` prevents opening folders or renaming files with a containing whitespace characters. It always leads to a 400 Bad Request error from apache. Removing the `NE` flag prevents this issue: `RewriteRule /jhub/(.*) ws://127.0.0.1:8000/jhub/$1 [P,L]`

[1] https://github.com/jupyterhub/jupyterhub/blob/main/docs/source/reference/config-proxy.md#apache

PR #3891, opened 2022-05-09
remove apache NE flag as it prevents opening folders and renaming fil…

bbrauns:
…es with whitespace characters. apache returns Bad Request

Closes https://github.com/jupyterhub/jupyterhub/issues/3890


minrk:
Thanks!

Issue #3897, opened 2022-05-11
Setting JupyterHub.admin_access does not change anything


tlehmann7:
### Bug description
By default, accessing other user Servers should be disabled. This is not the case. If I comment out the line in the config, so that it is set to False, its still possible to access user servers. Yes, I saved the config and restarted the Hub. Standard Installation without changes. 

#### Expected behaviour
Access to other servers is denied

#### Actual behaviour
Access is always possible if you are admin

### How to reproduce
try to set the setting in the config

### Your personal set up
Standard Installation using QuickStart. Ubuntu 20.04 LTS VM, 16 GB Ram, 8 vCPUs. Using default spawner and authenticator. Version 2.3.0. If you need more information, please ask.

<details><summary>Full environment</summary>

```
alembic==1.7.7
anyio==3.5.0
argon2-cffi==21.3.0
argon2-cffi-bindings==21.2.0
asttokens==2.0.5
async-generator==1.10
attrs==19.3.0
Automat==0.8.0
Babel==2.10.1
backcall==0.2.0
beautifulsoup4==4.11.1
bleach==5.0.0
blinker==1.4
certifi==2019.11.28
certipy==0.1.3
cffi==1.15.0
chardet==3.0.4
Click==7.0
cloud-init==22.1
colorama==0.4.3
command-not-found==0.3
configobj==5.0.6
constantly==15.1.0
cryptography==2.8
dbus-python==1.2.16
debugpy==1.6.0
decorator==5.1.1
defusedxml==0.7.1
distro==1.4.0
distro-info===0.23ubuntu1
entrypoints==0.3
executing==0.8.3
fastjsonschema==2.15.3
greenlet==1.1.2
httplib2==0.14.0
hyperlink==19.0.0
idna==2.8
importlib-metadata==4.11.3
importlib-resources==5.7.1
incremental==16.10.1
ipykernel==6.13.0
ipython==8.3.0
ipython-genutils==0.2.0
jedi==0.18.1
Jinja2==3.1.2
json5==0.9.8
jsonpatch==1.22
jsonpointer==2.0
jsonschema==3.2.0
jupyter-client==7.3.1
jupyter-core==4.10.0
jupyter-server==1.17.0
jupyter-telemetry==0.1.0
jupyterhub==2.3.0
jupyterlab==3.4.0
jupyterlab-pygments==0.2.2
jupyterlab-server==2.13.0
keyring==18.0.1
language-selector==0.1
[...]

Issue #3045, opened 2020-05-12
JupyterHub.admin_access parameter not transferred to singleuser-server when using LocalProcessSpawner


ghazpar:
### Bug description
Access not allowed by the NotebookApp when using either a service admin token or a user admin token, even after setting the hub's `admin_access` parameter (contrary to documentation). I observed this problem while using the `LocalProcessSpawner` (don't know if it applies to other spawners).

Depending on the kind of token, the access not allowed message is produced either here:
https://github.com/jupyterhub/jupyterhub/blob/72a38a599d3c623177fa0bb850ca47e5444f31ee/jupyterhub/services/auth.py#L866

Or here:
https://github.com/jupyterhub/jupyterhub/blob/72a38a599d3c623177fa0bb850ca47e5444f31ee/jupyterhub/services/auth.py#L883

I was able to bypass the problem by:
1. Setting an environment variable `JUPYTERHUB_ADMIN_ACCESS` in the hub's environment;
2. And adding the name of this variable to the `Spawner.env_keep` parameter.

The base spawner tries to do that here:
https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/spawner.py#L726

But it appears to also hardcode the `self.admin_access` to false here:
https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/spawner.py#L219

So the environment variable is never set.

Moreover, if using an environment variable is really the way to go (perhaps retrieving the hub's `admin_access` parameter value would be best?), then I believe the variable name `JUPYTERHUB_ADMIN_ACCESS` should also be added to the `Spawner.env_keep` base values here:
https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/spawner.py#L396

#### Expected behaviour
Access should be allowed according to documentation.

#### Actual behaviour
"Not allowing Hub service" or "Not allowing Hub user" message returned by the REST call.

### How to reproduce
Sending any REST request through the singleuser-server API.

### Your personal set up
Not relevant.

 - OS:
Not relevant.

 - Version:
 JupyterHub 1.1.0

 - Configuration:
 Basic JupyterHub configuration with the default `LocalProcessSpawner`.

Issue #3745, opened 2022-01-01
No links to access the users notebooks from the Admin menu


ykazakov:
### Bug description

Happy New Year!

I am trying to [enable admin access to user notebooks](https://jupyterhub.readthedocs.io/en/2.0.1/getting-started/authenticators-users-basics.html#give-admin-access-to-other-users-notebook-servers-admin-access):

```
JupyterHub.admin_access = True
```

#### Expected behaviour

As of Jupyterhub 1.5.0, I could access the users notebooks from the Admin menu:

<img width="610" alt="Screenshot 2022-01-01 at 23 04 24" src="https://user-images.githubusercontent.com/2140361/147861563-a4ecd2fe-558e-4021-b339-e3429b34fd87.png">

#### Actual behaviour

As of Jupyterhub 2.0.1, there is no link for accessing the users notebooks any longer:

<img width="689" alt="Screenshot 2022-01-01 at 23 05 54" src="https://user-images.githubusercontent.com/2140361/147861261-9b09a787-17b3-4328-89ae-9a7101b2eb84.png">

However, I could still access the user notebook by entering the url manually: `127.0.0.1:8000/user/zoe/`

### How to reproduce

1. Create the files with the following contents:
  - `Dockerfile`:
```
FROM jupyterhub/jupyterhub:2.0.1

RUN pip install jupyter jupyterlab --no-cache-dir \
    && useradd -m mal \
    && useradd -m zoe

COPY jupyterhub_config.py /etc/jupyterhub/

CMD jupyterhub -f /etc/jupyterhub/jupyterhub_config.py --debug
```
  - `jupyterhub_config.py`:
```
c.Authenticator.allowed_users = {'mal', 'zoe'}
c.Authenticator.admin_users = {'mal'}
c.JupyterHub.admin_access = True

c.JupyterHub.authenticator_class = 'dummy'
c.JupyterHub.spawner_class = 'simple'
c.Spawner.args = ["--allow-root"]
```
2. Create and run the docker container:
```
docker build -t jupyterhub-adminaccess . 
docker run -it -p 8000:8000 jupyterhub-adminaccess
```
3. Open the url: `http://127.0.0.1:8000/admin`
4. Enter Username: `mal` and press "Sign in"
5. Press "Start Server" next to `zoe`

### Your personal set up

- OS: MacOS 12.1
 - Version(s):
```
% docker --version    
Docker version 20.10.11, build dea9396
```

<details><summary>Logs</summary>

```
[...]

PR #3899, opened 2022-05-12
`admin_access` no longer works as it is overridden by RBAC scopes

manics:
`admin_access` no longer has any effect, see https://github.com/jupyterhub/jupyterhub/issues/3745#issuecomment-1038789932

The `admin_access` variable is still referenced elsewhere and I considered removing it, but I couldn't work out exactly how/if it's used so this is just a docs change for now.

Closes https://github.com/jupyterhub/jupyterhub/issues/3897
Closes https://github.com/jupyterhub/jupyterhub/issues/3045

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1317,11 +1317,14 @@ class JupyterHub(Application):
 
     admin_access = Bool(
         False,
-        help="""Grant admin users permission to access single-user servers.
+        help="""DEPRECATED since version 2.0.0.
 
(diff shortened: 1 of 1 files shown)


minrk:
Good catch, yes. We should definitely do another pass on the references to the deprecated value, and remove ~all of them. It's possible a few places should treat is as True (since full admins _can_ access servers), while others should be replaced with a direct permission check (e.g. admin page, which may be viewed with less than full admin permissions)

Issue #3144, opened 2020-08-14
Named Server names are not URL escaped


linhbngo:
### Bug description

A user created a named server with a name that contains a whitespace, this caused a malformed URL and subsequently crashed jupyterhub when it runs add_route()

#### Expected behaviour

JupyterHub is able to communicated with all spawned JupyterServer

#### Actual behaviour

Already started Jupyter servers remain running, but JupyterHub failed to restart. 

### How to reproduce

1. Launch a Jupyter server as normal. 
2. Launch an additional named server with white space in name. 

### Your personal set up

 - OS: CentOS 8.2
 
 - Version:
 
 - Configuration:

stv0g:
Thanks for @linhbngo for reporting this. I can reproduce this issue!

In our case the bug appeared when a user created a named server with a space in its name.
Interestingly, the bug did not cause a hard crash of the hub as long it was running.

Only when restarted, and the hub syncs its routes with the CHP, it became stuck in a crash-loop.

joe-eklund:
This is still happening and very easy to reproduce.

PR #3904, opened 2022-05-22
Escape named server name

manics:
This escapes the server name when used in paths, but keeps the original server name in the database.

Closes https://github.com/jupyterhub/jupyterhub/issues/3144

An alternative to this implementation is to require that server names are url-safe and error if they're not, i.e. the client UI would be responsible for converting the name submitted by a server to it's URL safe form before POSTing it to JupyterHub. However this leads to more complications since if the client encodes the name before submission (e.g. in javascript), then it'll be encoded a second time when it's sent in the URL request, and the JupyterHub database will always store an encoded name instead of the original.

- [x] This works in manual testing, ~but needs an automated test~ I've added some tests, though I couldn't work out how to trigger the `'/' in server_name` error in a test. I can trigger it in the browser, but if I add to the existing test it's accepted- is there a Mock getting in the way somewhere?

### DIFF
--- a/jupyterhub/handlers/pages.py
+++ b/jupyterhub/handlers/pages.py
@@ -118,9 +118,6 @@ class SpawnHandler(BaseHandler):
 
     @needs_scope("servers")
     async def _get(self, user_name, server_name):
-        if '/' in server_name:
-            raise web.HTTPError(400, f"Invalid server_name: {server_name}")
-
         for_user = user_name
 
         user = current_user = self.current_user
--- a/jupyterhub/handlers/pages.py
+++ b/jupyterhub/handlers/pages.py
@@ -14,7 +14,7 @@ from tornado.httputil import url_concat
 from .. import __version__
 from ..metrics import SERVER_POLL_DURATION_SECONDS, ServerPollStatus
 from ..scopes import needs_scope
-from ..utils import maybe_future, url_path_join
+from ..utils import maybe_future, url_escape_path, url_path_join
 from .base import BaseHandler
 
 
@@ -284,7 +284,10 @@ class SpawnHandler(BaseHandler):
         # which may get handled by the default server if they aren't ready yet
 
         pending_url = url_path_join(
-            self.hub.base_url, "spawn-pending", user.escaped_name, server_name
+            self.hub.base_url,
+            "spawn-pending",
+            user.escaped_name,
+            url_escape_path(server_name),
         )
 
         pending_url = self.append_query_parameters(pending_url, exclude=['next'])
@@ -353,6 +356,7 @@ class SpawnPendingHandler(BaseHandler):
         if server_name and server_name not in user.spawners:
             raise web.HTTPError(404, f"{user.name} has no such server {server_name}")
 
+        escaped_server_name = url_escape_path(server_name)
         spawner = user.spawners[server_name]
 
         if spawner.ready:
@@ -375,7 +379,7 @@ class SpawnPendingHandler(BaseHandler):
             exc = spawner._spawn_future.exception()
             self.log.error("Previous spawn for %s failed: %s", spawner._log_name, exc)
             spawn_url = url_path_join(
-                self.hub.base_url, "spawn", user.escaped_name, server_name
+                self.hub.base_url, "spawn", user.escaped_name, escaped_server_name
             )
             self.set_status(500)
             html = await self.render_template(
@@ -428,7 +432,7 @@ class SpawnPendingHandler(BaseHandler):
         # serving the expected page
         if status is not None:
             spawn_url = url_path_join(
-                self.hub.base_url, "spawn", user.escaped_name, server_name
+                self.hub.base_url, "spawn", user.escaped_name, escaped_server_name
             )
             html = await self.render_template(
                 "not_running.html",
--- a/jupyterhub/handlers/pages.py
+++ b/jupyterhub/handlers/pages.py
@@ -118,6 +118,9 @@ class SpawnHandler(BaseHandler):
 
     @needs_scope("servers")
     async def _get(self, user_name, server_name):
+        if '/' in server_name:
+            raise web.HTTPError(400, f"Invalid server_name: {server_name}")
+
         for_user = user_name
 
         user = current_user = self.current_user
--- a/jupyterhub/tests/test_named_servers.py
+++ b/jupyterhub/tests/test_named_servers.py
@@ -8,6 +8,7 @@ import pytest
 from requests.exceptions import HTTPError
 from tornado.httputil import url_concat
 
+from .. import orm
 from ..utils import url_escape_path, url_path_join
 from .mocking import FormSpawner, public_url
 from .test_api import TIMESTAMP, add_user, api_request, fill_user, normalize_user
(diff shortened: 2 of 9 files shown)


jupyterhub/utils.py
```diff
@@ -371,6 +372,11 @@ def compare_token(compare, token):
     return False
 
 
+def url_escape_path(value):
+    """Escape a value to be used in URLs, cookies, etc."""
+    return quote(value, safe='@~')
```
manics, line 377: This is taken from https://github.com/jupyterhub/jupyterhub/blob/b18a05c2c8a8ff86f9255bc2abaa356c418d3dd3/jupyterhub/user.py#L495-L497

jupyterhub/handlers/pages.py
```diff
@@ -118,6 +118,9 @@ def get(self, user_name=None, server_name=''):
 
     @needs_scope("servers")
     async def _get(self, user_name, server_name):
+        if '/' in server_name:
+            raise web.HTTPError(400, f"Invalid server_name: {server_name}")
```
minrk, line 122: This check is only on GET not on POST, which is why the test is passing. The test doesn't use the single-server GET request, it users the _full user_ (all servers) GET request. It might make sense to move it to the centralized `BaseHandler.spawn_single_user`, or on the User's `new_spawner`. Since there are quite a few places that would also need this check, but they all go through `spawn_single_user` eventually.

minrk (commented):
If we are escaping the URL, does it make sense to ban `/`? It should still work with the escaped path, right?

jupyterhub/tests/test_named_servers.py
```diff
@@ -121,7 +140,7 @@ async def test_create_named_server(app, named_servers):
                     'pending': None,
```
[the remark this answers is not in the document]
manics: You're right! I've updated the urls returned in API responses (not sure if I've found them all).

manics (commented):
> If we are escaping the URL, does it make sense to ban /? It should still work with the escaped path, right?

Ideally yes. Based on the logs I think JupyterHub now correctly handles it, but ConfigurableHttpProxy doesn't. If I disable the check:
```diff
diff --git a/jupyterhub/handlers/base.py b/jupyterhub/handlers/base.py
index c1685686..09ebe38f 100644
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -858,7 +858,7 @@ class BaseHandler(RequestHandler):
             if '/' in server_name:
                 error_message = f"Invalid server_name: {server_name}"
                 self.log.error(error_message)
-                raise web.HTTPError(400, error_message)
+                # raise web.HTTPError(400, error_message)
             user_server_name = f'{user.name}:{server_name}'
 
         if server_name in user.spawners and user.spawners[server_name].pending:
```
```
pytest -vs jupyterhub/tests/test_named_servers.py::test_create_invalid_named_server
```
```
[I 07:42.118 MockHub base:960] User walnut:a$/b took 1.126 seconds to start
[I 07:42.119 MockHub proxy:275] Adding user walnut to proxy /@/space%20word/user/walnut/a%24%2Fb/ => http://127.0.0.1:54521
[D 07:42.120 MockHub proxy:812] Proxy: Fetching POST http://127.0.0.1:8001/api/routes/@/space%20word/user/walnut/a%24%2Fb
23:07:42.125 [ConfigProxy] info: Adding route /@/space word/user/walnut/a$/b -> http://127.0.0.1:54521
23:07:42.127 [ConfigProxy] info: Route added /@/space word/user/walnut/a$/b -> http://127.0.0.1:54521
23:07:42.129 [ConfigProxy] info: 201 POST /api/routes/@/space%20word/user/walnut/a%24%2Fb 
[I 07:42.134 MockHub log:186] 201 POST /@/space%20word/hub/api/users/walnut/servers/a%24%2Fb (admin@127.0.0.1) 1232.41ms
```

We also prevent the use of `/` in usernames by default:
https://github.com/jupyterhub/jupyterhub/blob/9649a57e3457757d3a081c12fabb5fa6b712b67a/jupyterhub/auth.py#L253-L255
`validate_username()` can be overridden to allow it, but if you do then JupyterHub doesn't escape it:
```
[I 25:43.645 MockHub base:809] User logged in: wal/nut
[D 25:43.645 MockHub user:392] Creating <class 'jupyterhub.tests.mocking.MockSpawner'> for wal/nut:
[I 25:43.648 MockHub log:186] 302 POST /@/space%20word/hub/login -> /@/space%20word/hub/spawn (wal/nut@127.0.0.1) 16.87ms
[D 25:43.669 MockHub base:274] Recording first activity for <APIToken('7aae...', user='admin', client_id='jupyterhub')>
[D 25:43.672 MockHub base:274] Recording first activity for <User(admin 0/0 running)>
[D 25:43.680 MockHub base:319] Refreshing auth for admin
[W 25:43.685 MockHub log:186] 404 POST /@/space%20word/hub/api/users/wal/nut/servers/a%24%2Fb (admin@127.0.0.1) 20.22ms
```

jupyterhub/user.py
```diff
@@ -678,7 +686,8 @@ async def spawn(self, server_name='', options=None, handler=None):
             oauth_client = oauth_provider.add_client(
                 client_id,
                 api_token,
-                url_path_join(self.url, server_name, 'oauth_callback'),
+                url_path_join(self.url, url_escape_path(server_name), 'oauth_callback'),
+                # url_path_join(self.url, server_name, 'oauth_callback'),
```
minrk, line 690: looks like debug comment:

```suggestion
```

minrk:
Looks great! with a few minor edits

manics:
I think the root of the CHP problem is it fully decodes all API arguments even if they're not safe to use in a path:
https://github.com/jupyterhub/configurable-http-proxy/blob/d06de0f4a766d7ed8a198e6bfd448c83d3f46a3e/lib/configproxy.js#L641

This means CHP also can't handle `?` and `#` in servernames. Do you want me to also prohibit them on the JupyterHub side, or allow them (and `/`), and try and fix CHP instead?

minrk:
It's definitely a bug that CHP can't handle arbitrary escaped URLs. My hunch is that CHP shouldn't be escaping, but we'll need to make sure that it's handled the same on both sides. It's possible that is as simple as removing decode [here](https://github.com/jupyterhub/configurable-http-proxy/blob/d06de0f4a766d7ed8a198e6bfd448c83d3f46a3e/lib/configproxy.js#L385) and [here](https://github.com/jupyterhub/configurable-http-proxy/blob/d06de0f4a766d7ed8a198e6bfd448c83d3f46a3e/lib/configproxy.js#L641). One issue to deal with is _sometimes_ escaped chars like `@` should always be accepted.

For instance, adding a route for `a%2Fb` and a route for `a/b` should _not_ result in the same route, but I think it does right now.

I wish I'd declared server names to be url-safe (i.e. no escaping required), but implementing a transition for that is going to be a pain for folks who have picked names that need escaping in the meantime.

minrk:
Actually, now that I think of it - if names that need escaping have never worked, maybe the right thing to do is to require them to be url-safe, since that's an _implicit_ requirement now as any unsafe names have never worked?

minrk:
ok, if it's not broken, I don't want to break things by introducing strictness late in the game (at least until a major release). Let's go with what we have here then.

Thanks!

PR #3906, opened 2022-05-25
Force add existing certificates

fabianbaier:
When persisting certificates in e.g. a PV hub restarts throw a `certipy.certipy.CertExistsError` complaining the certificate exists already, instead of overwriting it and using the "old" one.

I think related issues have been around for a while (see e.g. https://discourse.jupyter.org/t/unable-to-combine-recreate-internal-ssl-false-with-external-ssl-authorities/10897)


minrk:
Thanks!

yuvipanda:
@MeeseeksDev backport 2.x

PR #3908, opened 2022-05-25
allow auth_state_hook to halt spawn

### LINKED ISSUE #3134: Spawner should stop on failed auth_state_hook
### Bug description
When a custom `auth_state_hook` is implemented for a spawner, and when the hook is executed, if an exception is raised, then spawning should stop. Currently if an exception is raised, it is logged, ignored, and the spawning continues. It seems to be due to [this part of the code](https://github.com/jupyterhub/jupyterhub/blob/b11c02c6e01531702e99d60317d5691999a0a4a6/jupyterhub/spawner.py#L1021)

#### Expected behaviour
Spawning should stop if the `auth_state_hook` raises an exception.

#### Actual behaviour
The spawning continues.

### How to reproduce
``` python
def my_auth_state_hook(spawner, auth_state):
    raise Exception()
```
Then in config set:
``` python
spawner.auth_state_hook = my_auth_state_hook
```

### Your personal set up

 - OS:
Linux
 - Version:
1.1.0

minrk:
hooks prior to start should raise and stop the whole thing

only hooks during cleanup need to be passed over

closes #3134

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -1142,10 +1142,7 @@ class Spawner(LoggingConfigurable):
     async def run_auth_state_hook(self, auth_state):
         """Run the auth_state_hook if defined"""
         if self.auth_state_hook is not None:
-            try:
-                await maybe_future(self.auth_state_hook(self, auth_state))
-            except Exception:
-                self.log.exception("auth_state_hook failed with exception: %s", self)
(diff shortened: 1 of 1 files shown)


consideRatio (approved):

consideRatio:
I marked this as a bugfix, I think that makes the most sense. This is the expected behavior I'd say.

PR #3909, opened 2022-05-25
include stopped servers in user model

### LINKED ISSUE #3901: Admin page not displaying named servers when they are stopped
### Bug description

The admin page only displays named servers that are actively running. Stopped servers are not displayed at all. This is a regression when compared to JupyterHub 1.x's functionality.

#### Expected behaviour

I would expect the admin page to list all servers on the hub, including named servers, regardless of whether they are stopped or running.

#### Actual behaviour

The admin page displays each user's default server regardless of status (stopped or running), but only displays named servers if they are running.

### How to reproduce

1. Configure JupyterHub to use named servers.
2. Create a named server and run it.
3. Log in as a different (admin) account and navigate to the admin page. If the named server is still running, you'll see it listed,
4. Shut down the named server. It will disappear once it is stopped.

### Your personal set up

I am running JupyterHub 2.3.0, installed from PIP. Using Python 3.9 (anaconda). The operating system is macOS 10.15.7.

### LINKED ISSUE #3981: Add "Last seen" column in admin page
### Proposed change

I just updated jupyterhub. There used to be a Last seen column in the admin panel, and this was a feature I used a lot. Now there is a "Last activity" column instead, which show "Never" if no server is running. As I use the culling system anyway, this column is rarely of any use. 

I would like to have the "Last seen" column back. 

I'm not fluent enough with the jupyterhub project to test a solution. I understand that the file to modify is probably this one https://github.com/jupyterhub/jupyterhub/blob/main/jsx/src/components/ServerDashboard/ServerDashboard.jsx around the line 261. Maybe switching "server.last_activity" to "user.last_activity" would do ? I couldn't find out how to test this.

minrk:
closes #3901 
closes #3995
closes #3981 

The user-list API doesn't list stopped servers, so there's no way to tell if e.g. a named server _doesn't exist_ or just isn't running.

We could just include this by default. I'm not sure what the performance cost would be. We could also make this a server-side config or a whole new endpoint to make it opt-in, but less tedious.

### DIFF
--- a/docs/source/_static/rest-api.yml
+++ b/docs/source/_static/rest-api.yml
@@ -139,6 +139,16 @@ paths:
             If unspecified, use api_page_default_limit.
           schema:
             type: number
+        - name: include_stopped_servers
+          in: query
+          description: |
+            Include stopped servers in user model(s).
+            Added in JupyterHub 3.0.
+            Allows retrieval of information about stopped servers,
+            such as activity and state fields.
+          schema:
+            type: boolean
+          allowEmptyValue: true
       responses:
         200:
           description: The Hub's user list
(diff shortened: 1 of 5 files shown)


consideRatio (approved):
Nice this LGTM!

PR #3910, opened 2022-05-25
use equality to filter token prefixes

### LINKED ISSUE #3070: performance issue about auth
### Bug description
There may be performance issue when auth.
in the method of get_current_user_token(), I find some code like 
`db.query(cls).filter(bindparam('prefix', prefix).startswith(cls.prefix))`
corresponding sql as follow 
`select ... FROM api_tokens WHERE ('0ee2' LIKE concat(api_tokens.prefix, '%'))`
explain this sql,
```
select_type: SIMPLE
type:ALL
```
it doesn't use the index api_tokens.ix_api_tokens_prefix.
So when this table has ten thousands of rows, it costs about 1000+ms. it may have performance issue.

#### Expected behaviour
this sql should use the index of api_tokens.ix_api_tokens_prefix.

#### Actual behaviour
Can we change it as follow
`select ... FROM api_tokens WHERE ('0ee2' = api_tokens.prefix)`

### How to reproduce
api_tokens has ten thousands of rows. All requests to jupyterhub always costs N * 1000ms, handred times than beginning.

### Your personal set up
jupyterhub 1.1.0

minrk:
otherwise, index isn't used (closes #3070)

note: this means changing the token prefix size requires revoking all tokens, where before only _increasing_ the token prefix size required doing that.

I ran a test with `sqlite:///:memory:` showing clear performance consequences:

visualization

where using the index, lookup stays around 150µs, it rapidly increased beyond 1ms per token lookup with the current scheme if you have at least ~10k tokens.

### DIFF
--- a/jupyterhub/orm.py
+++ b/jupyterhub/orm.py
@@ -528,9 +528,7 @@ class Hashed(Expiring):
         prefix = token[: cls.prefix_length]
         # since we can't filter on hashed values, filter on prefix
         # so we aren't comparing with all tokens
-        prefix_match = db.query(cls).filter(
-            bindparam('prefix', prefix).startswith(cls.prefix)
-        )
+        prefix_match = db.query(cls).filter_by(prefix=prefix)
         prefix_match = prefix_match.filter(
             or_(cls.expires_at == None, cls.expires_at >= cls.now())
         )


yuvipanda:
@MeeseeksDev backport 2.x

consideRatio:
This PR is part of the 2.3.1 release

PR #3914, opened 2022-05-29
Build admin app in setup.py

manics:
This removes the need for a manual step to build the admin app, and instead makes it part of setup.py alongside the existing NPM/CSS tasks.


setup.py
```diff
@@ -250,12 +263,74 @@ def run(self):
...
+    def should_run(self):
+        if os.getenv('READTHEDOCS'):
+            # yarn not available on RTD
+            return False
+
+        if not os.path.exists(self.js_target):
+            return True
+
+        js_target_mtime = mtime(self.js_target)
+        # The sdist doesn't contain the jsx directory so this always returns 0
```
minrk, line 281: I think this means we need have a test for this in release.yml. The sdist definitely should contain the `jsx` sources. I think this means MANIFEST.in needs

```
graft jsx
prune jsx/build
prune jsx/node_modules
```

manics: Added, including a test in release.yml.

minrk:
Awesome, thanks!

PR #3915, opened 2022-05-29
Update Contributing documentation

manics:
Adds yarn, moves most of CONTRIBUTING.md into https://jupyterhub.readthedocs.io/en/stable/contributing/index.html to reduce duplication

Includes the `python3 setup.py jsx` command added in https://github.com/jupyterhub/jupyterhub/pull/3914


minrk:
Nice!

PR #3918, opened 2022-05-31
set default_url via config

minrk:
avoids accidental overrides of `@default('default_url')` in subclasses, e.g. SingleUserLabApp by ensuring the env is loaded as _config_ rather than a _default_ in a base class.


yuvipanda:
@MeeseeksDev backport 2.x

PR #3919, opened 2022-06-01
ensure custom template is loaded with jupyter-server notebook extension

minrk:
our patches to page.html didn't affect nbclassic run with jupyter-srver, which gets its own jinja environment, due to how extensions are set up.

regression test included.

We could _only_ patch `notebook_jinja2_env` when we find it, instead of _also_ patching the base `jinja2_env`.

Had to shift template-patch to later in initialize, because the extension jinja env doesn't exist until extensions are loaded, which is after init_webapp.


yuvipanda:
@MeeseeksDev backport 2.x

PR #3921, opened 2022-06-02
pages.py: Remove unreachable code

manics:
Removes some unreachable code- the preceding `try` and `except` clauses both contain `return` statements.

### DIFF
--- a/jupyterhub/handlers/pages.py
+++ b/jupyterhub/handlers/pages.py
@@ -268,15 +268,6 @@ class SpawnHandler(BaseHandler):
             )
             self.finish(form)
             return
-        if current_user is user:
-            self.set_login_cookie(user)
-        next_url = self.get_next_url(
-            user,
-            default=url_path_join(
-                self.hub.base_url, "spawn-pending", user.escaped_name, server_name
-            ),
-        )
-        self.redirect(next_url)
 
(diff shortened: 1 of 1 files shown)


consideRatio:
Hmmm should the code be removed or be made reachable? Removing it is an improvement of sorts, but would hide the traces of a possible logic bug we may need to fix - code that is unreachable had a purpose originally I presume.

Do you think the right call is to remove this, or try make it reachable? I'm not confident about either atm.

manics:
Good question! That why I opened a standalone PR instead of including it in another one.

manics:
This is the commit that added a `return` to the `try` clause, making this code unreachable:
https://github.com/jupyterhub/jupyterhub/commit/3908c6d041987e69db7150dcf2041916053b863d

It's obvious why it wasn't noticed in the PR review- the return in the catch block is outside the displayed diff. do you know if there's a pre-commit hook we can use to catch unreachable code in future?

minrk:
> there's a pre-commit hook we can use to catch unreachable code in future?

I believe [pylint can find unreachable code](https://pylint.pycqa.org/en/latest/user_guide/messages/warning/unreachable.html). Not sure if any of the many checkers we already use do.

PR #3932, opened 2022-06-07
Fixes #3913 db uri parsing issue

### LINKED ISSUE #3913: sqlite DB URI error check shows incorrect path
When jupyterhub is configured with a sqlite DB URI (e.g. on Google Cloud's Dataproc Hub), if the sqlite DB connection fails, jupyterhub tries to detect DB path errors. The code below does a split on `:///`, which removes the root `/` character, so the check will incorrectly look for a relative path.

https://github.com/jupyterhub/jupyterhub/blob/9649a57e3457757d3a081c12fabb5fa6b712b67a/jupyterhub/app.py#L1801

I don't have a screenshot at the moment but this is an easy fix: split by `://` instead of `:///`, so that the correct absolute path is checked instead of an inexistent relative path.

stephenb101010:
Fixes jupyterhub/jupyterhub#3913

This change updates the parsing of sqlite URIs so that the root does not get parsed out.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1798,7 +1798,7 @@ class JupyterHub(Application):
             self.log.error("Failed to connect to db: %s", db_log_url)
             self.log.debug("Database error was:", exc_info=True)
             if self.db_url.startswith('sqlite:///'):
-                self._check_db_path(self.db_url.split(':///', 1)[1])
+                self._check_db_path(self.db_url.split('://', 1)[1])
             self.log.critical(
                 '\n'.join(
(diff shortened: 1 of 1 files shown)


minrk:
Thanks for the PR!

As mentioned in #3913, the parsing is correct. The [format for sqlite urls](https://docs.sqlalchemy.org/en/20/dialects/sqlite.html#connect-strings) is `sqlite:///[path]`, so 3 slashes for a relative path, 4 for absolute (except Windows, where absolute paths don't start with `/`).

PR #3937, opened 2022-06-08
switch to importlib_metadata for entrypoints

minrk:
standalone entrypoints package is deprecated now that similar functionality is in the stdlib

need importlib_metadata >= 3.6 backport on Python < 3.10

### DIFF
--- a/jupyterhub/traitlets.py
+++ b/jupyterhub/traitlets.py
@@ -3,7 +3,14 @@ Traitlets that are used in JupyterHub
 """
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
-import entrypoints
+import sys
+
+# See compatibility note on `group` keyword in https://docs.python.org/3/library/importlib.metadata.html#entry-points
+if sys.version_info < (3, 10):
+    from importlib_metadata import entry_points
+else:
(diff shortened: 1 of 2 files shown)


minrk:
cf https://github.com/jupyter/nbconvert/pull/1782

jupyterhub/traitlets.py
```diff
@@ -125,23 +132,23 @@ def help(self):
...
-        # make it case-insensitive
-        return {key.lower(): value for key, value in group.items()}
+        """Load my entry point group
+
+        Returns a dict whose keys are lowercase entrypoint names
+        """
+        group = entry_points(group=self.entry_point_group)
+        return {
+            entry_point.name.lower(): entry_point
+            for entry_point in entry_points(group=self.entry_point_group)
```
consideRatio, line 150:
```suggestion
            for entry_point in group
```

consideRatio: Maybe the name `group` could be `entry_points` or `group_entry_points` or similar.

consideRatio, line 150:
```suggestion
        entry_points = entry_points(group=self.entry_point_group)
        return {
            entry_point.name.lower(): entry_point
            for entry_point in entry_points
```

PR #3958, opened 2022-06-24
add correct autocomplete fields for login form

minrk:
This might help browsers do the right thing more often

In my experience, they already figure it out by the input types, but this may help it be right more often.


consideRatio:
Googling and finding mentions of how autocomplete can make sense to use at https://web.dev/sign-in-form-best-practices/#autofill, I think this makes sense!

PR #3960, opened 2022-06-28
Fix GET /api/proxy with pagination

cqzlxl:
It's an obvious bug.


minrk:
Added regression test to ensure this branch is covered. Thanks!

PR #3962, opened 2022-07-01
Add ConfigurableHTTPProxy.log_level

manics:
ConfigurableHTTPProxy has a debug flag that switches between `debug` and `info`, but there's no way to reduce the log-level beyond that.
Requested in https://discourse.jupyter.org/t/jupyterhub-log-level-not-working-as-expected/9961

### DIFF
--- a/jupyterhub/proxy.py
+++ b/jupyterhub/proxy.py
@@ -32,6 +32,7 @@ from traitlets import (
     Any,
     Bool,
     Dict,
+    Enum,
     Instance,
     Integer,
     TraitError,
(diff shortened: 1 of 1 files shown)


jupyterhub/proxy.py
```diff
@@ -466,7 +476,21 @@ def _default_semaphore(self):
     def _concurrency_changed(self, change):
         self.semaphore = asyncio.BoundedSemaphore(change.new)
 
+    # https://github.com/jupyterhub/configurable-http-proxy/blob/4.5.1/bin/configurable-http-proxy#L92
+    log_level = Enum(
+        ["debug", "info", "warn", "error", "DEBUG", "INFO", "WARN", "ERROR"],
```
manics, line 481: https://github.com/jupyterhub/configurable-http-proxy/blob/4.5.1/bin/configurable-http-proxy#L92 says `debug, info, warn, error` but upper-case (as used by Python logging) also seems to work.

minrk: Can use [CaselessStrEnum](https://traitlets.readthedocs.io/en/stable/trait_types.html#traitlets.CaselessStrEnum) to avoid repetition

manics: Thanks, I'm constantly amazed by all the hidden features in Traitlets :smiley:

minrk:
Thanks!

PR #3967, opened 2022-07-08
validate proxy.extra_routes

minrk:
- add trailing slash if missing, and warn (as we do with url prefixes elsewhere), since this one is unambiguous
- raise if leading slash is wrong (must not be present with host routing, must be present otherwise). Raises rather than fixing automatically like trailing slash because it is ambiguous whether the intention is host-based routing or prefix routing.
- Lightweight check for probably-urls (don't need to be strict here, just catch likely errors, such as omitting host/protocol entirely)

Came up in https://discourse.jupyter.org/t/oauth-flow-in-jupyterhub-on-z2jh/14779/8

### DIFF
--- a/examples/server-api/jupyterhub_config.py
+++ b/examples/server-api/jupyterhub_config.py
@@ -2,6 +2,8 @@
 # 1. start/stop servers, and
 # 2. access the server API
 
+c = get_config()  # noqa
+
 c.JupyterHub.load_roles = [
     {
         "name": "launcher",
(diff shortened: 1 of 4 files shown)


consideRatio (approved):

consideRatio (changes requested):
@minrk I see that this includes an actual test failure of relevance.

```
E                   ValueError: Proxy.extra_routes missing host component in /prefix/ (must not have leading '/') when using `JupyterHub.subdomain_host = 'http://localhost.jovyan.org:8000/'`
```

consideRatio (approved):

minrk:
@consideRatio thanks for reviewiing! Sorry I missed the test failure for a while.

PR #3968, opened 2022-07-08
pin nbclassic in dev requirements

minrk:
ensures `/tree/` handler is present.

There are no longer any authenticated non-API URLs served by both `notebook` and bare `jupyter-server + jupyterlab`. This is causing 404s in tests when no handler is registered for `/tree/`

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -9,7 +9,11 @@ cryptography
 html5lib  # needed for beautifulsoup
 jupyterlab >=3
 mock
-nbclassic  # ensures /tree/ handler is present as assumed by tests
+# nbclassic provides the '/tree/' handler, which we use in tests
+# it is a transitive dependency via jupyterlab,
+# but depend on it directly
+# 0.4 introduces some installation/loading problems (check again after 0.4.3)
+nbclassic<0.4
 pre-commit
 pytest>=3.3
 pytest-asyncio; python_version < "3.7"
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -9,6 +9,7 @@ cryptography
 html5lib  # needed for beautifulsoup
 jupyterlab >=3
 mock
+nbclassic
 pre-commit
 pytest>=3.3
 pytest-asyncio; python_version < "3.7"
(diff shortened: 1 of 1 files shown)


dev-requirements.txt
```diff
@@ -9,6 +9,7 @@ cryptography
 html5lib  # needed for beautifulsoup
 jupyterlab >=3
 mock
+nbclassic
```
consideRatio, line 12: Does this make sense? If not maybe you can tweak the comment to make sense or so. If nbclassic isn't directly imported as a dependency in tests, I think a comment is very helpful to have to clarify the situation.

```suggestion
nbclassic  # ensures /tree/ handler is present as assumed by tests
```

minrk:
This didn't appear to work. I'm not sure what's causing the 404 on `/tree/`. It shows `notebook_shim` as being loaded in the failing tests, so maybe it's a bug there?

minrk:
Apparently it's a bug in 0.4.x that nbclassic [doesn't enable itself on install](https://github.com/jupyter/nbclassic/pull/127). An explicit enable should do it, instead.

dev-requirements.txt
```diff
@@ -9,6 +9,7 @@ cryptography
 html5lib  # needed for beautifulsoup
 jupyterlab >=3
 mock
+nbclassic  # ensures /tree/ handler is present as assumed by tests
```
minrk, line 12: nbclassic is already a transitive dependency via jupyterlab, but let's be explicit since we call it directly and jupyterlab will probably drop the dependency someday

.github/workflows/test.yml
```diff
@@ -84,7 +84,7 @@ jobs:
           - python: "3.8"
             db: postgres
           - python: "3.8"
-            nbclassic: nbclassic
+            legacy_notebook: legacy_notebook
```
minrk, line 87: `nbclassic` is a bit confusing to call the matrix entry where we explicitly do _not_ use the package `nbclassic`, it's the legacy package `notebook`.

minrk:
Turns out 0.4 has various issues, so pinning <0.4 for now to get CI passing again

PR #3969, opened 2022-07-10
Fix disabling of individual page template announcements

### LINKED ISSUE #3965: `announcement_login` etc can't be set falsy to override `announcement`
My expectation was that I could set the `c.JupyterHub.template_vars` named `announcement_login` explicitly to a falsy value to disable an announcement for a specific page. It seems that the code is that only if `announcement_login` is Truthy, its value will override `announcement`.

The value of being able to do so, is to easily configure `announcement` in one place and set `announcement_login` to something falsy to disable it - practically declaring an annoncement in a single place for all authenticated pages.

https://github.com/jupyterhub/jupyterhub/blob/f22203f50e277b0aae89b079f59b24591547e914/share/jupyterhub/templates/login.html#L1-L4

If possible in a jinja template, I suggest the change like this, even though I assume it must be tweaked to match Jinja syntax.

```diff
  {% extends "page.html" %} 
- {% if announcement_login %} 
+ {% if announcement_login is not None %} 
    {% set announcement = announcement_login %} 
  {% endif %} 
```

consideRatio:
Closes #3965.

This could be written as `is not none` as well, but I figured there is no default value for `announcement_<page name>` in template_vars, this made sense as well, allowing people to declare a None value as well as a blank string, and not just a blank string, to disable an individual announcement.

## Final behavior

Setting template variables like:

```python
c.JupyterHub.template_vars = {
    "announcement": "something",
    "announcement_spawn": "something else",
    "announcement_login": "",
    # Setting a specific page's announcement variable to None doesn't do anything, but
    # it can be useful if merging template_vars dictionaries to unset a value.
    "announcement_home": None,
}
```

will cause the announcement `something` show up on all pages expect on the spawn page where it sais `something else` and on the login page where no announcement it shown.

## Related unchanged logic

https://github.com/jupyterhub/jupyterhub/blob/ab776e3989bffe9e1a9d0744c96c5f8e8d876988/share/jupyterhub/templates/page.html#L168-L174

### DIFF
--- a/share/jupyterhub/templates/home.html
+++ b/share/jupyterhub/templates/home.html
@@ -1,5 +1,5 @@
 {% extends "page.html" %}
-{% if announcement_home is defined %}
+{% if announcement_home is string %}
   {% set announcement = announcement_home %}
 {% endif %}
 
--- a/share/jupyterhub/templates/home.html
+++ b/share/jupyterhub/templates/home.html
@@ -1,5 +1,5 @@
 {% extends "page.html" %}
-{% if announcement_home %}
+{% if announcement_home is defined %}
   {% set announcement = announcement_home %}
 {% endif %}
 
--- a/share/jupyterhub/templates/login.html
+++ b/share/jupyterhub/templates/login.html
@@ -1,5 +1,5 @@
 {% extends "page.html" %}
-{% if announcement_login %}
+{% if announcement_login is defined %}
   {% set announcement = announcement_login %}
 {% endif %}
 
--- a/share/jupyterhub/templates/login.html
+++ b/share/jupyterhub/templates/login.html
@@ -1,5 +1,5 @@
 {% extends "page.html" %}
-{% if announcement_login is defined %}
+{% if announcement_login is string %}
   {% set announcement = announcement_login %}
 {% endif %}
 
--- a/share/jupyterhub/templates/logout.html
+++ b/share/jupyterhub/templates/logout.html
@@ -1,5 +1,5 @@
 {% extends "page.html" %}
-{% if announcement_logout %}
+{% if announcement_logout is defined %}
   {% set announcement = announcement_logout %}
 {% endif %}
 
--- a/share/jupyterhub/templates/logout.html
+++ b/share/jupyterhub/templates/logout.html
@@ -1,5 +1,5 @@
 {% extends "page.html" %}
-{% if announcement_logout is defined %}
+{% if announcement_logout is string %}
   {% set announcement = announcement_logout %}
 {% endif %}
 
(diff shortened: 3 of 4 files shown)


manics:
If someone sets it to `None` does tornado treat it as an empty string?

consideRatio:
> If someone sets it to None does tornado treat it as an empty string?

Hmmm... Is tornado aware about the jinja2 template rendering part before it serves the rendered HTML?

I think this happens:

1. template_vars are set
2. jinja2 templates are rendered based on template_vars passed
3. tornado servers the rendered html

minrk:
Yeah, tornado's not involved in the template rendering.

Maybe the right test is [`if x is string`](https://jinja.palletsprojects.com/en/3.0.x/templates/#jinja-tests.string) which will be True if it's any string. Then any non-string (e.g. None or undefined) will not take the branch.

Another option:

- `x is defined and x is not none` (explicitly treat undefined and None as the same)
- `x | default(none) is none` (same as above, style choice)

The only case I can imagine `is defined` alone having trouble is if some intermediate 'base' config (e.g. a helm chart) sets one of these and a user deployment wants to _un_set it. Setting back to None wouldn't have the same effect for `is defined`, and there is no 'set to unset' in merged config or helm values, etc. Non-strings like None don't really make sense to take the other branch anyway, so treating `None` the same as undefined makes sense.

share/jupyterhub/templates/home.html
```diff
@@ -1,5 +1,5 @@
 {% extends "page.html" %}
-{% if announcement_home %}
+{% if announcement_home is defined %}
```
minrk, line 2:
```suggestion
{% if announcement_home is string %}
```

share/jupyterhub/templates/login.html
```diff
@@ -1,5 +1,5 @@
 {% extends "page.html" %}
-{% if announcement_login %}
+{% if announcement_login is defined %}
```
minrk, line 2:
```suggestion
{% if announcement_login is string %}
```

share/jupyterhub/templates/logout.html
```diff
@@ -1,5 +1,5 @@
 {% extends "page.html" %}
-{% if announcement_logout %}
+{% if announcement_logout is defined %}
```
minrk, line 2:
```suggestion
{% if announcement_logout is string %}
```

share/jupyterhub/templates/spawn.html
```diff
@@ -1,5 +1,5 @@
 {% extends "page.html" %}
-{% if announcement_spawn %}
+{% if announcement_spawn is defined %}
```
minrk, line 2:
```suggestion
{% if announcement_spawn is string %}
```

consideRatio:
@minrk I've updated the original PR description to reflect the state of the PR as it is now.

manics (approved):

PR #3972, opened 2022-07-12
Test against python 3.10

consideRatio:
Related to #3973 - if a decision is made about that, I'll keep moving on fixing things based on findings here and then finally marking this PR as non-draft when tests pass.


minrk:
We should definitely add 3.10 and 3.11-beta, but we also _definitely_ shouldn't stop testing supported Python versions (back to 3.6, but should probably bump to 3.7 soon).

consideRatio:
Absolutely, we should test all versions. In this state, i wanted to ensure i caught all kinds of test errors if poython 3.10 is used before letting it run only for one test variation.

PR #3974, opened 2022-07-14
Avoid deprecated 'IOLoop.current' method

minrk:
Deprecated in tornado 6.2, matching asyncio's deprecation of get_current_loop (i.e. there is no longer any such thing as a 'current' loop that isn't running). Instead, only access running loop from inside coroutines.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -3292,16 +3292,19 @@ class JupyterHub(Application):
     def launch_instance(cls, argv=None):
         self = cls.instance()
         self._init_asyncio_patch()
-        loop = IOLoop.current()
-        task = asyncio.ensure_future(self.launch_instance_async(argv))
+        loop = IOLoop(make_current=False)
+
+        try:
+            loop.run_sync(self.launch_instance_async, argv)
+        except Exception:
+            loop.close()
+            raise
+
         try:
             loop.start()
         except KeyboardInterrupt:
             print("\nInterrupted")
         finally:
-            if task.done():
-                # re-raise exceptions in launch_instance_async
-                task.result()
             loop.stop()
             loop.close()
 
--- a/jupyterhub/tests/conftest.py
+++ b/jupyterhub/tests/conftest.py
@@ -132,16 +132,13 @@ def event_loop(request):
 
 
 @fixture(scope='module')
-def io_loop(event_loop, request):
+async def io_loop(event_loop, request):
     """Same as pytest-tornado.io_loop, but re-scoped to module-level"""
-    ioloop.IOLoop.configure(AsyncIOMainLoop)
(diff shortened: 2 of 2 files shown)


consideRatio (commented):
Hmmm I looked a while at this but struggled to come up with a clear understanding that this does or doesn't make sense.

@manics is this something you grasp well enough to review?

manics:
I'm afraid I don't understand it

minrk:
Had a chat with @consideRatio to go over this. It is a general change to asyncio (since 3.10), which has propagated to tornado (6.2) that it has proven problematic to allow accessing the 'current' event loop while it's not running, because what's 'current' can change (the problem largely arose with the introduction of `asyncio.run` which always creates a new event loop, breaking many applications that set up the loop, then started it with `asyncio.get_event_loop().run_forever()` which doesn't work with `asyncio.run`).

The high-level changes:

- `asyncio.get_event_loop()` (and related methods) and `IOLoop.current()` (and related methods) are deprecated _if_ called __when the event loop is not running__.
- If called when the loop is running, they are equivalent to `asyncio.get_running_loop()` (which raises RuntimeError if called outside the loop). Calling in this context is _not_ deprecated.
- Behavior is totally unchanged, only deprecated, so this is a relatively low-pressure change.
- Hooking up things to an event loop that isn't running is still fine, but it's now the _application's_ responsibility (i.e. JupyterHub and/or the test runner) to maintain the handle on the not-yet-running loop (`self.io_loop`), rather than calling e.g. `IOLoop.current()`

The gist of the required changes, in practice (applies to all asyncio/tornado applications, not just JupyterHub):

1. for access to the loop while it's not running, maintain and pass around our own handles (e.g. `self.io_loop`), removing explicit calls to managing the underlying 'current' event loop (e.g. `IOLoop.make_current()`, `asyncio.set_event_loop()`, etc.). In tests, these handles already exist and are the `event_loop` and `io_loop` fixtures.
2. where possible, call functions that reference the current loop from inside the loop (this is already true of ~everything except for sync test functions) because it's not always obvious when calls to `IOLoop.current()` may be made
4. For the most part, only _applications_ are affected (i.e. JupyterHub, not Spawners or Authenticators) because only _applications_ are responsible for creating/starting event loops. __Exception: if they start background threads which manage their own loops__
5. Test suites are technically separate applications, since they, too, start/stop event loops, usually in a different way from the actual application. The simplest way to ensure an event loop is running is to make a test or fixture `async` even if it has no awaits.

consideRatio (approved):
Thank you for your help explaining complicated topics @minrk!!! The video chat and the comment above (https://github.com/jupyterhub/jupyterhub/pull/3974#issuecomment-1199061543) helped me understand this a lot better!

This LGTM, lets go for a merge!

PR #3975, opened 2022-07-14
Bump Dockerfile base image to 22.04

minrk:


Dockerfile
```diff
@@ -21,7 +21,7 @@
 # your jupyterhub_config.py will be added automatically
 # from your docker directory.
 
-ARG BASE_IMAGE=ubuntu:focal-20200729
+ARG BASE_IMAGE=ubuntu:jammy-20220531
```
consideRatio, line 24: What do you think about omitting the date? I dont think this risks breaking things to leave unpinned, but could help us avoid very old versions with lots of known vulns etc.

Consider for example or focal version was two years old! Lets leave this unpinned - and maybe using the name 22.04 because its easier to know what it is compared to its code name.

```suggestion
ARG BASE_IMAGE=ubuntu:22.04
```

manics: How about if we keep the date and configure dependabot to bump it?

consideRatio: Is there value to that? Isnt this rebuilt with tags only anyhow - no master-branch releases right?

I figure it is helpful for requirements.txt ro see how things change, but the build date of ubuntu core image seems less relevant to track in version control.

I think that adds some overhead that isnt justified atm.

consideRatio: The key point is if there is value to doing something. @manics can you highlight the value provided by pinning a ubuntu build and bumping with dependabot based on date-formatted tags?

Im fine with added complexity and maintenance given that its has value to justify it - but does this has value?

manics: I'm personally not too fussed about whether we use the full version with date or just the `22.04` short tag. However there's a discussion in https://github.com/jupyterhub/team-compass/issues/502 about reproducible builds. If we're going that way then having reproducible (as much as possible) container builds seems a natural part of that.

consideRatio: Hmmm... I find that to be a different topic or aspect to some degree. That was about python package signatures, this regards docker images that when built determines their signature no matter what if we pin the base image used and have a frozen requirements.txt file etc.

Even though I suggest printing a hash as part of a CI build for a trace of what was built could be nice, I find this proposal would be adding significantly more work to setup and keep going long term than adding a print line.

Work:
- Making dependabot bump from `22.04-<date-here>` to `22.04-<more-recent-date-here>` automatically and reliably - maybe it works directly, but I'm skeptical dependabot recognizes the `22.04-YYYY0M0D` pattern and knows how to handle it.
- Merging those PRs regularly

I'm fine with any further action taken if someone wants to take action though, discussing this in depth seems like more work than going for whatever - but I do think its relevant to not lock us into a very old base image that won't get updated as that makes even new versions of this image end up with very old known vulnerabilities.

minrk: Given that we haven't bumped this tag in 2 years, I think removing the date will be better in practice. We can investigate more strictly reproducible builds perhaps as part of moving the images out of this repo.

consideRatio: Let's go for a commit on this code suggestion and a merge.

If dependabot bumping with pinned `22.04-<date>` bumping PRs are setup, that is also acceptable to me. My main concern is it isn't reliable, not that we press merge on reliably bumped PRs.

consideRatio: Marking this discussion as unresolved to increase the visibility of this discussion.

consideRatio (approved):

PR #3976, opened 2022-07-14
Require Python 3.7

minrk:
- Bumps CI Python versions to include 3.10 and 3.11 (closes #3972)
- Removes some special handling for 3.6

### DIFF
--- a/jupyterhub/utils.py
+++ b/jupyterhub/utils.py
@@ -94,13 +94,51 @@ def can_connect(ip, port):
         return True
 
 
-def make_ssl_context(keyfile, certfile, cafile=None, verify=True, check_hostname=True):
-    """Setup context for starting an https server or making requests over ssl."""
+def make_ssl_context(
+    keyfile,
+    certfile,
+    cafile=None,
+    verify=None,
+    check_hostname=None,
+    purpose=ssl.Purpose.SERVER_AUTH,
+):
+    """Setup context for starting an https server or making requests over ssl.
+
+    Used for verifying internal ssl connections.
+    Certificates are always verified in both directions.
+    Hostnames are checked for client sockets.
+
+    Client sockets are created with `purpose=ssl.Purpose.SERVER_AUTH` (default),
+    Server sockets are created with `purpose=ssl.Purpose.CLIENT_AUTH`.
+    """
     if not keyfile or not certfile:
         return None
-    purpose = ssl.Purpose.SERVER_AUTH if verify else ssl.Purpose.CLIENT_AUTH
+    if verify is not None:
+        purpose = ssl.Purpose.SERVER_AUTH if verify else ssl.Purpose.CLIENT_AUTH
+        warnings.warn(
+            f"make_ssl_context(verify={verify}) is deprecated in jupyterhub 2.4."
+            f" Use make_ssl_context(purpose={purpose}).",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+    if check_hostname is not None:
+        purpose = ssl.Purpose.SERVER_AUTH if check_hostname else ssl.Purpose.CLIENT_AUTH
+        warnings.warn(
+            f"make_ssl_context(check_hostname={check_hostname}) is deprecated in jupyterhub 2.4."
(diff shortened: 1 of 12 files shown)


minrk:
Can't test 3.11 on linux because sqlalchemy requires greenlet on linux (specifically because it's easy to install there), but greenlet can't be installed with Python 3.11 yet. Ref: https://github.com/gevent/gevent/issues/1867

sqlalchemy issue about how optional greenlet should be: https://github.com/sqlalchemy/sqlalchemy/issues/6136

minrk:
3.7 changed how `~` is handled in `urllib.quote` (no longer escaped), causing. I rewrote the dns escaping to be a bit more explicit, but it produces the same results consistently across Python versions, except for a few cases that wouldn't have produced valid domain names before.

manics:
Are you releasing this in 2.4.0 or 3.0.0?

minrk:
I opened #3982 about the 2.4 vs 3.0 issue. I was thinking 'next release' for this change, but I'm open to calling the next release 2.4 or 3.0.

In a JupyterHub context, I think dropping a Python version is _less_ disruptive than a db schema change, especially since JupyterHub tends to be more conservative in our Python version support than many other projects. For example, the widely adopted standard of [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) has already dropped support for 3.7 last December.

minrk:
The updated test matrix found and fixed two real bugs:

- internal-ssl wouldn't work with Python 3.10+ due to an ssl API change
- the extremely rare (but tested!) case of `~` in usernames with subdomain-per-user in Python 3.8+

jupyterhub/utils.py
```diff
@@ -82,13 +94,51 @@ def can_connect(ip, port):
...
+            f"make_ssl_context(verify={verify}) is deprecated in jupyterhub 2.4."
+            f" Use make_ssl_context(purpose={purpose}).",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+    if check_hostname is not None:
+        purpose = ssl.Purpose.SERVER_AUTH if check_hostname else ssl.Purpose.CLIENT_AUTH
+        warnings.warn(
+            f"make_ssl_context(check_hostname={check_hostname}) is deprecated in jupyterhub 2.4."
+            f" Use make_ssl_context(purpose={purpose}).",
```
consideRatio, line 128: I wondered how those objects would render like this, maybe rendering them as `ssl.Purpose.SERVER_AUTH` for example would make it clear. This change suggestion makes us render it like that.

image

```suggestion
            f" Use make_ssl_context(purpose=ssl.{str(purpose)}).",
            DeprecationWarning,
            stacklevel=2,
        )
    if check_hostname is not None:
        purpose = ssl.Purpose.SERVER_AUTH if check_hostname else ssl.Purpose.CLIENT_AUTH
        warnings.warn(
            f"make_ssl_context(check_hostname={check_hostname}) is deprecated in jupyterhub 2.4."
            f" Use make_ssl_context(purpose=ssl.{str(purpose)}).",
```

consideRatio (approved):
I made a suggestion related to a deprecation warnings string, and think this looks great overall besides that!

I also opened #3991 to track that we want to test against python 3.11 and why we don't do that based on your comment in this PR.

jupyterhub/utils.py
minrk: Went with `{purpose!s}`, which is the standard way to specify str-format conversion. I thought this was the default, but apparently it's `__format__()` which is [complex for enums](https://github.com/python/cpython/blob/v3.10.0/Lib/enum.py#L750-L767) (_most_ enums would produce `ssl.Purpose.SERVER_AUTH`, but not ssl.Purpose for weird reasons to do with `_member_type_`). After digging into why this wasn't what I expected, I also learned that the simpler behavior [has become the default in 3.11](https://github.com/python/cpython/blob/v3.11.0b5/Lib/enum.py#L1204-L1205).

I also just learned _today_ that it's `__format__` not `__str__` that sets the default substitution in f-strings and other similar situations. This has been true since Python 2.6!

PR #3977, opened 2022-07-14
unpin nbclassic

minrk:
0.4.3 is out, see if it fixes things


minrk:
Looks like the hang was caused by a typo in #3971. Seems all set now.

Issue #3858, opened 2022-04-07
Runtime / user-granted roles and scopes


minrk:
### Proposed change

For sharing and user-initiated collaboration, I think we want the ability for users to _grant_ other users permissions. Right now, to collaborate, admins must grant minimum sharing permissions _at all times_. This works for protyping purposes, and single-group deployments, but not in general

For simplicity, the first implementation of RBAC only allows roles to be defined in config. That means we always reconcile all role assignments to reflect config - roles cannot be created without restarting the hub, and their definitions only reside in config.

Example use cases:

1. start a server (e.g. a dashboard) and make it accessible by all users, unlike the user's default server (https://discourse.jupyter.org/t/refining-server-scopes-in-role-definitions/13760) 
2. user grants limited (e.g. read-only) collaboration access to their current server, _without_ requiring that the other user have that access all the time. The user should be able to revoke permissions they have granted.

The challenges are in reconciliation with config:

1. if we can grant access during runtime, we can't revert to config at each restart. We have to recognize what changes are from changed config and what changes are from runtime, priorities, etc.
2. what's the difference at startup between a role deleted from config (should be deleted) and a role created during runtime (should be left alone)?

### Who would use this feature?

Deployments that want selective and user-controlled collaboration.

### (Optional): Suggest a solution

- REST API for managing roles, role assignments
- need a way to track role changes from config vs runtime to avoid making a mess during hub startup. A single flag on the role might be the simplest way to track - config-managed roles could be read-only via API.
- need to track roles 'owned' by users
- Users may _create_ roles, granting other users permissions
- something like `oauth_allowed_roles`, but for users to govern which permissions they can grant others

meeseeksmachine:
This issue has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/refining-server-scopes-in-role-definitions/13760/4

ktaletsk:
> REST API for managing roles, role assignments

This would help a lot! Currently I have to run wrapper to restart JupyterHub to add individual sharing roles for each user. https://github.com/ktaletsk/jupyterhub-rtc-config-wrapper

minrk:
#4594 is now complete and ready for testing. I still think runtime-defined roles makes sense, so it won't close this issue, but it at least addresses the _primary_ use case.

krassowski:
Was having a `manage_roles` which would mirror `manage_groups` but for roles discussed before?

[...]

PR #3980, opened 2022-07-15
Add role handlers and endpoints for REST API

vladfreeze:
I needed to dynamically add/edit roles and permissions of groups and therefore created an endpoint which allows the admins to create, edit or delete roles using JupyterHub's REST API.

### DIFF
--- a/jupyterhub/apihandlers/__init__.py
+++ b/jupyterhub/apihandlers/__init__.py
@@ -1,6 +1,6 @@
-from . import auth, groups, hub, proxy, services, users
+from . import auth, groups, hub, proxy, services, users, roles
 from .base import *
 
 default_handlers = []
-for mod in (auth, hub, proxy, users, groups, services):
+for mod in (auth, hub, proxy, users, groups, services, roles):
     default_handlers.extend(mod.default_handlers)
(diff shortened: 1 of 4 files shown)


manics:
Relevant issue: https://github.com/jupyterhub/jupyterhub/issues/3858

vladfreeze:
After further testing I noticed that the roles added using the api are deleted when restarting the hub. Is this behaviour intended? It seems that only the default roles show up after restarting.

minrk:
As the issue @manics links mentioned, it is intended _now_. We have to be _extremely_ careful with the runtime creation of permissions. In particular, it's critical to be able to distinguish changes at runtime (should be preserved across restart) from changes in config (old values should be clobbered - only current behavior).

Specific case: if a role is present in the db now, but not in config, it's certain that it has been _deleted_ from config, and should be deleted from the db. That's indistinguishable from a runtime-created role right now, but we have to be able to tell the difference to make the right decision.

eeeeeeeason:
> After further testing I noticed that the roles added using the api are deleted when restarting the hub. Is this behaviour intended? It seems that only the default roles show up after restarting.

Please tell me how to solve this problem in the end. I am also doing dynamic permission control and need persistent role management.thanks alot

PR #3989, opened 2022-07-26
bump moment.js 2.29.4

rpwagner:
[CVE CVE-2022-31129](https://github.com/moment/moment/security/advisories/GHSA-wc69-rhjr-hc9g)


minrk:
Because JupyterHub doesn't use moment for user-generated strings, this is not a vulnerability that affects JupyterHub users. But always good to keep up!

PR #3992, opened 2022-07-29
Avoid IOLoop.current in singleuser mixins

minrk:
- define our own init_ioloop (added in jupyter-server, but not available in notebook)
- call it ASAP  (see https://github.com/jupyter-server/jupyter_server/pull/934)
- put init_httpserver in IOLoop.run_sync because instantiating a server accesses the current loop
- run cleanup via asyncio.run


minrk:
Looks like IOLoop.make_current is still needed for jupyter-server due to https://github.com/jupyter-server/jupyter_server/pull/934

consideRatio (approved):
Super helpful inline comments @minrk! I think I grasp this good enough to go for a merge, this LGTM!

PR #3994, opened 2022-07-29
3.0 changelog

minrk:
had to manually remove already-backported PRs, but I think I got it right.

highlights not complete yet, and I should probably move some of the detail (and add links) to existing docs, rather than just in the changelog.

Still to merge:

- [x] #3992
- [x] #3969
- [x] #3877
- [x] #3909

Then I think we can do 3.0 beta


minrk:
I added #3909 to what I think should be included in 3.0, because it's proven to be the cause of another issue (#3909)

consideRatio:
@minrk great work on many advanced topics, I'm learning quite a bit reviewing this!!!

Anything more you'd like to see part of the 3.0 release - beta or not?

minrk:
I'm going to push one more update here shortly, then it's ready to go from my side.

minrk:
@consideRatio just pushed the last update. I'll make 3.0b1 when this is merged.

minrk:
@consideRatio all green!

manics (approved):
LGTM

minrk:
3.0.0b1 tagged!

Issue #1490, opened 2017-10-18
Jupyterhub in a Resource Constrained Environment


ixxie:
I am planning to install Jupyterhub on the smallest Digital Ocean droplet, which has a single thread and only 512mb of memory. This will be for use by myself and a few friends, so I don't expect heavy compute loads or much simultaneous computation.

I am wondering if there are any ideas about the general minimal requirements for Jupyterhub, and also minimal requirements per simultaneous user. I understand this depends largely on what the user will do, and that limits can be set by the spawner, but I want to understand what suffices to cover the overhead + some minimal working resources.

yuvipanda:
Heya!

I think the overhead will be fairly minimal, compared to what your users are doing. It also heavily depends on what spawner & environment you will be using. 

From an example deployment I have, I see that the hub process is using 77M of RAM and the proxy about 18M of RAM. This is a somewhat large installation (several thousand users), and these RAM estimates also don't count shared library use (so the 77M of RAM for the hub is probably shared with many of the python processes already running).

Hope that's helpful, and good luck for your deployment!

PR #3995, opened 2022-08-01
Show user last activity not server in dashboard

### LINKED ISSUE #3981: Add "Last seen" column in admin page

EmericT:
As proposed on https://github.com/jupyterhub/jupyterhub/issues/3981

fixes #3981

### DIFF
--- a/jsx/src/components/ServerDashboard/ServerDashboard.jsx
+++ b/jsx/src/components/ServerDashboard/ServerDashboard.jsx
@@ -258,7 +258,7 @@ const ServerDashboard = (props) => {
           )}
         </td>
         <td data-testid="user-row-last-activity">
-          {server.last_activity ? timeSince(server.last_activity) : "Never"}
+          {user.last_activity ? timeSince(server.last_activity) : "Never"}
         </td>
         <td data-testid="user-row-server-activity">
           {server.started ? (


jsx/src/components/ServerDashboard/ServerDashboard.jsx
```diff
@@ -258,7 +258,7 @@ const ServerDashboard = (props) => {
           )}
         </td>
         <td data-testid="user-row-last-activity">
-          {server.last_activity ? timeSince(server.last_activity) : "Never"}
+          {user.last_activity ? timeSince(server.last_activity) : "Never"}
```
minrk, line 261: Change both the string and the condition:

```suggestion
          {user.last_activity ? timeSince(user.last_activity) : "Never"}
```

But this is going to use the user activity for _all_ servers, which I think isn't what we want.

Maybe what we want is really:

1. user.lastActivity _if_ server.name === "" _and_ server.last_activity is null.
2. else server.lastActivity

Which is probably best encapsulated in a new variable, around 226

```javascript
// show server activity for running servers (always for named servers)
// and user activity in the default row if the default server isn't running
const lastActivity = (server.lastActivity || server.name !== "") ? server.lastActivity : user.lastActivity;
```

What do you think?

EmericT: Sounds fine by me. 

In my setup, no user ever uses other server than its own, so I don't think there is a use case where this would make a difference (also sorry for the wrong PR obviously).

minrk:
It was bugging me that last activity was Never for stopped servers, until I realized that this is actually a symptom of #3901 - all information about stopped servers is excluded from the model. This should also be fixed by #1490.

PR #4000, opened 2022-08-02
[admin] update, clean jsx deps

minrk:
1. update some dependencies to avoid warnings
2. move build dependencies to `devDependencies`. Doesn't really make a difference since it's not a real package, but cleaner.
3. <del>remove use of outdated (since 2018) `recompose`, which was adding jh api functions via react `props`. Since these are pure, stateless functions, it's simpler to expose them as functions from a `jhapi` utility, rather than passing them around via props</del>

### DIFF
--- a/jsx/package.json
+++ b/jsx/package.json
@@ -41,7 +41,7 @@
     "react-redux": "^7.2.2",
     "react-router": "^5.2.0",
     "react-router-dom": "^5.2.0",
-    "recompose": "^0.30.0",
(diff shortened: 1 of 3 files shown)


minrk:
(draft because test mocks haven't been updated yet)

minrk:
Updated: instead of removing recompose (which I think we should still do, but it wasn't obvious to me how to update the mocks), update to a maintained fork.

PR #4008, opened 2022-08-09
Add draft capacity planning doc

minrk:
I think a general considerations doc is appropriate in the base repo, rather than just in z2jh. What do folks think?

TODO:

- [x] link from somewhere, toc
- [x] finish sections
- [x] more links to relevant sections in z2jh, tljh, k8s, etc.


docs/source/admin/capacity-planning.md
```diff
@@ -0,0 +1,246 @@
...
+### Oversubscribed CPU is okay, running out of memory is bad
+
+An important consideration when assigning resources to users is:
+
+> What happens when users need more than I've given them?
+
+A good summary to keep in mind: **when tasks don't get enough CPU, things are slow.
+When they don't get enough memory, things are broken.**
+Which means it's very important that users have enough memory,
+but much less important that they always have exclusive access to all the CPU they can use.
```
choldgraf, line 108:
```suggestion
An important consideration when assigning resources to users is: **What happens when users need more than I've given them?**

A good summary to keep in mind:

> When tasks don't get enough CPU, things are slow.
  When they don't get enough memory, things are broken.

This means it's **very important that users have enough memory**,
but much less important that they always have exclusive access to all the CPU they can use.
```

This highlights the "important part" a bit more clearly

docs/source/admin/capacity-planning.md
```diff
@@ -0,0 +1,246 @@
...
+
+This relates to [Limits and Requests](limits-requests),
+because these are the consequences of your limits and/or requests not matching what users actually try to use.
+
+A table of mismatched resource allocation situations and their consequences:
+
+| issue                                                    | consequence                                                                           |
```
choldgraf, line 115: this is great

docs/source/admin/capacity-planning.md
```diff
@@ -0,0 +1,246 @@
...
+| CPU oversubscribed (too-low request + too-high limit)    | Poor performance across the system; may crash, if severe                              |
+| Memory limit too low                                     | Servers killed by Out-of-Memory Killer (OOM); lost work for users                     |
+| Memory oversubscribed (too-low request + too-high limit) | System memory exhaustion - all kinds of hangs and crashes and weird errors. Very bad. |
+
+Note that the 'oversubscribed' problem case is where the request is lower than _typical_ usage,
+meaning that the total reserved resources isn't enough for the total _actual_ consumption.
+This doesn't mean that _all_ your users exceed the request,
+just that the _limit_ gives enough room for the _average_ user to exceed the request.
+
+### Example case for oversubscribe memory
```
choldgraf, line 128: Maybe we should point out somewhere really clearly that the most important number is the **Average CPU / Memory used per user on a node**. That's ultimately the number that decides whether you hit your node resource limits

docs/source/admin/capacity-planning.md
```diff
@@ -0,0 +1,246 @@
...
+This will assign 8 users to those 8G of RAM (remember: only requests are used for deciding when a machine is 'full').
+As long as the total of 8 users _actual_ usage is under 8G, everything is fine.
+But the _limit_ allows a total of 24G to be used,
+which would be a mess if everyone used their full limit.
+But _not_ everyone uses the full limit, which is the point!
+
+This pattern is fine if 1/8 of your users are 'heavy' because _typical_ usage will be ~0.7G,
+and your total usage will be ~5G (1 × 2 + 7 × 0.5 = 5.5).
+
+But if _50%_ of your users are 'heavy' you have a problem because that means your users will be trying to use 10G (4 × 2 + 4 × 0.5 = 10),
```
choldgraf, line 146:
```suggestion
But if _50%_ of your users are 'heavy' you have a problem because that means your users will be trying to use 10G (`4 × 2 + 4 × 0.5 = 10`),
```

I find it cleaner to read arithmetic in docs as inline code blocks, wdyt? (I do the same for numbers that are used often as well)

docs/source/admin/capacity-planning.md
```diff
@@ -0,0 +1,246 @@
...
+which you don't have.
+
+You can make guesses at these numbers, but the only _real_ way to get them is to measure (more [below](measuring)).
```
choldgraf, line 149:
```suggestion
You can make guesses at these numbers, but the only _real_ way to get them is to measure (see [](measuring)).
```

I think it's good to use "content-ful" link titles, rather than vague ones like `below` - by removing the title entirely it'll auto-populate the title with the section heading

docs/source/admin/capacity-planning.md
```diff
@@ -0,0 +1,246 @@
...
+- Long-running intense computations suggest higher requests.
+
+Again, using mybinder.org as an example—we run around 100 users on 8-core nodes,
+and still see fairly _low_ overall CPU usage on each user node.
+The limit here is actually Kubernetes' pods per node, not memory _or_ CPU.
+This is likely a extreme case, as many Binder users come from clicking links on webpages
+without any actual intention of running code.
+
+![mybinder.org node CPU usage is low with 50-150 users sharing just 8 cores](../images/mybinder-load5.png)
+
```
choldgraf, line 195:
````suggestion
```[figure} ../images/mybinder-load5.png
mybinder.org node CPU usage is low with 50-150 users sharing just 8 cores
```

````

File: docs/source/admin/capacity-planning.md [hunk 7]
```diff
@@ -0,0 +1,246 @@
...
+- If you see out-of-memory killer events, increase the limit (or talk to your users!)
+- If you see typical memory well below your limit, reduce the request (but not the limit)
+- If _nobody_ uses that much memory, reduce your limit
+- If CPU is your limiting scheduling factor and your CPUs are mostly idle,
+  reduce the cpu request (maybe even to 0!).
+- If CPU usage continues to be low, increase the limit to 2 or 4 to allow bursts of parallel execution.
+
+(measuring)=
+
+### Measuring user resource consumption
```
choldgraf, line 229: I'd also recommend adding a section on cloud provider cost monitoring and limits services. I think these are often the simplest way to limit your cloud bill (e.g. "Have AWS send me an email if I hit XXX spending trajectory on week 3 of the month")

choldgraf (commented):
Thanks for writing this up @minrk - this is really helpful. I made several comments and suggestions throughout, let me know what you think.

We should cross-reference this administrators guide from Z2JH so that it is more discoverable. Maybe we can keep the kubernetes-specific administration guide there, and general hub admin stuff here

docs/source/admin/capacity-planning.md
```diff
@@ -0,0 +1,246 @@
+# Capacity planning
+
+General capacity planning advice for JupyterHub is hard to give,
```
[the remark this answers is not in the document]
minrk: Point taken! I generally follow "semantic line breaks" for exactly that reason, but break on clauses rather than only full sentences. I can do sentences, though!

docs/source/admin/capacity-planning.md
```diff
@@ -0,0 +1,246 @@
...
+
+**There is no single answer to "I have X users, what resources do I need?" or "How many users can I support with this machine?"**
+
+Here are three _typical_ Jupyter use patterns that require vastly different resources:
+
+- negligible resources because computation is mostly idle,
+  e.g. students learning programming for the first time
+- very intense, sustained load, e.g. training machine learning models
+- _mostly_ idle, but needs a lot of resources for short periods of time
+  (interactive research often looks like this)
```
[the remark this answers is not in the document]
minrk: I like this, though I'm not sure the 'scaling out/up'  are the right labels for the first two. These are descriptions of _individual user_ behavior, not cluster behavior. Maybe "Reading/learning" and "Production code"?

docs/source/admin/capacity-planning.md
```diff
@@ -0,0 +1,246 @@
...
+
+(limits-requests)=
+
+### Limit vs Request
+
+Many scheduling tools like Kubernetes have two separate ways of allocating resources to users.
+A **Request** or **Reservation** describes how much resources are _set aside_ for each user.
+Often, this doesn't have any practical effect other than deciding when a given machine is considered 'full'.
+If you are using expandable resources like an autoscaling Kubernetes cluster,
+'requesting' more resources than fit on currently running nodes is when a new node is launched and added to the pool (a cluster **scale-up event**).
```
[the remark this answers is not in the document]
minrk: I found it a little hard to write ;)

docs/source/admin/capacity-planning.md
minrk: Yeah, I was surprised by the lack of block quote formatting. I think I have a bug report for our sphinx theme on the way.

docs/source/admin/capacity-planning.md
minrk:
```suggestion
### Example case for oversubscribing memory
```

docs/source/admin/capacity-planning.md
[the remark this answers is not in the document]
choldgraf: I am also a fan of semantic breaks :-) I've just found that for some reason, it is more common for me to want to edit at the sentence-level rather than the clause level, so diffing one line corresponding to a sentence is the compromise I use. Either one is fine, I just wanted to note it

docs/source/admin/capacity-planning.md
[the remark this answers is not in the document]
choldgraf: good point - in my mind, the amount of resources needed is _usually_ correlated with the total number of users, but I agree it's not strictly true

docs/source/admin/capacity-planning.md
minrk: Good call. Added.

docs/source/admin/capacity-planning.md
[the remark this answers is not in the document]
minrk: Yeah, I mostly thought about reading/editing a fixed diff, but sentences makes total sense in the context of tools like GitHub suggestions. I hadn't thought of that.

docs/source/admin/capacity-planning.md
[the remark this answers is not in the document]
minrk: > amount of resources needed is usually correlated with the total number of users

Yes, though it's more directly related to concurrent users. I just added a section on concurrent users which discusses some of this and links the idle culler.

I think there are 3 questions I expect most people coming this doc to have:

- I have X resources, how many users can I handle?
- I have Y users, what total resources do I need?
- What do my individual users need (already looking for what to put in the z2jh singleuser resource fields)?

and I want to get right away to the answer to all three: "It depends. What are your users are doing?"

The framing of this whole doc should be: understanding what your users actually do is key to accurately allocating resources.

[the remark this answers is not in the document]
minrk: fwiw, I went with:

- **Learning**
 - **Production code**
 - **Bursting**

minrk:
@choldgraf does this look okay for now?

choldgraf (approved):
This looks good to me - it's a big improvement and we can iterate in subsequent PRs if we want to improve stuff here and there. I'll merge this one, thanks for this awesome and helpful content!

choldgraf:
Random follow-up. I was planning to squash-and-merge this, since I find the commit message is easier to comprehend that way (it uses the PR title as the commit message which I like). But, I found that it wasn't allowed for this repository. Is there a strong reason for that? I don't think we should default to squash-and-merge or anything, but strictly disallowing it was unexpected to me.

PR #4013, opened 2022-08-10
Test 3.11

### LINKED ISSUE #3991: ci: test against python 3.11
As @minrk noted in https://github.com/jupyterhub/jupyterhub/pull/3976#issuecomment-1184975768, we can't yet do this

> Can't test 3.11 on linux because sqlalchemy requires greenlet on linux (specifically because it's easy to install there), but greenlet can't be installed with Python 3.11 yet. Ref: https://github.com/gevent/gevent/issues/1867
>
>sqlalchemy issue about how optional greenlet should be: https://github.com/sqlalchemy/sqlalchemy/issues/6136

minrk:
Try using a mock `greenlet` package to satisfy the install on 3.11

closes #3991


.github/workflows/test.yml
```diff
@@ -136,9 +135,19 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: "${{ matrix.python }}"
+
       - name: Install Python dependencies
         run: |
           pip install --upgrade pip
+
+          if [ "${{ matrix.python }}" == "3.11"* ]; then
```
consideRatio, line 143:
```suggestion
          if [[ "${{ matrix.python }}" == "3.11"* ]]; then
```

I think this is required.

PR #4018, opened 2022-08-19
reset offset to 0 on name filter change

### LINKED ISSUE #3957: A new search at /hub/admin should initialize page to 0
### Bug description

If I navigate to the second page of users (`/hub/admin#/?page=1`), then use the search field to try to find someone, search returns empty results. If I navigate back to the first page then initiate a search, it returns results matching the search term.

Search will return _some_ results on page > 0 if I use a short search time like `a`. So if I start the search while viewing page X, then I see page X of the search results.

#### Expected behaviour

I expected that when I initiate at search, it should always start at page=0.

#### Actual behaviour

Search returns no results, even if I know there to be a user matching the search term. It appears that search results are being paginated too and the initial page= offset is hiding results when the count of matching users is less than `limit`. For example if I search for "a" which matches more than 50 users on this particular hub, and if I'm starting out on the second page, then I do see some results. So search is working but the initial page offset was surprising and gives a false impression for non-pathological search terms.

### How to reproduce

1. Go to the first page of /hub/admin on a hub which has more than a page's worth of users (50 is the default?). Note the name of any user.
2. Click on Next to go to the next page.
3. Search for the user you noted in the first step. Alternatively search for any other user that you know to exist.
4. Observe empty results.

### Your personal set up

https://github.com/berkeley-dsep-infra/datahub

helm chart 1.1.3-n554.hfa81c67f
jupyterhub 2.3.0

<details><summary>Full environment</summary>

```
authenticator==14.2.0
oauthlib==3.2.0
onetimepass==1.0.1
packaging==21.3
pamela==1.0.0
pem==21.2.0
prometheus-client==0.14.1
psycopg2-binary==2.9.3
py-spy==0.3.12
pyasn1==0.4.8
pycparser==2.21
pycryptodome==3.14.1
pycryptodomex @ file:///tmp/pre-built-wheels/pycryptodomex-3.14.1-cp35-abi3-manylinux2010_x86_64.whl
pycurl @ file:///tmp/pre-built-wheels/pycurl-7.45.1-cp39-cp39-linux_x86_64.whl
pyjwkest==1.4.2
PyJWT==1.7.1
PyMySQL==1.0.2
pyOpenSSL==22.0.0
pyparsing==3.0.9
pyrsistent==0.18.1
python-dateutil==2.8.2
python-json-logger==2.0.2
python-slugify==6.1.2
PyYAML==6.0
requests==2.27.1
requests-oauthlib==1.3.1
ruamel.yaml==0.17.21
ruamel.yaml.clib @ file:///tmp/pre-built-wheels/ruamel.yaml.clib-0.2.6-cp39-cp39-manylinux1_x86_64.whl
six==1.16.0
SQLAlchemy==1.4.36
sqlalchemy-cockroachdb==1.4.3
statsd==3.3.0
text-unidecode==1.3
tornado==6.1
traitlets==5.2.1.post0
urllib3==1.26.9
wrapt==1.14.1
yarl==1.7.2```
</details>

<details><summary>Configuration</summary>
<!--
For JupyterHub, especially include information such as what Spawner and Authenticator are being used.
Be careful not to share any sensitive information.
You can paste jupyterhub_config.py below.
To exclude lots of comments and empty lines from auto-generated jupyterhub_config.py, you can do:
    grep -v '\(^#\|^[[:space:]]*$\)' jupyterhub_config.py
-->

```

[kubespawner](https://github.com/berkeley-dsep-infra/datahub/blob/staging/hub/values.yaml#L301)
[canvasauthenticator](https://github.com/berkeley-dsep-infra/datahub/tree/staging/images/hub/canvasauthenticator)```

</details>

minrk:
move offset to redux state, rather than independent, since it can come from two places (user_page and pagination footer). Seems to keep things in sync better.

Adds reducers for setting offset, name filter explicitly.

closes #3957

### DIFF
--- a/jsx/src/Store.js
+++ b/jsx/src/Store.js
@@ -1,23 +1,48 @@
 export const initialState = {
   user_data: undefined,
-  user_page: 0,
+  user_page: { offset: 0, limit: window.api_page_limit || 100 },
   name_filter: "",
   groups_data: undefined,
-  groups_page: 0,
-  limit: window.api_page_limit,
+  groups_page: { offset: 0, limit: window.api_page_limit || 100 },
+  limit: window.api_page_limit || 100,
 };
 
 export const reducers = (state = initialState, action) => {
   switch (action.type) {
     // Updates the client user model data and stores the page
+    case "USER_OFFSET":
+      return Object.assign({}, state, {
+        user_page: Object.assign({}, state.user_page, {
+          offset: action.value.offset,
+        }),
+      });
+
+    case "USER_NAME_FILTER":
+      // set offset to 0 if name filter changed,
+      // otherwise leave it alone
+      const newOffset =
+        action.value.name_filter !== state.name_filter ? 0 : state.name_filter;
+      return Object.assign({}, state, {
+        user_page: Object.assign({}, state.user_page, {
+          offset: newOffset,
+        }),
+        name_filter: action.value.name_filter,
+      });
+
     case "USER_PAGE":
       return Object.assign({}, state, {
         user_page: action.value.page,
         user_data: action.value.data,
-        name_filter: action.value.name_filter || "",
       });
 
-    // Updates the client group model data and stores the page
+    // Updates the client group user model data and stores the page
+    case "GROUPS_OFFSET":
+      return Object.assign({}, state, {
+        groups_page: Object.assign({}, state.groups_page, {
+          offset: action.value.offset,
+        }),
+      });
+
     case "GROUPS_PAGE":
       return Object.assign({}, state, {
         groups_page: action.value.page,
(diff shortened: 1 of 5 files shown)


minrk:
@naatebarber I've tested this and it's working in real tests. But I believe the test failing is because the useEffect is triggered by the reducer, and the tests don't actually enable the reducers. It's not clear to me how to make the test environment more realistic. Any idea?

consideRatio:
I tried to review this, but I can't yet =/ There is soo much code that I can't understand, and I end up recursively stepping in deeper into the code to and never manage to come back properly.

@minrk what do you think about disabling the non-functional tests for now to get this merged, and going for a 3.0.0 release (-> z2jh 2.0.0 release), and opening an issue in jupyterhub to represent the tests needed to be fixed?

I figure that could reduce the urgency of getting that done. I feel an urgency to get z2jh 2 out and would not like to reference a jupyterhub beta release.

minrk:
@consideRatio I updated the tests so they pass, with some notes about things that _should_ work but don't

consideRatio:
Wieee nice work on this @minrk, I hope to learn more to be able to help when needed with these parts as well.

PR #4022, opened 2022-08-23
Use correct expiration labels in drop-down menu on token page.

possiblyMikeB:
According to the REST API documentation for the POST end-point `/users/{name}/token`, the expected units of the `expires_in` attribute are seconds.  As such, the titles for the options of the "Token expires" drop-down menu from the template `share/jupyterhub/templates/token.html` specify the wrong length of time. 

This pull request corrects that by assuming the values for the options are correct, and updating each option's title.


consideRatio:
@possiblyMikeB thanks!!!

PR #4026, opened 2022-09-02
Add browser-based tests with Selenium

mouse1203:
adding Selenium tests - In Progress

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -87,6 +87,8 @@ jobs:
             subdomain: subdomain
           - python: "3.10"
             ssl: ssl
+          - python: "3.10"
+            selenium: selenium
           - python: "3.11.0-rc.1"
           - python: "3.10"
             main_dependencies: main_dependencies
@@ -215,16 +217,23 @@ jobs:
               DB=postgres bash ci/init-db.sh
           fi
       - name: Setup Firefox
+        if: matrix.selenium
         uses: browser-actions/setup-firefox@latest
         with:
           firefox-version: latest
 
       - name: Setup Geckodriver
+        if: matrix.selenium
         uses: browser-actions/setup-geckodriver@latest
+
+      - name: Configure selenium tests
+        if: matrix.selenium
+        run: echo "PYTEST_ADDOPTS=$PYTEST_ADDOPTS -m selenium" >> "${GITHUB_ENV}"
 
       - name: Run pytest
         run: |
           pytest --maxfail=2 --cov=jupyterhub jupyterhub/tests
+
       - name: Submit codecov report
         run: |
           codecov
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -215,13 +215,11 @@ jobs:
               DB=postgres bash ci/init-db.sh
           fi
       - name: Setup Firefox
-        if: matrix.os == 'ubuntu-latest'
         uses: browser-actions/setup-firefox@latest
         with:
           firefox-version: latest
 
       - name: Setup Geckodriver
-        if: matrix.os == 'ubuntu-latest'
         uses: browser-actions/setup-geckodriver@latest
 
       - name: Run pytest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -214,6 +214,15 @@ jobs:
               DB=postgres bash ci/docker-db.sh
               DB=postgres bash ci/init-db.sh
           fi
+      - name: Setup Firefox
+        if: matrix.os == 'ubuntu-latest'
+        uses: browser-actions/setup-firefox@latest
+        with:
+          firefox-version: latest
+
+      - name: Setup Geckodriver
+        if: matrix.os == 'ubuntu-latest'
+        uses: browser-actions/setup-geckodriver@latest
 
       - name: Run pytest
         run: |
--- a/jupyterhub/tests/selenium/conftest.py
+++ b/jupyterhub/tests/selenium/conftest.py
@@ -1,5 +1,6 @@
 import pytest
 from selenium import webdriver
+from selenium.webdriver.firefox.options import Options as FirefoxOptions
 
 
 @pytest.fixture()
--- a/jupyterhub/tests/selenium/conftest.py
+++ b/jupyterhub/tests/selenium/conftest.py
@@ -5,8 +5,8 @@ from selenium import webdriver
 @pytest.fixture()
 def browser():
     options = webdriver.FirefoxOptions()
-    options.set_headless()
-    driver = webdriver.Firefox(firefox_options=options)
+    options.headless = True
+    driver = webdriver.Firefox(options=options)
     yield driver
     driver.close()
     driver.quit()
--- a/jupyterhub/tests/selenium/conftest.py
+++ b/jupyterhub/tests/selenium/conftest.py
@@ -1,0 +1,10 @@
+import pytest
+from selenium import webdriver
+
+
+@pytest.fixture()
+def browser():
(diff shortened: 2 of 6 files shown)


minrk:
I'm working with @mouse1203 on getting selenium test infrastructure going so we can test actual pages. Welcome, @mouse1203!

jupyterhub/tests/selenium/test_browser.py
```diff
@@ -0,0 +1,417 @@
...
+        ("user", ""),
+        (" ", "password"),
+        ("user", "password"),
+    ],
+)
+async def test_invalid_credantials(app, browser, user, pass_w):
+
+    await open_url(app, browser)
+    await login(browser, user, pass_w)
+    time.sleep(5)
```
minrk, line 355: we shouldn't have any time.sleep in the tests because they don't allow the server to respond. If anything, they should be `await asyncio.sleep`. But ideally, we should always wait for a specific condition. Sleeps always have a chance to be too short _sometimes_.

.github/workflows/test.yml
```diff
@@ -214,6 +214,15 @@ jobs:
               DB=postgres bash ci/docker-db.sh
               DB=postgres bash ci/init-db.sh
           fi
+      - name: Setup Firefox
+        if: matrix.os == 'ubuntu-latest'
```
minrk, line 218: remove this 'if' here and in Setup Geckodriver

jupyterhub/tests/selenium/conftest.py
```diff
@@ -0,0 +1,10 @@
+import pytest
+from selenium import webdriver
+
+
+@pytest.fixture()
+def browser():
+    driver = webdriver.Firefox()
```
minrk, line 7: Probably want to make sure we use headless firefox to avoid issues on CI:
```suggestion
    options = webdriver.FirefoxOptions()
    options.set_headless()
    driver = webdriver.Firefox(firefox_options=options)
```

jupyterhub/tests/selenium/test_browser.py
```diff
@@ -0,0 +1,419 @@
...
+  </h1> <p>Jupyter has lots of moons, but this is not one...</p>
+    4. - 8.  the same as 3
+    """
+
+    """
+    print("url ---" +browser.current_url)
+
+    #print("user ()" +f"{user}")
+    print("user (type)" +str(user))
+    assert location in browser.current_url      """
```
consideRatio, line 341: This seems like debugging content or a partial docstring worth revising up before merge.

jupyterhub/tests/selenium/test_browser.py
```diff
@@ -0,0 +1,419 @@
...
+    'running, next_url, location, params',
+    [
+        # default URL if next not specified, for both running and not
+        (True, '', '', None),
+        (False, '', '', None),
+        # next_url is respected
+        (False, '/hub/admin', '/hub/admin', None),
+        (False, '/user/other', '/hub/user/other', None),
+        (False, '/absolute', '/absolute', None),
+        (False, '/has?query#andhash', '/has?query#andhash', None),
```
consideRatio, line 231: I love seeing these comments to help indicate what various test variations parts is about!!! Nice!

jupyterhub/tests/selenium/test_browser.py
```diff
@@ -0,0 +1,419 @@
...
+        if next_url.endswith(f"/user/{user}/") == False:
+            webdriver_wait(
+                browser, EC.url_to_be(ujoin(public_url(app), f"/user/{user}/"))
+            )
+        assert next_url.endswith(f"/user/{user}/")
+        print("Current Link 2: " + browser.current_url)
+
+
+"""TBD"""
+
```
consideRatio, line 219: Something to revise before merge

consideRatio (commented):
Wieee nice to see this kind of test coverage be added! Thank you for contributing with this @mouse1203!!! :heart: :tada:

I'm not yet confident about testing browser interactions etc, but I'm committed to contribute with maintenance to JupyterHub, so I wanted to highlight how valuable I find the in-code documentation you provide!!

If you can, I'd love to see even more to help people like me not so experienced with browser tests. I've written down some ideas on what I think could be helpful for me and other contributors that aren't experienced with these kinds of tests:
- [ ] A docstring at the top of this .py file summarizing briefly what kinds of tests can be done and is done in this file. I'm thinking something like "In these tests we use selenium and firefox to interact with a JupyterHub and verify that the returned HTML meets various criteria such as having a button rendered in the HTML.".
  Things relevant to convey that I currently have an open question about is:
  - Is selenium and/or firefox rendering HTML and letting javascript run, which in practice could influence the HTML as well?
  - What kinds of tests are done? We load and look at HTML in some way, but are we also interacting with buttons etc after that for example? Is that in or out of scope for these kinds of tests?
  - Maybe a link or two to relevant documentation for working with these kinds of tests
- [ ] Docstrings for helper functions (`webdriver_wait`, `in_thread`, `open_url`, ...)
  I think if these helper functions have some explanation to them in a docstring, maybe the test functions won't need them because it may be sufficiently easy to read the code itself to understand the details.

jupyterhub/tests/selenium/test_browser.py
```diff
@@ -0,0 +1,419 @@
...
+    elif user.active and not running:
+        # ensure not running
+        await user.stop()
+    time.sleep(5)
+    # open_url(app,browser)
+    await in_thread(browser.get, (ujoin(public_url(app), url)))
+    await login(browser, user='test_user', pass_w='test_user')
+
+    # await get_page(url, app, allow_redirects=False)
+    time.sleep(50)
```
minrk, line 293: Make sure to remove all `time.sleep`, which are blocking and prevent processing on the server. We _may_ need a few short `await asyncio.sleep`, but let's try to wait for specific conditions as much as possible.

jupyterhub/tests/selenium/test_browser.py
```diff
@@ -0,0 +1,228 @@
...
+)
+from jupyterhub.utils import exponential_backoff
+
+from ...utils import url_path_join
+from ..utils import public_host, public_url, ujoin
+
+pytestmark = pytest.mark.selenium
+
+
+async def webdriver_wait(driver, condition, timeout=30):
```
minrk, line 27: Docstring for wait - explain that it's an async wrapper for selenium's wait function, and that `condition` is something from selenium's expected_conditions

jupyterhub/tests/selenium/test_browser.py
```diff
@@ -0,0 +1,228 @@
...
+
+    via current event loop's run_in_executor
+
+    Returns asyncio.Future
+    """
+
+    return asyncio.get_event_loop().run_in_executor(None, partial(f, *args, **kwargs))
+
+
+async def open_url(app, browser):
```
minrk, line 46: Let's update the signature so the URL can be passed:

```python
async def open_url(app, browser, url="login")
```

jupyterhub/tests/selenium/test_browser.py
```diff
@@ -0,0 +1,228 @@
...
+    """initiating open the login page in the browser"""
+    url = url_path_join(public_host(app), app.hub.base_url, "login")
+    await in_thread(browser.get, url)
+    return url
+
+
+def click(browser, by_locator):
+    WebDriverWait(browser, 10).until(
```
minrk, line 54: Add a docstring here and below about "wait for element to be visible, then click on it"

minrk (commented):
A few suggestions for more clarifying docstrings, then I think this is good to go. Great job!

minrk (approved):
Terrific, thank you!

PR #4029, opened 2022-09-07
Update changelog for 3.0.0

consideRatio:
I wasn't sure if we should await something further 3.0.0 or it was time to go for a release, so I figured I'll update the changelog with recent PRs etc.

To spot changes since this PR was updated, use `github-activity --since 5e29605` (and ignore #4018 that shows up).


minrk:
Tag pushed, release assembling.

consideRatio:
Wieee!!

PR #4032, opened 2022-09-09
Use XSRF tokens for cross-site checks

### LINKED ISSUE #3556: Consider relaxing referer check on hub API requests
### Proposed change
Currently several hub API endpoints are protected by a referer check:
https://github.com/jupyterhub/jupyterhub/blob/c00c3fa28703669b932eb84549654238ff8995dc/jupyterhub/apihandlers/base.py#L53-L61

Which means API requests must originate from `http(s)://<hostname>/hub/`.

The RBAC work exposes more granular access to functions which were previously for admins only, which means a single-user extension might want to make hub API calls.

For example, in https://discourse.jupyter.org/t/plans-on-bringing-rtc-to-jupyterhub/9813/7 I made a user an admin of a group. I tried to get the list of users in the group using `GET /hub/api/groups/{group-name}` from a JupyterLab extension but this is blocked as the referer URL path starts with `/user/{username}` instead of `/hub`.

### Alternative options
Do nothing

Singleuser apps that want to take advantage of some functions that were previously limited to hub admins.

It would potentially also be useful for someone building an external JupyterHub interface.

Begin by describe the security risks this is intended to protect against. Then work out if those protections can be implemented in some other way.

### LINKED ISSUE #4024: Users unable to restart their own server
Hi, recently upgraded JupyterHub from 1.1.0 to 2.3.1. After upgrading our users are no longer able to use
File -> Hub Control Panel -> Stop My Server
the following error is shown
```
API request failed (403): Action is not authorized with current scopes; requires any of [delete:servers]
```

Looks like I might have to add a role to the jupyterhub_config.py to bestow "delete:servers" but I don't understand how to write one to apply to all users. Can you please help?

minrk:
## The problem

We want to validate that cookie-authenticated requests (requests authenticated with browser-stored credentials and not tokens) from the Hub are coming from the Hub, and not from pages served by e.g. services or single-user servers (those requests should use API tokens to authenticate).

Previously, we used various header checks to implement what are normally `Origin` header checks, but browsers have elected to send these only with explicitly cross-origin requests, rather than all requests, so we had to improvise a bit with a combination of headers. This has proved unreliable, especially when additional proxies are involved, all of which can modify the headers on the way by in different ways.

## The solution

There is another mechanism meant to protect against the same event: the XSRF token. I won't go into detail, but we use it in other Jupyter server API requests, and it is a mechanism to verify that the page sending the request has access to the credentials (cookies) for the target site.

## This PR

Removes all Referer checks, which have proven unreliable and have never been particularly strong. Referer's never been something that should be relied upon, since browsers can disable it entirely for privacy, and pages should still work when they do this.

We can set the XSRF cookie on paths for more robust inter-path protections, but it's still pretty easy to circumvent for the singleuser server -> hub case unless user subdomains are used (this isn't really different than our use of Referer, just the mechanism).

closes #3556 
closes #4024

## (potentially) Breaking change

This could be considered backward-incompatible in that cookie-authenticated API requests to the Hub need to now send the xsrf token. The reason I don't consider this a necessarily breaking change is that there aren't any supported extension points where cookie-authenticated API requests can really be made, except by our own page. Though _technically_ a fully custom HTML template could make all kinds of API requests. If they used our `jhapi` helper js, they wouldn't be affected. It will also break bypasses of the Referer checks, but that's fixing a bug, not breaking a valid use.

### DIFF
--- a/jupyterhub/apihandlers/auth.py
+++ b/jupyterhub/apihandlers/auth.py
@@ -15,6 +15,11 @@ from .base import APIHandler, BaseHandler
 
 
 class TokenAPIHandler(APIHandler):
+    def check_xsrf_cookie(self):
+        # no xsrf check needed here
+        # post is just a 404
+        return
+
     @token_authenticated
     def get(self, token):
         # FIXME: deprecate this API for oauth token resolution, in favor of using /api/user
@@ -378,31 +383,7 @@ class OAuthAuthorizeHandler(OAuthHandler, BaseHandler):
     @web.authenticated
     def post(self):
         uri, http_method, body, headers = self.extract_oauth_params()
-        referer = self.request.headers.get('Referer', 'no referer')
-        full_url = self.request.full_url()
-        # trim protocol, which cannot be trusted with multiple layers of proxies anyway
-        # Referer is set by browser, but full_url can be modified by proxy layers to appear as http
-        # when it is actually https
-        referer_proto, _, stripped_referer = referer.partition("://")
-        referer_proto = referer_proto.lower()
-        req_proto, _, stripped_full_url = full_url.partition("://")
-        req_proto = req_proto.lower()
-        if referer_proto != req_proto:
-            self.log.warning("Protocol mismatch: %s != %s", referer, full_url)
-            if req_proto == "https":
-                # insecure origin to secure target is not allowed
-                raise web.HTTPError(
-                    403, "Not allowing authorization form submitted from insecure page"
-                )
-        if stripped_referer != stripped_full_url:
-            # OAuth post must be made to the URL it came from
-            self.log.error("Original OAuth POST from %s != %s", referer, full_url)
-            self.log.error(
-                "Stripped OAuth POST from %s != %s", stripped_referer, stripped_full_url
-            )
-            raise web.HTTPError(
-                403, "Authorization form must be sent from authorization page"
-            )
+        # TODO: per-page xsrf token to verify origin on this page!
 
         # The scopes the user actually authorized, i.e. checkboxes
         # that were selected.
--- a/jupyterhub/apihandlers/auth.py
+++ b/jupyterhub/apihandlers/auth.py
@@ -383,8 +383,6 @@ class OAuthAuthorizeHandler(OAuthHandler, BaseHandler):
     @web.authenticated
     def post(self):
         uri, http_method, body, headers = self.extract_oauth_params()
-        # TODO: per-page xsrf token to verify origin on this page!
-
         # The scopes the user actually authorized, i.e. checkboxes
         # that were selected.
         scopes = self.get_arguments('scopes')
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -337,6 +337,15 @@ class UserAPIHandler(APIHandler):
 class UserTokenListAPIHandler(APIHandler):
     """API endpoint for listing/creating tokens"""
 
+    # defer check_xsrf_cookie so we can accept auth
+    # in the `auth` request field, which shouldn't require xsrf cookies
+    _skip_post_check_xsrf = True
+
+    def check_xsrf_cookie(self):
+        if self.request.method == 'POST' and self._skip_post_check_xsrf:
+            return
+        return super().check_xsrf_cookie()
+
     @needs_scope('read:tokens')
     def get(self, user_name):
         """Get tokens for a given user"""
@@ -374,6 +383,7 @@ class UserTokenListAPIHandler(APIHandler):
                 if isinstance(name, dict):
                     # not a simple string so it has to be a dict
                     name = name.get('name')
+                # don't check xsrf if we've authenticated via the request body
             except web.HTTPError as e:
                 # turn any authentication error into 403
                 raise web.HTTPError(403)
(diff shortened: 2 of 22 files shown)


julietKiloRomeo:
This sounds great and would fix some problems we have from running jupyterhub behind apache.

We have followed [the recommendations](https://jupyterhub.readthedocs.io/en/stable/reference/config-proxy.html). That leads to problems with some api requests being upgraded to https and subsequently caught in cross site checks. Two immediate symptoms are that the progress bar during spawn does not work and that the button for stopping a users server does not work.

Our patch looks like this:
```
--- jupyterhub/apihandlers/base.py.orig	2022-10-14 10:54:22.326789888 +0200
+++ jupyterhub/apihandlers/base.py	2022-10-14 10:56:49.783292624 +0200
@@ -80,10 +80,8 @@
             443 if referer_url.scheme == 'https' else 80
         )
         host_port = host_url.port or (443 if host_url.scheme == 'https' else 80)
-        if (
-            referer_url.scheme != host_url.scheme
-            or referer_url.hostname != host_url.hostname
-            or referer_port != host_port
+        if (# allow upgrade of http traffic in apache
+            referer_url.hostname != host_url.hostname
             or not (referer_url.path + "/").startswith(host_url.path)
         ):
             self.log.warning(

```
which is a small part of what your pull request covers :thumbsup:

We also patch the singleuser.mixin because it will overwrite our kerberos token (terminated at apache) when running jupyterlab.

```
--- jupyterhub/singleuser/mixins.py.orig	2022-10-14 10:54:11.151054679 +0200
+++ jupyterhub/singleuser/mixins.py	2022-10-14 10:56:15.452107441 +0200
@@ -746,7 +746,7 @@
 
         Only has effect on jupyterlab_server >=2.9
         """
-        page_config["token"] = self.hub_auth.get_token(handler) or ""
+        # allow lab to run behind apache with gssapi
         return page_config
 
     def patch_default_headers(self):
```

I don't see any changes to that file in the PR. Is that because the change to jupyterhub/handlers/base.py fixes that or will we still need to patch singleuser.mixin?

minrk:
@julietKiloRomeo Sorry, I missed the bit about the singleuser page config token when I read through that before. That line should likely not be removed, since it will change how and whether users are authenticated to make the right requests. How are you using a kerberos token in the JupyterLab API requests?

minrk:
includes #4258, needed for some cookie tests to pass (this PR is where I found the zombie cookie)

minrk:
_Finally_ got this done, and thanks to @mouse1203's selenium tests, I'm a lot more confident that this won't break things (it makes sure our various forms are submitted with the token properly). Also, now that we have a db revision merged, this will be a 4.0 change, which also makes me feel better.

jupyterhub/apihandlers/auth.py
```diff
@@ -378,31 +383,7 @@ async def get(self):
...
-        if stripped_referer != stripped_full_url:
-            # OAuth post must be made to the URL it came from
-            self.log.error("Original OAuth POST from %s != %s", referer, full_url)
-            self.log.error(
-                "Stripped OAuth POST from %s != %s", stripped_referer, stripped_full_url
-            )
-            raise web.HTTPError(
-                403, "Authorization form must be sent from authorization page"
-            )
+        # TODO: per-page xsrf token to verify origin on this page!
```
consideRatio, line 386: Is it correct that this is a todo not meant to be done as part of this PR?

jupyterhub/apihandlers/users.py
```diff
@@ -384,7 +394,14 @@ async def post(self, user_name):
...
+                raise web.HTTPError(403)
             requester = self.find_user(name)
+        else:
+            # perform delayed xsrf check
+            # if we aren't authenticating via the request body
+            self._skip_post_check_xsrf = False
+            self.check_xsrf_cookie()
         if requester is None:
             # couldn't identify requester
             raise web.HTTPError(403)
```
consideRatio, line 407: It think it could make sense to re-locate the if statement on line 388-390 within the initial if block starting under 369, not important though.

consideRatio (approved):
@minrk this looks good to me, but I'm not confident at all that I would have caught a bug or similar in this PR. I'm very happy that we have various tests setup that could catch issues related to this logic!

Thank you for introducing the topic in the PR description!!!

jupyterhub/apihandlers/auth.py
minrk: No, that's an old note. Removed.

jupyterhub/apihandlers/users.py
minrk: That would make sense and definitely still be correct. I like having it here, even though it can result in one redundant `if x is None` check because it's an unconditional block (if nobody identified, definitely stop), rather than having more complex branches. It helps me be more confident that the handler will not proceed past this point if auth has failed, when auth can be achieved via a few different paths.

jupyterhub/apihandlers/auth.py
minrk: For more historical info - this TODO represented an XSRF check of sorts - we required that the oauth form request was submitted from the oauth form page itself, and the new xsrf checks no longer preserve that restriction. Setting a _separate_ xsrf cookie for that page would have allowed us to keep the same check, but I don't think we need it.

consideRatio:
Thank you for your thorough work on this and help to explain how things work!!! :heart: :tada: :sunflower:

Issue #4033, opened 2022-09-10
admin-react.js not found in v3.0.0 docker image


takeshi-yashiro:
### Bug description

The latest version of official docker image, `jupyterhub/jupyterhub:3.0.0` does not contain admin-react.js file, which results in blank hub admin page.

The lack of this file can be confirmed by entering the container and running the following command.

```
root@d82a49a1c06f:/srv/jupyterhub# cd /usr/local/share/jupyterhub/static/js
root@d82a49a1c06f:/usr/local/share/jupyterhub/static/js# ls -al
total 44
drwxr-xr-x 2 root root 4096 Sep  9 06:35 .
drwxr-xr-x 7 root root 4096 Sep  9 06:35 ..
-rw-r--r-- 1 root root 8013 Sep  9 06:35 admin.js
-rw-r--r-- 1 root root 3293 Sep  9 06:35 home.js
-rw-r--r-- 1 root root 4446 Sep  9 06:35 jhapi.js
-rw-r--r-- 1 root root  398 Sep  9 06:35 not_running.js
-rw-r--r-- 1 root root 1463 Sep  9 06:35 token.js
-rw-r--r-- 1 root root 4383 Sep  9 06:35 utils.js
```

I assume `yarn build` and `yarn place` need to be added in `Dockerfile`, as adding `admin-react.js` generated with these command like below solves this problem. 

```
COPY admin-react.js /usr/local/share/jupyterhub/static/js/admin-react.js
```

#### Expected behaviour

Access to  /hub/admin shows user management table, without any HTTP errors.

#### Actual behaviour

Access to  /hub/admin results in a blank page, with 404 error for /hub/static/js/admin-react.js.

### Your personal set up

 - OS: Docker Desktop v4.11.1 on macOS 12.5.1
 
 - Version(s): `jupyterhub/jupyterhub:latest` and `jupyterhub/jupyterhub:3.0.0` images
 

<details><summary>Full environment</summary>

```
# paste output of `pip freeze` or `conda list` here
```
</details>

<details><summary>Configuration</summary>

```python
# jupyterhub_config.py
```
</details>

<details><summary>Logs</summary>

```
# paste relevant logs here, if any
```
</details>

benz0li:
> The latest version of official docker image, `jupyterhub/jupyterhub:3.0.0` does not contain admin-react.js file, which results in blank hub admin page.

@minrk I can confirm this issue.

Paul2708:
[...]

PR #4034, opened 2022-09-10
Fix Dockerfile yarn JSX build

manics:
Closes https://github.com/jupyterhub/jupyterhub/issues/4033

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,11 +38,11 @@ RUN apt-get update \
     python3-venv \
     nodejs \
     npm \
-    yarn \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/*
 
 RUN python3 -m pip install --upgrade setuptools pip build wheel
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -51,7 +52,7 @@ WORKDIR /src/jupyterhub
 
 # Build client component packages (they will be copied into ./share and
 # packaged with the built wheel.)
-RUN python3 setup.py bdist_wheel
+RUN python3 -m build --wheel
```
manics, line 55: This change isn't necessary, but it matches the pypi build
https://github.com/jupyterhub/jupyterhub/blob/02dead8ab1ad53655bdb461d9d7f94b52f2ecbec/.github/workflows/release.yml#L51-L53

consideRatio (approved):
LGTM, thanks @manics!

utkarshgupta137:
Can you please update 3.0 builds with this change?

PR #4035, opened 2022-09-10
setup.py: require npm, check that NPM CSS JSX commands succeed

manics:
Makes setup.py more robust to unexpected errors.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -62,6 +62,10 @@ jobs:
           pip install dist/*.whl
           ./ci/check_installed_data.py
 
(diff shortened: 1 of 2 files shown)


minrk:
I want to make sure `pip install .` from an sdist succeeds if node/yarn/npm are unavailable. Is that still true? We should probably add a check for this in `check_sdist.py`, though it's a bit tricky to mock _not_ finding a command on the PATH.

manics:
Yes, it works. I've added a test using the docker.io/library/python:3.9-slim-bullseye` container which doesn't include npm or yarn.

minrk:
Perfect, thanks!

PR #4042, opened 2022-09-14
Update documentation to refer to c.JupyterHub.bind_url

### LINKED ISSUE #4020: Documentation issue: c.JupyterHub.port is deprecated but not indicated in doc
### Bug description

In the docs in url: https://jupyterhub.readthedocs.io/en/stable/getting-started/networking-basics.html
it states that: 
```
Or by placing the following lines in a configuration file, jupyterhub_config.py:
c.JupyterHub.ip = '192.168.1.2'
c.JupyterHub.port = 443
```
But in the latest Jupyterhub 2.x.x the generated config file states the above property is deprecated:
```
## The public facing port of the proxy.
#
#          This is the port on which the proxy will listen.
#          This is the only port through which JupyterHub
#          should be accessed by users.
#
#          .. deprecated: 0.9
#              Use JupyterHub.bind_url
#  Default: 8000
# c.JupyterHub.port = 8000
```

#### Expected behaviour

The readthedocs document should mention the non-deprecated property not the deprecated property.

shubham0473:
Updated docs to reflect the new Proxy IP and port configuration property in the configuration file. This is my first contribution so please let me know if I'm missing anything or if something is wrong. It would be of great help to me.

closes #4020

### DIFF
--- a/docs/source/getting-started/networking-basics.md
+++ b/docs/source/getting-started/networking-basics.md
@@ -27,6 +27,7 @@ Or by placing the following lines in a **configuration file**,
 ```python
 c.JupyterHub.bind_ip = '192.168.1.2:443'
 ```
+
 Port 443 is used in the examples since 443 is the default port for SSL/HTTPS.
 
 Configuring only the main IP and port of JupyterHub should be sufficient for
@@ -46,7 +47,6 @@ c.JupyterHub.port = 443
 
 We strongly recommend using the `bind_ip` setting instead of the deprecated
 settings, `ip` and `port`.
-
 
 ## Set the Proxy's REST API communication URL (optional)
 
--- a/docs/source/getting-started/networking-basics.md
+++ b/docs/source/getting-started/networking-basics.md
@@ -25,19 +25,28 @@ Or by placing the following lines in a **configuration file**,
 `jupyterhub_config.py`:
 
 ```python
-c.JupyterHub.ip = '192.168.1.2'
-c.JupyterHub.port = 443
+c.JupyterHub.bind_ip = '192.168.1.2:443'
 ```
-
 Port 443 is used in the examples since 443 is the default port for SSL/HTTPS.
 
 Configuring only the main IP and port of JupyterHub should be sufficient for
 most deployments of JupyterHub. However, more customized scenarios may need
 additional networking details to be configured.
 
-Note that `c.JupyterHub.ip` and `c.JupyterHub.port` are single values,
+Note that `c.JupyterHub.bind_ip` and `c.JupyterHub.port` are single values,
 not tuples or lists – JupyterHub listens to only a single IP address and
 port.
+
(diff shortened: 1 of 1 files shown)


docs/source/getting-started/networking-basics.md
```diff
@@ -25,8 +25,7 @@ Or by placing the following lines in a **configuration file**,
 `jupyterhub_config.py`:
 
 ```python
-c.JupyterHub.ip = '192.168.1.2'
-c.JupyterHub.port = 443
+c.JupyterHub.bind_ip = '192.168.1.2:443'
```
minrk, line 28: it's `bind_url` here:

```suggestion
c.JupyterHub.bind_url = 'https://192.168.1.2:443'
```

docs/source/getting-started/networking-basics.md
```diff
@@ -35,10 +34,20 @@ Configuring only the main IP and port of JupyterHub should be sufficient for
 most deployments of JupyterHub. However, more customized scenarios may need
 additional networking details to be configured.
 
-Note that `c.JupyterHub.ip` and `c.JupyterHub.port` are single values,
+Note that `c.JupyterHub.bind_ip` and `c.JupyterHub.port` are single values,
 not tuples or lists – JupyterHub listens to only a single IP address and
```
minrk, line 38:
```suggestion
Note that `c.JupyterHub.bind_url` is a single string,
not a tuple or list – JupyterHub listens to only a single IP address and
```

docs/source/getting-started/networking-basics.md
```diff
@@ -35,10 +34,20 @@ Configuring only the main IP and port of JupyterHub should be sufficient for
...
 
+**Deprecated in 0.9:** Alternatively, Proxy's main IP address and port can be set by placing the following lines in the configuration file:
+
+```python
+c.JupyterHub.ip = '192.168.1.2'
+c.JupyterHub.port = 443
+```
+
+We strongly recommend using the `bind_ip` setting instead of the deprecated
+settings, `ip` and `port`.
```
minrk, line 49:
```suggestion
We recommend using the `bind_url` setting instead of `ip` and `port`,
since `bind_url` can set ip, port, and base_url in one place.
```

minrk (commented):
Thanks for the PR! I made some suggestions inline to refer to `bind_url` (not `bind_ip`) and update some of the language. Let me know what you think!

docs/source/getting-started/networking-basics.md
ebebpl: agree with @minrk but probably understood that http:// or https:// based on no-ssl or ssl used.

docs/source/getting-started/networking-basics.md
ebebpl: agree with @minrk correction

PR #4045, opened 2022-09-16
Fix crash when removing scopes attribute from an existing role

miwig:
JupyterHub currently crashes on startup with 
`KeyError: 'scopes'` at `if not set(role_spec['scopes']).issubset(old_role.scopes):`
if the `scopes` attribute was removed from a previously defined role. Since the documentation [explicitly states](https://jupyterhub.readthedocs.io/en/stable/rbac/roles.html#defining-roles) that all fields except `name` are optional, this fixes the crash and makes removal actually work (instead of silently using the old value).


minrk:
Thanks!

PR #4046, opened 2022-09-17
mentioning limitations of spawner.environment

Neeraj-Natu:
Having been tripped on sharing across a complex dictionary across from authenticator to spawner I think it would be helpful to mention the limitation of setting only string or byte like object within `spawner.environment`. A better option that i discovered was to use `auth_state_hook` within the spawner that reads the entire auth_state within spawner after it is initialised but before it is started. I hope this would have been the promoted way to share the `auth_state` across rather than  a limited set of environment variables.


docs/source/reference/authenticators.md
```diff
@@ -247,6 +247,9 @@ class MyAuthenticator(Authenticator):
         spawner.environment['UPSTREAM_TOKEN'] = auth_state['upstream_token']
 ```
 
+Please note the `spawner.environment` expects a string or byte array set in the environment but not dictionary.
+To set anything complex other than string please look at {meth}`Spawner.auth_state_hook` available in most of the spawners which can read the entire `auth_state` dictionary.
```
minrk, line 251: I'd like to understand this. Spawner.environment accepts everything valid for an environment variable in any context (which is a dict of `string:string`). I don't think env variables can be dictionaries anywhere; that's not a JupyterHub thing, that's a unix/windows/mac process thing.

minrk: ok, I think I understand. It's this statement I want to rephrase:

> To set anything complex other than string

Because there's nothing that lets you set anything other than a string. `auth_state_hook` has identical capabilities to the `pre_spawn_start`, but it's a newer API that avoids the `await user.get_auth_state()` step and can be set by _users_ via config, not just Authenticators. It looks like this comment is saying 'pre_spawn_hook can only set strings, but auth_state_hook can set other things', which isn't accurate.

How about this:

    auth state can also be used to configure the spawner via _config_ without subclassing
    by setting `c.Spawner.auth_state_hook`. This function will be called with `(spawner, auth_state)`,
    only when auth_state is defined.
    
    For example:
    
    ```python
    def auth_state_hook(spawner, auth_state):
        spawner.environment['UPSTREAM_TOKEN'] = auth_state['upstream_token']

    c.Spawner.auth_state_hook = auth_state_hook
    ```
    
    Note that environment variable names and values are always strings, so passing multiple values means setting multiple environment variables or serializing more complex data into a single variable, e.g. as a JSON string.

Neeraj-Natu: Agreed about `auth_state_hook` being newer and avoids `user.get_auth_state()` but it can be used by the spawner for example to pass the volume and volume mount details based on user authentication, thus it is of-course upon the `spawner/custom-spawner` which implements and has the need for anything other than string.

minrk: >  but it can be used by the spawner for example to pass the volume and volume mount details based on user authentication

Yes, and the same is true of `Authenticator.pre_spawn_hook`. They are given the same information at the same time, and can do all the same things. This isn't a difference.

> it is upon the spawner/custom-spawner which implements and has the need for anything other than string.

I think I got hung up on the `environment` vs anything _else_ that can be customized on the Spawner from auth state (in both of these hooks). So it wasn't about setting the _environment_ with non-strings (which nothing can do), it was customizations _that aren't setting environment variables at all_, like other Spawner customizations. That's what I missed and the KubeSpawner example illustrates that.

minrk:
Thanks for the PR and discussion!

docs/source/reference/authenticators.md
Neeraj-Natu: Correct yes it was about the other Spawner customizations. 
This was silly of me to have not tried the same thing with `Authenticator.pre_spawn_hook` it does make sense as both are similar and would have the same functionality.

PR #4047, opened 2022-09-18
Modified users.servers type from array to object in rest-api.yml

### LINKED ISSUE #4036: Servers returned by REST API users endpoint is a dictionary (not list) of dictionaries
### Bug description

[JupyterHub's REST API docs](https://jupyterhub.readthedocs.io/en/stable/reference/rest-api.html) include an example response from [`GET /users`](https://jupyterhub.readthedocs.io/en/3.0.0/reference/rest-api.html#/default/get_users), with the value of `"servers"` depicted as a list of dictionaries.  However, my Hub's API returns it as a single dictionary of dictionaries, with the keys being the server names.  The `User` model at the bottom of the API spec also indicates that it should be a list of dictionaries.

The dictionary appears to be what's in the user model, at [apihandlers/base in APIHandler's user_model()](https://github.com/jupyterhub/jupyterhub/blob/3.0.0/jupyterhub/apihandlers/base.py#L328).

#### Expected behaviour

Actually I thought the dictionary response was normal behavior, I'd never really looked that close at the documentation before!  Is it just a documentation issue?

#### Actual behaviour

Got dictionary, not a list.

### How to reproduce

Configure JupyterHub to support named servers, start a few, and then do a `GET /users` as admin and I think you'll see a dictionary not a list.

shree256:
Fixed issue in documentation, users.servers returns list of dict which now fixed to returning dict. 
Modified users.servers type from array to object in rest-api.yml 
closes #4036

### DIFF
--- a/docs/source/_static/rest-api.yml
+++ b/docs/source/_static/rest-api.yml
@@ -1169,14 +1169,13 @@ components:
           description: Timestamp of last-seen activity from the user
           format: date-time
         servers:
-          type: array
+          type: object
           description: |
             The servers for this user.
             By default: only includes _active_ servers.
             Changed in 3.0: if `?include_stopped_servers` parameter is specified,
             stopped servers will be included as well.
-          items:
-            $ref: "#/components/schemas/Server"
+          $ref: "#/components/schemas/Server"
         auth_state:
           type: object
           properties: {}
--- a/docs/source/_static/rest-api.yml
+++ b/docs/source/_static/rest-api.yml
@@ -1175,7 +1175,8 @@ components:
             By default: only includes _active_ servers.
             Changed in 3.0: if `?include_stopped_servers` parameter is specified,
             stopped servers will be included as well.
-          $ref: "#/components/schemas/Server"
+          items:
+            $ref: "#/components/schemas/Server"
         auth_state:
           type: object
           properties: {}


docs/source/_static/rest-api.yml
```diff
@@ -1169,14 +1169,13 @@ components:
...
-          type: array
+          type: object
           description: |
             The servers for this user.
             By default: only includes _active_ servers.
             Changed in 3.0: if `?include_stopped_servers` parameter is specified,
             stopped servers will be included as well.
-          items:
-            $ref: "#/components/schemas/Server"
+          $ref: "#/components/schemas/Server"
```
manics, line 1178: I think this needs to be a Dictionary of `<server-name>:Server`, not just a single `Server`

shree256: Do I need to rollback as before??
```
items:
   $ref: "#/components/schemas/Server"
```

manics: The CI checks include a link to preview the docs
https://jupyterhub--4047.org.readthedocs.build/en/4047/reference/rest-api.html#/default/get_users__name_
I think it's going to be more complicated since it's not a list of items, it needs a wildcard(?) fieldname as the key for each server.

shree256: I doubt, whether swagger gives option to define key value to be fieldname. Please correct me if I am wrong.

PR #4049, opened 2022-09-20
Set read-only permissions in github workflows

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -82,10 +82,6 @@ jobs:
           twine upload --skip-existing dist/*
 
   publish-docker:
-    permissions:
-      contents: read # to fetch code (actions/checkout)
-      metadata: read # to get tags
-
     runs-on: ubuntu-20.04
 
     services:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,6 +28,9 @@ on:
       - "**"
   workflow_dispatch:
 
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
 jobs:
   build-release:
     runs-on: ubuntu-20.04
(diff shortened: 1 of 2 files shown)


minrk:
Thanks for the PR! We already have this setting at the repo level, so this doesn't change the workflow permissions. Does it help to have duplicate config in the workflows themselves?

consideRatio:
> have this setting at the repo level

At the github organization level even:

image

I think the main case for adding it would be that there is no doubt by anyone that we have such restrictions in place already. This is not the first PR opened about this in the jupyterhub organization.

I'm inclined towards closing and just informing that we have the tightly scoped permissions by default in the organization if PRs are opened like this across the jupyterhub github organization.

sashashura:
Oh, I see. Sorry for the inconvenience. It is invisible externally. Feel free to close.

minrk:
No worries! it's not [_quite_](https://github.com/jupyterhub/jupyterhub/actions/runs/3104916188/jobs/5029929336#step:1:17) invisible, but pretty close.

I still think it's absolutely wild that elevated permissions don't need to be opt-in on a per-workflow basis. No other CI I'm aware of has permission to do things like this by default.

PR #4050, opened 2022-09-22
Adding manage_roles feature

vladfreeze:
Adds the manage_roles parameter, which, similarly to manage_groups, obtains custom role objects through the authentication and adds them to the database with the corresponding name, scopes, users, services and groups. 

We use this mainly for automatising the role management by adding group-specific roles with group-specific permissions.


minrk:
> We use this mainly for automatising the role management by adding group-specific roles with group-specific permissions.

Doesn't this mainly mean that group membership is all you need, since roles are granted to users by their group membership?


There's a significant problem to solve in JupyterHub startup, because all roles and role assignments will be reset on every Hub launch, and incorrect until the next login for every user. Runtime-defined roles and role assignments need to be differentiated from config-defined roles and role assignments so this can be handled correctly.

Make sure to add a test that covers:

1. user gets auth-managed roles
2. hub restarts/reloads roles from config
3. user still has the right permissions

vladfreeze:
Our groups are synced with AzureAD, which means that when a user is signing in after a Hub reset, the same groups and roles will be created, which are included in the authentication. In our case the role/group persistence is not an issue and we even use the sqlite-memory database type, which completely resets on deployment.
Here is an example of how we configure the roles to be added with the post_auth_hook. We first manage the groups with manage_groups and afterwards iterate through every group and create a group role with limited scopes (such as granting the members access to the admin dashboard only for the groups they are in):

image

jupyterhub/tests/test_auth.py
```diff
@@ -591,3 +591,79 @@ async def test_auth_managed_groups(
...
+            assert name in user_roles
+            role = orm.Role.find(app.db, name)
+            app.db.delete(role)
+
+    # Check if roles are deleted after restart
+    await hub.initialize()
+    all_roles = app.db.query(orm.Role).all()
+    all_roles_names = sorted(g.name for g in all_roles)
+    for name in expected_authenticated_roles_names:
+        assert name not in all_roles_names
```
minrk, line 669: I think this is what _shouldn't_ happen. The roles should not change across restart. If the roles are deleted on restart, then behavior is inconsistent. Hub restarts without configuration changes should be minimally disruptive (ideally not at all).

jupyterhub/tests/test_auth.py
```diff
@@ -591,3 +591,79 @@ async def test_auth_managed_groups(
...
+    # Check if user gets auth-managed roles
+    with mock.patch.dict(app.tornado_settings, {"authenticator": authenticator}):
+        assert not app.db.dirty
+        all_roles = app.db.query(orm.Role).all()
+        user_roles = sorted(g.name for g in user.roles)
+        expected_authenticated_roles_names = getRoleNames(expected_authenticated_roles)
+        for name in expected_authenticated_roles_names:
+            assert name in user_roles
+            role = orm.Role.find(app.db, name)
+            app.db.delete(role)
```
minrk, line 662: I don't think this `delete` should be here, since it doesn't a real situation.

jupyterhub/user.py
```diff
@@ -238,6 +238,14 @@ def count_active_users(self):
         return counts
 
 
+def getRoleNames(role_list):
+    return [role['name'] for role in role_list]
+
+
+def getRoleObject(role_list, role_name):
+    return [role for role in role_list if role['name'] == role_name][0]
```
minrk, line 246: Python standards don't use camelCase in Python names:

```suggestion
def _get_role_names(role_list):
    return [role['name'] for role in role_list]


def _get_role_object(role_list, role_name):
    return [role for role in role_list if role['name'] == role_name][0]
```

minrk (commented):
Thanks! I think there's more to do to ensure we have consistent behavior across restarts. If we have authenticator-managed roles, I think that means we have to ensure that _no_ roles come from config, which would necessarily conflict with authenticators managing role assignments.

Either that, or track that _individual_ roles are managed by the authenticator, and exclude those particular roles from any synchronization with config.

krassowski:
@vladfreeze would you mind if taken over this PR to get it over the finish line? I can work on top of your changes for now and later either open a new PR to the main branch, or open a PR against your fork (depending on whether makes more sense here).

jupyterhub/handlers/base.py
```diff
@@ -808,7 +808,12 @@ async def auth_to_user(self, authenticated, user=None):
             group_names = authenticated.get("groups")
             if group_names is not None:
                 user.sync_groups(group_names)
-
+        # apply authenticator-managed roles
+        if self.authenticator.manage_roles:
+            print(authenticated)
+            auth_roles = authenticated.get("roles")
+            if auth_roles is not None:
+                user.sync_roles(auth_roles)
```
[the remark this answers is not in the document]
krassowski: I infer that the original intent of this PR was to have something like:

```python
{
   'auth_state': ...,
   'name': 'elephant',
   'groups': ['animals'],
   'roles': [
      {
          'name': 'custom-role',
          'description': 'My custom role',
          'scopes': [...],
          'groups': ['animals']
      }
   ]
}
```

Bu that assumes that both `groups` and `roles` will be in sync and always returned together. Even for that use case the current logic still has the downside that if `custom-role` is extended to cover `['animals', 'fungi']` it would not be reflected without restart.

I also think the API should be designed with the least surprise principle, so that the `roles` are explicitly added for the current user.

I do not know if `'users`' key should be disallowed in the managed roles. It feels to me that it is not needed in this context. Thoughts?

jupyterhub/user.py
```diff
@@ -336,6 +344,69 @@ def sync_groups(self, group_names):
...
+                            if service is not None:
+                                services.append(service)
+                    users = []
+                    if 'users' in role.keys():
+                        for user_name in role['users']:
+                            user = orm.User.find(self.db, user_name)
+                            if user is not None:
+                                users.append(user)
+                    orm_role = orm.Role(
+                        name=role_name,
```
krassowski, line 400: Creating the role directly does not validate the `role_name`. I suspect that `create_role` was not used to avoid triggering database commit too early, but I think it's possibly better to reconcile the two.

jupyterhub/user.py
```diff
@@ -336,6 +344,69 @@ def sync_groups(self, group_names):
...
+        """Synchronize roles with database"""
+        current_roles = {g.name for g in self.db.query(orm.Role).all()}
+
+        auth_roles_names = _get_role_names(auth_roles)
+
+        new_roles_names = set(auth_roles_names).difference(current_roles)
+        if new_roles_names:
+            self.log.info(f"Adding new roles {new_roles_names} to the database")
+        if auth_roles:
+            roles = self.db.query(orm.Role).all()
```
krassowski, line 357: This is querying DB for the second time

jupyterhub/user.py
```diff
@@ -336,6 +344,69 @@ def sync_groups(self, group_names):
...
+            for role_name in auth_roles_names:
+                if role_name not in existing_roles:
+                    # create roles that don't exist yet
+                    self.log.info(f"Creating new role {role_name}")
+
+                    role = _get_role_object(auth_roles, role_name)
+
+                    role_name = []
+                    if 'name' in role.keys():
+                        role_name = role['name']
```
krassowski, line 368: This is spurious. `role_name ` is already defined (this is what we iterate over), and `_get_role_object` would not return the role if the `name` would not exist.

jupyterhub/tests/test_auth.py
krassowski: Do I understand the problem correctly: when hub restarts users are not required to login again, thus the role data obtained during login on the Hub side may be incomplete or missing altogether?

Would it be preferable to:
- keep track of all active sessions, and on Hub restart force refresh `auth` state for all users  who were active when Hub got shut down, or
- make use of the existing `Authenticator.refresh_pre_spawn` config option which appears to have similar effect (on the level of individual users) as it calls `refresh_auth` with `force=True` when enabled; I imagine that `Authenticator.refresh_pre_spawn` could get a new `auto` value (coded as enum or `None`) which would mean "True if `manage_roles` is enabled".

Issue #4038, opened 2022-09-12
Named server limit per user indeed


danilopeixoto:
Would it be doable to set `named_server_limit_per_user` type as Callable as well?

```python
def named_server_limit_per_user(user):
    if user.name == 'user':
        return 10

    return 0


c.JupyterHub.named_server_limit_per_user = named_server_limit_per_user
```

### Proposed change

Open to discussion.

### Alternative options

- Introduce `named_server_limit_hook`
- Introduce `named_server_limit` attribute per user
- Introduce general quota service

### Who would use this feature?

Administrator needs to set how much server each user is limited to spawn simultaneously.

minrk:
Yes, allowing a Callable here should be just fine.

PR #4053, opened 2022-09-26
Set named_server_limit_per_user type as integer or callable

danilopeixoto:
Related to #4038.


danilopeixoto:
Hi @minrk! Added this minimal implementation.

jupyterhub/handlers/base.py
```diff
@@ -248,6 +248,17 @@ def redirect_to_server(self):
...
+    async def get_current_user_named_server_limit(self):
+        """
+        Return named server limit for current user.
+        """
+        named_server_limit_per_user = self.named_server_limit_per_user
+
+        if callable(named_server_limit_per_user):
+            return await maybe_future(named_server_limit_per_user(self))
+        else:
+            return named_server_limit_per_user
```
GeorgianaElena, line 260: The `else` here can be removed for simplicity.

```suggestion
      return named_server_limit_per_user
```

GeorgianaElena (commented):
This looks good to me @dietmarw, thanks for opening the PR 🚀 

Can you please also update the docs [here](https://github.com/jupyterhub/jupyterhub/blob/main/docs/source/reference/config-user-env.md#named-servers) with an example and maybe also add a test for `named_server_limit_per_user`  as a callable?

dietmarw:
> This looks good to me @dietmarw, thanks for opening the PR rocket

@GeorgianaElena I guess you mean @danilopeixoto :-)

GeorgianaElena:
Yes 😅 Sorry for the cofusion @dietmarw and thanks for pointing this out! I've just updated my comment.

jupyterhub/handlers/base.py
danilopeixoto: Done. Thanks!

minrk:
Thanks @danilopeixoto! And thank you @GeorgianaElena for review!

PR #4063, opened 2022-10-06
Fully resolve requested scopes in oauth

minrk:
Previously, when checking if an oauth client was allowed to request a scope or if a user was allowed to issue a token with a scope, the user must have the _exact_ scopes listed, or they would be excluded. This includes filters.

With this change, the full intersection is calculated. A critical example is e.g. read-only access to user servers, where a user might have the permissions:

```
custom:execute!user
custom:read
```
(i.e. users can execute on their own servers, but read on anyone else's)

and the server would issue tokens with:

```
custom:read!server
custom:execute:!server
```

The resulting token for the owner of this server should have the full permissions (!server intersects with !user for the owner), while a visitor will only get `custom:read!server` (read-only access to just this server because !server is a subset of all servers, and !server does _not_ intersect with !user for the visitor).

This is actually a _regression_ from 2.0, where we had special cased allowing issuing of tokens if the user had _some_ access to a particular scope, and we relied on request-time application of the filters.

I think this finally delivers everything I was going for with the custom scopes.

I also cleaned up some implicit scope behaviors that shouldn't have any practical consequences.

### DIFF
--- a/jupyterhub/scopes.py
+++ b/jupyterhub/scopes.py
@@ -95,7 +95,7 @@ scope_definitions = {
     },
     'read:servers': {
         'description': 'Read users’ names and their server models (excluding the server state).',
-        'subscopes': ['read:users:name'],
+        'subscopes': [],
     },
     'delete:servers': {'description': "Stop and delete users' servers."},
     'tokens': {
@@ -463,14 +463,6 @@ def _expand_scope(scope):
         expanded_scopes = {
             f"{scope_name}!{filter_}" for scope_name in expanded_scope_names
         }
-        # special handling of server filter
-        # any read access via server filter includes permission to read the user's name
-        resource, _, value = filter_.partition('=')
-        if resource == 'server' and any(
-            scope_name.startswith("read:") for scope_name in expanded_scope_names
-        ):
-            username, _, server = value.partition('/')
-            expanded_scopes.add(f'read:users:name!user={username}')
     else:
         expanded_scopes = expanded_scope_names
 
--- a/jupyterhub/scopes.py
+++ b/jupyterhub/scopes.py
@@ -561,6 +561,71 @@ def expand_scopes(scopes, owner=None, oauth_client=None):
     return frozenset(reduce_scopes(expanded_scopes))
 
 
+def _resolve_requested_scopes(requested_scopes, have_scopes, user, client, db):
+    """Resolve requested scopes for an OAuth token
+
+    Intersects requested scopes with user scopes.
+
+    First, at the raw scope level,
+    then if some scopes remain, intersect expanded scopes.
+
+    Args:
+        requested_scopes (set):
+            raw scopes being requested.
+        user (orm.User):
+            user for whom the scopes will be issued
+        client (orm.OAuthClient):
+            oauth client which will own the token
+
+    Returns:
+        (allowed_scopes, excluded_scopes): sets of allowed scopes and excluded scopes from the request
+    """
+
+    allowed_scopes = requested_scopes.intersection(have_scopes)
+    excluded_scopes = requested_scopes.difference(have_scopes)
+
+    if not excluded_scopes:
+        # simple intersection worked, all scopes granted
+        return (allowed_scopes, excluded_scopes)
+
+    # if we got here, some scopes were excluded.
+    # resolve fully expanded scopes to make sure scope intersections are properly allowed.
+    expanded_allowed = expand_scopes(allowed_scopes, user, client)
+    expanded_have = expand_scopes(have_scopes, user, client)
+    # compute one at a time so we can keep the abbreviated scopes
+    # if they are a subset of user scopes (e.g. requested !server, have !user)
+    for scope in list(excluded_scopes):
+        expanded_excluded = expand_scopes({scope}, user, client)
+        # don't check already-allowed scopes
+        expanded_excluded -= expanded_allowed
+        if expanded_excluded:
+            allowed_intersection = _intersect_expanded_scopes(
+                expanded_excluded, expanded_have, db=db
+            )
+        else:
+            allowed_intersection = set()
+
+        if allowed_intersection == expanded_excluded:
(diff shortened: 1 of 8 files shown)


jupyterhub/scopes.py
```diff
@@ -569,6 +561,71 @@ def expand_scopes(scopes, owner=None, oauth_client=None):
...
+    First, at the raw scope level,
+    then if some scopes remain, intersect expanded scopes.
+
+    Args:
+        requested_scopes (set):
+            raw scopes being requested.
+        user (orm.User):
+            user for whom the scopes will be issued
+        client (orm.OAuthClient):
+            oauth client which will own the token
```
consideRatio, line 578: Can you add a description for `have_scopes`?

jupyterhub/scopes.py
```diff
@@ -569,6 +561,71 @@ def expand_scopes(scopes, owner=None, oauth_client=None):
...
+
+    Returns:
+        (allowed_scopes, excluded_scopes): sets of allowed scopes and excluded scopes from the request
+    """
+
+    allowed_scopes = requested_scopes.intersection(have_scopes)
+    excluded_scopes = requested_scopes.difference(have_scopes)
```
consideRatio, line 585: There is a mix of terminology here. `allowed` / `disallowed` is used outside this function, while inside it `allowed` and `excluded` is used.

I appreciate the `allowed` and `disallowed` overall, and currently lean towards thinking it would be an improvement to update the function naming to speak of `disallowed_scopes`.

I'm open to retaining naming as it is though if you think it makes sense.

consideRatio (commented):
Looks good from what I can tell in an initial review pass, small comments about documentation and namings. I'll try to get another review pass later today where I go through all the code, I have not done so at this time.

consideRatio:
Could you comment the change regarding the `read:users:name` scope that I think was given special treatment before? I don't understand that situation.

minrk:
> I appreciate the allowed and disallowed overall,

Yes, we had two naming conventions in the two places this was used because the use case was slightly different - in one case 'excluded' made sense, because we would proceed with the allowed subset, _excluding_ the disallowed subset. In the other case, if anything was disallowed, an error occurred. I've switched to disallowed everywhere, which is clearer, thanks for the feedback!

> Could you comment the change regarding the read:users:name scope that I think was given special treatment before?

Yes, and I'm reconsidering the change now. I've updated the PR.

`read:users:name` was added as a sub-scope to `read:servers` because that allowed access to the `/api/users/:name` endpoint, which is the only way to list all the servers of a user. There is special handling, because subscopes with filters preserve the filter (i.e. `read:servers!server=username/servername` expands to `read:users:name!server=username/servername`). But this doesn't grant access to the `/api/users/:name` endpoint if the filter is on `!server`, because the filter is on server, not the target resource, which is a user. Hence the special case of expanding `read:servers!server=username/servername` to include `read:users:name!user=username`.

I think that was a mistake, and my initial solution was to completely remove the connection from read:servers->read:users:name, because it still makes sense if you have `read:servers` permissions _at the level of user or higher_. But instead, I think the right answer is only to remove the special handling of elevating the filter from `read:servers!server` to `read:users:name!user`, so that the only change is that single-server permission `read:servers!server` does _not_ grant access to the user API, but all other cases (e.g. user or group filters or no filters) are unaffected. I've updated the PR to reflect this.

jupyterhub/tests/test_scopes.py
```diff
@@ -1200,3 +1208,88 @@ def format_scopes(scopes):
...
+
+    allowed, excluded = _resolve_requested_scopes(
+        requested_scopes,
+        have_scopes,
+        user=user.orm_user,
+        client=oauth_client,
+        db=db,
+    )
+    assert allowed == expected_allowed
+    assert excluded == expected_excluded
```
consideRatio, line 1295:
```suggestion
    expected_disallowed = format_scopes(expected_disallowed)

    allowed, disallowed = _resolve_requested_scopes(
        requested_scopes,
        have_scopes,
        user=user.orm_user,
        client=oauth_client,
        db=db,
    )
    assert allowed == expected_allowed
    assert disallowed == expected_disallowed
```

jupyterhub/tests/test_scopes.py
```diff
@@ -1200,3 +1208,88 @@ def format_scopes(scopes):
...
+    ],
+)
+def test_resolve_requested_scopes(
+    app,
+    user,
+    group,
+    requested_scopes,
+    have_scopes,
+    expected_allowed,
+    expected_excluded,
```
consideRatio, line 1255:
```suggestion
    expected_disallowed,
```

consideRatio (approved):
LGTM. I renamed "excluded" to "disallowed" also in the tests in some change suggestions.

jupyterhub/tests/test_scopes.py
consideRatio, line 1295:
```suggestion
    expected_disallowed = format_scopes(expected_disallowed)
    allowed, disallowed = _resolve_requested_scopes(
        requested_scopes,
        have_scopes,
        user=user.orm_user,
        client=oauth_client,
        db=db,
    )
    assert allowed == expected_allowed
    assert disallowed == expected_disallowed
```

jupyterhub/tests/test_scopes.py
```diff
@@ -1200,3 +1208,87 @@ def format_scopes(scopes):
     expanded = expand_scopes(scopes, owner=user.orm_user, oauth_client=oauth_client)
     assert isinstance(expanded, frozenset)
     assert sorted(expanded) == sorted(expected)
+
+
+@pytest.mark.parametrize(
+    "requested_scopes, have_scopes, expected_allowed, expected_excluded",
```
consideRatio, line 1214:
```suggestion
    "requested_scopes, have_scopes, expected_allowed, expected_disallowed",
```

minrk:
Thanks @consideRatio!

PR #4068, opened 2022-10-09
Update setup.rst

Goodiec:
This PR improves the setup page and is part of [issue #41](https://github.com/jupyterhub/outreachy/issues/41)

### DIFF
--- a/docs/source/contributing/setup.rst
+++ b/docs/source/contributing/setup.rst
@@ -135,7 +135,7 @@ To simplify testing of JupyterHub, it’s helpful to use
 authenticator and SimpleLocalProcessSpawner instead of the default spawner.
 
 There is a sample configuration file that does this in
-``testing/jupyterhub_config.py``. To launch Jupyterhub with this
+``testing/jupyterhub_config.py``. To launch JupyterHub with this
 configuration:
 
 .. code:: bash
--- a/docs/source/contributing/setup.rst
+++ b/docs/source/contributing/setup.rst
@@ -7,7 +7,7 @@ Setting up a development install
 System requirements
 ===================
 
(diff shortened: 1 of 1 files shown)


docs/source/contributing/setup.rst
```diff
@@ -137,7 +135,7 @@ To simplify testing of JupyterHub, it’s helpful to use
 authenticator and SimpleLocalProcessSpawner instead of the default spawner.
 
 There is a sample configuration file that does this in
-``testing/jupyterhub_config.py``. To launch jupyterhub with this
+``testing/jupyterhub_config.py``. To launch Jupyterhub with this
```
minrk, line 138:
```suggestion
``testing/jupyterhub_config.py``. To launch JupyterHub with this
```

minrk (commented):
Thank you for the suggestions! I made one inline suggestion, and then we can merge.

docs/source/contributing/setup.rst
GeorgianaElena: @Goodiec, no worry, it's hard to spot a small difference of only one letter in these diffs.

`JupyterHub` has the upper letter `H` in the middle of the name, from `Hub`. 
Min's suggestion is to change  the name to `JupyterHub` (notice the upper letter `H`). In your commit it's currently `Jupyterhub` with the small letter `h`.

Goodiec: Thank you @GeorgianaElena, I've made the correction.

GeorgianaElena: > Also, is it Ok if I help review the PRs of others?

Of course, that would be great! Thanks @Goodiec

PR #4070, opened 2022-10-09
Modifications to Technical Overview Docs

Mackenzie-OO7:
In this PR, I:
- fixed typos and restructured some sentences to be better understood 
- fixed the link leading to the Security Settings Page
- embedded links containing definitions to terms that might be unfamiliar or unknown

### DIFF
--- a/docs/source/reference/technical-overview.md
+++ b/docs/source/reference/technical-overview.md
@@ -110,7 +110,7 @@ working directory:
   This file needs to persist so that a **Hub** server restart will avoid
   invalidating cookies. Conversely, deleting this file and restarting the server
   effectively invalidates all login cookies. The cookie secret file is discussed
-  in the [Cookie Secret section of the Security Settings document](../getting-started/security-basics.md).
+  in the [Cookie Secret section of the Security Settings document](../getting-started/security-basics.rst).
 
 The location of these files can be specified via configuration settings. It is
 recommended that these files be stored in standard UNIX filesystem locations,
--- a/docs/source/reference/technical-overview.md
+++ b/docs/source/reference/technical-overview.md
@@ -2,7 +2,7 @@
 
 The **Technical Overview** section gives you a high-level view of:
 
-- JupyterHub's Subsystems: Hub, Proxy, Single-User Notebook Server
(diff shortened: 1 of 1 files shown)


docs/source/reference/technical-overview.md
```diff
@@ -102,7 +101,7 @@ only.
 By default, starting JupyterHub will write two files to disk in the current
 working directory:
 
-- `jupyterhub.sqlite` is the SQLite database containing all of the state of the
+- `jupyterhub.sqlite` is the SQLite database containing all of the states of the
```
minrk, line 104:
```suggestion
- `jupyterhub.sqlite` is the SQLite database containing all of the state of the
```

Usually we call the 'state' of an application a single thing.

docs/source/reference/technical-overview.md
```diff
@@ -72,20 +71,20 @@ When a user accesses JupyterHub, the following events take place:
...
   the single-user notebook server.
 
-The single-user server identifies the user with the Hub via OAuth:
+How does the single-user server identify the user with the Hub via OAuth?
 
-- on request, the single-user server checks a cookie
-- if no cookie is set, redirect to the Hub for verification via OAuth
-- after verification at the Hub, the browser is redirected back to the
+- On request, the single-user server checks a cookie
+- If no cookie is set, it redirects to the Hub for verification via OAuth
```
minrk, line 83: Maybe a good idea to minimize pronouns when there are so many things being talked about: 

```suggestion
- If no cookie is set, the single-user server redirects to the Hub for verification via OAuth
```

Even though it's redundant with the previous bullet, in this particular doc I think it's good to minimize possible sources of "Wait, what is 'it' here?"

minrk (commented):
Thank you, this is great! I left some suggestions in-line, which you can accept or modify.

GeorgianaElena:
I believe this is now ready to go 🚀 Thank you @Mackenzie-OO7, great work! 🎉

Mackenzie-OO7:
> I believe this is now ready to go 🚀 Thank you @Mackenzie-OO7, great work! 🎉

Thank you too! It was fun😂

PR #4071, opened 2022-10-09
Improve rbac/roles documentation

NPDebs:
This PR fixes, in part, issue [#41](https://github.com/jupyterhub/outreachy/issues/41).
In order to promote comprehension of the documentation, I read and made the following changes to "roles.md" file: 

- Fix up grammar and eliminate typos,
- Remove redundant/repetitive words,
- Adjust punctuation,
- Use initial caps where necessary,
- Add meaning of abbreviation.

### DIFF
--- a/docs/source/rbac/roles.md
+++ b/docs/source/rbac/roles.md
@@ -1,4 +1,5 @@
 (roles)=
+
 # Roles
 
 JupyterHub provides four (4) roles that are available by default:
@@ -42,6 +43,7 @@ A group does not require any role, and has no roles by default. If a user is a m
 A token’s permissions are evaluated based on their owning entity. Since a token is always issued for a user or service, it can never have more permissions than its owner. If no specific scopes are requested for a new token, the token is assigned the scopes of the `token` role.
 
 (define-role-target)=
+
 ## Defining Roles
 
 Roles can be defined or modified in the configuration file as a list of dictionaries. An example:
@@ -110,6 +112,7 @@ If no scopes are defined for _new role_, JupyterHub will raise a warning. Provid
 In case the role with a certain name already exists in the database, its definition and scopes will be overwritten. This holds true for all roles except the `admin` role, which cannot be overwritten; an error will be raised if trying to do so. All the role bearers permissions present in the definition will change accordingly.
 
 (overriding-default-roles)=
+
 ### Overriding Default Roles
 
 Role definitions can include those of the "default" roles listed above (admin excluded),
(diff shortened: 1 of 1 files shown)


docs/source/rbac/roles.md
```diff
@@ -1,8 +1,6 @@
-(roles)=
-
```
minrk, line 2: I believe this  is actually required for cross-linking from other documents, and shouldn't be removed.

docs/source/rbac/roles.md
```diff
@@ -13,11 +11,11 @@ JupyterHub provides four roles that are available by default:
 **These roles cannot be deleted.**
 ```
 
-These default roles have a default collection of scopes,
-but you can define the scopes associated with each role (excluding admin) to suit your needs,
+These roles are so named because they have a default collection of scopes.
```
minrk, line 14:
```suggestion
We call these 'default' roles because they are available by default and have a default collection of scopes.
```

docs/source/rbac/roles.md
```diff
@@ -31,19 +29,17 @@ Roles can be assigned to the following entities:
...
-Services do not have a default role. Services without roles have no access to the guarded API end-points, so most services will require assignment of a role in order to function.
+Services do not have a default role. Services without roles have no access to the guarded API end-points. So, most services will require assignment of a role in order to function.
 
 **Groups** \
 A group does not require any role, and has no roles by default. If a user is a member of a group, they automatically inherit any of the group's permissions (see {ref}`resolving-roles-scopes-target` for more details). This is useful for assigning a set of common permissions to several users.
 
 **Tokens** \
 A token’s permissions are evaluated based on their owning entity. Since a token is always issued for a user or service, it can never have more permissions than its owner. If no specific scopes are requested for a new token, the token is assigned the scopes of the `token` role.
 
-(define-role-target)=
```
minrk, line 45: Same here and everywhere - none of these should be removed.

A good tip in general: If you're not sure what a bit of code is, probably don't remove it. It's likely there for a reason. It's always okay to ask!

minrk (commented):
Thanks for the suggestions! I've left some feedback for you to incorporate.

NPDebs:
> Thanks for the suggestions! I've left some feedback for you to incorporate.

Thank you, @minrk for taking the time to thoroughly review this PR. I'll make the necessary changes right away.

docs/source/rbac/roles.md
NPDebs: Ah, yes! I was going to ask.

I saw this, and didn't know what the purpose was. Naturally, I took it out...instead of asking for clarification. 🤦🏽‍♀️

In hindsight, that's a really dumb thing to do.

docs/source/rbac/roles.md
NPDebs: Well, here's a tip that might save my job one day. 😅 Thanks a million!

NPDebs:
> Thank you, @NPDebs!

Yaaay! Thanks for merging. 🙈🤸🏽‍♀️🚀

You're welcome. 🙌🏼

PR #3566, opened 2021-08-03
explicit DB rollback for 500 errors

nsshah1288:
This PR adds explicit `try`, `except`, and `db.rollback()` when JupyterHub DB is in a bad state. We made these changes in a 'brute force' approach, by searching our logs and adding an explicit `db.rollback()` wherever the error first occurred. It has worked well (no manual restarts required) in our production JupyterHub over the last month, where we are hosted on EKS, using Amazon Aurora PostgreSQL Serverless and JupyterHub version 1.2.1.

Posts related to this issue:
- https://discourse.jupyter.org/t/jupyterhub-500-error-ssl-connection-has-been-closed-unexpectedly-tornado-ioloop/9106/5
- https://github.com/jupyterhub/jupyterhub/issues/1626

Please let us know if there is a better or more standardized approach that we should follow for this. Thanks!

Issue #1626, opened 2018-01-19
Jupyterhub is not resilient with external MySQL database


ckbhatt:
We are using Jupyterhub with an external mariadb(open source fork of mysql) database. In the event of a temporary network failure, Jupyterhub gets stuck and do not recover.

**How to reproduce the issue on mac**

1. Install Mariadb locally
`brew install mariadb`
2. Remove from launchctl for ease of stop and start
`launchctl unload -w ~/Library/LaunchAgents/homebrew.mxcl.mariadb.plist`
3. Start mariadb
`brew services start mariadb`
4. Create a database named jh_test
5. Create a user test/test in mac.
6. Run jupyter hub locally with following config file.
```
c = get_config()
c.Authenticator.whitelist = {'test'}
c.Authenticator.admin_users = {'test'}
c.JupyterHub.db_url = 'mysql+mysqlconnector://{}:{}@{}/{}{}'.format("root","","localhost","jh_test","")
```
7. Login using test/test and start jupyter server.
8. Stop mysql
`brew services stop mariadb`
9. Go to browser and click on Control Panel button. You'll get following exception:
```
[E 2018-01-19 04:14:19.920 JupyterHub web:1518] Exception in exception handler
    Traceback (most recent call last):
      File "/Users/bhatt/anaconda3/lib/python3.6/site-packages/tornado/web.py", line 1516, in _execute
        self._handle_request_exception(e)
      File "/Users/bhatt/anaconda3/lib/python3.6/site-packages/tornado/web.py", line 1570, in _handle_request_exception
        self.send_error(500, exc_info=sys.exc_info())
      File "/Users/bhatt/anaconda3/lib/python3.6/site-packages/tornado/web.py", line 1040, in send_error
        self.finish()
      File "/Users/bhatt/anaconda3/lib/python3.6/site-packages/jupyterhub/handlers/base.py", line 105, in finish
        super().finish(*args, **kwargs)
      File "/Users/bhatt/anaconda3/lib/python3.6/site-packages/tornado/web.py", line 993, in finish
        self._log()
      File "/Users/bhatt/anaconda3/lib/python3.6/site-packages/tornado/web.py", line 1540, in _log
        self.application.log_request(self)
      File "/Users/bhatt/anaconda3/lib/python3.6/site-packages/tornado/web.py", line 2053, in log_request
        self.settings["log_function"](handler)
[...]

PR #4076, opened 2022-10-10
Backport PR #3566 to 1.x series

mriedem:
This is a backport of PR #3566 to the 1.x branch in
order to resolve the same issue seen in the 1.5.0 release.

There was a minor merge conflict in c29a5ca4c but the
resolution was trivial.

Related #1626

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@ repos:
     hooks:
       - id: reorder-python-imports
   - repo: https://github.com/psf/black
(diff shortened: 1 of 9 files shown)


mriedem:
My hope here is that a 1.5.1 release could contain this, but I'd also understand if the 1.x series is closed for non-security issues.

minrk:
Since it's auto-formatting, I'm not worried about it. Thanks for wading through that!

I can try to do a 1.5.1 later next week, perhaps.

PR #4077, opened 2022-10-10
rephrased a word

ArafatAbdussalam:


docs/source/reference/websecurity.md
```diff
@@ -16,7 +16,7 @@ JupyterHub is designed to be a _simple multi-user server for modestly sized
 groups_ of **semi-trusted** users. While the design reflects serving semi-trusted
 users, JupyterHub is not necessarily unsuitable for serving **untrusted** users.
 
-Using JupyterHub with **untrusted** users does mean more work by the
+Using JupyterHub with **untrusted** users does involves more work by the
```
Mackenzie-OO7, line 19: I think "Administrator" here is single, so you should use "involve" instead of "involves"

PR #4079, opened 2022-10-10
Improve the documentation about log messages

mahamtariq58:
Kindly review my pull request and suggest how to improve.

### DIFF
--- a/docs/source/admin/log-messages.md
+++ b/docs/source/admin/log-messages.md
@@ -4,7 +4,7 @@ When debugging errors and outages, looking at the logs emitted by
 JupyterHub is very helpful. This document intends to describe some common
 log messages, what they mean and what are the most common causes that generated them, as well as some possible ways to fix them.
 
-## Failing suspected API request to run-down server
+## Failing suspected API request to not-running server
 
 ### Example
 
--- a/docs/source/admin/log-messages.md
+++ b/docs/source/admin/log-messages.md
@@ -2,7 +2,7 @@
 
 When debugging errors and outages, looking at the logs emitted by
 JupyterHub is very helpful. This document intends to describe some common
-log messages, and what necessary actions to take.
+log messages, what they mean and what are the most common causes that generated them, as well as some possible ways to fix them.
 
 ## Failing suspected API request to run-down server
 
--- a/docs/source/admin/log-messages.md
+++ b/docs/source/admin/log-messages.md
@@ -48,7 +48,7 @@ This log message is benign, and there is usually no action for you to take.
 JupyterHub requires the `jupyterhub` python package installed inside the image or
 environment, the user server starts in. This message indicates that the version of
 the `jupyterhub` package installed inside the user image or environment is not
-the same as the JupyterHub server itself. This is not necessarily always a
+the same as the JupyterHub server's version itself. This is not necessarily always a
 problem - some version drift is mostly acceptable, and the only two known cases of
 breakage are across the 0.7 and 2.0 version releases. In those cases, issues pop
 up immediately after upgrading your version of JupyterHub, so **always check the JupyterHub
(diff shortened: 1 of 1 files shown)


docs/source/admin/log-messages.md
```diff
@@ -1,33 +1,30 @@
 # Interpreting common log messages
 
 When debugging errors and outages, looking at the logs emitted by
-JupyterHub is very helpful. This document tries to document some common
-log messages, and what they mean.
+JupyterHub is very helpful. This document intends to describe some common
+log messages, and what necessary actions to take.
```
GeorgianaElena, line 5: This section of the docs focus on these three main areas: explain the log message, list a some of of the most common causes that generate this message, and what should be the first steps to take in order to fix this cause.

I believe it makes sense to mention all three of them:

```suggestion
log messages, what they mean and what are the most common causes that generated them, as well as some possible ways to fix them.
```

docs/source/admin/log-messages.md
```diff
@@ -1,33 +1,30 @@
...
 
-## Failing suspected API request to not-running server
+## Failing suspected API request to run-down server
```
GeorgianaElena, line 7: Let's keep the same naming here, to match the one in the actual log message (you can find it in the example below). This will help with making this section more visible for people when searching after a key phrase in their logs.

```suggestion
## Failing suspected API request to not-running server
```

docs/source/admin/log-messages.md
```diff
@@ -1,33 +1,30 @@
...
 
-This likely means is that the user's server has stopped running but they
-still have a browser tab open. For example, you might have 3 tabs open, and shut
-your server down via one. Or you closed your laptop, your server was
-culled for inactivity, and then you reopen your laptop again! The
-client side code (JupyterLab, Classic Notebook, etc) does not know
-yet that the server is dead, and continues to make some API requests.
+The possible reason may be that the user's server has stopped running
+but they still have a browser tab open. For example, you might have 3 tabs open and you shut
+the server down via one. The other possible reason could be that you closed your laptop and the server was culled for inactivity, then reopen the laptop! However, the client-side code (JupyterLab, Classic Notebook, etc) doesn't interpret the shut-down server and continues to make some API requests.
```
GeorgianaElena, line 21: Let's put each main idea into it's own row to make it more easier to read and keep the same tense of verbs (`reopen` ➡️  `reopened`) 

```suggestion
the server down via one.
The other possible reason could be that you closed your laptop and the server was culled for inactivity, then reopened the laptop!
However, the client-side code (JupyterLab, Classic Notebook, etc) doesn't interpret the shut-down server and continues to make some API requests.
```

docs/source/admin/log-messages.md
```diff
@@ -47,9 +44,9 @@ This log message is benign, and there is usually no action for you to take.
 ### Cause
 
 JupyterHub requires the `jupyterhub` python package installed inside the image or
-environment the user server starts in. This message indicates that the version of
+environment, the user server starts in. This message indicates that the version of
 the `jupyterhub` package installed inside the user image or environment is not
-the same version as the JupyterHub server itself. This is not necessarily always a
+the same as the JupyterHub server itself. This is not necessarily always a
```
GeorgianaElena, line 49: I know it's a bit repetitive, but I believe it's worth being as explicit as possible to eliminate all confusion.

```suggestion
the same as the JupyterHub server's version itself. This is not necessarily always a
```

GeorgianaElena (commented):
Thank you @mahamtariq58, this is great 🌻! I left some suggestions in-line, which you can accept or modify.

mahamtariq58:
@GeorgianaElena  Thank you so much for reviewing. I will surely incorporate your suggestions.

GeorgianaElena:
Thank you @mahamtariq58 🎉

PR #4086, opened 2022-10-11
Updated log message docs

ArafatAbdussalam:
I improved the log message subsection of the documentation by making the page more user friendly


docs/source/admin/log-messages.md
```diff
@@ -1,22 +1,20 @@
 # Interpreting common log messages
 
-When debugging errors and outages, looking at the logs emitted by
-JupyterHub is very helpful. This document tries to document some common
-log messages, and what they mean.
+In this subsection of the documentation, you will get to understand the meaning of common log messages and how to resolve them. When debugging errors and outages, it is very helpful to look at the logs emitted by JupyterHub.
```
minrk, line 3: I think the earlier order was preferable: problem statement ("when debugging...") before solution ("explaining common log messages")

docs/source/admin/log-messages.md
```diff
@@ -1,22 +1,20 @@
...
 
 ## Failing suspected API request to not-running server
 
 ### Example
 
-Your logs might be littered with lines that might look slightly scary
+When your screen displays the log message as shown below, you need not to get scared.
```
minrk, line 9: It's not about the screen, it's the logs. If we're keeping "scary", think it's better to keep it as a property of the logs, rather than assigning it to the reader.

docs/source/admin/log-messages.md
```diff
@@ -1,22 +1,20 @@
...
 
 ```
 [W 2022-03-10 17:25:19.774 JupyterHub base:1349] Failing suspected API request to not-running server: /hub/user/<user-name>/api/metrics/v1
 ```
 
-### Most likely cause
+### Cause
 
-This likely means is that the user's server has stopped running but they
+This likely cause is that the user's server has stopped running but they
```
minrk, line 17: We need either "means that" or "The likely cause is", but not "This likely cause".

I like "This means":

```suggestion
This likely means that the user's server has stopped running but they
```

docs/source/admin/log-messages.md
ArafatAbdussalam: Thanks for the correction mnirk. I have re-written it.

docs/source/admin/log-messages.md
ArafatAbdussalam: Resolved

docs/source/admin/log-messages.md
ArafatAbdussalam: Changed

minrk:
Thank you!

PR #4091, opened 2022-10-12
Update config-user-env.md

Goodiec:
This PR improves the 'Configuring user environments' doc and is part of [issue 41](https://github.com/jupyterhub/outreachy/issues/41)

### DIFF
--- a/docs/source/reference/config-user-env.md
+++ b/docs/source/reference/config-user-env.md
@@ -7,7 +7,7 @@ environment in some way.
 Since the `jupyterhub-singleuser` server extends the standard Jupyter notebook
 server, most configuration and documentation that applies to Jupyter Notebook
 applies to the single-user environments. Configuration of user environments
-typically does not occur through JupyterHub itself, but rather through the system-wide 
+typically does not occur through JupyterHub itself, but rather through the system-wide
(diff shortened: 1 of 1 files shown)


docs/source/reference/config-user-env.md
```diff
@@ -30,7 +30,7 @@ system-wide or in a shared environment.
 This installation location should always be in the same environment that
 `jupyterhub-singleuser` itself is installed in, and must be _readable and
 executable_ by your users. If you want users to be able to install additional
-packages, it must also be _writable_ by your users.
+packages, they must also be _writable_ by your users.
```
GeorgianaElena, line 33: `it` is actually the correct pronoun here, because the statement refers to the `instalation location` that must be `writable` and not to the users. Let's instead be explicit to avoid any future confusion:

```suggestion
packages, the installation location must also be _writable_ by your users.
```

GeorgianaElena (commented):
Thank you @Goodiec, this is great! I left a suggestion in-line, which you can accept or modify.

docs/source/reference/config-user-env.md
Goodiec: Done, thank you!

minrk:
Thank you, and thanks @GeorgianaElena for the review!

PR #4094, opened 2022-10-12
Fixed typos and added punctuations

Busayo-ojo:


docs/source/getting-started/security-basics.rst
```diff
@@ -233,7 +233,7 @@ jupyterhub-session-id
 This is a random string, meaningless in itself, and the only cookie
 shared by the Hub and single-user servers.
 
-Its sole purpose is to coordinate logout of the multiple OAuth cookies.
+Its sole purpose is to coordinate the logout of multiple OAuth cookies.
```
GeorgianaElena, line 236: Let's keep both the `the` articles here instead. I believe it makes sense in this context to emphasize the number of cookies.

```suggestion 
Its sole purpose is to coordinate the logout of the multiple OAuth cookies.
```

GeorgianaElena (commented):
Thank you @Busayo-ojo 🌻 ! I left a suggestions in-line, which you can accept or modify.

docs/source/getting-started/security-basics.rst
Busayo-ojo: Hi @GeorgianaElena Alright thank you. I'll take a look at the suggestions.

PR #4097, opened 2022-10-12
Modifications to URLs docs

Mackenzie-OO7:
@GeorgianaElena In this PR, I:

- properly punctuated and restructured some sentences to be better understood
- embedded links containing definitions to terms that might be unfamiliar or unknown
- fixed typos


minrk:
Wonderful, thank you @Mackenzie-OO7!

PR #4101, opened 2022-10-12
Updated use-cases.md

emmanuella194:
Added reference to [RBCA framework]
Emphasized on the solution on [service to cull idle servers]

### DIFF
--- a/docs/source/contributing/docs.rst
+++ b/docs/source/contributing/docs.rst
@@ -1,13 +1,13 @@
 .. _contributing/docs:
 
 ==========================
-Contributing Documentation
+#Contributing Documentation
 ==========================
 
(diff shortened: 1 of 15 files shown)


emmanuella194:
I added reference to the rbca framework and I emphasized(bolden) the service to cull idle servers

minrk:
@emmanuella194 this PR now edits a very large number of files, which has introduced conflicts. Do you want to perhaps revert back to only changing use-cases.md?

emmanuella194:
> @emmanuella194 this PR now edits a very large number of files, which has introduced conflicts. Do you want to perhaps revert back to only changing use-cases.md?

Alright, i would revert to the changes made on use-cases.md

PR #4102, opened 2022-10-12
Updated ReadMe

KaluBuikem:


README.md
```diff
@@ -56,7 +56,7 @@ Basic principles for operation are:
   servers.
 
 JupyterHub also provides a
-[REST API][]
+[REST API]:https://jupyterhub.readthedocs.io/en/latest/reference/rest-api.html
```
Mackenzie-OO7, line 59: Hi, I suggest the REST API link be embedded in "REST API".

KaluBuikem: Thanks a lot for the suggestion I would implement it immediately

Mackenzie-OO7: > Thanks a lot for the suggestion I would implement it immediately

You're welcome 👍🏾

README.md
```diff
@@ -56,10 +56,10 @@ Basic principles for operation are:
   servers.
 
 JupyterHub also provides a
-[REST API][]
+[REST API](https://jupyterhub.readthedocs.io/en/latest/reference/rest-api.html)
```
minrk, line 59: Using markdown references this way was deliberate. This is not a change we should make.

PR #4106, opened 2022-10-13
Edited and restructured `server-api` file

alwasega:
I have edited and restructured some parts of the page to make it tighter and as per the Diataxis model. 

- I would suggest changing the title of the page to `Managing servers using the JupyterHub API` and not `Starting servers using the JupyterHub API`. `Starting` a server is only one of the steps covered in the article and it does not, therefore, give a comprehensive picture of what the article is about. However, I have not changed the title as doing so might break something else somewhere. 
- Later in the article, I have used `Progress API` and not `progress API`, which breaks the sentence case styling of (sub)headings. However, I feel using a proper noun for `Progress` makes more sense given the context. I am open to corrections on this though.


docs/source/reference/server-api.md
```diff
@@ -1,37 +1,29 @@
...
-This document provides an example of working with the JupyterHub API to
-manage servers for users.
-In particular, we will cover how to:
-
-1. [check status of servers](checking)
-2. [start servers](starting)
-3. [wait for servers to be ready](waiting)
-4. [communicate with servers](communicating)
-5. [stop servers](stopping)
+Sometimes, such as when working with applications such as [BinderHub](https://binderhub.readthedocs.io), it may be necessary to launch Jupyter-based services on behalf of your users. Doing so can be achieved through JupyterHub's [REST API](../reference/rest.md), which allows one to launch and manage servers on behalf of users through API calls instead of through the JupyterHub UI. In doing so, you can take advantage of other user/launch/lifecycle patterns that are not natively supported by JupyterHub, all without the need to develop the server management features of JupyterHub Spawners and/or Authenticators.
```
minrk, line 3: Let's avoid `such as` twice in one clause. Maybe `like BinderHub...`? Or just remove the first `such as` for `Sometimes, when working with...`

docs/source/reference/server-api.md
```diff
@@ -1,37 +1,29 @@
...
+
+This tutorial goes through the processes involved while working with the JupyterHub API to manage servers for users. In particular, it covers how to:
+
+1. [Check the status of servers](checking)
+2. [Start servers](starting)
+3. [Wait for servers to be ready](waiting)
+4. [Communicate with servers](communicating)
+5. [Stop servers](stopping)
+
+In the end, we also provide a sample Python code that can be used to implement these steps.
```
minrk, line 13:
```suggestion
In the end, we also provide sample Python code that can be used to implement these steps.
```

docs/source/reference/server-api.md
```diff
@@ -119,67 +101,50 @@ while the server is not ready yet:
   }
 ```
 
-Note that `ready` is false and `pending` is `spawn`.
-This means that the server is not ready
-(attempting to access it may not work)
-because it isn't finished spawning yet.
-We'll get more into that below in [waiting for a server][].
+Note that `ready` is assigned `false` and `pending` is assigned `spawn`, meaning that the server is not ready and attempting to access it may not work as it is still in the process of spawning.
```
minrk, line 104: I'm not sure `assigned` is helpful here. I would still say `ready is false`. But if you do find we need different language, I would probably pick something like 'has the value'. But in short, I think `is` is clear and concise and standard language for a field having a value.

Can you speak more to removing the link? It seems useful to keep it here that we're going to get more into that later.

docs/source/reference/server-api.md
```diff
@@ -1,37 +1,29 @@
...
+
+In the end, we also provide a sample Python code that can be used to implement these steps.
 
 (checking)=
 
 ## Checking server status
 
-Requesting information about a user includes a `servers` field,
-which is a dictionary.
+First, request information about a particular user using a GET request as below.
```
minrk, line 19: I don't think we need 'following' or 'as below' when discussing things that immediately appear. A colon flows it into the sentence:

```suggestion
First, request information about a particular user with a GET request:
```

docs/source/reference/server-api.md
```diff
@@ -119,67 +101,50 @@ while the server is not ready yet:
...
+Note that `ready` is assigned `false` and `pending` is assigned `spawn`, meaning that the server is not ready and attempting to access it may not work as it is still in the process of spawning.
 
 [waiting for a server]: waiting
 
 (starting)=
 
 ## Starting servers
 
-To start a server, make the request
+To start a server, make the following API request:
```
minrk, line 112:
```suggestion
To start a server, make the API request:
```

docs/source/reference/server-api.md
```diff
@@ -119,67 +101,50 @@ while the server is not ready yet:
...
 
 (waiting)=
 
-## Waiting for a server
+## Waiting for a server to start
 
-If you are starting a server via the API,
-there's a good change you want to know when it's ready.
-There are two ways to do with:
+After receiving a `202 Accepted` response, you have to wait for it to start. Two approaches can be applied to establish when the server is ready:
```
minrk, line 133: FWIW, It's helpful for collaboration and review to wrap lines at sentence (or clause) boundaries,
rather than wrapping whole paragraphs onto one line.

docs/source/reference/server-api.md
```diff
@@ -119,67 +101,50 @@ while the server is not ready yet:
...
 
 1. {ref}`Polling the server model <polling>`
-2. the {ref}`progress API <progress>`
+2. {ref}`Using the Progress API <progress>`
```
minrk, line 136: Thanks for your explanation for this change! I don't think it qualifies as a proper noun, though: progress is a generic term. If Progress were a resource type, I might agree, but it's not - it's one endpoint that yields progress information.

```suggestion
2. {ref}`Using the progress API <progress>`
```

docs/source/reference/server-api.md
```diff
@@ -119,67 +101,50 @@ while the server is not ready yet:
...
 (polling)=
 
 ### Polling the server model
 
-The simplest way to check if a server is ready
-is to request the user model.
+The simplest way to check if a server is ready is to programmatically query the server model until the following two conditions are true:
 
-If:
+: 1. The server name is contained in the `servers` response, and
```
minrk, line 144: These colons are meant for definition lists. I suspect they should be removed. You can check the resulting rendered docs [here](https://jupyterhub--4106.org.readthedocs.build/en/4106/reference/server-api.html#polling-the-server-model).

minrk (commented):
Thanks for all the great suggestions! I've left some comments inline that I think will improve the final result.

alwasega:
> Thanks for all the great suggestions! I've left some comments inline that I think will improve the final result.

Thanks @minrk . I will update the file accordingly.

docs/source/reference/server-api.md
alwasega: Resolved

docs/source/reference/server-api.md
alwasega: I have removed `assigned` as suggested. I have also returned the link as your explanation makes sense. I removed it as I figured the reader will get to the section either way, so it did not add any value having the information.

docs/source/reference/server-api.md
alwasega: I have used "this" rather than "the"

docs/source/reference/server-api.md
alwasega: Do you mean to break sentences using the "Enter" button here? So that we do not have long paragraphs? I have not really understood what you mean by this comment but I have broken the two sentences here.

docs/source/reference/server-api.md
alwasega: Thanks. I agree that if it some generic term, then it does not count as a proper noun.

docs/source/reference/server-api.md
alwasega: Thank you.

docs/source/reference/server-api.md
```diff
@@ -202,27 +167,23 @@ def server_ready(hub_url, user, server_name="", token):
         return False
 ```
 
-You can keep making this check until `ready` is true.
+The function polls the server until `ready` is true.
```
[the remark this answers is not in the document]
alwasega: I have returned the initial text.

docs/source/reference/server-api.md
```diff
@@ -202,27 +167,23 @@ def server_ready(hub_url, user, server_name="", token):
...
 
 ```
 GET /hub/api/users/:user/servers/:servername/progress
 ```
 
 **Required scope: `read:servers`**
 
-This is an [EventStream][] API.
-In an event stream, messages are _streamed_ and delivered on lines of the form:
+The Progress API is an example of a [EventStream][] API. Therefore, as in a typical event stream, messages are _streamed_ and delivered in the form:
```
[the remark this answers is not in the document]
alwasega: Deleted the "as in a typical event stream" part.

docs/source/reference/server-api.md
```diff
@@ -300,23 +259,19 @@ Servers can be stopped with a DELETE request:
 DELETE /hub/api/users/:user/servers/[:servername]
 ```
 
-**Required scope: `servers`**
```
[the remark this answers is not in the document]
alwasega: I have returned it in the latest commit.

alwasega:
Hi @minrk. 
Please check commit #9f5f19cb264bcaac334d828367c501abb5e94f8e above. 
Thank you.

docs/source/reference/server-api.md
```diff
@@ -302,21 +262,20 @@ DELETE /hub/api/users/:user/servers/[:servername]
...
+: This code means your request was accepted, but is not yet completely processed.
 The server has `pending: 'stop'` at this point.
 
-Unlike start, there is no progress API for stop.
-To wait for stop to finish, you must poll the user model
-and wait for the server to disappear from the user `servers` model.
+There is no progress API for checking when a server actually stops.
+Thus, the only available alternative is to poll the server and wait for it to disappear from the user `servers` model.
+
+This Python code snippet can be used to check if a server stops:
```
minrk, line 278: That's not quite what the code does. It _stops_ the server, and then waits for it to finish stopping.

docs/source/reference/server-api.md
```diff
@@ -338,29 +296,27 @@ a token must be owned by the same user as the server,
...
 
-The URL returned from a server model is the url path suffix,
+The URL returned from a server model is the URL path suffix,
 e.g. `/user/:name/` to append to the jupyterhub base URL.
 
 For instance, `{hub_url}{server_url}`,
-where `hub_url` would be e.g. `http://127.0.0.1:8000` by default,
-and `server_url` `/user/myname`,
-for a full url of `http://127.0.0.1:8000/user/myname`.
+where `hub_url` would be such as `http://127.0.0.1:8000` by default,
```
minrk, line 303: 'For example' (e.g.) is more natural than 'such as' here

```suggestion
where `hub_url` would be e.g. `http://127.0.0.1:8000` by default,
```

docs/source/reference/server-api.md
```diff
@@ -338,29 +296,27 @@ a token must be owned by the same user as the server,
...
+and `server_url` is `/user/myname.` When combined, the two give a full URL of `http://127.0.0.1:8000/user/myname`.
```
minrk, line 304: Period shouldn't be inside literals:

```suggestion
and `server_url` is `/user/myname`.
When combined, the two give a full URL of `http://127.0.0.1:8000/user/myname`.
```

docs/source/reference/server-api.md
```diff
@@ -338,29 +296,27 @@ a token must be owned by the same user as the server,
...
 
 ## Python example
 
 The JupyterHub repo includes a complete example in {file}`examples/server-api`
-tying all this together.
+tying all the above mentioned steps together.
```
minrk, line 309:
```suggestion
tying all these steps together.
```

docs/source/reference/server-api.md
```diff
@@ -338,29 +296,27 @@ a token must be owned by the same user as the server,
...
 
-To summarize the steps:
+In summary, the steps involved while managing servers on behalf of users are:
```
minrk, line 311:
```suggestion
In summary, the steps involved in managing user servers via the JupyterHub API are:
```

minrk (commented):
Great! A few more suggestions for language refinements, then I think it's all set.

docs/source/reference/server-api.md
alwasega: I have also rephrased the sentence.

docs/source/reference/server-api.md
alwasega: I have rephrased the sentence. Please check whether it is accurate.

docs/source/reference/server-api.md
```diff
@@ -270,9 +230,9 @@ there will only be one event that looks like:
 }
 ```
 
-where `ready` and `url` are the same as in the server model (`ready` will always be true).
+In this case, `ready` and `url` are the same as in the server model, and `ready` will always be true.
```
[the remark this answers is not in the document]
alwasega: Noted and thank you.

docs/source/reference/server-api.md
```diff
@@ -254,11 +216,9 @@ ready
 : present and true only for the last event when the server is ready
 
 url
-: only present if `ready` is true; will be the server's url
+: only present if `ready` is true; will be the server's URL
 
-the progress API can be used even with fully ready servers.
-If the server is ready,
-there will only be one event that looks like:
+The progress API can be used even with fully ready servers. In such a situation, there will only be one event response of the form:
```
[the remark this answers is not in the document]
alwasega: Thank you. I will use less of the filler texts.

alwasega:
Hi @minrk. Is commit 3da4dc66c9848375df41513c97e459032b80b661 now okay?

minrk:
Wonderful, thank you!

alwasega:
Thank you too @minrk for your review and guidance. I have truly learnt a lot from this experience.

PR #4125, opened 2022-10-17
Using python3 doesn't work on most windows python installations even …

softkeldozy:
…when having dual installation but instead using just python works all the time

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,7 @@ conda install notebook
 
 #### Using `pip`
 
+Using python3 doesn't work on most windows python installations even when having dual installation but instead using just python works all the time:
 JupyterHub can be installed with `pip`, and the proxy with `npm`:
 
 ```bash


manics:
Hi! Please could you explain a bit more about this change? Windows doesn't come with Python, so whether `python` or `python3` works will depend on how it's installed.

softkeldozy:
Yes, windows don't come with python, but there are times when using python3 throws an error.
I recently worked on a Django project where I have about 5 Windows 10 Pro Pc (including mine), all having python installed, and that command didn't work. But when I used just python, it worked.
In my case, I had 2 versions installed on my Pc, and that command didn't work either. So I had to uninstall both, did a clean installation. Next, I ran python3 still nothing, so I just used python, and it worked perfectly. My colleagues had the same issue recently, and running python fixed it.

minrk:
How did you install Python? Every Windows Python install I have, `python3` works.

Windows is also not supported by JupyterHub, so I'm not sure a Windows-specific caveat that affects _some_ Windows Python environments to basic install docs is appropriate.

minrk:
Thanks for your contribution! I'm going to close this, since the statement it adds about "doesn't work on most Windows" is not correct, and unusual Windows caveats aren't appropriate for top-level docs in a project that doesn't support Windows.

PR #4126, opened 2022-10-17
Improve clarity in troubleshooting doc

PoorvajaRayas:
Improved the documentation to make it clearer

### DIFF
--- a/docs/source/troubleshooting.md
+++ b/docs/source/troubleshooting.md
@@ -1,6 +1,6 @@
 # Troubleshooting
 
-While troubleshooting, you may see unexpected behaviors or receive an error
+When troubleshooting, you may see unexpected behaviors or receive an error
 message. This section provides links for identifying the cause of the
 problem and how to resolve it.
 
@@ -8,14 +8,14 @@ problem and how to resolve it.
 
 ### JupyterHub proxy fails to start
 
-If you have tried to start the JupyterHub proxy and it fails to run:
+If you have tried to start the JupyterHub proxy and it fails to start:
 
 - check if the JupyterHub IP configuration setting is
   `c.JupyterHub.ip = '*'`; if it is, try `c.JupyterHub.ip = ''`
 - Try starting with `jupyterhub --ip=0.0.0.0`
 
 **Note**: If this occurs on Ubuntu/Debian, check that you are using a
-recent version of [Node](https://nodejs.org). Some versions of Ubuntu/Debian come with an older version
+recent version of [Node](https://nodejs.org). Some versions of Ubuntu/Debian come with a very old  version
 of Node and it is necessary to update Node.
 
 ### sudospawner fails to run
@@ -37,7 +37,7 @@ to the config file, `jupyterhub_config.py`.
 When nothing is given for these lists, there will be no admins, and all users
 who can authenticate on the system (i.e. all the Unix users on the server with
 a password) will be allowed to start a server. The allowed username set lets you limit
-this to a particular set of users, and admin_users let you specify who
+this to a particular set of users, and admin_users lets you specify who
 among them may use the admin interface (not necessary, unless you need to do
 things like inspect other users' servers or modify the user list at runtime).
 
@@ -87,7 +87,7 @@ sudo MY_ENV=abc123 \
 
 ### Error 500 after spawning my single-user server
 
-You receive an Error 500 while accessing the URL `/user/<your_name>/...`.
+You receive a 500 error while accessing the URL `/user/<your_name>/...`.
 This is often seen when your single-user server cannot verify your user cookie
 with the Hub.
 
@@ -100,7 +100,7 @@ There are two likely reasons for this:
 #### Symptoms
 
 The main symptom is a failure to load _any_ page served by the single-user
-server, met with an Error 500. This is typically the first page at `/user/<your_name>`
+server, met with a 500 error. This is typically the first page at `/user/<your_name>`
 after logging in or clicking "Start my server". When a single-user notebook server
 receives a request, the notebook server makes an API request to the Hub to
 check if the cookie corresponds to the right user. This request is logged.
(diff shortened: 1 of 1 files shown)


docs/source/troubleshooting.md
```diff
@@ -1,22 +1,22 @@
 # Troubleshooting
 
-When troubleshooting, you may see unexpected behaviors or receive an error
+While troubleshooting, you may see unexpected behaviors or receive an error
```
minrk, line 3: I would keep 'when'

docs/source/troubleshooting.md
```diff
@@ -1,22 +1,22 @@
...
 message. This section provides links for identifying the cause of the
 problem and how to resolve it.
 
 ## Behavior
 
 ### JupyterHub proxy fails to start
 
-If you have tried to start the JupyterHub proxy and it fails to start:
+If you have tried to start the JupyterHub proxy and it fails to run:
```
minrk, line 11: I think 'start' is better here

docs/source/troubleshooting.md
```diff
@@ -1,22 +1,22 @@
...
 
 - check if the JupyterHub IP configuration setting is
   `c.JupyterHub.ip = '*'`; if it is, try `c.JupyterHub.ip = ''`
 - Try starting with `jupyterhub --ip=0.0.0.0`
 
 **Note**: If this occurs on Ubuntu/Debian, check that you are using a
-recent version of [Node](https://nodejs.org). Some versions of Ubuntu/Debian come with a version
-of Node that is very old, and it is necessary to update Node.
+recent version of [Node](https://nodejs.org). Some versions of Ubuntu/Debian come with an older version
```
minrk, line 18: This dramatically weakens the language - we want to communicate that it is very old. At the very least, it should say that it's "old" not "older"

docs/source/troubleshooting.md
```diff
@@ -37,11 +37,11 @@ to the config file, `jupyterhub_config.py`.
 When nothing is given for these lists, there will be no admins, and all users
 who can authenticate on the system (i.e. all the Unix users on the server with
 a password) will be allowed to start a server. The allowed username set lets you limit
-this to a particular set of users, and admin_users lets you specify who
+this to a particular set of users, and admin_users let you specify who
```
minrk, line 40: `admin_users` is a singular configuration value, so it should be `lets` not `let.

docs/source/troubleshooting.md
```diff
@@ -37,11 +37,11 @@ to the config file, `jupyterhub_config.py`.
...
 among them may use the admin interface (not necessary, unless you need to do
 things like inspect other users' servers or modify the user list at runtime).
 
-### JupyterHub Docker container not accessible at localhost
+### What to do when JupyterHub Docker container is not accessible at localhost?
```
minrk, line 44: This doesn't fit most of the other headings, which are statements about what's going wrong.

docs/source/troubleshooting.md
```diff
@@ -87,7 +87,7 @@ sudo MY_ENV=abc123 \
 
 ### Error 500 after spawning my single-user server
 
-You receive a 500 error when accessing the URL `/user/<your_name>/...`.
```
minrk, line 90: I would call it a '500 error' not 'error 500'

docs/source/troubleshooting.md
```diff
@@ -71,7 +71,7 @@ After successfully logging in to JupyterHub with a compatible authenticator, I g
 This issue occurs when the authenticator requires a local system user to exist. In these cases, you need to use a spawner
 that does not require an existing system user account, such as `DockerSpawner` or `KubeSpawner`.
 
-### How can I run JupyterHub with sudo but use my current env vars and virtualenv location?
+### How can I run JupyterHub with sudo but use my current environment variables and virtualenv location?
```
minrk, line 74: 👍

docs/source/troubleshooting.md
```diff
@@ -167,7 +167,7 @@ When your whole JupyterHub sits behind an organization proxy (_not_ a reverse pr
 
 [JupyterHub services](https://jupyterhub.readthedocs.io/en/stable/reference/services.html) allow processes to interact with JupyterHub's REST API. Example use-cases include:
 
-- **Secure Testing**: provide a canonical Jupyter Notebook for testing production data to reduce the number of entry points into production systems.
+- **Secure Testing**: provides a canonical Jupyter Notebook for testing production data to reduce the number of entry points into production systems.
```
minrk, line 170: `provide` is correct here

docs/source/troubleshooting.md
```diff
@@ -324,7 +324,7 @@ Or use syslog:
 
 ### Toree integration with HDFS rack awareness script
 
-The Apache Toree kernel will have an issue when running with JupyterHub if the standard HDFS
+The Apache Toree kernel will have an issue while running with JupyterHub if the standard HDFS
```
minrk, line 327: I think 'when' is better here

docs/source/troubleshooting.md
```diff
@@ -371,7 +371,7 @@ You can also tail logs to view them in real-time using the `-f` option:
 ## Troubleshooting commands
 
 The following commands provide additional detail about installed packages,
-versions, and system information that may be helpful when troubleshooting
+versions, and system information that may be helpful while troubleshooting
```
minrk, line 374: Again, I think when is better here

minrk (commented):
Thanks for your contribution! I think several of these changes should be reverted, but some are helpful. I've left suggestions inline.

docs/source/troubleshooting.md
PoorvajaRayas: @minrk I think it was started off with answering questions which then moved on to stating the problems in the form of headings and providing solutions . So either I have to change those initial questions into headings to make everything look well structured or the other way around.

minrk: Good point. I think for the most part, the headings under 'behavior' should be _a behavior_ where possible, (i.e. "JupyterHub Docker container is not accessible at localhost"). If you see ways to rephrase one or more of the questions to match, that would be appropriate.

But not all situations quite fit that.

minrk:
Great job, thank you!

PoorvajaRayas:
@minrk Thanks a lot for merging! :)

PR #4130, opened 2022-10-18
refine text in template docs.

lumenCodes:
As part of my outreachy contribution, I updated the [templates.md](https://github.com/jupyterhub/jupyterhub/blob/main/docs/source/reference/templates.md) file by:
- adding a section reference for easy navigation
- rephrasing some sentences to be clearer and more understandable.


docs/source/reference/templates.md
```diff
@@ -1,18 +1,21 @@
...
 
 The pages of the JupyterHub application are generated from
-[Jinja](http://jinja.pocoo.org/) templates. These allow the header, for
-example, to be defined once and incorporated into all pages. By providing
-your own templates, you can have complete control over JupyterHub's
+[Jinja](http://jinja.pocoo.org/) templates. This allows the header to be defined once and incorporated into all pages. By providing your templates, you can have complete control over JupyterHub's
 appearance.
+This document will provide information on how to:
+
+- [add custom templates](#custom-templates)
```
minrk, line 8: Let's not add in-body table-of-contents links. We already have these links that will be auto-generated and correct in the sidebar of the docs, whereas these links will break without warning if the spelling of a heading is changed.

docs/source/reference/templates.md
```diff
@@ -22,16 +25,15 @@ Jinja provides a mechanism to [extend templates](http://jinja.pocoo.org/docs/2.1
...
+make extensive use of blocks, which allows you to customize parts of the interface easily.
 
 In general, a child template can extend a base template, `page.html`, by beginning with:
 
 ```html
 {% extends "page.html" %}
 ```
 
-This works, unless you are trying to extend the default template for the same
+This works if you are trying to extend the default template for the same
```
minrk, line 36: This change inverts the logic making it inaccurate. It works _unless_ you do this. It _doesn't_ work _if_ you are doing it.

docs/source/reference/templates.md
lumenCodes: Okay, I have noted the reviews and will implement them. 

Thank you

docs/source/reference/templates.md
lumenCodes: okay, thanks.

lumenCodes:
@minrk Please I have made the requested changes.

lumenCodes:
Done @minrk

PR #4135, opened 2022-10-19
update websecurity.md

Christiandike:
- fix typos/edit text
- add important links

PR #4136, opened 2022-10-19
Update wording in web security docs

yamakat:


docs/source/reference/websecurity.md
```diff
@@ -1,24 +1,21 @@
 # Security Overview
 
-The **Security Overview** section helps you learn about:
-
-- the design of JupyterHub with respect to web security
-- the semi-trusted user
-- the available mitigations to protect untrusted users from each other
-- the value of periodic security audits
+The **Security Overview** section helps you learn about: the semi-trusted user;
```
minrk, line 3: I think the bulleted list delivers this info better

docs/source/reference/websecurity.md
```diff
@@ -1,24 +1,21 @@
...
-works.
+works, as well as its design with respect to web security.
 
 ## Semi-trusted and untrusted users
 
 JupyterHub is designed to be a _simple multi-user server for modestly sized
-groups_ of **semi-trusted** users. While the design reflects serving semi-trusted
-users, JupyterHub is not necessarily unsuitable for serving **untrusted** users.
+groups_ of **semi-trusted** users. Its design reflects serving semi-trusted
+users, however, JupyterHub is not necessarily unsuitable for serving **untrusted** users.
```
minrk, line 14: I prefer the `While...` construction here. Can you discuss why you think the `however` works better?

minrk (commented):
Thanks for your contribution! I’ve left some suggestions inline for you to incorporate. You can update this PR with the suggestions and we will come back to review again.

Note that this will likely conflict with #4135 which touches the same file, and need resolution prior to merge.

docs/source/reference/websecurity.md
yamakat: I've tried to make the items on the list reference links to the corresponding headings or even turning it to a table of contents but couldn't get to do it, then thought about this alternative... I agree the bulleted list is simpler and makes it more rapidly comprehensible though.

docs/source/reference/websecurity.md
yamakat: The phrase sounds confusing to me, so I added the `however` so it could bring an idea of opposition between semi-trusted and untrusted user's cases.

Reading it again and again, maybe my discomfort is on `is not necessarily unsuitable`. It could be, then, `While the design reflects serving semi-trusted users, JupyterHub can also be suitable for serving **untrusted** users.`?

minrk:
#4135 has been merged, so there are now conflicts to resolve

yamakat:
> #4135 has been merged, so there are now conflicts to resolve

Got it. Thanks, Min!

docs/source/reference/websecurity.md
```diff
@@ -2,23 +2,22 @@
...
 The **Security Overview** section helps you learn about:
 
-- the design of JupyterHub with respect to web security
 - the semi-trusted user
 - the available mitigations to protect untrusted users from each other
 - the value of periodic security audits
 
 This overview also helps you obtain a deeper understanding of how JupyterHub
-works.
+works, as well as its design with respect to web security.
```
minrk, line 10: Can I ask why this item is moved out of the bullets? It makes sense to me for this to be the first item, rather than the last.

docs/source/reference/websecurity.md
```diff
@@ -2,23 +2,22 @@
...
 JupyterHub is designed to be a _simple multi-user server for modestly sized
-groups_ of **semi-trusted** users. While the design reflects serving semi-trusted
-users, JupyterHub is not necessarily unsuitable for serving **untrusted** users.
+groups_ of **semi-trusted** users. While the design reflects serving
+semi-trusted users, JupyterHub can also be suitable for serving **untrusted** users.
 
-Using JupyterHub with **untrusted** users does mean more work for the
-administrator. Much care is required to secure a Hub, with extra caution on
-protecting users from each other, since the Hub serves untrusted users.
+That way, using JupyterHub with **untrusted** users means more work by the
```
minrk, line 18: Perhaps a better connection:

```suggestion
As a result, using JupyterHub with **untrusted** users means more work by the
```

docs/source/reference/websecurity.md
yamakat: It was due to me trying to use reference links in the topics and headings before. Web security's design encompasses all the page, so I've put it this way, but since I couldn't use reference links to the headings, it makes sense to be put as a first item, yes.

minrk:
Great job, thanks for your contribution!

PR #4137, opened 2022-10-19
update websecurity.md #4135

KaluBuikem:
Implemented the changes @sgibson91 suggested in #4135


minrk:
Thanks for the PR! But during review, these changes are already made in #4135

PR #4139, opened 2022-10-20
highlight note about the docker image scope

Joel-Ando:
Restructure words by making them bold so that users will actually understand their meaning

### DIFF
--- a/docs/source/quickstart-docker.rst
+++ b/docs/source/quickstart-docker.rst
@@ -1,10 +1,16 @@
 Install JupyterHub with Docker
 ==============================
 
+
 The JupyterHub `docker image <https://hub.docker.com/r/jupyterhub/jupyterhub/>`_ is the fastest way to set up Jupyterhub in your local development environment.
 
-The ``JupyterHub`` docker image runs the Hub service only. It does not provide other Jupyter components such as Notebook installation, which are needed by the single-user servers.
-To run the single-user servers, which may be on the same system as the Hub or not, Jupyter Notebook or Jupyter Lab must be installed.
+.. note::
+    This ``jupyterhub/jupyterhub`` docker image is only an image for running
+    the Hub service itself. It does not provide the other Jupyter components,
+    such as Notebook installation, which are needed by the single-user servers.
+    To run the single-user servers, which may be on the same system as the Hub or
+    not, `JupyterLab <https://jupyterlab.readthedocs.io/>`_ or Jupyter Notebook must be installed.
+
 
 .. important::
     We strongly recommend that you follow the `Zero to JupyterHub`_ tutorial to
@@ -23,6 +29,7 @@ To pull the latest JupyterHub image and start the `jupyterhub` container, run th
 ::
    
     docker run -d -p 8000:8000 --name jupyterhub jupyterhub/jupyterhub jupyterhub
+
 
 This command exposes the Jupyter container on port:8000. Navigate to `http://localhost:8000` in a web browser to access the JupyterHub console.
 


docs/source/quickstart-docker.rst
```diff
@@ -1,11 +1,17 @@
 Install JupyterHub with Docker
 ==============================
 
+
+**important:**
```
minrk, line 5: This is now duplicated.

This same message is already highlighed via the `.. important::` below:

Screen Shot 2022-10-21 at 10 20 01

docs/source/quickstart-docker.rst
```diff
@@ -16,6 +22,14 @@ Prerequisites
 You should have `Docker`_ installed on a Linux/Unix based system.
 
 
+
+**note:**
```
minrk, line 26: In rst, notes are created with `.. note::`, not `**note:**`.

This duplicates the note, rather than updating it.

Joel-Ando: Ok Sir @minrk thanks

docs/source/quickstart-docker.rst
```diff
@@ -1,11 +1,17 @@
...
 ==============================
 
+
+**important:**
+   We highly recommend following the `Zero to JupyterHub`_ tutorial for
+   installing JupyterHub.
+
 The JupyterHub `docker image <https://hub.docker.com/r/jupyterhub/jupyterhub/>`_ is the fastest way to set up Jupyterhub in your local development environment.
 
 The ``JupyterHub`` docker image runs the Hub service only. It does not provide other Jupyter components such as Notebook installation, which are needed by the single-user servers.
```
Joel-Ando, line 11: sir @minrk do you think the notebook installation should be a link if there is a doc on it?

minrk: Sure, you can link to https://jupyterlab.readthedocs.io

docs/source/quickstart-docker.rst
```diff
@@ -16,6 +18,14 @@ Prerequisites
 You should have `Docker`_ installed on a Linux/Unix based system.
 
 
+
+.. note::
+    This ``jupyterhub/jupyterhub`` docker image is only an image for running
```
minrk, line 23: This note is still duplicated from the above. Do you mean to move the note, or replace it? Because it shouldn't be in both places.

Joel-Ando: > This note is still duplicated from the above. Do you mean to move the note, or replace it? Because it shouldn't be in both places.

I mean to remove it sir

minrk: Is there a reason you moved the note about what the image is for to the 'prerequisites' section?

minrk: Yes, it is added to the subsection "Prerequisites". I'm not sure I understand the reason for moving it there, since it isn't related to the prerequisites.

Joel-Ando: > Yes, it is added to the subsection "Prerequisites". I'm not sure I understand the reason for moving it there, since it isn't related to the prerequisites.

you are right sir. it does make sense being under the prerequisites

docs/source/quickstart-docker.rst
```diff
@@ -1,10 +1,16 @@
...
 The JupyterHub `docker image <https://hub.docker.com/r/jupyterhub/jupyterhub/>`_ is the fastest way to set up Jupyterhub in your local development environment.
 
-The ``JupyterHub`` docker image runs the Hub service only. It does not provide other Jupyter components such as Notebook installation, which are needed by the single-user servers.
-To run the single-user servers, which may be on the same system as the Hub or not, Jupyter Notebook or Jupyter Lab must be installed.
+.. note::
+    This ``jupyterhub/jupyterhub`` docker image is only an image for running
+    the Hub service itself. It does not provide the other Jupyter components,
+    such as Notebook installation, which are needed by the single-user servers.
+    To run the single-user servers, which may be on the same system as the Hub or
+    not, `Jupyter Notebook <https://jupyterlab.readthedocs.io/>`_ or greater must be installed.
```
minrk, line 12: `or greater` here doesn't work - or greater would follow a version number. Plus, the text is "Jupyter Notebook" but the link is for JupyterLab. Let's say: "JupyterLab (link) or Jupyter Notebook (no link) must be installed"

Joel-Ando: > `or greater` here doesn't work - or greater would follow a version number. Plus, the text is "Jupyter Notebook" but the link is for JupyterLab. Let's say: "JupyterLab (link) or Jupyter Notebook (no link) must be installed"

ok sir @minrk thanks

minrk:
Thanks!

PR #4141, opened 2022-10-20
Remove trailing commas in code blocks

Joel-Ando:
Edited code block

### DIFF
--- a/docs/source/rbac/roles.md
+++ b/docs/source/rbac/roles.md
@@ -61,7 +61,7 @@ c.JupyterHub.load_roles = [
    'scopes': ['servers'],
    'users': ['alice', 'bob'],
    'services': ['idle-culler'],
-   'groups': ['admin-group'],
+   'groups': ['admin-group']
  }
 ]
 ```
@@ -128,7 +128,7 @@ use the scope `inherit` (for 'inheriting' the owner's permissions):
 c.JupyterHub.load_roles = [
  {
    'name': 'server',
-   'scopes': ['inherit'],
+   'scopes': ['inherit']
  }
 ]
 ```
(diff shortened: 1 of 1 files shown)


Christiandike (commented):
The use of trailing commas is a convention in Python.
You can learn more about it [here](https://blog.miguelgrinberg.com/post/video-using-trailing-commas-in-python)

Joel-Ando:
> The use of trailing commas is a convention in Python. You can learn more about it [here](https://blog.miguelgrinberg.com/post/video-using-trailing-commas-in-python)

ok wow thanks bro @Christiandike

minrk:
Yup, trailing commas are preferred just about everywhere they are allowed (these days, ~only JSON doesn't allow them, which has been a source of problems for years) because the improve version control behavior and readability.

Since that is the only change here, feel free to close this PR and try a new one.

A good tip contributing to projects in general: when there's a a systematic change like this and it changes every instance of something, it's good to stop and ask yourself why that might be before opening a PR. If something is internally consistent but doesn't follow the pattern you are applying, it's more likely to be a _choice_  than an error made over and over again, and the changes won't be accepted. When in doubt, it's a good idea to ask first.

Joel-Ando:
> Yup, trailing commas are preferred just about everywhere they are allowed (these days, ~only JSON doesn't allow them, which has been a source of problems for years) because the improve version control behavior and readability.
> 
> Since that is the only change here, feel free to close this PR and try a new one.
> 
> A good tip contributing to projects in general: when there's a a systematic change like this and it changes every instance of something, it's good to stop and ask yourself why that might be before opening a PR. If something is internally consistent but doesn't follow the pattern you are applying, it's more likely to be a _choice_ than an error made over and over again, and the changes won't be accepted. When in doubt, it's a good idea to ask first.

thanks very much sir @minrk

PR #4142, opened 2022-10-20
Update authenticators.md

Joel-Ando:
edited authenticators.md code block


minrk:
This is the same change as in #4141, so we can close for the same reason.

PR #4147, opened 2022-10-20
oauth doc: added punctuations and capitalized words where necessary

Teebarh:
I capitalized the 'oa' in 'oauth' to words missing it and added punctuation marks in some sentences.

### DIFF
--- a/docs/source/reference/oauth.md
+++ b/docs/source/reference/oauth.md
@@ -30,19 +30,19 @@ Some relevant points:
 Here are some key definitions to keep in mind when we are talking about OAuth.
 You can also read more detail [here](https://www.oauth.com/oauth2-servers/definitions/).
 
-- **provider** the entity responsible for managing identity and authorization,
+- **provider**: The entity responsible for managing identity and authorization,
   always a web server.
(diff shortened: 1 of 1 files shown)


docs/source/reference/oauth.md
```diff
@@ -56,8 +56,8 @@ A single oauth flow generally goes like this:
 
 1. A **browser** makes an HTTP request to an oauth **client**.
 2. There are no credentials, so the client _redirects_ the browser to an "authorize" page on the oauth **provider** with some extra information:
-   - the oauth **client id** of the client itself
-   - the **redirect uri** to be redirected back to after completion
+   - the oauth **client id** of the client itself.
```
minrk, line 59: fwiw, when bullets/lists are sentence _fragments_, adding a period on the end is not required or necessarily correct. But since _one_ item here is multiple sentences, I think this is okay here.

minrk (commented):
Great, thank you!

docs/source/reference/oauth.md
Teebarh: I'll take note of that, thank you so much!

PR #4150, opened 2022-10-21
grammar improvements in quickstart.md

liliyao2022:
Modified some grammar issues.

### DIFF
--- a/docs/source/quickstart.md
+++ b/docs/source/quickstart.md
@@ -4,7 +4,7 @@
 
 Before installing JupyterHub, you will need:
 
-- a Linux/Unix based system
+- a Linux/Unix-based system
 - [Python](https://www.python.org/downloads/) 3.6 or greater. An understanding
   of using [`pip`](https://pip.pypa.io) or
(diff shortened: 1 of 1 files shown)


docs/source/quickstart.md
```diff
@@ -17,7 +17,7 @@ Before installing JupyterHub, you will need:
 
   - If you are using **`pip`**, install a recent version of
     [nodejs/npm](https://docs.npmjs.com/getting-started/installing-node).
-    For example, install it on Linux (Debian/Ubuntu) using:
+    For example, install it on Linux (Debian/Ubuntu) using the following:
```
minrk, line 20: I don't think it's beneficial to add `the following` when a code block fits into the flow of a sentence, e.g. "install it...using: [code block]". Adding `following` is helpful when it's further removed from the code block, as in

> Running the following code will do such and such:

I know we've been incosistent with this, but I'd like to mostly move _away_ from habitual "with the following:"

minrk:
```suggestion
    For example, install it on Linux (Debian/Ubuntu) using:
```

PR #4153, opened 2022-10-21
Consistent capitalization of Authenticator

Teniola-theDev:
There were some inconsistencies like Authenticator and authenticator so I went for Authenticator instead.   
I also brought out the link for the word "roles" just to ensure that it was easier to click because links can not be clicked in a codeblock.  
I changed the word from pam (which seemed more like a word) to PAM (which seemed more like an abbreviation), I also added the full meaning.


minrk (commented):
Thanks for the contribution! I made a suggestion to fix the roles link instead of removing it.

docs/source/getting-started/authenticators-users-basics.md
```diff
@@ -72,11 +74,11 @@ After starting the Hub once, it is not sufficient to **remove** a user
 from the allowed users set in your config file. You must also remove the user
 from the Hub's database, either by deleting the user from JupyterHub's
 admin page, or you can clear the `jupyterhub.sqlite` database and start
-fresh.
+afresh.
```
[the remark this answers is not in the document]
minrk:
```suggestion
fresh.
```

docs/source/getting-started/authenticators-users-basics.md
```diff
@@ -25,10 +25,12 @@ If this configuration value is not set, then **all authenticated users will be a
 ```{note}
 As of JupyterHub 2.0, the full permissions of `admin_users`
 should not be required.
-Instead, you can assign [roles](https://jupyterhub.readthedocs.io/en/stable/rbac/roles.html#define-role-target) to users or groups
+Instead, you can assign [roles](define-role-target) to users or groups
 with only the scopes they require.
 ```
 
+**[roles](https://jupyterhub.readthedocs.io/en/stable/rbac/roles.html#define-role-target)**
```
minrk, line 32: This duplicate link should be removed.

docs/source/getting-started/authenticators-users-basics.md
```diff
@@ -25,10 +25,12 @@ If this configuration value is not set, then **all authenticated users will be a
 ```{note}
 As of JupyterHub 2.0, the full permissions of `admin_users`
 should not be required.
-Instead, you can assign [roles](https://jupyterhub.readthedocs.io/en/stable/rbac/roles.html#define-role-target) to users or groups
+Instead, you can assign [roles](define-role-target) to users or groups
 with only the scopes they require.
 ```
 
+roles
```
minrk, line 32: It looks like there's still some leftover text here.

minrk:
Great, thank you!

PR #4156, opened 2022-10-22
Update dockerfile README

liliyao2022:
Just for reviews.

### DIFF
--- a/dockerfiles/README.md
+++ b/dockerfiles/README.md
@@ -1,12 +1,12 @@
 ## What is Dockerfile.alpine
 
-Dockerfile.alpine contains base image for jupyterhub. It does not work independently, but only as part of a full jupyterhub cluster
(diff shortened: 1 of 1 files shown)


dockerfiles/README.md
```diff
@@ -1,12 +1,12 @@
...
 
-Dockerfile.alpine contains base image for jupyterhub. It does not work independently, but only as part of a full jupyterhub cluster
+Dockerfile.alpine contains the base image for jupyterhub. It does not work independently, but only as part of a full jupyterhub cluster
 
 ## How to use it?
 
 1. A running configurable-http-proxy, whose API is accessible.
 2. A jupyterhub_config file.
-3. Authentication and other libraries required by the specific jupyterhub_config file.
+3. Authentication and other libraries are required by the specific jupyterhub_config file.
```
minrk, line 9: I think the clarification is that the list should be preceded with "You will need:"

Then this `are` should be left out, and it will match the others.

PR #4175, opened 2022-10-26
Update test.rst

emmanuella194:
I made the Test Organisation bold for emphasis

### DIFF
--- a/docs/source/contributing/setup.rst
+++ b/docs/source/contributing/setup.rst
@@ -1,10 +1,10 @@
 .. _contributing/setup:
 
 ================================
- #Setting up a development install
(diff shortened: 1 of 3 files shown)


minrk (commented):
Thanks for your contribution!

This text is already rendered as a heading (as seen [here](https://jupyterhub.readthedocs.io/en/stable/contributing/tests.html#test-organisation)), so we shouldn't add bold to it. Heading styling is handled by the theme, and should be consistent across headings. Since that's the only change here, I suggest we close this PR.

emmanuella194:
Thanks for the clarification minrk

PR #4176, opened 2022-10-27
Update CreateGroup.test.js

liliyao2022:
Redux createStore() is deprecated, so I replaced it with configureStore().


minrk:
This is failing with:

>  TypeError: (0 , _redux.configureStore) is not a function

So I think something's not right with the patch.

minrk:
Closing this one, since it doesn't appear to be a correct fix. Feel free to try again another time!

PR #4177, opened 2022-10-27
Update EditUser.test.js

liliyao2022:
Redux createStore() is deprecated, so I replaced it with configureStore().


minrk:
Thanks for the PR! Closing this in favor of #4176 which should have all of the same change in one PR.

PR #4172, opened 2022-10-26
Deleted unused failRegexEvent

liliyao2022:
Deleted a  variable because it is declared but its value is never read.

PR #4181, opened 2022-10-27
Update EditUser.test.js

liliyao2022:
Deleted unused failRegexEvent() function and replaced createStore() with configureStore().


minrk:
This is a duplicate of #4177 and #4172, so I'm closing here. Please complete the changes in those PRs instead of opening a new PR for every file.

PR #4182, opened 2022-10-27
Update AddUser.jsx

liliyao2022:
Deleted an unused function.


minrk:
@liliyao2022 for future reference, it's very helpful to have a title for pull requests that describes what they do. If you could update the titles of your PRs to be brief descriptions, that would help very much. 'update adduser.jsx' isn't a description, the 'deleted unused failRegexEvent' would be more appropriate for a title.

minrk:
Thanks for the pull request! I'm closing this in favor of #4182 which should have all of these changes in one PR.

PR #4193, opened 2022-11-02
Accesibility: Added eslint-jsx-a11y plugin

Biki-das:
I tried to add the eslint-jsx-a11y plugin that will help us to catch static code errors in regard to accessibility


Biki-das:
test failures seems to be out of scope from my code as i have not changed any of the core python files.

PR #4195, opened 2022-11-02
Accessibility:-Added eslint-plugin-jsx-a11y package as linter for accessibility

Biki-das:
Adding a linter for accessibility can really help us to catch static code errors in regards to accessibility , This PR adds the Eslint-plugin-jsx-a11y package and also set the config for the same.

### DIFF
--- a/jsx/.eslintrc.json
+++ b/jsx/.eslintrc.json
@@ -1,5 +1,5 @@
 {
-  "extends": ["plugin:react/recommended"],
+  "extends": ["plugin:react/recommended", "plugin:jsx-a11y/recommended"],
   "parserOptions": {
     "ecmaVersion": 2018,
     "sourceType": "module",
@@ -12,7 +12,7 @@
       "version": "detect"
     }
   },
-  "plugins": ["eslint-plugin-react", "prettier", "unused-imports"],
+  "plugins": ["eslint-plugin-react", "prettier", "unused-imports", "jsx-a11y"],
   "env": {
     "es6": true,
     "browser": true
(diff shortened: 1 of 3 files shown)


liliyao2022:
Hey @Biki-das, after doing some research, I found out that you used the wrong package name,  it should be "eslint-plugin-jsx-a11y", instead of "eslint-plugin-jsx-ally".

liliyao2022:
> Hey @Biki-das, after doing some research, I found out that you used the wrong package name, it should be "eslint-plugin-jsx-a11y", instead of "eslint-plugin-jsx-ally" in your title.

Biki-das:
> Hey @Biki-das, after doing some research, I found out that you used the wrong package name, it should be "eslint-plugin-jsx-a11y", instead of "eslint-plugin-jsx-ally".

yeah i made the mistake in the PR name, rest the package is a valid one and it works as i have already set the config for the same in the .eslint.rc.json file

minrk:
Accessibility testing is being discussed in #4078. We should resolve that discussion about how to approach accessibility testing, what tools to use, and how to integrate it with CI before proceeding with implementation.

Biki-das:
> Accessibility testing is being discussed in #4078. We should resolve that discussion about how to approach accessibility testing, what tools to use, and how to integrate it with CI before proceeding with implementation.

thanks for the review!, yes as you said it will require a proper discussion on what approach we would be taking to deal with accessibility testing

minrk:
Thanks for the PR! Closing due to conflicts, pending resolution of tool selection in #4078.

PR #4196, opened 2022-11-02
Added gitpod support

Biki-das:
Tried to add gitpod support for this repo, so developers can easily spin up the repo and run it locally


minrk:
Thanks for the PR! In general, before adding a new tool to the development of a project, it is best to start with an Issue discussing whether it's appropriate and fits with the project, rather than contributing a new toolchain that the project must support maintain.

We may want to do this, but there's a discussion to be had first about what problems it solves, and what we need to do to support, maintain, and test it.

minrk:
Closing in favor of discussion issue: #4565

PR #4199, opened 2022-11-07
Dockerfile: add python-is-python3 to hub image

thetorpedodog:
Adds the `python-is-python3` Debian package so that the `python` binary on the system PATH refers to the Python 3 installation.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -68,6 +68,7 @@ RUN apt-get update \
     curl \
     gnupg \
     locales \
+    python-is-python3 \
     python3-pip \
     python3-pycurl \
     nodejs \


manics:
Hi! Please could you explain a bit more about why this is useful, especially since this is a self-contained image? User notebooks are expected to be in a separate container.

thetorpedodog:
It’s a convenience for people (like me) who end up doing work in the hub image and may be surprised (or have scripts unexpectedly not work) due to there not being a binary named `python` (e.g. `#!/usr/bin/env python` fails). `python-is-python3` just creates a symlink from `/usr/bin/python` to `python3` so that regular old `python` does what you would expect.

It’s not huge and I have no problem if you’d rather not, but in my experience it makes life a tiny bit better.

consideRatio (approved):
I'm +1 for this. When using the `python:3.11` image for example as we do here and there in jupyterhub, `python` will work. So, having it work here also is consistent for us maintainers.

minrk:
This change was applied in #4394

PR #2705, opened 2019-08-27
Reset _jupyterhub_user after login cookie cleared

aar0nTw:
This change will fix #1765
Screen Shot 2019-08-27 at 10 12 05 PM

PR #4202, opened 2022-11-09
Added prehandle_logout

wesuuu:
This pull request is to allow subclasses the ability to perform some action before starting the logout process. Use cases include: calling `self.current_user` right before logout. 

For example, app may need to perform some action with the user object (like removing information from a cache) before its unset via `self.default_logout()`. `self.current_user` is not exposed to the override-able method `self.handle_logout()` if `self.shutdown_on_logout` is specified.

### DIFF
--- a/jupyterhub/handlers/login.py
+++ b/jupyterhub/handlers/login.py
@@ -45,6 +45,14 @@ class LogoutHandler(BaseHandler):
         self.clear_login_cookie()
         self.statsd.incr('logout')
 
+    async def prehandle_logout(self):
+        """Custom user action before starting logout
+
+        By default a no-op, this function should be overriden in subclasses
+        to have JupyterHub take a custom action before starting logout
+        """
+        return
+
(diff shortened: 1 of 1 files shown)


minrk:
Thanks for the PR! Here's a question: if we fixed `self.current_user` so it's available in `handle_logout()`, would that solve your problem? If so, I might go for that rather than defining a new hook.

wesuuu:
@minrk Thanks for the feedback and I think that would solve the problem. How that would be implemented since `default_handle_logout()` sets `self.current_user` to `None`?

minrk:
It doesn't _really_ need to do that. This was introduced in #2705 because the 'successful logout page' still showed the login button instead because the template makes that decision based on whether current_user is defined. Instead of doing this _always_, we can do it specifically for the logout handler prior to rendering the page. #4230 does this.

wesuuu:
Awesome. Thanks for the explanation. I'll go ahead and close this pull request

PR #4207, opened 2022-11-11
more selenium test cases

mouse1203:
- revoking tokens
- logout
- token page


jupyterhub/tests/selenium/test_browser.py
```diff
@@ -110,11 +127,9 @@ async def login(browser, user, pass_w):
     await webdriver_wait(browser, EC.url_changes(browser.current_url))
 
 
-async def test_submit_login_form(app, browser):
-    user = "test_user"
-    pass_w = "test_user"
+async def test_submit_login_form(app, browser, user="test_user", pass_w="test_user"):
```
minrk, line 130: It doesn't make sense to have default values for arguments in test functions. Arguments to test functions are always fixture names.

jupyterhub/tests/selenium/test_browser.py
```diff
@@ -46,10 +57,10 @@ def in_thread(f, *args, **kwargs):
     return asyncio.get_event_loop().run_in_executor(None, partial(f, *args, **kwargs))
 
 
-async def open_url(app, browser, url="login"):
-    """initiating open the login page in the browser"""
+async def open_url(app, browser, param):
```
minrk, line 60: I think `url` or `path` make more sense here than `param`.

jupyterhub/tests/selenium/test_browser.py
```diff
@@ -193,11 +209,11 @@ async def test_open_url_login(
         assert next_url.endswith("spawn?param=value")
         assert f"user/{user}/" not in next_url
     else:
-        if next_url.endswith(f"/user/{user}/") == False:
-            await webdriver_wait(
-                browser, EC.url_to_be(ujoin(public_url(app), f"/user/{user}/"))
-            )
-        assert next_url.endswith(f"/user/{user}/")
+        u = app.users[orm.User.find(app.db, user)]
```
minrk, line 212: `user` fixture is already a User object, so we can avoid these lookups

jupyterhub/tests/selenium/test_browser.py
```diff
@@ -209,36 +225,686 @@ async def test_open_url_login(
         ("user", "password"),
     ],
 )
-async def test_invalid_credantials(app, browser, user, pass_w):
-    await open_url(app, browser)
+async def test_login_with_invalid_credantials(app, browser, user, pass_w):
+    await open_url(app, browser, param="login")
     await login(browser, user, pass_w)
-    await asyncio.sleep(0.1)
+    await asyncio.sleep(0.5)
```
minrk, line 231: Let's try to add a comment explaining what we are waiting for with each sleep, so we can maybe replace them with specific conditions in the future.

jupyterhub/tests/selenium/test_browser.py
```diff
@@ -209,36 +225,686 @@ async def test_open_url_login(
...
     """adding for a catching of the reflected error"""
```
minrk, line 232: docstring has moved away from the signature

jupyterhub/tests/selenium/test_browser.py
```diff
@@ -209,36 +225,686 @@ async def test_open_url_login(
...
+        ("/hub/token", "test_user", "test_user"),
+        # the spawn-pending page: verify if links work on the top bar
+        ("hub/spawn", "test_user", "test_user"),
+        # "hub/not" = any url that is not existed: verify if links work on the top bar
+        ("hub/not", "test_user", "test_user"),
+        # the login page: verify if links work on the top bar
+        ("", None, None),
+    ],
+)
+async def test_menu_bar(app, browser, slow_spawn, value, user, pass_w):
```
minrk, line 262: Probably don't need parametrize on `user, pass_w` if it's always the same. Maybe a True/False for logged-in?

Let's also use a more informative name for `value`, maybe `page`?

jupyterhub/tests/selenium/test_browser.py
```diff
@@ -209,36 +225,686 @@ async def test_open_url_login(
...
+        if index == 0:
+            assert links_bar_url.endswith("/hub/")
+            if not value:
+                assert (
+                    f"hub/login?next={url_escape(app.base_url)}" in browser.current_url
+                )
+            else:
+                u = app.users[orm.User.find(app.db, user)]
+                while not u.spawner.ready:
+                    await asyncio.sleep(1.5)
```
minrk, line 305: Let's shorten the interval:

```suggestion
                    await asyncio.sleep(0.1)
```

jupyterhub/tests/selenium/test_browser.py
```diff
@@ -209,36 +225,686 @@ async def test_open_url_login(
...
+                    assert str(logs_list[0]) == "Server requested"
+                    assert str(logs_list[1]) == "Spawning server..."
+                    assert (
+                        str(logs_list[2])
+                        == f"Server ready at {app.base_url}user/{user}/"
+                    )
+                    assert str(logs_list[2]) == progress_messages
+        except NoSuchElementException:
+            break
+        except StaleElementReferenceException:
```
minrk, line 439: This can be one except with multiple exceptions to catch:

```suggestion
        except (NoSuchElementException, StaleElementReferenceException):
```

jupyterhub/tests/selenium/test_browser.py
```diff
@@ -209,36 +225,686 @@ async def test_open_url_login(
...
+            break
+        await asyncio.sleep(0.001)
+
+
+async def test_spawn_pending_launch_server(
```
minrk, line 444: Let's maybe name these spawn pending tests based on the server state (stopped, start_pending, ready) rather than what should happen.

jupyterhub/tests/selenium/test_browser.py
```diff
@@ -209,36 +225,686 @@ async def test_open_url_login(
...
+
+
+async def open_home_page(app, browser, user, pass_w):
+    """function to open the home page"""
+
+    home_page = url_escape(app.base_url) + "hub/home"
+    await open_url(app, browser, param="/login?next=" + home_page)
+    await login(browser, user, pass_w)
+    while '/hub/home' not in browser.current_url:
+        await asyncio.sleep(0.5)
```
minrk, line 473: This can be shorter:

```suggestion
        await asyncio.sleep(0.1)
```

jupyterhub/tests/selenium/test_browser.py
```diff
@@ -209,36 +225,686 @@ async def test_open_url_login(
...
+    await in_thread(browser.get, home_page)
+    assert u.spawner.ready
+    assert is_displayed(browser, HomePageLocators.BUTTON_START_SERVER)
+    assert is_displayed(browser, HomePageLocators.BUTTON_STOP_SERVER)
+
+
+# HOME PAGE
+
+
+async def open_home_page(app, browser, user, pass_w):
```
minrk, line 466: open_home_page can probably take a _default_ user/pass_w so you don't need to specify it every time

jupyterhub/tests/selenium/test_browser.py
```diff
@@ -209,36 +225,686 @@ async def test_open_url_login(
...
+    await in_thread(browser.get, next_url)
+
+    buttons = browser.find_elements(*HomePageLocators.BUTTONS_SERVER)
+    if len(buttons) != 2:
+        await webdriver_wait(browser, EC.visibility_of_all_elements_located(buttons))
+
+    u = app.users[orm.User.find(app.db, user)]
+    while not u.spawner.ready:
+        await asyncio.sleep(0.1)
+    """add this stop click event is registred in JS to verify that the poccess is not still pending"""
```
minrk, line 544: docstring in the middle

jupyterhub/tests/selenium/test_browser.py
```diff
@@ -209,36 +225,686 @@ async def test_open_url_login(
...
-    await in_thread(browser.get, url)
-    redirected_url = url_path_join(public_host(app), app.base_url, '/hub/token')
+    token_page = url_escape(app.base_url) + "hub/token"
+    await open_url(app, browser, param="/login?next=" + token_page)
+    await login(browser, user, pass_w)
+    while '/hub/token' not in browser.current_url:
+        await asyncio.sleep(0.5)
+
+
+def elements_table(browser):
```
minrk, line 577: These might want more descriptive names, since they are token-page specific

jupyterhub/tests/selenium/test_browser.py
```diff
@@ -209,36 +225,686 @@ async def test_open_url_login(
...
+    return body.find_elements(*TokenPageLocators.TABLE_API_ROWS_BY_CLASS)
+
+
+def elements_row_head(browser):
+    table = browser.find_element(*TokenPageLocators.TABLE_API)
+    head = table.find_element(*TokenPageLocators.TABLE_API_HEAD)
+    return head.find_elements(*TokenPageLocators.TABLE_API_ROWS)
+
+
+async def test_elements_of_token_page_server_not_started(
```
minrk, line 598: token page isn't really sensitive to server running, instead let's create a couple of tokens with e.g. `user.new_api_token` and check the page contents. We don't need two separate tests.

jupyterhub/tests/selenium/test_browser.py
```diff
@@ -209,36 +225,686 @@ async def test_open_url_login(
...
+    elif token_opt == "1 Week":
+        assert Dict_body_columns.get(3) == "in 7 days"
+    # verify that the button for revoke is presented
+    buttons = elements_table_body(browser).find_elements(
+        *TokenPageLocators.BUTTON_REVOKE
+    )
+    for button in buttons:
+        button = button.text
+    assert button == columns_body[len(columns_body) - 1].text
+    assert int(len(buttons)) == 1
```
minrk, line 823: Let's test at the end with a db query that the tokens were actually created with the right expiration.

jupyterhub/tests/selenium/test_browser.py
```diff
@@ -209,36 +225,686 @@ async def test_open_url_login(
...
+        assert int(len(buttons)) == 2
+        while int(len(buttons)) != 0:
+            await asyncio.sleep(0.5)
+            click(browser, TokenPageLocators.BUTTON_REVOKE)
+            await asyncio.sleep(0.5)
+            buttons = elements_table_body(browser).find_elements(
+                *TokenPageLocators.BUTTON_REVOKE
+            )
+            assert int(len(buttons)) == int(len(elements_row_body(browser)))
+        assert int(len(buttons)) == 0
```
minrk, line 878: Same here - verify that the token we revoked is actually gone from the db

jupyterhub/tests/test_browser.py
```diff
@@ -0,0 +1,41 @@
+import asyncio
```
minrk, line 1: I accidentally added this file. Let's remove it.

mouse1203: done

minrk:
Great! Tests are passing. Next steps are:

1. I think we can lose a lot of the complexity of the token table parsing with a single table-as-dict function, something like:

  ```python
  def token_table_as_dict(browser):
    # get table
    return [
        [element.text for element in row.find_elements("td")]
        for row in find_elements('tr')
    ]
  ```

2. figure out the races in the progress test to get it passing on CI

3. I think we could also do with less abstraction in the Locators, especially for generic locators like `By.TAG, 'table'`. Using the locators directly in the tests will be clearer and simpler a lot of the time, I think.

minrk (approved):

PR #4209, opened 2022-11-14
avoid contraction in setup.rst

ArafatAbdussalam:
When I needed to set up Linux on my windows, i found out the easier to do so was through Windows Subsystem for Linux. Hence, I needed to add the guide to the development setup


minrk:
@ArafatAbdussalam this PR doesn't appear to make the changes you described. It only changes "it's" to "it is". Is that what you intended, or did you intend to describe setting up with WSL?

PR #4214, opened 2022-11-15
Add active users prometheus metrics

yuvipanda:
These are *extremely useful* for people advocating for more resources for their JupyterHubs, but a little difficult to calculate without a full scale log ingestion and analytics pipeline (such as ELK or equivalent). However, these are easy to calculate on the JupyterHub side at any given instance - these are fairly quick SQL queries. Prometheus can capture and store this as a timeseries, and provide valuble advocacy data that is hard to get otherwise.

This turns the metrics on by default, but only updates them every hour - which seems fine for metrics that don't change that often. This should reduce performance impact. Admins can also turn this off if needed.

I've had to implement this in many different ways in many different contexts, and it's also important to be able to *trust* these. Getting this from other grafana data can be tricky to validate - we had an experimental one in our grafana dashboards at some point in the past that we had to kill due to it being hard to validate (https://github.com/jupyterhub/grafana-dashboards/pull/45). This metric will provide an authoritative source of truth for this.

There is a `period` parameter to `active_users` that provides the time period for which the active users are calculated. It provides '24h', '7d', '30d' - accurate, matches other time descriptors used in prometheus, and far more accurate than 'daily' or 'monthly'.

Ref https://github.com/2i2c-org/infrastructure/issues/1888

### DIFF
--- a/jupyterhub/metrics.py
+++ b/jupyterhub/metrics.py
@@ -23,7 +23,7 @@ from datetime import datetime, timedelta
 from enum import Enum
 
 from prometheus_client import Gauge, Histogram
-from tornado.ioloop import IOLoop, PeriodicCallback
+from tornado.ioloop import PeriodicCallback
 from traitlets import Any, Bool, Integer
 from traitlets.config import LoggingConfigurable
 
--- a/jupyterhub/metrics.py
+++ b/jupyterhub/metrics.py
@@ -19,9 +19,15 @@ them manually here.
 
     added ``jupyterhub_`` prefix to metric names.
 """
+from datetime import datetime, timedelta
 from enum import Enum
 
 from prometheus_client import Gauge, Histogram
+from tornado.ioloop import IOLoop, PeriodicCallback
+from traitlets import Any, Bool, Integer
+from traitlets.config import LoggingConfigurable
+
+from . import orm
 
 REQUEST_DURATION_SECONDS = Histogram(
     'jupyterhub_request_duration_seconds',
@@ -43,6 +49,14 @@ RUNNING_SERVERS = Gauge(
 )
 
 TOTAL_USERS = Gauge('jupyterhub_total_users', 'total number of users')
+
+DAILY_ACTIVE_USERS = Gauge(
+    'jupyterhub_daily_active_users', 'number of users who were active in the last 24h'
+)
+
+MONTHLY_ACTIVE_USERS = Gauge(
+    'jupyterhub_monthly_active_users', 'number of users who were active in the last 30d'
+)
 
 CHECK_ROUTES_DURATION_SECONDS = Histogram(
     'jupyterhub_check_routes_duration_seconds',
@@ -200,3 +214,74 @@ def prometheus_log_method(handler):
         handler=f'{handler.__class__.__module__}.{type(handler).__name__}',
         code=handler.get_status(),
     ).observe(handler.request.request_time())
+
+
+class PeriodicMetricsCollector(LoggingConfigurable):
+    """
+    Collect metrics to be calculated periodically
+    """
+
+    active_users_enabled = Bool(
+        True,
+        help="""
+        Enable daily_active_users and monthly_active_users prometheus metric.
+
+        daily_active_users reports number of users who have registered *some* kind of activity
+        in the last 24h. monthly_active_users reports it for the last 30 days.
+        """,
+        config=True,
+    )
+
+    active_users_update_period = Integer(
+        60 * 60,
+        help="""
+        Number of seconds between updating daily_active_users and monthly_active_users metric.
+
+        To avoid extra load on the database, this is only calculated periodically rather than
+        at per-minute intervals. Defaults to once an hour.
+
+        Both the metrics are updated at the same time so they provide a consistent snapshot of
+        stats at that point in time.
+        """,
+        config=True,
+    )
+
+    db = Any(help="SQLAlchemy db to use for performing queries")
+
+    def update(self):
+        """
+        Update all these metrics!
+        """
+        # daily cutoff
+        daily_cutoff = datetime.now() - timedelta(days=1)
+        monthly_cutoff = datetime.now() - timedelta(days=30)
+
+        daily_active_users = (
+            self.db.query(orm.User)
+            .filter(orm.User.last_activity >= daily_cutoff)
+            .count()
+        )
+        monthly_active_users = (
+            self.db.query(orm.User)
+            .filter(orm.User.last_activity >= monthly_cutoff)
+            .count()
+        )
+
+        DAILY_ACTIVE_USERS.set(daily_active_users)
+        MONTHLY_ACTIVE_USERS.set(monthly_active_users)
+        self.log.info(f'Found {daily_active_users} active users in the last 24h')
+        self.log.info(f'Found {monthly_active_users} active users in the last 30d')
+
+    def start(self):
+        """
+        Start the periodic update process
+        """
+        if self.active_users_metric_enabled:
+            # Setup periodic refresh of the metric
+            pc = PeriodicCallback(
+                self.update, self.active_users_update_period * 1000, jitter=0.01
+            )
+            pc.start()
+
+            # Update the metrics once on startup too
+            self.update()
--- a/jupyterhub/metrics.py
+++ b/jupyterhub/metrics.py
@@ -221,7 +221,7 @@ class PeriodicMetricsCollector(LoggingConfigurable):
     Collect metrics to be calculated periodically
     """
 
-    active_users_enabled = Bool(
+    active_users_metrics_enabled = Bool(
         True,
         help="""
         Enable daily_active_users and monthly_active_users prometheus metric.
@@ -232,7 +232,7 @@ class PeriodicMetricsCollector(LoggingConfigurable):
         config=True,
     )
 
-    active_users_update_period = Integer(
+    active_users_metrics_update_period = Integer(
         60 * 60,
         help="""
         Number of seconds between updating daily_active_users and monthly_active_users metric.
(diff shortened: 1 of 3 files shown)


yuvipanda:
From looking at the lastactivity handler code, it looks like the user objects last_activity field is updated anytime last_activity is updated for any of its servers, which matches what I expect.

minrk:
Makes sense! Do you think 1d and 30d are standard enough intervals that we shouldn't provide config to allow deployments to _specify_ the intervals (e.g. 7d)?

consideRatio:
> Do you think 1d and 30d are standard enough intervals that we shouldn't provide config to allow deployments to specify the intervals (e.g. 7d)?

Hmm hmmm, I think if we opt to make it configurable, we would make it a bit harder to name the metrics etc and perhaps provide standarized jupyterhub dashboards. Is it worth providing this customization if it means providing a reliable standard solution harder?

manics (commented):
> provide config to allow deployments to specify the intervals (e.g. 7d)

That was my first thought too! Could make it take an array of intervals (in hours, or maybe seconds, so people can have more fine grained usage? Useful for analysing use over a working day)

And the metric could be something like `jupyterhub_active_users_per_$N`?

minrk:
> we would make it a bit harder to name the metrics

If we e.g. limit it to a number of days, we could name them like:

```
active_users_last_{days}d
```

where the number of days is in the metric name. Then the dashboard would only have to assume that one or more values for days have been chosen.

Alternately, we could us a single metric and store the interval as a _label_:

```
active_users_last_days{days=30}
```

which could then probably make a single chart with a line for each interval, no matter what intervals are collected. I _think_ that might be less efficient for prometheus? I'm not sure.


Because of this complexity, I don't think we should add configuration for it if it's not likely folks want it (we don't have any other configurably-defined metrics). I just wanted to get folks' opinions on how likely we think that is, as weighed against the complexity of dealing with it.

Another possible criticism of this metric: it won't work for BinderHub or any deployment that deletes inactive users, since it only tells you how many users _still in the database_ who were active in that interval. To allow counting 'deleted' users, we'd need trash-like mechanism for deleting users: mark users as deleted to remove them from _most_ API endpoints, but actually remove from the db after some interval, e.g. 31d.

yuvipanda:
I would prefer to not make this configurable. I think daily and monthly is good enough for most use cases. We can add a *yearly* later if anyone wants, but IMO that's not super common.

I think the TODO items here are:

1. Make sure to use utc here
2. Add a test

I don't think we should make this super configurable (lots of complexity!) or try to solve *all* cases (like binder) here.

jupyterhub/metrics.py
```diff
@@ -200,3 +214,74 @@ def prometheus_log_method(handler):
...
+        self.log.info(f'Found {monthly_active_users} active users in the last 30d')
+
+    def start(self):
+        """
+        Start the periodic update process
+        """
+        if self.active_users_metrics_enabled:
+            # Setup periodic refresh of the metric
+            pc = PeriodicCallback(
+                self.update, self.active_users_metrics_update_period * 1000, jitter=0.01
```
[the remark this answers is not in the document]
yuvipanda: @minrk in our case, there's only one process hitting the db so probably not. However, I think all periodic callbacks should have jitter just in case anyway, so I kept it in.

consideRatio:
I lack experience here, but I'm considering the difference between multiple metrics like `active_users_last_1day` `active_users_last_30days` and what @minrk considered with a label (with a predefined set of values - 1, 30).

> Alternately, we could us a single metric and store the interval as a _label_:
> 
> ```
> active_users_last_days{days=30}
> ```
> 
> which could then probably make a single chart with a line for each interval, no matter what intervals are collected. I _think_ that might be less efficient for prometheus? I'm not sure.

Hoping its reasonable to use a label like that to avoid defining multiple metric names, thinking it will be easier to consume in jupyterhub/grafana-dashboards.

minrk:
Makes sense! I also think we shouldn't implement the rigorous 'monthly active users' metric here, I was trying to note a case where expectations from the metric name/description might not match the behavior.

If we're not making it configurable, does 7d make sense? I'd usually use a multiple of weeks for a metric like this, since it captures the natural user activity cycle. But if nobody would use it, we should leave it out.

consideRatio:
pre-commit.ci run

consideRatio:
Reopened to get tests against latest main branch, which should make selenium test not fail.

yuvipanda:
@minrk @consideRatio @manics so I've renamed this to `active_users`, and added a `period` label that can take one of `24h`, `7d`, `30d` as labels. This is more accurate than 'daily' and 'monthly', and adds the '7d' as well. Thank you for the valuble input :)

If this design looks ok, I can work on adding tests. Here's an example of how the metrics are:

```
# HELP jupyterhub_active_users number of users who were active in the given time period
# TYPE jupyterhub_active_users gauge
jupyterhub_active_users{period="24h"} 1.0
jupyterhub_active_users{period="7d"} 494.0
jupyterhub_active_users{period="30d"} 3140.0
```

manics:
```
jupyterhub_active_users{period="24h"} 1.0
jupyterhub_active_users{period="7d"} 494.0
jupyterhub_active_users{period="30d"} 3140.0
```
I like this, and it's easily extensible in future 🌞

consideRatio:
> If this design looks ok, I can work on adding tests. Here's an example of how the metrics are:
> 
> ```
> # HELP jupyterhub_active_users number of users who were active in the given time period
> # TYPE jupyterhub_active_users gauge
> jupyterhub_active_users{period="24h"} 1.0
> jupyterhub_active_users{period="7d"} 494.0
> jupyterhub_active_users{period="30d"} 3140.0
> ```

@yuvipanda looks great!!! :tada:

minrk:
Sounds great, @yuvipanda!

consideRatio:
@yuvipanda you have not pushed recent changes but I suspect you may have intended to do so

yuvipanda:
@consideRatio whoops good catch, fixed!

jupyterhub/metrics.py
```diff
@@ -44,6 +50,16 @@
...
+ACTIVE_USERS = Gauge(
+    'jupyterhub_active_users',
+    'number of users who were active in the given time period',
+    ['period'],
+)
+
+MONTHLY_ACTIVE_USERS = Gauge(
+    'jupyterhub_monthly_active_users', 'number of users who were active in the last 30d'
+)
+
```
consideRatio, line 62:
```suggestion
```

jupyterhub/metrics.py
```diff
@@ -200,3 +230,69 @@ def prometheus_log_method(handler):
...
+        To avoid extra load on the database, this is only calculated periodically rather than
+        at per-minute intervals. Defaults to once an hour.
+        """,
+        config=True,
+    )
+
+    db = Any(help="SQLAlchemy db to use for performing queries")
+
+    def update(self):
```
consideRatio, line 264: It seems fine that the class is designed ahead of time to be general purpose for other metrics to updated periodically, but I think the update function needs to be specific to active_users. So `update_active_users` for example?

The `start` function is called, and checks if active_users_metrics_enabled is truthy, and then a periodic callback is scheduled to run this `update` function. So that makes me think it should be `update_active_users` rather than a general `update` function.

minrk:
@yuvipanda I'd like to make 3.1 this week, do you think you'll have a chance to finish this? If not, I can have a look at the tests.

yuvipanda:
@minrk i don't think I can add tests this week, would *LOVE* if you could finish it up!!!!

jupyterhub/metrics.py
minrk: I'm torn! I think `update` is the right name. This is a generic class for periodically updating metrics. Active users happens to be the only kind we have right now, so that's the only checks we have, but adding more wouldn't change what the update method should be called. It might move exactly where the `enabled` checks happen, though.

minrk (approved):

minrk:
I added the tests and all looks good. One outstanding question to @consideRatio above, then happy to merge.

jupyterhub/metrics.py
manics: It sounds like this class is trying to be both an Abstract class (`start()` and `update()` methods) and a concrete implementation (defines `active_users_metrics_enabled` and `active_users_metrics_update_period` which will be inherited by any subclasses of `PeriodicMetricsCollector` even if it's used for some other metrics. Would it make sense to separate out the Abstract class, or is that over engineering things?

consideRatio: Update can be the name, but then the active_users_metrics_update_period is a bit off.

Since its now a class designed for future metrics to be updated periodically, i think it should be natural on how to go about it. But, there can only be ome update period as its done now right - that is why either the update name or the active_users_metrics_update_period doesnt feel right.

consideRatio: In other words, pretend another metrics is added to use this class. How should it be added? Maybe it was meant as a base class, and when a new class is added we need to extract active_users_metric, or maybe it was meant as a single class with one update interval, or a class that should handle multiple update intervals. A decision should be made now, or a comment indicating that whoever adds something in the future can do as they think makes sense.

minrk: > either the update name or the active_users_metrics_update_period doesnt feel right.

👍 I agree, they should probably match. Since @yuvipanda named the config value with `active_users`, I made the method name match, so presumably adding another periodically updated metric will mean a new config for enabled/interval and a new update_$metrics method.

> It sounds like this class is trying to be both an Abstract class

@manics I don't think it's meant to be an abstract class at all, I think it's meant to be the one and only periodic metrics updater instance. So if we have other periodically updated metrics, they would be added to the update in this class, not a new class. I think the only question is whether the period should be shared across all metrics or per metric group, and naming follows from that. Naming of methods is a private implementation detail and not something I'd worry about, but naming of config is a user-facing API we need to consider because it's a pain to change.

consideRatio (approved):
Wieee looks great!

jupyterhub/metrics.py
yuvipanda: Yep, it was meant to be sort of a singleton, rather than an abstract class. I am extremely happy with the state it was merged in! If we add more of these metrics in the future, i think it makes sense to add individual traitlet knobs for those, and we can even coalesce periodcallbacks if needed.

Thank you so much @minrk for taking this to completion.

yuvipanda:
Thank you so much for taking this through, @minrk!!!

yuvipanda:
https://github.com/jupyter-server/jupyter_server/pull/1471 adds a similar metric for jupyter_server!

Issue #3982, opened 2022-07-18
Preparing for 3.0


minrk:
I believe the updated scopes work is ready to release with https://github.com/jupyterhub/jupyterhub/pull/3877, which means starting the release process. The first decision in preparing this release is: do we call it 2.4 or 3.0?

Todo list:

- [x] Review/merge remaining PRs
- [x] Review docs for new/changed features
- [x] changelog PR
- [x] Decide 2.4 vs 3.0 (details below)

[@manics](https://github.com/jupyterhub/jupyterhub/pull/3976#issuecomment-1185296439) re: dropping Python 3.6 support:

> Are you releasing this in 2.4.0 or 3.0.0?

The next release has relatively little user-facing changes (mostly custom scopes - a pretty advanced feature), and _should_ have no backward-compatible changes. _But_ there are reasons to call it 3.0 anyway:

- It includes a schema change for the roles->scopes migration, which means db upgrade.
- Dropping a Python version support _could_ be considered a backward-incompatble change, but with tooling (pip, conda, etc.) this doesn't have any real issues as older Python users simply won't get the new version as an installation candidate.

I think we can _decide_ that for JupyterHub:

- Schema change (excluding wholly new tables which have no migration operation) only happens in major versions
- We can similarly _decide_ that we only drop Python versions in major versions.
 
Making either of these choices would make the next release 3.0.

Neither of these are strictly required by semantic versioning, and we can always decide on a case by case basis at release time, as we're doing here.

A [github compare](https://github.com/jupyterhub/jupyterhub/compare/2.3.1...main) is not useful because of the switch to isort in pre-commit.

To me, the biggest reason in favor of 2.x is that it will _feel_ like a minor release, and anything that breaks is a bug and not an intentional change.

And the biggest reason in favor of 3.0 is the db upgrade step, and deciding we won't do that except in major revisions, since it adds an upgrade step for non-sqlite users, and makes rolling back more difficult (for everyone).

manics:
I think 3.0 makes sense given the DB change.

From the point of view of maintaining our distributions (Z2JH, TLJH) do you think any JupyterHub upgrade that requires a manual step by an administrator (in this case `jupyterhub upgrade-db`) could count as a major change?

consideRatio:
I lean towards 3.0 due to the need to do a db-upgrade =/

I consider the following:
[...]

PR #4215, opened 2022-11-15
Add an index to last_activity for user

yuvipanda:
This is queried by https://github.com/jupyterhub/jupyterhub/pull/4214, and indexing would make this much faster! The query is probably already fast enough, but this would help.


consideRatio:
Do you think this will require a database upgrade or similar @yuvipanda?

Overall, adding an index is a compromise right? I can imagine its a compromise related to database size (larger), write speeds (slower?), read speeds (faster).

I also understad that last_activity is a column in the User table for the database, making it scale with the amount of users, and that seems acceptable.

consideRatio:
I went ahead and added "maintenance" label, thinking this isn't relevant to surface as "new" in the changelog as its just a "new under the hood" feature.

minrk:
@consideRatio for classification, I'd usually call a performance improvement an 'enhancement'

consideRatio:
@minrk makes sense! I removed the maintenance label to ensure that there is only one category for github-activity to summarize this PR.

manics:
> it would only show up for new deployments with a fresh database

Will the new database have a different Alembic version/id? Or would we have two DB schemas with the same version?

yuvipanda:
@minrk hmm, I feel like we should have a different alembic version (so an upgrade *does* happen), but keep a minor version bump only - this isn't a particularly significant change. However, I'm not super tied to it :) And given a db migration is required anyway, I put an index on the spawner last_activity too, so hopefully we can provide an API for filtering servers by last_activity for more efficient spawning.

@consideRatio it's primarily a tradeoff with *very slightly* (like, very) write times and much faster read times! I think at the scales we are operating at, and the datatype this index is, I think of it as *almost* being free :D

yuvipanda:
@minrk I've added an index to the spawner last_activity, and added an alembic migration. Looks like a couple unique constraint fixes were not added to a migration earlier, so they're part of this now.

I have a weak feeling that we don't need to bump major version for this, but I'm happy to just take other people's lead here :)

yuvipanda:
uh oh!

`NotImplementedError: No support for ALTER of constraints in SQLite dialect. Please refer to the batch mode feature which allows for SQLite migrations using a copy-and-move strategy.`

manics:
The main reason for a major release with a schema change is it's disruptive to admins. You can't just upgrade JupyterHub and restart, you also need to upgrade the database.

minrk:
Interesting! How did you create the revision? When I ran the alembic command above, it didn't add any of those, I just got one line:

```python
    # ### commands auto generated by Alembic - please adjust! ###
    op.create_index(op.f('ix_users_last_activity'), 'users', ['last_activity'], unique=False)
    # ### end Alembic commands ###
```

What version of alembic/sqla did you use? I think we might need more developer documentation and/or tooling for creating these revisions. "I don't really know how it works, but it's always worked okay for me!" isn't a great situation to be in.

> I have a weak feeling that we don't need to bump major version for this, but I'm happy to just take other people's lead here :)

I mostly agree, but @manics and @consideRatio persuaded me in #3982. I think the argument for major bump for _any_ db upgrade is that it doesn't really matter to users _what's_ in the db upgrade, any db upgrade has the same consequences:

- explicit admin step needed to back up and upgrade db (automatic for sqlite, not others)
- rolling back the upgrade requires a db restore from backup or a tricky downgrade sequence, which is nontrivial and potentially requires a lot of manual intervention _while the hub is down_

The second point is actually because we've never implemented a smooth workflow for db downgrade. I think right now, you have to:

1. make sure you still have the _newer_ jupyterhub installed that you aren't going to use anymore
2. make sure hub isn't running
3. figure out what revision to rollback to (not documented!)
4. run `python -m jupyterhub.dbutil alembic downgrade <old_revision>`
5. actually downgrade the jupyterhub package
6. start jupyterhub with the rolled back version

This applies to _all_ implementations, including sqlite. Restoring the old db may be easier (copy a file in sqlite), but will lead to inconsistency if there are running servers, etc. created since the upgrade.

So the major bump is largely a signal that this upgrade ought to be a one-way trip, because going back is a big pain, regardless of what the actual db changes are.

But I'm open to it being a minor revision, too! It's not a clear choice, to me.

minrk:
If it matters, this index doesn't seem to make a difference in performance of the queries in #4214 up to 100k users (with sqlite, I haven't run with postgres or mysql, which may well do better with indices):

visualization

[produced with this script](https://gist.github.com/ea0c2102ffe3f09165f8970e647b9976)

minrk:
Managed to run it with postgres and mysql, also no apparent impact from the index:

visualization-4

yuvipanda:
@minrk PERFECT! Let's skip it completely then :)

PR #4216, opened 2022-11-16
ci: flake8, cleanup unused/redundant config

consideRatio:
- `E###` warnings were already ignored by E
- Sorting of import warnings won't matter as we have black and isort
- D400: First line should end with a period

  This is not flake8 configuration, but related to `pydocstyle`, which
  isn't used.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -3,11 +3,12 @@
 # E: style errors
 # W: style warnings
 # C: complexity
+# D: docstring warnings (unused pydocstyle extension)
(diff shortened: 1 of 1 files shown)


minrk:
Can we leave the D (maybe just `D` instead of `D400`)? It can still get picked up in folks' editors if they have flake8 integration and docstring plugins installed.

consideRatio:
@minrk deal, done!

PR #4219, opened 2022-11-16
flake8: check unused/wildcard imports and redefinitions

consideRatio:
- flake8: check F403 about import *
- flake8: check F811 about redefinition of unused name
- flake8: check F401 about unused imports


examples/service-fastapi/app/service.py
```diff
@@ -51,7 +51,7 @@ async def me(user: User = Depends(get_current_user)):
 
 
 @router.get("/debug")
-async def index(request: Request, user: User = Depends(get_current_user)):
+async def index(request: Request, user: User = Depends(get_current_user)):  # noqa
```
minrk, line 54: This one could be called `debug` for the page name. It's not technicaly a _bug_ that two handler functions have the same name, but it's not intentional and probably due to a copy/paste.

minrk (approved):
Looks great! Would you like to try to make this PR with [autoflake](https://github.com/PyCQA/autoflake) and see if that makes things easier? ('no' is a fine answer!)

It does appeal to me to have rules enforced automatically where possible.

I checked and the [default args](https://github.com/PyCQA/autoflake/blob/main/.pre-commit-hooks.yaml#L9-L11) are a bit harsh, so might need to be trimmed to `--in-place`, so the actual options can be specified in pyproject.toml.

minrk:
If we want to try autoflake, I think it would look like:

```toml
# pyproject.toml
[tool.autoflake]
in_place = true
remove_all_unused_imports = true
ignore_init_module_imports = true
```

and 

```yaml
  - repo: https://github.com/PyCQA/autoflake
    rev: v1.7.7
    hooks:
      - id: autoflake
        args:
          - --in-place
```

I'd be closer to 50-50 on F841 if we used autoflake for unused variables, rather than opposed. I care a lot less about code churn like that if it's not _me_ who needs to do the churning :)

consideRatio:
While we could put things in the pyproject.toml, I've opted to put this in pre-commit-config.yaml. I see it as a compromise. The downside is that someone having autoflake stuff built into their editor won't make use of this. The upsides is that someone wanting to get this into another project doesn't have to parts of configuration together, and that the configuration is available in direct association with pre-commit which is blamed if something goes wrong and someone looks to understand what is it about.

@minrk what do you think about pre-commit-config.yaml for autoflake vs pyproject.toml?

I realized that when autoflake removes unused imports, it replaces them with `pass` statements. These are then typically removed by another pass of autoflakes, unless `--ignore-pass-statements`.

For now, lets not enforce F841. If its to be enforced via autoflakes we need to do some manual inspections I'd rather not do now at least.

minrk:
> what do you think about pre-commit-config.yaml for autoflake vs pyproject.toml?

I think it's always better to put tool config in the standard location, rather than only where it will be picked up by pre-commit. That way it behaves appropriately no matter how it's invoked (e.g. via editor integration or manual invocatino). I think the only thing that should go in pre-commit is what's necessary for pre-commit itself to work (e.g. flags like in-place, check), but the tool should always pick up the same options both inside and outside pre-commit.

I don't think it was a good choice for the pre-commit hook itself to [default to several non-default options](https://github.com/PyCQA/autoflake/blob/main/.pre-commit-hooks.yaml#L9-L11), so the hook is _not_ like calling the tool directly.

> For now, lets not enforce F841. If its to be enforced via autoflakes we need to do some manual inspections I'd rather not do now at least.

👍

testing/jupyterhub_config.py
```diff
@@ -4,7 +4,7 @@
 to enable testing without administrative privileges.
 """
 
-c = get_config()  # noqa
+c = get_config()
```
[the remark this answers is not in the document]
consideRatio: > This should keep # noqa for the undefined name on get_config()

It is a manual commit where I remove it, after concluding that we have `get_config()` as an excluded thing in config.

Do you think we should retain it because people may copy-paste it and not have our flake8 config for example, or that it could help people get its something magic about it?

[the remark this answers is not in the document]
consideRatio: Decision: remove or retain ade8999?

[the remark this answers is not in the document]
consideRatio: https://github.com/jupyterhub/jupyterhub/blob/5ea93add289f9905e4957f8a1a8d6d01ce5ab634/.flake8#L12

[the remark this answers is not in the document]
minrk: Yes, that's precisely why - we exclude it, but folks who _use_ the examples won't carry over our config. (we could maybe even remove our `builtins` flake8 config to ensure we add these, but no need in this PR).

traitlets generated config files now do the same thing: https://github.com/ipython/traitlets/pull/750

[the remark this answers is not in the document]
consideRatio: :+1: sorry for the noise! Rebased and didn't pick that commit now.

minrk (approved):

consideRatio:
@manics this is ready for review!

Looking forward to testing the automation by writing `pre-commit.ci run` in #4214 to see if pre-commit.ci will cleanup an unused IOLoop import that made me end up thinking about this in the first place.

examples/service-fastapi/app/__init__.py
```diff
@@ -1 +1 @@
-from .app import app
+from .app import app  # noqa
```
manics, line 1:
```suggestion
from .app import app

__all__ = ["app"]
```
The PEP8  suggestion is to use `__all__`
https://peps.python.org/pep-0008/#public-and-internal-interfaces

jupyterhub/__init__.py
```diff
@@ -1 +1 @@
-from ._version import __version__, version_info
+from ._version import __version__, version_info  # noqa
```
manics, line 1:
```suggestion
from ._version import __version__, version_info

__all__ = ["__version__", "version_info"]
```

jupyterhub/singleuser/__init__.py
```diff
@@ -2,8 +2,8 @@
 
 Contains default notebook-app subclass and mixins
 """
-from .app import SingleUserNotebookApp, main
-from .mixins import HubAuthenticatedHandler, make_singleuser_app
+from .app import SingleUserNotebookApp, main  # noqa
+from .mixins import HubAuthenticatedHandler, make_singleuser_app  # noqa
```
manics, line 6:
```suggestion
from .app import SingleUserNotebookApp, main
from .mixins import HubAuthenticatedHandler, make_singleuser_app

__all__ = ["SingleUserNotebookApp", "main", "HubAuthenticatedHandler", "make_singleuser_app"]
```

manics (commented):
I made some suggestions regarding using `__all__`, but the otherwise looks good to me!

examples/service-fastapi/app/__init__.py
consideRatio: I've not been so keen on `__all__` because it require us to use more code in these situations, even though just having two variables makes it something of little importance. I figure the key benefit of `__all__` is when you have things declared in `__init__.py` that isn't supposed to be exported, and you can omit them in an `__all__` list.

We have three `from ... import *` statements in two files, something the PEP also suggest we avoid. The `autoflake` autoformatter can do something related to star imports, documented like this:

```
  --expand-star-imports
                        expand wildcard star imports with undefined names; this only triggers if there is only one star import in the file; this is skipped if there are any uses of `__all__` or `del` in the file
```

Ah I figure we shouldn't overthink this detail though, so lets go with what you suggested - either way is fine as long as we don't have massive lists that are manually maintained in my mind!

manics: I did originally add a comment along the lines of "should we have some team code style guide for the use of `import *`", then decided this wasn't the time to start that discussion :smile:

manics (approved):
Thanks!

PR #4223, opened 2022-11-18
maint: add test to extras_require, remove greenlet workaround, test final py311, misc cleanup

consideRatio:
## Add test to extras_require

Following this PR, the test requirements are now installable via `pip install -e ".[test]"`, and the other tooling previously found in dev-requirements.txt (tbump, github-activity, pre-commit) are now mentioned in RELEASE.md and the contributing documentation.

## Additional changes

- I added `jsonschema` as a test requirements as it were and removed an outdated workaround related to `attrs` in the test requirements
- I saw `urllib3` added in tests, but I saw no reference to that being used so I removed it.
- I saw html5lib added explicitly for beautifulsoup4 and figured it made more sense to install via beautifulsoup4's extras_require, allowing it to put version constraints on it if needed etc.
- I made us test against py311 instead of a release candidate, and converted a few test cases with python 3.10 to python 3.11
- I removed a mock of greenlet used before it became Python 3.11 compatible


manics:
I think having `doc` as a separate target is fine, though the split between `dev` and `test` is confusing, since testing is an integral part of development.

consideRatio:
> since testing is an integral part of development.

I agree that logically development implies testing, and its probably always relevant to do `pip install -e ".[dev,test]"` and not just `pip install -e ".[dev]"`, unless you opt to run tests systamtically in the CI etc...

I feel quite fine about the situation though, as our contributing doc now suggests that you do `pip install -e ".[dev,test]"` together, and its never been suggested to just install "dev" or "test" by themselves in our contributing docs. So, I figure the practical confusion is that someone just reads about the "doc" extras_require from the setup.py or similar and assumes it implies test capability.

The practical benefits as I see them are these I think, but I don't see a practical downside:
- avoid pulling more than needed from pypi / faster setup
- documentation through configuration, helping users distinguish the purpose of github-activity, pre-commit, tbump from the test related dependencies

consideRatio:
@manics is there a comment or similar you think could help avoid the confusion aspect? Perhaps an inline comment next to the respective dev/doc/test definition clarifying what should be listed under each respectively?

manics:
How about rename `dev` to `release` (or `release-tools`) since the deps are mostly for the release manager not developers? Then rename `test` to `dev`, and move `pre-commit` to `dev`. A releaser shouldn't need pre-commit, as the code should already be merged, whereas developers are expected to ensure their code passes pre-commit before merging.

consideRatio:
I think a bit attached to the idea of having `test` as its even more clear than `dev` and is a name I've often seen used for this. `dev` could imply just tests, or something more, or something else. I'm less attached to the idea of having `dev`.

Hmm so `github-activity` and `tbump` is only used for RELEASE.md, and `pre-commit` will require manual steps anyhow to be useful.

@manics what do you think about:
- removing `dev` entirely and then letting `test` remain `test`
- optionally letting `github-activity` and `tbump` be a `release-tools` extras_require grouping

manics:
That sounds fine to me! I'm not too fussed over the naming, as long as it's a single target for dev/test.

manics:
How about keep the requirements in `*.txt` files, but read them in `setup.py` to populate the install targets? Then you get the benefits of both. E.g.
https://github.com/ansible/ansible/blob/1bda6750f5f4fb8b01de21d1949b02d7547ff838/setup.py#L10
```py
install_requires = (here / 'requirements.txt').read_text(encoding='utf-8').splitlines()`

setup(
    install_requires=install_requires,
```
Might need to call `.strip()` and remove lines begining `#`.

manics:
Have you tried dynamic requirements in `pyproject.toml`?
https://stackoverflow.com/a/73600610

consideRatio:
> Have you tried dynamic requirements in pyproject.toml?
https://stackoverflow.com/a/73600610

Okay true then we could retain docs/requirements.txt from the perspective setup.py / pyproject.toml still without issues.

**Decision point**
- We stop installing the jupyterhub python package and ensure we setup jupyterhub to be on the path, and let docs requirements reference the root folders requirements.txt
- We have it similar to what we have now, with a `doc` in extras_require
  - Either with or without a docs/requirements.txt file referencing "..[doc]"

minrk:
1 is the simplest, right? If it works without too much (ideally any) RTD setup, let's go with that.

I've [done 3 in ipyparalllel with hatch](https://github.com/ipython/ipyparallel/blob/ec4ad17d61d9b1ab7bd11058e7ac37d76886e21b/hatch_build.py#L18). It's not too complex, but perhaps more than we want to deal with here.

minrk (approved):

minrk:
Thanks, @consideRatio!

consideRatio:
With this approved by @minrk and where I think @manics review input is addressed, I'll go for a merge and submit another PR to: refresh conf.py for readability etc, add opengraph, add rediraffe.

PR #4230, opened 2022-11-29
make current_user available to handle_logout hook

minrk:
Clear the user prior to rendering the logout page, rather than in clear_login_cookie. ensures current_user is accessible in `handle_logout` hook, so the hook can take the user into account.

closes #4202


yuvipanda:
Thanks @minrk

PR #4232, opened 2022-11-30
changelog for 3.1.0

minrk:
I think we're due for a release, especially given the admin bugs in the docker image. It's a small one.

I'd like to get the following PRs in as well:

- [x] #4230
- [x] #4214


consideRatio:
Maybe also include the 1.5.1 changelog in this before this is merged so its seen from the latest tagged docs build as well.

consideRatio (commented):
With the changelog list updated to include recent merged PRs, and the 1.5.1 entry in, this LGTM and can be merged -> a release can be made.

Possibly also update the date to include the day of the release and not just the month.

minrk:
Thanks for the reviews! I'm home sick today, but if I (and everyone else in my house) feel better on Monday, I'll cut 1.5.1 first, then 3.1.0.

minrk:
OK, I've imported the 1.5.1 changelog and updated the PR list. Making the release shortly.

PR #4233, opened 2022-11-30
ci: update database image versions

minrk:
mysql: 8.0
postgres: 15.1

### DIFF
--- a/ci/docker-db.sh
+++ b/ci/docker-db.sh
@@ -22,7 +22,7 @@ if [[ "$DB" == "mysql" ]]; then
     # ref server: https://hub.docker.com/_/mysql/
     # ref client: https://dev.mysql.com/doc/refman/5.7/en/setting-environment-variables.html
     #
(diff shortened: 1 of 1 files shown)


jupyterhub/tests/test_db.py
```diff
@@ -35,7 +35,9 @@ def generate_old_db(env_dir, hub_version, db_url):
     check_call([env_py, populate_db, db_url])
 
 
-@pytest.mark.parametrize('hub_version', ['1.0.0', "1.2.2", "1.3.0", "1.5.0", "2.1.1"])
+# changes to this version list must also be reflected
+# in ci/init-db.sh
+@pytest.mark.parametrize('hub_version', ["1.1.0", "1.2.2", "1.3.0", "1.5.0", "2.1.1"])
```
minrk, line 40: Had to bump the oldest version to test from because 1.0 can't create a db with recent mysql, due to #2712

PR #4234, opened 2022-11-30
changelog for 1.5.1

minrk:
just backporting the db fix (#4076)

cc @mriedem

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,7 +11,7 @@ repos:
     rev: v2.2.1
     hooks:
       - id: prettier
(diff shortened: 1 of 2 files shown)


mriedem (approved):
lgtm, thanks!

consideRatio:
Nice LGTM! There is an action point remaining to update the changelog in the main branch as well now right?

I'm on mobile on a bus so i'll not go for that now myself.

minrk:
Yes, adding it to #4232 as you suggested makes sense.

minrk:
1.5.1 is published.

PR #4236, opened 2022-11-30
deprecate JupyterHub.extra_handlers

minrk:
I don't think we should encourage or support in-process extensions for JupyterHub. Point to services, instead.

I think we've gently discouraged this for a long time, but haven't had an official deprecation/warning. I've no plans to remove the feature, but I think it's appropriate to have a "please don't do this" message, and keep it out of the documentation.

closes #3053 because I don't think we should add docs encouraging something I don't think folks should do.


GeorgianaElena (approved):
Thanks @minrk!

PR #4237, opened 2022-11-30
Do not ignore a configured `default_server_name`

behrmann:
I'm currently writing a spawner (actually) and wondered why with `c.JupyterHub.default_server_name` set to something different than the empty string when accessing `self.name` in the spawner. This change let me remove the kludges where I hardcode a default name.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -2708,7 +2708,7 @@ class JupyterHub(Application):
             statsd=self.statsd,
             implicit_spawn_seconds=self.implicit_spawn_seconds,
             allow_named_servers=self.allow_named_servers,
-            default_server_name=self._default_server_name,
+            default_server_name=self.default_server_name,
             named_server_limit_per_user=self.named_server_limit_per_user,
             oauth_provider=self.oauth_provider,
             oauth_no_confirm_list=oauth_no_confirm_list,


minrk:
Do you have `JupyterHub.allow_named_servers = True`? The default_server_name config is ignored unless allow_named_servers is True, and if it is True, `_default_server_name` has the same value.

behrmann:
Yes, I do. Sorry, I should have specified that. 

Currently I do [`self.name or "default"`](https://github.com/behrmann/systemdspawner/blob/staticspawner/systemdspawner/staticsystemdspawner.py#L31) in my spawner (with `"default"` being the value I set `c.JupyterHub.default_server_name` to). If that's not intended to work I'd be happy about a hint for a different workaround.

minrk:
Sorry if I'm misunderstanding, but you _do_ have `allow_named_servers = True`? Because if that's the case, this PR should have no effect (`app._default_server_name == app.default_server_name` is always True). This PR as it is only has the effect to enforce always using named servers when named servers are disabled, which should be mutually exclusive, and I suspect will create some errors. Changing this value affects lots of URLs, e.g. user urls will be `/user/name/default/` instead of `/user/name/`, it's not just the `server.name` attribute.

To try to rephrase a bit: the default server name configuration selects which _named server_ is launched by default, it does not change the name of the server whose name was previously `''`, that server still exists and still has the same name (`''`) in this situation.

If an empty name is not valid for your template purposes, to avoid collisions you'll need to use a name that can't be produced by escaping a valid server name. I don't know how systemd-escape works, but with escapism, you can do this by inserting the escape character somewhere that isn't followed by an escape code (e.g. escapism will never produce `default-server` if the escape char is `-`, because `escape('default-server')` would produce `default-2dserver`).

If named servers aren't enabled, then using `'default'` is just fine.

behrmann:
> Sorry if I'm misunderstanding, but you do have allow_named_servers = True?

As I said, I do, so, the changed PR title is incorrect.

> To try to rephrase a bit: the default server name configuration selects which named server is launched by default, it does not change the name of the server whose name was previously `''`, that server still exists and still has the same name (`''`) in this situation.

Thanks for the clarification. The docs are a bit lacking in that regard and I misunderstood the intent of the option to be to change exactly this default `''`.

> If an empty name is not valid for your template purposes, to avoid collisions you'll need to use a name that can't be produced by escaping a valid server name. I don't know how systemd-escape works, but with escapism, you can do this by inserting the escape character somewhere that isn't followed by an escape code (e.g. escapism will never produce default-server if the escape char is -, because escape('default-server') would produce default-2dserver).

Any string for an instance name is fine, except the empty one (and within a total length limit on the file name). Therefore I wanted to force every server to have a name, but it seems to me, that's not possible. Not much of a problem. It's more cases, but then I'll have to support the default unnamed and named servers separately.

Thanks for the input!

minrk:
Sorry! I know it's complicated, and the empty default name causes problems in some places. #4240 adds some details to the docs and warnings. I hope that helps!

minrk:
I think it's also a valid feature request to _disable_ the `name=''` server. Then you'd be able to assume spawner.name is non-empty, at least for a deployment. No Spawner implementation could safely assume this, though.

behrmann:
> No Spawner implementation could safely assume this, though.

That could be yet another feature request, maybe, a `sanity_check` class method for spawners that the Hub calls on startup for the configured spawner, so that spawners could communicate whether they don't support a specific feature or expect a certain option to be set.

PR #4245, opened 2022-12-02
doc: fix formatting of spawner env-vars

manics:
Some env-vars on https://jupyterhub.readthedocs.io/en/stable/reference/spawners.html are incorrectly formatted.

### DIFF
--- a/docs/source/reference/spawners.md
+++ b/docs/source/reference/spawners.md
@@ -297,36 +297,36 @@ Additional variables can be specified via the `Spawner.environment` configuratio
 
 The process environment is returned by `Spawner.get_env`, which specifies the following environment variables:
 
-- JUPYTERHUB_SERVICE_URL - the \_bind* URL where the server should launch its HTTP server (`http://127.0.0.1:12345`).
+- `JUPYTERHUB_SERVICE_URL` - the _bind_ URL where the server should launch its HTTP server (`http://127.0.0.1:12345`).
   This includes `Spawner.ip` and `Spawner.port`; _new in 2.0, prior to 2.0 IP, port were on the command-line and only if specified_
-- JUPYTERHUB_SERVICE_PREFIX - the URL prefix the service will run on (e.g. `/user/name/`)
-- JUPYTERHUB_USER - the JupyterHub user's username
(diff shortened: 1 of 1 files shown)


docs/source/reference/spawners.md
```diff
@@ -297,8 +297,8 @@ Additional variables can be specified via the `Spawner.environment` configuratio
 
 The process environment is returned by `Spawner.get_env`, which specifies the following environment variables:
 
-- JUPYTERHUB*SERVICE_URL - the \_bind* URL where the server should launch its HTTP server (`http://127.0.0.1:12345`).
-  This includes `Spawner.ip` and `Spawner.port`; _new in 2.0, prior to 2.0 IP, port were on the command-line and only if specified_
+- JUPYTERHUB*SERVICE_URL - the \_bind\* URL where the server should launch its HTTP server (`http://127.0.0.1:12345`).
+  This includes `Spawner.ip` and `Spawner.port`; \_new in 2.0, prior to 2.0 IP, port were on the command-line and only if specified*
```
minrk, line 301: Still [renders a bit off](https://jupyterhub--4245.org.readthedocs.build/en/4245/reference/spawners.html#environment-variables-and-command-line-arguments). Maybe autoformatter's getting something mixed up with `_/*`?

```suggestion
- JUPYTERHUB_SERVICE_URL - the *bind* URL where the server should launch its HTTP server (`http://127.0.0.1:12345`).
  This includes `Spawner.ip` and `Spawner.port`; new in 2.0, prior to 2.0 IP, port were on the command-line and only if specified.
```

minrk:
`*` seems to have been injected into JUPYTERHUB_SSL below, too.

manics:
It's a bug in prettier so none of the above workarounds work: https://github.com/prettier/prettier/issues/12965

How do you feel about the use of `` `backticks` `` to format the env-vars as code? That's what I normally use in my own docs.

consideRatio:
> How do you feel about the use of `backticks` to format the env-vars as code?

:+1:!

minrk:
I somewhat prefer not using code formatting for env variables since they are already offset plenty with all caps, but if the autoformatter is confused by the interior underscores, it seems like the right fix. Go for it!

consideRatio (approved):

PR #4251, opened 2022-12-07
Test docs

minrk:
- Runs sphinx build with error-on-warn so we notice problems in the sphinx build. RTD (rightly) doesn't stop on warnings, but we still want to notice. It would be nice if we could get RTD to report these errors as CI failures while still completing and publishing the build. But running them in an action is easy enough.
- runs linkcheck to keep our links fresh (and fixes several broken links)

includes #4249 to avoid conflicts


minrk:
The failure is showing that it caught the problem fixed by #4250

minrk:
rebased

consideRatio (approved):

PR #4256, opened 2022-12-07
docs: refresh Makefile/make.bat

consideRatio:
Working with sphinx built doc, I love the use of sphinx-autobuild, but have never learned how to invoke it directly but instead relied on a makefile target like the `devenv` target I'm adding in this PR.

These Makefile/make.bat files have been generated by spinx-quickstart, and the reason we can cut down so significantly on the size of this file - without loosing functionality, is because of a new "catch all" target that the modern sphinx-quickstart generates for the Makefile and make.bat files.

I see significant value in having the complexity of these files reduced to help anyone inspecting them understand what goes on better.


minrk:
Sure! I've ~never looked in these files, but happy to have them refreshed.

PR #4259, opened 2022-12-09
set max depth on api/index toctree

### LINKED ISSUE #4254: docs: 3.1.0 vs 3.0.0 has different layout in api/index.html
I'm not sure what caused this, but the index looks better like in 3.0.0 I think.

3.0.0|3.1.0
-|-
image|image

minrk:
default maxdepth seems to have changed?

closes #4254


consideRatio:
Looks great now: 
image

PR #4271, opened 2022-12-14
only run testing config on localhost

minrk:
avoids listening on the network with dummy auth

### DIFF
--- a/testing/jupyterhub_config.py
+++ b/testing/jupyterhub_config.py
@@ -16,3 +16,6 @@ c.JupyterHub.authenticator_class = DummyAuthenticator
 from jupyterhub.spawner import SimpleLocalProcessSpawner
 
 c.JupyterHub.spawner_class = SimpleLocalProcessSpawner
+
+# only listen on localhost for testing
+c.JupyterHub.bind_url = 'http://127.0.0.1:8000'


testing/jupyterhub_config.py
```diff
@@ -16,3 +16,6 @@
 from jupyterhub.spawner import SimpleLocalProcessSpawner
 
 c.JupyterHub.spawner_class = SimpleLocalProcessSpawner
+
+# only listen on localhost for testing
+c.JupyterHub.bind_url = 'http://127.0.0.1:8000'
```
manics, line 21:
```suggestion
c.JupyterHub.bind_url = 'http://localhost:8000'
```
I don't know if this works with ipv6... but if it does then `localhost` should be more robust.

minrk: On my machine (mac), changing this results in listening _only_ on IPv6 (connecting on 127 doesn't work). I'm not sure if that's better?

When I try to curl localhost, it tries and fails IPv4 first, then IPv6, so if something else is listening on 127, that will get the requests for localhost instead.

```
$ curl -v http://localhost:8000
*   Trying 127.0.0.1:8000...
* connect to 127.0.0.1 port 8000 failed: Connection refused
*   Trying [::1]:8000...
* Connected to localhost (::1) port 8000 (#0)
```

manics: Hmmm.... you decide then! Merge if you think it's fine as is.

minrk: I don't know! I'll go with 127 for now, since it seems to result in simpler behavior. This is only for local testing, anyway.

Issue #4270, opened 2022-12-14
Accessibility of the 404 page


bl-aire:
### Details from report:

Based on evaluation using the WAVE accessibility tool, the following are responsible for accessibility issues on the 404 page:

The errors on this page are the same with those on the previous pages evaluated(Home, Login and Token) and can be fixed with the same solutions discussed earlier:

- **Language missing or invalid**
- **Skipped heading level**
- **Possible heading**
- **Noscript element**

### Report:

<img width="1280" alt="Screenshot 2022-12-14 at 11 59 47" src="https://user-images.githubusercontent.com/84882370/207596167-2ca5e07b-81f7-43ab-8dca-e52d9e771812.png">

@minrk

minrk:
Great! Do you want to make a PR for the general issues (language, noscript) before evaluating more pages?

I don't quite know what the issue is with noscript.

bl-aire:
> Great! Do you want to make a PR for the general issues (language, noscript) before evaluating more pages?

Yes. That would prevent unnecessary repetition in every issue.

> I don't quite know what the issue is with noscript.

I have been trying to understand it too. This is what I saw on WAVE:

> **Why It Matters**
Content within <noscript> is presented if JavaScript is disabled. Because nearly all users (including users of screen readers and other assistive technologies) have JavaScript enabled, <noscript> cannot be used to provide an accessible version of inaccessible scripted content.
**How to Fix It**
Ensure that scripted content is accessible. The <noscript> content will be presented to very few users, but must be accessible if used.

minrk:
We don't use noscript to present anything about accessibility - it's a message that the pages mostly won't work, so I don't really understand the comments. It seems irrelevant.

tpu01yzx:
I have the same problem. After login and redirect to the spamed jupyter_server page it shows 404 with URL _http://XX.XX.XX:280/user/yzx/tree?_.



> 404 : Not Found
> 
> You are requesting a page that does not exist!

The config is

> c = get_config()  #noqa
> 
> import sys
> 
> c.Authenticator.admin_users = {'root', 'yzx'}  # 管理员用户
> c.JupyterHub.admin_access = True
> c.PAMAuthenticator.open_sessions = False # 解决多用户同时登录问题。
> c.Spawner.args = ['--allow-root']  # 允许root用户使用
> c.LocalAuthenticator.create_system_users = True  # 允许创建其他用户
[...]

Issue #4265, opened 2022-12-12
Accessibility of the Token page


bl-aire:
### Details from report:

Based on evaluation using the WAVE accessibility tool, the following are responsible for accessibility issues on the Token page:

- **Language missing or invalid, Missing first level heading, Noscript element, Very low contrast, Aria button**: These are the same with the previous issues and can be fixed the same way.

- **Skipped heading level**: The current hierarchy of header elements is **_h2 --> h4._** This is what caused this issue. Normally, the hierarchy should go from the highest to the lowest so that.....After a `h1` element is added to the page to fix the "missing first level heading" error, the modal title(which is currently a `h4` element) can be changed to a `h2` and resized. So the new hierarchy will be **_h1 --> h2_.**

- **Layout table present**: 

> Layout tables are used to position content visually - to create columns, insert spacing, or align content neatly for sighted users. They should not be used in HTML5 as they introduce reading and navigation order issues. Screen readers may interpret them as data tables (i.e., announcing column and row numbers), especially if they contain table header `<th>` cells. This introduces significant overhead on screen reader users.

To achieve desired visual presentation of layout tables, flexbox or grid can be used. But if the table contains tabular data(_which I think is the case here_), appropriate header `<th>` cells should be provided. The current table does not have `<th>`.

- **Possible heading**: 

### Report:

<img width="1280" alt="Screenshot 2022-12-10 at 12 26 53" src="https://user-images.githubusercontent.com/84882370/207054729-ea81935a-d0c8-4ae9-b1c4-e6925a9d4022.png">

<img width="1280" alt="Screenshot 2022-12-10 at 12 27 29" src="https://user-images.githubusercontent.com/84882370/207054764-590710dd-cf4d-4511-b318-36cb65dc3522.png">

@minrk

bl-aire:
Hi @minrk, please I'd like to know how I can add the accessibility label to my issues. I tried to but I can't. The settings icon is not showing beside 'Labels' in my issues.

minrk:
Yes, th elements are definitely appropriate.

Let's skip the white-on-orange button issue across the board for now.

I'm not certain about the h1, whether it's a good idea to add a title to these pages, or make a hidden h1. There's been some discussion about this previously, specifically in the context of these pages: https://github.com/jupyter/accessibility/discussions/88

minrk:
> I'd like to know how I can add the accessibility label to my issues

[...]

Issue #4262, opened 2022-12-10
Accessibility of the Home page


bl-aire:
### Details from report:

Based on evaluation using the WAVE accessibility tool, the following are responsible for accessibility issues on the Login page:

- **Language missing or invalid**, **Missing first level heading** and **Noscript element**: These are accessibility issues that are common between the login page and home page #4260 so they can be fixed the same way.

- **Very low contrast**: The contrast between the navigation items (Home, Token and username) and their container's background colour is too low so it won't be legible for people with low vision or colour deficiencies except the text's shade is made darker.

- **Aria button**: This means element (other than a button) has `role="button"`. Elements with `role="button"` must function like native buttons. These should typically be replaced with true button elements or one should ensure the element is navigable, can be activated with Enter or Space key presses (key detection is typically necessary), and provides a visible focus indicator.

### Report:

<img width="1280" alt="Screenshot 2022-12-09 at 14 36 02" src="https://user-images.githubusercontent.com/84882370/206853307-51245a46-ddb6-4dff-8bfe-e89bfcb0aee8.png">

@minrk

minrk:
Let's change the link color to `#222`, which should be high enough contrast. Let me know if you have any questions about finding that in the LESS.

Let's try making our button links button _elements_, and see what happens.

bl-aire:
Okay. 
I wil do that

bl-aire:
Hello @minrk ,

I have a question regarding the pages I am evaluating. I have been able to evaluate pages like:
- Home
- Login
- Token

But I haven't figured out how to evaluate templates like:

- **Spawn_pending**: this displays when the **Start my server** is clicked but stays for a couple of seconds so I haven't been able to access it with wave.
- **Oath** , **not_running**: the former should display when a third-party application wants access to data associated with a JupyterHub account while the latter should display when attempts to start the server fails but I don't know how to recreat these yet.
- **Logout**: this is not displaying on my instance when I logout.


How can I have these pages display on my local instance of JupyterHub?

I considered creating an issue for questions related to the project so they can be tracked easily. If that's okay, I can create it when you approve.

minrk:
Spawn pending can be tested more easily by running a custom spawner that delays start:

```python
import asyncio
from jupyterhub.spawner import SimpleLocalProcessSpawner

class SlowSpawner(SimpleLocalProcessSpawner):
    async def start(self):
        await asyncio.sleep(10)
        return await super().start()

[...]

Issue #4260, opened 2022-12-09
Accessibility of the Login page


bl-aire:
### Details from report:
Based on evaluation using the WAVE accessibility tool, the following are responsible for accessibility issues on the Login page:

- **Language missing or invalid**: Currently, the language attribute `lang="en"` which helps search engines switch language profiles so they can provide correct accent/pronunciation is missing in the opening html tag `<html>` .

- **Very low contrast**: The contrast of the form's header background colour, its button background colour and their text is too low. The same applies to the error message displayed when a user's password or username are not correct. The background colours of affected elements can be made darker to fix this so their text is legible for anyone with low vision or colour deficiencies.

- **Missing first level header**: This accessibility issue is as a result of the page having no `h1` element. 
The form's header was created with this code `<div>Sign in</div>`. I would propose nesting a `h1` in the `div` and reducing the font-size and possibly the font-weight of the header to fix it.

- **Positive tabindex values present**: 
Tabindex values of 1 or greater places the element the tabindex is applied first in the keyboard tabbing order. Thus, creating a confusing experience for anyone who relies on the keyboard’s Tab key to navigate.
Interactive elements come with native browser behavior already built in that allows them to be navigated to via Tab without the need for a tabindex declaration. In this case, the elements are in the reading order of the document so the Tabindex is not necessary.

- **Noscript element**

**Suggestion:** The alt text of the logo is "jupyterhub". I am suggesting that it is renamed to "jupyterhub logo".  This would give users an idea of what the image's purpose is.

### Report:

<img width="1280" alt="Screenshot 2022-12-08 at 13 43 16" src="https://user-images.githubusercontent.com/84882370/206695138-a0534019-306c-451e-9cdd-ec856243628d.png">

Hello @minrk,
As discussed, I have evaluated and given a summary of the login page. I am open to a discussion on how best they can be fixed.

minrk:
> Currently, the language attribute lang="en" 

Yes, this would be in `page.html` and fix every page. We should make sure it can be overridden, in case deployments localize their pages in other languages. We can do this with a template variable, I think.

>  The alt text of the logo is "jupyterhub". I am suggesting that it is renamed to "jupyterhub logo"

👍 

For the h1 question (which will come up for all pages), I asked the accessibility team about that a while back https://github.com/jupyter/accessibility/discussions/88. Their recommendations there will be useful. I think making the `Sign in` heading an h1, and then adding CSS to keep it a sensible size makes sense.

[...]

PR #4274, opened 2022-12-16
Fix reoccurring accessibility issues in JupyterHub's pages

bl-aire:
After reviewing the accessibility of some of JupyterHub's pages, a few accessibility issues have been reoccurring. They can be seen here:

- Login page: #4260

- Home page: #4262

- Token page: #4265

- 404 page: #4270

With this PR, solutions to a few of them like **Lang missing or invalid** and **Missing first level heading** which were discussed earlier in the issues are implemented. Also the logo's **Alt text** is renamed from _jupyterhub_ to _jupyterhub logo_ so it clearly states the purpose of the image to people who rely on screen readers.

### Report: 

The report below shows the Login page with the changes above implemented

Screenshot 2022-12-16 at 14 17 01

@minrk

### DIFF
--- a/share/jupyterhub/static/less/login.less
+++ b/share/jupyterhub/static/less/login.less
@@ -19,7 +19,6 @@
     vertical-align: middle;
     margin: auto auto 20% auto;
     width: 350px;
-    font-size: large;
   }
 
   .input-group,
@@ -50,11 +49,17 @@
     color: #fff;
     background: @jupyter-orange;
     border-radius: @border-radius-large @border-radius-large 0 0;
+    font-size: large;
+  }
+
+  .auth-form-header > h1 {
+    font-size: inherit;
+    font-weight: inherit;
+    margin: 0;
   }
 
   .auth-form-body {
     padding: 20px;
-    font-size: 14px;
     border: thin silver solid;
     border-top: none;
     border-radius: 0 0 @border-radius-large @border-radius-large;
--- a/share/jupyterhub/static/less/variables.less
+++ b/share/jupyterhub/static/less/variables.less
@@ -4,7 +4,14 @@
 @navbar-height: 40px;
 @grid-float-breakpoint: @screen-xs-min;
 
+@navbar-default-color: #222;
+@navbar-default-link-color: @navbar-default-color;
+// darken background on hover, no change to text
+@navbar-default-link-hover-color: @navbar-default-color;
+@navbar-default-link-hover-bg: darken(@navbar-default-bg, 10%);
+
 @jupyter-orange: #f37524;
+
 @jupyter-red: #e34f21;
 // color blind-friendly alternative to red/green
 // from 5-class RdYlBu via colorbrewer.org
--- a/share/jupyterhub/templates/home.html
+++ b/share/jupyterhub/templates/home.html
@@ -5,6 +5,7 @@
 
 {% block main %}
 <div class="container">
+  <h1 class="sr-only">JupyterHub home page</h1>
   <div class="row">
     <div class="text-center">
       {% if default_server.active %}
--- a/share/jupyterhub/templates/login.html
+++ b/share/jupyterhub/templates/login.html
@@ -27,7 +27,7 @@
 {% else %}
 <form action="{{authenticator_login_url}}" method="post" role="form">
   <div class="auth-form-header">
-    Sign in
+    <h1>Sign in</h1>
   </div>
   <div class='auth-form-body'>
 
--- a/share/jupyterhub/templates/page.html
+++ b/share/jupyterhub/templates/page.html
@@ -22,7 +22,7 @@
 
 
 <!DOCTYPE HTML>
-<html>
+<html lang="en">
 
 <head>
     <meta charset="utf-8">
@@ -103,7 +103,7 @@
       <div class="navbar-header">
         {% block logo %}
         <span id="jupyterhub-logo" class="pull-left">
-            <a href="{{logo_url or base_url}}"><img src='{{base_url}}logo' alt='JupyterHub' class='jpy-logo' title='Home'/></a>
+            <a href="{{logo_url or base_url}}"><img src='{{base_url}}logo' alt='JupyterHub logo' class='jpy-logo' title='Home'/></a>
         </span>
         {% endblock %}
         {% if user %}
--- a/share/jupyterhub/templates/token.html
+++ b/share/jupyterhub/templates/token.html
@@ -3,6 +3,7 @@
 {% block main %}
 
 <div class="container">
+  <h1 class="sr-only">Manage JupyterHub Tokens</h1>
   <div class="row">
     <form id="request-token-form" class="col-md-offset-3 col-md-6">
       <div class="text-center">


share/jupyterhub/templates/login.html
```diff
@@ -10,6 +10,7 @@
 
 {% block login %}
 <div id="login-main" class="container">
+  <h1 class="sr-only">Login page</h1>
```
minrk, line 13: I _think_ this h1 sould replace the current `div.auth-form-header` and visible, rather than hidden. Then we can make the `auth-form-header` style applied to the h1 instead of the div, but the same element.

share/jupyterhub/static/less/page.less
```diff
@@ -44,6 +44,12 @@
   padding: 8px;
 }
 
+/* #thenavbar{
+  ul li a {
+    color: #222;
+  }
+}*/
```
minrk, line 51: empty css, commented-out? Maybe there's a bootstrap LESS variable we should be setting to control the link color, rather than these specific navbar elements.

share/jupyterhub/templates/home.html
```diff
@@ -5,6 +5,7 @@
 
 {% block main %}
 <div class="container">
+  <h1 class="sr-only">Home page</h1>
```
minrk, line 8: Probably: `JupyterHub home page`

share/jupyterhub/templates/page.html
```diff
@@ -103,7 +103,7 @@ <h4 class="modal-title" id="{{key}}-label">{{title}}</h4>
       <div class="navbar-header">
         {% block logo %}
         <span id="jupyterhub-logo" class="pull-left">
-            <a href="{{logo_url or base_url}}"><img src='{{base_url}}logo' alt='JupyterHub' class='jpy-logo' title='Home'/></a>
+            <a href="{{logo_url or base_url}}"><img src='{{base_url}}logo' alt='JupyterHub logo' class='jpy-logo' title='Home'/></a>
```
minrk, line 106: 👍

share/jupyterhub/templates/token.html
```diff
@@ -3,6 +3,7 @@
 {% block main %}
 
 <div class="container">
+  <h1 class="sr-only">Token page</h1>
```
minrk, line 6: Rather than `Token Page`, let's do something more descriptive like `Manage JupyterHub Tokens`.

minrk (commented):
Great! I left some suggestions to move this forward.

share/jupyterhub/templates/token.html
bl-aire: Okay. I will rename the h1 tags now

share/jupyterhub/templates/home.html
bl-aire: Noted

share/jupyterhub/templates/login.html
bl-aire: > I _think_ this h1 sould replace the current `div.auth-form-header` and visible, rather than hidden. Then we can make the `auth-form-header` style applied to the h1 instead of the div, but the same element.

Okay. That's better. I will make changes

share/jupyterhub/static/less/page.less
bl-aire: **File:**....share/jupyterhub/static/components/bootstrap/less/variables.less**

Changing the `@navbar-default-link-color:` and `@navbar-default-link-hover-color: ` in variables.less fixed this.

**Report:**
Screenshot 2022-12-16 at 17 42 33

minrk: Good find! Looks like this hasn't been added yet, though. Still working on a commit?

bl-aire: Yes, I am. On running `git status`, I get the prompt `Your branch is up to date with origin/main`.

I am inferring this is because share/jupyterhub/static/components is in the .gitignore file. Please what can I do?

share/jupyterhub/templates/login.html
```diff
@@ -10,6 +10,7 @@
 
 {% block login %}
 <div id="login-main" class="container">
+  <!--<h1 class="sr-only">Login page</h1>-->
```
minrk, line 13: Let's not leave commented-out HTML in here

```suggestion
```

share/jupyterhub/static/less/login.less
```diff
@@ -52,6 +52,11 @@
     border-radius: @border-radius-large @border-radius-large 0 0;
   }
 
+  h1 {
+    font-size: 1.03em;
```
minrk, line 56: 1.03? Not 1?

Maybe it should be `.auth-form-header > h1` to keep it specific.

share/jupyterhub/templates/login.html
bl-aire: Okay.

share/jupyterhub/static/less/login.less
bl-aire: 1.03 was the closest to the previous size. But I can make it 1em, if that's better

Okay

share/jupyterhub/static/less/page.less
minrk: Bootstrap variables should be overridden in `share/jupyterhub/static/less/variables.less`. Files in components should not be edited.

share/jupyterhub/static/less/login.less
minrk: Left the comment [earlier](https://github.com/jupyterhub/jupyterhub/pull/4274#discussion_r1053305552), but it should be in `share/jupyterhub/static/less/variables.less`

share/jupyterhub/static/less/page.less
bl-aire: > Bootstrap variables should be overridden in `share/jupyterhub/static/less/variables.less`. Files in components should not be edited.

Okay. Thanks

share/jupyterhub/static/less/variables.less
```diff
@@ -4,7 +4,11 @@
 @navbar-height: 40px;
 @grid-float-breakpoint: @screen-xs-min;
 
+@navbar-default-link-color: #222;
+@navbar-default-link-hover-color: #777;
```
minrk, line 8: I think lightening the text this much on hover makes it pretty low contrast. Instead, let's try darkening the background a bit and leaving the text the same:

```suggestion
@navbar-default-color: #222;
@navbar-default-link-color: @navbar-default-color;
// darken background on hover, no change to text
@navbar-default-link-hover-color: @navbar-default-color;
@navbar-default-link-hover-bg: darken(@navbar-default-bg, 10%);
```

minrk (commented):
Great! I left two small suggestions in how to implement some of the styling, but otherwise looking good.

share/jupyterhub/static/less/login.less
minrk: I don't think any changes to the padding are needed. I think you might be getting misled by WAVE increasing the size by inserting its warning markers. The padding is already correct at `10px 20px` and I think it shouldn't change here. This is what it looks like in the main branch:

Screenshot 2022-12-21 at 14 04 18

vs increasing padding to 20px:


Screenshot 2022-12-21 at 14 06 21

bl-aire: Okay. I will stick to 10px then

minrk (approved):
Great! I squashed the history since this is a small change and there were lots of commits, but all the changes look good!

bl-aire:
> Great! I squashed the history since this is a small change and there were lots of commits, but all the changes look good!

Yaaaayyy! 💯🎉

I have a question..For subsequent PRs, would you advise I keep making commits of little changes as I work or group them? For example, having one commit per issue or page

bl-aire:
> It's always fine to keep adding commits during the PR review process (that makes changes easier to review than constant rebasing, especially if the changes aren't small). I only squash history like this if the commits have gone significantly out of proportion to the changes, as they did here. It's more a part of the merge process than part of the contribution process.
> 
> 
> 
> If the changes are truly unrelated, they should usually be made in separate pull requests. But one commit per discrete change is fine, as is a bigger commit of _related_ changes (I wouldn't object either way on a PR like this).

Okay. It's clearer for me now. Thank you

PR #4288, opened 2023-01-04
added note on `Spawner.name_template` setting

stevejpurves:
Addresses issue jupyterhub/dockerspawner#475

Adds a note to the docs to mention setting the `name_template` on the spawner to include the server name when named servers are enabled.

In my TLJH deployment this was necessary to be able to launch more than 1 server. I'm assuming this is a necessary step?

I arrived at this fix based on this discussion here: https://discourse.jupyter.org/t/multiple-jupyterhub-servers-per-user-named-servers-does-not-work/2357/5

### DIFF
--- a/docs/source/reference/config-user-env.md
+++ b/docs/source/reference/config-user-env.md
@@ -202,7 +202,7 @@ This can be useful for quota service implementations. The example above limits t
 
 If `named_server_limit_per_user` is set to `0`, no limit is enforced.
 
-When using a spawner such as `DockerSpawner`, it is necessary to customize the name template to include the server name. Failing to do so can result in server errors when attempting to launch subsequent servers. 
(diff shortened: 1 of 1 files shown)


minrk:
Thanks! I think it's appropriate to have a _general_ note about extra config _possibly_ needed for named servers. I wouldn't get _too_ specific about a particular issue, though (e.g. the specific referred-to behavior should be considered a bug in DockerSpawner and ought to be fixed, which would make the JupyterHub docs outdated, independent of the JupyterHub version). But mentioning that Spawners may need specific configuration to take the server name into account is definitely appropriate, and users shoudl be directed to their specific Spawner's docs, such as DockerSpawner's name_template (which doesn't work by default), or KubeSpawner's [pod_name_template](https://jupyterhub-kubespawner.readthedocs.io/en/latest/spawner.html#kubespawner.KubeSpawner.pod_name_template) (which does).

minrk:
Thanks!

Issue #1932, opened 2018-05-30
Support HA (High Availability)


elgalu:
What would it take for the Hub, and the Proxy to support HA?

i.e. to have more than 1 pod (multi-master), ideally 3 or more so in case of some AWS EC2 instances going down the service survives?

minrk:
There are two main obstacles to HA:

1. The proxy is a single-point-of failure. A proxy implementation that is itself HA should be used instead of the default. kubespawner [provides](https://github.com/jupyterhub/kubespawner/pull/64) an experimental implementation based on kubernetes ingress. [Implementing a proxy](https://jupyterhub.readthedocs.io/en/stable/reference/proxy.html) using other technology should be doable by writing a few Python functions (assuming you already know how to implement updating your chosen proxy).
2. The Hub itself is a single-point-of-failure and **must own the database** it talks to. Other Hub instances may not be running and talking to the same database, or there will be errors and inconsistent state. So you cannot easily have two Hub replicas and failover between them.

The way services have achieved this today is to run multiple *wholly independent* JupyterHub instances (including a proxy for each), and maintaining sessions such that users are always routed to the same Hub (so that the Hub databases are independent), and failing over to new Hubs when the Hub a user is associated with has gone.

If, however, your interpretation of HA allows brief downtime when a node disappears, using a technology like Kubernetes allows the Hub (and other processes) to automatically be relocated to a new node when the previous node goes away. This isn't HA *per se* in that there is nonzero downtime, but the downtime is on the scale of the time it takes for the node loss to be noticed and for the new Hub pod to start (typically < 1-2 minutes), and no human intervention is required.

clkao:
@minrk can you elaborate how the session-aware meta proxy is achieved? and can it also be done with an ingress-based proxy implementation similar to the current experimental one?

minrk:
The end-goal is that a given user, once assigned to a Hub, is always routed to that Hub. Typically, this is accomplished by setting an affinity cookie that is checked in the edge proxy (not the hub's proxy, but a layer before that) before routing. The data-8x implementation is largely [here](https://github.com/berkeley-dsep-infra/data8xhub/blob/d0f9ff322039625cc0e63a6db3a369bc94656fb6/outer-edge/templates/configmap.yaml)

> can it also be done with an ingress-based proxy implementation similar to the current experimental one?

[...]

PR #4292, opened 2023-01-09
expand database docs

minrk:
add notes on what's in the database, why there's a database, and how it relates to performance

### DIFF
--- a/docs/source/reference/database.md
+++ b/docs/source/reference/database.md
@@ -14,7 +14,7 @@ A lot of the operations in JupyterHub are also **relationships**, which is exact
 For example:
 
 - Given an API token, what user is making the request?
-- Which users who don't have running servers?
+- Which users don't have running servers?
 - Which servers belong to user X?
 - Which users have not been active in the last 24 hours?
 
@@ -52,7 +52,7 @@ Database operations are also synchronous, so while JupyterHub is waiting on a da
 This allows us to avoid complex locking mechanisms, because transaction races can only occur during an `await`, so we only need to make sure we've completed any given transaction before the next `await` in a given request.
 
 :::{note}
-We are slowing working to remove these assumptions, and moving to a more traditional db session per-request pattern.
+We are slowly working to remove these assumptions, and moving to a more traditional db session per-request pattern.
 This will enable multiple Hub instances and enable scaling JupyterHub, but will significantly reduce the number of active users a single Hub instance can serve.
 :::
 
--- a/docs/source/reference/database.md
+++ b/docs/source/reference/database.md
@@ -80,7 +80,7 @@ Additionally, there is usually _very_ little load on the database itself.
 By far the most taxing activity on the database is the 'list all users' endpoint, primarily used by the [idle-culling service](https://github.com/jupyterhub/jupyterhub-idle-culler).
 Database-based optimizations have been added to make even these operations feasible for large numbers of users:
 
-1. State filtering on GET /users with [?state=active](./rest-api.html#/default/get_users),
+1. State filtering on [GET /users](./rest-api.md) with `?state=active`,
    which limits the number of results in the query to only the relevant subset (added in JupyterHub 1.3), rather than all users.
 2. [Pagination](api-pagination) of all list endpoints, allowing the request of a large number of resources to be more fairly balanced with other Hub activities across multiple requests (added in 2.0).
 
--- a/docs/source/reference/database.md
+++ b/docs/source/reference/database.md
@@ -1,29 +1,120 @@
 # The Hub's Database
 
-JupyterHub uses a database to store information about users, services, and other
-data needed for operating the Hub.
+JupyterHub uses a database to store information about users, services, and other data needed for operating the Hub.
+This is the **state** of the Hub.
 
-## Default SQLite database
+## Why does JupyterHub have a database?
 
-The default database for JupyterHub is a [SQLite](https://sqlite.org) database.
-We have chosen SQLite as JupyterHub's default for its lightweight simplicity
-in certain uses such as testing, small deployments and workshops.
+JupyterHub is a **stateful** application (more on that 'state' later).
+Updating JupyterHub's configuration or upgrading the version of JupyterHub requires restarting the JupyterHub process to apply the changes.
+We want to minimize the disruption caused by restarting the Hub process, so it can be a mundane, frequent, routine activity.
+Storing state information outside the process for later retrieval is necessary for this, and one of the main thing databases are for.
+
+A lot of the operations in JupyterHub are also **relationships**, which is exactly what SQL databases are great at.
+For example:
+
+- Given an API token, what user is making the request?
+- Which users who don't have running servers?
+- Which servers belong to user X?
+- Which users have not been active in the last 24 hours?
+
+Finally, a database allows us to have more information stored without needing it all loaded in memory,
+e.g. supporting a large number (several thousands) of inactive users.
+
+## What's in the database?
+
+The short answer of what's in the JupyterHub database is "everything."
+JupyterHub's **state** lives in the database.
+That is, everything JupyterHub needs to be aware of to function that _doesn't_ come from the configuration files, such as
+
+- users, roles, role assignments
+- state, urls of running servers
+- Hashed API tokens
+- Short-lived state related to OAuth flow
+- Timestamps for when users, tokens, and servers were last used
+
+### What's _not_ in the database
+
+Not _quite_ all of JupyterHub's state is in the database.
+This mostly involves transient state, such as the 'pending' transitions of Spawners (starting, stopping, etc.).
+Anything not in the database must be reconstructed on Hub restart, and the only sources of information to do that are the database and JupyterHub configuration file(s).
+
+## How does JupyterHub use the database?
+
+JupyterHub makes some _unusual_ choices in how it connects to the database.
+These choices represent trade-offs favoring single-process simplicity and performance at the expense of horizontal scalability (multiple Hub instances).
+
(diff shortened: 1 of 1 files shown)


docs/source/reference/database.md
```diff
@@ -1,29 +1,120 @@
...
+Anything not in the database must be reconstructed on Hub restart, and the only sources of information to do that are the database and JupyterHub configuration file(s).
+
+## How does JupyterHub use the database?
+
+JupyterHub makes some _unusual_ choices in how it connects to the database.
+These choices represent trade-offs favoring single-process simplicity and performance at the expense of horizontal scalability (multiple Hub instances).
+
+We often say that the Hub 'owns' the database.
+This ownership means that we assume the Hub is the only process that will talk to the database.
+This assumption enables us to make several caching optimizations that dramatically improve JupyterHub's performance (i.e. data written recently to the database can be read from memory instead of fetched again from the database) that would not work if multiple processes could be interacting with the database at the same time.
```
consideRatio, line 49: No action point or issue seems relevant to create as a followup, just felt like sharing this brief thought!

This got me thinking that maybe a way to incrementally work towards supporting multiple replicas of a JupyterHub is to make such information requests flip from being read in memory to being read from the database or similar.

docs/source/reference/database.md
```diff
@@ -1,29 +1,120 @@
...
+
+Database operations are also synchronous, so while JupyterHub is waiting on a database operation, it cannot respond to other requests.
+This allows us to avoid complex locking mechanisms, because transaction races can only occur during an `await`, so we only need to make sure we've completed any given transaction before the next `await` in a given request.
```
consideRatio, line 52: (This comment requires no action)

Aaaaaaah I'm so happy I can understand this! Excellent technical input!!!

My understanding is that JupyterHub executes code just at one place at the time, and switches only when running `await` somewhere, but since database calls are not `await`-ed this is not a problem as long as all transactional logic that needs to be done against the database is done without an `await` in between.

docs/source/reference/database.md
```diff
@@ -1,29 +1,120 @@
...
+The default database backend for JupyterHub is [SQLite](https://sqlite.org).
+We have chosen SQLite as JupyterHub's default because it's simple (the 'database' is a single file) and ubiquitous (it is in the Python standard library).
+It works very well for testing, small deployments, and workshops.
 
 For production systems, SQLite has some disadvantages when used with JupyterHub:
 
-- `upgrade-db` may not work, and you may need to start with a fresh database
+- `upgrade-db` may not always work, and you may need to start with a fresh database
 - `downgrade-db` **will not** work if you want to rollback to an earlier
   version, so backup the `jupyterhub.sqlite` file before upgrading
```
consideRatio, line 110: Hmmm wait what does downgrade-db do? To me it sounds like upgrade-db / downgrade-db is about upgrading the database schema based on an associated jupyterhub version or similar.

consideRatio (approved):
Brilliant, this was very helpful reading for me!

I'm not merging as there is a pending review comment about linking to the dynamically created documentation section and since its a very recently opened PR.

docs/source/reference/database.md
minrk: Yeah, cooperative multi-tasking (coroutines) are actually a _huge_ help in terms of the complexity of race conditions. Races can only _possibly_ happen on lines with an `await`, so as long as you've set whatever lock you need (e.g. the Spawner.pending attribute) before the next await, you're all set. This was still plenty hard enough! It was a _big_ win in complexity reduction for the maintenance of the project early on.

I don't relish making concurrent requests to launch the same server on multiple Hubs robust when there can be two or more fully concurrent Hub instances.

docs/source/reference/database.md
minrk: It's entirely possible this comment was outdated and should be tested and/or removed. IIRC, weirdly sqlite can _add_ data to the schema, but it couldn't remove it. So adding things almost always works with sqlite (upgrade-db), while removing things almost never did (downgrade-db).

docs/source/reference/database.md
minrk: Yes, I've outlined the steps involved [here](https://github.com/jupyterhub/jupyterhub/issues/1932#issuecomment-658031232) and [here](https://figshare.com/articles/poster/The_cloud_is_the_limit_scaling_limits_of_JupyterHub/14937954), but it's roughly:

1. take away direct db access from Authenticator and Spawner (already deprecated, _almost_ never used)
2. remove the in-memory Users dict
3. open a db session per-request instead of using a shared session

These are all that's needed for the db session. It will have a pretty large performance cost for idle-culling, because listing all 'active' users is an iteration over a Python dict of already-constructed objects, rather than recreating a bunch of ORM objects from a db query, which it will _have_ to be going forward. That's orders of magnitude faster. _But_ it's also a scaling issue because all 'active' users must fit in memory.

The main challenge is that User objects are ORM wrappers and Spawners have access to these, and Spawners are long-lived objects. Providing them with a read-only static snapshot of the User data ought to be fine for almost everyone, but it will need to be a breaking API change. Cases that will be hard are e.g. checking user group membership, which some folks do now in hooks. That will mean not only fetching the user, but fetching all their _relationships_ that we want to support Spawners _possibly_ accessing, rather than what we have now, which is allowing them to request anything as-needed. It's probably better in the long run! But it will be a hard transition. And if we don't enable multiple replicas in the same release, it will be a purely _worse_ JupyterHub as a result.

That's largely why I haven't started work on it - most of the incremental changes, while possible, are _detrimental_ to JupyterHub until they reach the final goal of enabling replicas. So I don't really want to start it until I have the support needed to finish it.

docs/source/reference/database.md
```diff
@@ -1,29 +1,120 @@
...
+JupyterHub is a **stateful** application (more on that 'state' later).
+Updating JupyterHub's configuration or upgrading the version of JupyterHub requires restarting the JupyterHub process to apply the changes.
+We want to minimize the disruption caused by restarting the Hub process, so it can be a mundane, frequent, routine activity.
+Storing state information outside the process for later retrieval is necessary for this, and one of the main thing databases are for.
+
+A lot of the operations in JupyterHub are also **relationships**, which is exactly what SQL databases are great at.
+For example:
+
+- Given an API token, what user is making the request?
+- Which users who don't have running servers?
```
GeorgianaElena, line 17: Or maybe something is missing?

```suggestion
- Which users don't have running servers?
```

docs/source/reference/database.md
GeorgianaElena: I find this kind of information about _current_ plans about the project _future_ very useful for creating a mental picture about the general direction of where the work should be put into and how to make sustainable technical choices and that don't harm the general plan. But currently, as it is, as GitHub issues, this info is pretty hard to find when searching for them intentionally unless, you already know some keywords and have some prior context.

So a question popped into my head when reading this, that I feel was discussed before (?) is "Where should we store such project plan in a place that is findable, iterable and easy to update when things change?" Is it discourse, the documentation itself, or GitHub issues / projects?

(cc @sgibson91 as having such place, I believe might be of help for bringing in, getting feedback from and keeping new contributors in the community)

GeorgianaElena (approved):
I'm also very helpful for the information in this PR.
This, together with the info  in https://jupyterhub.readthedocs.io/en/stable/reference/oauth.html are some of the things that really broght a lot of light on the overall picture of the hub internals and how they interact ✨ Thank you @minrk for taking the time to write them!

docs/source/reference/database.md
manics: @GeorgianaElena There's an out of date roadmap https://jupyterhub.readthedocs.io/en/stable/contributing/roadmap.html 
Might be worth a team-compass discussion for updating or replacing it?

sgibson91: Yeah, roadmaps are a great tool for this but they need regular(ish, depending on how long the roadmap is intended to last) tending/updating. So then next after the team-compass discussion on updating it, we'd need another discussion about how to organise regular updates to the roadmap. It can be a time-consuming process that can be difficult to coordinate for, but very beneficial for the project and contributors.

docs/source/reference/database.md
```diff
@@ -1,29 +1,120 @@
...
+
+We often say that the Hub 'owns' the database.
+This ownership means that we assume the Hub is the only process that will talk to the database.
+This assumption enables us to make several caching optimizations that dramatically improve JupyterHub's performance (i.e. data written recently to the database can be read from memory instead of fetched again from the database) that would not work if multiple processes could be interacting with the database at the same time.
+
+Database operations are also synchronous, so while JupyterHub is waiting on a database operation, it cannot respond to other requests.
+This allows us to avoid complex locking mechanisms, because transaction races can only occur during an `await`, so we only need to make sure we've completed any given transaction before the next `await` in a given request.
+
+:::{note}
+We are slowing working to remove these assumptions, and moving to a more traditional db session per-request pattern.
```
ajpower, line 55:
```suggestion
We are slowly working to remove these assumptions, and moving to a more traditional db session per-request pattern.
```

docs/source/reference/database.md
minrk: I wonder if GitHub projects might be a better way to represent some of these things?

I do think the hub team is due for a 'roadmap workshop' to review/reset/communicate medium and long-term goals across the project(s).

sgibson91: Oooh, I like both of these ideas!

manics: I'm in favour of not using a static RTD page for the roadmap.... that's why I suggested a new team-compass issue, instead of sidetracking this PR 😄

PR #4297, opened 2023-01-13
docs: fix linkcheck in gallery

minrk:
- jupyter.chameleoncloud SSL is failing (I can reproduce with conda curl, but not /usr/bin/curl, so seems to be a CA issue)
- remove dead arnesund tag link (keep single article link)


minrk:
This fixes the affected links, but linkcheck is still failing because virtualbox's whole website is down.

PR #4301, opened 2023-01-17
Created folders to house the restructured documentation

alwasega:
Hello @minrk, @sgibson91, and everyone. 
I have created folders and index files for the new sub-folders that will house the restructured JupyterHub documentation as mapped [here](https://docs.google.com/spreadsheets/d/1Nnwji40YXMklnQ2yvO3rcfGnMP8S_LsVlkKdobo8ZjM/edit#gid=1020565119). Please review the draft PR and let me know if any changes are necessary.

### DIFF
--- a/docs/source/diataxis-FAQ/index.md
+++ b/docs/source/diataxis-FAQ/index.md
@@ -1,7 +1,6 @@
 # FAQs
 
-This section will contain documentation that contain frequently-asked questions. 
-
+This section will contain documentation that contain frequently-asked questions.
 
 ```{toctree}
 :maxdepth: 2
--- a/docs/source/diataxis-FAQ/index.md
+++ b/docs/source/diataxis-FAQ/index.md
@@ -1,8 +1,0 @@
-# FAQs
-
-This section will contain documentation that contain frequently-asked questions.
-
-```{toctree}
-:maxdepth: 2
-
-```
--- a/docs/source/diataxis-FAQ/index.md
+++ b/docs/source/diataxis-FAQ/index.md
@@ -1,0 +1,9 @@
+# FAQs
+
+This section will contain documentation that contain frequently-asked questions. 
+
+
+```{toctree}
+:maxdepth: 2
+
+```
--- a/docs/source/diataxis-explanation/index.md
+++ b/docs/source/diataxis-explanation/index.md
@@ -1,8 +1,0 @@
-# Explanation
-
-This section will contain restructured documentation that fit into the `Explanation` part of the Diataxis model.
-
-```{toctree}
-:maxdepth: 2
-
-```
--- a/docs/source/diataxis-explanation/index.md
+++ b/docs/source/diataxis-explanation/index.md
@@ -1,0 +1,8 @@
+# Explanation
+
+This section will contain restructured documentation that fit into the `Explanation` part of the Diataxis model. 
+
+```{toctree}
+:maxdepth: 2
+
+```
--- a/docs/source/diataxis-explanation/index.md
+++ b/docs/source/diataxis-explanation/index.md
@@ -1,6 +1,6 @@
 # Explanation
 
-This section will contain restructured documentation that fit into the `Explanation` part of the Diataxis model. 
+This section will contain restructured documentation that fit into the `Explanation` part of the Diataxis model.
 
 ```{toctree}
 :maxdepth: 2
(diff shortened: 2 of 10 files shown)


minrk:
Great! Some notes:

- let's not put `diataxis` in the folder names, or use it in the headline of each section
- let's avoid capitalizing folder or file names, which can have peculiar behaviors on case-sensitive vs case-insensitive filesystems and URLs

i.e. folders should be `faq`, `reference`, `howto`, etc.

For the index headlines, let's use a for-readers summary of what's in each section, rather than referencing a framework readers may not be familiar with.

 Readers shouldn't need to know about diataxis to interact with the docs, so it shouldn't show up often as a term. That's for us as documentation _authors_.  A single short explainer about how we organize with links to diataxis (and any relevant 'interpretations' of diataxis specific to jupyterhub) seems like the right level.

alwasega:
Thanks, @minrk.

alwasega:
Please check the updates in the latest commit @minrk and @sgibson91.

minrk (approved):
Makes sense!

I think  the doc build is failing because the new index pages aren't referenced in the top-level index yet.

sgibson91:
This is much better - thanks!

What are your next plans for this PR? :)

sgibson91:
>  I noted the fails as well and this could be the reason. Can I create the references in the top-level index in the next PR?

I wouldn't merge this while the docs test is failing, as that means that your changes won't actually get deployed :/

alwasega:
> Ok, RTD has built the preview, so I guess it's how much appetite we have for merging a broken test

Please advise on how best to proceed. I could update the top-level index file here or keep in mind to do so whenever I make any changes that affect it going forward.

manics:
I don't think we should deliberately break a test unless there's a good justification- could you perhaps add that justification as a comment?

alwasega:
Thanks @manics. 
The purpose of this PR was to create the folders that would house the restructured documentation and I was avoiding having to edit any other files here. However, seemingly, the new folders also have to be referenced in the top-level index file for all tests to pass. I can do so, only that I was trying to avoid making changes to other existing files at the moment.

minrk:
Adding one new toctree which references the new folders to the top-level `docs/source/index.md` should solve the failure. Try adding this directly below `## Contents`:

````markdown
Documentation sections (reorganization in-progress)

```{toctree}
:maxdepth: 1

faq/index.md
tutorial/index.md
howto/index.md
explanation/index.md
reference/index.md
```
````

alwasega:
> Try adding this directly below ## Contents:

I have done this but the test still fails.

sgibson91:
> I have done this but the test still fails. Seems to be an issue with broken links in the `authenticator-users-basics` file.

Rather than a file in this repo, I think these broken links are do with changes made in oauthenticator.readthedocs.io I know a bunch of work has been happening over there that could have caused this disruption.

Since the linkcheck is failing on _external_ links, and nothing that has been changed by this PR, I would be happy to merge this in and address the broken external links in a new PR (Allan you don't have to make that PR!)

sgibson91:
I think this is the upstream PR that has caused the broken links https://github.com/jupyterhub/oauthenticator/pull/561

GeorgianaElena:
Yes, I believe [this PR](https://github.com/jupyterhub/oauthenticator/pull/561) is the one causing the issues.

There are some redirects set with sphinx-rediraffe however in
https://github.com/jupyterhub/oauthenticator/blob/82b1627fc87360f244350fe23d3cb17491cac697/docs/source/conf.py#L182-L204

But I don't think they know how to redirect when a header is also present in the link. Fox example:
- this redirect works https://oauthenticator.readthedocs.io/en/latest/api/gen/oauthenticator.auth0.html
- but this doesn't https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.auth0.html#module-oauthenticator.auth0

@consideRatio, I was thinking to fix this on the `oauthenticator` side by explicitly listing these anchor links for rediraffe? WDYT? Are you aware of any other, more simple rediraffe features that we might use to fix this instead?

GeorgianaElena:
I checked and adding the explicit links with anchors to oautnenticator results in circular redirects. So I'm opening a PR to fix the link on the `jupyterhub`side now.

Also this statement
~"_but this doesn't https://oauthenticator.readthedocs.io/en/latest/api/gen/oauthenticator.auth0.html#module-oauthenticator.auth0_"~ is wrong, and it actually redirects correctly. So, the issue might be with how `sphinx-build linkcheck` does things.

sgibson91:
Thank you Georgiana! I'm going to merge this and we can address anything other problems with the linkcheck separately :)

alwasega:
Thank you all for your help!

PR #4302, opened 2023-01-18
sqlalchemy 2 compatibility

minrk:
- avoid backref warnings by adding objects to session explicitly before creating any relationships
- remove unnecessary `[]` around scalar query
- use `text()` wrapper on connection.execute
- engine.execute is removed
- update import of declarative_base
- raise on any sqlalchemy 2 removal warnings during testing

as sqlalchemy 2 is in release candidate, this probably warrants a backport to 3.1 at least, and maybe also 2.0?

closes #4312
closes #4313

PR #4303, opened 2023-01-19
make sure event-schemas are installed

minrk:
when not installed, event publishing won't work because they will show as 'unregistered', since the schemas are unavailable.

revealed in #4302 because I switched it to testing against an install instead of the repo


consideRatio (approved):

minrk:
@meeseeksdev please backport to 3.x

PR #4304, opened 2023-01-19
Fix the oauthenticator docs api links

GeorgianaElena:
Fixes the `oauthenticator` links to match their new location and also removes the anchors, which were pointing to the beginning of the doc page anyway.

Reference https://github.com/jupyterhub/jupyterhub/pull/4301#issuecomment-1396859946


minrk:
Merging to unblock things, but maybe these should be using intersphinx references instead?

PR #4310, opened 2023-01-25
Verify port number in JUPYTERHUB_SERVICE_URL

twalcari:
When implementing a custom spawner, a sneaky bug can appear when Spawner.port is not implemented: when calling `def get_env()` the `JUPYTERHUB_SERVICE_URL` is generated with port number 0. While this was ignored by _nbviewer_ in the past, it now results in difficult to diagnose behavior where the URL on which the Jupyter server starts defaults to port 80 ( See https://github.com/jupyter/docker-stacks/issues/1862).

I think it is preferable that the Spawner crashes instead.

Instructions to solve the issue: either define a static port in your Spawner:

```python
    @default("port")
    def _default_port(self):
        return 8888
```

or initialize the port to a random port number, as happens in `LocalProcessSpawner.start`: https://github.com/twalcari/jupyterhub/blob/cbce162e7cfa7da7997df6a6c2c37248d06ac527/jupyterhub/spawner.py#L1662-L1663

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -994,6 +994,9 @@ class Spawner(LoggingConfigurable):
             # this should only occur in mock/testing scenarios
             base_url = '/'
 
+        if self.port < 1 or self.port > 65535:
+            raise ValueError("Port must be between 1 and 65535")
+
         proto = 'https' if self.internal_ssl else 'http'
         bind_url = f"{proto}://{self.ip}:{self.port}{base_url}"
         env["JUPYTERHUB_SERVICE_URL"] = bind_url


jupyterhub/tests/test_spawner.py
```diff
@@ -56,6 +56,7 @@ def new_spawner(db, **kwargs):
     kwargs.setdefault('term_timeout', 1)
     kwargs.setdefault('kill_timeout', 1)
     kwargs.setdefault('poll_interval', 1)
+    kwargs.setdefault('port', 5555)
```
minrk, line 59: Since starting a new spawner will generally not be passed a port, we shouldn't need a port here. If this is required for tests to pass, we may need to think about what's changing in a breaking way.

jupyterhub/services/service.py
```diff
@@ -123,6 +123,8 @@ def user_env(self, env):
 
     def start(self):
         """Start the process"""
+        if self.port == 0:
+            self.port = random_port()
```
minrk, line 127: services already have this logic, because it's inherited [from LocalProcessSpawner here](https://github.com/jupyterhub/jupyterhub/blob/e255ada1691a8300509085238707b0e21af7c4d5/jupyterhub/spawner.py#L1659-L1660).

jupyterhub/spawner.py
```diff
@@ -994,6 +994,9 @@ def get_env(self):
             # this should only occur in mock/testing scenarios
             base_url = '/'
 
+        if self.port < 1 or self.port > 65535:
```
minrk, line 997: This is a breaking change that takes away the option of letting the launched process pick the port, as is done in e.g. batchspawner, which phones home the resulting port after launch - a pattern I think we want _more_ spawners to be able to do, since it will select ports more reliably than a random number on the Hub's system which is typically not the same port namespace.

Ideally, this could be communicated better and fail _when not supported_, which is most of the time.

How about this:

- validate `0 <= port <= 65535`
- if port is 0, _log_ that the launched process will be picking the port, and that the Spawner better be able to handle that.

Or, we could have a small breaking change that defines an overrideable method `validate_port` that can be overridden to allow for port 0 (the default would be your check). Maybe even with a boolean attribute on the spawner class to indicate whether port 0 should be allowed.

WDYT?

jupyterhub/services/service.py
twalcari: You override the `start()` function in this class, and don't call `super.start()` anywhere.

Hence this is why the tests started failing, and I copied the logic into this class: https://github.com/jupyterhub/jupyterhub/actions/runs/4005957446/jobs/6876931281#step:10:421

jupyterhub/spawner.py
twalcari: The reason that I raise an error instead of logging an error is that in the case of `port == 0` an invalid `JUPYTERHUB_SERVICE_URL` will be generated on the next lines.  Invalid in the sense that `jupyter/docker-stacks` is tripping over it and starting the server on port 80 instead of the default port 8888 or even an explicitely set JUPYTER_PORT.

Maybe part of the solution is  to add a check in `nbviewer` to make sure that a valid port is used? I can create a pull request for this commit: https://github.com/twalcari/nbviewer/commit/ef2ac83d7fe8b006f53ec6e0c0eb7b23fdc8b44e

jupyterhub/services/service.py
minrk: Ah, forgot about the lack of super start. Then this change is definitely correct. Thanks!

jupyterhub/spawner.py
minrk: It doesn't produce an invalid string, though. It produces a string instructing the server to bind on port 0, which is asking the server to pick its own port. To work properly with JupyterHub, this must be _combined_ with the logic required to pass the port info back to the Hub after starting, as is done in BatchSpawner.

For _services_, which don't allow custom spawners, the `0->random port` above will fix this. The problem with the validation here that it affects _all_ Spawners, prohibiting a valid use case, when it should only be affecting the service case, and thus more appropriately confined to the Service class.

> Maybe part of the solution is to add a check in nbviewer to make sure that a valid port is used? 

Yeah, I think that's a good idea.

minrk:
Do you have your configuration for the nbviewer service? It should have a `url` field in it, which should ultimately populate this field.

I think the bug might really be that $JUPYTERHUB_SERVICE_URL is defined _at all_ when nbviewer is registered as a service without a URL. It shouldn't be possible to specify that a service has a web endpoint without specifying the port where it will run. Unlike a regular Spawner, there are services that don't have URLs (this is the default if a URL is not specified). An `else:` branch [here](https://github.com/jupyterhub/jupyterhub/blob/3.1.0/jupyterhub/services/service.py#L390-L392) that explicitly sets `env['JUPYTERHUB_SERVICE_URL'] = ''` if `self.url` is not set would be an indicator that it's launching a service it doesn't expect to have a URL. That should also make it impossible for this event to occur, I think.

twalcari:
Some context: I maintain a custom spawner for our JupyterHub service which works on top of 'GPULab' at Ghent University. We run the `jupyter/docker-stacks` Docker images as they perfectly fit our needs. 

Our users all use JupyterLab. nbviewer is not used. I do observe that changing the port of the `JUPYTERHUB_SERVICE_URL` to 0 causes JupyterLab to launch on port 80. When another port is set (for example with a `JUPYTERHUB_SERVICE_URL` like `http://127.0.0.1:5555/user/twalcari@ilabt.imec.be/`) then JupyterHub will launch on port 5555 instead. 

I honestly don't fully grasp what this environment variable does in the context of JupyterLab and the `jupyter/dockerstacks` Docker images. Why is it influencing the port on which JupyterLab launches on in the first place?

**Steps to reproduce:**

1. Launch a webserver which answers to all requests (I use the example code from [this gist](https://gist.github.com/bradmontgomery/2219997))

2. Run `docker run --rm -p 8888:8888 --user 1000 --env-file env-test jupyter/minimal-notebook` with the content of the `env-test` file being:

```
JPY_API_TOKEN=0c1f95e5d1384faeaa9b945bf21b8e3b
JUPYTERHUB_ACTIVITY_URL=http://172.24.124.101:19998/hub/api/users/twalcari@ilabt.imec.be/activity
JUPYTERHUB_API_TOKEN=0c1f95e5d1384faeaa9b945bf21b8e3b
JUPYTERHUB_API_URL=http://172.24.124.101:19998/hub/api
JUPYTERHUB_BASE_URL=/
JUPYTERHUB_CLIENT_ID=jupyterhub-user-twalcari%40ilabt.imec.be
JUPYTERHUB_OAUTH_ACCESS_SCOPES=["access:servers!server=twalcari@ilabt.imec.be/", "access:servers!user=twalcari@ilabt.imec.be"]
JUPYTERHUB_OAUTH_CALLBACK_URL=/user/twalcari@ilabt.imec.be/oauth_callback
JUPYTERHUB_OAUTH_CLIENT_ALLOWED_SCOPES=[]
JUPYTERHUB_OAUTH_SCOPES=["access:servers!server=twalcari@ilabt.imec.be/", "access:servers!user=twalcari@ilabt.imec.be"]
JUPYTERHUB_SERVICE_PREFIX=/user/twalcari@ilabt.imec.be/
JUPYTERHUB_SERVICE_URL=http://127.0.0.1:0/user/twalcari@ilabt.imec.be/
#JUPYTERHUB_SERVICE_URL=http://127.0.0.1:5555/user/twalcari@ilabt.imec.be/
JUPYTERHUB_USER=twalcari@ilabt.imec.be
```

minrk:
If you're launching a singleuser server, the $JUPYTERHUB_SERVICE_URL->port is handled [here](https://github.com/jupyterhub/jupyterhub/blob/eca9c00872cbf5faa2f9b81b1ff9f5f30f610554/jupyterhub/singleuser/mixins.py#L334-L344) and there is a bug in that `if url.port` will behave the same for `http://host/path` and `http://host:0/path` which is not correct - port 0 should mean bind to port 0 and let the OS assign a port.

I do think it may be time for a small backward-incompatible change:

1. add an attribute on Spawner, e.g. `Spawner.supports_server_assigned_port`
2. make it opt-in (False on the base Spawner class)
3. reject port 0 `if not self.supports_server_assigned_port`

This will result in port 0 failing informatively and by default, as you're doing here, but still allowing for the 'user server decides' pattern to work going forward.

twalcari:
Ok, after reading a bit about `JUPYTERHUB_SERVICE_URL` I guess that it also applies to JupyterLab as that is also a _service_ in the eyes of JupyterHub? So the case where a valid port (ex. `5555`) is provided should indeed result in JupyterLab starting on port 5555.

The reason that I opened the issue on `jupyter/docker-stacks` is due to the changed behavior when port `0` is being sent. Previously it launched JupyterLab on port 8888, which is also the default when no environment variables have been passed. Now (due to [this change](https://github.com/jupyter/docker-stacks/commit/0d324bc0b38ccd9feac269d521abcf6476324e42#diff-e7e599e8c9c2032a64bd44cdb51bb5dadfa6540102cbe6f822c5799e1b5e2befL12)?) this behavior has changed and now it launches on port 80 instead. This seems unwanted, but I don't understand how this behavior is triggered, and searching for `JUPYTERHUB_SERVICE_URL` in the organisations 'jupyter' and 'jupyterlab' does not yield any useful results.

yuvipanda:
Thoughts on how to move this PR forward (or close it) given it's been more than a year and a half?

Issue #4312, opened 2023-01-27
JupyterHub is broken due to latest release of SqlAlchemy 


dclong:
### Bug description

JupyterHub installed using `pip3 install jupyterhub` fails to launch with the following error messages due to a breaking release of SqlAlchemy (2.0.0). 
```
[I 2023-01-26 17:04:14.612 JupyterHub app:2810] Running JupyterHub version 3.1.0
[I 2023-01-26 17:04:14.612 JupyterHub app:2840] Using Authenticator: jupyterhub.auth.PAMAuthenticator-3.1.0
[I 2023-01-26 17:04:14.612 JupyterHub app:2840] Using Spawner: jupyterhub.spawner.LocalProcessSpawner-3.1.0
[I 2023-01-26 17:04:14.612 JupyterHub app:2840] Using Proxy: jupyterhub.proxy.ConfigurableHTTPProxy-3.1.0
[I 2023-01-26 17:04:14.613 JupyterHub app:1694] Writing cookie_secret to /root/jupyterhub_cookie_secret
[E 2023-01-26 17:04:14.618 JupyterHub app:3334]
    Traceback (most recent call last):
      File "/usr/local/lib/python3.10/dist-packages/jupyterhub/app.py", line 3331, in launch_instance_async
        await self.initialize(argv)
      File "/usr/local/lib/python3.10/dist-packages/jupyterhub/app.py", line 2852, in initialize
        self.init_db()
      File "/usr/local/lib/python3.10/dist-packages/jupyterhub/app.py", line 1823, in init_db
        self.session_factory = orm.new_session_factory(
      File "/usr/local/lib/python3.10/dist-packages/jupyterhub/orm.py", line 1061, in new_session_factory
        check_db_revision(engine)
      File "/usr/local/lib/python3.10/dist-packages/jupyterhub/orm.py", line 953, in check_db_revision
        current_table_names = set(inspect(engine).get_table_names())
      File "/usr/local/lib/python3.10/dist-packages/sqlalchemy/inspection.py", line 111, in inspect
        ret = reg(subject)
      File "/usr/local/lib/python3.10/dist-packages/sqlalchemy/engine/reflection.py", line 304, in _engine_insp
        return Inspector._construct(Inspector._init_engine, bind)
      File "/usr/local/lib/python3.10/dist-packages/sqlalchemy/engine/reflection.py", line 237, in _construct
        init(self, bind)
[...]

PR #4313, opened 2023-01-27
Restrict the SqlAlchemy dependency to be <2.0.0

dclong:
fixes https://github.com/jupyterhub/jupyterhub/issues/4312


consideRatio:
- I think @minrk wishes to resolve this by #4302

minrk:
I'll make a 3.1.1 with #4302 today

PR #4319, opened 2023-01-27
require sqlalchemy 1.4

minrk:
removes some workarounds needed for sqlalchemy 1.1 + 2.0 support

1.4 backports most 2.0 behavior, keeping it off-by-default for an easier opt-in transition. #4302 needed several conditionals to avoid bumping the dependency version, which I didn't want to do in the same PR to make backporting easier.

There doesn't appear to be a great way to avoid the SADeprecationWarning on the engine_connect signature until we require sqla 2, even though our callback accepts the new signature. This is because the warning condition only checks if we accept the _old_ signature and takes that path, it doesn't seem to allow us to opt-in to the new signature while still accepting the old signature.

### DIFF
--- a/jupyterhub/orm.py
+++ b/jupyterhub/orm.py
@@ -8,9 +8,7 @@ from datetime import datetime, timedelta
 
 import alembic.command
 import alembic.config
-import sqlalchemy
 from alembic.script import ScriptDirectory
-from packaging.version import parse as parse_version
 from sqlalchemy import (
     Boolean,
     Column,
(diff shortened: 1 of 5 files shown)


jupyterhub/orm.py
```diff
@@ -1067,6 +1048,8 @@ def new_session_factory(
     elif url.startswith('mysql'):
         kwargs.setdefault('pool_recycle', 60)
 
+    kwargs.setdefault("future", True)
```
manics, line 1051: What are the circumstances where a caller should be allowed to pass `future=False`?

minrk: Good question! If someone were using sqlalchemy 1.4 and found a _problem_ with our switch to 2.0 lifecycles, this gives them an escape hatch to keep the older behavior without needing to downgrade to JupyterHub. It should have no practical effect if everything's been done correctly. The alternatives are:

- unconditional `kwargs['future'] = True` (always get 2.0 behavior, even on 1.4, no escape hatch but always consistent hardcoded behavior)
- leave it out (status quo - get deprecated but default 1.x behavior on 1.4, 2.0 behavior on upgrade or user opt-in)

This is a no-op on 2.0.

manics: Thanks, I'll let you decide!

manics (approved):

PR #4321, opened 2023-01-29
Sample Playwright test

bl-aire:
This commit tests if:

- The Login page's title contains the string "JupyterHub"

- If the page has a Sign-in button

> The PR isn't up for being merged yet. I thought to create it so we have a way to keep track of the conversion regarding using Playwright or not. If Playwright gets selected, I will make necessary changes on this branch.

@minrk


manics:
I tried playwright for testing https://github.com/manics/zero-to-jupyterhub-k8s-examples/ in GitHub (e.g. see https://github.com/manics/zero-to-jupyterhub-k8s-examples/blob/main/ldap-singleuser/tests/test_jupyterhub_ldap.py). I found it's a lot nicer to use than Selenium, and the [recording mode](https://playwright.dev/docs/codegen-intro) is very helpful for writing new tests. The main downside to be aware of is (IIRC) it uses special instrumented builds of browsers.

minrk:
Thanks @bl-aire! This is part of a discussion on whether we want to move the new selenium tests @mouse1203 has added this Fall to playwright, which seems to be popular. Our browser testing needs are pretty light, but the lack of needing a browser install is definitely nice. And the fact that playwright has its own async API, which we had to make ourselves for selenium.

We were definitely burned by a testing tool bundling its own browser builds (phantomjs, which bundled an increasingly outdated QtWebKit, I think), because it didn't behave like a real-world browser at all in several important ways, and that lead to loads of headaches solving problems that could never affect real users. I don't think playwright is in that situation.

manics:
I'd support switching to Playwright, I've found configuring Selenium and Geckodriver for Firefox on my own projects to be a pain.

consideRatio:
I've not heard about playwright before and greatly appreciate the though of not needing to think about selenium/geckodriver/firefox.

For reference, also note that [jupyterhub/jupyter-server-proxy](https://github.com/jupyterhub/jupyter-server-proxy/tree/main/tests/acceptance) are setup with [robotframework](https://robotframework.org/) together with [robotframework-jupyterlibrary](https://robotframework-jupyterlibrary.readthedocs.io/) to start servers etc, render things, inspect outcomes, etc. geckodriver/firefox is required to be setup for this.

bl-aire:
Hi @minrk 

I added Eslint-plugin-jsx-ally to have a first-hand experience of what's possible with it and I was able to access the admin page(jsx components) of JupyterHub using:

- ESLint command line interface(CLI)

Screenshot 2023-02-06 at 19 37 06

- VS Code's ESLint extension: 
Screenshot 2023-02-06 at 20 09 50

minrk:
We've completed the playwright migration. Thanks, @bl-aire!

PR #4324, opened 2023-01-30
Remove no longer relevant notice in readme

consideRatio:
- Reverts #3506 as I think its no longer needed.

Note the pre-commit.ci failure can be resolved by bumping isort, something that needs doing across all our repos to resolve :'(

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,15 +8,6 @@
 
 ---
 
-Please note that this repository is participating in a study into the sustainability of open source projects. Data will be gathered about this repository for approximately the next 12 months, starting from 2021-06-11.
-
-Data collected will include the number of contributors, number of PRs, time taken to close/merge these PRs, and issues closed.
-
-For more information, please visit
-[our informational page](https://sustainable-open-science-and-software.github.io/) or download our [participant information sheet](https://sustainable-open-science-and-software.github.io/assets/PIS_sustainable_software.pdf).
-
----
-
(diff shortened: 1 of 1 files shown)


consideRatio:
@minrk what is the logic behind letting this be "maintenance" rather than "docs"? I don't mind, but I'd like align with your mental model on how to label for consistency.

My model said "this PR only touch a .md file, so its a documentation pr". If this comment would have been made as an inline comment in a .py file, I'd maybe labelled this as maintenance just because a .py file was touched.

minrk:
I'm not sure! (you can see I did both.) I felt since it was a note about the status of the repo itself in the README (not the docs), it didn't feel like a documentation change (in no practical way did our documentation of jupyterhub change), so I put maintenance as "when browsing the changelog, this PR should be in the lowest priority section because it's extremely unlikely to be relevant." I'm fine with the docs label, too, since it is a change to a document.

Having to make classification decisions like this are actually part of why I don't really like generated changelogs - not all patches really belong in an changelog, and usually almost all (if not all) PRs in the maintenance category don't contain information of relevance to anyone reading a changelog, so are safely and rightly ignored.

Changelogs serve two purposes, though! One is collecting useful information for readers considering upgrades (or selecting minimum required versions, etc.), and the other is contributor-recognition. I can see this thought process prioritizing the former resulting in not great results for the latter, which is maybe even more important.

PR #4326, opened 2023-01-30
Improve contrast on muted text

bl-aire:
On accessing the pages with axe DevTools again, I found text that had a low contrast which I did not notice previously.

This PR increases the contrast of muted text on the Token page.

**Before:**

Screenshot 2023-01-30 at 13 18 17

**After:**

Screenshot 2023-01-30 at 12 16 18

@minrk

### DIFF
--- a/share/jupyterhub/static/less/variables.less
+++ b/share/jupyterhub/static/less/variables.less
@@ -20,6 +20,8 @@
 @brand-success: @brand-primary;
 @brand-danger: #d7191c;
 
+@text-muted: #222;
+
 .btn-jupyter {
   .button-variant(#fff; @jupyter-orange; @jupyter-red);
 }


share/jupyterhub/static/less/page.less
```diff
@@ -70,3 +70,7 @@
   border-color: @jupyter-orange;
   outline-color: @jupyter-orange;
 }
+
+.text-muted {
```
minrk, line 74: Is this class required? I suspect that assigning the variable value [does this already](https://github.com/twbs/bootstrap/blob/v3.4.1/less/type.less#L105).

minrk (commented):
Thanks!

Looks like a `jupyterhub.sqlite` file was accidentally committed. Can you remove it?

I think defining the variable is sufficient, without also using it again in the class. I linked above to where twitter bootstrap should already do this.

Can you also share before/after screenshots when changing CSS to show the changes?

share/jupyterhub/static/less/page.less
bl-aire: I commented it out and got the same result so I have removed it.

bl-aire:
> Looks like a `jupyterhub.sqlite` file was accidentally committed. Can you remove it?

I checked my .gitignore on the branch and saw `jupyterhub.sqlite` is listed there. I don't know how it got committed. How can I fix it? `git rm`?
 
> I think defining the variable is sufficient, without also using it again in the class. I linked above to where twitter bootstrap should already do this.

I fixed this.

> Can you also share before/after screenshots when changing CSS to show the changes?

I will add that now

minrk:
Thanks!

PR #4311, opened 2023-01-25
Moved first half of HowTo documentation

alwasega:
I have moved the first half of the docs marked for the `HowTo` section of the restructured documentation.

PR #4331, opened 2023-01-31
Added rediraffe using auto redirect builder

alwasega:
Hi @GeorgianaElena and @sgibson91, 
I have set up rediraffe as discussed yesterday and created a `redirects.txt` file to which the redirects have been written using auto redirect builder. 
I based the changes on commit #[0873902](https://github.com/jupyterhub/jupyterhub/pull/4301/commits/0873902a18bfcddc0525876f80385399eb992f82).
Please check and advise if all is okay.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -214,11 +214,28 @@ ogp_use_first_image = True
 # -- Options for the rediraffe extension -------------------------------------
 # ref: https://github.com/wpilibsuite/sphinxext-rediraffe#readme
 #
-# This extensions help us relocated content without breaking links. If a
-# document is moved internally, a redirect like should be configured below to
+# This extension helps us relocate content without breaking links. If a
+# document is moved internally, a redirect link should be configured as below to
 # help us not break links.
 #
-rediraffe_branch = "main"
+# The workflow for adding redirects can be as follows: 
+# 1. Change "rediraffe_branch" below to point to the commit/ branch you
+#   want to make the changes based off. 
+# 2. Option 1: run "make rediraffecheckdiff"
+#       a. Analyze the output of this command
+#       b. Manually add the redirect entries to the "redirects.txt" file.
+#    Option 2: run "rediraffewritediff"
+#       a. rediraffe will then automatically add the obvious redirects to redirects.txt
+#       b. Analyze the output of the command for broken links. 
+#       c. Check the "redirects.txt" file for any files that were moved/ renamed but are not listed. 
+#       d. Manually add the redirects that havebeen mised by the automatic builder to "redirects.txt"
+#    Option 3: Do not use the commands above and, instead, do everything manually - by taking 
+#    note of of the files you have moved or renamed and adding them to the "redirects.txt" file.
+#  
+# If you are basing changes off another branch/ commit, always change back
+# rediraffe_branch to main before pushing your changes upstream. 
+#
+rediraffe_branch = "main" 
 rediraffe_redirects = "redirects.txt"
 # rediraffe_redirects = {
 # "old-file": "new-folder/new-file-name",
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -220,6 +220,6 @@ ogp_use_first_image = True
 #
 rediraffe_branch = "main"
 rediraffe_redirects = "redirects.txt"
-#rediraffe_redirects = {
-    # "old-file": "new-folder/new-file-name",
-#}
+# rediraffe_redirects = {
+# "old-file": "new-folder/new-file-name",
+# }
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -220,17 +220,17 @@ ogp_use_first_image = True
 #
 # The workflow for adding redirects can be as follows:
 # 1. Change "rediraffe_branch" below to point to the commit/ branch you
-#   want to make the changes based off.
+#    want to base off the changes.
 # 2. Option 1: run "make rediraffecheckdiff"
-#       a. Analyze the output of this command
+#       a. Analyze the output of this command.
 #       b. Manually add the redirect entries to the "redirects.txt" file.
-#    Option 2: run "rediraffewritediff"
-#       a. rediraffe will then automatically add the obvious redirects to redirects.txt
+#    Option 2: run "make rediraffewritediff"
+#       a. rediraffe will then automatically add the obvious redirects to redirects.txt.
 #       b. Analyze the output of the command for broken links.
 #       c. Check the "redirects.txt" file for any files that were moved/ renamed but are not listed.
-#       d. Manually add the redirects that havebeen mised by the automatic builder to "redirects.txt"
+#       d. Manually add the redirects that have been mised by the automatic builder to "redirects.txt".
 #    Option 3: Do not use the commands above and, instead, do everything manually - by taking
-#    note of of the files you have moved or renamed and adding them to the "redirects.txt" file.
+#    note of the files you have moved or renamed and adding them to the "redirects.txt" file.
 #
 # If you are basing changes off another branch/ commit, always change back
 # rediraffe_branch to main before pushing your changes upstream.
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -219,6 +219,7 @@ ogp_use_first_image = True
 # help us not break links.
 #
 rediraffe_branch = "main"
-rediraffe_redirects = {
+rediraffe_redirects = "redirects.txt"
+#rediraffe_redirects = {
     # "old-file": "new-folder/new-file-name",
-}
+#}
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -218,24 +218,24 @@ ogp_use_first_image = True
 # document is moved internally, a redirect link should be configured as below to
 # help us not break links.
 #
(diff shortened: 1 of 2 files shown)


docs/source/conf.py
```diff
@@ -219,6 +219,7 @@ def setup(app):
 # help us not break links.
```
consideRatio, line 219: I'm new to using the automation to figure out redirects and this comment is now outdated.

How should one update this file? Manually or using a tool? I'll understand the manual part, but I never learned how tools could be used.

If using a command line tool, could you share a one-liner or similar how you used the tool in the config comment here?

alwasega: Generally, if one's aim is to update redirects, there are two ways to go about it based on the branch/ commit specified in the `rediraffe_branch` line in the conf.py file. : 

1. Manually add the old and new file links to the dict given in the `conf.py` file. 
2. Or run `make rediraffewritediff` and the redirects will be automatically written to an external `redirects.txt` file.

I created the `redirects.txt` file and commented out the dict given originally (lines 223 - 225) as rediraffe auto redirect builder only writes to an external file and not the dict.
 
Going forward, you can manually add redirects to this `redirects.txt` file or, alternatively, run `make rediraffewritediff` and the changes will (hopefully) reflect in the `redirexts.txt` file. 

Please advise further @GeorgianaElena.

GeorgianaElena: > How should one update this file? Manually or using a tool? I'll understand the manual part, but I never learned how tools could be used.

@consideRatio, I only figured out yesteday how to use the rediraffe diff checker and the auto redirect builder, but I believe they might be useful.

@alwasega is right, if there is a separate text file where the redirects are stored, instead of having them in a dict in `conf.py`, then:

### running the diff checker
will output to stdout what are the redirects that should be added to the list, by comparing the current state of the files (where HEAD is pointing) against the state of the files in `rediraffe_branch` (this can also hold a commit, not just a branch).

  ```bash
  make rediraffecheckdiff
  ```

### running the autobuilder
will append (I _think_ based [on code](https://github.com/wpilibsuite/sphinxext-rediraffe/blob/d8961c2402015dafa6039ed27d973ad835d7fc09/sphinxext/rediraffe.py#L386), because it's not documented ) to the redirects file, the redirects for which there's an 100% accuracy and print to stdout the others.
   ```bash
    make rediraffewritediff
   ```
- example output:
Screenshot 2023-01-31 at 11 20 15

### Manual steps

Because this automation tools don't work with deleted files, I _believe_ we should also checkout the redirects with accuracy < 100% and manually add the ones that make sense.

### Documentation
We should document the workflow to modify `redirects.txt` and the `rediraffecheckdiff` and  `rediraffewritediff` either in `conf.py` or `redirects.txt` or both.

alwasega: Thanks @GeorgianaElena. I see a number of redirects with the `Error: [broken]` tag. Do you mean that we should manually add these files to the `redirects.txt` file if they are valid docs and have not, for example, been deleted?

GeorgianaElena: @alwasega, I belive so, yes. The workflow for adding redirects can then be something like the following:

- one adds a commit (or multiple ones) that moves files around, renames them, etc
- before these commits get merged into the main branch
- they:
   1. **option 1**: run `make rediraffecheckdiff`
     a. analise the output of this command
     b. and manually add the redirect entries to `redirects.txt`
   2. **option 2**: run `make rediraffewritediff`
     a. rediraffe will then automatically add the obvious redirects to `redirects.txt`
    b. they also analise the output of the command
    c. and manually add the redirects that make sense but have been missed by the automatic builder to `redirects.txt`
   3. **option 3**: do everything manual, by taking notes of the files that they've moved or renamed

GeorgianaElena (commented):
This looks good to me @alwasega! ✨  Thank you

I believe the next steps would be to:
- make sure that `rediraffewritediff` didn't miss any file:
   - do an an overview of the things it failed to add itself
   - and a manual inspection of the files that have been moved in previous PRs
- update the comment below with instructions on how to use the rediraffe's diff checker and auto builder and how to modify the `redirects.txt` file https://github.com/jupyterhub/jupyterhub/blob/e85b91cd9bc24db6e09d0c6d17f1e754dc08c72c/docs/source/conf.py#L214-L220

alwasega:
I have updated the redirects that autobuilder missed @GeorgianaElena and added comments on the `conf.py` file on how to update the `redirects.txt` file.

docs/source/conf.py
```diff
@@ -214,11 +214,29 @@ def setup(app):
...
+# 1. Change "rediraffe_branch" below to point to the commit/ branch you
+#   want to make the changes based off.
+# 2. Option 1: run "make rediraffecheckdiff"
+#       a. Analyze the output of this command
+#       b. Manually add the redirect entries to the "redirects.txt" file.
+#    Option 2: run "rediraffewritediff"
+#       a. rediraffe will then automatically add the obvious redirects to redirects.txt
+#       b. Analyze the output of the command for broken links.
+#       c. Check the "redirects.txt" file for any files that were moved/ renamed but are not listed.
+#       d. Manually add the redirects that havebeen mised by the automatic builder to "redirects.txt"
```
sgibson91, line 231:
```suggestion
#       d. Manually add the redirects that have been missed by the automatic builder to "redirects.txt"
```

docs/source/conf.py
```diff
@@ -214,11 +214,29 @@ def setup(app):
...
+#    Option 3: Do not use the commands above and, instead, do everything manually - by taking
+#    note of of the files you have moved or renamed and adding them to the "redirects.txt" file.
```
sgibson91, line 233:
```suggestion
#    note of the files you have moved or renamed and adding them to the "redirects.txt" file.
```

sgibson91 (approved):
A few suggestion to clean up the comment, but otherwise this LGTM!

alwasega:
> A few suggestion to clean up the comment, but otherwise this LGTM!

Thanks @sgibson91. Fixed the typos :)

docs/source/conf.py
consideRatio: Thanks for looking into this @alwasega and @GeorgianaElena!

> I think based [on code](https://github.com/wpilibsuite/sphinxext-rediraffe/blob/d8961c2402015dafa6039ed27d973ad835d7fc09/sphinxext/rediraffe.py#L386), because it's not documented

If we end up needing to document things for ourselves beyond something small, it sould probably instead be attempted to document it upstream. Its more work improving documentation upstream, but it would be great!!

I think you are best informed to make a decision on what to do with regards to how we work with the redirects in this repo for now. If experience is gained and you start feeling confident on a practice, we try go for that more systematically in other repos long term.

I think the options on how to work with the redirects are:

1. Fully manual as now, where we figure out ourselves if we need to add a redirect
2. Mostly manual, where we can go fully manual but also suggest that `make rediraffecheckdiff` can be used to get informed on what redirects we should make.
3. Semi automated, where we suggest using `make rediraffewritediff` to update a file (not overwrite right?)
4. Mostly automated, where we have a github workflow triggered when docs change to inspect if a redirect needs to be setup and otherwise fails to let us know.
5. Almost fully automated, where we also push a commit with the relevant redirects.

If I understand this correctly, we need to consider this ourselves no matter what because the tools can't figure it out fully in all cases. That makes me think we shouldn't go for option 5. Maybe option 2-4? No matter what, I'm a big :+1: for going for something you feel like going for!

If a redirects.txt file is added, its a big plus I think if its possible to have a comment in its top helping anyone opening and wondering what its about to learn what its about, for example by "For more information, see rediraffe configuration in docs/source/conf.py" or similar.

manics: If it's automated (or routinely run) should we only run it immediately before a release? If docs are moved twice we don't need a redirect for an intermediate edit.

alwasega: @minrk and @sgibson91, please weigh in on this. 

My thoughts are that:
 
- As we presently move the docs, we realize breaks in links as in https://github.com/jupyterhub/jupyterhub/pull/4311. Internally, we are fixing the breakages using appropriate MyST references. 
- For external purposes, just using MyST references may not suffice. So, as a backup, we are also using rediraffe, which will traverse all links and point a user to one that works for a doc that was moved/ renamed. 


- Generally, I think that since we are moving the docs now, we should also ensure that both MyST and rediraffe-generated links are updated this time instead of later. I would mostly go for 2 as given by @consideRatio: applying `make rediraffecheckdiff` to check what redirects to make and add them to the `redirects.txt` file as appropriate.  

> If a redirects.txt file is added, its a big plus I think if its possible to have a comment in its top helping anyone opening and wondering what its about to learn what its about, for example by "For more information, see rediraffe configuration in docs/source/conf.py" or similar.

I agree with this as it will help anyone making any further documentation moves/ renames.

sgibson91: I think the discussion of automating the redirect generation is out of scope for the this PR and should be moved into a new issue.

@alwasega If you push another commit that adds a comment to the top of redirects.txt, then we can merge I think.

alwasega: Okay, @sgibson91. I have added some comments to the file.

sgibson91: I opened https://github.com/jupyterhub/jupyterhub/issues/4333 from Erik's comment above

sgibson91 (approved):

PR #4332, opened 2023-01-31
simplify some async fixtures

minrk:
use async fixtures for simpler event-loop integration

several of these fixtures were written before fixtures themselves could be async, but now they can, which means we can use async/await instead of run_sync.

related to https://github.com/jupyterhub/pytest-jupyterhub/issues/30

### DIFF
--- a/jupyterhub/tests/conftest.py
+++ b/jupyterhub/tests/conftest.py
@@ -29,7 +29,6 @@ import asyncio
 import copy
 import os
 import sys
-from functools import partial
 from getpass import getuser
 from subprocess import TimeoutExpired
 from unittest import mock
@@ -58,7 +57,7 @@ def ssl_tmpdir(tmpdir_factory):
 
 
 @fixture(scope='module')
-def app(request, io_loop, ssl_tmpdir):
+async def app(request, io_loop, ssl_tmpdir):
     """Mock a jupyterhub app for testing"""
     mocked_app = None
     ssl_enabled = getattr(
@@ -70,10 +69,6 @@ def app(request, io_loop, ssl_tmpdir):
 
     mocked_app = MockHub.instance(**kwargs)
 
-    async def make_app():
-        await mocked_app.initialize([])
-        await mocked_app.start()
-
     def fin():
         # disconnect logging during cleanup because pytest closes captured FDs prematurely
         mocked_app.log.handlers = []
@@ -84,7 +79,8 @@ def app(request, io_loop, ssl_tmpdir):
             print("Error stopping Hub: %s" % e, file=sys.stderr)
 
     request.addfinalizer(fin)
-    io_loop.run_sync(make_app)
+    await mocked_app.initialize([])
+    await mocked_app.start()
     return mocked_app
 
 
@@ -131,7 +127,12 @@ def event_loop(request):
 
 @fixture(scope='module')
 async def io_loop(event_loop, request):
-    """Same as pytest-tornado.io_loop, but re-scoped to module-level"""
+    """Mostly obsolete fixture for tornado event loop
+
+    Main purpose is to register cleanup (close) after we're done with the loop.
+    The main reason to depend on this fixture is to ensure your cleanup
+    happens before the io_loop is closed.
+    """
     io_loop = AsyncIOMainLoop()
     assert asyncio.get_event_loop() is event_loop
     assert io_loop.asyncio_loop is event_loop
@@ -144,11 +145,13 @@ async def io_loop(event_loop, request):
 
 
 @fixture(autouse=True)
-def cleanup_after(request, io_loop):
+async def cleanup_after(request, io_loop):
     """function-scoped fixture to shutdown user servers
 
     allows cleanup of servers between tests
     without having to launch a whole new app
+
+    depends on io_loop to ensure it runs before things are closed.
     """
 
     try:
@@ -167,10 +170,10 @@ def cleanup_after(request, io_loop):
             for name, spawner in list(user.spawners.items()):
                 if spawner.active:
                     try:
-                        io_loop.run_sync(lambda: app.proxy.delete_user(user, name))
+                        await app.proxy.delete_user(user, name)
                     except HTTPError:
                         pass
-                    io_loop.run_sync(lambda: user.stop(name))
+                    await user.stop(name)
             if user.name not in {'admin', 'user'}:
                 app.users.delete(uid)
         # delete groups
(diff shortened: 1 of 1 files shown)


GeorgianaElena:
@Sheila-nk, do you want to help out with reviewing Min's PR?

I believe it could help with putting the info in https://github.com/jupyterhub/pytest-jupyterhub/issues/30 into context, but also identify what are the confusion points, so that we can document them. Feel free to ping and ask any additional clarification. Asking questions is an important piece in the review process.

What do you think?

minrk:
I tried to remove the io_loop fixture, which I don't believe we _really_ need anymore, but it did cause some failures. The only thing the io_loop fixture really does is call [`io_loop.close(all_fds=True)`](https://github.com/jupyterhub/jupyterhub/blob/e85b91cd9bc24db6e09d0c6d17f1e754dc08c72c/jupyterhub/tests/conftest.py#L140) to ensure everything's cleaned up and FDs don't leak from one test module to the next. This isn't necessary if we actually clean up everything, which we _mostly_ do. But I did see failures such that at least _sometimes_ we rely on this to close the http socket for the server, so I decided to leave it alone for now. I think if the `app` fixture more reliably cleaned up everything it started, this would not be necessary.

Sheila-nk:
@minrk @GeorgianaElena 👋 I have some questions first.

> In JupyterHub, we use tornado, which is a wrapper around asyncio's event loop. It's unlikely that the tornado io_loop will be needed directly in tests. It's very important that the tornado io loop is 'hooked up to' the main asyncio event loop, otherwise you'll get mysterious hangs, as events you are waiting for just never happen because they are hooked up to an event loop that's not running.

I was trying to understand why the `io_loop` is no longer needed considering JupyterHub uses tornado as you had mentioned. When you say that tornado is a wrapper around `asyncio` does that mean that pytest-asyncio can be used to write tests for Tornado applications? And is 'hooking' to the main event loop necessary anymore?

Second,  why is this `io_loop` still calling the `event_loop` fixture even with the `async` keyword provided? Is it necessary because of the override?

Lastly, from studying the changes to the `app` fixture, my understanding is that the `make_app` function had to go since the `initialize` and `start` functions are already registered coroutines, so they can easily be called using the `await` keyword within the `app` coroutine. Am I correct in saying this?

minrk:
> I was trying to understand why the io_loop is no longer needed considering JupyterHub uses tornado as you had mentioned

To be clear, the tornado IOLoop _object_ is still used. What may not be needed is the _fixture_ to do special setup and teardown of the loop. Instantiating JupyterHub calls `IOLoop.current()` which gets the IOLoop associated with the current asyncio event loop, and creates it if necessary. _If_ the MockHub cleanup fully closes and stops everything, the extra io_loop teardown doesn't need to happen. That doesn't appear to be _quite_ always the case.

> When you say that tornado is a wrapper around asyncio does that mean that pytest-asyncio can be used to write tests for Tornado applications?

Yes, there's not generally any need for special handling of tornado _unless_ you also want/need to _close_ the IOLoop's resources at the end. Tornado calls will generally work fine if run from inside asyncio without doing any special tornado setup beforehand. This has changed since we started writing the jupyterhub test suite, and there are still some remnants of the old ways. It's much simpler now.

>  And is 'hooking' to the main event loop necessary anymore?

The 'hooking' is done in tornado itself, we mainly assert that it happens, just in case something goes wrong in tornado setup which would cause the two-loops-problem. Calling `AsyncIOMainLoop()` from inside a coroutine (why we `async def io_loop()`) will always create an IOLoop attached to the _current_ event_loop.

> Second, why is this io_loop still calling the event_loop fixture even with the async keyword provided? Is it necessary because of the override?

Only for the assertion, really. We could remove it an assert against `asyncio.get_running_loop()`. But putting the reference there also ensures that setup/teardown happens in the right order.

> from studying the changes to the app fixture, my understanding is that the make_app function had to go since the initialize and start functions are already registered coroutines, so they can easily be called using the await keyword within the app coroutine. Am I correct in saying this?

Pretty much. I wouldn't say it _had to go_, but it _could_ go because we can now use top-level awaits. We only had the function because we need functions to pass to `run_sync`. Since moving to `async def` means we can have top-level awaits, we no longer need that extra layer for wrapping `await`s, and can do the steps we wanted in order in a simpler way without a wrapper.

jupyterhub/tests/conftest.py
```diff
@@ -322,26 +325,26 @@ def cleanup():
...
+            await service.server.wait_up(http=True)
     return service
 
 
 @fixture
-def mockservice(request, app):
+async def mockservice(request, app):
     """Mock a service with no external service url"""
-    yield _mockservice(request, app, url=False)
+    yield await _mockservice(request, app, url=False)
```
Sheila-nk, line 335: I have not encountered a scenario where `yield` and `await` are used in this way. Can you please help me understand the usage?

Sheila-nk, line 335: I have not encountered a scenario where `yield` and `await` are used in this way. Can you please help me understand the usage?

minrk: Absolutely! Fixtures are used to do one or both of:

1. set something up before the test
2. clean up after the test

If you only need to do setup, the fixture function can `return` what should be the value of the fixture (if any) in tests that use it.

If you need to _cleanup_, then you can `yield` what should be passed to the functions (can be nothing, if it's purely setup/teardown and has no object to be passed to tests). You can read more in the [pytest docs on fixture teardown/cleanup](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#yield-fixtures-recommended).

So a fixture with setup _and_ teardown generally looks like:

```python
@pytest.fixture
def my_fixture():
    setup_before()
    yield value_for_tests_if_any
    cleanup_after()
```

The `await` doesn't change anything here. When you `yield <expression>`, what's yielded is the _result_ of the expression, which means the `awaited` result if there's an `await` in there. That means

```python
yield await async_thing
```

is the same as

```python
thing = await async_thing
yield thing
```

if that makes things clearer.


If you run [this sample test](https://gist.github.com/d6fa0aaaa35ec8bad55db5136499c728) with:

```
pytest -vs example.py
```

you'll get:

```
example.py::test_sample in setup
in async function
myfixture='sync result'
PASSEDin teardown
```

Sheila-nk: I definitely understand this better. Thank you for the explanation.
When I run the code I get this though:
```python
example.py::test_sample <async_generator object myfixture at 0x7f22b8408af0>
PASSED
```

minrk: Forgot about my global config. It should be:

```
pytest -vs --asyncio-mode auto example.py

Sheila-nk: Yeah! This works now. Thank you!

Sheila-nk:
@minrk Thank you for taking the time to answer my questions. Getting to review this PR was insightful.
 I think this PR is ready for merging 🚀 
@GeorgianaElena I will leave that to you

GeorgianaElena:
Thanks @Sheila-nk and @minrk

PR #4340, opened 2023-02-06
Moved Explanation/Background files

alwasega:


alwasega:
@minrk , @sgibson91 , 
Please check this PR in which I have moved the docs marked for the Explanation/ Background section of the documentation. Of note:
 
1. I have renamed the section to `Background` from `Explanation` to accommodate the docs under the new `About JupyterHub` subsection. As `Background` is still a correct term to use under Diataxis, I feel it is more appropriate here as docs such as _A Gallery of JupyterHub deployments_ do not explain a concept but just give some background information. 
Screenshot 2023-02-06 at 11 52 54


2. I have retained the `RBAC` docs as a folder outside Background as I was encountering a `Sphinx error (builder name xxx not registered)` error when I changed the following lines to point to a new path within `Makefile`: 
Screenshot 2023-02-06 at 11 46 13

I figure it is something to do with the paths in the `generate-scope-table.py` file, but retaining the `RBAC` folder where it is and referencing its contents within the `background/index.md` file also works.

sgibson91:
A few quick thoughts from me:

- I think it's totally okay to rename Explanation to Background so long as a reader is clear what the sections are for
- I think pages like "A Gallery of JupyterHub Deployments" should actually live in the Reference section rather than Background. This is because you'd look up a deployment on that page (i.e. reference it), but you don't actually learn much about any of the deployments specifically.
- I think it's also ok to update paths in any Python scripts that generate content if we want to move them too. But that can be done in a fresh PR if you feel it needs to be.

alwasega:
Thanks @sgibson91. 

> I think it's also ok to update paths in any Python scripts that generate content if we want to move them too. But that can be done in a fresh PR if you feel it needs to be.

I won't change the paths as I found referencing the folder within the `background/index.md` file works. 

> I think pages like "A Gallery of JupyterHub Deployments" should actually live in the Reference section rather than Background. This is because you'd look up a deployment on that page (i.e. reference it), but you don't actually learn much about any of the deployments specifically.

Your explanation makes sense :). I will move this back to its original folder and shift it to `reference` in a different PR. 

> I  think it's totally okay to rename Explanation to Background so long as a reader is clear what the sections are for. 

I will change this back to `Explanation` as that makes more sense once I remove the docs I have in the `About JupyterHub` section in the above commit.

sgibson91:
> I won't change the paths as I found referencing the folder within the `background/index.md` file works.

So long as it makes sense from the viewpoint of a maintainer working on the docs as well. For example, would they know to look for those files there in terms of the directory structure of the docs folder?

alwasega:
> For example, would they know to look for those files there in terms of the directory structure of the docs folder? We can signpost using comments in the Markdown files.

Let me add the comments @sgibson91. They would really be of value to a maintainer.

minrk (approved):

minrk:
This is really great! I think it will be worthwhile to split the rbac docs explicitly into their reference and explanation subsections, but that will require some rewriting, so it's okay to mark it as 'todo' for now.

alwasega:
Thanks @minrk. I agree that we would need to split the rbac section further as it contains some specialized technical reference docs as well.

sgibson91 (approved):
Thanks @alwasega!

alwasega:
Thank you too @sgibson91 and @minrk :)

PR #4343, opened 2023-02-09
Restructured references section of the docs

alwasega:
Restructured the references documentation by moving files into various subfolders. I have also separated the references into two main sections: Technical references and API references.


alwasega:
@sgibson91, please advise how I can resolve the failing test above.

sgibson91:
@alwasega I reran the test and it passes now. I think this was a transient error: they JupyterHub instance that is deployed for testing was not in the correct state to pass that particular test, rather than the test actually being broken. Even if it was, the break would not have been caused by any of your changes, so you wouldn't have needed to resolve it anyway.

alwasega:
Thank you @sgibson91. Please check the restructured docs and if what I have done is appropriate.

sgibson91:
For future readers: The reason we saw the tests triggered and transiently fail in this PR was because pre-commit made changes to some of the test files in this commit [`fed4cd5` (#4343)](https://github.com/jupyterhub/jupyterhub/pull/4343/commits/fed4cd5e4024c7ab241cee2ea2f27ed3058ad601)

sgibson91 (approved):

Issue #4344, opened 2023-02-10
push singleuser to localhost registry has started to hang at push step


minrk:
The release action has started to fail in a really weird way. The build-push-action just _stops_ [at the very end](https://github.com/jupyterhub/jupyterhub/actions/runs/4124177212/jobs/7123116179#step:15:425), pushing the image, terminating after 6 hours (normally, this step takes about 2 minutes).

It's suspicious that this started to happen shortly after #4339, which was a major bump to the now failing action. But weird things:

- it succeeded two times after the merge, so it's not simply that change in the action
- the same action reliably succeeds earlier, building the jupyterhub/jupyterhub image. It's reliably only the jupyerhub/singleuser push that is failing
- the build succeeds fine, it's only the push/export that's failing

I'm going to explore this, but it's a mystery

PR #4347, opened 2023-02-10
debug: build-push hang

minrk:
debugging what's up with #4344

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -84,6 +84,7 @@ jobs:
 
   publish-docker:
     runs-on: ubuntu-20.04
+    timeout-minutes: 30
 
     services:
       # So that we can test this in PRs/branches
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -150,7 +150,7 @@ jobs:
           branchRegex: ^\w[\w-.]*$
 
       - name: Build and push jupyterhub
-        uses: docker/build-push-action@v3.1.1
+        uses: docker/build-push-action@3b5e8027fcad23fda98b2e3ac259d8d67585f671
         with:
           context: .
           platforms: linux/amd64,linux/arm64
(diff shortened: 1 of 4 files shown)


minrk:
wild! The change in the jupyter/base-notebook from 2023-01-30 to 2023-02-09 prevents the push from succeeding (not the build, though)! I cannot imagine what could be the cause of that, but at least we hav a short-term fix.

mathbunnyru:
Wow, this is surprising for me as well, because there are not many things that have changed between these images.
If you haven't already done this, let's make sure, that the issue is reproducible.

I mean, the push process might be stuck randomly and it might not depend on the Jupyter Docker Stacks image at all.
Can you please close this PR, wait for 10 secs and reopen it to see if it always fails with fresh image?

minrk:
@mathbunnyru I have, I opened this PR because we've had reliable (100% of the time) failures on the push since the image update. Check out the failures over the last week [here](https://github.com/jupyterhub/jupyterhub/actions/workflows/release.yml), starting on February 8, and resolved only in the recent commits that pinned `base-notebook:2023-01-30`:

Screenshot 2023-02-13 at 09 32 58

I'm sure it's nothing you did, and probably something wonky in the infrastructure, but it is an absolute mystery to me how push could hang like this so I don't have any idea how to pursue it. I've replicated the build process locally as best I can, and it pushes just fine.

I don't want to waste anyone's time debugging what's hopefully a bizarre anomaly, so my plan is to wait a while and try updating to a later build in some days or weeks, and just hope it goes away by itself :)

consideRatio:
I suspect that this relates to the docker push action introduced a breaking change in 3.3.0 that they reverted in 3.3.1, then added in 4.0.0. There are notes about it in https://github.com/docker/build-push-action/releases/tag/v4.0.0.

In practice, I guess what has happened is that the jupyter/docker-stacks images started building with the new breaking change, and then when we rely on them, we observe issues.

minrk:
@consideRatio great idea! I suspected that, too, and noted that we had two successful builds and pushes _after_ merging the update to v4.0.0, and docker-stacks doesn't appear to use that action. But there could well be an interaction I don't understand there.

Whatever the ultimate issue is, certainly a docker build bug that whatever's failing manifests as a hang rather than an error.

minrk:
neither setting provenance=false and downgrading to 3.1.1 fixed it here, so it really seems that whatever changed, changed in the upstream image.

minrk:
It appears the wait-and-see strategy has paid off! 2022-02-28 works without a hang, and we hopefully never need to figure out what was so unique about the early February builds.

PR #4354, opened 2023-02-15
Make singleuser server-extension default

minrk:
For JupyterHub 4.0, when run under jupyter-server 2.0, jupyterhub-singleuser will use the `singleuser.extension` path, rather than the subclass/monkeypatch approach.

The old path is 'deprecated' insomuch as it can be removed when we require Jupyter Server 2.0, but will not be removed until that point and there's no pressure to do that any time soon. I don't think It's worth warning anyone about.

Follow-up to #3888

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,7 +28,6 @@ on:
 env:
   # UTF-8 content may be interpreted as ascii and causes errors without this.
   LANG: C.UTF-8
-  PYTEST_ADDOPTS: "--verbose --color=yes"
   SQLALCHEMY_WARN_20: "1"
 
 permissions:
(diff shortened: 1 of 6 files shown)


minrk:
Added test coverage of subdomains and SSL revealed some things that needed fixing, so I added two matrix entries to ensure ssl and subdomains are both tested for both the singleuser extension and mixin approaches.

Because this adds a couple of matrix entries, I also added a `subset` argument to pass to `pytest -k`, since we really don't need to run _all_ the tests for each server variant, as only the `singleuser` tests should really have different results with different versions.

Our tests don't take super long (~8 minutes), so we don't need to go too far optimizing coverage without too much redundancy, but this might save a bit of waste. By far our longest task is the docker image build, which I still want to get into a standalone repo (https://github.com/jupyterhub/team-compass/issues/474).

minrk:
All set, I think!

consideRatio (approved):

consideRatio:
:heart: :tada: :sunflower: looks great!

Issue #4353, opened 2023-02-13
Can we with a metric measure if jupyterhub is CPU constrained?


consideRatio:
I've watched a few grafana dashboards, most recently [examples from here](https://github.com/jupyterhub/grafana-dashboards/pull/59) where hub response times can peak to seconds. Seeing that I got thinking that the hub pod from time to time get starved on CPU.

In practice though, it is perhaps rather that some responses are stuck waiting on something, like stopping a server, before it completes. So, it would be incorrect to draw the conclusion that the hub pod is overburdened by requests when we see long response times.

The question I raise is then instead, can we understand if jupyterhub is overburdened somehow using exposed prometheus metrics? If not, can we develop a new metric for this somehow?

yuvipanda:
I think the most useful metric here is 'tick duration' - how long does a single 'tick' of the event loop take? I know that when we were on the tornado ioloop @minrk has this be printed during debug, and it was *super* helpful. However when we moved to asyncio ioloop we lost this. A tick duration that's too long usually means there's too much CPU bound stuff going on that's not async enough, and causes responsiveness problems.

I don't know if we can measure that though.

minrk:
> when we moved to asyncio ioloop we lost this

To be clear, it's _tornado_ that switched to asyncio.

asyncio has its own [debug mode](https://docs.python.org/3/library/asyncio-dev.html) that logs similar info. I don't think it quite includes tick duration, but it does note each _individual coroutine_ that blocks for more than 100ms (configurable threshold). Lacking the 'tick duration' metric means the asyncio debug log wouldn't capture the 'everything individually is async and cooperative and fast, but there's just too damn many coroutines right now' case.

 I'm not sure how we could hook it up to metrics. Maybe we could hook up the asyncio debug logger and count 'long task block' events? It would be hard to get labels on them, I think.

minrk:
Looking into asyncio, we _could_ patch asyncio's `_run_once` as is done [in the asyncio test](https://github.com/python/cpython/blob/8a2b7ee64d1bde762438b458ea7fe88f054a3a88/Lib/test/test_asyncio/test_events.py#L1708-L1714) to add a timer for the tick, equivalent to what we had in tornado. This is a private API, but private APIs in the standard library are very stable, so can generally be safely written against for debugging purposes, as long as you gracefully handle their eventual removal, which should be easy enough for a metric like this.

Something like:

```python
loop = asyncio.get_running_loop()
original_run_once = loop._run_once
[...]

PR #4362, opened 2023-02-20
[DO NOT MERGE] add example metric for asyncio loop duration

minrk:
Not a real proposal as-is, but this is what a metric for #4353 could look like.

https://github.com/python/cpython/issues/101946 includes proposals for two possibly more useful metrics, which could show _which_ coroutines are slow (rather than merely _that_ something is slow).

Notably, this does _not_ directly indicate that a Hub is CPU-bound, merely a measure of 'how blocking it is', which is perhaps more relevant, but less directly actionable as it measures a directly user-experienced property of the Hub (responsiveness), but doesn't include any information about the cause beyond what may be in the surrounding logs.

Also of note: you can actually define and enable this metric _purely from jupyterhub_config.py_ if someone wants to experiment with it.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -70,6 +70,7 @@ from .emptyclass import EmptyClass
 from .handlers.static import CacheControlStaticFilesHandler, LogoHandler
 from .log import CoroutineLogFormatter, log_request
 from .metrics import (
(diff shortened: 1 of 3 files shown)


minrk:
closed in favor of #4615

PR #4363, opened 2023-02-20
JupyterHub sphinx theme

minrk:
Try out the new theme

The only thing I immediately noticed was the links in the corner needed updating, since discourse was added by default and the css class names needed to change to get the right size.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -9,9 +9,9 @@
 --editable .
 
 autodoc-traits
+jupyterhub-sphinx-theme
 myst-parser
 pre-commit
-pydata-sphinx-theme
 pytablewriter>=0.56
 ruamel.yaml
 sphinx>=4
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -34,6 +34,7 @@ extensions = [
     "sphinx-jsonschema",
     "sphinxext.opengraph",
     "sphinxext.rediraffe",
+    "jupyterhub_sphinx_theme",
     "myst_parser",
 ]
 root_doc = "index"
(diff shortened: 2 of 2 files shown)


choldgraf:
the icon links looked OK to me in the preview (at least they were the same size as all the other buttons):

image

minrk:
Sorry, I should have been clearer. The second commit fixes the links so everything looks good now. That was the only thing I noticed that needed changing after the theme change.

minrk:
`make linkcheck` appears to be failing in a relevant way:

```
Running Sphinx v5.3.0
[15](https://github.com/jupyterhub/jupyterhub/actions/runs/4224203153/jobs/7334868623#step:7:16)
making output directory... done
[16](https://github.com/jupyterhub/jupyterhub/actions/runs/4224203153/jobs/7334868623#step:7:17)
loading intersphinx inventory from https://docs.python.org/3/objects.inv...
[17](https://github.com/jupyterhub/jupyterhub/actions/runs/4224203153/jobs/7334868623#step:7:18)
loading intersphinx inventory from https://www.tornadoweb.org/en/stable/objects.inv...
[18](https://github.com/jupyterhub/jupyterhub/actions/runs/4224203153/jobs/7334868623#step:7:19)

[19](https://github.com/jupyterhub/jupyterhub/actions/runs/4224203153/jobs/7334868623#step:7:20)
Extension error (jupyterhub_sphinx_theme):
[20](https://github.com/jupyterhub/jupyterhub/actions/runs/4224203153/jobs/7334868623#step:7:21)
Handler <function set_config_defaults at 0x7f9c7b047c10> for event 'builder-inited' threw an exception (exception: 'CheckExternalLinksBuilder' object has no attribute 'theme_options')
[21](https://github.com/jupyterhub/jupyterhub/actions/runs/4224203153/jobs/7334868623#step:7:22)
make: *** [Makefile:27: linkcheck] Error 2
```

choldgraf:
Ah yeah, we just need to add a check for "if app.builder.theme_options" and skip it if it doesn't exist (or we could check to make sure it's the HTML builder of some kind). I'm not going to be at a computer for a while but happy to review and merge a PR from my phone!

minrk:
Success with 0.0.3!

PR #4365, opened 2023-02-21
admin panel: fix condition for start/stop buttons on user servers

### LINKED ISSUE #4361: admin page shows stop/access links for stopped servers
It appears that JupyterHub has arrived to a faulty state about active servers in a jupyterhub 3.1.1 installation I manage (z2jh based, latest dev release of helm chart).

There may be a need to protect for this happening, and/or recovering from it happening. I hope this issue report can help as a clue to fixing a small bug that this is, as it doesn't seem to cause much issues.

Since the metrics gets it right, and the /hub/admin gets it wrong, it seems reasonable that JupyterHub could figure that the state is incosistent by itself.

### Logs from hub startup one day ago

```
[I 2023-02-19 22:52:56.478 JupyterHub app:2539] redacted-user-a still running
[I 2023-02-19 22:52:56.478 JupyterHub app:2539] redacted-user-b still running
[I 2023-02-19 22:52:56.479 JupyterHub app:2879] Initialized 2 spawners in 0.055 seconds
```

### Incorrect state reported

image

### Access server presents this

image

### Metrics of `jupyterhub_running_servers` is correct

```
# HELP jupyterhub_running_servers the number of user servers currently running
# TYPE jupyterhub_running_servers gauge
jupyterhub_running_servers 2.0
```

### Starting server works

image

### Stopping started server works

image

minrk:
`.ready` is the right switch for the links, not `.started` which can be defined even after they stop

closes #4361

### DIFF
--- a/jsx/src/components/ServerDashboard/ServerDashboard.jsx
+++ b/jsx/src/components/ServerDashboard/ServerDashboard.jsx
@@ -298,7 +298,7 @@ const ServerDashboard = (props) => {
           {server.last_activity ? timeSince(server.last_activity) : "Never"}
         </td>
         <td data-testid="user-row-server-activity">
-          {server.started ? (
+          {server.ready ? (
             // Stop Single-user server
             <>
               <StopServerButton serverName={server.name} userName={user.name} />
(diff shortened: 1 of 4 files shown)


consideRatio (approved):
Nice LGTM!

consideRatio (approved):

minrk:
Managed to find and fix the selenium failure, I think.

PR #4371, opened 2023-02-22
Expose `oauth_no_confirm_list` setting

### LINKED ISSUE #4370: Expose `oauth_no_confirm_list` configuration to users
### Proposed change
Currently, the `needs_oauth_confirm` method of `OAuthAuthorizeHandler` checks for the oauth clients that don't need confirmation. It ignores the clients in the `oauth_no_confirm_list` set 
(https://github.com/jupyterhub/jupyterhub/blob/ce7085b720f690144f9fe987eb373c3a0bd83db2/jupyterhub/apihandlers/auth.py#L180-L189)

But the `oauth_no_confirm_list`  is generated automatically at startup and is not configurable by users. 

https://github.com/jupyterhub/jupyterhub/blob/ce7085b720f690144f9fe987eb373c3a0bd83db2/jupyterhub/app.py#L2722-L2729

It would be useful if users could define the initial value of this set before populating it with the other services. And since it is a white list, we could support the `regex` pattern in this list to allow more freedom on service names.
 
### Alternative options
I don't know if we have other ways to populate this `oauth_no_confirm_list` set.

In our use case, we use `JupyterHub` as an OAuth provider for external applications which are not managed by `JupyterHub`, this feature will help us to bypass the confirmation screen of all applications at once.

- Allow users to define `oauth_no_confirm_list`  in the configuration file with a list of services name or regex patterns, e.g:

```python
c.JupyterHub.oauth_no_confirm_list = ['my-oauth-service', '^(my-prefix)']
```
- Update the current logic of populating the `oauth_no_confirm_list` set and the verification in the `needs_oauth_confirm` method.

trungleduc:
Closes #4370

### DIFF
--- a/jupyterhub/apihandlers/auth.py
+++ b/jupyterhub/apihandlers/auth.py
@@ -10,7 +10,7 @@ from oauthlib import oauth2
 from tornado import web
 
 from .. import orm, roles, scopes
-from ..utils import get_browser_protocol, token_authenticated
+from ..utils import get_browser_protocol, match_regex_patterns, token_authenticated
 from .base import APIHandler, BaseHandler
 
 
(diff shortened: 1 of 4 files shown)


trungleduc:
Closing this PR the feature is provided by #4381.

Issue #4364, opened 2023-02-21
Documentation To-Dos after restructruring using the Diataxis framework


alwasega:
Over the past weeks, we (@sgibson91 and myself) have been restructuring the JupyterHub docs as outlined in #4257. Based on this exercise, there are some pending tasks that would need looking into: 

1. In #4272 , @minrk suggested that: 
> Looking at the layout, I think we probably do want more tutorial content, and perhaps one way to do that is to place some of the docs like 'basics' in tutorials and mark them for rewriting with the diataxis tutorial definition in mind.

Based on this suggestion, we moved the following docs to the `Tutorial` section: 
- Configuration Basics
- Networking Basics
- Authentication and User Basics
- Security settings
- External services
- Spawners and single-user notebook servers
However, these docs do not read like the typical Diataxis tutorials (lessons that take a reader - who is unfamiliar with the product - by hand through a series of steps to complete a project). Hence there is need to further refine or expand them. 

2. The following `Reference` docs can be split further or refined:
- Authenticators  (into **Base Authenticator** and **PAM Authenticator**, for example)
- Spawners (into **base Spawner class** and **default implementation**, for example)
- Services - can we have Tutorials or How-Tos that takes a reader through examples of working with **managed** and **unmanaged** services? 

3. The `RBAC` files can be split further into separate Explanation and References, based on @minrk's comment in #4340, i.e.
> I think it will be worthwhile to split the rbac docs explicitly into their reference and explanation subsections, but that will require some rewriting

4.  Ensure consistency across all docs, for example, use sentence case in all titles. 

5. Resolve #4333 on how to work with documentation redirects from rediraffe.

minrk:
Great! I think the main thing we should have prior to a 4.0 release is:

1. all file relocations are done, _not_ including those that should be rewritten or split, and
2. redirects are in place for all the moved files

That will let us iterate as we have time/availability on the docs that should be split.

I think it may be good to actually put a call to action _in the docs_ that we already know should be split, indicating that it could be a good place for folks to contribute.

sgibson91:
ref https://github.com/jupyterhub/jupyterhub/issues/4350

alwasega:
> all file relocations are done, not including those that should be rewritten or split

@minrk, the restructuring of the docs is almost done, and these files have already been moved. Should we then un-move them? Alternatively, I could add attention call-outs at the top of the files mentioning the need to split/ rewrite them.

sgibson91:
[...]

PR #4377, opened 2023-02-28
reduce nested hierarchy in docs organization

alwasega:
related to https://github.com/jupyterhub/jupyterhub/issues/4364#issuecomment-1447966100


docs/source/redirects.txt
```diff
@@ -17,5 +17,5 @@
 "reference/rest.md" "howto/rest.md"
```
minrk, line 17: Looks like we need a redirect for events/index.md

docs/source/reference/index.md
```diff
@@ -12,117 +12,27 @@ This section is divided into two broad subsections:
...
 
-rest-api
-api-reference/app
-api-reference/auth
-api-reference/spawner
-api-reference/proxy
-api-reference/user
-api-reference/service
-api-reference/services.auth
+reference-api/index.md
```
minrk, line 37: This can probably go in the main toctree above, now that it's a separate index. And can we the folder `api` (so it's `reference/api` and just `api/index` here).

minrk (commented):
Looks great! I left a few comments, but altogether, I think this will make it easier to get an overview without spreading things quite so thin.

alwasega:
@minrk, please check if the commit resolves your comments above.

minrk:
Great, thanks!

PR #4379, opened 2023-03-01
remove pin from singleuser

### LINKED ISSUE #4344: push singleuser to localhost registry has started to hang at push step

minrk:
mysterious upstream image problem appears to have been fixed just as mysteriously

closes #4347 
closes #4344

### DIFF
--- a/singleuser/Dockerfile
+++ b/singleuser/Dockerfile
@@ -1,7 +1,7 @@
 # Build as jupyterhub/singleuser
 # Run with the DockerSpawner in JupyterHub
 
-ARG BASE_IMAGE=jupyter/base-notebook:2023-01-30
+ARG BASE_IMAGE=jupyter/base-notebook
 FROM $BASE_IMAGE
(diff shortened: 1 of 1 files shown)


consideRatio:
:tada: haha okay!

PR #4381, opened 2023-03-01
Add REST API for managing services at runtime

trungleduc:
### Proposed change
This PR allows users with authorized scope to create and remove services dynamically at runtime. Since we will have two ways to define a service (via the config file and via REST API), a column is added to the `services` table to distinguish the origin of services.

#### New scope
A new scope `admin:services` is added, this scope is required to be able to modify the services at runtime.

#### Add a new service via REST API
To add a new service, send a POST request to this endpoint
```
POST /hub/api/services/:servicename
```
**Required scope: `admin:services`**
**Payload**: The payload should contain the definition of the service to be created. The endpoint supports the same properties as services defined in the config file.
**Possible responses**
- `201 Created`: The service and related objects are created (and started in the case of a Hub-managed one) successfully.
- `400 Bad Request`: The payload is invalid or JupyterHub can not create the service.
- `409 Conflict`: The service with the same name already exists.

#### Remove an existing service via REST API
To remove a new service, send a DELETE request to this endpoint
```
DELETE /hub/api/services/:servicename
```
**Required scope: `admin:services`**
**Payload**: `None`
**Possible responses**
- `200 OK`: The service and related objects are removed (and stopped in case of a Hub-managed one) successfully.
- `400 Bad Request`: JupyterHub can not remove the service.
- `404 Not Found`: The requested service does not exist.
- `405 Not Allowed`: The requested service is created from the config file, it can not be removed at runtime.

### Todo

- [x] Update the `service` database schema
  - [x]  Distinguish services registered at runtime from services that came from config.
  - [x]  Ensure the content parity between the services defined from the config file and from the database.
- [x] Logic to reconcile runtime changes from config changes at startup
  - [x] Remove services from the db when removed from config
  - [x] Leave services created at runtime alone during startup
- [x] Initialize services with config defined in the database.
     - [x] Create services from `orm.Service`
- [x] Add scope for managing services by REST API.
- [x] Add handlers for service CRUD operations.
- [x] Logic to handle non-managed/managed services at run time.
     - [x] Prevent changing services via the API that were registered via config
     - [x] Update health check logic for services added at runtime
     - [x] Start/stop managed services on adding/removing services
     - [x] Add servers, `oauth` clients, API tokens to the database on adding services.
     - [x] Remove servers, `oauth` clients, API tokens after removing services
     - [x] Update the `oauth_no_confirm_list` list
- [x] Update documentation.
- [x] Update test.
- [x] Review the necessity of https://github.com/jupyterhub/jupyterhub/pull/4371

### Reference
https://github.com/jupyterhub/jupyterhub/issues/4370
https://discourse.jupyter.org/t/is-there-a-facility-for-dynamically-adding-services-to-the-hub/15352/3


minrk:
Thanks for opening this! Looking great so far, and I think this same simple `.from_config` approach will be extendable to roles in #3980

trungleduc:
Hi @minrk,
I'm having an issue with the database test, especially the migration from the 2.1.1 version 
(`test_db.py::test_upgrade[2.1.1]`).
The problem comes from using the ORM classes inside the migration script. In the `651f5419b74d_api_token_scopes.py` script, we invoke the `scopes.access_scopes` function which then will use the `orm.Service` class.
https://github.com/jupyterhub/jupyterhub/blob/f14be3df65cb1a3ac5fb466a7f161d8301e3f07d/jupyterhub/alembic/versions/651f5419b74d_api_token_scopes.py#L101-L104
Since I added a new column (`from_config`) to the `service` table but we're migrating from the 2.1.1 version, the database does not contain this column and it causes the failed test with the following error
```
 sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) no such column: services.from_config 
```
I can make the test pass by replacing the `scopes.access_scopes` call with a local version defined inside the migration script without using `orm.Service`, e.g.
```python
def access_scopes(oauth_client: orm.OAuthClient, db: Session):
    """Return scope(s) required to access an oauth client
    This is a clone of `scopes.access_scopes` without using
    the `orm.Service`
    """
    scopes = set()
    if oauth_client.identifier == 'jupyterhub':
        return frozenset()
    spawner = oauth_client.spawner
    if spawner:
        scopes.add(f'access:servers!server={spawner.user.name}/{spawner.name}')
    else:
        statement = f"SELECT * FROM services WHERE oauth_client_id = '{oauth_client.identifier}'"
        service = db.execute(text(statement)).fetchall()
        if len(service) > 0:
            scopes.add(f'access:services!service={service[0].name}')
    return frozenset(scopes)
```
What do you think about this solution?

minrk:
@trungleduc that solution in the migration script sounds reasonable!

Really, the migration steps shouldn't import anything from 'current' jupyterhub, which can change, they should bundle copies from their appropriate versions. It was short-sighted of me to do it that way, and we're seeing here exactly why!

jupyterhub/alembic/versions/3c2384c5aae1_add_from_config_column_to_the_services_.py
```diff
@@ -0,0 +1,30 @@
...
+from alembic import op
+
+
+def upgrade():
+    engine = op.get_bind().engine
+    tables = sa.inspect(engine).get_table_names()
+    if 'services' in tables:
+        op.add_column(
+            'services',
+            sa.Column('from_config', sa.Boolean, nullable=True, default=True),
```
manics, line 24: Is there a situation where NULL is required?

jupyterhub/alembic/versions/651f5419b74d_api_token_scopes.py
```diff
@@ -13,16 +13,35 @@
...
+    """
+    scopes = set()
+    if oauth_client.identifier == "jupyterhub":
+        return frozenset()
+    spawner = oauth_client.spawner
+    if spawner:
+        scopes.add(f"access:servers!server={spawner.user.name}/{spawner.name}")
+    else:
+        statement = f"SELECT * FROM services WHERE oauth_client_id = '{oauth_client.identifier}'"
+        service = db.execute(text(statement)).fetchall()
```
minrk, line 36: Rather than format strings, let's use [safe variable injection](https://docs.sqlalchemy.org/en/14/core/tutorial.html#using-textual-sql):

```suggestion
        statement =f"SELECT * FROM services WHERE oauth_client_id = :identifier"
        service = db.execute(text(statement), {"identifier": oauth_client.identifier}).fetchall()
```

trungleduc:
The `check links` failed test is not related to this PR.

trungleduc:
@minrk @manics What do you think about this implementation? Is there anything else I should beware of?

trungleduc:
Thank @minrk for the remark, I'll update the PR to take it into account.

docs/source/reference/services.md
```diff
@@ -174,6 +174,47 @@ c.JupyterHub.services = [
 In this case, the `url` field will be passed along to the Service as
 `JUPYTERHUB_SERVICE_URL`.
 
+## Adding or removing services at runtime
+
+Both Hub-Managed services and Externally-Managed services can be added at runtime by using JupyterHub’s REST API.
```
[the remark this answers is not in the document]
manics: I think limiting this to externally managed services makes sense. It also reduces the chance of needing to manually edit the DB, e.g. if a managed service disrupts the hub to the point where the API is unresponsive so you can't remove the service.

jupyterhub/app.py
```diff
@@ -2358,6 +2359,195 @@ async def init_api_tokens(self):
...
+            Defaults to `True`.
+            domain (Optional[str]): The domain of Hub session. Defaults to None.
+            host (Optional[str]): The host of Hub session. Defaults to None.
+
+        Returns:
+            Optional[Service]: The created service
+        """
+
+        if domain is None:
+            if self.domain:
```
manics, line 2430: This is always False

jupyterhub/apihandlers/services.py
```diff
@@ -28,6 +33,153 @@ def get(self, service_name):
...
+        new_service = await self.add_service(data)
+        self.write(json.dumps(self.service_model(new_service)))
+        self.set_status(201)
+
+    @needs_scope('admin:services')
+    async def delete(self, service_name: str):
+        service, orm_service = self.find_service(service_name)
+
+        if service is None:
+            raise web.HTTPError(404, f"Service {service_name} does not exists")
```
manics, line 110:
```suggestion
            raise web.HTTPError(404, f"Service {service_name} does not exist")
```

jupyterhub/app.py
```diff
@@ -2358,6 +2359,195 @@ async def init_api_tokens(self):
...
+                    " You can assign now assign roles via `JupyterHub.load_roles` configuration."
+                    " If you specify services in the admin role configuration, "
+                    "the Service admin flag will be ignored."
+                )
+                roles.update_roles(self.db, entity=orm_service, roles=['admin'])
+        else:
+            # Do nothing if the config file tries to modify a API-base service
+            # or vice versa.
+            if orm_service.from_config != from_config:
+                return
```
manics, line 2464: What's your thinking behind returning None instead of throwing an Exception? Should this log an self.log.error message?

jupyterhub/app.py
```diff
@@ -2358,6 +2359,195 @@ async def init_api_tokens(self):
...
+            if key not in traits:
+                raise AttributeError("No such service field: %s" % key)
+            setattr(service, key, value)
+            orm_service.update_column(key, value)
+
+        if service.api_token:
+            self.service_tokens[service.api_token] = service.name
+        elif service.managed:
+            # generate new token
+            # TODO: revoke old tokens?
```
manics, line 2491: @minrk this comment is copied from the existing code. Since we're refactoring services is there something we can do to resolve this TODO?

jupyterhub/app.py
```diff
@@ -2358,6 +2359,195 @@ async def init_api_tokens(self):
...
+            service.api_token = service.orm.new_api_token(note="generated at startup")
+
+        if service.url:
+            parsed = urlparse(service.url)
+            if parsed.port is not None:
+                port = parsed.port
+            elif parsed.scheme == 'http':
+                port = 80
+            elif parsed.scheme == 'https':
+                port = 443
```
manics, line 2501: else: raise an Exception?

docs/source/reference/services.md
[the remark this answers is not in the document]
trungleduc: Sorry for my late reply, I was busy with JupyterCon last week. And I also think that limiting the scope of runtime-created services is the way to go. I'll update the PR in this direction.

jupyterhub/app.py
trungleduc: Sorry, I do not catch that. Do you mean the check `if self.domain` is always `False`? I thought that `self.domain` is configurable.

jupyterhub/app.py
trungleduc: Throwing an exception here would prevent the startup of Jupyterhub. I added a log message to clarify the error. Returning `None` here helps the API handler to detect if the service is not created properly.

docs/source/reference/services.md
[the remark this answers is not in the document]
trungleduc: I implemented the scope check in https://github.com/jupyterhub/jupyterhub/pull/4381/commits/7704d2cf9b670b97e24268fe199597ee95513876

trungleduc:
> 

I implemented the scope check in https://github.com/jupyterhub/jupyterhub/pull/4381/commits/7704d2cf9b670b97e24268fe199597ee95513876

trungleduc:
Hi @minrk and @manics, could you do another round of review? I addressed all previous comments.

jupyterhub/oauth/provider.py
```diff
@@ -668,6 +668,15 @@ def add_client(
         self.db.commit()
         return orm_client
 
+    def remove_client(self, client_id):
+        """Remove a client by its id if it is existed."""
+        orm_client = (
+            self.db.query(orm.OAuthClient).filter_by(identifier=client_id).one_or_none()
+        )
+        if orm_client is not None:
+            self.db.delete(orm_client)
```
minrk, line 677: Add an `app_log.info` for deleting the client and `app_log.warning` if there was no client to delete.

jupyterhub/orm.py
```diff
@@ -423,6 +444,58 @@ def find(cls, db, name):
...
+        if hasattr(self, column_name):
+            column = getattr(Service, column_name)
+            if hasattr(column, 'foreign_keys'):
+                return len(column.foreign_keys) == 0
+            else:
+                return False
+        else:
+            return False
+
+    def update_column(self, column_name: str, value: any) -> bool:
```
minrk, line 467: I don't quite follow what the get/update_column bits are for. I don't really think we should have logic based on whether fields are foreign keys or not. Can you explain more about that? I think the `input=True` trait metadata on the Service might already serve the purpose here, but I'm not sure.

minrk (commented):
Sorry for the delayed review! I wrote up a review earlier, but failed to submit it. I think we're almost there.

trungleduc:
Thank! I'll update this PR next week.

jupyterhub/orm.py
trungleduc: I need to save service data coming from the API to the service table in order to recreate the service from the database in the next Hub startup. And the `input=True` trait metadata and the non-foreign key are not in sync. 
For example, the `oauth_client_id` is an `input` trait of the `Service` object:
```python
    oauth_client_id = Unicode(
        help="""OAuth client ID for this service.

        You shouldn't generally need to change this.
        Default: `service-<name>`
        """
    ).tag(input=True)
```
But in the `Service` ORM, it comes from an `OAuthClient` table so I don't need to save this column.
```python
    oauth_client_id = Column(
        Unicode(255),
        ForeignKey(
            'oauth_clients.identifier',
            ondelete='SET NULL',
        ),
    )
```
When I recreate the service from the service database at Hub startup, for the input trait that I can read from the table, I can set it directly to the service object, and for the remaining traits, I need to set it manually.

```python
        traits = service.traits(input=True)
        for key in traits:
            value = orm_service.get_column(key)
            if value is not None:
                setattr(service, key, value)

        if orm_service.oauth_client is not None:
            service.oauth_redirect_uri = orm_service.oauth_client.redirect_uri
            service.oauth_client_id = orm_service.oauth_client.identifier
            service.oauth_redirect_uri = orm_service.oauth_client.redirect_uri
```

minrk:
Coming back from vacation, will have a look soon.

martinRenou:
Thanks!

jupyterhub/app.py
minrk: Sorry for not clarifying `domain` is defined as domain of the hub, while `self.domain` is also the domain of the hub, so these two variables should have the same value. I don't think there's a reason to have these args, and instead always take the `if domain is None:` path for domain here and host below.

jupyterhub/app.py
minrk: I think it might be right for the Hub to halt. Invalid configuration errors like this should usually prevent loading the configuration. This method should not return without error if it did nothing, it should be up to the caller to catch that error, if it should at all.

jupyterhub/app.py
minrk: Let's do that in its own PR. It's finicky because we want to revoke the previous token, not _all_ tokens associated with the service, which means we have to unambiguously identify when that happens.

jupyterhub/app.py
minrk: > I want to avoid raising an exception in this function to keep Jupyterhub from crashing at startup because of badly configured services.

I don't think we should do that. If config is invalid and not loaded, it's a lot more surprising to see deployment and everything succeed and your config just have no effect than it is to see an error.

Given how JupyterHub is usually deployed, logs don't tend to get seen until _after_ a problem is noticed by real users. So admin errors like invalid service config should probably prevent startup, because otherwise bad config is more likely to go unnoticed long enough to affect _users_.

jupyterhub/app.py
```diff
@@ -3094,6 +3203,90 @@ async def update_last_activity(self):
...
+                self.log.error(
+                    "Cannot connect to %s service %s at %s. Is it running?",
+                    service.kind,
+                    service_name,
+                    service.url,
+                )
+                return False
+        return True
+
+    def toggle_service_health_check(self) -> None:
```
minrk, line 3272: Stateful `toggle` means if any call is missed (e.g. an unhandled error), we will stay always in the wrong state. Let's split this into explicit `start` and `stop` methods.

Alternately, even a context manager for `pause_service_health_check`, or even a boolean flag in the health check callback to skip checks that fire while editing, but without deleting/recreating the callback itself.

minrk (commented):
Finally had a chance to look through, thanks @trungleduc! I think this is almost there. I realize it's been a long time, so let me know if you'd like me to finish this up.

jupyterhub/app.py
trungleduc: Thank @minrk for your suggestion! And indeed it would be great if you can help me to push this PR over the line.

trungleduc:
Thank @minrk for your suggestion! And indeed it would be great if you can help me to push this PR over the line.

minrk:
Great! Applied my review and all looks good. Thank you!

PR #4387, opened 2023-03-06
fix inclusion of singleuser/templates/page.html in wheel

consideRatio:


ci/check_installed_data.py
```diff
@@ -12,7 +12,7 @@
 DATA_FILES_PATH = Path(DATA_FILES_PATH)
 assert DATA_FILES_PATH.is_dir(), DATA_FILES_PATH
 for subpath in (
-    "templates/page.html",
```
consideRatio, line 15: As this is just one template out of many, I adjusted this to reduce risk of confusion with singleuser/template/page.html.

minrk:
Thanks!

PR #4393, opened 2023-03-13
Selenium: updating test_oauth_page

mouse1203:
Changing is in test_oauth_page:
- rows with assertions for client_id and redirect_url have been rewritten

### DIFF
--- a/jupyterhub/tests/selenium/test_browser.py
+++ b/jupyterhub/tests/selenium/test_browser.py
@@ -2,6 +2,7 @@
 
 import asyncio
 import json
+import urllib.parse
 from functools import partial
 
 import pytest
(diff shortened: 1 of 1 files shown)


jupyterhub/tests/selenium/test_browser.py
```diff
@@ -921,8 +922,15 @@ async def test_oauth_page(
...
-    expected_redirect_url = app.base_url + f"servises/{service.name}/oauth_callback"
-    assert expected_client_id, expected_redirect_url in browser.current_url
+    expected_redirect_url = url_path_join(
+        app.base_url + f"services/{service.name}/oauth_callback"
+    )
+    # decode the URL
+    decoded_browser_url = urllib.parse.unquote(
+        urllib.parse.unquote(browser.current_url)
+    )
+    assert f"client_id=service-{expected_client_id}" in decoded_browser_url
```
minrk, line 932: Since we're asserting the values of url params, let's try parsing the URL using [urlparse](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse) and then [parse_qs](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.parse_qs) on the query, rather than running unquote twice on the full URL.

mouse1203: thank you, done

minrk:
Thanks!

PR #4394, opened 2023-03-14
Reduce size of jupyterhub image

alekseyolg:
Reduced the number of layers, optimized the assembly, reduced the size of the final image, removed the logs, removed unnecessary commands.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -73,6 +73,7 @@ RUN apt-get update \
     curl \
     gnupg \
     locales \
+    python-is-python3 \
     python3-pip \
     python3-pycurl \
     nodejs \
--- a/Dockerfile
+++ b/Dockerfile
@@ -56,7 +56,8 @@ ENV DEBIAN_FRONTEND=noninteractive \
     SHELL=/bin/bash \
     LC_ALL=en_US.UTF-8 \
     LANG=en_US.UTF-8 \
-    LANGUAGE=en_US.UTF-8
+    LANGUAGE=en_US.UTF-8 \
+    PYTHONDONTWRITEBYTECODE=1
 
 EXPOSE 8000
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,7 +27,7 @@ FROM $BASE_IMAGE AS builder
 ENV DEBIAN_FRONTEND=noninteractive
 WORKDIR /src/jupyterhub
 
-RUN apt-get update \
+RUN apt-get update -q \
  && apt-get install -yq --no-install-recommends \
     build-essential \
     ca-certificates \
@@ -66,7 +66,7 @@ LABEL org.jupyter.service="jupyterhub"
 
 WORKDIR /srv/jupyterhub
 
-RUN apt update \
+RUN apt update -q \
  && apt-get install -yq --no-install-recommends \
     ca-certificates \
     curl \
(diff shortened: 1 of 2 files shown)


alekseyolg:
Image size reduced by 10%
image

Dockerfile
```diff
@@ -39,65 +43,50 @@ RUN apt-get update \
...
-# compromise between needing to rebuild and maintaining
-# what needs to be part of the build
-COPY . /src/jupyterhub/
-WORKDIR /src/jupyterhub
+ && rm -rf /var/lib/apt/lists/* \
+ && python3 -m pip install --no-cache-dir --upgrade setuptools pip build wheel \
+ && npm install --global yarn \
+ # Build client component packages (they will be copied into ./share and
+ # packaged with the built wheel.)
+ && python3 -m build --wheel \
```
minrk, line 51: I will note that reducing layer count in the build stage only has a downside in build caching. It shouldn't affect the size of the resulting image. I don't think this change is an improvement.

Dockerfile
```diff
@@ -39,65 +43,50 @@ RUN apt-get update \
...
-
-RUN mkdir -p /srv/jupyterhub/
-WORKDIR /srv/jupyterhub/
-
-EXPOSE 8000
-
-LABEL maintainer="Jupyter Project <jupyter@googlegroups.com>"
-LABEL org.jupyter.service="jupyterhub"
+ && rm -rf /var/lib/apt/lists/* /var/log/* /var/tmp/* ~/.npm \
+ && find / -type d -name '*__pycache__' -prune -exec rm -rf {} \;
```
minrk, line 90: This is probably simplified with

```
ENV PYTHONDONTWRITEBYTECODE=1
```

prior to installation, without any pruning. This also has the benefit of preventing re-writing these files on every container startup.

Removing bytecode generally has a (small) negative impact on startup performance, which I guess is traded for pull-time performance - so the first startup+pull _might_ be quicker, but all startups after that are slower. This impact is doubled if PYTHONDONTWRITEBYTECODE is not set in the container, because all the missing bytecode files will be written out - never to be used - on _every start_.

Dockerfile
```diff
@@ -39,65 +43,50 @@ RUN apt-get update \
...
 
-FROM $BASE_IMAGE
+LABEL maintainer="Jupyter Project <jupyter@googlegroups.com>"
+LABEL org.jupyter.service="jupyterhub"
 
-USER root
+WORKDIR /srv/jupyterhub
 
-ENV DEBIAN_FRONTEND=noninteractive
+COPY --from=builder /src/jupyterhub/wheelhouse /tmp/wheelhouse
```
minrk, line 68: I think if we're optimizing image size, using [mounts as we do in other images](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/75f19bc56ed7d82d2c0a586286300fd56c41367a/images/hub/Dockerfile#L66-L67) instead of COPY prevents duplicating the wheels into the build image. That might have a bigger impact than the rest of the changes.

minrk (commented):
Thanks for the PR! I suggested some simplifications and further improvements.

alekseyolg:
If any of my colleagues have suggestions or comments on my proposal, I will be glad to know them.

PR #4399, opened 2023-03-22
add some more detail and examples to database doc

minrk:
include actual configuration samples for postgres/mysql


minrk:
looks like I didn't quite get the mysqlclient bit right

consideRatio (approved):
Looks reasonable to me - not verified though!

PR #4400, opened 2023-03-22
add Spawner.server_token_scopes config

minrk:
consistent behavior with oauth_client_allowed_scopes, where the _intersection_ of requested and owner-held permissions is granted, instead of failing

Enables different users to have different permissions in $JUPTYERHUB_API_TOKEN, either via callables or via requesting as much as you may want and only granting the subset.

Additionally, the !server filter can now be correctly applied to the server token, which means we _could_ update the default server role to more correctly have the `!server` filter instead of the `!user` filter, which grants it access to other servers owned by the same user

default behavior is unchanged

### DIFF
--- a/jupyterhub/roles.py
+++ b/jupyterhub/roles.py
@@ -52,7 +52,7 @@ def get_default_roles():
             'description': 'Post activity only',
             'scopes': [
                 'users:activity!user',
-                'access:servers!user',
+                'access:servers!server',
             ],
         },
         {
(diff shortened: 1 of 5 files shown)


minrk:
Yeah, I feel pretty good about it. WDYT about updating the default server permissions? We are preparing a major release right now, after all. For context, this would be the permissions I _would have_ made default, if it were possible when tokens could only be assigned roles, but it is possible now.

consideRatio (approved):

consideRatio:
Merging on test success!

consideRatio:
I've triggered a re-run of a selenium test failure, not sure if its an unrelated intermittent test issue

Issue #4415, opened 2023-04-13
Attribute lists in admin UI can be too wide


ryanlovett:
### Bug description

In /hub/admin, when you expand a user's attributes, if a user is a member of a lot of lists and/or if the group names are long, the user table runs off the page:
<img width="1382" alt="Screenshot 2023-04-13 at 1 31 34 PM" src="https://user-images.githubusercontent.com/10248125/231876149-ff7f4f16-2458-4e45-88c0-e089a4452072.png">

The admin must scroll off to the side and scan the string horizontally when looking for a specific group.

In [jsx/src/components/ServerDashboard/ServerDashboard.jsx](https://github.com/jupyterhub/jupyterhub/blob/main/jsx/src/components/ServerDashboard/ServerDashboard.jsx#L240) I see that attributes are assembled together in a comma-separated string.

#### Expected behaviour

I think it'd be better if the attribute list could be displayed as an unordered list. It would make it easier to quickly scan for values and preserves the semantics of "a list of things".

#### Actual behaviour

Long attribute lists are concatenated into a very long string.

### How to reproduce

1. Visit /hub/admin.
2. Choose a user who is a member of either a lot of groups, groups with long names, or both.
3. Click the drop-down arrow next to their username.

### Your personal set up

z2jh, jupyterhub 4.0.0b2

<details><summary>Full environment</summary>

```
Package                          Version
-------------------------------- ---------
aiohttp                          3.8.4
aiosignal                        1.3.1
alembic                          1.10.2
async-generator                  1.10
async-timeout                    4.0.2
attrs                            22.2.0
bcrypt                           4.0.1
certifi                          2022.12.7
certipy                          0.1.3
cffi                             1.15.1
charset-normalizer               3.1.0
cryptography                     39.0.2
escapism                         1.0.1
frozenlist                       1.3.3
greenlet                         2.0.2
idna                             3.4
Jinja2                           3.1.2
jsonschema                       4.17.3
jupyter-telemetry                0.1.0
jupyterhub                       4.0.0b2
jupyterhub-canvasoauthenticator  0.1
jupyterhub-firstuseauthenticator 1.0.0
jupyterhub-hmacauthenticator     1.0
jupyterhub-idle-culler           1.2.1
jupyterhub-kubespawner           4.3.0
jupyterhub-ldapauthenticator     1.3.2
jupyterhub-ltiauthenticator      1.4.0
jupyterhub-nativeauthenticator   1.1.0
[...]

PR #4417, opened 2023-04-14
admin page: improve display of long lists (groups, etc.)

manics:
Closes https://github.com/jupyterhub/jupyterhub/issues/4415

image

Ideally this would be a plain vertical list without indices, but I don't have enough React knowledge to do this. Also I'm not sure this is better than having everything in one line due to the wasted space.


ryanlovett:
Thanks @manics!

> Ideally this would be a plain vertical list without indices, but I don't have enough React knowledge to do this.

I agree. While I think the vertical arrangement would be better for my purposes, I do think the word wrapping is still an improvement over the current display.

> Also I'm not sure this is better than having everything in one line due to the wasted space.

I think if everyone's attributes were displayed by default, then it would waste a lot of vertical space for the case where a user is in a lot of groups and the groups are in a vertical list. Since all of the per-user tables are not visible by default, and you have to click each user's down arrow to view the attributes, it wouldn't have wasted a lot of space overall.

I like your suggestion about a text background rather than commas. There is also a [horizontal listgroup](https://react-bootstrap.github.io/components/list-group/#horizontal) display.

manics:
After disappearing into a React wormhole I've managed to get this:
image

`<ListGroup horizontal>` didn't work due to some interaction with the table component, but I figure out how to create a new component `RowListItem`. I've also vendored the `ReactObjectTableViewer` since it's pretty small, and modified it to handle React components, previously it only handled JSON and plain text fields.

ryanlovett:
@manics That looks great!

PR #4420, opened 2023-04-17
Begin setup of playwright tests

mouse1203:
To simplify web testing and avoid a problem with async waiting of web elements as Playwright has own async API.

The next files are changed:
test.yml - added a configuration for Playwright, rename selenium to browser
pytest.ini - rename marker for browser testing to "browser"
setup.py - added "playwright"

New files:
tests/browser/conftest.py
tests/browser/test_browser.py

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -233,6 +233,10 @@ jobs:
         if: matrix.selenium
         run: echo "PYTEST_ADDOPTS=$PYTEST_ADDOPTS -m selenium" >> "${GITHUB_ENV}"
 
+      - name: Install playwright module
+        run: |
+          pip install playwright
+
       - name: Ensure browsers are installed
         run: python -m playwright install --with-deps
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -99,9 +99,7 @@ jobs:
             noextension: noextension
             subset: singleuser
           - python: "3.11"
-            selenium: selenium
-          - python: "3.11"
-            playwright: playwright
+            browser: browser
           - python: "3.11"
             main_dependencies: main_dependencies
 
@@ -231,21 +229,17 @@ jobs:
               DB=postgres bash ci/init-db.sh
           fi
 
-      - name: Configure selenium tests
-        if: matrix.selenium
-        run: echo "PYTEST_ADDOPTS=$PYTEST_ADDOPTS -m selenium" >> "${GITHUB_ENV}"
+      - name: Configure browser tests
+        if: matrix.browser
+        run: echo "PYTEST_ADDOPTS=$PYTEST_ADDOPTS -m browser" >> "${GITHUB_ENV}"
 
       - name: Install playwright module
         run: |
           pip install playwright
 
       - name: Ensure browsers are installed for playwright
-        if: matrix.playwright
+        if: matrix.browser
         run: python -m playwright install --with-deps
-
-      - name: Configure playwright tests
-        if: matrix.playwright
-        run: echo "PYTEST_ADDOPTS=$PYTEST_ADDOPTS -m playwright" >> "${GITHUB_ENV}"
 
       - name: Run pytest
         run: |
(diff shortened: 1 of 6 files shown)


.github/workflows/test.yml
```diff
@@ -233,6 +235,18 @@ jobs:
         if: matrix.selenium
         run: echo "PYTEST_ADDOPTS=$PYTEST_ADDOPTS -m selenium" >> "${GITHUB_ENV}"
 
+      - name: Install playwright module
+        run: |
+          pip install playwright
```
minrk, line 240: `playwright` can be added to `dev-requirements.txt`

.github/workflows/test.yml
```diff
@@ -100,6 +100,8 @@ jobs:
             subset: singleuser
           - python: "3.11"
             selenium: selenium
```
minrk, line 102: let's use `browser: browser` instead of an implementation-specific name

jupyterhub/tests/browser/test_browser.py
```diff
@@ -0,0 +1,31 @@
...
+    """filling the login form by user and pass_w parameters and iniate the login"""
+
+    await browser.get_by_label("Username:").click()
+    await browser.get_by_label("Username:").fill(username)
+    await browser.get_by_label("Password:").click()
+    await browser.get_by_label("Password:").fill(password)
+    await browser.get_by_role("button", name="Sign in").click()
+
+
+async def test_open_login_page(app, browser):
```
minrk, line 24: For migration, let's _remove_ each selenium test we add here, so that it's easier to keep track of the migration progress.

pytest.ini
```diff
@@ -7,7 +7,7 @@
 asyncio_mode = auto
 
 # jupyter_server plugin is incompatible with notebook imports
-addopts = -p no:jupyter_server -m 'not selenium' --color yes --durations 10 --verbose
+addopts = -p no:jupyter_server -m 'not selenium and not playwright' --color yes --durations 10 --verbose
```
minrk, line 10: also here: _rename_ selenium to more generic `browser` rather than adding a new playwright mark

minrk (commented):
Great to get this started!

pytest.ini
mouse1203: thank you, the changes have been done

.github/workflows/test.yml
mouse1203: thank you, the changes have been done

.github/workflows/test.yml
```diff
@@ -233,6 +235,18 @@ jobs:
         if: matrix.selenium
         run: echo "PYTEST_ADDOPTS=$PYTEST_ADDOPTS -m selenium" >> "${GITHUB_ENV}"
```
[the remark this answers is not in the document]
mouse1203: thank you, the changes have been done

.github/workflows/test.yml
mouse1203: thank you, the changes have been done in setup.py under test section

jupyterhub/tests/browser/conftest.py
```diff
@@ -0,0 +1,14 @@
+import pytest
+from playwright.async_api import async_playwright
+
+
+@pytest.fixture()
+async def browser():
+    # browser_type in ["chromium", "firefox", "webkit"]
```
manics, line 7: Is this just a reminder of which browsers are available? Or are you intending to support all three?

jupyterhub/tests/browser/test_browser.py
mouse1203: this simple case is absent in the selenium scope, but I will follow your approach with removing in the next PR, it is a good point

.github/workflows/test.yml
minrk: Then this step can be removed.

jupyterhub/tests/browser/conftest.py
minrk: In the future, we could make this controllable by an environment variable or CLI arg

.github/workflows/test.yml
mouse1203: thank you, those rows have been removed, changes have been added

PR #4423, opened 2023-04-19
add remaining redirects for docs reorg

### LINKED ISSUE #4333: How do we work with redirects in the repo?
[...] to make a decision on what to do with regards to how we work with the redirects in this repo for now. If experience is gained and you start feeling confident on a practice, we try go for that more systematically in other repos long term.

I think the options on how to work with the redirects are:

1. Fully manual as now, where we figure out ourselves if we need to add a redirect
2. Mostly manual, where we can go fully manual but also suggest that `make rediraffecheckdiff` can be used to get informed on what redirects we should make.
3. Semi automated, where we suggest using `make rediraffewritediff` to update a file (not overwrite right?)
4. Mostly automated, where we have a github workflow triggered when docs change to inspect if a redirect needs to be setup and otherwise fails to let us know.
5. Almost fully automated, where we also push a commit with the relevant redirects.

If I understand this correctly, we need to consider this ourselves no matter what because the tools can't figure it out fully in all cases. That makes me think we shouldn't go for option 5. Maybe option 2-4? No matter what, I'm a big :+1: for going for something you feel like going for!

If a redirects.txt file is added, its a big plus I think if its possible to have a comment in its top helping anyone opening and wondering what its about to learn what its about, for example by "For more information, see rediraffe configuration in docs/source/conf.py" or similar.

_Originally posted by @consideRatio in https://github.com/jupyterhub/jupyterhub/pull/4331#discussion_r1092331738_

### LINKED ISSUE #4376: Add missing redirects for files moves since 3.1
We should make sure to run `make rediraffecheckdiff` with `rediraffe_branch = "3.1.0"` to ensure we have redirects for all the moved files in the diataxis reorg. We have several redirects in place, but definitely not all of them. In particular, the index files from the old sections have no targets (cf #4374).

Adding this out of #4364 as a discrete task which must be done prior to releasing 4.0 (#4350)

minrk:
- ran `make rediraffecheckdiff` with rediraffe_branch=3.1.1
- add a marker in the redirect file indicating that redirects are up-to-date for 4.0
- add redireaffecheck to CI, checking against:
  - current PR (should be the only one that fails going forward, but _removing_ longstanding redirects wouldn't be caught by this)
  - latest tag, so `stable` links always have verified redirects
  - a fixed version (3.0.0), so older links always have verified redirects

closes #4376
closes #4333


docs/source/redirects.txt
```diff
@@ -1,23 +1,44 @@
...
+
 "reference/api-only.md" "howto/api-only.md"
 "reference/config-ghoauth.md" "howto/configuration/config-ghoauth.md"
 "reference/config-proxy.md" "howto/configuration/config-proxy.md"
-"admin/log-messages.md" "howto/log-messages.md"
+"reference/database.md" "explanation/database.md"
+"reference/oauth.md" "explanation/oauth.md"
 "reference/proxy.md" "howto/proxy.md"
 "reference/templates.md" "howto/templates.md"
-"quickstart-docker.md" "tutorial/quickstart-docker.md"
```
minrk, line 12: I sorted these alphabetically  by the origin, grouped by top-level folder, so they were easier for me to skim

consideRatio (approved):

consideRatio (changes requested):
Hmm hmmm, I suspect that the failure in CI is relevant. Do you think so as well?

We currently have `- uses: actions/checkout@v3`, that may not have captured all historical info of relevance.

```
Running Sphinx v6.1.3
making output directory... done
loading pickled environment... done
fatal: ambiguous argument 'main': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'

Exception occurred:
  File "/opt/hostedtoolcache/Python/3.9.[16](https://github.com/jupyterhub/jupyterhub/actions/runs/4742862923/jobs/8421659004#step:8:17)/x64/lib/python3.9/subprocess.py", line 528, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command 'git -C /home/runner/work/jupyterhub/jupyterhub/docs/source diff --name-status --diff-filter=R main' returned non-zero exit status 128.
The full traceback has been saved in /tmp/sphinx-err-_i7u8kue.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
make: *** [Makefile:[27](https://github.com/jupyterhub/jupyterhub/actions/runs/4742862923/jobs/8421659004#step:8:28): rediraffecheckdiff] Error 2
Error: Process completed with exit code 2.
```

It seems to me that there is a `git` command reporting that `main` wasn't recognized:

```
git -C /home/runner/work/jupyterhub/jupyterhub/docs/source diff --name-status --diff-filter=R main
```

Reading up on `-C` it is to emulate running `git` from another directory, so this is like running:

```
cd docs/source
git diff --name-status --diff-filter=R main
```

and then getting an error like:

```
fatal: ambiguous argument 'main': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

Due to this, I figure we do need something like this:

```yaml
      - uses: actions/checkout@v3
        with:
          # chartpress requires git history to set chart version and image tags
          # correctly
          fetch-depth: 0
```

consideRatio:
Hmmm, now run against origin/main, where we assume origin to be jupyterhub/jupyterhub, but if we push to a fork's branch, the origin is for example consideratio/jupyterhub, and my fork's main then seen as origin/main won't be updated necessarily. Due to that, we may find ourselves with test docs failures in forks etc now.

I think maybe we should tweak the call to actions/checkout further, or make sure that the workflow not run on forks.

consideRatio (approved):

consideRatio (changes requested):
Argh a final issue: `git describe --tags --abbrev 0` gives us...

```
fatal: Not a valid object name 0
```

consideRatio (approved):
Pweh! This LGTM now and is ready to be merged I think. I'll go for it!

minrk:
@consideRatio thanks for working on this! I added one more commit to use the PR base ref for the PR diffcheck, so it should always resolve, no matter what is checked out or what repo is used.

minrk:
I believe this is the last thing before making 4.0, so I'll do a changelog PR tomorrow or Friday and make the actual release!

.github/workflows/test-docs.yml
```diff
@@ -72,3 +77,29 @@ jobs:
...
+      - name: check redirects since last release
+        if: github.repository == 'jupyterhub/jupyterhub'
+        run: |
+          cd docs
+          export REDIRAFFE_BRANCH=$(git describe --tags --abbrev=0)
+          make rediraffecheckdiff
+
+      # longer-term redirect check (fixed version) for older links
+      - name: check redirects since 3.0.0
+        if: github.repository == 'jupyterhub/jupyterhub'
```
consideRatio, line 101: @minrk the `${{ }}` part isn't needed for `if` statements, but will be fine to have.

I added two more `github.repository == 'jupyterhub/jupyterhub'` checks below that needs to be updated as well. Maybe they aren't critical? Or are they?

I think maybe we will end up failing now when we check against `main`, but I'm not confident. I figure maybe we just checkout a branch or similar, and `main` isn't a checked out local branch, so we would be forced to compare with `origin/main` - or?

Hmmm... This is quite tricky to understand I think.

consideRatio: I think we should lean towards having `if` conditions for all three, not just the first. And if we check against `main` by default in conf.py, maybe we need to do `export REDIRAFFE_BRANCH=origin/main` in the first step in this CI job.

.github/workflows/test-docs.yml
```diff
@@ -72,3 +77,29 @@ jobs:
...
```
consideRatio, line 101:
```suggestion
        if: github.event_name == 'pull_request'
        run: |
          cd docs
          export REDIRAFFE_BRANCH=origin/${{ github.base_ref }}
          make rediraffecheckdiff

      # this should check currently published 'stable' links for redirects
      - name: check redirects since last release
        if: github.event_name == 'pull_request'
        run: |
          cd docs
          export REDIRAFFE_BRANCH=$(git describe --tags --abbrev=0)
          make rediraffecheckdiff

      # longer-term redirect check (fixed version) for older links
      - name: check redirects since 3.0.0
        if: github.event_name == 'pull_request'
```

.github/workflows/test-docs.yml
minrk: Yes, I think if for all makes sense - for the pr it resolves the ref regardless of origin (this run makes no sense for non PRs) and the other two because forks don't always have up-to-date tags.

I think "main" is the right default in the file (as it was before) because that will be used only by local runs, where we don't want to assume how people manage their remotes.

minrk: > I think maybe we will end up failing now when we check against main, but I'm not confident.

To be clear: we won't ever check against main implicitly on CI. It's always a tag or the explicit base ref (which is usually main, but always guaranteed to exist and be correct when that is the case).

.github/workflows/test-docs.yml
minrk: I think your conditions were correct the first time - the repo check helps ensure the tags are defined, which is relevant for the latter two checks. The PR check is only for the base ref, which doesn't exist if it's not a PR.

minrk: Updated the comment to explain the reason for the two different `if`s (and removed the unnecessary `${{}}`, thanks!)

consideRatio (approved):

Issue #4425, opened 2023-04-20
Add version-specific upgrading notes to 'upgrading' doc


minrk:
### Proposed change

Add a dedicated 'upgrading to X.Y' section for each release to `docs/source/howto/upgrading.md` that covers the relevant changes and steps (e.g. upgrade-db). We should usually need these only for major versions.

The changelog _contains_ this information, but because it contains all changes, it is extremely noisy and I don't really expect anyone to read it. Focusing a section on "these are the things you should know to upgrade" should help people see what they need to see.

### Alternative options

Keep this info mostly in the changelog as we have done, which I think is a reasonable place to find it, but since it serves many more audiences answering a single "what changed?" question, it's harder to answer more specific questions like upgrading clearly.

### Who would use this feature?

- Users upgrading jupyterhub checking what they need to know first
- Maintainers preparing documentation for a release

manics:
👍  We did something similar for Z2JH https://z2jh.jupyter.org/en/stable/administrator/upgrading/upgrade-1-to-2.html

minrk:
we have this for v4 and v5

PR #4426, opened 2023-04-20
add upgrade note for 4.0 to changelog

minrk:
I also opened #4425 because I think the changelog isn't really the best place for upgrading instructions, but that's a bigger task.

### DIFF
--- a/docs/source/reference/changelog.md
+++ b/docs/source/reference/changelog.md
@@ -14,8 +14,7 @@ command line for details.
 
 4.0 is a major release, but a small one.
 
-:::{note}
-**Upgrade note**
+:::{admonition} Upgrade note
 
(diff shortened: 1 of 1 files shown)


minrk:
linkcheck should be fixed by #4427

docs/source/reference/changelog.md
```diff
@@ -13,6 +13,14 @@ command line for details.
 ### 4.0.0 - 2023-04-20
 
 4.0 is a major release, but a small one.
+
+:::{note}
+**Upgrade note**
```
consideRatio, line 18:
```suggestion
:::{admonition} Upgrade note
```

Like this, we get a note section with the title "Upgrade note" as compared to having it titled "Note" followed by the bold text "Upgrade note" inside it.

consideRatio: image

Compared to

image

consideRatio: Either seems fine, merge directly if you want

consideRatio (approved):

PR #4427, opened 2023-04-20
Fix some public URL links within the docs

minrk:
there shouldn't be any links to jupyterhub.readthedocs.io

several of these links were to stale URLs, but linkcheck only failed when the 4.0.0 became 'stable'

### DIFF
--- a/docs/source/contributing/setup.md
+++ b/docs/source/contributing/setup.md
@@ -130,8 +130,8 @@ configuration:
 jupyterhub -f testing/jupyterhub_config.py
 ```
 
-The default JupyterHub [authenticator](https://jupyterhub.readthedocs.io/en/stable/reference/authenticators.html#the-default-pam-authenticator)
-& [spawner](https://jupyterhub.readthedocs.io/en/stable/api/spawner.html#localprocessspawner)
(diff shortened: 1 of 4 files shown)


consideRatio (approved):

consideRatio:
I'd label this as `documentation`, but I see how you can think its `maintenance`. It makes either docs or maintenance section in the changelog a bit cleaner in a way.

Sometimes when debugging, I end up going through the list of maintenance PRs as those may have touched something that could have caused a bug, while I wouldn't scan the documentation section.

I don't think its important to enforce one practice here though - so no action point required

minrk:
Thanks for your perspective! I wouldn't have thought of looking through the maintenance PR list for debugging, since I don't think I've ever actually looked at one. I've made this one documentation. I don't have a good answer, other than I don't believe our short list of classifications really encompasses every kind of change, and I don't really like lumping all maintenance of the docs infrastructure in with actual content changes.

consideRatio:
I've never gone through this repo's merged PRs to catch a regression, but for other projects where titles may have omitted a change of relevance - that was actually breaking. :)

PR #4429, opened 2023-04-23
docs: fix missing redirects for api to reference/api

consideRatio:
I saw a broken link in oauthenticators documentation:

image

### DIFF
--- a/docs/source/redirects.txt
+++ b/docs/source/redirects.txt
@@ -39,6 +39,15 @@
 "reference/server-api.md" "tutorial/server-api.md"
 "reference/websecurity.md" "explanation/websecurity.md"
 
+"api/app.md" "reference/api/app.md"
+"api/auth.md" "reference/api/auth.md"
+"api/index.md" "reference/api/index.md"
+"api/proxy.md" "reference/api/proxy.md"
+"api/service.md" "reference/api/service.md"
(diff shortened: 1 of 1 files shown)


manics:
Do our docs CI tests automatically check the destination of the redirects is valid, or is this something a reviewer should check?

consideRatio:
> Do our docs CI tests automatically check the destination of the redirects is valid, or is this something a reviewer should check?

The rediraffe extension makes checks of redirects automatically during docs build, specifically that the new destination exist I think.

@minrk also developed a check in the test-docs workflow recently to help catch failure to setup redirects for files that were found to be relocated or similar, but apparently it didn't catch this. It would fail to detect something for example if a file was largely remade, then it could have been seen as a deleted file and an entirely new file instead.

I went ahead and manually tested for all redirects configured that:

1. There was a page before at the redirect location: https://jupyterhub.readthedocs.io/en/3.0.0/api/<page>.html was found
2. There was a page we got redirected to after: https://jupyterhub--4429.org.readthedocs.build/en/4429/api/<page>.html redirected correctly

minrk:
Thanks! I wonder why the checkdiff didn't find these.

I believe rediraffecheckdiff does _not_ validate targets (I introduced some typos while working on the redirect PR and it didn't complain, IIRC).

PR #4431, opened 2023-04-25
Migrate some tests from selenium to playwright

mouse1203:
Continue migration from Selenium to Playwright:
Added Playwright tests which are covered Login, Spawning, Home and Token pages instead of Selenium cases which are covered the same scope


minrk:
Great!

PR #4432, opened 2023-04-26
add HUNT into research institutions

matuskosut:
Hi all,

I would like to add HUNT into the list of research institutions. We are providing virtual labs with Jupyterhub deployments for students, researchers, and institutions working with sensitive data in Norway but also internationally. You can find more info on our profile https://github.com/huntdatacenter#welcome-to-hunt-cloud or feel free to ask me.

Is there other specific requirements for adding an institution to the list?

### DIFF
--- a/docs/source/faq/institutional-faq.md
+++ b/docs/source/faq/institutional-faq.md
@@ -66,7 +66,7 @@ Here is a sample of organizations that use JupyterHub:
 - **Universities and colleges**: UC Berkeley, UC San Diego, Cal Poly SLO, Harvard University, University of Chicago,
   University of Oslo, University of Sheffield, Université Paris Sud, University of Versailles
 - **Research laboratories**: NASA, NCAR, NOAA, the Large Synoptic Survey Telescope, Brookhaven National Lab,
-  Minnesota Supercomputing Institute, ALCF, CERN, Lawrence Livermore National Laboratory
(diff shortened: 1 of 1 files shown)


minrk:
I think it's fine to add you. It seems like the 'institution' ought to be NTNU instead of HUNT, is that right?

:wave: from Oslo

matuskosut:
Hi @minrk 

we were aiming for recognition of HUNT. I see it similar to _Minnesota Supercomputing Institute_ being listed on behalf of _University of Minnesota_. HUNT is well known in health research as one of the largest cohort studies in the world. And here we come into the play by providing virtual labs for working with sensitive data (health data being one of the most sensitive) to many institutions in Norway and abroad. As HUNT/HUNT Cloud we are also iso certified and typically we do not provide these labs for general research (non-sensitive data).

Do you travel up north sometimes? You should definitely stop by for a chat if you happen to be in Trondheim. We would be happy to share a bit about our work.

manics:
@matuskosut that sounds really cool! I hadn't heard of HUNT before but I'm interested to hear more about how you're using JupyterHub (and everything else) with sensitive data- I'm in the UK working with trusted research environments.

matuskosut:
@manics cool, we are also in on TREs! PMed you OOB.

minrk:
Fair enough, thanks!

I don't get up often, but have visited a few years ago.

PR #4438, opened 2023-05-01
Remove old admin JS code

yuvipanda:
We have a new react based admin, and this JS was just loading and doing nothing.


minrk:
Thanks!

PR #4442, opened 2023-05-05
add config to /hub/info endpoint

### LINKED ISSUE #4406: Provide an API endpoint to get current effective traitlets configuration
When extending JupyterHub via services, it is quite difficult to figure out what exactly is the effective configuration of the hub. For example, there isn't a way to know what image is going to be launched, based on traitlets configuration! It would be extremely helpful to have an API endpoint that just dumps out traitlets values. This would drastically increase the ability of things like [the jupyterhub configurator](https://github.com/yuvipanda/jupyterhub-configurator) which currently can't tell what the default image is, or the resource limits, etc.

However, dumping all traitlets info is probably a bad idea, because it can contain sensitive info. So I propose we have an allow-list of traitlet values that will be exposed *readonly* via this config API. So it would be something like:

```python
c.JupyterHub.config_api_allowed_traitlets = [
  "KubeSpawner.image",
  "KubeSpawner.profile_list",
  "Spawner.memory_limit"
]
```

And then if you hit a `/api/config` endpoint, you'll get back the current calculated values of all these allowed traitlets.

minrk:
adds `JupyterHub.api_include_config` list for selecting which config should be included, which includes nothing by default.

instead of adding a new `/api/config` endpoint, I elected to add a field to the existing `/api/info` endpoint, which already tells things about the Hub, and config seems like it fits with what's already there.

Key issues:

- unrecognized classes (e.g. kubespawner Reflectors) will get values from config, not instances, so no default values will be reported, since JupyterHub doesn't know how to find the instances. A discoverability system for configured instances could help here.
- instantiating Spawners with a mock user may be fiddly, depending on assumptions in the Spawner. This should usually be okay, though, and handled semi-gracefully where failure can occur.

closes #4406

WDYT @yuvipanda?

### DIFF
--- a/jupyterhub/_config.py
+++ b/jupyterhub/_config.py
@@ -1,0 +1,140 @@
+"""Utilites for getting the resolved Hub config
+
+Used in the /api/info endpoint
+"""
+
+from dataclasses import dataclass, field
+
+from traitlets.log import get_logger
+
+from jupyterhub import crypto
+from jupyterhub.user import User
+
+
+@dataclass
+class _MockORMUser:
+    """Mock orm.User API, so we can instantiate a Spawner without a real user"""
+
+    id: int = 0
+    name: str = "username"
+    admin: bool = False
+    encrypted_auth_state: bytes | None = None
+    state: dict = field(default_factory=dict)
+
+    # relationships likely to be used
+    orm_spawners: dict = field(default_factory=dict)
+    groups: list = field(default_factory=list)
+
+    # other fields technically exist, but shouldn't be accessed by Spawners
+    # these likely won't be exposed when we separate Spawners from db access
+    created = None
+    last_activity = None
+    cookie_id: str = ""
+    roles: list = field(default_factory=list)
+
+
+class _MockUser(User):
+    """Mock orm.User API, so we can instantiate a Spawner without a real user"""
+
+    def _new_orm_spawner(self, servername=""):
+        pass
+
+
+class _MockDB:
+    """Mock out deprecated db access"""
+
+    def add(self, obj):
+        pass
+
+    def commit(self):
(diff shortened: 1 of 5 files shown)


jupyterhub/_config.py
```diff
@@ -0,0 +1,141 @@
...
+    full_class_filters = set()
+    class_trait_filters = {}
+    for f in trait_filters:
+        if "." not in f:
+            full_class_filters.add(f)
+        else:
+            cls_name, _, trait_name = f.partition(".")
+            if cls_name not in full_class_filters:
+                cls_filters = class_trait_filters.setdefault(cls_name, set())
+                cls_filters.add(trait_name)
```
sgaist, line 77: Nitpick: single letter variable. I would just use `filter`

jupyterhub/_config.py
```diff
@@ -0,0 +1,141 @@
...
+
+    all_classes = full_class_filters | set(class_trait_filters.keys())
+
+    # track any sections we didn't look at via instances
+    unused_sections = set() | all_classes
```
sgaist, line 82: Isn't `all_classes` guaranteed to be a set ?
`full_class_filters` is one even if empty and if so, `all_classes` is a set of `class_trait_filters.keys`.

sgaist (approved):
Only minor things, otherwise, it looks good.

yuvipanda:
I'm now convinced that traitlets are too dynamic for us to realistically do this, and vote we close this one :D I'm not sure if I'll even be able to fully trust the answers that come out of this API, given how dynamic traitlets can be.

minrk:
@yuvipanda since you requested the feature and don't plan to use it, I'm happy to close. We can always pick it up again!

PR #4443, opened 2023-05-05
Update jsx dependencies as much as possible

manics:
Updates dependencies in `jsx/` to the latest possible without requiring code changes (React 18 has breaking changes).

`ubuntu:22.04` installs Nodejs 12 which is not supported by many of the updated dependencies, however Nodejs 12 is fine in the runtime image since that only runs configurable-http-proxy.

For mysterious reasons using the upgraded Node led to a large increase in build times, so I've refactored the Dockerfile to build the JupyterHub wheel on the [native architecture only](https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/) which brings the build time below what it was previously.


Dockerfile
```diff
@@ -22,36 +22,73 @@
...
+# The JupyterHub wheel is pure Python so can be built once on any architecture
+# TODO: re-enable BUILDPLATFORM instead of forcing to linux/amd64
+#FROM --platform=${BUILDPLATFORM:-linux/amd64} base-builder AS jupyterhub-builder
+FROM --platform=linux/amd64 base-builder AS jupyterhub-builder
+
+# Need ARG to be able to view values
+# https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope
+ARG BUILDPLATFORM
+ARG TARGETPLATFORM
+RUN echo "BUILDPLATFORM=${BUILDPLATFORM} TARGETPLATFORM=${TARGETPLATFORM}"
```
manics, line 70: on linux/arm64:
```
2023-05-06T00:00:12.2369673Z #24 [linux/arm64 jupyterhub-builder 1/2] RUN echo "BUILDPLATFORM=linux/amd64 TARGETPLATFORM=linux/arm64"
2023-05-06T00:00:12.2370126Z #0 0.088 BUILDPLATFORM=linux/amd64 TARGETPLATFORM=linux/arm64
2023-05-06T00:00:12.3870583Z #24 DONE 0.1s
```
as expected..... so the JupyterHub build should be fast, but it's as if it's still being done on emulated arm64 :disappointed:

minrk (approved):
Great! Love the quicker builds.

Have you verified that the bootstrap 5 update hasn't introduced any visual issues on the admin page? That wouldn't be caught by CI. If you've checked, feel free to merge.

manics:
Yes, I've checked!

consideRatio:
You are amazing @manics, thank you for all your work!!

PR #4444, opened 2023-05-06
Remove Dockerfile.alpine

manics:
It's obviously not maintained:

> `ARG JUPYTERHUB_VERSION=1.3.0`


minrk:
Good call! I think we should restore this file as part of https://github.com/jupyterhub/team-compass/issues/474

PR #4452, opened 2023-05-12
Fix docs on setting named_server_limit_per_user config

NarekA:
Since the value is a property, it should be patched as a property.


NarekA:
Looks like this breaks the current deployment, should leave it as-is

PR #4454, opened 2023-05-16
Add CERN to Gallery of JupyterHub Deployments

goseind:
As the team managing the JupyterHub deployments at CERN, we would gladly add it to this list.


consideRatio (approved):
Nice :heart: :tada:

Issue #4447, opened 2023-05-10
Move ready_event to Spawner class


kreuzert:
Hi everyone,
we're running a JupyterHub with our own Spawner Class. We use an overwritten `.progress()` function to show the user the current status of their starting process. We would like to change the `ready_event` shown to the user, once JupyterLab is started.

### Proposed change
Move `ready_event` from [apihandlers/users.py](https://github.com/jupyterhub/jupyterhub/blob/abe1136cba84a1dc30d7c7b296a5bc2c21271c22/jupyterhub/apihandlers/users.py#L714) to [spawner.py](https://github.com/jupyterhub/jupyterhub/blob/abe1136cba84a1dc30d7c7b296a5bc2c21271c22/jupyterhub/spawner.py#L1243) as property. Maybe one should do the same for the `failed_event`, too.

### Who would use this feature?

An administrator could give additional information (like in all other events in `Spawner.progress()` ), e.g. adding a timestamp to the html_message.

### (Optional): Suggest a solution

I think different ways of implementation are possible. I will show an approach similar to `Spawner.options_form`. One could also use an implementation like `pre_spawn_hook`/`post_stop_hook`.  
  
Maybe the `failed_event` would be useful, too. I've added it to my suggested solution. You can still overwrite the error message via `html_message = getattr(exc, "jupyterhub_html_message", "")` , but you can customize the default message.
  
```
diff -Naurx .git -x node_modules -x share -x __pycache__ ../jupyterhub/jupyterhub/apihandlers/users.py ../jupyterhub-patched/jupyterhub/apihandlers/users.py
--- ../jupyterhub/jupyterhub/apihandlers/users.py	2023-05-10 16:15:15.301756624 +0200
+++ ../jupyterhub-patched/jupyterhub/apihandlers/users.py	2023-05-10 16:36:33.111613061 +0200
@@ -19,8 +19,6 @@
     isoformat,
     iterate_until,
     maybe_future,
-    url_escape_path,
-    url_path_join,
 )
 from .base import APIHandler
 
@@ -710,15 +708,8 @@
         # - spawner not running at all
         # - spawner failed
         # - spawner pending start (what we expect)
-        url = url_path_join(user.url, url_escape_path(server_name), '/')
-        ready_event = {
-            'progress': 100,
-            'ready': True,
-            'message': f"Server ready at {url}",
-            'html_message': 'Server ready at <a href="{0}">{0}</a>'.format(url),
-            'url': url,
-        }
[...]

PR #4455, opened 2023-05-17
Add Spawner.progress_ready_hook for customizing the ready event

kreuzert:
discussion: https://github.com/jupyterhub/jupyterhub/issues/4447

### DIFF
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -4,6 +4,7 @@
 import asyncio
 import inspect
 import json
+from copy import deepcopy
 from datetime import datetime, timedelta, timezone
 
 from async_generator import aclosing
@@ -715,13 +716,14 @@ class SpawnProgressAPIHandler(APIHandler):
 
         async def get_ready_event():
             url = url_path_join(user.url, url_escape_path(server_name), '/')
-            ready_event = original_ready_event = {
+            ready_event = {
                 'progress': 100,
                 'ready': True,
                 'message': f"Server ready at {url}",
                 'html_message': 'Server ready at <a href="{0}">{0}</a>'.format(url),
                 'url': url,
             }
+            original_ready_event = deepcopy(ready_event)
             if spawner.progress_ready_hook:
                 try:
                     ready_event = spawner.progress_ready_hook(spawner, ready_event)
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -4,7 +4,6 @@
 import asyncio
 import inspect
 import json
-from copy import deepcopy
 from datetime import datetime, timedelta, timezone
 
 from async_generator import aclosing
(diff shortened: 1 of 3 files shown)


minrk:
Looks sensible to me, thank you! Can you add a test for this in `test_api.py` around the other `test_progress_` tests?

kreuzert:
I've added three tests (async, sync, exception) for `progress_ready_hook` and switched to `deepcopy` to create the `original_ready_event`. Otherwise, modifications of `ready_event` in `progress_ready_hook` would not have been reverted.
  
Please let me know what you think about it.

jupyterhub/apihandlers/users.py
```diff
@@ -710,19 +712,32 @@ async def get(self, user_name, server_name=''):
...
+        async def get_ready_event():
+            url = url_path_join(user.url, url_escape_path(server_name), '/')
+            ready_event = {
+                'progress': 100,
+                'ready': True,
+                'message': f"Server ready at {url}",
+                'html_message': 'Server ready at <a href="{0}">{0}</a>'.format(url),
+                'url': url,
+            }
+            original_ready_event = deepcopy(ready_event)
```
minrk, line 726: deepcopy is fine here, but I don't think it is strictly required _in this case_ because all of the values are immutable scalars. So there isn't really a difference between a shallow and deep copy of this dict:

```python
In [8]: d = {"html_message": "Server ready"}

In [9]: d2 = d.copy()

In [10]: d2["html_message"] += " extra"

In [11]: d2["extra_key"] = "ok"

In [12]: d
Out[12]: {'html_message': 'Server ready'}

In [13]: d2
Out[13]: {'html_message': 'Server ready extra', 'extra_key': 'ok'}
```

deepcopy only makes a difference when the _values_ of the dict are themselves mutable, like lists, dicts, etc.

jupyterhub/spawner.py
```diff
@@ -840,6 +840,28 @@ def _deprecate_percent_u(self, proposal):
...
+        ready event, which will be shown to the user once a service
+        is ready.
+
+        This can be set independent of any concrete spawner implementation.
+
+        This maybe a coroutine.
+
+        Example::
+
+            from subprocess import check_call
```
minrk, line 855:
```suggestion
```

jupyterhub/spawner.py
```diff
@@ -840,6 +840,28 @@ def _deprecate_percent_u(self, proposal):
         """,
     ).tag(config=True)
 
+    progress_ready_hook = Any(
+        help="""
+        An optional hook function that you can implement to modify the
+        ready event, which will be shown to the user once a service
```
minrk, line 846:
```suggestion
        ready event, which will be shown to the user on the spawn progress page when their server
```

minrk (commented):
Looks great, thanks for the test! It's still marked as draft, is there anything you'd like to add?

My only suggestions are in the hook docstring.

kreuzert:
Thanks for your feedback, I've updated it. From my side, we can merge it.

minrk:
Thanks!

PR #4457, opened 2023-05-19
Delete server button on admin page

### LINKED ISSUE #4174: No more delete server button on admin page
### Bug description

Since react admin page there is no more "delete server" button on stopped server.

servers

#### Expected behaviour

As admin, I need the possibility to delete stopped servers.

#### Actual behaviour

Can't delete stopped servers in UI.

diocas:
Fixes #4174 

Hi, I was in the Jupytercon Sprints last week, with no clear fix to contribute, so decided to start with a "good first issue".

This PR adds a "delete server" button to named servers (personal servers cannot be deleted, according to the API) + basic tests. ~~The button is always visible but enabled/disabled depending on the case.~~

Screenshot 2023-05-19 at 16 40 56

Let me know if this is ok. Tried linting and fixing the tests, hope I didn't mess up the local configuration.

### DIFF
--- a/jsx/src/components/ServerDashboard/ServerDashboard.jsx
+++ b/jsx/src/components/ServerDashboard/ServerDashboard.jsx
@@ -45,11 +45,11 @@ const ServerDashboard = (props) => {
     adminAsc = (e) => e.sort((a) => (a.admin ? 1 : -1)),
     dateDesc = (e) =>
       e.sort((a, b) =>
-        new Date(a.last_activity) - new Date(b.last_activity) > 0 ? -1 : 1
+        new Date(a.last_activity) - new Date(b.last_activity) > 0 ? -1 : 1,
       ),
     dateAsc = (e) =>
       e.sort((a, b) =>
-        new Date(a.last_activity) - new Date(b.last_activity) > 0 ? 1 : -1
+        new Date(a.last_activity) - new Date(b.last_activity) > 0 ? 1 : -1,
       ),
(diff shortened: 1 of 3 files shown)


minrk:
Thanks! Can you make the delete button gone/invisible instead of disabled for the default servers? Named servers aren't used very often, so most of the time 100% of the 'delete server' buttons on the screen will be disabled. I think this will work better if they aren't there, instead.

Don't worry about the lint - pre-commit is supposed to be applying it automatically, but apparently isn't. Will have to look into why, exactly.

diocas:
Are these pytests flaky? Doesn't seem related to my change.

minrk:
Thanks! Looks like we do have at least one somewhat flaky test, passed after restarting.

I think there's room for improvement in the button conditions because we still show the 'start server' button when a server is _pending_ start, during which the delete will fail.

The conditions are:

- 'pending' - no actions can be taken, but spawn page may be visited (currently: start, delete buttons are shown but do not work). Might be worth a non-interactive indicator in the UI. `pending` is a _key_ in the server model and may have the values 'start', 'stop', and null (nothing pending).
- 'ready' - start, delete cannot be done, stop & access (conditions are correct here)
- 'inactive' - start, delete may be done (correct as well)

For convenience, we also have an 'active' state, which is really `ready | pending`.

I'm going to merge this one, as it's completed what it says on the tin, but would you like to try a PR for handing the 'pending' state more precisely?

diocas:
Yea, sure. Can you create a ticket?

PR #4463, opened 2023-05-31
Reorder token request docs

minrk:
- suggest token page first
- remove caveat about JupyterHub 0.8, which can be assumed now
- undocument `jupyterhub token`
- refresh token page screenshots, and remove duplicate screenshot of the token page
- minor improvements to language in token page itself, such as matching language so it reads "Expires in 1 week" instead of "Expires at in 1 week".


docs/source/howto/rest.md
```diff
@@ -74,6 +51,40 @@ JupyterHub's token page after successfully requesting a token.
...
+so that both JupyterHub and the service have access to the same value.
+
+First, you need to generate a good random secret.
+A good way of generating an API token is by running:
+
+```bash
+openssl rand -hex 32
+```
+
+This `openssl` command generates a potential token that can then be added to JupyterHub configuration in `jupyterhub_config.py`.
```
manics, line 67:
```suggestion
This `openssl` command generates a random token that can be added to the JupyterHub configuration in `jupyterhub_config.py`.
```

docs/source/howto/rest.md
```diff
@@ -74,6 +51,40 @@ JupyterHub's token page after successfully requesting a token.
...
+    },
+]
+```
+
+At this point, requests authenticated with the token will be associated with The service `my-service`.
+
+```{note}
+You can also load additional tokens for users via the `JupyterHub.api_tokens` configuration.
+
+However, this option has been softly deprecated since the introduction of services.
```
manics, line 85:
```suggestion
However, this option has been deprecated since the introduction of services.
```

manics (commented):
Looks mostly fine to me

PR #4465, opened 2023-06-01
Initial support for OTP MFA

opoplawski:
This is pretty basic but it seems to work.  Probably want to enhance it by enabling the configuration of OTP support and the displaying of the OTP password prompt.  It also requires the just released pamela 1.1.0


share/jupyterhub/templates/login.html
```diff
@@ -64,6 +64,15 @@ <h1>Sign in</h1>
       id="password_input"
       tabindex="2"
     />
+    <label for='otp_input'>OTP (if required):</label>
```
minrk, line 67: Nice! Can we put this behind an option, so OTP isn't presented as an input if it's not going to be expected? I think that might be confusing for some. Maybe `Authenticator.expect_otp`?

opoplawski:
So this is my crude attempt at making this configurable - but I clearly don't understand how Authenticator config settings can make it down to the LoginHandler.  Help please?

jupyterhub/tests/test_auth.py
```diff
@@ -19,31 +19,31 @@
 async def test_pam_auth():
     authenticator = MockPAMAuthenticator()
     authorized = await authenticator.get_authenticated_user(
-        None, {'username': 'match', 'password': 'match'}
+        None, {'username': 'match', 'password': 'match', 'otp': ''}
```
minrk, line 22: These should pass without 'otp' being specified if `request_otp` is False

jupyterhub/handlers/login.py
```diff
@@ -105,6 +105,8 @@ def _render(self, login_error=None, username=None):
                     'next': self.get_argument('next', ''),
                 },
             ),
+            "authenticator_request_otp": self.settings['request_otp'],
```
minrk, line 108: Here's a question - should the login form template get the `authenticator` itself? Seems like it might be useful for custom login forms, but is a bit of a can of worms supported-API-wise.

jupyterhub/handlers/login.py
```diff
@@ -105,6 +105,8 @@ def _render(self, login_error=None, username=None):
...
+            "authenticator_otp_prompt": self.settings['otp_prompt'],
```
minrk, line 109: These are attributes on the Authenticator:

```suggestion
            "authenticator_request_otp": self.authenticator.request_otp,
            "authenticator_otp_prompt": self.authenticator.otp_prompt,
```


By exposing a `config=True` trait on a class, configuration ultimately sets the attribute values on the instances themselves. It's like giving users the ability to do `self.authenticator.request_otp = "x"` on an approved set of attributes in their config files.

jupyterhub/auth.py
```diff
@@ -1105,7 +1124,10 @@ def authenticate(self, handler, data):
         username = data['username']
         try:
             pamela.authenticate(
-                username, data['password'], service=self.service, encoding=self.encoding
+                username,
+                (data['password'], data['otp']),
```
minrk, line 1128: data['otp'] should only be expected when `self.request_otp`, I think. Alternately, it can be added if present in `data`, but shouldn't be assumed present.

minrk (commented):
Thanks! Just a few more details. Let me know if you have any questions, or if you'd like us to take over and finish it up, that's always fine, too.

opoplawski:
I'm afraid I don't have any more time to pursue this.  If you could take it over, that would be great.  Thanks.

minrk:
Applied suggestions, thanks!

opoplawski:
Thank you very much for finishing this up.

PR #4471, opened 2023-06-06
add JupyterHub.subdomain_hook

### LINKED ISSUE #4282: Dot should not be an allowed domain name character
### Bug description

If a username contains a period and subdomains are enabled, they will get a sub-subdomain now because this code doesn't mind sub-subdomains:

https://github.com/jupyterhub/jupyterhub/blob/49c518940bbdedbaf9037d463bcf455c794e5e3f/jupyterhub/user.py#L58-L60

I believe, however, that this doesn't match a common use case. For example, wildcard certificates don't support multilevel nesting, if I understand correctly.

#### Expected behaviour

Dot isn't considered a `_dns_safe` character.

### LINKED ISSUE #4306: Services are moved to a subdomain if subdomain_host is set
### Bug description

Specifying `c.JupyterHub.subdomain_host` moves the services URL to `//services.<subdomain_host>/services/`. This was a bit tricky to figure out, and I in the end only discovered it by querying the proxy via its API.

#### Expected behaviour

- Either services should stay at the main hub domain
- Or the behavior should be documented when describing the service behavior and additionally the admin panel should take this into account when computing service URLs (now it does not).

### How to reproduce

1. Setup a whoami service
2. Configure the jupyterhub to use a subdomain_host
3. Check proxy paths

### Your personal set up

Jupyterhub 3.1.0

minrk:
overrideable so folks can use custom schemes

and new <del>opt-in</del> __default__ 'idna' hook that should always produce valid domains, and separate domains for each service.

closes #4306 by fixing the nav links (@akhmerov) and making service domains customizable

also closes #4282 by excluding '.' from the safe chars in the idna scheme and giving the hook as the escape hatch


jupyterhub/app.py
```diff
@@ -737,6 +739,55 @@ def _domain_default(self):
...
+        Hook for constructing subdomains for users and services.
+
+        Default: 'legacy'
+        Use 'idna' to select a newer, more robust scheme,
+        which will become the default in the future.
+
+        A custom subdomain hook should have the signature:
+
+        def subdomain_hook(name, domain, kind) -> str:
+            ...
```
consideRatio, line 756: I don't understand this config so well from the docstring alone, I think the following could be concrete suggestions:

- Mention that `"legacy"` and `"idna"` string options refer to _pre-defined hook functions_
- Clarify what the pre-defined hook function does in this docstring or link out to read about them elsewhere.
- I think maybe this config is relevant only if another config option is enabled, is it? I've never worked with a jupyterhub providing subdomains for each user, and figure that it must be opted-in one way or another.

minrk: All great points that should be addressed in the docstring, thanks!

akhmerov:
I would like to bring up an alternative sanitization strategy, that seems to be impossible with this hook:
1. `john.smith -> john_smith`
2. `john_smith1 -> john_smith1`
3. `john-smith -> john_smith2`
i.e. one could resolve sanitized name conflicts by checking which names already exist.

minrk:
Yes, sanitizing collisions _only if they actually happen_ wouldn't really fit with this scheme, as the scheme is responsible for guaranteeing uniqueness. Such a scheme would be order dependent, though - the URL for each colliding user could change over time based on who logged in first that day, etc. That is, unless we persist the result of each normalization to the database.

You could implement the persistent state in your function, I suppose, by using e.g. an sqlite-backed LRU cache.

<details>
<summary>like this</summary>

```python
import sqlite3
import re


def sqlite_cached(f):
    conn = sqlite3.connect("domains.sqlite")
    conn.execute(
        """
        CREATE TABLE IF NOT EXISTS domains (
            name TEXT,
            domain TEXT,
            kind TEXT,
            result TEXT UNIQUE,
            PRIMARY KEY(name, domain, kind)
        )
        """
    )

    def cached_f(name, domain, kind):
        cur = conn.execute(
            "SELECT result FROM domains WHERE name=? AND domain=? AND kind=? LIMIT 1",
            (name, domain, kind),
        )
        row = cur.fetchone()
        if row:
            print("cache hit", name, row[0])
            return row[0]
        result = original_result = f(name, domain, kind)
        # lookup collision
        matched = conn.execute(
            "SELECT result FROM domains WHERE result=? LIMIT 1", (result,)
        ).fetchone()
        i = 0
        while matched:
            print("collision", name, result)
            i += 1
            result = f"{original_result}-{i}"
            matched = conn.execute(
                "SELECT result FROM domains WHERE result=? LIMIT 1", (result,)
            ).fetchone()

        conn.execute(
            "INSERT INTO domains VALUES(?, ?, ?, ?)", (name, domain, kind, result)
        )
        conn.commit()

        return result

    return cached_f


non_letter = re.compile(r'[^a-z]+')


@sqlite_cached
def collision_detecing_hook(name, domain, kind):
    return non_letter.sub("", name) + "." + domain


for name in ["foo bar", "foobar", "foo.bar"]:
    print(name, collision_detecing_hook(name, 'host.tld', 'user'))
```

</details>

which implements the scheme you proposed, while meeting the spec provided here. I think I'm satisfied that the scheme requires uniqueness and if the scheme wants to use state as part of that implementation, it's up to the scheme implementation to manage that state, rather than making it part of JupyterHub's own state.

minrk:
The biggest downside of the new scheme is outlined in https://github.com/jupyterhub/jupyterhub/issues/4282, that the collection of domains chosen (`{user}.users.{domain}` and `{service).services.{domain}`) _cannot_ be served by a single wildcard certificate. But it solves a collision issue where `services` was in the same namespace as the usernames.

Another approach to the collision issue would be to use a prefix/suffix (`user-{username}.{domain}` and `service-{servicename}.{domain}`, or apply the prefix only to services, and use a delimiter like `--` that ensures it cannot collide with a username because usernames with `--` always go via the trim-and-hash route.

minrk:
I went with the `--user` suffix so that users will be on `myname--user.hub.horse`, to avoid the wildcard SSL problem. There would still be no collisions if we only added `--service` to service subdomains, and let users be the 'default'. This would mean no domain change at all for the vast majority of usernames that require no encoding, which would be a plus. The use of `--` as a delimiter means there can be no collisions with services or any other suffix, as `--` in any username forces re-encoding and therefore cannot collide. The advantage of `--user` is that the namespacing is explicit.

jupyterhub/utils.py
```diff
@@ -779,3 +783,149 @@ def get_browser_protocol(request):
...
+    for c in unique_chars:
+        if c not in _dns_needs_replace:
+            label = label.replace(c, f"%{ord(c):x}")
+
+    # underscore is our escape char -
+    # it's not officially legal in hostnames,
+    # but is valid in _domain_ names (?),
+    # and seems to always work in practice.
+    # FIXME: We should consider switching to proper IDNA encoding
+    # in a major release
```
consideRatio, line 820:
```suggestion
```

The fixme note is part of a helper function used by `subdomain_hook_legacy`, which is used if the `somdomain_hook` config makes use of `legacy`. The docstring of `subdomain_hook` already captures the idea that this is to be changed in the future.

consideRatio (approved):
Wow nice work, this looks great!!

minrk:
Updated from discussion with @consideRatio:

Since this is going to be new in 5.0 instead of 4.1, made the following changes:

- default is now 'idna'
- user domains do not have `--user` suffix, meaning no practical change for usernames that do not need escaping, the vast majority of domains governed by this hook

consideRatio (approved):

Issue #4473, opened 2023-06-07
Custom log_function in custom tornado_settings in config file is not used in SingleUserApp


grios-stratio:
### Bug description

Setting a custom log_function via custom tornado_settings in the config file for the singleUserApp has no effect

#### Expected behaviour

When setting a custom log_function in the tornado_settings for the singleUserApp in the config file, this function should be used by tornado to log the requests (this is the case for the hub server, but it does not happen in the singleUserApp). 

#### Actual behaviour

The "default" log_function is used, instead of the custom one defined in the configuration file

### How to reproduce

1. set up a config file `~/.jupyter/jupyter_server_config.py`
```python
# jupyter_server_config.py
import sys

# to check that the config file is being loaded
c.ServerApp.log_format = 'testing => [%(name)s] %(message)s'

def custom_log_fun(handler):
    sys.stderr.write(f'hello from custom log fun {handler.request.uri}')

c.ServerApp.tornado_settings = {'log_function': custom_log_fun}
```
2. launch jupyterhub
3. login
4. see in the logs that the requests are not being logged with the custom function (should say 'hello from custom log fun'):
```
testing => [SingleUserLabApp] 302 GET /user/grios/oauth_callback?code=[secret]&state=[secret] -> /user/grios/lab?redirects=1 (@::1) 45.42ms
testing => [SingleUserLabApp] 200 GET /user/grios/lab?redirects=1 (grios@::1) 9.08ms
testing => [SingleUserLabApp] 200 GET /user/grios/api/kernelspecs?1686149978280 (grios@::1) 4.49ms
testing => [SingleUserLabApp] 200 GET /user/grios/lab/api/settings?1686149978285 (grios@::1) 79.00ms
testing => [SingleUserLabApp] 200 GET /user/grios/api/kernels?1686149978289 (grios@::1) 2.97ms

```

### Your personal set up

 - OS: Ubunto 20.04
 - Version(s): I think it happens in all jupyterhub versions, but I have checked it explicitly on 4.0.0 and 2.3.0 (python 3.9)
 

### more info
I believe the problem is that the log function is ovewriten in the tornado_settings without checking if it has been set by the user, both in the SingleUserNotebookAppMixin:
https://github.com/jupyterhub/jupyterhub/blob/d64d916abc012ce5e80ce54fb17277e7a52045bd/jupyterhub/singleuser/mixins.py#L668-L673

and in the JupyterHubSingleUser app:
[...]

PR #4475, opened 2023-06-08
Allow setting custom log_function in tornado_settings in SingleUserServer

grios-stratio:
Fixes https://github.com/jupyterhub/jupyterhub/issues/4473. 

The problem there was that setting a custom log_function for the singleUserServer in the tornado_settings in the jupyter_server_config.py had no effect since it was overwritten during singleUserServer initialization. This PR fixes that by checking if the log_function in the settings is already defined before setting it.

### DIFF
--- a/jupyterhub/singleuser/extension.py
+++ b/jupyterhub/singleuser/extension.py
@@ -617,8 +617,8 @@ class JupyterHubSingleUser(ExtensionApp):
         app.web_app.settings[
             "page_config_hook"
         ] = app.identity_provider.page_config_hook
-        # allow to override log_function with tornado_settings set in user config
-        if not app.web_app.settings.get("log_function"):
+        # if the user has configured a log function in the tornado settings, do not override it
+        if not 'log_function' in app.config.ServerApp.get('tornado_settings', {}):
             app.web_app.settings["log_function"] = log_request
         # add jupyterhub version header
         headers = app.web_app.settings.setdefault("headers", {})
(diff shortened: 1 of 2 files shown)


minrk (commented):
Thanks! When I test, `app.web_app.settings['log_function']` is defined here (set in jupyter-server [here](https://github.com/jupyter-server/jupyter_server/blob/d48843eea78fe2ee37608a767ceea3d0d2f71763/jupyter_server/serverapp.py#L343), before the extension is loaded) in the default case, so this has the effect of removing the default behavior, which is not quite what we want.

I think you'll have to check for `log_function` in `app.config.tornado_settings` to check if it was configured at a higher level:

```python
if not ('tornado_settings' in app.config.ServerApp and 'log_function' in app.config.ServerApp.tornado_settings):
```

grios-stratio:
> Thanks! When I test, `app.web_app.settings['log_function']` is defined here (set in jupyter-server [here](https://github.com/jupyter-server/jupyter_server/blob/d48843eea78fe2ee37608a767ceea3d0d2f71763/jupyter_server/serverapp.py#L343), before the extension is loaded) in the default case, so this has the effect of removing the default behavior, which is not quite what we want.
> 
> I think you'll have to check for `log_function` in `app.config.tornado_settings` to check if it was configured at a higher level:
> 
> ```python
> if not ('tornado_settings' in app.config.ServerApp and 'log_function' in app.config.ServerApp.tornado_settings):
> ```

Ahhh you are absolutely right! I didn't quite understood you when you commented just that in the issue  :man_facepalming:, sorry!   (and also I was testing the thing backwards).

minrk:
Thanks!

minrk:
@meeseeksdev backport to 4.x

PR #4476, opened 2023-06-11
admin page: buttons handle pending status correctly

### LINKED ISSUE #4466: admin server buttons don't have quite the right conditions
### Bug description

Follow-up from #4457

I think there's room for improvement in the button conditions because we still show the 'start server' button when a server is pending start, during which the delete will fail.

The conditions are:

- 'pending' - no actions can be taken, but spawn page may be visited (currently: start, delete buttons are shown but do not work). Might be worth a non-interactive indicator in the UI. pending is a key in the server model and may have the values 'start', 'stop', and null (nothing pending).
- 'ready' - start, delete cannot be done, stop & access (conditions are correct here)
- 'inactive' - start, delete may be done (correct as well)

For convenience, we also have an 'active' state, which is really `ready | pending`.

#### Expected behaviour

Start Server and Delete Server buttons are only visible while the server is 'inactive', which is when they can work. Alternately: visible and inactive with informative tooltip when 'pending'. When pending start, the spawn page is a valid destination, which will show the progress while it waits for startup. When pending stop, there's nothin really to be done.

The simplest solution, which would be to move the current condition line from 'ready' to 'active', as the access server button will work and the stop button should fail informatively (and may work if you wait a bit), but it could be a little better than that to take pending into account.

#### Actual behaviour

- Start Server and Delete Server buttons are visible while the server is pending, and they will not work

### How to reproduce

1. begin launching a server that takes a while to start
2. visit admin page
3. see 'Start Server' and 'Delete Server' buttons
4. click one, it won't work

diocas:
Fix #4466

This makes the buttons more aware of the pending state. This is how it looks:

Screenshot 2023-06-11 at 19 09 14

This implementation has one detail: if you click "Start server", the button gets disabled but the model does not get updated. That only happens on success/failure, as before. What this means is that the message "Server is pending" only shows up if we refresh the page. Is this ok?

I've also made some changes to the `webpack-dev-server` configuration. This allows opening the admin panel without starting JH while also allows "watching" for code changes.

Not sure why, but `yarn.lock` doesn't seem up to date.

Please let me know if this is ok as it is or if you want changes. I'll update the tests afterwards.

(Sorry for taking so long to reply, but only managed to find time today)


diocas:
One last comment: if there's something pending, the model is not updated until another refresh of the page (or we do another action). Should we automate this?

diocas:
I'm having some issues with the jest tests, not sure if anyone can help me. For some reason, at the end, the Stop button is not rendering, and it still shows Start. If I put some `console.log(random_non_existent_var)` before `dispatchPageUpdate` (`ServerDashboard.jsx`), then it works...

minrk:
Thanks!

PR #4479, opened 2023-06-14
add JupyterHub.public_url config

minrk:
When specified, removes all guesses based on the request, which aren't always correct in proxy situations (especially Host and protocol)

idea came up in https://github.com/jupyterhub/ltiauthenticator/pull/166

todo:

- [x] figure out some interactions with subdomain_host
- [x] pass JUPYTERHUB_PUBLIC_URL to spawners


jupyterhub/app.py
```diff
@@ -704,6 +704,21 @@ def _bind_url_default(self):
...
+        config=True,
+        help="""Set the public URL of JupyterHub
+
+        This will skip any detection of URL and protocol from requests,
+        which isn't always correct when JupyterHub is behind
+        multiple layers of proxies, etc.
+        Usually the failure is detecting http when it's really https.
+
+        Should include the full, public URL of JupyterHub,
+        including the public-facing base_url prefix.
```
yuvipanda, line 718: Can you specify if this should or should not include a trailing slash?

yuvipanda (changes requested):

jupyterhub/app.py
minrk: Indeed it should, and we can add a validator to _ensure_ it does, so that common mistakes can't cause problems.

yuvipanda: +1

yuvipanda:
Thanks Min! This would indeed be useful to pass to spawners!

How would this work with subdomains?

minrk:
Not sure where my comment went that I thought I posted, but I think I worked it out: setting subdomain_host is _equivalent_ to setting public_url, since the domain is necessarily the public URL. So:

- setting subdomain_host sets public_url _and_ enables user subdomains
- setting public_url directly influences things like the cookie `secure` flag without enabling subdomains
- subdomain_host and public_url should match

There are _two_ public URLs passed to servers/services

- JUPYTERHUB_PUBLIC_URL - the public URL of the server/service that receives the env (available as Spawner.public_url). This is what should influence cookies, etc.
- JUPYTERHUB_PUBLIC_HUB_URL - the public URL of the full deployment (available as JupyterHub.public_url). This information is already available when using subdomains as JUPYTERHUB_HOST (used to construct login/logout URLs), but PUBLIC_HUB_URL contains `base_url` while JUPYTERHUB_HOST does not.

Naming these things is hard, so if anyone has strong feelings, feel free to suggest alternatives.

minrk:
Relevant to sharing in #4594, and others where we have 'url' fields that are actually just url paths (like `server.url = '/users/name/'`), with this defined, we could start putting actual _full_ URLs in the API responses. For backward compatibility, this should probably be in new fields (e.g. `url='/path/', full_url='$public_url/path/'`).

docs/source/reference/services.md
```diff
@@ -138,6 +138,14 @@ JUPYTERHUB_OAUTH_SCOPES:   JSON-serialized list of scopes to use for allowing ac
                            (deprecated in 3.0, use JUPYTERHUB_OAUTH_ACCESS_SCOPES).
 JUPYTERHUB_OAUTH_ACCESS_SCOPES: JSON-serialized list of scopes to use for allowing access to the service (new in 3.0).
 JUPYTERHUB_OAUTH_CLIENT_ALLOWED_SCOPES: JSON-serialized list of scopes that can be requested by the oauth client on behalf of users (new in 3.0).
+JUPYTERHUB_PUBLIC_URL: the public URL of the service,
+  e.g. `https://jupyterhub.horse/services/name/`.
```
manics, line 142:
```suggestion
  e.g. `https://jupyterhub.example.org/services/name/`.
```
I think it's best to use official example domains, just in case something ends up on `jupyterhub.horse` in future. If you want a TLD `jupyterhub.example` [also works](https://datatracker.ietf.org/doc/html/rfc2606).

The definition of `JUPYTERHUB_SERVICE_PREFIX` uses `:service-name` instead of `name`, either is fine but I think it should be consistent.

docs/source/reference/services.md
```diff
@@ -138,6 +138,14 @@ JUPYTERHUB_OAUTH_SCOPES:   JSON-serialized list of scopes to use for allowing ac
...
+  Empty if no public URL is specified (default).
+  Will be available if subdomains are configured.
+JUPYTERHUB_PUBLIC_HUB_URL: the public URL of JupyterHub as a whole,
+  e.g. `https://jupyterhub.horse/`.
```
manics, line 146:
```suggestion
  e.g. `https://jupyterhub.example.org/`.
```

jupyterhub/app.py
```diff
@@ -698,6 +698,60 @@ def _bind_url_default(self):
...
+        help="""Set the public URL of JupyterHub
+
+        This will skip any detection of URL and protocol from requests,
+        which isn't always correct when JupyterHub is behind
+        multiple layers of proxies, etc.
+        Usually the failure is detecting http when it's really https.
+
+        Should include the full, public URL of JupyterHub,
+        including the public-facing base_url prefix
+        (i.e. it should include a trailing slash).
```
manics, line 713:
```suggestion
        (i.e. it should include a trailing slash), e.g.
        https://jupyterhub.example.org/prefix/
```

jupyterhub/app.py
```diff
@@ -698,6 +698,60 @@ def _bind_url_default(self):
...
+        if not url:
+            # explicitly empty (default)
+            return url
+        if not url.endswith("/"):
+            # ensure we have a trailing slash
+            # for consistency with base_url
+            url = url + "/"
+        if not url.endswith(self.base_url):
+            if not urlparse(url).path.strip("/"):
+                # no path specified, add base_url and warn
```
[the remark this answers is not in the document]
minrk: Since  this is a URL prefix by definition, I think adding the slash is the right thing to do. If it were a canonical URL for a single page, that would be different, but it is a prefix so adding a slash is for our internal convenience and consistency when joining with sub-paths.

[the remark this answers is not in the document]
manics: Adding the slash is fine, it's the addition of `base_url` I was wondering about. Not a blocker though.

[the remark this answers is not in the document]
minrk: Ah, gotcha. Yeah, that seems okay to raise. What we really care about here is the public _origin_ (proto://host:port), as base_url config already exists and is defined elsewhere, so specifying just the origin seems like sensible behavior as opposed to having to change base_url in multiple places. However, if you might be _hoping_ that public_url allows specifying a public URL that _doesn't_ match base_url would allow proxying, that's not going to work because we use absolute paths with base_url (you can't do e.g. proxy_pass with rewriting paths by adding or removing a prefix, for example), in which case an error could be helpful.

consideRatio (approved):
Wieee looks fine!

PR #4480, opened 2023-06-14
Remove check for allowed_users for users created via role or group config

### LINKED ISSUE #4461: Can we override `Authenticator.check_allowed` in jupyterhub/oauthenticator and assume authentication argument be passed?
In [oauthenticator #594](https://github.com/jupyterhub/oauthenticator/pull/594) we are currently planning to override the `check_allowed` as considered and proposed in https://github.com/jupyterhub/oauthenticator/pull/594#issuecomment-1519086377 and the comments below. The [`check_allowed` function signature](https://github.com/jupyterhub/jupyterhub/blob/4.0.0/jupyterhub/auth.py#L432-L438) is...

```python
    def check_allowed(self, username, authentication=None):
        """Check if a username is allowed to authenticate based on configuration

        Return True if username is allowed, False otherwise.
        No allowed_users set means any username is allowed.

        Names are normalized *before* being checked against the allowed set.
```

What came as a surprise to me now was that check_allowed was also being called by JupyterHub without passing the authentication dictionary (aka. `auth_model`), and having error messages related to specifically that `allowed_users` check failing.

It seems that also the [LocalAuthenticator](https://github.com/jupyterhub/jupyterhub/blob/main/jupyterhub/auth.py#L871-L875) class overrides `check_allowed`, and adds a check for beloning to either allowed_users or allowed_groups very similar to what we want to do. Our unique challenge is that we rely on being passed `authentication` aka `auth_model`, and that isn't passed at all times when check_allowed is used.

In practice it seems that only if the deprecated `JupyterHub.api_tokens` is used or if `JupyterHub.load_roles` is configured, and they involve user api tokens or similar, then `check_allowed` will be called without an authentication dictionary.

### Using `JupyterHub.load_roles`

[`JupyterHub.init_role_assignment`](https://github.com/jupyterhub/jupyterhub/blob/4.0.0/jupyterhub/app.py#L2274-L2295)

```python
    async def init_role_assignment(self):
        """
        This function loads config from `JupyterHub.load_roles`
        """
        # tokens are added separately
        kinds = ['users', 'services', 'groups']

        # ...

            # ...
        
            # add users, services, and/or groups,
            # tokens need to be checked for permissions
            for kind in kinds:
                orm_role_bearers = []
                if kind in role_spec:
                    for name in role_spec[kind]:
                        if kind == 'users':
                            name = self.authenticator.normalize_username(name)
                            if not (
                                await maybe_future(
                                    self.authenticator.check_allowed(name, None)
                                )
                            ):
                                raise ValueError(
                                    f"Username {name} is not in Authenticator.allowed_users"
                                )
```

### Using `JupyterHub.api_tokens`

[`JupyterHub.api_tokens`](https://github.com/jupyterhub/jupyterhub/blob/4.0.0/jupyterhub/app.py#L1049-L1060)

```python
    api_tokens = Dict(
        Unicode(),
        help="""PENDING DEPRECATION: consider using services

        Dict of token:username to be loaded into the database.

        Allows ahead-of-time generation of API tokens for use by externally managed services,
        which authenticate as JupyterHub users.

        Consider using services for general services that talk to the JupyterHub API.
        """,
    ).tag(config=True)
```

[`JupyterHub.init_api_tokens`](https://github.com/jupyterhub/jupyterhub/blob/4.0.0/jupyterhub/app.py#L2345-L2349)

```python
    async def init_api_tokens(self):
        """Load predefined API tokens (for services) into database"""

        await self._add_tokens(self.service_tokens, kind='service')
        await self._add_tokens(self.api_tokens, kind='user')
        
        # ...
```

[`JupyterHub._add_token`](https://github.com/jupyterhub/jupyterhub/blob/4.0.0/jupyterhub/app.py#L2274-L2295)

```python
    async def _add_tokens(self, token_dict, kind):
        """Add tokens for users or services to the database"""
        if kind == 'user':
            Class = orm.User
        elif kind == 'service':
            Class = orm.Service
        else:
            raise ValueError("kind must be user or service, not %r" % kind)

        db = self.db
        for token, name in token_dict.items():
            if kind == 'user':
                name = self.authenticator.normalize_username(name)
                if not (
                    await maybe_future(self.authenticator.check_allowed(name, None))
                ):
                    raise ValueError(
                        "Token user name %r is not in Authenticator.allowed_users"
                        % name
                    )
                if not self.authenticator.validate_username(name):
                    raise ValueError("Token user name %r is not valid" % name)
```

minrk:
adding users via config anywhere makes them allowed

previously, this was _required_, so that it was always true for working config, but config which allowed some users but declared others in groups or roles was forbidden.

Now, declaring a user in roles or groups _ensures_ the user is allowed rather than _enforcing_ it.

Upside: avoids issues where `check_allowed` requires auth info (closes #4461), as check_allowed will now always be called with auth_info, never None.

Downside: some config inconsistencies will be assumed correct instead of raising, e.g.

```python
c.Authenticator.allowed_users = {'minrk'}
c.JupyterHub.load_roles = [
  {"name": "role",
   "users": ["minkr"],
  }
]
```

will change from raising "minkr is not allowed" to creating both users 'minrk' and 'minkr'.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -2041,21 +2041,20 @@ class JupyterHub(Application):
 
         TOTAL_USERS.set(total_users)
 
-    async def _get_or_create_user(self, username):
+    async def _get_or_create_user(self, username, hint):
         """Create user if username is found in config but user does not exist"""
-        if not (await maybe_future(self.authenticator.check_allowed(username, None))):
-            raise ValueError(
-                "Username %r is not in Authenticator.allowed_users" % username
-            )
         user = orm.User.find(self.db, name=username)
         if user is None:
             if not self.authenticator.validate_username(username):
                 raise ValueError("Username %r is not valid" % username)
-            self.log.info(f"Creating user {username}")
+            self.log.info(f"Creating user {username} found in {hint}")
             user = orm.User(name=username)
             self.db.add(user)
             roles.assign_default_roles(self.db, entity=user)
             self.db.commit()
+            f = self.authenticator.add_user(user)
+            if f:
+                await maybe_future(f)
         return user
 
     async def init_groups(self):
@@ -2064,7 +2063,9 @@ class JupyterHub(Application):
 
         if self.authenticator.manage_groups and self.load_groups:
             raise ValueError("Group management has been offloaded to the authenticator")
+
         for name, contents in self.load_groups.items():
+            self.log.debug("Loading group %s", name)
             group = orm.Group.find(db, name)
 
             if group is None:
@@ -2080,9 +2081,11 @@ class JupyterHub(Application):
             if 'users' in contents:
                 for username in contents['users']:
                     username = self.authenticator.normalize_username(username)
-                    user = await self._get_or_create_user(username)
+                    user = await self._get_or_create_user(
+                        username, hint=f"group: {name}"
+                    )
                     if group not in user.groups:
-                        self.log.debug(f"Adding user {username} to  group {name}")
+                        self.log.debug(f"Adding user {username} to group {name}")
                         group.users.append(user)
 
             if 'properties' in contents:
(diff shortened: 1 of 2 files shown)


consideRatio (approved):
Looks great, and I think the PR description is very good!

PR #4491, opened 2023-06-26
avoid counting failed requests to not-running servers as 'activity'

minrk:
For example, when a user stops and a JupyterLab tab is left open, requests will be made to `/user/:name/api/...`. When these are handled by the Hub, they should not be considered 'activity', contributing to users' last activity metrics, etc.

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -1433,6 +1433,12 @@ class UserUrlHandler(BaseHandler):
     # accept token auth for API requests that are probably to non-running servers
     _accept_token_auth = True
 
+    # don't consider these redirects 'activity'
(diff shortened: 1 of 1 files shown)


minrk:
For example, I have this in the jupyterhub-grafana-dashboards, because 'active users' is actually running servers (max 1), while 'daily active users' is any kind of hub activity (minimum 7), even if no servers are running.

Screenshot 2023-06-26 at 11 39 35


I looked into the logs, and most of the activity for the other users was these 424 errors for a left-open tab. I believe after this change, the two metrics would end up more aligned.

consideRatio (approved):

minrk:
@meeseeksdev backport to 4.x

Issue #4485, opened 2023-06-16
Performance problems on start from N+1 queries


liamstewart:
### Bug description

We are running a large JupyterHub installation on kubernetes using z2jh 2.0.0 (so JupyterHub 3.0.0) and noticing performance issues at and after startup. In particular, restarts of the hub take minutes and, after restart, initial requests to `GET /users` API requests for large batches (hundreds) take seconds to return and, while processing, block all other requests to the hub. We've debugged against the JupyterHub 3.0.0 code, but, looking at the latest code, the issues aren't restricted to that version and are present in main.

#### Expected behaviour

- Restarts take a handful of seconds
- Initial `GET /users` requests for large batches take fractions of a second and don't block concurrent requests

#### Actual behaviour

##### Restarts take minutes, the time taken scales with the number of users in the database

The significant part of the start time behaviour appears to be due to an N+1 query and an uncached role in role assignment, in particular https://github.com/jupyterhub/jupyterhub/blob/161cdcd7e711845d2c4c157c150f251261a5f925/jupyterhub/app.py#L2252-L2260

The logs don't exactly pinpoint it but do point to that:

```
[I 2023-06-01 00:16:14.984 JupyterHub app:2775] Running JupyterHub version 3.0.0
[D 2023-06-01 00:16:14.985 JupyterHub app:1783] Connecting to db: postgresql+psycopg2://<redacted>
[D 2023-06-01 00:16:15.135 JupyterHub orm:998] database schema version found: 651f5419b74d
...
[D 2023-06-01 00:16:16.750 JupyterHub app:2110] Loading role assignments from config
[D 2023-06-01 00:16:16.767 JupyterHub app:2181] Assigning XXXXX allowed_users to the user role
[D 2023-06-01 00:19:13.278 JupyterHub app:2433] Initializing spawners
...
[I 2023-06-01 00:19:13.551 JupyterHub app:2844] Initialized XX spawners in 0.273 seconds
```

Notice the 3 minutes it takes to get from assigning users to initializing spawners. `roles.grant_role` should be a no-op because each user is already part of the user role in the db. It's still done once for each user. At least from reads of the code, my understanding is that users have been previously loaded from the database. Looking at AWS Performance Insights for the period of blocking from ~16-19 shows the top queries with over 200 queries / second each to be

```sql
SELECT roles.id AS roles_id, roles.name AS roles_name, roles.description AS roles_description, roles.scopes AS roles_scopes
FROM roles, user_role_map
[...]

PR #4494, opened 2023-06-27
Improve query performance with eager loading

minrk:
Reduces the number of queries by using eager loading as appropriate.

Using an ORM lets us hide queries behind property access (convenient), but this can get unintentionally expensive when we are accessing properties as if they are local in very large queries (#4485).

Switches ORM from using the legacy 'backref' to the more explicit 'back_populates'. This has no change on the database or behavior, but makes both directions of each relationship explicit, rather than defining it on only one side. This helps us specify the chosen lazy-loading options in the right places.

More info on [loading relationships in sqlalchemy](https://docs.sqlalchemy.org/en/20/tutorial/orm_related_objects.html#tutorial-orm-loader-strategies).

In general, the choices are:

- eager loading defined on the relationship itself, so that it's fetched by default when we fetch the object. This makes sense when the relationship is almost certainly used when the object is used (e.g. Spawner.server, APIToken.user).
  Currently this applies to:
  - User.groups, User.roles (used for permissions resolution)
  - Group.roles
  - Spawner.server, user
  - Service.server, roles
  - APIToken.user, service, oauth_client
  - OAuthCode.user, client
  - OAuthClient.spawner, service
- eager loading defined on a particular query via `query.options`. This is useful when a specific request is likely to look at a lot of the relationships (e.g. `GET /api/users` should resolve group/spawner/server relationships)
  - GET /api/users: adds User.spawners
  - May apply to getting user from APIToken, if we don't always get groups/roles from user above
- _raise_ loading to essentially assert that there's no lazy loading going on. I'm using this now for debugging, and hope to enable it for tests in particular, but don't think we should use it in production, since it turns inefficiencies into errors.

In general, this _increases_ the cost of the initial queries, at the expense of reducing the follow-up queries if/when we know they are going to happen, so we need to be careful to not get _too_ eager, loading things we aren't going to use (e.g. I need to make sure we aren't making GET /api/groups too expensive by fetching lots of relationships we aren't going to use about users)

I need to run some profiling to validate that this is actually improving things, not making it worse.

TODO:

- [x] reduce queries in startup by only updating things that need changing (#4496)
- [x] reduce queries in GET /api/users
- [x] reduce queries in cheecking request permissions
- [x] check GET /api/groups for inefficiencies
- [x] profile improvements to make sure this actually helps
- [x] remove raiseload/debugging

I may pull the startup performance out to its own PR, since it turns out to be fairly independent of the other changes.


minrk:
working on the profiling, but this currently has the GET /users for 5k users (paginated requests) down from 14,000 db operations to ~600, a reduction of 95%, and reducing from 14 seconds to 0.6, assuming 1ms latency for each db operation (local sqlite is much quicker than that, but a remote db may not be).

minrk:
Everything works, except I need to tweak something for sqlalchemy 1.4 compatibility.

minrk:
This could perhaps use some refinement, but I think it's a dramatic improvement. db queries are down to 5 per GET /users (regardless of the number of users), from ~5 _per user_ per GET /users (down from 20160 queries to GET 5000 inactive users in 50 pages to 257 for the same requests and same results).

Much further loading optimizations are hard to do while we have a shared db session. This can be further improved as we move to per-request db sessions.

ryanlovett:
Hi @minrk , is this going to be added to 4.x branch for a 4.x release?

PR #4496, opened 2023-06-28
Reduce O(N) queries during startup

minrk:
- avoid lookup-by-name of user and admin roles each time they are assigned
- during admin assignment, filter users-to-update to only those that need updating, which will usually be empty

No change in behavior, but dramatically fewer queries

pulled out of #4494

one of the issues raised in #4485

One thing that would be handy - emit a metric for the number of queries performed for each test. I'm not sure how to do that.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -2281,23 +2281,22 @@ class JupyterHub(Application):
                 self.log.info(f"Setting admin=True on {is_admin}")
                 is_admin.admin = True
 
-            if has_admin_role_spec[kind]:
-                # role membership specified exactly in config,
-                # already populated above
-                # make sure user.admin matches admin role
-                # setting .admin=False for anyone no longer in admin role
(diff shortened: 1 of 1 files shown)


minrk:
Sample startup performance with an artificial 1ms query latency for 5000 active users shows reducing startup time from 50 seconds to ~2 seconds, and (more realistically) 5000 users with few active from ~20 seconds to ~0.4 seconds.

image

https://gist.github.com/minrk/aba0d612b04f7603bc5688b7d8bbd187

benchmark script: https://github.com/minrk/jupyterhub-bench

yuvipanda:
@ryanlovett @shaneknapp I think this would be a big benefit to the berkeley hubs' restart performance.

yuvipanda:
I read through the code, and all the changes make sense to me - thank you for the comments, @minrk! Given all the tests pass, let's go for this!

PR #4497, opened 2023-06-28
orm: trade backref for back_populates

minrk:
relationship.backref is [considered legacy](https://docs.sqlalchemy.org/en/20/orm/backref.html), in favor of the more explicit "back_populates". This changes no behavior at all, only changing how bi-directional relationships are defined, trading:

```python
class A:
  b = relationship("B", backref="a")

class B:
  # has B.a dynamically defined from the above backref without anything declared on 'B' itself
```

for the more explicit and symmetrical

```python
class A:
  b = relationship("B", back_populates="a")

class B:
  a = relationship("A", back_populates="b")
```

Extracted from #4494 so that I can rebase that one to show only actual changes in behavior, rather than including a refactor that has no changes.

Also includes a tiny repr change on orm.Group, which was costly because it fetched all users in the group!

### DIFF
--- a/jupyterhub/orm.py
+++ b/jupyterhub/orm.py
@@ -29,7 +29,6 @@ from sqlalchemy import (
 )
 from sqlalchemy.orm import (
     Session,
-    backref,
     declarative_base,
     interfaces,
     object_session,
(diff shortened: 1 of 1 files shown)


minrk:
Checklist for review, making sure each removed backref has a corresponding back_populates partner:

- [ ] User.roles
- [ ] Service.roles
- [ ] Group.roles
- [ ] User.groups
- [ ] Spawner.user
- [ ] APIToken.user
- [ ] OAuthCode.user
- [ ] Server.spawner
- [ ] OAuthClient.spawner
- [ ] APIToken.service
- [ ] Server.service
- [ ] OAuthClient.service
- [ ] APIToken.oauth_client
- [ ] OAuthCode.client

PR #4503, opened 2023-07-03
set root_dir when using singleuser extension

### LINKED ISSUE #4493: Jupyter 4.0.1 - Unable to access filesystem outside of home directory/notebook directory
### Bug description
I've also posted to Discord.  please forgive me as this is my first bug report.  LMK the preferred method github v discord. ty

I’ve been using jupyterhub for a few years now. I’ve recently tried updating to 4.0.1 but it no longer seems to honor my Spawner.default_url and Spawner.notebook_dir. My system uses the base local single user spawner along with NativeAuthenticator. As of 4.0.1, I’m always locked within the home dir of the user which is not desirable as I have group dirs for users to share behind their home dirs. When I downgrade back to jupyterhub 3.1.1, I can move around as needed

I’ve also even tried this article [Troubleshooting — JupyterHub documentation](https://jupyterhub.readthedocs.io/en/stable/faq/troubleshooting.html#i-want-access-to-the-whole-filesystem-and-still-default-users-to-their-home-directory). This doesn’t appear to work at all as the redirects dup the users homedir path in the url

#### Expected behaviour

c.Spawner.notebook_dir should start at path specified.  Currently only drops user into homedir as "/" and therefore locked within that dir.  

#### Actual behaviour

c.Spawner.notebook_dir = ''/some/path"  should start the user at some specified path instead of forced homedir

### How to reproduce
user homedir ex....   /my_dir/username/nbs

install 4.x jupyter hub
set config:
c.Spawner.notebook_dir = "/my_dir"
c.Spawner.default_url = '/lab'

forces user into home '/my_dir/username'.  I want users to be able to go a level back for shared folders

reverting to Jupyterhub 3.1.1 fixes issue

### Your personal set up
Docker container cuda 11.8 ubuntu 20.08
running jupyterhub in conda base
default local spawner
NativeAuthenticator for auth

minrk:
and make sure Spawner.notebook_dir config is tested

closes #4493

### DIFF
--- a/jupyterhub/tests/test_singleuser.py
+++ b/jupyterhub/tests/test_singleuser.py
@@ -2,6 +2,7 @@
 import os
 import sys
 from contextlib import nullcontext
+from pprint import pprint
 from subprocess import CalledProcessError, check_output
 from unittest import mock
 from urllib.parse import urlencode, urlparse
@@ -171,9 +172,7 @@ async def test_disable_user_config(request, app, tmpdir, full_spawn):
     )
     r.raise_for_status()
     info = r.json()
-    import pprint
-
-    pprint.pprint(info)
+    pprint(info)
     assert info['disable_user_config']
     server_config = info['config']
     settings = info['settings']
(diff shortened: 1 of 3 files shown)


jupyterhub/tests/test_singleuser.py
```diff
@@ -198,6 +197,70 @@ def assert_not_in_home(path, name):
...
+@pytest.mark.parametrize("notebook_dir", ["", "~", "~/sub", "ABS"])
+async def test_notebook_dir(
+    request, app, tmpdir, user, full_spawn, extension, notebook_dir
+):
+    token = user.new_api_token(scopes=["access:servers!user"])
+    headers = {"Authorization": f"Bearer {token}"}
+
+    spawner = user.spawner
+    if extension:
+        user.spawner.environment["JUPYTERHUB_SINGELUSER_EXTENSION"] = "1"
```
manics, line 210:
```suggestion
        user.spawner.environment["JUPYTERHUB_SINGLEUSER_EXTENSION"] = "1"
```
Alternatively since the `extension` param is only used in this conditional you could set the param to `["0", "1"]` instead of `[True, False]` and get rid of the `if`

jupyterhub/tests/test_singleuser.py
```diff
@@ -198,6 +197,70 @@ def assert_not_in_home(path, name):
...
+    else:
+        user.spawner.environment["JUPYTERHUB_SINGELUSER_EXTENSION"] = "0"
```
manics, line 212:
```suggestion
        user.spawner.environment["JUPYTERHUB_SINGLEUSER_EXTENSION"] = "0"
```

consideRatio (approved):

consideRatio:
Test failure in https://github.com/jupyterhub/jupyterhub/actions/runs/5453721711/jobs/9922909198#step:11:472 still

```
  File "/opt/hostedtoolcache/Python/3.8.17/x64/lib/python3.8/runpy.py", line 185, in _run_module_as_main
    mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
  File "/opt/hostedtoolcache/Python/3.8.17/x64/lib/python3.8/runpy.py", line 144, in _get_module_details
    return _get_module_details(pkg_main_name, error)
  File "/opt/hostedtoolcache/Python/3.8.17/x64/lib/python3.8/runpy.py", line 111, in _get_module_details
    __import__(pkg_name)
  File "/home/runner/work/jupyterhub/jupyterhub/jupyterhub/singleuser/__init__.py", line 65, in <module>
    from .extension import main
  File "/home/runner/work/jupyterhub/jupyterhub/jupyterhub/singleuser/extension.py", line 35, in <module>
    from jupyter_server.auth import Authorizer, IdentityProvider, User
ImportError: cannot import name 'IdentityProvider' from 'jupyter_server.auth' (/opt/hostedtoolcache/Python/3.8.17/x64/lib/python3.8/site-packages/jupyter_server/auth/__init__.py)
```

consideRatio (approved):

minrk:
I can do a 4.0.2 backport PR with this and one or two other bugfixes. Can't release 4.0.x from main due to #4495 and a couple others.

consideRatio:
Min if the 5.0 release is made without more major-version causing changes than py38, it would enable z2jh / tljh to bump to jh 5 without going for another major release because they are guaranteed to have py38+ where hub is installed.

Is the main branch for jh including more than py38 making a major version bump relevant currently?

minrk:
I haven't checked in detail. There are a couple of new features, I don't think anything else ought to be breaking.

Do you think we should do a tiny major release? There are a few other major changes currently in review, so 6.0 would likely be not too long after with actual substantial changes.

PR #4509, opened 2023-07-13
Improve Auth module code documentation

Ph0tonic:
Improve documentation of `auth.py:authenticate` with missing overridable functions.


minrk:
Thanks!

PR #4511, opened 2023-07-14
Rename parameter for post_auth_hook to be clearer

yuvipanda:
`auth_model` matches what we call this in oauthenticator, and IMO is clearer than 'authentication'.

Ref https://github.com/2i2c-org/infrastructure/pull/2809#discussion_r1263404964

### DIFF
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -299,8 +299,9 @@ class Authenticator(LoggingConfigurable):
         This function is called after the user has passed all authentication checks
         and is ready to successfully authenticate. This function must return the
         auth_model dict reguardless of changes to it.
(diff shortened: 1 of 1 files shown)


jupyterhub/auth.py
```diff
@@ -298,27 +298,28 @@ def validate_username(self, username):
 
         This function is called after the user has passed all authentication checks
         and is ready to successfully authenticate. This function must return the
-        authentication dict reguardless of changes to it.
+        auth_model dict reguardless of changes to it.
```
minrk, line 301: By changing the name, we are making it explicit that we will pass these args _positionally_ (because if we passed it by keyword, this would be a breaking change). Maybe add that fact to the help string?

minrk (approved):

yuvipanda:
Thanks @minrk!

PR #4516, opened 2023-07-19
Add "address already in use" troubleshooting hints

sgaist:
While it seems trivial, this can be a bit convoluted to debug on macOS because some of the services might not be visible to the user logged in.

The solution is simple however knowing why it is needed is a good thing.

### DIFF
--- a/docs/source/contributing/setup.md
+++ b/docs/source/contributing/setup.md
@@ -173,3 +173,46 @@ python3 setup.py js    # fetch updated client-side js
 python3 setup.py css   # recompile CSS from LESS sources
 python3 setup.py jsx   # build React admin app
 ```
+
+### Failed to bind XXX to `http://127.0.0.1:<port>/<path>`
+
+This error can happen when there's already an application or a service using this
+port.
+
+Use the following command to find out which service is using this port.
+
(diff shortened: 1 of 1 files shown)


docs/source/contributing/setup.md
```diff
@@ -173,3 +173,25 @@ python3 setup.py js    # fetch updated client-side js
 python3 setup.py css   # recompile CSS from LESS sources
 python3 setup.py jsx   # build React admin app
 ```
+
+### Failed to bind hub to http://127.0.0.1:8081/hub/
+
+This error can happen when there's already an application or a service using this port. Use the following command to find out which one it is.
```
minrk, line 179: Worth pointing out that this applies to any port, not just 8081.

minrk (commented):
Makes sense! I'd generalize the tip to describe any port, not just the hub port which isn't any more or less likely to conflict than the others.

docs/source/contributing/setup.md
sgaist: Good point, I reworded and it made the hub an example.

minrk:
Thanks!

PR #4520, opened 2023-07-28
Support Jupyverse as a single-user server

davidbrochart:
With https://github.com/jupyter-server/jupyverse/pull/335, this will allow to use Jupyverse as a single-user server.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -183,11 +183,7 @@ jobs:
               pip install "jupyter_server==${{ matrix.jupyter_server }}"
           fi
           if [ "${{ matrix.jupyverse }}" != "" ]; then
-              pip install "git+https://github.com/davidbrochart/jupyverse.git@jupyterhub#egg=fps_auth_jupyterhub&subdirectory=plugins/auth_jupyterhub"
-              pip install "git+https://github.com/jupyter-server/jupyverse.git#egg=jupyverse_api&subdirectory=jupyverse_api" --ignore-installed
-              pip install "git+https://github.com/jupyter-server/jupyverse.git#egg=fps_kernels&subdirectory=plugins/kernels"
               pip install "jupyverse[jupyterlab,auth-jupyterhub]"
-              pip install -e .
           fi
           if [ "${{ matrix.db }}" == "mysql" ]; then
               pip install mysqlclient
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -184,6 +184,7 @@ jobs:
           fi
           if [ "${{ matrix.jupyverse }}" != "" ]; then
               pip install "jupyverse[jupyterlab,auth-jupyterhub]"
+              pip install -e .
           fi
           if [ "${{ matrix.db }}" == "mysql" ]; then
               pip install mysqlclient
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -183,6 +183,9 @@ jobs:
               pip install "jupyter_server==${{ matrix.jupyter_server }}"
           fi
           if [ "${{ matrix.jupyverse }}" != "" ]; then
+              pip install "git+https://github.com/davidbrochart/jupyverse.git@jupyterhub#egg=fps_auth_jupyterhub&subdirectory=plugins/auth_jupyterhub"
+              pip install "git+https://github.com/jupyter-server/jupyverse.git#egg=jupyverse_api&subdirectory=jupyverse_api" --ignore-installed
+              pip install "git+https://github.com/jupyter-server/jupyverse.git#egg=fps_kernels&subdirectory=plugins/kernels"
               pip install "jupyverse[jupyterlab,auth-jupyterhub]"
               pip install -e .
           fi
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -65,7 +65,7 @@ jobs:
         #   unencrypted HTTP
         #
         # main_dependencies:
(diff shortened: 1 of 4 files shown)


jupyterhub/singleuser/__init__.py
```diff
@@ -28,15 +28,20 @@
         # - extension, if jupyter server 2
         # - older subclass app, otherwise
         try:
-            import jupyter_server
+            import jupyverse_api  # noqa: F401
```
minrk, line 31: The priority here shouldn't change, which would be breaking. Let's add jupyverse as opt-in only via JUPYTERHUB_SINGLEUSER_APP=jupyverse.

minrk (changes requested):
I think it's sensible to support it, but the PR needs to be updated, since it currently changes the _default_ server to jupyverse.

The defaults and priority should be kept as they are, but adding a new value for JUPYTERHUB_SINGLEUSER_APP=jupyverse so it's opt-in seems AOK.

Make sure to add a test matrix entry with this env so it's exercised.

jupyterhub/singleuser/__init__.py
davidbrochart: Thanks, https://github.com/jupyterhub/jupyterhub/pull/4520/commits/f342dd2540fd21abff5c4ae280ee101595cdfa7b implements that, and I also moved all the Jupyverse launching logic to the Jupyverse plugin.
I will add a test to this PR when https://github.com/jupyter-server/jupyverse/pull/335 is merged and released.

jupyterhub/tests/test_singleuser.py
```diff
@@ -167,6 +173,10 @@ async def test_disable_user_config(request, app, tmpdir, full_spawn):
     )
     assert r.status_code == 200
 
+    if IS_JUPYVERSE:
+        # the remaining is jupyter-server specific
+        return
```
minrk, line 178: Rather than returning, this should verify that no jupyverse config files are loaded from $HOME.

jupyterhub/tests/test_singleuser.py
```diff
@@ -63,6 +65,8 @@ async def test_singleuser_auth(
         await user.spawn(server_name)
         await app.proxy.add_user(user, server_name)
     spawner = user.spawners[server_name]
+    if IS_JUPYVERSE:
+        spawner.default_url = "/lab"
```
minrk, line 69: What happens if this is missing? Is it because the default url for jupyverse is not `/tree`?

jupyterhub/tests/test_singleuser.py
```diff
@@ -202,6 +212,9 @@ def assert_not_in_home(path, name):
 async def test_notebook_dir(
     request, app, tmpdir, user, full_spawn, extension, notebook_dir
 ):
+    if IS_JUPYVERSE:
+        pytest.skip("jupyter-server specific test")
```
minrk, line 216: This test is not jupyter-server specific, it's verifying if the root directory config is respected. The mechanism of the test (specifically the use of the test extension utility) may need different handling, but the API calls to the contents API should produce the same results.

minrk (commented):
Thanks! There are two tests that are currently skipped that I think should be fixed instead of skipped. Once those pass, I think this should be fine to land.

jupyterhub/tests/test_singleuser.py
davidbrochart: I still think this is jupyter-server specific. Jupyverse uses the [Asphalt application framework](https://github.com/asphalt-framework/asphalt) where [configuration files](https://asphalt.readthedocs.io/en/latest/userguide/deployment.html) are passed explicitly, so there is no notion of looking up directories for configuration files.

jupyterhub/tests/test_singleuser.py
davidbrochart: Yes, just like JupyterLab has no `/tree` endpoint.

jupyterhub/tests/test_singleuser.py
minrk: Ah, then I'd just update the comment to say that jupyverse never loads any configuration from the home directory, and the skip can move to the top of the test, since there's nothing to check for jupyverse.

jupyterhub/tests/test_singleuser.py
davidbrochart: Jupyverse has no root directory config (yet), what do you think about re-enabling this test if/when there is support for it?

davidbrochart: @minrk what do you think about merging this?

.github/workflows/test.yml
```diff
@@ -176,6 +182,13 @@ jobs:
           if [ "${{ matrix.jupyter_server }}" != "" ]; then
               pip install "jupyter_server==${{ matrix.jupyter_server }}"
           fi
+          if [ "${{ matrix.jupyverse }}" != "" ]; then
+              pip install "git+https://github.com/davidbrochart/jupyverse.git@jupyterhub#egg=fps_auth_jupyterhub&subdirectory=plugins/auth_jupyterhub"
```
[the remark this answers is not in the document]
davidbrochart: > I think this PR's been merged?

Ah no, you're right. I'll make a jupyverse release and update the install command.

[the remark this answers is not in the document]
davidbrochart: @minrk it should be good to go now.

minrk:
Thanks!

davidbrochart:
Thanks, do you plan to release soon?

PR #4525, opened 2023-08-05
Add `JUPYTERHUB_METRICS_PREFIX` environment variable to customize metrics prefix

danilopeixoto:
Allow metrics prefix customization using `JUPYTERHUB_METRICS_PREFIX` environment variable.

### DIFF
--- a/docs/generate-metrics.py
+++ b/docs/generate-metrics.py
@@ -22,14 +22,8 @@ class Generator:
         for name in dir(jupyterhub.metrics):
             obj = getattr(jupyterhub.metrics, name)
             if obj.__class__.__module__.startswith('prometheus_client.'):
-                description = obj.describe()[0]
-                table_rows.append(
-                    [
-                        description.type,
-                        description.name,
-                        description.documentation,
-                    ]
-                )
+                for metric in obj.describe():
+                    table_rows.append([metric.type, metric.name, metric.documentation])
         return table_rows
 
     def prometheus_metrics(self):
--- a/docs/generate-metrics.py
+++ b/docs/generate-metrics.py
@@ -23,11 +23,13 @@ class Generator:
             obj = getattr(jupyterhub.metrics, name)
             if obj.__class__.__module__.startswith('prometheus_client.'):
                 description = obj.describe()[0]
-                table_rows.append([
-                    description.type,
-                    description.name,
-                    description.documentation,
-                ])
+                table_rows.append(
+                    [
+                        description.type,
+                        description.name,
+                        description.documentation,
+                    ]
+                )
         return table_rows
 
     def prometheus_metrics(self):
--- a/docs/generate-metrics.py
+++ b/docs/generate-metrics.py
@@ -22,8 +22,12 @@ class Generator:
         for name in dir(jupyterhub.metrics):
             obj = getattr(jupyterhub.metrics, name)
             if obj.__class__.__module__.startswith('prometheus_client.'):
-                for metric in obj.describe():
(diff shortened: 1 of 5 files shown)


danilopeixoto:
@nreith We are using the new `namespace` parameter available in the `prometheus_client>=0.5.0` metrics to assemble the names. Prometheus Client will assemble the names as previously (`jupyterhub_<metric_name>`) and no changes will be applied by default (assuming `JUPYTERHUB_METRICS_PREFIX` not set) . The string concatenation was delegated to Prometheus to avoid explicit `fstrings` and `+` operations. This also helps to document the Prometheus convention. The `jupyterhub` namespace previously hard coded in the metric name is now set by `namespace` parameter.

jupyterhub/metrics.py
```diff
@@ -30,49 +32,66 @@
 from . import orm
 from .utils import utcnow
 
+metrics_prefix = os.getenv('JUPYTERHUB_METRICS_PREFIX', 'jupyterhub')
```
minrk, line 35: It makes more sense to me to keep `jupyterhub_` in the metric names and make the default metrics prefix `''` (same as the default 'namespace' value). What do you think? i.e. custom prefix is _in addition to_ `jupyterhub_`, not _instead of_.

minrk (commented):
Makes sense! Just a few questions about implementation.

docs/generate-metrics.py
```diff
@@ -22,8 +22,14 @@ def _parse_metrics(self):
         for name in dir(jupyterhub.metrics):
             obj = getattr(jupyterhub.metrics, name)
             if obj.__class__.__module__.startswith('prometheus_client.'):
-                for metric in obj.describe():
```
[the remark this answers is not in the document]
danilopeixoto: It makes sense. Updated. I did also adapt the test to be consistent to this situation of get name, not otherwise.

jupyterhub/metrics.py
danilopeixoto: In the context of projects like [ml-tooling/ml-hub](https://github.com/ml-tooling/ml-hub) having the freedom to define the metrics prefix could be helpful for maintaining consistency within their ecosystem. While keeping `jupyterhub_` as a constant identifier in the metric names can provide a clear origin, offering full control over the metrics prefix could indeed enhance the value of this feature. Users can redefine it completely or set options like `jupyterhub_prod` and `jupyterhub_dev`. It is also common for the application name to be the namespace as the Prometheus documentation suggests.

Unfortunately, this particular option stands out as one of the limited settings that cannot be configured by Traitlets.

consideRatio:
> I agree that would be ideal, but the way prometheus-client works, metrics are instantiated and configured at import time, which makes traitlets config come too late.

Ah, that is a very good case for this config being exposed as an environment variable though!

minrk:
I dug through, and there doesn't appear to be a simple way to do it with prometheus_client, without getting into som private APIs or janky string rewriting.

To make it a traitlet config, we'd need to change the metrics from top-level import constants to something created within a function and passed around as variables. That doesn't mean we shouldn't do it that way, but I don't really want to make that change here.

manics:
@minrk Do you think the metrics should be refactored into a seperate class in a new PR first, then rebase this on top? Or merge this, and optionally remove the env-var in the refactoring PR?

minrk:
I think we should merge this as-is. If someone wants to work in exposing it in traitlets config, that's fine, but I wouldn't call it a high priority.

minrk (approved):

consideRatio (approved):

minrk:
Thanks @danilopeixoto!

PR #4534, opened 2023-08-10
Changelog for 4.0.2

minrk:
Mostly for #4503, which is the main user-facing regression we've seen in 4.0, but also improves several error messages and communications of compatibility issues with things like notebook 7 (compatibility isn't affected, but the error messages are much better).

Will backport and release once this goes in.

### DIFF
--- a/docs/source/reference/changelog.md
+++ b/docs/source/reference/changelog.md
@@ -10,6 +10,34 @@ command line for details.
 
 ## 4.0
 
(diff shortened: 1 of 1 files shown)


consideRatio (approved):
Wieee nice @minrk!

minrk:
Thanks!

minrk:

PR #4536, opened 2023-08-10
Mention NomadSpawner

### LINKED ISSUE #3860: Spawner  support for nomad
### Proposed change
NoamdSpawner

mxab:
A while back I started to work on a Nomad Spawner for Jupyterhub. I'd like to mention it in the docs here

closes #3860


minrk:
Cool! I have to admit, I'm not super motivated to promote anything for HashiCorp right now, but I think at the very least a project would need a license before we can link it (and please don't choose BSL ;)).

mxab:
Good point, I added one https://github.com/mxab/jupyterhub-nomad-spawner/blob/main/LICENSE

minrk:
Thanks!

PR #4537, opened 2023-08-10
add example for utilizing multiple docker hosts

oboki:
Based on a common dockerspawner, I created an example that can utilize multiple docker hosts.

I made it for my team, but I thought it would be helpful for the community if I registered it as an example.

Please review this.


manics:
Thanks for sharing that. We've got a repository for Docker examples:
https://github.com/jupyterhub/jupyterhub-deploy-docker

I think that's a more suitable place for it. Please add a CI test too to help with future maintenance. Thanks!

PR #4542, opened 2023-08-22
Fix include_stopped_servers in paginated next_url

### LINKED ISSUE #4524: /hub/api/users?include_stopped_servers doesn't include ?include_stopped_servers in the next_url when paginating
### Bug description

/hub/api/users?include_stopped_servers doesn't include ?include_stopped_servers in the next_url when paginating. This means the first page has stopped servers, but no subsequent pages do. I'd assume that all query params aside from offset should be replicated to the next_url.

#### Expected behaviour

?include_stopped_servers should be in the pagination url

#### Actual behaviour

?include_stopped_servers is not in the url so only the first page has stopped servers

### How to reproduce

1. On 2 different users create 2 stopped named servers.
2. Make an API Request: /hub/api/users?include_stopped_servers&limit=1
3. Follow the pagination url: /hub/api/users?limit=1&offset=1
4. No servers are listed for second user

### Your personal set up

Zero 2 jupyterhub 2.0.

jabbera:
Here is a test that demo's https://github.com/jupyterhub/jupyterhub/issues/4524. It's likely because include_stopped_servers in not key value pair and stands alone. Might I suggest include_stopped_servers=[1,0] for enabled or disabled?

```
pagination = response["_pagination"]
              if include_stopped_servers:
  >               assert "include_stopped_servers" in pagination["next"]["url"]
  E               AssertionError: assert 'include_stopped_servers' in 'http://127.0.0.1:39887/@/space%20word/hub/api/users?limit=1&offset=1'
```

closes #4524

### DIFF
--- a/jupyterhub/apihandlers/base.py
+++ b/jupyterhub/apihandlers/base.py
@@ -532,7 +532,7 @@ class APIHandler(BaseHandler):
         if next_offset < total_count:
             # if there's a next page
             next_url_parsed = urlparse(self.request.full_url())
-            query = parse_qs(next_url_parsed.query)
+            query = parse_qs(next_url_parsed.query, keep_blank_values=True)
             query['offset'] = [next_offset]
             query['limit'] = [limit]
             next_url_parsed = next_url_parsed._replace(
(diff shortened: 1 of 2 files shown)


minrk:
@meeseeksdev backport to 4.x

PR #4546, opened 2023-09-01
Fix typo in comment in spawner.py

umka1332:


consideRatio (approved):
Thank you @umka1332!!

PR #4550, opened 2023-09-05
Fix external-oauth example jupyterhub_config.py

yuvipanda:
- Roles need to be explicitly granted, otherwise you get a
  403. This example predates roles.
- Explicitly set bind_url - without this, JupyterHub itself doesn't seem to bind anywhere, and so you just get a 404 when you visit whatever port configurable-http-proxy lands on. This is probably a separate bug to be investigated, but in the meantime copying this from testing/jupyterhub_config.py makes this example actually work
- Set DummyAuthenticator as the default, so users can get started with this example

### DIFF
--- a/examples/external-oauth/jupyterhub_config.py
+++ b/examples/external-oauth/jupyterhub_config.py
@@ -1,16 +1,5 @@
 import os
 
-# Allow anyone to authenticate
-from jupyterhub.auth import DummyAuthenticator
-
-c.JupyterHub.authenticator_class = DummyAuthenticator
-
-# Optionally set a global password that all users must use
-# c.DummyAuthenticator.password = "your_password"
-
-# only listen on localhost for testing.
-c.JupyterHub.bind_url = 'http://127.0.0.1:8000'
-
 # get the oauth client's API token.
 # this could come from anywhere
 api_token = os.getenv("JUPYTERHUB_API_TOKEN")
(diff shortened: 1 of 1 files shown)


examples/external-oauth/jupyterhub_config.py
```diff
@@ -1,5 +1,16 @@
 import os
 
+# Allow anyone to authenticate
```
minrk, line 3: In some other examples that have this, I've [put it at the bottom](https://github.com/jupyterhub/jupyterhub/blob/f4163069139251c2be2d7752607518e97bd31079/examples/server-api/jupyterhub_config.py#L54-L57) to separate it as clearly the "this isn't relevant to the topic of the example, but is needed to make the example runnable". I think that makes it easier to "stop reading here..."

minrk (approved):
👍 after moving the test boilerplate to the bottom

> without this, JupyterHub itself doesn't seem to bind anywhere

That would definitely be a separate bug, but I'm not sure there is one here. This config is setting the bind URL for the proxy itself. The Hub process or what is proxied has not been affected. The default is `http://:8000`, which binds to all available interfaces. My _guess_ is that perhaps you had something running on that port already on 127.0.0.1, and binding to `''` weirdly _succeeds_ if there are any other services bound to more specific interfaces, and only gets requests not matched by another, more specific listening socket.

So you can do:

```python
listen('127.0.0.1', port)
listen('1.2.3.4', port)
listen('', port)
```

and all will succeed, but the `''` will not get any requests unless you make requests to an ip not already bound by another one.

examples/external-oauth/jupyterhub_config.py
yuvipanda: @minrk I've moved it to the bottom!

yuvipanda:
@minrk I've also opened https://github.com/jupyterhub/jupyterhub/issues/4557 with repro steps for `bind_url` issues.

PR #4552, opened 2023-09-06
Small fixes to services docs

consideRatio:


minrk (approved):

minrk:
Weird, the class directive seems to have triggered some duplicate indexing issues. That doesn't make sense to me, since it _should_ only be a reference.

yuvipanda:
I fixed it up by getting rid of the `~`. It still links to the correct place.

yuvipanda:
Thanks, @consideRatio!

PR #4555, opened 2023-09-06
fail if external oauth service lacks required oauth_redirect_uri

### LINKED ISSUE #4551: "Invalid client_id" error message when real problem is that oauth_redirect_uri is missing
### Bug description

When configuring JupyterHub to run an external service, I forgot to set `oauth_redirect_uri` in the JupyterHub service config. However, this resulted in an error message saying 'invalid client_id' for the end user, rather than 'invalid oauth_redirect_uri'. This was confusing and led us down a couple blind alleys before debugging.

#### Expected behaviour

Show an error message saying 'invalid oauth_redirect_uri'

#### Actual behaviour

A client error message saying 'invalid client_id'. No details in the JupyterHub logs.

### How to reproduce

1. Check out https://github.com/jupyterhub/jupyterhub/pull/4550 (or `main` if that PR is merged)
2. Go to `examples/external-oauth`
3. Follow instructions in the README to run the service. Note that this *works*, and you can access the external service.
4. In `examples/external-oauth/jupyterhub_config.py`, uncomment `oauth_redirect_uri` in the services config. Restart JupyterHub to load new config. In your browser, clear all cookies for `127.0.0.1:5555` by going to the 'Storage' tab under developer tools - the example oauth process doesn't have logout functionality.
5. Try accessing the the external service now.

You should get an error saying:

image

minrk:
and log service creation with oauth enabled/disabled

closes #4551

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -2501,6 +2501,11 @@ class JupyterHub(Application):
         if orm_service.oauth_client is not None:
             service.oauth_client_id = orm_service.oauth_client.identifier
             service.oauth_redirect_uri = orm_service.oauth_client.redirect_uri
+            oauth_msg = f"with ouath_client_id={orm_service.oauth_client.identifier}"
+        else:
+            oauth_msg = "without oauth"
+
+        self.log.info(f"Loaded service {service.name} from database {oauth_msg}.")
 
         self._service_map[name] = service
 
@@ -2626,6 +2631,15 @@ class JupyterHub(Application):
             service.orm.server = None
 
         if service.oauth_available:
+            self.log.info(
+                f"Creating service {service.name} with oauth_client_id={service.oauth_client_id}"
+            )
+            if not service.oauth_redirect_uri:
+                # redirect uri has a default value if a URL is configured,
+                # but must be specified explicitly for external services
+                raise ValueError(
+                    f"Service {service.name} has oauth configured, but is missing required oauth_redirect_uri."
+                )
             allowed_scopes = set()
             if service.oauth_client_allowed_scopes:
                 allowed_scopes.update(service.oauth_client_allowed_scopes)
@@ -2655,7 +2669,11 @@ class JupyterHub(Application):
             allowed_scopes.update(scopes.access_scopes(oauth_client))
             oauth_client.allowed_scopes = sorted(allowed_scopes)
         else:
+            self.log.info(f"Creating service {service.name} without oauth.")
             if service.oauth_client:
+                self.log.warning(
+                    f"Deleting unused oauth client for service {service.name} with client_id={service.oauth_client.identifier}"
+                )
                 self.db.delete(service.oauth_client)
 
         self._service_map[name] = service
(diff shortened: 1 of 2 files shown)


jupyterhub/app.py
```diff
@@ -2626,6 +2631,15 @@ def service_from_spec(
...
         if service.oauth_available:
+            self.log.info(
+                f"Creating service {service.name} with oauth_client_id={service.oauth_client_id}"
+            )
+            if not service.oauth_redirect_uri:
+                # redirect uri has a default value if a URL is configured,
+                # but must be specified explicitly for external services
+                raise ValueError(
+                    f"Service {service.name} has oauth configured, but is missing required oauth_redirect_uri."
+                )
```
consideRatio, line 2642: This is asymetric now, and I think it shouldn't be. We don't error with a nice message if `oauth_redirect_uri` is configured but `oauth_client_id` isn't now, only the other way around.

minrk: oauth_client_id can't be undefined, since it has a default value. oauth_redirect_uri has a default value unless no URL is configured (i.e. this whole situation can only come up for external but not proxied services).

jupyterhub/services/service.py
```diff
@@ -342,12 +344,20 @@ def _default_redirect_uri(self):
             return ''
         return self.host + url_path_join(self.prefix, 'oauth_callback')
 
+    @observe("oauth_client_id", "oauth_redirect_uri")
```
consideRatio, line 347: There is also the oauth config `oauth_no_confirm`, maybe it belongs here also?

minrk: Good call, added.

jupyterhub/services/service.py
```diff
@@ -342,12 +344,20 @@ def _default_redirect_uri(self):
...
+    def _oauth_config_set(self, change):
+        # record that some oauth config is specified
+        self._oauth_specified.append(change.name)
```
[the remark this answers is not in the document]
minrk: Added a check for these

jupyterhub/services/service.py
```diff
@@ -342,12 +344,20 @@ def _default_redirect_uri(self):
...
     @property
     def oauth_available(self):
         """Is OAuth available for this client?
 
         Returns True if a server is defined or oauth_redirect_uri is specified manually
         """
+        if self._oauth_specified:
+            # if any oauth config is set, oauth should be available
+            return True
         return bool(self.server is not None or self.oauth_redirect_uri)
```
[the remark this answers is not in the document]
minrk: Right, removed and replaced with _oauth_specified

jupyterhub/app.py
```diff
@@ -2501,6 +2501,11 @@ def service_from_orm(
         if orm_service.oauth_client is not None:
             service.oauth_client_id = orm_service.oauth_client.identifier
             service.oauth_redirect_uri = orm_service.oauth_client.redirect_uri
+            oauth_msg = f"with ouath_client_id={orm_service.oauth_client.identifier}"
```
yuvipanda, line 2504: nice!

yuvipanda:
Just tested, works ok! Thanks @minrk!

minrk:
@meeseeksdev backport to 4.x

Issue #4553, opened 2023-09-06
An update of the JupyterHub Services documentation would be good


consideRatio:
I think the definition of a service which leads the [JupyterHub Services docs section](https://jupyterhub.readthedocs.io/en/stable/reference/services.html) would benefit from an update.

Currently it starts out with a definition of what a service is which is great, but I think its a bit limiting. It sais that a service is something that always interacts with the JupyterHub API and provides two examples of that. I think we should try to broaden the scope.

For example, a service...

- can register an `api_token` to acquire credentials to use the jupyterhub REST API
- can register an `url` to get a proxy route `/services/<service-name>`, and a button to visit it if `display` is true
- can register a `command` to have a process started by jupyterhub

Practically this makes services usable to...

- acquire credentials
  - to do anything via the REST API if combined with configuring permissions for the service
  - to use jupyterhub as an identity provider (registering an "oauth application": the service name is the "client id", and the `api_token` is the "client secret")
- proxy traffic to some url and optionally provide a link to it
- start some process

Currently I think our service docs fails to convey all these use cases by the limiting definition and its examples.

consideRatio:
I think this can be considered closed with work done by @minrk in #4556, thank you @minrk!!!!

PR #4556, opened 2023-09-06
add some more service credential docs

minrk:
working toward #4553

@consideRatio what do you think is still missing from #4553?

### DIFF
--- a/docs/source/reference/services.md
+++ b/docs/source/reference/services.md
@@ -49,8 +49,8 @@ A Service may have the following properties:
   Skipping the confirmation page is useful for admin-managed services that are considered part of the Hub
   and shouldn't need extra prompts for login.
 - `oauth_client_id: str (default - 'service-$name')` -
-  This never needs to be set,
-  but you can specify a service's OAuth client id.
+  This never needs to be set, but you can specify a service's OAuth client id.
+  It must start with `service-`.
 - `oauth_redirect_uri: str (default: '/services/$name/oauth_redirect')` -
   Set the OAuth redirect URI.
   Required if it differs from the default proxied prefix URI,
@@ -219,7 +219,7 @@ This token enables the service to act _on behalf of_ the user.
 
 When an oauthenticated service makes a request to the Hub (or other Hub-authenticated service), it has two credentials available to authenticate the request:
 
-- the service's own api*token, which acts \_as* the service,
+- the service's own `api_token`, which acts _as_ the service,
   and is governed by the service's own role assignments.
 - the user's oauth token issued to the service during the oauth flow,
   which acts _as_ the user.
--- a/docs/source/reference/services.md
+++ b/docs/source/reference/services.md
@@ -4,20 +4,19 @@
 
 ## Definition of a Service
 
-When working with JupyterHub, a **Service** is defined as a process that interacts
-with the Hub's REST API. A Service may perform a specific
-action or task. For example, the following tasks can each be a unique Service:
+When working with JupyterHub, a **Service** is defined as something (usually a process) that can interact with the Hub's REST API.
+A Service may perform a specific action or task.
+For example, the following tasks can each be a unique Service:
 
-- shutting down individuals' single user notebook servers that have been idle
-  for some time
-- registering additional web servers which should use the Hub's authentication
-  and be served behind the Hub's proxy.
+- shutting down individuals' single user notebook servers that have been idle for some time
+- an additional web application which uses the Hub as an OAuth provider to authenticate and authorize user access
+- a script run once in a while, which performs any API action
+- automating requests to running user servers, such as activity data collection
 
-Two key features help define a Service:
+Two key features help differentiate Services:
 
 - Is the Service **managed** by JupyterHub?
-- Does the Service have a web server that should be added to the proxy's
-  table?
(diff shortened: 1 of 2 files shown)


docs/source/reference/services.md
```diff
@@ -30,15 +29,16 @@ Currently, these characteristics distinguish two types of Services:
 A Service may have the following properties:
 
 - `name: str` - the name of the service
-- `admin: bool (default - false)` - whether the service should have
-  administrative privileges
-- `url: str (default - None)` - The URL where the service is/should be. If a
-  url is specified for where the Service runs its own web server,
+- `url: str (default - None)` - The URL where the service is/should be.
+  If a url is specified for where the Service runs its own web server,
```
consideRatio, line 33: I don't understand the URL description so well.

> If a url is specified for where the Service runs its own web server

docs/source/reference/services.md
```diff
@@ -30,15 +29,16 @@ Currently, these characteristics distinguish two types of Services:
...
-- `api_token: str (default - None)` - For Externally-Managed Services you need to specify
-  an API token to perform API requests to the Hub
+- `api_token: str (default - None)` - For Externally-Managed Services,
+  you need to specify an API token to perform API requests to the Hub.
+  For Hub-managed services, this token is generated at startup,
+  and available via `$JUPYTERHUB_API_TOKEN`.
+  For OAuth services, this is the client secret.
 - `display: bool (default - True)` - When set to true, display a link to the
-  service's URL under the 'Services' dropdown in user's hub home page.
+  service's URL under the 'Services' dropdown in users' hub home page.
```
consideRatio, line 41: display is related to url, without an url, display is irrelevant I think.

docs/source/reference/services.md
```diff
@@ -30,15 +29,16 @@ Currently, these characteristics distinguish two types of Services:
 A Service may have the following properties:
```
consideRatio, line 29: This list doesn't mention `oauth_client_id` or `oauth_redirect_uri` yet

minrk: Good point, Added.

docs/source/reference/services.md
minrk: Added that note

docs/source/reference/services.md
minrk: Took another stab at it. Let me know what you think.

docs/source/reference/services.md
```diff
@@ -30,44 +29,54 @@ Currently, these characteristics distinguish two types of Services:
...
   and shouldn't need extra prompts for login.
+- `oauth_client_id: str (default - 'service-$name')` -
+  This never needs to be set,
+  but you can specify a service's OAuth client id.
+- `oauth_redirect_uri: str (default: '/services/$name/oauth_redirect')` -
+  Set the OAuth redirect URI.
+  Required if it differs from the default proxied prefix URI,
+  e.g. the redirect handler is not at `/$prefix/oauth_redirect`,
+  and/or the service is not to be added to the proxy at `/services/$name`
+  (i.e. `url` is not set, but there is still a public web service using OAuth).
```
consideRatio, line 59: - Is `$prefix` the same as `services/$name`? I think we should avoid introducing `$prefix` and just be explicit about that if so.
- `/services/$name` is written here, but its written as `/services/:name` above

docs/source/reference/services.md
```diff
@@ -30,44 +29,54 @@ Currently, these characteristics distinguish two types of Services:
...
   skip the OAuth confirmation page when users access this service.
-
   By default, when users authenticate with a service using JupyterHub,
   they are prompted to confirm that they want to grant that service
   access to their credentials.
   Skipping the confirmation page is useful for admin-managed services that are considered part of the Hub
   and shouldn't need extra prompts for login.
+- `oauth_client_id: str (default - 'service-$name')` -
+  This never needs to be set,
+  but you can specify a service's OAuth client id.
```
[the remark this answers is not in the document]
consideRatio:
```suggestion
- `oauth_client_id: str (default - 'service-$name')` -
  This never needs to be set, but you can specify a service's OAuth client id.
  It must start with `service-`.
```

docs/source/reference/services.md
consideRatio: I think this section could be omitted as it confused me more than helped me understand.

> proxied prefix URI, e.g. the redirect handler is not at `/$prefix/oauth_redirect`

docs/source/reference/services.md
```diff
@@ -174,6 +183,106 @@ c.JupyterHub.services = [
...
+
+- owned by the authenticated user
+- associated with the oauth client of the service
+- governed by the service's `oauth_client_allowed_scopes` configuration
+
+This token enables the service to act _on behalf of_ the user.
+
+When an oauthenticated service makes a request to the Hub (or other Hub-authenticated service), it has two credentials available to authenticate the request:
+
+- the service's own api*token, which acts \_as* the service,
```
consideRatio, line 222: Formatting issue here I think

consideRatio (commented):
I just read through the service credentials examples and I think they are great!!! I've slowly gained more and more understanding about this, and this helped me understad the permission system a bit better once more.

consideRatio:
❤️ 🎉 wieee thank you for working this!!

PR #4560, opened 2023-09-11
singleuser extension: persist token from ?token=... url in cookie

minrk:
matches pre-extension behavior

consolidated code to HubAuth, so it's easier to re-use, but logic is unchanged.

Reported [on the forum](https://discourse.jupyter.org/t/websocket-http-error-403/21141/8)

### DIFF
--- a/jupyterhub/services/auth.py
+++ b/jupyterhub/services/auth.py
@@ -686,6 +686,31 @@ class HubAuth(SingletonConfigurable):
         """Check whether the user has required scope(s)"""
         return check_scopes(required_scopes, set(user["scopes"]))
 
+    def _persist_url_token(self, handler):
+        """Persist ?token=... from URL in cookie
+
+        for use in future cookie-authenticated requests.
+
+        Allows initiating an authenticated session
+        via /user/name/?token=abc...,
+        otherwise only the initial request will be authenticated.
+        """
+        url_token = handler.get_argument('token', '')
+        if not url_token:
+            # no token to persist
+            return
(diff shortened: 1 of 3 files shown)


consideRatio (approved):

consideRatio (approved):
I reflected a bit about `_persist_url_token` checking if `token` query param was passed (and more), but that the code invoking it checked if `token` query param was passed as well (but not more).

I don't arrive at a suggestion though, so I think we should go for a merge - I'll let you do it just in case you arrived at something you thought would be an improvement.

minrk:
Yes, it's unclear to me as well! That's why the check is in both places. The function is really "persist url token if there is one and it's the one that ultimately determined authentication", but the name suggested to me that I should check if there's a token to persist before calling it.

I decided to rename it `_persist_url_token_if_set` which made me more comfortable leaving the conditions all inside the method.

minrk:
@meeseeksdev backport to 4.x

kzgrzendek:
@minrk thank you for solving that issue that caused imo random latencies on our side bacause of websocket authentication issues popping in our logs.

Unfortunately, with Zero2JupyterHub on a vanilla K8S, the workaround you're proposing makes our spawned lab crashing at the start.

I see that your patch has been merged after the latest JHub release ; do you have an idea about when an official JHub realease taking this fix in account will be available?

Thanks again for your work! :)

minrk:
I'm working on the release now, hopefully soon.

kzgrzendek:
> I'm working on the release now, hopefully soon.

That's perfect, thank you!

Issue #4557, opened 2023-09-06
JupyterHub route not added to CHP when `bind_url` is not set


yuvipanda:
### Bug description

When `JupyterHub.bind_url` is not set, jupyterhub doesn't seem to add a default route to CHP to itself. This leads to a 404 when accessing the URL where CHP is listening, as there's no route to JupyterHub.

### How to reproduce

1. Open `testing/jupyterhub_config.py` file, and comment out the line that sets `bind_url`
2. Run `jupyterhub -f testing/jupyterhub_config.py`
3. Access jupyterhub via CHP at `127.0.0.1:8000`

#### Expected behaviour

JupyterHub login screen shows up

#### Actual behaviour

There's a 404, as there is no route. 

Logs from JupyterHub:

```
➜ python -m jupyterhub -f testing/jupyterhub_config.py
[I 2023-09-06 10:29:10.848 JupyterHub app:3065] Running JupyterHub version 5.0.0.dev[I 2023-09-06 10:29:10.848 JupyterHub app:3095] Using Authenticator: jupyterhub.auth.DummyAuthenticator-5.0.0.dev
[I 2023-09-06 10:29:10.848 JupyterHub app:3095] Using Spawner: jupyterhub.spawner.SimpleLocalProcessSpawner-5.0.0.dev
[I 2023-09-06 10:29:10.848 JupyterHub app:3095] Using Proxy: jupyterhub.proxy.ConfigurableHTTPProxy-5.0.0.dev
[I 2023-09-06 10:29:10.850 JupyterHub app:1734] Loading cookie_secret from /Users/yuvipanda/code/jupyterhub/jupyterhub_cookie_secret
[I 2023-09-06 10:29:10.872 JupyterHub proxy:556] Generating new CONFIGPROXY_AUTH_TOKEN
[I 2023-09-06 10:29:10.875 JupyterHub app:2054] Not using allowed_users. Any authenticated user will be allowed.
[I 2023-09-06 10:29:10.900 JupyterHub app:3134] Initialized 0 spawners in 0.003 seconds
[I 2023-09-06 10:29:10.903 JupyterHub metrics:303] Found 0 active users in the last ActiveUserPeriods.twenty_four_hours
[I 2023-09-06 10:29:10.903 JupyterHub metrics:303] Found 0 active users in the last ActiveUserPeriods.seven_days
[...]

PR #4561, opened 2023-09-12
Improve debugging when waiting for servers

minrk:
Add debug messages and timers for start and end waiting for servers

and improve logic for awaiting proxy endpoints using concurrency primitives instead of a for-loop

this might have improved debugging of #4557

Sample debug output:

```
[D 2023-09-12 09:27:23.233 JupyterHub utils:251] Waiting 10s for server at heavy:8000
[D 2023-09-12 09:27:23.243 JupyterHub utils:98] Server at heavy:8000 not ready: [Errno 61] Connection refused
[D 2023-09-12 09:27:23.243 JupyterHub utils:251] Waiting 10s for server at 127.0.0.1:8001
[D 2023-09-12 09:27:23.243 JupyterHub utils:98] Server at 127.0.0.1:8001 not ready: [Errno 61] Connection refused
09:27:23.327 [ConfigProxy] info: Proxying http://*:8000 to (no default)
09:27:23.329 [ConfigProxy] info: Proxy API at http://127.0.0.1:8001/api/routes
[D 2023-09-12 09:27:23.384 JupyterHub utils:261] Server at 127.0.0.1:8001 responded in 0.14s
[D 2023-09-12 09:27:23.416 JupyterHub utils:261] Server at heavy:8000 responded in 0.18s
[D 2023-09-12 09:27:23.416 JupyterHub proxy:799] Proxy started and appears to be up
```

### DIFF
--- a/jupyterhub/proxy.py
+++ b/jupyterhub/proxy.py
@@ -48,7 +48,7 @@ from jupyterhub.traitlets import Command
 from . import utils
 from .metrics import CHECK_ROUTES_DURATION_SECONDS, PROXY_POLL_DURATION_SECONDS
 from .objects import Server
-from .utils import AnyTimeoutError, exponential_backoff, url_escape_path, url_path_join
+from .utils import exponential_backoff, url_escape_path, url_path_join
 
 
 def _one_at_a_time(method):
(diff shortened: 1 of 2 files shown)


consideRatio (commented):
I think the logic is correct, but I had comments about readability

- The function no longer has a `_` prefix, shouldn't it have that still?
- name and docstring didn't help me understand what the function did so well, and with the function not being prefixed with `_` I think we should make sure to get this tightly defined.
  - based on name/docstring, I was surprised to see the externally defined servers_ready awaited
- A variable name `process_ended` doesn't meet my expectations, currently understanding it to represents that the watching of the proxy initialization had ended or similar.
- The comment `# whichever comes first` wasn't representing what `wait_for -> gather (A, b)` does, but its kind of true anyhow because to await A is to await `a` or `b`.

I'm not sure if this would work out correctly, but with regards to the "wait for first", maybe we can use:
- one awaitable for early termination of proxy process (sleeps 0.5 second between poll)
- one awaitable for servers ready
- use of `asyncio.wait` with a `return_when` argument set to `FIRST_COMPLETED`, more info in reference https://docs.python.org/3/library/asyncio-task.html#waiting-primitives
  - manually cancelling waited futures if returned or timeout...
  
Hmmm... I'm not sure, we can merge this as it if but I think this was quite tricky to understand and therefore maintain longterm as it is now.

minrk:
Thanks for the feedback! I've pushed an updated implementation that uses the lower-level `wait` instead of relying on `wait_for`, I think it should be more clear and explicit, if more complex in terms of cases to handle explicitly.

> The function no longer has a _ prefix, shouldn't it have that still?

I don't think it makes sense to use private naming prefixes for functions defined in closures. Nobody can ever call this function so it's not an API. If Python had anonymous blocks, that's what I think would be most appropriate here.

> A variable name process_ended doesn't meet my expectations, currently understanding it to represents that the watching of the proxy initialization had ended or similar.

It is a Future that resolves only when the process has ended, but cancels itself when it's no longer needed. That cancelling itself is weird, I admit! I'll look at avoiding it.

> The comment `# whichever comes first` wasn't representing what wait_for -> gather (A, b) does, but its kind of true anyhow because to await A is to await a or b.

I think this accurately describes what happens here, but here's a more detailed description of the possible situations:

1. servers wait_up complete with success (wait_for returns immediately)
2. process exits before both wait_ups finish (wait_for raises RuntimeError immediately)
3. _either_ wait_up raises an error (wait_for raises error immediately)
4. one or both wait_up doesn't finish within the deadline (wait_for raises TimeoutError)

and the use of `wait_for` ensures cancellation of whichever task doesn't finish first.

I can rewrite it to use `wait` instead of `wait_for`, but then we are reimplementing all the logic of wait_for (cancel pending tasks, raise TimeoutError, etc.). Maybe that would be clearer and more explicit, though.

consideRatio:
Yikes this cascaded to significant additional work for you, I'm sorry! Thank you for the clear followups and effort into this.

> I don't think it makes sense to use private naming prefixes for functions defined in closures. Nobody can ever call this function so it's not an API. If Python had anonymous blocks, that's what I think would be most appropriate here.

Oh I didn't realize this was just defined inside another function, makes sense!

> I can rewrite it to use wait instead of wait_for, but then we are reimplementing all the logic of wait_for (cancel pending tasks, raise TimeoutError, etc.). Maybe that would be clearer and more explicit, though.

Ah yes this certainly added complexity as well.

minrk:

PR #4562, opened 2023-09-12
Use `user.stop` to cleanup spawners that stopped while Hub was down

### LINKED ISSUE #4519: Behavior if server has stopped while hub was down (no Spawner.stop call)
Hi everyone,

JupyterHub checks during the startup if a server is still running. If this is not the case, JupyterHub does not [call](https://github.com/jupyterhub/jupyterhub/blob/3fbdf02cc5c6574432dd28710ed2fc64b2353081/jupyterhub/app.py#L2607) `user.stop(name)` . The server will be set to `None` and removed from the database.  
  
This can cause problems, because `Spawner.post_stop_hook` will never be called in this scenario. One might want to use the configurable `post_stop_hook` as a custom cleanup function, to revert things done previously in `Spawner.pre_spawn_hook`. Or is there a better place to do some cleanup afterwards?  
  
An alternative would be to implement the cleanup in `Spawner.poll` (for `Spawner` subclasses), since this is the only function called during the startup check.

minrk:
for more consistent behavior with failed start or crash while Hub is up

closes #4519

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -2785,9 +2785,13 @@ class JupyterHub(Application):
                         "%s appears to have stopped while the Hub was down",
                         spawner._log_name,
                     )
-                    # remove server entry from db
-                    db.delete(spawner.orm_spawner.server)
-                    spawner.server = None
+                    try:
+                        await user.stop(name)
+                    except Exception:
+                        self.log.exception(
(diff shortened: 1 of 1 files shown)


consideRatio (approved):

minrk:
@meeseeksdev backport to 4.x

PR #4563, opened 2023-09-12
only set 'domain' field on session-id cookie

minrk:
this is the only cookie we share with subdomains on purpose, rather than setting it on all cookies. We used to do this on more cookies before everything was oauth.

Notably, domain is the _reverse_ of path: setting domain explicitly [shares the cookie with subdomains](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#define_where_cookies_are_sent), whereas omitting domain restricts it to the setting host, which is what we want for everything but the session id.

Cookie paths should generally handle the restriction already, but it seems good to limit where we set these things.


minrk:

PR #4058, opened 2022-10-01
c.JupyterHub.port indicated in the docs

BhavyaT-135:
Fixes: #4020

PR #4564, opened 2023-09-12
undeprecate JupyterHub.ip/port/base_url

### LINKED ISSUE #4020: Documentation issue: c.JupyterHub.port is deprecated but not indicated in doc

minrk:
Setting the full bind_url seems increasingly cumbersome, as often one only wants to change the url prefix or the ip, rather than configuring the whole endpoint, which requires _looking up_ the other defaults to avoid changing more things. This is partly motivated by seeing docs PRs to address the deprecation, which I think make things _less_ clear.

There's no strong reason to deprecate these options other than the 'only one way to do things' and I think allowing the configuration of base_url or ip without needing to configure the other is more important than the complexity of being able to set them in two ways.

closes #4042
closes #4058
closes #4020

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -624,9 +624,6 @@ class JupyterHub(Application):
         This is the address on which the proxy will listen. The default is to
         listen on all interfaces. This is the only address through which JupyterHub
         should be accessed by users.
-
-        .. deprecated: 0.9
-            Use JupyterHub.bind_url
         """,
     ).tag(config=True)
 
@@ -637,9 +634,6 @@ class JupyterHub(Application):
         This is the port on which the proxy will listen.
         This is the only port through which JupyterHub
         should be accessed by users.
-
-        .. deprecated: 0.9
-            Use JupyterHub.bind_url
         """,
     ).tag(config=True)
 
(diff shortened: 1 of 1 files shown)


yuvipanda (approved):
Me likey, especially as this doesn't actually increase *current* complexity.

I'll leave this open for a day or two, happy to merge after if nobody objects

consideRatio (approved):

isaprykin:
thank you @manics

Issue #4566, opened 2023-09-13
Jupyterhub Version Check does not work in k8s helm chart deploy


jabbera:
### Bug description

Every time a server starts I see this:

```
  [D 2023-09-13 03:06:09.612 JupyterHub utils:278] Server at http://10.15.40.229:8888/user/mbarry/ responded with 302
> [W 2023-09-13 03:06:09.612 JupyterHub _version:37] Single-user server has no version header, which means it is likely < 0.8. Expected 4.0.2
```

This is because the 302 redirect doesn't include the header X-JupyterHub-Version as defined here: https://github.com/jupyterhub/jupyterhub/blob/f4426ae0df4bff35bdb809a1c447d6707b979cc8/jupyterhub/user.py#L935

### How to reproduce

Install helm chart 3.0.2 and start a server

Notice the warning. Curl the same URL as the log entry:

```
curl http://10.15.40.229:8888/user/mbarry/ --verbose

*   Trying 10.15.40.229:8888...
* Connected to 10.15.40.229 (10.15.40.229) port 8888 (#0)
> GET /user/mbarry/ HTTP/1.1
> Host: 10.15.40.229:8888
> User-Agent: curl/8.0.1
> Accept: */*
>
< HTTP/1.1 302 Found
< Server: TornadoServer/6.3.3
< Content-Type: text/html; charset=UTF-8
< Date: Wed, 13 Sep 2023 03:36:17 GMT
< Location: /user/mbarry/lab/tree/home/mbarry?
< Content-Length: 0
<
* Connection #0 to host 10.15.40.229 left intact
```

Notice no header. You need to follow the redirect to get the header with the version:

```
*   Trying 10.15.40.229:8888...
* Connected to 10.15.40.229 (10.15.40.229) port 8888 (#0)
> GET /user/mbarry/lab/tree/home/mbarry? HTTP/1.1
> Host: 10.15.40.229:8888
> User-Agent: curl/8.0.1
> Accept: */*
>
< HTTP/1.1 302 Found
< Server: TornadoServer/6.3.3
< Content-Type: text/html; charset=UTF-8
< Date: Wed, 13 Sep 2023 03:40:08 GMT
< X-Content-Type-Options: nosniff
< X-Jupyterhub-Version: 4.0.2
< Content-Security-Policy: frame-ancestors 'self'; report-uri /user/mbarry/api/security/csp-report
```

#### Expected behaviour

Version check returns 4.0.2

#### Actual behaviour

[...]

PR #4567, opened 2023-09-13
Use {base_url}/api for checking hub version

jabbera:
Disclosure: I've not actually run this.

Fixes: https://github.com/jupyterhub/jupyterhub/issues/4566

### DIFF
--- a/jupyterhub/objects.py
+++ b/jupyterhub/objects.py
@@ -166,7 +166,7 @@ class Server(HasTraits):
             name=self.__class__.__name__, url=self.url, bind=self.bind_url
         )
 
-    def wait_up(self, timeout=10, http=False, ssl_context=None):
+    def wait_up(self, timeout=10, http=False, ssl_context=None, extra_path=""):
         """Wait for this server to come up"""
         if http:
             ssl_context = ssl_context or make_ssl_context(
@@ -174,7 +174,7 @@ class Server(HasTraits):
             )
 
             return wait_for_http_server(
-                urljoin(self.url, "api"), timeout=timeout, ssl_context=ssl_context
+                urljoin(self.url, extra_path), timeout=timeout, ssl_context=ssl_context
             )
         else:
             return wait_for_server(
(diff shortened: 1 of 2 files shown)


minrk:
Rather than following redirects, perhaps we should fix the URL to be `{base_url}/api/`?

jabbera:
@minrk all set.

jupyterhub/objects.py
```diff
@@ -174,7 +174,7 @@ def wait_up(self, timeout=10, http=False, ssl_context=None):
             )
 
             return wait_for_http_server(
-                self.url, timeout=timeout, ssl_context=ssl_context
+                urljoin(self.url, "api"), timeout=timeout, ssl_context=ssl_context
```
minrk, line 177: This `/api/` bit is specific to waiting for user servers, which the Server is for any http server, so I think either `wait_up` needs to add a path argument or the Server on which `wait_up` is called needs to be constructed via `Server.from_url(urljoin(server.url, "api")` to add it to `self.url`.

jabbera:
@minrk I tried to grok the self.url change but the extra_path on wait_up seemed less risky and more generally useful as an api feature.

jupyterhub/objects.py
```diff
@@ -3,7 +3,7 @@
 # Distributed under the terms of the Modified BSD License.
 import socket
 import warnings
-from urllib.parse import urlparse, urlunparse
+from urllib.parse import urljoin, urlparse, urlunparse
```
minrk, line 6: Let's use `jupyterhub.utils.url_path_join` here, which I think is more likely to have the desired effect. `urljoin` behaves like relative URL resolution on a webpage, whereas `url_path_join` does a simpler path-style join.

For example:

```python
In [9]: urljoin("a/b", "c")
Out[9]: 'a/c'

In [10]: url_path_join("a/b", "c")
Out[10]: 'a/b/c'
```

minrk:
extra_path makes sense! One minor comment about what join function to use, but 👍 after that.

PR #4571, opened 2023-09-15
TST: apply lower-bound pins during install, rather than later

minrk:
helps pip make a valid env, since it doesn't downgrade packages broken by pin-down.

I think this will fix the oldest_dependencies test, which is not actually getting its pins applied, as the `install notebook` is picking up several more recent packages, including upgrading traitlets.


consideRatio (approved):
Test failure expected? Code changes lgtm!

minrk:
Sort of. The failure is because by doing these two installs at the same time, we're now revealing an actual conflict (`jupyterhub[test]` has `jupyterlab>=3`, while `oldest_dependencies` pins down some packages like importlib_metadata that are actually too old for jupyterlab 3.0.0 and some of its dependencies). This is tricky to resolve, because we use a single environment for whole test suite. Up to now, pip was happily creating an env with broken dependencies, because it only tries to solve the requested install, it ignores already-installed packages, and only warns about conflicts after the fact.

So it is right that the oldest_dependencies install fails, but I'm not quite sure what the fix should be.

Backing up a bit, the goal of the `oldest_dependencies` test entry is to verify that we don't introduce any assumptions in our code that dependencies are newer than our declared oldest-supported versions. So we try to (as maintainably as possible) install an environment where the actual lowest-satisfying versions of our `>=` dependencies are installed. This works fine for direct dependencies that aren't shared by any of our test dependencies. But we are actually pinning down some shared dependencies (importlib_metadata and jinja2 being some).

This is creating exactly the "never pin down only direct dependencies" problem, so maybe what we should do is have a fully frozen environment for the 'oldest depenencies' case, instead of the `sed`-based pin-down logic that we have now.

consideRatio (approved):

consideRatio (approved):

PR #4573, opened 2023-09-15
temp ci fix: pin down traitlets in oldest_dependencies test

minrk:
because the bigger fix is proving more complicated (#4571), this should get CI passing in the meantime.


minrk:
closed in favor of #4571

PR #4578, opened 2023-09-25
Improve requests for tokens with scopes

### LINKED ISSUE #4576: How to create token with limited permissions?
The [Assigning permissions to a token](https://jupyterhub.readthedocs.io/en/stable/howto/rest.html#assigning-permissions-to-a-token) section of the documentation says:

> In JupyterHub 2.0, specific permissions are now defined as ‘scopes’, and can be assigned both at the user/service level, and at the individual token level.
> 
> This allows e.g. a user with full admin permissions to request a token with limited permissions.

But the documentation on _how_ to actually obtain a token with limited permissions/scopes is missing. The web UI allows to set only the expiration date and it also does not show which tokens have which permissions/scopes.

minrk:
- Docs update with specific example requesting token with scopes via API (closes #4576)
- Handler update for better error handling of invalid scopes (some invalid inputs would register as 500 instead of 400)
- Add input for permissions in token request form
- Add permissions column to token tables

I had been holding off on permissions inputs in the form since it really _should_ be a nice checkbox-list form of scopes with explanations (much like the permissions form on the oauth authorization page). This is tricky to get right, especially when exposing all the filter options, like !user, !user=someoneelse, !group=g, !server, !server=specific. But rather than waiting for that hard problem to be solved before we have _any_ way to request tokens with scopes, do the simple thing of a free text field so folks can type the scopes they want. Not the nicest, but at least it's fully capable, and the nice UI can be a later enhancement.

https://github.com/jupyterhub/jupyterhub/assets/151929/1c2c32dd-03cc-4e99-ab6a-f3323cc4477c

### DIFF
--- a/docs/source/_static/rest-api.yml
+++ b/docs/source/_static/rest-api.yml
@@ -552,7 +552,11 @@ paths:
         - oauth2:
             - read:tokens
     post:
-      summary: Create a new token for the user
+      summary: |
+        Create a new token for the user.
+        Permissions can be limited by specifying a list of `scopes` in the JSON request body
+        (starting in JupyerHub 3.0; previously, permissions could be specified as `roles` could be specified,
+        which is deprecated in 3.0).
       parameters:
         - name: name
           in: path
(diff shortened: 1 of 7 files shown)


minrk:
linkcheck errors fixed by #4586

PR #4596, opened 2023-10-03
fix package_data in docker images

minrk:
- verify package_data and data_files installation in docker build stage
- include .git in build stage, setuptools_scm section to ensure setuptools_scm gets its file metadata correct

closes #4593

PR #4597, opened 2023-10-03
update nodesource installation in docker

minrk:
the setup script has been deprecated and adds a 60s wait

includes #4596 to avoid conflict on  `git` vs `gnupg` in the apt install list

updates node runtime to 20 in the build. This should be inconsequential since it doesn't affect the final image, which doesn't use nodesource.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,6 +45,7 @@ RUN apt-get update -qq \
     ca-certificates \
     curl \
     git \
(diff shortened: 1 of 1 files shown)


consideRatio (approved):

consideRatio:
LGTM, feel free to merge if expected tests pass!

PR #4598, opened 2023-10-03
jsx: trade yarn for npm

minrk:
we're already using npm at one level, it's a bit confusing to use a different package manager in a subdirectory. Yarn doesn't really give us anything that npm doesn't, so be consistent.

For consistency, stop blocking `package-lock.json` from git, now that we are dropping yarn.lock.


jsx/README.md
```diff
@@ -7,11 +7,11 @@ admin dashboard codebase.
...
 
-- `yarn build`: Installs all dependencies and bundles the application
-- `yarn hot`: Bundles the application and runs a mock (serverless) version on port 8000
-- `yarn lint`: Lints JSX with ESLint
-- `yarn lint --fix`: Lints and fixes errors JSX with ESLint / formats with Prettier
-- `yarn place`: Copies the transpiled React bundle to /share/jupyterhub/static/js/admin-react.js for use.
+- `npm run build`: Installs all dependencies and bundles the application
+- `npm run hot`: Bundles the application and runs a mock (serverless) version on port 8000
+- `npm run lint`: Lints JSX with ESLint
+- `npm run lint --fix`: Lints and fixes errors JSX with ESLint / formats with Prettier
```
minrk, line 13: I want to trade the lint steps here for pre-commit, but not in this PR.

yuvipanda:
I ran `python -m build` to test this and verified that npm is invoked `jupyterhub-5.0.0.dev0/share/jupyterhub/static/js/admin-react.js` is part of the built tarball.

minrk:
@yuvipanda thanks! We also have a [test](https://github.com/jupyterhub/jupyterhub/blob/70717dc9ab163b568c51f70369374d35b91ba56f/ci/check_installed_data.py#L18) for that, having mixed it up a few times in the past.

PR #4602, opened 2023-10-13
Change `db_url` schema in docs from `postgres` to `postgresql`

johncf:
SQLAlchemy no longer supports `postgres` schema


yuvipanda:
Thanks for the PR, @johncf! The test failure is unrelated and I have opened https://github.com/jupyterhub/jupyterhub/pull/4603 to fix it.

PR #4608, opened 2023-10-19
move service oauth state from cookies to memory

minrk:
only use a random token as the actual oauth state, and use a local cache dict to store the extra info like cookie_name, next_url, anything else we might want.

This is related to https://github.com/jupyterhub/oauthenticator/pull/671 which makes a similar change, but only modifying the _url parameter_ state, not the cookie state.

this avoids the state field getting too big and passing local browser-server info to anyone else. 

The only downside of storing this extra info in memory is that it won't work for multiple replicas, but I don't think services can have multiple replicas. We could take the _exact_ approach in https://github.com/jupyterhub/oauthenticator/pull/671, which does work for multiple replicas.

Prompted by [this discussion](https://discourse.jupyter.org/t/error-431-headers-too-big/21833/4) where our fairly large state cookies are not getting cleared (cause still unknown), and which motivated a few other changes:

- prevent expiration-related cookie_options from overriding the state cookie's short life

Related to that discourse topic, I considered also purging _all_ state cookies upon successful completion of login. This would change one test which exercises collisions between multiple concurrent logins, but it is perhaps a good thing to be able to say "no oauth state cookies will be set after successful login, no matter how many oauth flows have started", but that could end up reintroducing #1438.


yuvipanda:
> but I don't think services can have multiple replicas

I don't *fully* understand this I think - does this mean external services can't be HA?

minrk:
> I don't fully understand this I think - does this mean external services can't be HA?

Not in general, but if they use specifically this bundled implementation of oauth, then some features won't work.

If you have two replicas of a service and there's no stickiness on routing of requests:

- [replica 1] visit url, not authenticated. stores request url as next_url associated with state id _in memory_ (previously in oauth state itself), begins oauth redirect
- [hub as oauth provider] oauth authorized, redirect back to service
- [replica 2] receives oauth redirect, completes oauth, looks up next_url for state_id, finds nothing, sends to default url

So when oauth completes, it should still _succeed_, but you won't get redirected to the original visit URL, you'll end up at the default URL.

If the service uses any standard oauth implementation, it will be unaffected and work fine (and probably ignore the requested URL, as most do, which bugs the heck out of me!)

minrk:
This now clears all oauth cookies on completion of any oauth flow, which should prevent leftover oauth state cookies in some cases.

If two simultaneously outstanding oauths both complete (I think this is quite unlikely), instead of the _second_ one taking precedence, the _first_ one will. The main thing we need to deal with is multiple oauth flows _starting_ preventing any from completing (i.e. background requests starting oauth redirects that will never finish, which caused #1438)

jupyterhub/services/auth.py
```diff
@@ -854,37 +904,32 @@ async def _token_for_code(self, code):
...
+        
+        Governs both oauth state cookie Max-Age,
+        as well as the in-memory _oauth_states cache.
+        """,
+    )
+    _oauth_states = Instance(
+        _ExpiringDict,
+        allow_none=False,
+        help="""
+        Store oauth state info, such as next_url
```
sgaist, line 922: Should we have something like `Store oauth state info per service` to make things clearer ?

sgaist (approved):
The implementation looks sane beside the minor property documentation question.

consideRatio:
@minrk merge conflict resolution needed, I'll try to get reviews in tomorrow of this and other PRs!

consideRatio:
@minrk this is a tradeoff PR since we no longer handle next_url reliably for a service with multiple replicas relying on this implementation - right? Due to this, maybe a note or similar can be added somewhere about it?

minrk:
Yeah, it is. We could consider making the state persistence overrideable public methods. But if folks think next_url working for multiple replicas is important, then I think we should probably make the default implementation work instead of making it overrideable.

We can do that, we just need to persist in cookies (which potentially reintroduces the "too-big cookie" problem), and possibly get better at preemptively clearing state cookies instead of clearing only on success.

I do think the clear-on-success behavior here will already improve that situation.

consideRatio:
I think this this is an improvement overall, let's go with it as it is now with the note added!

consideRatio:
Thank you for contributing with reviews @sgaist!!!!!!!!! ❤️ 🎉 🌻 👍

jupyterhub/services/auth.py
minrk: added more detail here

PR #4615, opened 2023-10-26
add event_loop_interval_seconds metric

### LINKED ISSUE #4353: Can we with a metric measure if jupyterhub is CPU constrained?

minrk:
attempts to measure event loop responsiveness (closes #4353)

uses a Histogram to track how many iterations fit in a given bucket.

this collects more information than the dask average-tick-duration version by counting every tick, rather than the average every second. The same average can be computed from the histogram with `sum() / count()`, but the resolution of the average will be the prometheus collection window, rather than the internal collection window (1s for dask).

I could go with matching the dask metric exactly, which is simpler to visualize as it's a simple line, or do both, but I opted for the histogram because I think 'bubbles' of blocking activity are more likely to be relevant for us, and are not well represented by the dask approach. The dask approach captures very well lots of small tasks adding up, but not very well fewer large tasks having a big impact. Those will be represented more clearly in a histogram.

Question:

- use dask average instead of (or in addition to) histogram?
- custom buckets for histogram?

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1665,6 +1665,8 @@ class JupyterHub(Application):
         config=True,
     )
 
+    metrics_collector = Any()
+
     def init_handlers(self):
         h = []
         # load handlers from the authenticator
@@ -3180,8 +3182,10 @@ class JupyterHub(Application):
                 await self.proxy.check_routes(self.users, self._service_map)
 
             asyncio.ensure_future(finish_init_spawners())
-        metrics_updater = PeriodicMetricsCollector(parent=self, db=self.db)
-        metrics_updater.start()
+        metrics_collector = self.metrics_collector = PeriodicMetricsCollector(
+            parent=self, db=self.db
+        )
+        metrics_collector.start()
 
     async def cleanup(self):
         """Shutdown managed services and various subprocesses. Cleanup runtime files."""
@@ -3617,6 +3621,8 @@ class JupyterHub(Application):
             return
         if self.http_server:
             self.http_server.stop()
+        if self.metrics_collector:
+            self.metrics_collector.stop()
         self.io_loop.add_callback(self.shutdown_cancel_tasks)
 
     async def start_show_config(self):
--- a/jupyterhub/metrics.py
+++ b/jupyterhub/metrics.py
@@ -23,12 +23,13 @@ them manually here.
 """
 
 import os
+import time
 from datetime import timedelta
 from enum import Enum
 
 from prometheus_client import Gauge, Histogram
 from tornado.ioloop import PeriodicCallback
-from traitlets import Any, Bool, Integer
+from traitlets import Any, Bool, Dict, Float, Integer
 from traitlets.config import LoggingConfigurable
 
 from . import orm
@@ -231,6 +232,13 @@ for s in ActiveUserPeriods:
     ACTIVE_USERS.labels(period=s.value)
 
 
+EVENT_LOOP_INTERVAL_SECONDS = Histogram(
+    'event_loop_interval_seconds',
+    'Distribution of measured event loop intervals',
+    namespace=metrics_prefix,
+)
+
+
 def prometheus_log_method(handler):
     """
     Tornado log handler for recording RED metrics.
@@ -281,6 +289,37 @@ class PeriodicMetricsCollector(LoggingConfigurable):
         config=True,
     )
 
+    event_loop_interval_enabled = Bool(
+        True,
+        config=True,
+        help="""
+        Enable event_loop_interval_seconds metric.
+        
+        Measures event-loop responsiveness.
+        """,
+    )
+    event_loop_interval_resolution = Float(
+        0.02,
+        config=True,
+        help="""
+        Interval (in seconds) on which to measure the event loop interval.
+        
+        This is the _sensitivity_ of the event_loop_interval metric.
+        Setting it too low (e.g. below 10ms) can end up slowing down the whole event loop
+        by measuring too often,
+        while setting it too high (e.g. above 1s) will limit its resolution and usefulness.
+        """,
+    )
+    event_loop_interval_log_threshold = Float(
+        1,
+        config=True,
+        help="""Log when the event loop blocks for at least this many seconds.""",
+    )
+
+    # internal state
+    _last_tick = Float()
+    _periodic_callbacks = Dict()
+
     db = Any(help="SQLAlchemy db session to use for performing queries")
 
     def update_active_users(self):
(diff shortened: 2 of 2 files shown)


yuvipanda:
This is great! I tested it out, and the buckets look like this:

```
jupyterhub_event_loop_interval_seconds_bucket{le="0.005"} 2.0
jupyterhub_event_loop_interval_seconds_bucket{le="0.01"} 3.0
jupyterhub_event_loop_interval_seconds_bucket{le="0.025"} 177.0
jupyterhub_event_loop_interval_seconds_bucket{le="0.05"} 184.0
jupyterhub_event_loop_interval_seconds_bucket{le="0.075"} 184.0
jupyterhub_event_loop_interval_seconds_bucket{le="0.1"} 184.0
jupyterhub_event_loop_interval_seconds_bucket{le="0.25"} 184.0
jupyterhub_event_loop_interval_seconds_bucket{le="0.5"} 184.0
jupyterhub_event_loop_interval_seconds_bucket{le="0.75"} 184.0
jupyterhub_event_loop_interval_seconds_bucket{le="1.0"} 184.0
jupyterhub_event_loop_interval_seconds_bucket{le="2.5"} 184.0
jupyterhub_event_loop_interval_seconds_bucket{le="5.0"} 184.0
jupyterhub_event_loop_interval_seconds_bucket{le="7.5"} 184.0
jupyterhub_event_loop_interval_seconds_bucket{le="10.0"} 184.0
jupyterhub_event_loop_interval_seconds_bucket{le="+Inf"} 184.0
```

I do think it would be useful to have better resolution at the lower end. There's a big jump between `0.01` and `0.025` as well. I would suggest 0.005 intervals there maybe? I don't think the cost to adding more buckets is that high, so we may as well do that all the way to `0.05`?

yuvipanda:
I also tested this works by putting a `time.sleep(20)` in a `get_env` call, and this resulted in:

```
jupyterhub_event_loop_interval_seconds_bucket{le="7.5"} 503.0
jupyterhub_event_loop_interval_seconds_bucket{le="10.0"} 503.0
jupyterhub_event_loop_interval_seconds_bucket{le="+Inf"} 504.0
```

Which is exactly as expected.

Great job, @minrk!

minrk:
Thanks for testing!

Having to decide on buckets was the big downside of histogram vs scalar gauge ;)

Is it useful to distinguish between 45 and 50 ms, vs 40-50? Or 25-50, even? Or is it more about resolving the distribution? Maybe only increase resolution below 25ms, e.g.: 2.5, 5, 10, 15, 20, 25, 50, 75, ...rest as default? Or throw in 30, 40 so we at least have 10ms resolution below 50?

To avoid spending too long on something that probably won't make much difference, I'll propose 3 choices, so we can pick one and move on (but feel free to add more proposals if you have them!):

1. 5ms resolution to 50ms (your initial suggestion)
2. 5ms resolution to 25ms, then 50, 75, etc. 
3. 5ms resolution to 30ms, then 40, 50, 75, etc. (10ms resolution to 50)

all adding the 2.5ms line, in part to match dask, but that's also the baseline of what we are measuring, so it should include all the 'fully idle' ticks.

minrk:
Wait, I got my orders of magnitude mixed up - we measure a _20_ millisecond resolution, not _2_, so it doesn't really make sense to have any ticks below 25. How about 5ms from 25-50, and drop below 25?

minrk:
Updated PR to apply 5ms from 25-50, followed by the standard intervals.

yuvipanda:
@minrk that makes sense! However, since that 20ms is configurable, should we make the buckets based on whatever that configurable is? Or not make that number configurable?

minrk:
My _initial_ inclination would be to leave the buckets as they are, and let folks tune the interval, which might leave some empty buckets. Making buckets configurable is a bit of a pain, I think, given that we declare metrics as module globals, and histogram buckets aren't a simple public attribute that can be modified after instantiation. I think we'd need to delete and re-register a new Histogram object.

In leaving the metric tunable, I imagined imagine that folks tuning the sensitivity would only _decrease_ the sensitivity if anyone felt like it was having a performance cost, but they still wanted to catch bigger hangs, so the result would only be some empty buckets (seems like no big deal). If folks instead wanted to _increase_ the sensitivity, that essentially wouldn't work because the smallest bucket size is set. I could also hardcode the buckets for 5ms intervals down to 10ms, which would leave 3 empty buckets with the default value, but allow increased sensitivity:

- Downside: somewhat misleading histogram by default, because what looks like a 20-25ms bucket is actually a 0-25ms bucket.
- Upside: resolves finer observations without needing custom buckets

Upside of no configurable buckets:

- doesn't encourage too fine-grained observations, which can be costly and of arguable benefit
- consistent buckets make it easier to make portable charts in e.g. jupyterhub-grafana
- simpler to implement

So for the moment, I've gone with:

- leave event_loop_interval_resolution configurable, but add caveats to docstring explaining why _increasing_ sensitivity doesn't make much sense
- leave buckets hardcoded based on the default, for now

WDYT?

yuvipanda:
I like this! Thanks a lot for working on this, @minrk!

minrk:
btw, you can try out this metric in an existing deployment by adding to your jupyterhub_config.py:

<details>
<summary>jupyterhub_config.py</summary>

```python
import time

from prometheus_client import Histogram
from tornado.ioloop import PeriodicCallback
from traitlets.log import get_logger

c = get_config()  # noqa
EVENT_LOOP_INTERVAL_SECONDS = Histogram(
    'event_loop_interval_seconds',
    'Distribution of measured event loop intervals',
    namespace="jupyterhub",
    # Increase resolution to 5ms below 50ms
    # because this is where we are most sensitive.
    # No need to have buckets below 25, since we only measure every 20ms.
    buckets=[
        # 5ms from 25-50ms
        25e-3,
        30e-3,
        35e-3,
        40e-3,
        45e-3,
        50e-3,
        # from here, default prometheus buckets
        75e-3,
        0.1,
        0.25,
        0.5,
        0.75,
        1,
        2.5,
        5,
        7.5,
        10,
        float("inf"),
    ],
)


class EventLoopMetric:
    event_loop_interval_resolution = 20e-3
    event_loop_interval_log_threshold = 1
    def _event_loop_tick(self):
        """Measure a single tick of the event loop
    
        This measures the time since the last tick
        """
        now = time.perf_counter()
        tick_duration = now - self._last_tick
        self._last_tick = now
        EVENT_LOOP_INTERVAL_SECONDS.observe(tick_duration)
        if tick_duration >= self.event_loop_interval_log_threshold:
            # warn about slow ticks
            self.log.warning("Event loop was unresponsive for %.2fs!", tick_duration)
        
    
    def start(self):
        self.log = get_logger()
        self.log.info("starting!")
        now = time.perf_counter()
        self._last_tick = self._last_tick_collect = now
        pc = PeriodicCallback(
            self._event_loop_tick,
            self.event_loop_interval_resolution * 1000,
        )
        pc.start()

metric = EventLoopMetric()
metric.start()
```

</details>

If anyone wants to give that a try and give feedback on the buckets, etc.

minrk:
Got [something](https://grafana.mybinder.org/d/b8de32f1-59c3-4fce-b8ac-f70b070f2053/event-loop-diagnostics?orgId=1&var-cluster=000000004&from=now-3h&to=now) that seems to work:

Screenshot 2023-11-10 at 13 19 06

- shows heatmap of all buckets _excluding the smallest_, which includes 99.99% of all ticks, usually, making it impossible to see any others. Uses absolute count, so you know how many times a given tick occurred in a time window
- timeseries of average, and 95%, 99% of ticks

Still probably some to figure out, but it at least seems to surface useful info about 'typical' performance and should notice delaysw

PR #4616, opened 2023-10-27
test: ensure test server is added to proxy before talking to it

minrk:
this should fix some flaky test failures in test_singleuser like [this one](https://github.com/jupyterhub/jupyterhub/actions/runs/6665391840/job/18114866017#step:11:1074):

```
FAILED jupyterhub/tests/test_singleuser.py::test_token_url_cookie - assert 302 == 200
   +  where 302 = <Response [302]>.status_code
```

I think the reason it doesn't always fail is that `proxy.check_routes()` is running in the background and will add the route when it sees it's missing.

### DIFF
--- a/jupyterhub/tests/test_singleuser.py
+++ b/jupyterhub/tests/test_singleuser.py
@@ -363,6 +363,7 @@ def test_singleuser_app_class(JUPYTERHUB_SINGLEUSER_APP):
 async def test_nbclassic_control_panel(app, user, full_spawn):
     # login, start the server
     await user.spawn()
+    await app.proxy.add_user(user)
     cookies = await app.login_user(user.name)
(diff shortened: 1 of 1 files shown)


consideRatio (approved):

consideRatio:
Not merging as tests are running still and the PR was opened 5 minutes ago - but this looks great!

minrk:
we have another flaky test in test_browser test_revoke_token, which I'll try to hit in another PR:

```
        if token_type in {"server_up", "request_by_user"}:
>           await expect(revoke_btns).to_have_count(1)
E           AssertionError: Locator expected to have count '1'
E           Actual value: 0 
```

minrk:
@meeseeksdev backport to 4.x

PR #4617, opened 2023-10-27
try to improve reliability of test_external_proxy

minrk:
- use random ports instead of pre-specified numbers, which may collide
- mark test as flaky to give it a second chance to succeed

This test has [failed intermittently](https://github.com/jupyterhub/jupyterhub/actions/runs/6574430927/job/18080283121) with port collisions, like this one:

```
12:05:28.756 [ConfigProxy] info: Proxy API at http://127.0.0.1:55432/api/routes
  12:05:28.758 [ConfigProxy] error: Uncaught Exception: listen EADDRINUSE: address already in use 127.0.0.1:55432
  12:05:28.758 [ConfigProxy] error: Error: listen EADDRINUSE: address already in use 127.0.0.1:55432
      at Server.setupListenHandle [as _listen2] (net.js:1331:16)
      at listenInCluster (net.js:1379:12)
      at doListen (net.js:1516:7)
      at processTicksAndRejections (internal/process/task_queues.js:83:21)
```

it's unclear exactly what it's colliding with, but using random ports may fix the issue, and marking the test as flaky should make it quite unlikely to fail (it's unlikely that _just_ marking it as flaky would fix it, since the collision is likely due to something outside the test allocating the requested port and it would just collide again if it tried the test on the same port a second time)

### DIFF
--- a/jupyterhub/tests/test_proxy.py
+++ b/jupyterhub/tests/test_proxy.py
@@ -9,6 +9,7 @@ import pytest
 from traitlets import TraitError
 from traitlets.config import Config
 
+from ..utils import random_port
 from ..utils import url_path_join as ujoin
 from ..utils import wait_for_http_server
 from .mocking import MockHub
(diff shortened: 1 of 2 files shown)


consideRatio (approved):
Aha nice!

minrk:
#4616 should fix two of those, and #4618 should hopefully fix the others. Going to rebase and see if they go away.

consideRatio (approved):
A slow arm build caused a failure, I think that is unrelated!

minrk:
@meeseeksdev backport to 4.x

PR #4618, opened 2023-10-27
browser test: wait for token request to finish before reloading

minrk:
this may be the cause of flaky tests where there are no tokens to list

This _might_ fix failures like [this one](https://github.com/jupyterhub/jupyterhub/actions/runs/6665976493/job/18116613669?pr=4616) in `test_revoke_token[request_by_user]`:

```
        if token_type in {"server_up", "request_by_user"}:
>           await expect(revoke_btns).to_have_count(1)
E           AssertionError: Locator expected to have count '1'
E           Actual value: 0 
E           Call log:
E           LocatorAssertions.to_have_count with timeout 5000ms
```

where the failure suggests that the token wasn't actually requested, and the logs appear to confirm that the request for a new token is not received by the server. My hunch is that awaiting the button click doesn't await the resulting fetch, and reloading the page prevents it from actually firing. A second option is that the javascript click event isn't hooked up before we trigger the click, so clicking doesn't do anything.

### DIFF
--- a/jupyterhub/tests/browser/test_browser.py
+++ b/jupyterhub/tests/browser/test_browser.py
@@ -428,7 +428,7 @@ async def test_token_request_form_and_panel(app, browser, user):
     await expect(token_area_heading).to_have_text(expected_panel_token_heading)
     token_result = browser.locator('#token-result')
     await expect(token_result).not_to_be_empty()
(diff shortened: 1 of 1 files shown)


consideRatio (approved):

minrk:
@meeseeksdev backport to 4.x

PR #4630, opened 2023-11-13
avoid setting unused oauth state cookies on API requests

### LINKED ISSUE #4629: singleuser sets unused oauth state cookie when not redirecting
Attempting to start oauth login too many times fill browser cookies, eventually ending in [431 headers too big](https://discourse.jupyter.org/t/error-431-headers-too-big-during-singleuser-oauth/21833/15).

Background:

- tornado's `@web.authenticated` checks if a user is authenticated, and redirects to `get_login_url()`
- jupyterlab makes lots of API requests, even while not logged in, which can't follow those redirects
- oauth state means that the login URL includes the oauth state, which is also stored in a cookie
- jupyter-server [overrides get_login_url](https://github.com/jupyter-server/jupyter_server/blob/v2.10.0/jupyter_server/base/handlers.py#L755-L762) for API requests to return 403 instead of redirecting
- HubOAuth _also_ [overrides get_login_url](https://github.com/jupyter-server/jupyter_server/blob/v2.10.0/jupyter_server/base/handlers.py#L755-L762) in order to inject the oauth state into the login URL

This combination means that when login url is _computed_, the oauth state cookie is set, even if the request doesn't actually get redirected, [producing logs like this](https://discourse.jupyter.org/t/error-431-headers-too-big-during-singleuser-oauth/21833/14):

```
2023-11-08T13:29:06.511Z | No user identified
-- | --
2023-11-08T13:29:06.511Z | Detected unused OAuth state cookies
2023-11-08T13:29:06.511Z | Redirecting to login url: /hub/api/oauth2/authorize?client_id=jupyterhub-user-[user]&redirect_uri=%2Fuser%2F[user]%2Foauth_callback&response_type=code&state=[state]
2023-11-08T13:29:06.511Z | wrote error: ‘Forbidden’
2023-11-08T13:29:06.511Z | 403 GET /user/[user]/api/contents/notebooks/jump_detection/replay_jump_detection.ipynb?content=0&1699450145824 (@10.49.241.58) 2.80ms
```

What we need to do is avoid setting the oauth state cookie until _after_ we know that it will be redirected to.

Best solution: move the cookie-setting step to inside an overridded `redirect` method, to ensure it's really truly overridden in the event of a redirect. Downside: I don't think I can solve this without breaking a public API in HubAuth (that the cookie is computed only by setting it)

Second-best solution that's doable without any breakage: append `state` to the login URL after computing it with `super()`. I think this should be pretty straightforward.

minrk:
get_login_url may raise 403 to short-circuit redirects, so don't create the oauth state cookie until _after_ that has happened.

As discussed in #4629, a more robust approach would be to move the cookie setting to inside an overridden `.redirect()` method so it's only ever used if redirect actually occurs, but that would require changing the HubAuth API from computing the cookie in `set_state_cookie` to computing the cookie on one API and setting it in anther, which is technically a breaking change.

closes #4629

### DIFF
--- a/jupyterhub/services/auth.py
+++ b/jupyterhub/services/auth.py
@@ -1079,19 +1079,30 @@ class HubAuthenticated:
     def hub_auth(self, auth):
         self._hub_auth = auth
 
+    _hub_login_url = None
+
     def get_login_url(self):
         """Return the Hub's login URL"""
-        login_url = self.hub_auth.login_url
-        if isinstance(self.hub_auth, HubOAuth):
-            # add state argument to OAuth url
-            state = self.hub_auth.set_state_cookie(self, next_url=self.request.uri)
-            login_url = url_concat(login_url, {'state': state})
+        if self._hub_login_url is not None:
+            # cached value, don't call this more than once per handler
+            return self._hub_login_url
         # temporary override at setting level,
         # to allow any subclass overrides of get_login_url to preserve their effect
         # for example, APIHandler raises 403 to prevent redirects
-        with mock.patch.dict(self.application.settings, {"login_url": login_url}):
-            app_log.debug("Redirecting to login url: %s", login_url)
-            return super().get_login_url()
+        with mock.patch.dict(
+            self.application.settings, {"login_url": self.hub_auth.login_url}
+        ):
+            login_url = super().get_login_url()
+        app_log.debug("Redirecting to login url: %s", login_url)
+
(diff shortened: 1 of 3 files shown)


mbiette:
Hello, do you know when will be the next release that would include this fix?

mbiette:
@minrk do you know when there will be a release with that PR? Is there anything I can help with?

minrk:
@meeseeksdev backport to 4.x

PR #4632, opened 2023-11-16
simplify, avoid errors in parsing accept headers

minrk:
Certain Accept header values weren't handled (most critically, supplying no Accept header at all) could result in an error attempting to parse the header. q-value priority was also inverted.

The main change here is to remove a bunch of parsing of unused values and avoid server-side errors for empty or invalid values.

### DIFF
--- a/jupyterhub/utils.py
+++ b/jupyterhub/utils.py
@@ -661,58 +661,34 @@ def utcnow():
 
 def _parse_accept_header(accept):
     """
-    Parse the Accept header *accept*
+    Parse the Accept header
 
-    Return a list with 3-tuples of
-    [(str(media_type), dict(params), float(q_value)),] ordered by q values.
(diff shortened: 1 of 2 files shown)


jupyterhub/utils.py
```diff
@@ -661,58 +661,34 @@ def utcnow():
 
 def _parse_accept_header(accept):
     """
-    Parse the Accept header *accept*
-
-    Return a list with 3-tuples of
-    [(str(media_type), dict(params), float(q_value)),] ordered by q values.
-    If the accept header includes vendor-specific types like::
-        application/vnd.yourcompany.yourproduct-v1.1+json
```
manics, line 669: Was this originally for some legacy browsers that sent unusual headers, or is it just old code copied from somewhere else that was never needed?

manics (approved):

PR #4633, opened 2023-11-16
Authenticator reference doc: update authenticate return

manics:
The authenticate reference doc is out of date, the return value can also be a dict https://jupyterhub.readthedocs.io/en/stable/reference/api/auth.html#jupyterhub.auth.Authenticator.authenticate

I've ignored `admin` on the assumption it's deprecated and replaced by RBAC (maybe we should update the API docs too?)

### DIFF
--- a/docs/source/reference/authenticators.md
+++ b/docs/source/reference/authenticators.md
@@ -70,7 +70,9 @@ If authentication is successful the `authenticate` method must return either:
 - the username (non-empty str) of the authenticated user
 - or a dictionary with fields:
   - `name`: the username
-  - `auth_state`: optional, a dictionary of of [auth state that will be persisted](authenticator-auth-state)
+  - `admin`: optional, a boolean indicating whether the user is an admin.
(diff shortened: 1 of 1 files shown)


minrk:
Great!

`admin` is not exactly deprecated at this level. Using the user admin _attribute_ to determine permission _is_ deprecated in favor of using scopes (i.e. no deployments should need any admin users), but the `admin` field in the authenticate model as a mechanism to assign the admin _role_ is not deprecated. Groups are a more generalized approach to achieve the same thing, so I think it's fine to omit it here. But support for the `admin` field has no plans for removal, and still makes sense in RBAC, it just means something slightly different implementation-wise.

Where `admin` is documented should probably come with a note/link to RBAC that it shouldn't be _needed_ anymore, despite the acknowledgement that it's never going away.

manics:
Thanks for the explanation! I think it makes sense to include admin, with a link to RBAC

PR #4634, opened 2023-11-17
Add workflow to update registry overviews

mathbunnyru:
This will update registry overviews automatically.
I intentionally use Docker Hub currently - I would like to add a message that Quay.io needs to be used, and then we can switch this workflow to Quay.io

Unfortunately, I can't test this workflow in a branch properly - there is no access to secrets in PRs.
I created a similar workflow in `jupyter/docker-stacks` and it worked well for Docker Hub and now works well with Quay.io: https://github.com/jupyter/docker-stacks/blob/main/.github/workflows/registry-overviews.yml

### DIFF
--- a/.github/workflows/registry-overviews.yml
+++ b/.github/workflows/registry-overviews.yml
@@ -1,0 +1,52 @@
+name: Update Registry overviews
+
+env:
+  OWNER: ${{ github.repository_owner }}
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - ".github/workflows/registry-overviews.yml"
+
+      - "README.md"
+      - "onbuild/README.md"
+  workflow_dispatch:
+
+jobs:
+  update-overview:
+    runs-on: ubuntu-latest
+    name: update-overview (${{matrix.image}})
+    if: github.repository_owner == 'jupyter'
+
+    steps:
+      - name: Checkout Repo ⚡️
+        uses: actions/checkout@v4
+
+      - name: Push README to Registry 🐳
+        uses: christian-korneck/update-container-description-action@d36005551adeaba9698d8d67a296bd16fa91f8e8 # v1
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/registry-overviews.yml
```diff
@@ -0,0 +1,54 @@
...
+          provider: dockerhub
+          short_description: ${{ matrix.description }}
+          readme_file: ${{ matrix.readme_file }}
+
+    strategy:
+      matrix:
+        include:
+          - image: jupyterhub
+            description: "JupyterHub: multi-user Jupyter notebook server"
+            readme_file: README.md
```
mathbunnyru, line 45: This seems to be more-or-less consistent with current Docker Hub readme.

.github/workflows/registry-overviews.yml
```diff
@@ -0,0 +1,54 @@
...
+          - image: jupyterhub-onbuild
+            description: onbuild version of JupyterHub images
+            readme_file: onbuild/README.md
```
mathbunnyru, line 48: This file seems is more up-to-date than the current Docker Hub text.

.github/workflows/registry-overviews.yml
```diff
@@ -0,0 +1,54 @@
...
+          - image: jupyterhub-demo
+            description: Demo JupyterHub Docker image with a quick overview of what JupyterHub is and how it works
+            readme_file: demo-image/README.md
```
mathbunnyru, line 51: Docker Hub text is completely missing

.github/workflows/registry-overviews.yml
```diff
@@ -0,0 +1,54 @@
...
+          - image: singleuser
+            description: "single-user docker images for use with JupyterHub and DockerSpawner see also: jupyter/docker-stacks"
+            readme_file: singleuser/README.md
```
mathbunnyru, line 54: This file seems to be more-or-less consistent with current Docker Hub text.

PR #4642, opened 2023-11-24
Use jupyterhub repository_owner in registry-overviews

mathbunnyru:
Follow-up on: https://github.com/jupyterhub/jupyterhub/pull/4634/

### DIFF
--- a/.github/workflows/registry-overviews.yml
+++ b/.github/workflows/registry-overviews.yml
@@ -20,7 +20,7 @@ jobs:
   update-overview:
     runs-on: ubuntu-latest
     name: update-overview (${{matrix.image}})
-    if: github.repository_owner == 'jupyter'
+    if: github.repository_owner == 'jupyterhub'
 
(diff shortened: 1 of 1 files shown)


mathbunnyru:
@minrk sorry, I didn't change `jupyter` -> `jupyterhub`.

minrk:
Thanks!

mathbunnyru:
@minrk we need to create new access token with `Read, Write, Delete` permission and then use it here.
Could you do that please?
I have no permissions in `jupyterhub` Docker Hub / GitHub.

minrk:
Apparently, it needs full admin permissions to edit the readme, including permanently deleting the repo! I don't really want the bot to have that level of permission. Is that the only way?

minrk:
Got it. Updated permissions and re-ran the action. All green!

mathbunnyru:
Great, seems to work 👍

PR #4643, opened 2023-11-28
add warning when an oauth client is used after its secret is deleted

minrk:
the client cannot be used for much anymore, so this situation is likely caused by an error, so we should notice it.

### DIFF
--- a/jupyterhub/oauth/provider.py
+++ b/jupyterhub/oauth/provider.py
@@ -424,6 +424,7 @@ class JupyterHubRequestValidator(RequestValidator):
         if orm_client is None:
             return False
         if not orm_client.secret:
+            app_log.warning("OAuth client %s present without secret", client_id)
             return False
         request.client = orm_client
         return True


consideRatio (approved):
I'd label this as `maintenance` thinking that if this isn't maintenance, doesn't almost everything become `enhancement` ?

minrk:
I was on the fence, but went with enhancement. I don't believe the distinction is particularly relevant, but it's improved, user-facing behavior, so I went with enhancement. 

> doesn't almost everything become enhancement ?

Not really. The vast majority of maintenance tasks are related to CI, code formatting, deprecations, tests, etc. with no user-facing changes. I think improving log output is a user-facing change.

For the most part, I'd hold maintenance for things that affect the repo, but has no real effect on an installation of jupyterhub.

consideRatio:
Ah okay, that is a sharper deliniation than I've adopted so far, and I think I appreciate it more - I'll adjust my interpretation to yours!

PR #4645, opened 2023-11-28
explicitly require groups in auth model when Authenticator.manage_groups is enabled

minrk:
avoids the experience of silently ignored config for Authenticators that don't support `manage_groups`, which leads to confusion like https://github.com/jupyterhub/oauthenticator/issues/706.

technically breaking, because the previously unspecified behavior allowed Authenticators to omit the groups field which is interpreted as "no change to group membership." That will no longer be allowed.

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -820,7 +820,16 @@ class BaseHandler(RequestHandler):
 
         # apply authenticator-managed groups
         if self.authenticator.manage_groups:
-            group_names = authenticated.get("groups")
+            if "groups" not in authenticated:
+                # to use manage_groups, group membership must always be specified
+                # Authenticators that don't support this feature will omit it,
+                # which should fail here rather than silently not implement the requested behavior
+                auth_cls = self.authenticator.__class__.__name__
+                raise ValueError(
+                    f"Authenticator.manage_groups is enabled, but auth_model for {username} specifies no groups."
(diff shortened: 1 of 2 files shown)


jupyterhub/handlers/base.py
```diff
@@ -820,7 +820,16 @@ async def auth_to_user(self, authenticated, user=None):
...
+                # Authenticators that don't support this feature will omit it,
+                # which should fail here rather than silently not implement the requested behavior
+                auth_cls = self.authenticator.__class__.__name__
+                raise ValueError(
+                    f"Authenticator.manage_groups is enabled, but auth_model for {username} specifies no groups."
+                    f" Does {auth_cls} support manage_groups=True?"
+                )
+            group_names = authenticated["groups"]
             if group_names is not None:
                 user.sync_groups(group_names)
```
consideRatio, line 834: Should `authenticated["groups"]` be allowed to be `None` - if it is, then its currently treated differently from `[]`, where setting it to `None` means its not getting synced, while setting it to `[]` means its getting synced to no groups.

minrk: Yes, it's explicitly "no change" if None. I'll add that to the doc.

I'm mostly thinking of `refresh_user` here, where it may not have what it needs to refresh group membership, in which case 'no change' is still okay, but it should be _explicit_.

consideRatio (approved):

PR #4648, opened 2023-11-30
add Spawner.poll_jitter

### LINKED ISSUE #4646: Add jitter to poll checks
Some deployments use spawners that may consume resources to execute Spawner.poll() calls.  For instance, SlurmSpawner in [batchspawner](https://github.com/jupyterhub/batchspawner) makes a call to Slurm via [`squeue`](https://github.com/jupyterhub/batchspawner/blob/e89bf6a69ccfb9119b7066ec49df35694afb2c9e/batchspawner/batchspawner.py#L724) to get the status of notebook jobs.

If there are a large number of these jobs, and the Hub is redeployed, poll checks for all the notebook jobs synchronize, so every `poll_interval` seconds the Slurm controller gets a large number of status requests in a short amount of time, which Slurm does not like.  Over time the poll checks smear out, but this happens very slowly.

### Proposed change

The smear-out could be accelerated if poll checks could be jittered.  I think the very first poll check when the Hub starts up maybe cannot/should not be jittered, but subsequent ones could be, by setting the `jitter` argument to the [`PeriodicCallback`](https://www.tornadoweb.org/en/stable/ioloop.html#tornado.ioloop.PeriodicCallback) setup in [`Spawner.start_polling`](https://github.com/jupyterhub/jupyterhub/blob/73b2307b367f8582e015fb6f8da5dd213523a943/jupyterhub/spawner.py#L1407).  By default this could be 0, but could be adjusted by adding a Spawner config parameter like `poll_jitter`.

### Alternative options

* Another option would be to proxy all notebook server state through a reflector service like I think is used in KubeSpawner.  This would entail implementation changes to spawners that wanted to use that service.
* Don't allow "too many" servers to run in the first place (set an active_server_limit), or, more aggressively cull idle servers to cut down on the number running.

Adding jitter to the poll interval isn't the best solution maybe, but it is very simple to implement and we could maintain the current behavior by keeping the default jitter to 0 seconds. 

Anybody with poll operations that consume any kind of resource in the course of execution.

I'd add a `poll_jitter` parameter to Spawner and then set up `start_polling` to use it.  I thought for sure this conversation had happened in the past though but I can't find it in the issues (maybe issues where there are large numbers of servers running) but I might have missed it.  There may be reasons I haven't thought of that make this not a good idea.

minrk:
to avoid alignment of poll callbacks, e.g. after Hub restart

default: 10%, or +/- 3 seconds for the default poll interval of 30 seconds.

closes #4646

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -490,6 +490,20 @@ class Spawner(LoggingConfigurable):
         """,
     ).tag(config=True)
 
+    poll_jitter = Float(
+        0.1,
+        min=0,
+        max=1,
+        help="""
+        Jitter fraction for poll_interval.
+        
+        Avoids alignment of poll calls for many Spawners,
+        e.g. when restarting JupyterHub, which restarts all polls for running Spawners.
+
+        `poll_jitter=0` means no jitter, 0.1 means 10%, etc.
+        """,
+    ).tag(config=True)
+
     _callbacks = List()
     _poll_callback = Any()
 
@@ -1405,7 +1419,9 @@ class Spawner(LoggingConfigurable):
         self.stop_polling()
 
         self._poll_callback = PeriodicCallback(
-            self.poll_and_notify, 1e3 * self.poll_interval
+            self.poll_and_notify,
+            1e3 * self.poll_interval,
+            jitter=self.poll_jitter,
         )
         self._poll_callback.start()
 


manics (approved):
@rcthomas does this look good to you?

rcthomas:
Yes!  Thanks!

PR #4658, opened 2023-12-09
Compare major hub and singleuser versions only

manics:
JupyterHub uses semantic versioning and has been >1.0.0 for a long time. It should be fine for the hub and singleuser versions to differ in their minor component.


consideRatio (approved):
Looks right to me! I fully agree on not warning if there is only a minor version mis-matched.

minrk:
poking CI after #4663

minrk (approved):

PR #4662, opened 2023-12-17
docs: Remove non-actionable step from developer setup

yuvipanda:
I just went through these with @jmunroe, and found the db step a little confusing - there is no action to really be taken here, as pretty much everyone just uses sqlite for development (and even production). So I've just removed that step, as python almost always ships with sqlite built into it.


yuvipanda:
I actually did a very small livestream of me making this PR! https://www.youtube.com/watch?v=k5uGD0eQOFQ. I'm hoping to do more of these :)

minrk:
Makes good sense.

PR #4665, opened 2023-12-19
avoid deprecated datetime.utcnow

minrk:
deprecated in Python 3.12

replace with equivalent `utils.utcnow(with_tz=False)`.

We probably should be using integer timestamps everywhere internally rather than datetimes, but that's a bigger, more sensitive change.


consideRatio (approved):
Looks right, I've only reviewed chances seen in PR from mobile so its not reviewed for completeness.

PR #4676, opened 2024-01-18
Add appropriate scope in examples/service-whoami-flask

rschroll:
I ran into some trouble trying to get services to work with an updated version of JupyterHub.  I went to this example for help, but it also failed.  I think this is the appropriate solution -- the `user` role needs a scope to access the service.  There seems to be an analogous fix already in examples/service-whoami.

In the process of testing this, I got a Flask error from the `return 403` lines in the example.  I don't know if that was ever valid Flask code, but it doesn't work now.  I've changed those so something that should be backwards-compatible.

### DIFF
--- a/examples/service-whoami-flask/README.md
+++ b/examples/service-whoami-flask/README.md
@@ -15,14 +15,19 @@ After logging in with your local-system credentials, you should see a JSON dump 
 ```json
 {
   "admin": false,
-  "last_activity": "2016-05-27T14:05:18.016372",
+  "groups": [],
+  "kind": "user",
   "name": "queequeg",
-  "pending": null,
-  "server": "/user/queequeg"
+  "scopes": [
+    "access:services!service=whoami",
+    "read:users:groups!user=queequeg",
+    "read:users:name!user=queequeg"
+  ],
+  "session_id": "a32e59cdd7b445759c58c48e47394a38"
 }
 ```
 
-This relies on the Hub starting the whoami service, via config (see [jupyterhub_config.py](./jupyterhub_config.py)).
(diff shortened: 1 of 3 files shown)


minrk:
@rschroll thanks! Can you run the autoformatters with `pre-commit run --all-files` (CI tried to fix it automatically, but doesn't have permission to push to your PR).

minrk (approved):

rschroll:
I was able to get the black formatter to run, but prettier failed to run.  It seems to need node >= 14, while I appear to have 12.22.  My experience with wrangling node versions is that it quickly leads to well of infinite sadness, so I'm hesitant to try that on my laptop.  I'll see if I can get it working in a container.

As an aside: I'd argue that the black formatting is wrong if this file is to be used pedagogically.  Sticking everything on one line makes it harder for learners to grasp what's going on.  I'm afraid it's too easy to miss the "self" scope, for example.

rschroll:
Looks like I got it to pass.  It turns out that if I add some comments, black seems to get what I'm trying to do. :)

I tried to get prettier to run in both the latest jupyterhub image from quay.io as well as an image I build from the Dockerfile in the repo.  Both had node 12.22 installed, and therefore prettier refused to work.  I'm not sure if this is a bug _per se_, but it does strike me as odd.  It does look like a newer version of node is installed as one of the stages; perhaps it should be installed in the final image as well, to allow development there?

For anyone facing the same problem, for now you can just paste the code in https://prettier.io/playground/ and paste the result back in your editor.

minrk:
Thanks! Sorry for the trouble with prettier and thank you for documenting how you solved it.

I wish we could trade it for mdformat, but can't really until https://github.com/executablebooks/mdformat-myst/issues/13 is addressed

PR #4677, opened 2024-01-19
Improve validation, docs for token.expires_in

### LINKED ISSUE #4660: HTTP 500 error in Docker Jupyterhub when requesting user tokens
Look at this problem ,  it not works with official documents.

https://discourse.jupyter.org/t/http-500-error-in-docker-jupyterhub-when-requesting-user-tokens/18872

minrk:
#4660 describes a problem where the generated rest api docs use 0 as a default number in examples, but expires_in=0  is interpreted as create a token that is immediately expired, rather than one that never expires.

Collection of small changes to improve this situation:

- accept 0 meaning no expiration, since folks have tried to use it that way
- clear error message for invalid (e.g. negative) values
- specify nonzero example in rest api doc so it doesn't default to invalid `0` (which is now valid)
- better error logging if orm token fails to be retrieved

closes #4660

### DIFF
--- a/docs/source/_static/rest-api.yml
+++ b/docs/source/_static/rest-api.yml
@@ -572,9 +572,11 @@ paths:
               properties:
                 expires_in:
                   type: number
+                  example: 3600
                   description:
                     lifetime (in seconds) after which the requested token
                     will expire.
+                    Omit or specify null for no expiration.
                 note:
                   type: string
                   description: A note attached to the token for future bookkeeping
--- a/docs/source/_static/rest-api.yml
+++ b/docs/source/_static/rest-api.yml
@@ -576,7 +576,7 @@ paths:
                   description:
                     lifetime (in seconds) after which the requested token
                     will expire.
-                    Omit or specify null for no expiration.
+                    Omit, or specify null or 0 for no expiration.
                 note:
                   type: string
                   description: A note attached to the token for future bookkeeping
--- a/docs/source/_static/rest-api.yml
+++ b/docs/source/_static/rest-api.yml
@@ -575,8 +575,7 @@ paths:
                   example: 3600
                   description:
                     lifetime (in seconds) after which the requested token
-                    will expire.
-                    Omit, or specify null or 0 for no expiration.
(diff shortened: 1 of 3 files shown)


docs/source/_static/rest-api.yml
```diff
@@ -572,9 +572,11 @@ paths:
               properties:
                 expires_in:
                   type: number
+                  example: 3600
                   description:
                     lifetime (in seconds) after which the requested token
                     will expire.
+                    Omit or specify null for no expiration.
```
manics, line 579:
```suggestion
                    Omit, or specify null or 0 for no expiration.
```

jupyterhub/orm.py
```diff
@@ -813,7 +814,18 @@ def new(
...
             orm_token.service = service
-        if expires_in is not None:
+        if expires_in:
+            if not isinstance(expires_in, numbers.Real):
+                raise TypeError(
+                    f"expires_in must be a positive integer or null, not {expires_in!r}"
+                )
+            expires_in = int(expires_in)
+            # tokens must always expire in the future
+            if expires_in < 1:
```
manics, line 824:
```suggestion
            if expires_in < 0:
```

manics:
`test_rest_api_rbac_scope_descriptions_are_updated` is failing, I think https://github.com/jupyterhub/jupyterhub/blob/main/docs/source/rbac/generate-scope-table.py needs to be run to reformat the file.

manics (commented):
`test_rest_api_rbac_scope_descriptions_are_updated` is failing, I think https://github.com/jupyterhub/jupyterhub/blob/main/docs/source/rbac/generate-scope-table.py needs to be run to reformat the file.

jupyterhub/orm.py
minrk: This is nonzero on purpose - to prevent e.g. `0.5` from rounding down to 0 and getting through.

consideRatio (approved):

minrk:
@meeseeksdev backport to 4.x

PR #4679, opened 2024-01-20
Unescape jinja username

manics:
If a username contains characters such as `<>` you can login and launch a server, but you can't stop it as the stop command is sent as a background request that uses an incorrectly escaped username:
https://github.com/jupyterhub/jupyterhub/blob/ab588c28ce06c13352acf0dcfe197d4b67e69431/share/jupyterhub/static/js/jhapi.js#L67-L71
where `user` is defined in
https://github.com/jupyterhub/jupyterhub/blob/ab588c28ce06c13352acf0dcfe197d4b67e69431/share/jupyterhub/templates/page.html#L71

### DIFF
--- a/share/jupyterhub/templates/page.html
+++ b/share/jupyterhub/templates/page.html
@@ -68,7 +68,16 @@
         base_url: "{{base_url}}",
         prefix: "{{prefix}}",
         {% if user %}
-        user: "{{user.json_escaped_name}}",
+        {#
+          Autoescaping in templates is turned on in JupyterHub, need to unescape
+          https://github.com/pallets/markupsafe/blob/2.1.4/src/markupsafe/_native.py#L6
+        #}
+        user: "{{user.json_escaped_name}}"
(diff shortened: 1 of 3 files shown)


share/jupyterhub/templates/page.html
```diff
@@ -68,7 +68,16 @@ <h1 class="modal-title" id="{{key}}-label">{{title}}</h1>
         base_url: "{{base_url}}",
         prefix: "{{prefix}}",
         {% if user %}
-        user: "{{user.json_escaped_name}}",
```
manics, line 71: I couldn't decide whether `{{user.json_escaped_name|safe}}` is safe

minrk: `json_escaped_name | safe` ought to be correct, because the escaped value cannot close the quote (arguably json_escaped_name should have _included_ the outer quotes, but it's okay).

minrk:

manics:
Sorry, forgot about this!

PR #4682, opened 2024-01-25
Note that you can throw a 403 from authenticator methods

yuvipanda:
From https://github.com/jupyterhub/oauthenticator/pull/719#pullrequestreview-1838530737

We *may* be doing this in https://github.com/jupyterhub/oauthenticator/pull/719#pullrequestreview-1838530737 - and
if we do, we should change upstream doc to match.

### DIFF
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -453,7 +453,10 @@ class Authenticator(LoggingConfigurable):
     def check_allowed(self, username, authentication=None):
         """Check if a username is allowed to authenticate based on configuration
 
-        Return True if username is allowed, False otherwise.
+        Return True if username is allowed, False otherwise. Subclasses can also
+        raise a `web.HTTPError(403, message)` to immediately *deny* access and show
+        the end user the given message.
+
         No allowed_users set means any username is allowed.
 
         Names are normalized *before* being checked against the allowed set.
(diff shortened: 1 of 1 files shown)


jupyterhub/auth.py
```diff
@@ -453,7 +453,10 @@ def normalize_username(self, username):
     def check_allowed(self, username, authentication=None):
         """Check if a username is allowed to authenticate based on configuration
 
-        Return True if username is allowed, False otherwise.
+        Return True if username is allowed, False otherwise. Subclasses can also
+        raise a `web.HTTPError(403, message)` to immediately *deny* access and show
+        the end user the given message.
```
minrk, line 458: I definitely agree that documenting the 403 error to enable propagating error messages is a good thing.

But perhaps it is better to point to `check_blocked_users` for _blocking_ users. Since `allowed` may have multiple passes (as oauthenticator's implementation does), putting a rejection in one makes the outcome order-dependent, and may not have the desired effect if another earlier pass granted access. This isn't the case in the base JupyterHub API, which has only the one call. Putting a rejection in `check_blocked_users` will always have the desired effect.

consideRatio: In oauthenticator, we are also currently raising an error during the authentication step before check_ functions are called, when a user arrives from an idp that isn't preconfigured via allowed_idps (naming of this is not suitable, its now allowing users from these idps, only accepting users from pre-configured idps)

manics: Instead of documenting this here or in other selected methods could we move this to a higher level, and say all methods called during `authenticate()` may raise 403?

minrk: Yeah, that should be generally true in authenticate, check_allowed/check_blocked_users.

consideRatio (approved):
Title update needed but lgtm

PR #4684, opened 2024-01-29
enable log capture in pytest, compatibility with pytest 8

minrk:
- better periodic callback cleanup to reduce warnings
- fix log-assertion tests to use caplog instead of pytest.warns, which wasn't correct

the log assertion tests started failing with pytest 8 because they were using pytest.warns in a deprecated way. It turned out that these were not actually exercising the intended code, as they are meant to check for log messages, not warnings.

This should fix compatibility with pytest 8.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1725,6 +1725,7 @@ class JupyterHub(Application):
     )
 
     metrics_collector = Any()
+    _periodic_callbacks = Dict()
 
     def init_handlers(self):
         h = []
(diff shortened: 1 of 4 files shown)


jupyterhub/app.py
```diff
@@ -3555,16 +3557,17 @@ async def start(self):
         # Check services health
         self._check_services_health_callback = None
         if self.service_check_interval:
-            self._check_services_health_callback = PeriodicCallback(
```
consideRatio, line 3558: This callback is never set now i think, a mistake?

minrk: No, these handles only existed for cleanup in `stop`, but then needed to be tracked by name in both start and stop. Putting them in a dictionary means stop doesn't need to be updated when these are added, removed, or changed.

consideRatio: Screenshot_20240129-143135

I think maybe the assignment to None should be cleaned up as well? Reviewing via mobile, a bit limited to looking this up thoroughly

minrk: Good catch! Removed that as well.

consideRatio (approved):

PR #4699, opened 2024-02-06
clarify some points where users can disable security for their own servers

minrk:
makes explicit that selecting images is equivalent to the installing packages issue.

@consideRatio this is my attempt at addressing https://github.com/jupyterhub/jupyterhub/pull/4594#issuecomment-1929293237


consideRatio (approved):
Excellent! Thank you @minrk!!!

Letting you decide on when to merge to help you avoid possible merge conflict issues etc. Do merge right away if that makes sense!

PR #4701, opened 2024-02-10
Add Authenticator config `allow_all` and `allow_existing_users`

### LINKED ISSUE #4484: Authenticator config to toggle allowing users by default (when no other allow config is specified)
If `Authenticator.allowed_users` isn't specified, and for example `GitHubOAuthenticator.allowed_organizations` isn't specified either, we have currently in OAuthenticator defaulted to allowing all users not part of `Authenticator.blocked_users`.

I think this behavior should change by default for at least some authenticators, and I wonder if we should introduce a flag in `Authenticator` that for the various OAuthenticator classes could make use of to toggle this with explicit config. I think explicit config tends to be more self-explanatory config, and I think self-explanatory config is the best kind of documentation for config.

### Idea for new Authenticator base class config

- **Naming and option**
  - Draft name proposal `Authenticator.allow_users_by_default`
  - Allowed values are None, False/True
- **Partial misconfiguration protection**
  We introduce a validation of the config value, making False/True only be something that works if the specific authenticator declared it supports it. This protects users from assuming this config works with an authenticator not supporting it, but it doesn't protect users of old versions of JupyterHub using an authenticator that doesn't support it.
- **Behavior modes**
  - None (default config value), the current behavior is retained
  - True, the current behavior is retained
  - False, users are required to configure explicitly what users are allowed because otherwise no users will have access

### Adoption in other Authenticators

If we go for this we it will be available in JupyterHub 4.x.y, but we can support it in authenticators like various OAuthenticator classes with JupyterHub versions older than JupyterHub 4.x.y.

### Action points

- Comment with feedback

### LINKED ISSUE #4487: Ideas for new authenticator config
I agreed with @minrk in a the recent June jupyterhub team meeting to describe an idea of new config without coupling it to suggesting how it should be implemented.

I want directly configured behaviors instead of behavior following from if `allowed_users` is truthy or falsy. Currently the following behaviors can follow from declaring `allowed_users`:

1. Declaring `allowed_users` truthy currently [implies allowing existing users](https://github.com/jupyterhub/jupyterhub/blob/4.0.0/jupyterhub/auth.py#L647-L648) to login as well.
2. Declaring `allowed_users` falsy can for some authenticators in certain configurations imply allowing all authenticated users to login.

I suggest new authentication config like below to provide a way to directly configure the otherwise implied behavior:

1. `allow_existing_users`
2. `allow_all`

I think such new config could make:
- a deployments jupyterhub config easier to understand without reading the docs
- easier to provide as secure by default
- easier to document, maintain, and implement

## Related

- Discussion related to `allow_existing_users`: #4483
- Discussion related to `allow_all`: #4484
- Discussion related to defaulting to not allowing existing users in OAuthenticator, introducing `allow_existing_users` and defaulting to False: https://github.com/jupyterhub/oauthenticator/issues/619
- Discussion related to defaulting to not allowing all users in OAuthenticator when no users is explicitly allowed, introducing `allow_all` and defaulting to False: https://github.com/jupyterhub/oauthenticator/issues/620

consideRatio:
<details><summary>Initial outdated PR text</summary>

Configuring `Authenticator.allowed_users` truthy makes other existing users in JupyterHub's database be allowed access, this could come as a surprise. This new config is meant to help avoid such surprise. With this new config, a JupyterHub admin is able to directly declare if the existing users in JupyterHub's database is to be granted access or not.

If `allow_existing_users` isn't explicity set, the default value will be computed to True or False depending on if `allowed_users` is Truthy, which makes the introduction of this config a non-breaking change.

This configuration was initially introduced in jupyterhub/oauthenticator via https://github.com/jupyterhub/oauthenticator/pull/631, and is with this PR being upstreamed to the base Authenticator class.

## Review notes

- __Should be non-breaking for oauthenticator__
  I've concluded that defining a dedicated `@default` function for `allow_existing_users`, which we don't in jupyterhub/oauthenticator, doesn't result in that having precedence in oauthenticator or similar - so introducing this in Authenticator for JupyterHub 5 won't be breaking for oauthenticator that re-defines it (and the default value - even if we have a dedicated `@default` dectorated function.

</details>

## Todo

- [x] Add allow_all config in this PR also
- [x] Add tests
- [x] Check for documentation to update

---

- closes #4484 
- closes #4487

### DIFF
--- a/docs/source/reference/authenticators.md
+++ b/docs/source/reference/authenticators.md
@@ -41,6 +41,11 @@ When testing, it may be helpful to use the
 password unless a global password has been set. Once set, any username will
 still be accepted but the correct password will need to be provided.
 
+:::{versionadded} 5.0
+The DummyAuthenticator's default `allow_all` is True,
+unlike most other Authenticators.
+:::
+
 ## Additional Authenticators
 
 Additional authenticators can be found on GitHub
@@ -81,6 +86,7 @@ Writing an Authenticator that looks up passwords in a dictionary
 requires only overriding this one method:
 
 ```python
+from secrets import compare_digest
 from traitlets import Dict
 from jupyterhub.auth import Authenticator
 
@@ -91,8 +97,14 @@ class DictionaryAuthenticator(Authenticator):
     )
 
     async def authenticate(self, handler, data):
-        if self.passwords.get(data['username']) == data['password']:
-            return data['username']
+        username = data["username"]
+        password = data["password"]
+        check_password = self.passwords.get(username, "")
+        # always call compare_digest, for timing attacks
+        if compare_digest(check_password, password) and username in self.passwords:
+            return username
+        else:
+            return None
 ```
 
 #### Normalize usernames
@@ -190,10 +202,10 @@ via `jupyterhub --generate-config`.
 When dealing with logging in, there are generally two _separate_ steps:
 
 authentication
-: identifying who is logged in, and
+: identifying who is trying to log in, and
 
 authorization
-: deciding whether an authenticated user is logged in
+: deciding whether an authenticated user is allowed to access your JupyterHub
 
 {meth}`Authenticator.authenticate` is responsible for authenticating users.
 It is perfectly fine in the simplest cases for `Authenticator.authenticate` to be responsible for authentication _and_ authorization,
@@ -203,71 +215,137 @@ However, Authenticators also have have two methods {meth}`~.Authenticator.check_
 
 If `check_blocked_users()` returns False, authorization stops and the user is not allowed.
 
-If `check_allowed()` returns True, authorization proceeds.
+If `Authenticator.allow_all` is True OR `check_allowed()` returns True, authorization proceeds.
 
 :::{versionadded} 5.0
-{attr}`Authenticator.allow_all` and {attr}`Authenticator.allow_existing_users` are new in JupyterHub 5.0.
+{attr}`.Authenticator.allow_all` and {attr}`.Authenticator.allow_existing_users` are new in JupyterHub 5.0.
 
-By default, `allow_all` is True when `allowed_users` is empty,
-and `allow_existing_users` is True when `allowed_users` is not empty.
-This is to ensure backward-compatibility, but subclasses are free to pick more restrictive defaults.
+By default, `allow_all` is False,
+which is a change from pre-5.0, where `allow_all` was implicitly True if `allowed_users` was empty.
 :::
 
 ### Overriding `check_allowed`
 
+:::{versionchanged} 5.0
+`check_allowed()` is **not called** is `allow_all` is True.
+:::
+
+:::{versionchanged} 5.0
+Starting with 5.0, `check_allowed()` should **NOT** return True if no allow config
+is specified (`allow_all` should be used instead).
+
+:::
+
 The base implementation of {meth}`~.Authenticator.check_allowed` checks:
 
-- if `allow_all` is True, return True
 - if username is in the `allowed_users` set, return True
 - else return False
 
-If a custom Authenticator defines additional sources of `allow` configuration,
-such as membership in a group or other information,
-it should override `check_allowed` to account for this.
-`allow_` configuration should generally be _additive_,
-i.e. if permission is granted by _any_ allow configuration,
-a user should be authorized.
-
-:::{note}
-For backward-compatibility, it is the responsibility of `Authenticator.check_allowed()` to check `.allow_all`.
-This is to avoid the backward-compatible default values from granting permissions unexpectedly.
-:::
-
-If an Authenticator defines additional `allow` configuration, it must at least:
-
-1. override `check_allowed`, and
-2. override the default for `allow_all`
-
-The default for `allow_all` in a custom authenticator should be one of `False` or a dynamic default matching something like `if not any allow configuration specified`.
-False is recommended for authenticators which source much larger pools of users than are _typically_ allowed to access a Hub (e.g. generic OAuth providers like Google, GitHub, etc.).
-
-For example, here is how `PAMAuthenticator` extends the base class to add `allowed_groups`:
+:::{versionchanged} 5.0
+Prior to 5.0, the check was
 
 ```python
-from traitlets import default
-
-@default("allow_all")
-def _allow_all_default(self):
-    if self.allowed_users or self.allowed_groups:
-        # if any allow config is specified, default to False
-        return False
-    return True
-
-def check_allowed(self, username, authentication=None):
-    if self.allow_all:
-        return True
-    if self.check_allowed_groups(username, authentication):
-        return True
-    return super().check_allowed(username, authentication)
+if (not allowed_users) or username in allowed_users:
 ```
 
-Important points to note:
+but the implicit `not allowed_users` has been replaced by explicit `allow_all`, which is checked _before_ calling `check_allowed`.
+`check_allowed` **is not called** if `allow_all` is True.
 
-- overriding the default for `allow_all` is required to avoid `allow_all` being True when `allowed_groups` is specified, but `allowed_users` is not.
-- `allow_all` must be checked inside `check_allowed`
-- `allowed_groups` strictly expands who is authorized,
-  it does not apply restrictions `allowed_users`.
-  This is recommended for all `allow_` configuration added by Authenticators.
+If your Authenticator subclass similarly returns True when no allow config is defined,
+this is fully backward compatible for your users, but means `allow_all = False` has no real effect.
+
+You can make your Authenticator forward-compatible with JupyterHub 5 by defining `allow_all` as a boolean config trait on your class:
+
+```python
+class MyAuthenticator(Authenticator):
+
+    # backport allow_all from JupyterHub 5
+    allow_all = Bool(False, config=True)
+
+    def check_allowed(self, username, authentication):
+        if self.allow_all:
+            # replaces previous "if no auth config"
+            return True
+        ...
+```
+
+:::
+
+If an Authenticator defines additional sources of `allow` configuration,
+such as membership in a group or other information,
+it should override `check_allowed` to account for this.
+
+:::{note}
+`allow_` configuration should generally be _additive_,
+i.e. if access is granted by _any_ allow configuration,
+a user should be authorized.
+
+JupyterHub recommends that Authenticators applying _restrictive_ configuration should use names like `block_` or `require_`,
+and check this during `check_blocked_users` or `authenticate`, not `check_allowed`.
+:::
+
+In general, an Authenticator's skeleton should look like:
+
+```python
+class MyAuthenticator(Authenticator):
+    # backport allow_all for compatibility with JupyterHub < 5
+    allow_all = Bool(False, config=True)
+    require_something = List(config=True)
+    allowed_something = Set()
+
+    def authenticate(self, data, handler):
+        ...
+        if success:
+            return {"username": username, "auth_state": {...}}
+        else:
+            return None
+
+    def check_blocked_users(self, username, authentication=None):
+        """Apply _restrictive_ configuration"""
+
+        if self.require_something and not has_something(username, self.request_):
+            return False
+        # repeat for each restriction
+        if restriction_defined and restriction_not_met:
+            return False
+        return super().check_blocked_users(self, username, authentication)
+
+    def check_allowed(self, username, authentication=None):
+        """Apply _permissive_ configuration
+
+        Only called if check_blocked_users returns True
+        AND allow_all is False
+        """
+        if self.allow_all:
+            # check here to backport allow_all behavior
+            # from JupyterHub 5
+            # this branch will never be taken with jupyterhub >=5
+            return True
+        if self.allowed_something and user_has_something(username):
+            return True
+        # repeat for each allow
+        if allow_config and allow_met:
+            return True
+        # should always have this at the end
+        if self.allowed_users and username in self.allowed_users:
+            return True
+        # do not call super!
+        # super().check_allowed is not safe with JupyterHub < 5.0,
+        # as it will return True if allowed_users is empty
+        return False
+```
+
+Key points:
+
+- `allow_all` is backported from JupyterHub 5, for consistent behavior in all versions of JupyterHub (optional)
+- restrictive configuration is checked in `check_blocked_users`
+- if any restriction is not met, `check_blocked_users` returns False
+- permissive configuration is checked in `check_allowed`
+- if any `allow` condition is met, `check_allowed` returns True
+
+So the logical expression for a user being authorized should look like:
+
+> if ALL restrictions are met AND ANY admissions are met: user is authorized
 
 #### Custom error messages
 
--- a/docs/source/reference/authenticators.md
+++ b/docs/source/reference/authenticators.md
@@ -211,7 +211,7 @@ authorization
 It is perfectly fine in the simplest cases for `Authenticator.authenticate` to be responsible for authentication _and_ authorization,
 in which case `authenticate` may return `None` if the user is not authorized.
 
-However, Authenticators also have have two methods {meth}`~.Authenticator.check_allowed` and {meth}`~.Authenticator.check_blocked_users`, which are called after successful authentication to further check if the user is allowed.
+However, Authenticators also have two methods, {meth}`~.Authenticator.check_allowed` and {meth}`~.Authenticator.check_blocked_users`, which are called after successful authentication to further check if the user is allowed.
 
 If `check_blocked_users()` returns False, authorization stops and the user is not allowed.
 
@@ -227,7 +227,7 @@ which is a change from pre-5.0, where `allow_all` was implicitly True if `allowe
 ### Overriding `check_allowed`
 
 :::{versionchanged} 5.0
-`check_allowed()` is **not called** is `allow_all` is True.
+`check_allowed()` is **not called** if `allow_all` is True.
 :::
 
 :::{versionchanged} 5.0
@@ -242,14 +242,10 @@ The base implementation of {meth}`~.Authenticator.check_allowed` checks:
 - else return False
 
 :::{versionchanged} 5.0
-Prior to 5.0, the check was
+Prior to 5.0, this would also return True if `allowed_users` was empty.
 
-```python
-if (not allowed_users) or username in allowed_users:
-```
-
-but the implicit `not allowed_users` has been replaced by explicit `allow_all`, which is checked _before_ calling `check_allowed`.
-`check_allowed` **is not called** if `allow_all` is True.
+For clarity, this is no longer the case. A new `allow_all` property (default False) has been added which is checked _before_ calling `check_allowed`.
+If `allow_all` is True, this takes priority over `check_allowed`, which will be ignored.
 
 If your Authenticator subclass similarly returns True when no allow config is defined,
 this is fully backward compatible for your users, but means `allow_all = False` has no real effect.
@@ -349,7 +345,7 @@ So the logical expression for a user being authorized should look like:
 
 #### Custom error messages
 
-Any of these authentication and authorization methods may
+Any of these authentication and authorization methods may raise a `web.HTTPError` Exception
 
 ```python
 from tornado import web
--- a/docs/source/reference/authenticators.md
+++ b/docs/source/reference/authenticators.md
@@ -37,14 +37,14 @@ A [generic implementation](https://github.com/jupyterhub/oauthenticator/blob/mas
 ## The Dummy Authenticator
 
 When testing, it may be helpful to use the
-{class}`jupyterhub.auth.DummyAuthenticator`. This allows for any username and
-password unless if a global password has been set. Once set, any username will
+{class}`~.jupyterhub.auth.DummyAuthenticator`. This allows for any username and
+password unless a global password has been set. Once set, any username will
 still be accepted but the correct password will need to be provided.
 
 ## Additional Authenticators
 
-A partial list of other authenticators is available on the
-[JupyterHub wiki](https://github.com/jupyterhub/jupyterhub/wiki/Authenticators).
+Additional authenticators can be found on GitHub
+by searching for [topic:jupyterhub topic:authenticator](https://github.com/search?q=topic%3Ajupyterhub%20topic%3Aauthenticator&type=repositories).
 
 ## Technical Overview of Authentication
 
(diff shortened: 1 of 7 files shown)


minrk (commented):
Nice!

I do think it makes sense to promote these to the base class

[this discussion on oauthenticator](https://github.com/jupyterhub/oauthenticator/pull/729) reveals that one of the tests to add is that a user added to `admin_users` can actually login when `allow_existing_users` is False. I'm guessing it won't [without a change to initialization of users](https://github.com/jupyterhub/jupyterhub/blob/2c932997643c631f3a7874ab39ee51f6e040622d/jupyterhub/app.py#L2087-L2107).

It may also make sense to add a stage in reconciling [users already in the db](https://github.com/jupyterhub/jupyterhub/blob/2c932997643c631f3a7874ab39ee51f6e040622d/jupyterhub/app.py#L2130) based on this, to actually revoke permissions, rather than the current unconditional `auth.add_user()`, which currently _assumes_ `allow_existing_users=True`

jupyterhub/auth.py
```diff
@@ -141,6 +141,54 @@ def _deprecated_db(self):
...
+
+        .. versionadded:: 5.0
+        """,
+    )
+
+    @default("allow_existing_users")
+    def _allow_existing_users_default(self):
+        """
+        Computes the default value of allow_existing_users based on if
+        allowed_users to align with original behavior not introduce a breaking
```
consideRatio, line 185: Fix grammar

jupyterhub/auth.py
```diff
@@ -141,6 +141,54 @@ def _deprecated_db(self):
...
+        change.
+        """
+        if self.allowed_users:
+            return True
+        return False
```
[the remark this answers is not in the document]
consideRatio: Deal i'll expand this pr to include allow_all, giving it a dynamic default value that is true/false to not be breaking.

I'm not 100% confident its not breaking though if used with various authenticator logics, i'll have to think a bit more on if its non-breaking.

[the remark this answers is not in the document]
minrk: Since we're preparing a major release, as long as it's not _usually_ breaking and desirable behavior is readily recoverable, and default behavior is at least as restrictive as now, I think we're okay.

I think that will be true, but thinking through the implications of the change with Authenticators is a good plan nonetheless, and should probably end up in the docs.

I think we should explicitly document that while `allow_all` is `not bool(allowed_users)` by default, Authenticator subclasses should:

1. be False by default for any authenticator that has a pool of users much larger than those who should have access (e.g. general user accounts, federated login, etc.),
2. need a custom dynamic default if they add any allow source beyond `allowed_users`

and include an example of a custom `@default("allow_all")`.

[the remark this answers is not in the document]
minrk: I think probably the biggest, clearest change to make that simpler is to make allow_all only explicitly opt-in (I.e. effectively require all JupyterHub to have at least one allow config for anyone to be able to use it). I'm not a fan of that, but it does solve a hypothetical problem.

I'm less concerned about the scenario of removing the last user and then allowing unintentionally, because that doesn't seem very realistic, but it does make sense in principle. One way could be to make a clear distinction between _unset_ and specified but empty. I don't necessarily agree that it's surprising that going from specifying one user to not specifying allow_users at all would have this effect. But it _does_ seem reasonable to expect explicitly specifying an empty set to mean not allowing any users by name. We can investigate what that might look like, and what else might break as a result (e.g. use None as default).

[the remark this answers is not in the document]
minrk: I'll also note that the defaults here are unchanged from current behavior. This PR aims to encode longstanding behavior in variables that can be overridden explicitly for clearer, predictable behavior in edge cases. So I think this PR ought to be purely an improvement in terms of simplicity and clarity without any changes to defaults.

consideRatio:
I'm finding myself very time constrained @minrk, I think I can't get this to land this in a while =/

minrk:
No worries, do you want me to pick it up? I keep doing other things, but I really want to get 4.1 out, then I'll come back to this one if you want. I think after our conversations I know what to do.

consideRatio:
You are most welcome to pick it up, but I can try to find time to work it in the future as well. I'll prioritize review efforts if you pick it up, i never want to run out of capacity to review to unblock others work.

minrk:
I finished some other things waiting on review, so I can pick this one up now.

minrk:
I've added allow_all, added as many test combinations as I could think of, and updated the docs. I think this is ready for review.

minrk (commented):
I think the main remaining decision is whether it makes sense to simplify the default `allow_all` to be just False.

Personally, I think it's a better and more logical experience for most Authenticators to allow all users who can authenticate by default when no auth config is provided. The exception being Authenticators that reference large pools of users unrelated to the Hub (i.e. OAuthenticator), where allow_all is still common, but makes sense to be opt-in.

Choices:

1. set allow_all = False by default, and
  a. make True always opt-in for users, or
  b override default where True makes sense (PAM, dummy, LDAP, FirstUse, etc.)
2. preserve current default behavior (if no auth config, allow all)

To me, the biggest advantage of strict `allow_all = False` by default with no dynamic logic is that it would allow JupyterHub to check it _outside_ `check_allowed()`, meaning it would be immediately available to all Authenticators, without needing to update `check_allowed` for support.

And the biggest disadvantage of this PR as it is is that `allow_all` has no effect for Authenticators that have overridden `check_allowed`, to ensure that existing configuration does not change who is allowed after upgrade.

jupyterhub/auth.py
```diff
@@ -144,6 +144,98 @@ def _deprecated_db(self):
...
+           JupyterHub admin panel at `/hub/admin`. You may find users existing
+           there because they have previously been declared in config such as
+           `allowed_users` or allowed to sign in.
+
+        .. warning::
+
+           When this is enabled and you wish to remove access for one or more
+           users previously allowed, you must make sure that they
+           are removed from the jupyterhub database. This can be tricky to do
+           if you stop allowing a group of externally managed users for example.
```
consideRatio, line 219:
```suggestion
           if you stop allowing an externally managed group of users for example.
```

jupyterhub/auth.py
```diff
@@ -677,25 +768,33 @@ def add_user(self, user):
         """Hook called when a user is added to JupyterHub
 
         This is called:
-         - When a user first authenticates
-         - When the hub restarts, for all users.
+         - When a user first authenticates, _after_ all allow and block checks have passed
+         - When the hub restarts, for all users in the database (i.e. users previously allowed)
```
consideRatio, line 772:
```suggestion
         - When the hub restarts, for all users in the database (i.e. users previously allowed)
         - When a user is added to the database, either via configuration or REST API
```

jupyterhub/auth.py
```diff
@@ -677,25 +768,33 @@ def add_user(self, user):
...
+        allow_existing_users is true.
 
         Subclasses may do more extensive things, such as adding actual unix users,
         but they should call super to ensure the allowed_users set is updated.
 
         Note that this should be idempotent, since it is called whenever the hub restarts
         for all users.
 
+        .. versionchanged:: 5.0
+           Now adds users to the allowed_users set if allow_all is False,
```
consideRatio, line 786:
```suggestion
           Now adds users to the allowed_users set if allow_existing_users is True,
```

jupyterhub/auth.py
```diff
@@ -144,6 +144,98 @@ def _deprecated_db(self):
...
+        .. versionadded:: 5.0
+        """,
+    )
+
+    @default("allow_all")
+    def _default_allow_all(self):
+        if self.allowed_users:
+            return False
+        else:
+            return True
```
consideRatio, line 191: We now have a dynamic default for `allow_existing_users` and `allow_all`.

Both `allow_all` and `allow_existing_users` in the Authenticator class get their default based on if `allowed_users` is configured truthy or not.

If someone would configure only `allow_existing_users`, they would still get `allow_all` set to True by default, making `allow_existing_users` pointless as everyone is allowed anyhow.

I'm not sure how to handle that, but its certainly very messy to have dynamic defaults. On the other hand, if we fix the values we end up needing to think about how both existing and new authenticators are influenced when using new/old versions of jupyterhub - which is very complicated as well.

I think it may be safe to configure `allow_all` to True `if not self.allowed_users and not self.allow_existing_users` - what do you think?

jupyterhub/auth.py
```diff
@@ -902,23 +1001,25 @@ def _add_user_cmd_default(self):
...
-    def _allowed_groups_changed(self, change):
-        """Log a warning if mutually exclusive user and group allowed sets are specified."""
-        if self.allowed_users:
-            self.log.warning(
-                "Ignoring Authenticator.allowed_users set because Authenticator.allowed_groups supplied!"
-            )
+    @default("allow_all")
+    def _allow_all_default(self):
+        if self.allowed_users or self.allowed_groups:
+            # if any allow config is specified, default to False
```
consideRatio, line 1013: Should include `allow_existing_users` also, right? Discussed also in https://github.com/jupyterhub/jupyterhub/pull/4701/files#r1531617284

jupyterhub/auth.py
minrk: > I think it may be safe to configure allow_all to True if not self.allowed_users and not self.allow_existing_users - what do you think?

I think that's true, good catch. The general rule should be "True if no other allow config", and I totally forgot that `allow_existing_users` _on its own_ could be used to _only_ admit users already in the db, so it definitely qualifies.

WDYT about the case for allow_all=False after this? I think if we set allow_all=False unconditionally, a lot gets simpler because then JupyterHub can be responsible for interpreting `allow_all`, and Authenticators don't need to know about it at all. The only reason we don't do that is to avoid the dynamic default allowing users access. The only downside I see is it's the same breaking change we made in oauthenticator (which was a bit painful), and it makes less sense to prevent that for most other Authenticators. We can still set it to true by default for Dummy (and maybe PAM).

Another mechanism for the dynamic default is to use traitlets metadata to tag all config options that explciitly admit users (e.g. `allowed_users = Set(...).tag(allows_access=True)`), and then allow_all can be False if any of those traits are set. We wuld still have the problem of Authenticators needing to be updated to support this.

consideRatio: > WDYT about the case for allow_all=False after this?

I'm very happy about the idea of getting to a state of no longer providing dynamic defaults, but I also find reviewing the consequences of it very complicated. I figure we need to consider the combinations of jupyterhub before/after the change and authenticator classes before/after adjusting to such change.

minrk: OK, I'll take a stab at simple False and update the docs to match to hopefully cover all of that.

consideRatio: OK! The aim is for this to be shipped with 5.x only now right?

minrk: Yes, definitely. Breaking changes here are okay, as long as they actually improve things for users and don't _expand_ access for any existing config.

jupyterhub/tests/mocking.py
```diff
@@ -243,7 +243,8 @@ def __init__(self, *args, **kwargs):
             cert_location = kwargs['internal_certs_location']
             kwargs['external_certs'] = ssl_setup(cert_location, 'hub-ca')
         super().__init__(*args, **kwargs)
-        self.config.Authenticator.allow_all = True
+        if 'allow_all' not in self.config.Authenticator:
+            self.config.Authenticatorallow_all = True
```
consideRatio, line 247:
```suggestion
            self.config.Authenticator.allow_all = True
```

minrk:
Okay, I believe I've updated the code and docs for `allow_all=False` by default. I think it is clearer, and there's a substantial benefit to allow_all being handled _outside_ `check_allowed` (all Authenticators get it immediately), which we can only get if we don't have a dynamic default that needs to understand each Authenticator's config.

I've gone with only applying allow_all=True only to the DummyAuthenticator and not PAM. I think applying it to PAM would be sensible and results in less surprising default _behavior_, but is more complex to communicate about when the _default_ implementation has different default behavior from the _base_ implementation. And the cost of "nobody is allowed at all" is immediately apparent and easy to toggle with one line of backward-compatible config, so I think it is minimally disruptive (though not _not_ disruptive). Plus, the cost of unexpected defaults (which can always go either way) makes sense in the more restrictive direction rather than the more permissive direction. Surprisingly restrictive is both immediately apparent and safe, while surprisingly permissive is _not_ immediately apparent and a security problem, and we don't really get to decide which direction is surprising.

docs/source/reference/authenticators.md
```diff
@@ -188,6 +195,170 @@ Additionally, configurable attributes for your authenticator will
...
+: identifying who is trying to log in, and
+
+authorization
+: deciding whether an authenticated user is allowed to access your JupyterHub
+
+{meth}`Authenticator.authenticate` is responsible for authenticating users.
+It is perfectly fine in the simplest cases for `Authenticator.authenticate` to be responsible for authentication _and_ authorization,
+in which case `authenticate` may return `None` if the user is not authorized.
+
+However, Authenticators also have have two methods {meth}`~.Authenticator.check_allowed` and {meth}`~.Authenticator.check_blocked_users`, which are called after successful authentication to further check if the user is allowed.
```
manics, line 214:
```suggestion
However, Authenticators also have two methods, {meth}`~.Authenticator.check_allowed` and {meth}`~.Authenticator.check_blocked_users`, which are called after successful authentication to further check if the user is allowed.
```

docs/source/reference/authenticators.md
```diff
@@ -188,6 +195,170 @@ Additionally, configurable attributes for your authenticator will
...
+{attr}`.Authenticator.allow_all` and {attr}`.Authenticator.allow_existing_users` are new in JupyterHub 5.0.
+
+By default, `allow_all` is False,
+which is a change from pre-5.0, where `allow_all` was implicitly True if `allowed_users` was empty.
+:::
+
+### Overriding `check_allowed`
+
+:::{versionchanged} 5.0
+`check_allowed()` is **not called** is `allow_all` is True.
```
manics, line 230:
```suggestion
`check_allowed()` is **not called** if `allow_all` is True.
```

docs/source/reference/authenticators.md
```diff
@@ -188,6 +195,170 @@ Additionally, configurable attributes for your authenticator will
...
+:::
+
+:::{versionchanged} 5.0
+Starting with 5.0, `check_allowed()` should **NOT** return True if no allow config
+is specified (`allow_all` should be used instead).
```
manics, line 235: Is there anything wrong with relying on `check_allowed()` to do everything in a custom authenticator?

docs/source/reference/authenticators.md
```diff
@@ -188,6 +195,170 @@ Additionally, configurable attributes for your authenticator will
...
+
+- if username is in the `allowed_users` set, return True
+- else return False
+
+:::{versionchanged} 5.0
+Prior to 5.0, the check was
+
+```python
+if (not allowed_users) or username in allowed_users:
+```
```
manics, line 249:
```suggestion
Prior to 5.0, this would also return True if `allowed_users` was empty.
```

docs/source/reference/authenticators.md
```diff
@@ -188,6 +195,170 @@ Additionally, configurable attributes for your authenticator will
...
+
+but the implicit `not allowed_users` has been replaced by explicit `allow_all`, which is checked _before_ calling `check_allowed`.
```
manics, line 251:
```suggestion
For clarity this is no longer the case. A new `allow_all` property (default False) has been added which is checked _before_ calling `check_allowed`.
```

docs/source/reference/authenticators.md
```diff
@@ -188,6 +195,170 @@ Additionally, configurable attributes for your authenticator will
...
+`check_allowed` **is not called** if `allow_all` is True.
```
manics, line 252:
```suggestion
If `allow_all` is True this takes priority over `check_allowed` which will be ignored.
```

docs/source/reference/authenticators.md
```diff
@@ -188,6 +195,170 @@ Additionally, configurable attributes for your authenticator will
...
+If your Authenticator subclass similarly returns True when no allow config is defined,
+this is fully backward compatible for your users, but means `allow_all = False` has no real effect.
+
+You can make your Authenticator forward-compatible with JupyterHub 5 by defining `allow_all` as a boolean config trait on your class:
+
+```python
+class MyAuthenticator(Authenticator):
+
+    # backport allow_all from JupyterHub 5
+    allow_all = Bool(False, config=True)
```
manics, line 263: I can't remember how Traitlets works- if someone sets `c.Authenticator.allow_all = True|False` will this also set `MyAuthenticator.allow_all` if the property is redefined in the subclass?

docs/source/reference/authenticators.md
```diff
@@ -188,6 +195,170 @@ Additionally, configurable attributes for your authenticator will
...
+- permissive configuration is checked in `check_allowed`
+- if any `allow` condition is met, `check_allowed` returns True
+
+So the logical expression for a user being authorized should look like:
+
+> if ALL restrictions are met AND ANY admissions are met: user is authorized
+
+#### Custom error messages
+
+Any of these authentication and authorization methods may
```
manics, line 352:
```suggestion
Any of these authentication and authorization methods may raise a `web.HTTPError` Exception
```

jupyterhub/auth.py
```diff
@@ -121,6 +121,54 @@ def _deprecated_db(self):
         """
     ).tag(config=True)
 
+    any_allow_config = Bool(
+        False,
+        help="""Is there any allow config?
+        
+        Used to show a warning if it looks like nobody can access the HUb,
```
manics, line 128:
```suggestion
        Used to show a warning if it looks like nobody can access the Hub,
```

jupyterhub/auth.py
```diff
@@ -121,6 +121,54 @@ def _deprecated_db(self):
...
+        authenticator configuration problems.
+
+        .. versionadded:: 5.0
+        """
+        if not self.any_allow_config:
+            self.log.warning(
+                "No allow config found, it's possible that nobody can login to your Hub!\n"
+                "You can set `c.Authenticator.allow_all = True` to allow any user who can login to access the Hub,\n"
+                "or e.g. `allowed_users` to a set of users who should have access."
+            )
```
manics, line 170: I we're mentioning suggested steps here should we also mention `any_allow_config`?

jupyterhub/auth.py
```diff
@@ -144,6 +192,83 @@ def _deprecated_db(self):
...
     ).tag(config=True)
 
+    allow_all = Bool(
+        False,
+        config=True,
+        help="""
+        Allow every user who can successfully authenticate to access JupyterHub.
+        
+        False by default, which means for most Authenticators,
+        _some_ allow-related configuration is required to allow any users to log in.
```
manics, line 202:
```suggestion
        _some_ allow-related configuration is required to allow users to log in.
```

jupyterhub/auth.py
```diff
@@ -144,6 +192,83 @@ def _deprecated_db(self):
...
+
+        Authenticator subclasses may override the default with e.g.::
```
manics, line 204: Are you mentioning this just for backwards compatibility? If so I think it's worth saying that, for new authenticators is it better for them to require explicit configuration?

docs/source/tutorial/getting-started/authenticators-users-basics.md
```diff
@@ -68,26 +105,54 @@ group. For example, we can let any user in the `wheel` group be an admin:
...
-then admins have permission to log in _as other users_ on their
-respective machines for debugging. **As a courtesy, you should make
-sure your users know if admin_access is enabled.**
+By default, only the deprecated `admin` role has global `access` permissions.
+**As a courtesy, you should make sure your users know if admin access is enabled.**
 
 ## Add or remove users from the Hub
 
+:::{versionadded} 5.0
+`c.Authenticator.allow_existing_users` is added in 5.0 and enabled by default.
```
manics, line 133: `enabled by default` is unclear, since the value is conditional

docs/source/reference/authenticators.md
minrk: Not _really_. The only real downside is that it becomes impossible to specify "nobody should be allowed." if the total absence of config results in `allow_all`. I'd stand by this statement, though, as a recommendation to Authenticators.

docs/source/reference/authenticators.md
minrk: Yes, redefining the trait will still inherit the parent config. Config is loaded if it's tagged as `config=True` on the loading class and the config section name matches any parent Configurable class. Only if MyAuthenticator redefined it without `config=True` would it not load `Authenticator.allow_all`.

jupyterhub/auth.py
minrk: I don't know that it's better! I think it's for Authenticators to decide what makes sense, so I'd leave this as a statement of fact that it _may_ be overridden.

docs/source/tutorial/getting-started/authenticators-users-basics.md
minrk: Good catch. This is a remnant from _either_ allow_existing or allow_all always being True, so _in practice_ this was always True by default. Now it's more strictly conditional.

minrk: Clarified.

minrk:
Thanks for the reviews! I think everything is addressed as far as I can tell. Anything folks want to look at or talk about more?

jupyterhub/tests/test_auth.py
```diff
@@ -593,3 +606,195 @@ async def test_auth_managed_groups(
...
+    if allowed_letters:
+        kwargs["allowed_letters"] = tuple(allowed_letters.split(','))
+
+    authenticator = AllowAllIgnoringAuthenticator(**kwargs)
+
+    # load one user from db
+    existing_user = add_user(app.db, app, name="existing")
+    authenticator.add_user(existing_user)
+
+    if allowed:
```
manics, line 779: The `allowed` test parameter is always non-empty.

manics (commented):
I've got a couple of comments on the tests, but given there are no further comments from anyone else I think we should merge this! We can refine it prior to the 5.0.0 release if necessary.

jupyterhub/tests/test_auth.py
```diff
@@ -593,3 +606,195 @@ async def test_auth_managed_groups(
         assert not app.db.dirty
         groups = sorted(g.name for g in user.groups)
         assert groups == expected_refresh_groups
+
+
+@pytest.mark.parametrize(
+    "allowed_users, allow_all, allow_existing_users",
```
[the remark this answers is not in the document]
minrk: yeah, artifact of when this had a dynamic default. removed the conditions.

jupyterhub/tests/test_auth.py
minrk: True. I'll think about more test cases where that might be empty, but I think it's okay for now (I think it can only be the case when `allow_all=False`, which means changing the parametrization).

minrk:
Thanks @consideRatio and @manics!

consideRatio:
Wow thank you @minrk and @manics for working this to completion!!

Issue #4653, opened 2023-12-04
Minor broken docs: reference to wiki


benjimin:
Minor broken docs: [this page](https://jupyterhub.readthedocs.io/en/stable/reference/spawners.html#examples) has a link titled "Jupyterhub Wiki", but it directs to GitHub, _not_ to any wiki. 

(I think the docs should be changed to update either the url or label used by the link, but I'm not sure which.. does a wiki exist?)

minrk:
Thanks, it does link to the wiki, but since we stopped using the wiki and a disabled wiki link redirects back to the original repo. I think this link can just be removed.

rizz-sd:
Hi fellow members.
I am new to Open source community. Given that wiki link doesn't exist anymore and we just need to remove the link reference and modify some text, this looks like an easy fix. Can I take this up?

minrk:
@rizz-sd yes, please go ahead!

PR #4706, opened 2024-02-13
Update allowed_users docstring

minrk:
conflict less with Authenticator subclasses that extend allow behavior.

@consideRatio feel free to take this and update it for `allow_all` in #4653, if you want, rather than as a standalone PR.


consideRatio:
Thank you @minrk!!

PR #4713, opened 2024-02-27
resolve paths in disable_user_config

### LINKED ISSUE #3804: Unable to disable user config with Jupyter Server
The hub administrator is supposed to be able to prevent per-user notebook configuration scripts from running by setting

```
c.Spawner.disable_user_config = True
```

In the `jupyterhub_config.py` config.  This sets the environment variable `JUPYTERHUB_DISABLE_USER_CONFIG=1` for the spawned notebook server.  However this seems to be being ignored?

<details>
<summary>Using this Dockerfile</summary>

```
FROM jupyterhub/jupyterhub:2

RUN python3 -m pip install --no-cache jupyterlab

RUN \
    adduser -q --gecos "" --disabled-password user1     && \
    echo user1:user1 | chpasswd

ADD jupyterhub_config.py .

RUN mkdir -p /home/user1/.jupyter
ADD jupyter_notebook_config.py /home/user1/.jupyter/.
RUN chown -R user1:user1 /home/user1/.jupyter

CMD ["jupyterhub"]
```

</details>

<details><summary>
with this `jupyterhub_config.py` and example notebook config for `user1`:
</summary>

```
c.Spawner.disable_user_config = True
```

```
import os

print("HELLO FROM THE NOTEBOOK CONFIG")
print(os.getenv("JUPYTERHUB_DISABLE_USER_CONFIG"))

c.ServerApp.shutdown_no_activity_timeout = 600
c.MappingKernelManager.cull_idle_timeout = 600
c.TerminalManager.cull_inactive_timeout = 600
```

</details>

I see the "HELLO" message and the value 1 printed when the notebook starts up, and the timeout message indicating that my config setting is in effect:

```
[I 2022-02-22 22:35:23.167 SingleUserLabApp serverapp:2161] Will shut down after 600 seconds with no kernels or terminals.
```

Am I misunderstanding exactly what config files are excluded?  I see there's a test for this but I wonder is it actually verifying that the config is being ignored?

minrk:
handles home directory being a symlink, which would not be excluded before.

Workaround is to resolve $HOME to its real path:

```bash
export HOME=$(realpath $HOME)
```

prior to launching single-user servers

closes #3804

### DIFF
--- a/jupyterhub/singleuser/_disable_user_config.py
+++ b/jupyterhub/singleuser/_disable_user_config.py
@@ -20,8 +20,22 @@ rather than keeing these monkey patches around.
 """
 
 import os
+from pathlib import Path
 
 from jupyter_core import paths
+
+
+def _is_relative_to(path, prefix):
+    """
+    Backport Path.is_relative_to for Python < 3.9
+
+    added in Python 3.9
+    """
+    if hasattr(path, "is_relative_to"):
+        # Python >= 3.9
+        return path.is_relative_to(prefix)
+    else:
+        return path == prefix or prefix in path.parents
 
 
 def _exclude_home(path_list):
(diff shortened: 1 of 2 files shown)


consideRatio (approved):

minrk:

PR #4717, opened 2024-03-01
allow callable values in c.JupyterHub.template_vars

kreuzert:
Hi everyone,

we're currently facing some limitations regarding the jinja template variables in JupyterHub. We would like to pass variables in the Jinja templates, that might change during runtime. 

In the current JupyterHub code, the `c.JupyterHub.template_vars` are defined while the config is loaded. This pull request would allow us to pass callable objects. 
I've also added the option to pass the current user as argument, if the callable value in `template_vars` requires an argument. This gives us even more options, to pass user specific values as jinja template variable to the frontend.

Example jupyterhub_config.py:
```
import yaml
def load_file():
    changing_file = "/mnt/.../changing_file.yaml"
    with open(changing_file, "r") as f:
        ret = yaml.full_load(f)
    return ret

def load_file_user_specific(current_user):
    changing_file = "/mnt/.../changing_file.yaml"
    with open(changing_file, "r") as f:
        ret = yaml.full_load(f)
    ret = ret.replace("<user>", current_user.name)
    return ret

c.JupyterHub.template_vars = {
    "changing_file_content": load_file,
    "changing_file_content_user_specific": load_file_user_specific
}
```

Use Cases:
In the kubernetes "Zero2JupyterHub" context, one can use a `ConfigMap` Resource in the frontend and change it's value, without restarting the hub pod. 
Example helm chart value.yaml for z2jh:
```
hub:
  extraVolumes:
    - name: changing-file
      configMap:
        defaultMode: 400
  extraVolumeMounts:
    - name: changing-file
      mountPath: /mnt/...
```

Downside of this patch, which might (or might not) require some attention:
- The `.update` function of python supports tuples as argument. This would not be possible right now. But since `c.JupyterHub.template_vars` is of type `traitlets.Dict` anyway, it would not change JupyterHubs behaviour right now.

Please let me know your thoughts about this small change.


jupyterhub/handlers/base.py
```diff
@@ -1348,7 +1349,13 @@ def template_namespace(self):
             xsrf=self.xsrf_token.decode('ascii'),
         )
         if self.settings['template_vars']:
-            ns.update(self.settings['template_vars'])
+            for key, value in self.settings['template_vars'].items():
+                if callable(value):
```
minrk, line 1353: So far, all callable hooks have a single defined signature, rather than variable by inspection, which simplifies things a bit. What about calling it with `f(user)` unconditionally?

And should they be allowed to be async? If so, add:

```python
            if inspect.isawaitable(value):
                value = await value
```

minrk: ah, I recalled that templates may be async, but for backward-compatibility, `render_template` must be _allowed_ to be called synchronously. I'll think about that. Let's go with sync-only for now.

I think if you return awaitables from the template namespace, you'll be able to await them _in_ the template, assuming the page is rendered async (all default pages are, I believe).

kreuzert:
Please let me know, if anything else is recommended or required. 
Otherwise I would leave this suggestion as it is.

minrk:
Looks great, thank you!

PR #4718, opened 2024-03-02
[WIP] Add system monitor for API endpoint

mtekman:
### Motivation

I want to quickly peek at my JupyterHub machine to see what resources I have free before I spawn a new user server.

I also want to show this information to my users at the spawn page, to make them more aware of the finite resources we actually have.

### Usage

```bash
curl -H "Authorization: token <read:hub token> https://<server>/hub/api/sysmon
```

```json
{
   "last_updated": "Sat Mar  2 12:57:57 2024", 
   "seconds_interval": 20,
   "ram_free_gb": 192.120967168, 
   "ram_used_gb": 0.851726336,
   "ram_total_gb": 202.312220672,
   "ram_free_percent": 95,
   "ram_used_percent": 1.3,
   "cpu_usage_percent": 0,
   "cpu_count": 36
}
```

These results are cached and only updated every `seconds_interval` which is by default 10, but can be easily configured via `c.JupyterHub.sysmon_interval` setting.

Issue #3818, opened 2022-03-09
View/Limit memory usage per user via admin view


alejandrofm:
### Proposed change
To have more information/control of the instance resources used by Jupyterhub it would be nice and UX friendly to have in the admin panel: http://jupyterhub2.bi.etermax.net/admin Information about current memory usage, and as a plus it would be great to limit that for each user to that limit shows on each server via the jupyterlab-system-monitor plugin or similar.

### Alternative options
This way if the instance has 128gb of memory I could assign 100gb as max, so no one will be hoarding ram on processes that need more than 128gb.

### Who would use this feature?
Jupyterhub users

minrk:
This will require quite a bit of new JupyterHub APIs, since JupyterHub doesn't understand memory usage, nor a generic way to apply limits on a per-user basis.

I proposed adding generic metrics some time ago in https://github.com/jupyterhub/jupyterhub/issues/3101 which would help with one part, but I'm not sure how to turn that into applying limits via UI.

You _can_ set memory limits (if the Spawner supports it) via config with `c.Spawner.mem_limit = '2G'`. This can be a callable that takes username or group membership into account.

WilliamJudge94:
@alejandrofm, @minrk I have created a repository to do just this. By running the jupyterhub_auto.sh script in a terminal you can automatically do the following:

- Install conda
- Install JupyterHub
- Install systemdspawner for resource managment
- Dynamically allocated RAM + CPU usage per user
- Creates a shared folder
- Allows for customization of JupyterHub
- Adds SSL encryption to Jupyterhub
- Creates a jupyterhub service
- Allows for dynamic user privileges between the MANAGER and SharedFiles group

You can check out the code here: https://github.com/WilliamJudge94/jupyterhub-auto

minrk:
@WilliamJudge94 that's great! Maybe we can use that as a reference to figure out what can be generalized across Spawners, which is the hard part for adoption in JupyterHub itself.

FWIW, we also can't use any code directly from that repo in jupyterhub or any of our Spawners, due to the choice of the viral GPL license. I'm not sure if that was a deliberate choice or not.

WilliamJudge94:
@minrk, I have updated the license. Hopefully, bits and pieces of my code can be helpful to the JupyterHub community!

minrk:
Awesome! I'm sure it will be useful to specific deployments, and hopefully we can figure out which parts make sense to upstream so they can be effectively shared where implementations are feasible.

[...]

PR #4719, opened 2024-03-02
Add system monitor for API endpoint

mtekman:
Continuation of: #4718 

### Motivation

I want to quickly peek at my JupyterHub machine to see what resources I have free before I spawn a new user server.

I also want to show this information to my users at the spawn page, to make them more aware of the finite resources we actually have.

### Usage

```bash
curl -H "Authorization: token <read:hub token> https://<server>/hub/api/sysmon
```

```json
{
   "last_updated": "Sat Mar  2 12:57:57 2024", 
   "seconds_interval": 20,
   "ram_free_gb": 192.120967168, 
   "ram_used_gb": 0.851726336,
   "ram_total_gb": 202.312220672,
   "ram_free_percent": 95,
   "ram_used_percent": 1.3,
   "cpu_usage_percent": 0,
   "cpu_count": 36
}
```

These results are cached and only updated every `seconds_interval` which is by default 10, but can be easily configured via `c.JupyterHub.sysmon_interval` setting.

### DIFF
--- a/docs/source/_static/rest-api.yml
+++ b/docs/source/_static/rest-api.yml
@@ -127,7 +127,7 @@ paths:
                     description: |
                       The percent of CPU usage, rounded to nearest integer
                       The output of psutil.cpu_percent()
-                  cpu_count :
+                  cpu_count:
                     type: number
                     description: The number of CPUs, as output by psutil.cpu_count()
       security:
--- a/docs/source/_static/rest-api.yml
+++ b/docs/source/_static/rest-api.yml
@@ -80,6 +80,59 @@ paths:
       security:
         - oauth2:
             - read:hub
+  /sysmon:
+    get:
+      summary: Get system monitor information
+      description: |
+        CPU and memory usage of the server as a whole
+      responses:
+        200:
+          description: CPU and Memory usage, availability, and totals in GB or percent
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  last_updated:
+                    type: string
+                    description: The local date and time, as returned by time.ctime
+                  seconds_interval:
+                    type: number
+                    description: The interval in which the metrics are updated in seconds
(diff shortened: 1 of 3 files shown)


mtekman:
I guess the legacy_notebook tests are expected to fail? Should I add `psutil` to the requirements of the old dependencies?

manics:
This is a nice idea! Unfortunately I don't think it's suitable for JupyterHub at the moment.

This only works for the single case where JupyterHub and all singleuser servers runs on a single server with the default LocalProcessSpawner, and users are familiar with how CPU/memory is managed. Many JupyterHubs use a different spawner, for example to run singleuser servers in Docker on a single or multiple hosts, Kubernetes, on HPC, etc.

JupyterHub is already designed to integrate with Prometheus metrics
https://github.com/jupyterhub/jupyterhub/blob/9c3f98d4273c7ebc7a4d8bc5e05dc0d97c4e4d3e/jupyterhub/metrics.py#L38

Also see this related issue for a bit more background
https://github.com/jupyterhub/jupyterhub/issues/3818

mtekman:
This just shows the usage stats of the machine that JupyterHub is installed on, and does not concern itself with the machine(s) where the JupyterLab/Notebook kernels are spawned.

Even using SwarmDocker or Kubernetes, surely the Hub itself is still installed on a single machine? The other servers are merely job runners, right?

mtekman:
Also, the metrics that I'm seeing at "`/metrics`" endpoint do not seem to yield any RAM or CPU info, unless I'm interpreting the output wrong. Is there a way to configure Prometheus to show this? Sorry if this is all in the wiki

mtekman:
Okay, I will close this -- it looks like I need to setup a graphing library like Grafana to probe the Prometheus metrics.

I wonder if we can get per user metrics working, I will watch that other issue with interest

Issue #3816, opened 2022-03-04
Sorting option on the admin page considers only the servers listed on the current page


szazs89:
### Bug description
Sorting option on the JupyterHub admin page considers only the servers listed on the current page.

#### Expected behaviour
All the servers should be sorted first and then paginated. (It may require to reload the list of servers...)

#### Actual behaviour
Clicking on one of the sorting option (User or Last Activity, e.g.) on the Admin page, sorts the list displayed on the current or actual page, not the whole.

### How to reproduce
1. Go to `/hub/admin` Admin page
2. Click on one of the up/down arrows in the header of the table
3. Click on Next at the bottom of the table
4. The pages are sorted separately

### Your personal set up
`docker image pull jupyterhub/jupyterhub`

 - OS: Ubuntu 20.04.4 LTS
 - Version(s): JupyterHub 2.0.0, Python 3.8.10
- <details><summary>Full environment</summary>
  
  ```
  alembic==1.7.3
  argon2-cffi==21.3.0
  argon2-cffi-bindings==21.2.0
  asttokens==2.0.5
  async-generator @ file:///tmp/wheelhouse/async_generator-1.10-py3-none-any.whl
  attrs @ file:///tmp/wheelhouse/attrs-21.2.0-py2.py3-none-any.whl
  backcall==0.2.0
  black==21.12b0
  bleach==4.1.0
  certifi @ file:///tmp/wheelhouse/certifi-2021.10.8-py2.py3-none-any.whl
  certipy @ file:///tmp/wheelhouse/certipy-0.1.3-py3-none-any.whl
  cffi @ file:///tmp/wheelhouse/cffi-1.15.0-cp38-cp38-manylinux_2_12_x86_64.manylinux2010_x86_64.whl
  charset-normalizer @ file:///tmp/wheelhouse/charset_normalizer-2.0.8-py3-none-any.whl
  click==8.0.3
  cryptography @ file:///tmp/wheelhouse/cryptography-36.0.0-cp36-abi3-manylinux_2_24_x86_64.whl
  cycler==0.11.0
  debugpy==1.5.1
  decorator==5.1.1
  defusedxml==0.7.1
  entrypoints @ file:///tmp/wheelhouse/entrypoints-0.3-py2.py3-none-any.whl
  executing==0.8.2
  fonttools==4.28.5
[...]

PR #4720, opened 2024-03-04
admin: persist page view parameters in url

### LINKED ISSUE #4709: modifyable api_page_default_limit variable on the admin page
### Proposed change

It is rather annoying that the number of the listed servers (`api_page_default_limit`?) cannot be changed on the admin page,
but only use the number fixed in `jupyterhub_config.py`.

It would help the administration of jupyter HUB very much if this number could be changed on the admin page.

### Alternative options

The variable `api_page_default_limit` can be changed in `jupyterhub_config.py`, however, this requires restarting the server.

Most of the users with _admin_ access.

Completing the admin page template(?) with a choice box (50|100|200|500) or input box.

minrk:
- persist offset, limit, name_filter in URL parameters, so they are stable across page reload
- add UI element to specify items per page (closes #4709)

This allows specifying a URL (e.g. `/hub/admin#/?limit=50&name_filter=student`), which will show a specific view of a page of users. it also means the view persists across a reload.

It specifically does _not_ address #3816, which I want to do in one or two separate PRs. That one's trickier because it needs to rewrite how sorting is implemented, since client-side sorting should be entirely removed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,5 @@ docs/source/reference/metrics.rst
 oldest-requirements.txt
 jupyterhub-proxy.pid
 examples/server-api/service-token
+
+*.hot-update*
--- a/jsx/package.json
+++ b/jsx/package.json
@@ -39,8 +39,7 @@
     "react-icons": "^4.8.0",
     "react-multi-select-component": "^4.3.4",
     "react-redux": "^7.2.8",
-    "react-router-dom": "^5.3.4",
-    "react-router-dom-v5-compat": "^6.22.2",
+    "react-router-dom": "^6.22.2",
     "recompose": "npm:react-recompose@^0.33.0",
     "redux": "^4.2.1",
     "regenerator-runtime": "^0.13.11"
--- a/jsx/package.json
+++ b/jsx/package.json
@@ -40,6 +40,7 @@
     "react-multi-select-component": "^4.3.4",
     "react-redux": "^7.2.8",
     "react-router-dom": "^5.3.4",
+    "react-router-dom-v5-compat": "^6.22.2",
     "recompose": "npm:react-recompose@^0.33.0",
     "redux": "^4.2.1",
     "regenerator-runtime": "^0.13.11"
--- a/jsx/package.json
+++ b/jsx/package.json
@@ -7,8 +7,8 @@
   "license": "BSD-3-Clause",
   "scripts": {
     "build": "webpack",
+    "build:watch": "webpack watch",
     "hot": "webpack && webpack-dev-server",
-    "place": "cp build/admin-react.js* ../share/jupyterhub/static/js/",
     "test": "jest --verbose",
     "snap": "jest --updateSnapshot",
     "lint": "eslint --ext .jsx --ext .js src/",
--- a/jsx/package.json
+++ b/jsx/package.json
@@ -65,7 +65,6 @@
     "jest": "^29.5.0",
     "jest-environment-jsdom": "^29.5.0",
     "prettier": "^2.8.7",
-    "sinon": "^15.0.3",
     "style-loader": "^3.3.2",
     "webpack": "^5.79.0",
     "webpack-cli": "^5.0.1",
--- a/jsx/src/App.jsx
+++ b/jsx/src/App.jsx
@@ -5,7 +5,7 @@ import { createStore } from "redux";
 import { compose } from "recompose";
 import { initialState, reducers } from "./Store";
 import withAPI from "./util/withAPI";
-import { HashRouter, Switch, Route } from "react-router-dom";
+import { HashRouter, Routes, Route } from "react-router-dom";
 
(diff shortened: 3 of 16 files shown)


minrk:
everything seems to work, but I need to understand what needs changing in the tests. The test scaffolding is very complex, and doesn't appear to reflect real use very accurately.

I think part of it is because we have multple events causing state changes from multiple sources.

jsx/package.json
```diff
@@ -7,8 +7,8 @@
   "license": "BSD-3-Clause",
   "scripts": {
     "build": "webpack",
+    "build:watch": "webpack watch",
     "hot": "webpack && webpack-dev-server",
-    "place": "cp build/admin-react.js* ../share/jupyterhub/static/js/",
```
minrk, line 11: this helped development, because I could run `npm run build:watch` with a real, running jupyterhub and keep reloading. I found this more useful than `npm run hot`, which doesn't use real JupyterHub responses.

.gitignore
```diff
@@ -38,3 +38,5 @@ docs/source/reference/metrics.rst
 oldest-requirements.txt
 jupyterhub-proxy.pid
 examples/server-api/service-token
+
+*.hot-update*
```
minrk, line 42: this is related to the updated webpack build destination

jsx/src/App.jsx
```diff
@@ -23,34 +23,17 @@ const App = () => {
     <div className="resets">
       <Provider store={store}>
         <HashRouter>
-          <Switch>
+          <Routes>
```
minrk, line 26: update of react-router-dom to v6 (needed for `useSearchParams`) necessitates changing `<Switch>` to `<Routes>` and a few other small changes. More info in the [upgrade docs](https://reactrouter.com/en/main/upgrading/v5).

jsx/src/Store.js
```diff
@@ -1,48 +1,20 @@
...
   groups_data: undefined,
-  groups_page: { offset: 0, limit: window.api_page_limit || 100 },
+  groups_page: undefined,
   limit: window.api_page_limit || 100,
 };
 
 export const reducers = (state = initialState, action) => {
   switch (action.type) {
     // Updates the client user model data and stores the page
-    case "USER_OFFSET":
```
minrk, line 13: this is removing the pagination state from the global React 'state'

jsx/src/components/PaginationFooter/PaginationFooter.jsx
```diff
@@ -41,17 +41,33 @@ const PaginationFooter = (props) => {
             <span className="inactive-pagination">Next</span>
           </button>
         )}
+        <label>
+          Items per page:
+          <FormControl
```
minrk, line 46: The new input for the users per page

jsx/src/components/ServerDashboard/ServerDashboard.test.js
```diff
@@ -9,39 +10,44 @@ import {
...
-          startServer={spy}
-          stopServer={spy}
-          startAll={spy}
-          stopAll={spy}
-        />
-      </Switch>
-    </HashRouter>
-  </Provider>
-);
+const serverDashboardJsx = (props) => {
```
minrk, line 32: This file has a lot of changes, mostly around making this function reusable, rather than needing the whole `<Provider...` repeated many times.

jsx/src/components/ServerDashboard/ServerDashboard.test.js
```diff
@@ -350,25 +349,25 @@ test("Shows server details with button click", async () => {
 
   await act(async () => {
     fireEvent.click(button);
+    jest.runAllTimers();
   });
-  clock.tick(400);
```
minrk, line 354: clock.tick outside `act` caused problems. `runAllTimers` is a more robust, built in solution to the same problem.

jsx/src/util/paginationParams.js
```diff
@@ -0,0 +1,66 @@
+import { debounce } from "lodash";
+import { useSearchParams } from "react-router-dom";
+
+export const usePaginationParams = () => {
+  // get offset, limit, name filter from URL
```
minrk, line 5: This is the shared code for storing pagination info in URL (for any paginated page)

jsx/webpack.config.js
```diff
@@ -28,7 +28,7 @@ module.exports = {
   output: {
     publicPath: "/",
     filename: "admin-react.js",
-    path: path.resolve(__dirname, "build"),
+    path: path.resolve(__dirname, "../share/jupyterhub/static/js/"),
```
minrk, line 31: This is what removes the need for `npm run place`. Set the destination to the actual destination.

minrk (commented):
did some self-review to hopefully help folks navigate the meaningful changes

consideRatio (approved):
Wieee looks good to me! Thank you for providing self review notes @minrk, and especially thankful for the gif/video showing the functionality!

PR #4721, opened 2024-03-04
Implement sort order in GET /users

minrk:
support sorting by id (current, default), name, last_activity

Names and definitions of `sort`, `direction` match GitHub REST API.

This is the server-side part of #3816. With and #4720, we can include the sort info in the URL

### DIFF
--- a/docs/source/_static/rest-api.yml
+++ b/docs/source/_static/rest-api.yml
@@ -135,6 +135,27 @@ paths:
           schema:
             type: boolean
           allowEmptyValue: true
+        - name: sort
+          in: query
+          description: |
+            Sort users by this key (new in JupyterHub 5.0).
+            Default: id (opaque but stable internal order).
+          schema:
+            type: string
+            enum:
+              - id
+              - name
+              - last_activity
+        - name: direction
+          in: query
+          description: |
+            Direction of sorted results (new in JupyterHub 5.0).
+            Default: 'asc' (ascending).
+          schema:
+            type: string
+            enum:
+              - asc
+              - desc
       responses:
         200:
           description: The Hub's user list
--- a/docs/source/_static/rest-api.yml
+++ b/docs/source/_static/rest-api.yml
@@ -139,23 +139,26 @@ paths:
           in: query
           description: |
             Sort users by this key (new in JupyterHub 5.0).
-            Default: id (opaque but stable internal order).
+            Default: `id` (opaque but stable internal order).
+            Supported fields: `id`, `name`, `last_activity`.
+            Results will be in ascending order.
+            Descending order can be requested with a `-` prefix,
+            e.g. `sort=-last_activity` for most recently active users first.
           schema:
             type: string
-            enum:
-              - id
-              - name
-              - last_activity
-        - name: direction
-          in: query
-          description: |
-            Direction of sorted results (new in JupyterHub 5.0).
-            Default: 'asc' (ascending).
-          schema:
-            type: string
-            enum:
-              - asc
-              - desc
+          examples:
+            default:
+              value: id
+              summary: |
+                default: sort by internal id
+            by_name:
+              value: name
+              summary: |
+                sort by user.name
+            recent_activity:
+              value: "-last_activity"
+              summary: |
+                sort by last activity, most recent first
       responses:
         200:
           description: The Hub's user list
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -88,15 +88,20 @@ class UserListAPIHandler(APIHandler):
     def get(self):
         state_filter = self.get_argument("state", None)
         name_filter = self.get_argument("name_filter", None)
-        sort = self.get_argument("sort", "id")
-        sort_direction = self.get_argument("direction", "asc")
+        sort = sort_by_param = self.get_argument("sort", "id")
+        sort_direction = "asc"
+        if sort[:1] == '-':
+            sort_direction = "desc"
+            sort = sort[1:]
+
         offset, limit = self.get_api_pagination()
 
         if sort in {"id", "name", "last_activity"}:
             sort_column = getattr(orm.User, sort)
         else:
             raise web.HTTPError(
-                400, f"sort must be 'id', 'name', or 'last_activity', not '{sort}'"
+                400,
+                f"sort must be 'id', 'name', or 'last_activity', not '{sort_by_param}'",
             )
 
         # NULL is sorted inconsistently, so make it explicit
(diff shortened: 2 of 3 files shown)


minrk:
Of _course_, sort order for NULL and timestamp is undefined, and postgres sorts differently from everything else:

```
>       assert sorted_fields == expected_order
E       AssertionError: assert ['early', 'mi...ate', 'never'] == ['never', 'ea...ddle', 'late']
E         
E         At index 0 diff: 'early' != 'never'
E         
E         Full diff:
E           [
E         -     'never',
E               'early',
E               'middle',
E               'late',
E         +     'never',
E           ]
```

I think we can be explicit about this.

minrk:
aaaah, `nulls_first` fixes postgres, but produces a syntax error on mysql. Databases are not my friend.

docs/source/_static/rest-api.yml
```diff
@@ -135,6 +135,27 @@ paths:
...
+          description: |
+            Sort users by this key (new in JupyterHub 5.0).
+            Default: id (opaque but stable internal order).
+          schema:
+            type: string
+            enum:
+              - id
+              - name
+              - last_activity
+        - name: direction
```
consideRatio, line 149: I'm considering the query parameters naming, they are named very briefly.

`direction` isn't self explanatory, and `sort` was something I asked myself "is this a boolean?" about, but learn from reading more that its a `sort key`.

I'm considering options like `sort_by` instead of `sort` and `sort_order` or `sort_direction`instead of `direction` as the query parameter names. We can leave it as it is as well, but I think self-explanatory naming could be worth more than being brief here.

consideRatio (commented):
Super happy to see this being worked!!

I was earlier today wanting to be able to rely on describing to hub admins how they could "sort by running servers" and press "shut down all" in order to terminate all running servers - but realized it doesn't work reliably until we have fixed these things.

docs/source/_static/rest-api.yml
minrk: I didn't choose them to be brief, I chose this to be exactly the names and values for the same thing in the GitHub API. I don't know if consistency across familiar APIs is better than trying to be clear and verbose in our own names, though. If I were picking my own, I'd do `sort_by` and `sort_direction` (or boolean `sort_ascending`, but query parameters don't have super clear ways to specify bools). WDYT?

manics: I think following a naming convention from an existing well known API (I don't have a preference for which one) is better than coming up with our own. Otherwise it's effectively the N+1 standards problem 😁.

minrk:
Yeah, I actually have that all working locally and will open a PR once this and #4720 land. I'm struggling with the jsx tests, but it all works in manual testing.

docs/source/_static/rest-api.yml
consideRatio: I looked around from a search "sorting rest api best practices" and then went onwards to look at examples from some big companies exposing REST API's:

- [Ref 1](https://www.moesif.com/blog/technical/api-design/REST-API-Design-Filtering-Sorting-and-Pagination/#sorting), mentions `sort` and `sort_by`, and examplifies how the sort direction can provided by the value of this parameter as well
- [Ref 2](https://www.atatus.com/blog/rest-api-design-filtering-sorting-and-pagination/#sorting), mentions `sort` and `order`, and examplifies multi column sorting by `sort` bundled to handle order as well with a prepended `-` sign for descending.
- [Ref 3](https://stackoverflow.blog/2020/03/02/best-practices-for-rest-api-design/#h2-ef7a235718c20), mentions `sort` with `+-` prefixes to columns to determine order.
- [Ref 4, twitter](https://developer.twitter.com/en/docs/twitter-ads-api/sorting), `sort_by` including suffix `-asc` or `-desc`
- [Ref 5, some facebook api](https://developers.facebook.com/docs/marketing-api/insights/#sorting), `sort` including `_ascending` or `_descending` suffix.

---

There doesn't seem to be a common naming practice to adhere to. Due to this, I'm currently favor we pick something that is a bit more self explanatory, and possibly also a pattern that allows us to re-use it in multi-column sorting situations even though we won't support it. I think currently lean towards favoring a single query parameter `sort_by` with a use of prefix `+` or `-` (`+` / ascending by default), thinking that we won't need a separator to use such prefixes, while we may need a separator if we suffix `-desc` or similar.

@minrk maybe you can make a call at this point and we'll go with it?

consideRatio (approved):
This looks right! We can merge this as it is, or optionally iterate on the naming as well first.

docs/source/_static/rest-api.yml
minrk: I really wish naming things wasn't so hard, which is why I always try to follow someone else's choice instead of starting the naming discussion from scratch. [jsonapi](https://jsonapi.org/format/) uses the `-` prefix for descending and defines no prefix as ascending. The reason not to use `+` prefix is that `+` in a query string actually means a space character. I don't think we should suggest using any characters that could get confused by url escaping.

I do like the possibility of chaining sorts with commas in the future, so I'll go with the jsonapi spec for now (`sort=id` or `sort=-last_activity`, no `+`), though I won't implement `sort=-last_activity,name` multi-level sort here. It does give us a clear way to do it in the future.

I really don't think anyone could be confused by the absence of `_by`, since nobody will see the `sort` parameter without also seeing at least one example of a value, which removes all ambiguity. There definitely does seem to be consensus among sampling that `sort` is expected over `sort_by`, even if there are some exceptions, so I think `sort` is the better choice for the parameter name.

minrk:
This now implements the `jsonapi` [definition](https://jsonapi.org/format/#fetching-sorting) of the sorting parameter, which is approximately @consideRatio's suggestion (`sort` not `sort_by`, and no `+` for ascending).

consideRatio (approved):
The rest-api.yml still references `direction`, but everything else looks right!

docs/source/_static/rest-api.yml
consideRatio: Excellent find of referencing jsonapi - I'm also happy that this syntax allows for multi-sorting if we want to implement support for that.

docs/source/_static/rest-api.yml
```diff
@@ -135,6 +135,27 @@ paths:
...
+          in: query
+          description: |
+            Sort users by this key (new in JupyterHub 5.0).
+            Default: id (opaque but stable internal order).
+          schema:
+            type: string
+            enum:
+              - id
+              - name
+              - last_activity
```
consideRatio, line 148: I think this needs an update as well, so that it mentions the `-` prefix. Also, this schema won't make a request fail if its passes `-id` instead of the listed enums that only includes `id` right?

minrk: right, forgot to push. I debated that. I don't think adding `-` to enum looks or feels nice, so I removed the enum and added examples, instead. This is what it looks like after #4700:

Screenshot 2024-03-06 at 13 26 43

minrk: > Also, this schema won't make a request fail if its passes -id instead of the listed enums that only includes id right?

Right, should clarify: this schema exists solely for the purposes of generating documentation for consumption by _humans_. It has no effect on JupyterHub behavior, and is not intended or appropriate for code generation or validation purposes. We exclusively use OpenAPI as a _rendering medium_ for documentation, not as a strict specification of our API. Deviations from the spec where they produce nicer-looking or clearer docs are preferred to strict adherence via complex things like `oneOf`, etc.

consideRatio:
Wieee nice!

PR #4723, opened 2024-03-08
admin: update navigation for react-router v6

minrk:
navigation wasn't working (edit user, etc.) because it used `history` property, rather than v6's navigate/location

mostly a big find/replace, but also fixed the hierarchy of the group property editor,

which was previously computing a nested transpose table, where the table actually had only one row, but in each _cell_ of the table was a number of rows containing keys, values, etc.

```
<tr>
  <td>
  <tr>
    <td>
```

which is now fixed to be a standard one-row-per-row table, and the form input is in the table, so it matches.


consideRatio (approved):
Amazing progress @minrk! Looks right from what i can tell, and improves a now broken state - let's go!

PR #4724, opened 2024-03-11
switch to ruff for lint, format

minrk:
seems to be the popular tool these days, and it's quite a bit quicker to run (one hook for several steps, etc.)

minor differences in some formatting, and a few lint rules that caught code that could be improved. Otherwise, no change.

ruff has lots more lint rules we could consider opting in to, but this keeps our lint rules _almost_ unchanged.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -24,8 +24,6 @@ from textwrap import dedent
 from typing import Optional
 from urllib.parse import unquote, urlparse, urlunparse
 
-if sys.version_info[:2] < (3, 3):
-    raise ValueError("Python < 3.3 not supported: %s" % sys.version)
 
 import tornado.httpserver
 import tornado.options
@@ -385,9 +383,7 @@ class JupyterHub(Application):
     def _validate_config_file(self, proposal):
         if not self.generate_config and not os.path.isfile(proposal.value):
             print(
-                "ERROR: Failed to find specified config file: {}".format(
-                    proposal.value
-                ),
+                f"ERROR: Failed to find specified config file: {proposal.value}",
                 file=sys.stderr,
             )
             sys.exit(1)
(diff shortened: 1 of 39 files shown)


minrk:
need to disable the `==` -> `is` lint, which is not a safe conversion for ORM comparisons.

manics:
Sounds good to me, I've tried Ruff in one of my other projects and it's definitely a lot faster!

manics (approved):

jupyterhub/app.py
```diff
@@ -24,9 +25,6 @@
 from typing import Optional
 from urllib.parse import unquote, urlparse, urlunparse
 
-if sys.version_info[:2] < (3, 3):
-    raise ValueError("Python < 3.3 not supported: %s" % sys.version)
-
```
consideRatio, line 29: :D

consideRatio (approved):
This makes sense I think.

The configuration below is ignored by ruff, right? Let's remove it as well

```toml
# autoflake is used for autoformatting Python code
#
# ref: https://github.com/PyCQA/autoflake#readme
#
[tool.autoflake]
ignore-init-module-imports = true
remove-all-unused-imports = true
remove-duplicate-keys = true
#remove-unused-variables = true
# isort is used for autoformatting Python code
#
# ref: https://pycqa.github.io/isort/
#
[tool.isort]
profile = "black"
# black is used for autoformatting Python code
#
# ref: https://black.readthedocs.io/en/stable/
#
[tool.black]
skip-string-normalization = true
# target-version should be all supported versions, see
# https://github.com/psf/black/issues/751#issuecomment-473066811
target_version = [
    "py38",
    "py39",
    "py310",
    "py311",
    "py312",
]
```

minrk:
@consideRatio removed. I initially left it because I didn't want an editor configured to run black to automatically make a mess, but I guess that's unavoidable when there are differences. Auto-formatting in editors without detecting project configuration is just not a good idea!

PR #4725, opened 2024-03-11
Fix typo in the `rest-api.yml`

aktech:
Minor typo, just noticed here: https://jupyterhub.readthedocs.io/en/latest/reference/rest-api.html#operation/delete-shares-server

cc @minrk


minrk:
Thank you!

PR #4736, opened 2024-03-19
Bump required Python version in contributing setup to 3.8

krassowski:
Since https://github.com/jupyterhub/jupyterhub/pull/4495 the required Python version is 3.8 or greater:

https://github.com/jupyterhub/jupyterhub/blob/1db5e5e95cf5201e7b524223e8468545d35abaa1/pyproject.toml#L18

The contributing documentation still mentioned 3.6. This bumps it to 3.8


minrk:
Thanks!

PR #4738, opened 2024-03-20
run browser tests in subdomain

minrk:
make sure everything is working


minrk:
self-merging to unblock other things

PR #4741, opened 2024-03-21
Consistently use minimum Python version in docs

manics:
We had some mentions of Python 3.6 as the minimum version.

Also adds a Sphinx variable for Node.js


minrk:
nice!

Issue #3313, opened 2020-12-22
Documentation of jupyterhub.dbutil and alembic revisions


consideRatio:
As discussed during a meeting with @minrk and @ivanah8, it would be useful to have some documentation in the alembic/README.md file.

Here are ideas of what can be documented.

- How to create new migration scripts that are found in the folder
- Guidelines for creating those migration scripts: one per release rather than between each PR
- Briefly about what jupyterhub.dbutil is and when/how its useful

In general, I'm thinking a few sentences rather than an entire page of text etc.

PR #4748, opened 2024-03-23
Add authenticator-managed roles (`manage_roles`)

krassowski:
:wave: I would greatly appreciate any feedback on whether the approach taken looks like the right way forward. I'm happy to rework the PR if it needs to be changed.

- addresses https://github.com/jupyterhub/jupyterhub/issues/3858
- builds upon and closes https://github.com/jupyterhub/jupyterhub/pull/4050

Compared to #4050:
- using both `manage_roles` and `load_roles` is disallowed (tests were added)
- documentation was added
- `sync_roles` was rewritten  to always grant/strip roles to the current user and to update all role attributes for each rule
- the existing `create_role` function is used (over direct ORM manipulation) which checks for validity of the role name, and guards against overwriting the admin role

To be done:
- [x] rework tests carried over from #4050
- [x] add more tests for granting/stripping roles
- [x] maybe add a method to sync all roles from the Authenticator on startup


jupyterhub/auth.py
```diff
@@ -721,7 +721,18 @@ def delete_user(self, user):
...
+    manage_roles = Bool(
+        False,
+        config=True,
+        help="""Let authenticator manage roles
+
+        If True, Authenticator.authenticate and/or .refresh_user
+        may return a list of roles in the 'roles' field,
+        which will be added to the database.
 
+        All role-assignment APIs are disabled if this is True.
```
minrk, line 733: Probably mention `load_roles` config conflict. There aren't currently any role-assignment APIs, but this will be true. Could be a more general statement that all role management will happen through the Authenticator.

jupyterhub/roles.py
```diff
@@ -214,7 +232,7 @@ def _existing_only(func):
     """Decorator for checking if roles exist"""
 
     @wraps(func)
-    def _check_existence(db, entity, role=None, *, rolename=None):
+    def _check_existence(db, entity, role=None, commit=True, *, rolename=None):
```
minrk, line 235: can this be keyword-only (after the `*`)?

jupyterhub/user.py
```diff
@@ -301,6 +301,84 @@ def sync_groups(self, group_names):
...
+
+        granted_roles = new_user_roles.difference(current_user_roles)
+        stripped_roles = current_user_roles.difference(new_user_roles)
+
+        if granted_roles:
+            self.log.info(f"Granting user {self.name} roles(s): {granted_roles}")
+        if stripped_roles:
+            self.log.info(f"Stripping user {self.name} roles(s): {stripped_roles}")
+
+        all_roles = {r.name for r in self.db.query(orm.Role).all()}
```
minrk, line 319: This might be costly to fetch all roles that exist every time, when most won't be relevant. Instead, you can do a filtered query for:

```python
self.db.query(orm.Role).filter(orm.Role.name.in_(granted_roles))
```

to retrieve only the roles that are to be granted and already exist (or a similar query if you're going to look at different collections)

minrk (commented):
This is great!

I do think we want a mechanism for Authenticators to initialize and check existing roles at startup.

Goals:

- populate initial roles
- check/validate/update/remove past roles that may be outdated

Considerations:

- credentials may not be available to check a user's role membership, so we can't do the simple approach of running through `sync_roles`
- Calling `sync_roles` on every user at startup will get prohibitively costly, we shouldn't do that.
- I'd like to keep the Authenticator from having access to the database, so this should likely be done through role descriptions dicts as seen in `load_roles`, not orm.Role objects, queries, etc.


The first and simplest option is an `Authenticator.load_managed_roles()` _method_ that is treated very similar to (perhaps exactly the same as) the `load_roles` config.

I think perhaps we should still support `load_roles` config for services and overriding the scopes of the default roles. Maybe it's only user/group role assignment that's prohibited when Authenticator.manage_roles is defined, rather than `load_roles` as a whole.

For the check/refresh role, what about calling an `Authenticator.refresh_managed_role(role_dict)` for each role. This is likely to be a much smaller number than the number of users.

for role in db.query(orm.Role):
    role_dict = role_to_dict(role)
    initial_role_dict = copy.deepcopy(role_dict)
    updated_role_dict = await authenticator.refresh_managed_role(role_dict)
    if updated_role_dict is None:
        db.delete(role)
    else:
        handle_differences(updated_role_dict, initial_role_dict)
```

But if we _do_ still support roles from config, we'll need to only do this for roles actually managed by the Authenticator.

File: jupyterhub/user.py (hunk 3 above)
Comment (krassowski) [reply jupyterhub/user.py]: Done in 26906cca07c18e2e20b5b6c6d7ee134876ad2832.

File: jupyterhub/roles.py (hunk 2 above)
Comment (krassowski) [reply jupyterhub/roles.py]: Done in 0d6778f95584f3e5fc200125bf185c2809a60358.

File: jupyterhub/auth.py (hunk 1 above)
Comment (krassowski) [reply jupyterhub/auth.py]: Addressed in baf6e03c469afce48d4b8195bb89894ad2950bb9 but may need further tuning as we go.

**Comment by krassowski** [comment]
> The first and simplest option is an `Authenticator.load_managed_roles()` _method_ that is treated very similar to (perhaps exactly the same as) the `load_roles` config.

Done in https://github.com/jupyterhub/jupyterhub/pull/4748/commits/b7d68ca255717854a1d0a130fd4c53f8f32e21eb

> I think perhaps we should still support `load_roles` config for services and overriding the scopes of the default roles. Maybe it's only user/group role assignment that's prohibited when Authenticator.manage_roles is defined, rather than `load_roles` as a whole.

Done in https://github.com/jupyterhub/jupyterhub/pull/4748/commits/b7d68ca255717854a1d0a130fd4c53f8f32e21eb

> For the check/refresh role, what about calling an `Authenticator.refresh_managed_role(role_dict)` for each role. This is likely to be a much smaller number than the number of users.

Apologies, I do not follow this bit. When would `Authenticator.refresh_managed_role` be called?

**Comment by krassowski** [comment]
Two additional questions:
- should the default roles be stripped if not included in the list of roles returned by authenticator? I think that this should be symmetrical with whether the default roles can be granted by being included in the authenticator response, hence I think that returning an empty list `rules: []` should result in stripping of all roles, including `user` or `admin`.
- regarding _but if we do still support roles from config, we'll need to only do this for roles actually managed by the `Authenticator`_, does it imply that we should persist in the database if a role was created from `Authenticator` by adding a new column to the table?

**Comment by krassowski** [comment]
> > For the check/refresh role, what about calling an `Authenticator.refresh_managed_role(role_dict)` for each role. This is likely to be a much smaller number than the number of users.
>
> Apologies, I do not follow this bit. When would `Authenticator.refresh_managed_role` be called?

Expanding on my "I do not understand":
- there is `User.refresh_user()` method, but if we went calling `[auth.refresh_managed_role(role) for role in roles]`, we would have a number of updates up to (number of users) * (number of roles)
- I do not see another existing method that would be called to sync more general state from the `Authenticator`
- this PR is syncing state from `Authenticator` is in `User.sync_roles()` on user-by-user basis as needed (as they login or as their sessions need refreshing), which is consistent with the existing behaviour for groups (it also loads roles using `Authenticator.load_managed_roles` at startup but this appears irrelevant)
- while I do see a high value in having a REST API to call from my Authenticator module which says "roles/groups changed, please sync everything", I am not sure if that fits into what @minrk was thinking of and more generally into the scope of this PR.

I am probably missing something obvious here :thinking:

**Comment by minrk** [comment]
> should the default roles be stripped if not included in the list of roles returned by authenticator?

In other contexts we distinguish `None` (no change, e.g. insufficient info) from an empty list (strip all). I think the same thing makes sense here. An empty list should generally strip all roles. I'm a little on the fence about stripping `users`, but I'm leaning toward yes. I do think it should strip `admin`. The question for `users` is: does it make sense for a user to successfully login, but not have the `users` role? Maybe it doesn't.

> does it imply that we should persist in the database if a role was created from Authenticator by adding a new column to the table?

That is an option. That's what we do for services created via the REST API, for example. The main problem to solve is: when the hub starts up, how does it ensure that the role assignments are correct and stale assignments are removed? For example:

1. configure jupyterhub with `load_roles` config
2. start jupyterhub
3. _change_ config to `Authenticator.manage_roles = True`, remove `load_roles` config
4. restart jupyterhub 

what should happen to the roles and role assignments present in the database because they _were_ defined by configuration that is no longer valid? Presumably, they should all be removed. But we also need to be able to distinguish this from a role defined by a user's past login via managed_roles.

A role in the database can have these states at startup:

- define by load_roles config and in sync with config (no action)
- define by load_roles config and out of sync with config (needs update or removal)
- defined by manage_roles and up-to-date (no action)
- defined by manage_roles and out-of-date (needs update or removal)

The important thing is that we can tell the difference between these, or at least that behavior is well defined for each (e.g. with manage_groups, we don't handle the last case, but it is at least well-defined that group membership lasts until the user logs in again or refresh_user is called).

> When would Authenticator.refresh_managed_role be called?

I was thinking this would be called for each _role_ at Hub start, not any other time. This would essentially be the Authenticator-managed version of the config<->database reconciliation step we currently have. It gives the Authenticator a chance to refresh role definitions and assignments:

```python
def init_role_creation(self):
    if self.authenticator.manage_roles:
        for role in self.db.query(orm.Role):
            # filter to only authenticator-managed?
            role_dict.role_to_dict(role)
            new_role_dict = await self.authenticator.refresh_role(new_role_dict.copy())
            if new_role_dict is None:
                delete_role
            elif new_role_dict != role_dict:
                refresh_role
            else:
                no_change
``` 

It may not need to be implemented as this method, but I want to make sure we have a way for roles defined by manage_roles to be able to _remove_ stale roles at startup. We don't have this for groups, and I don't particularly think we need it for groups (it might be nice, but isn't a big deal), but I think it's more important for permission-granting things like roles.

minrk:
Another, perhaps simpler, option is to have a _boolean_ config that says whether to remove all roles not returned by `load_managed_roles` (i.e. indicating whether `load_managed_roles` returns a comprehensive list). `refresh_managed_role` would have the same capability but more flexibility. It would be more expensive if it should just remove them all.

krassowski:
Thank you for the explanations and patience, I believe I now understand the problem.

> Another, perhaps simpler, option is to have a _boolean_ config that says whether to remove all roles not returned by `load_managed_roles` (i.e. indicating whether `load_managed_roles` returns a comprehensive list).

I think that it would be possibly easier to understand and less surprising for users (developers) if the `load_managed_roles` result was always defined to be comprehensive. Is there a valid scenario in which you would want to have the proposed _boolean_ flag set to "do not delete stale managed roles"?

While I now understand `refresh_managed_role` idea I am not convinced. If I am an author of an `Authenticator` and set `manage_roles = True` I would need to:
- populate the `roles` attribute in auth data, which includes the role definition in case it was newly created
- implement `refresh_managed_role` to remove stale roles or update existing ones
- implement `load_managed_roles` to pre-populate the JupyterHub with new roles that I know I will want to use _and_ update existing ones

It strikes me that `refresh_managed_role` is one more thing that the author of `Authenticator` would need to worry about, compared to declaring `load_managed_roles` result as always comprehensive, in which case it would be:
- populate the `roles` attribute in auth data, which includes the role definition in case it was newly created
- implement `load_managed_roles` to pre-populate the JupyterHub with new roles, update existing roles, and implicitly mark any non-included managed roles for removal

In the latter approach we also eliminated a duplication in responsibility as  in the former both `refresh_managed_role` and `load_managed_roles` were responsible for updating the existing roles.

The former approach would be perhaps more compelling if we offloaded more responsibility to `refresh_managed_role`, this is the `Authenticator` author would: 
- populate the `roles` attribute in auth data _with role names only_
- implement `refresh_managed_role` to remove stale roles _and fetch definitions of roles as passed in `roles` in auth data_
- implement `load_managed_roles` to pre-populate the JupyterHub with new roles that I know I will want to use _and_ update existing ones

I am not super enthusiastic about this idea either because it makes the `Authenticator` behaviour more complex and would require `refresh_managed_role` to potentially callback to the service to get the role details extending the login sequence.

>  It [`refresh_managed_role`] would be more expensive [than boolean] if it should just remove them all.

I tried to understand this one because I think there are assumptions hidden here. With my own set of assumptions (/biases) I think I would not expect a significant performance difference.

Core assumption: most often the Hub would restart with the roles being the same as before and no changes will be required:
- a) a naive implementation of the boolean config would just remove all managed roles before loading the roles returned by `load_managed_roles`. It would be fast to remove the old roles (a single "delete" query) but would still have a cost in re-creating all the roles, so there would be a constant cost proportional to the number of roles if nothing changed.
- b) if we instead compare the old and new managed roles role-by-role, the cost for "nothing changed" scenario would be equivalent to a single "query all roles including role assignments" and then some iterations in Python. Comparing role-by-role is inherent in `refresh_managed_role` option too, and that option is only slightly more expensive than "boolean with role-by-role comparison" due to additional Python calls.

I suspect that querying roles (and their assignments) is faster than creating them, so role-by-role comparison as in (b) should be faster than (a). Hence, there would be no significant difference between "boolean flag" and `refresh_managed_role` options in this scenario.

minrk:
> Is there a valid scenario in which you would want to have the proposed boolean flag set to "do not delete stale managed roles"?

Yes, when the roles are not or cannot be known at Hub start time. I believe there are real cases where roles are created during the lifetime (e.g. creating special per-user or per-class roles that are only known once users login and should not require hub restart to create). I actually imagine this to be most of the time - I would expect load_managed_roles to _usually_ be empty and not be implemented by most Authenticators, and rely on authentication-time to populate the roles.

> While I now understand refresh_managed_role idea I am not convinced. 

I'm not either! I don't know what the best solution is, but I do think we need _a_ mechanism by which roles created by the Authenticator (or outdated config) _can_ be deleted. This was my first idea of a 'punt to the Authenticator a hook that can do anything' so we can at least know that the Authenticator _can_ do whatever it needs to, even though it may be cumbersome. I'm open to any kind of suggestions, but the absence of a role-management API means that the APIs we create here are the _only_ ones for creating _and deleting_ roles and role assignments, so a deployment/Authenticator must be sufficient to do both _somehow_. Any version of _somehow_ is on the table right now.

> - implement refresh_managed_role
> - implement load_managed_roles

I imagine these will usually be empty. Essentially, I expect an Authenticator would _either_ implement `load_managed_roles` _or_ populate roles during authenticate. There's likely little reason to do both unless you need to pre-populate a specific set for management operations to work.

So I imagine that most Authenticator implementations would use the default `load_managed_roles` (load no roles), and the default `refresh_managed_role` (`lambda x: x`  - do nothing, leaves stale roles untouched). But this at least gives Authenticators the _opportunity_ to purge stale roles. Only Authenticators and/or deployments where stale roles are _likely_ need to implement it.

So we have some choices:

- load_managed_roles is the _only_ source of role definitions and authenticate can only do role _assignments_ (simple for JupyterHub, but I think is insufficient for many managed_roles use cases) - this way role removal is the same as we currently have with load_roles config
- `load_managed_roles` is comprehensive _if defined_, and can return None to mean no change (i.e. `load_managed_roles` is the only way to delete roles, but if undefined, stale roles are left untouched)
- authenticate can define roles, and
  - assignments can only be updated at successful login and no mechanism to remove roles themselves short of direct db operations (our current situation with managed groups, actually)
  - provide some hook to the Authenticator to refresh or remove them at startup
  - some other mechanism for role assignments to _expire_ and require a fresh login

Thanks for talking this through with me!

krassowski:
@minrk what do you think about 633aa69623c7d4d6a092c5f56ca30bb00f99ff38?

jupyterhub/orm.py
```diff
@@ -163,28 +163,32 @@ def __repr__(self):
 # lots of things have roles
 # mapping tables are the same for all of them
 
-_role_map_tables = []
+role_associations = {}
```
minrk, line 166: this should remain private

jupyterhub/orm.py
```diff
@@ -163,28 +163,32 @@ def __repr__(self):
...
+            f'{entity}_id',
+            ForeignKey(f'{entity}s.id', ondelete='CASCADE'),
             primary_key=True,
         ),
         Column(
             'role_id',
             ForeignKey('roles.id', ondelete='CASCADE'),
             primary_key=True,
         ),
+        Column('managed_by_auth', Boolean, default=False),
```
minrk, line 186: I don't believe the role assignments need this column, only the roles themselves

jupyterhub/app.py
```diff
@@ -2240,6 +2240,37 @@ async def init_role_creation(self):
...
+                )
+
+        roles_to_load = self.load_roles
+
+        if self.authenticator.manage_roles and self.load_roles:
+            for role_spec in roles_to_load:
+                user_role_assignments = role_spec.get('users', [])
+                group_role_assignments = role_spec.get('groups', [])
+                if user_role_assignments or group_role_assignments:
+                    raise ValueError(
```
minrk, line 2263: Make sure to show the offending role and assignment(s) so users know what config to look for

jupyterhub/app.py
```diff
@@ -2303,6 +2334,20 @@ async def init_role_assignment(self):
...
+                entity_name = kind[:-1]
+                association_class = orm.role_associations[entity_name]
+                deleted = (
+                    db.query(association_class)
+                    .filter(association_class.managed_by_auth == True)
+                    .delete()
+                )
+                if deleted:
+                    self.log.info(
+                        f"Deleted {deleted} stale {entity_name} role assignments previously added by an authenticator"
```
minrk, line 2349: This is a _highly_ disruptive operation - all users lose all permissions and are forced to login again to regain their access.

If this is the way we go, this flag should have a lot of caveats in its help that it should only be True only once in a while to reset permissions and usually switch back to False. Either that, or it should only be combined with a comprehensive `load_managed_roles`, and not be used otherwise.

minrk (commented):
Thanks for implementing this! I'm not sure deleting everything on a boolean is gong to be the best approach. I think it is going to get quite expensive, because if there are a lot of users, all managed assignments are deleted and (hopefully) recreated. The number of role assignments is going to be on the order of the number of users (can be several thousand), while the number of _roles_ is likely to be much lower, and in the vast majority of cases, there should be no change to make. This makes the most common case (restart with everything up-to-date) the most expensive, rather than the least expensive.

So far, I think this hard reset is probably more complex and less flexible for Authenticators than `refresh_managed_role`, because it means the only way Authenticators can reset roles is to _also_ always have a completely comprehensive `load_managed_roles`, which I don't think is likely to be true.

On the other hand, I am intrigued by the possibility of tracking the `managed` state of role _assignments_. I hadn't thought of that. It is useful, because it means deleting 'unmanaged' assignments is easy - delete them all! But that's only really useful if there can be unmanaged assignments to managed roles or vice versa - if all assignments to a managed role are via manage_roles and all assignments to an unmanaged role are when manage_roles=False, then the flag on the role is always equivalent to the flag on each assignment, and we wouldn't have to do any assignment-level queries or operations. I am convinced by this that the `auth_managed` column is a good idea on the Roles themselves, either way.

What do you think? Is there something we get from tracking the managed state of role _assignments_ that we don't get from having it just on the roles themselves?

jupyterhub/app.py
krassowski: Right, especially because earlier we defined that `load_managed_roles` does not do role assignments, only role definitions. I will think about it more.

minrk: > Right, especially because earlier we defined that load_managed_roles does not do role assignments, only role definitions. I will think about it more.

If it makes things simpler, I think it's okay to allow load_managed_roles to do assignments.

krassowski:
Thanks for getting back to me quickly!

> Is there something we get from tracking the managed state of role assignments that we don't get from having it just on the roles themselves?

I thought - but I may be wrong - that if a user is granted an existing role, say `admin` through the authenticator in the login sequence, then this allows us to revoke that role assignment, even though the role itself will not be deleted (because default roles and roles from `load_roles` are never marked as managed).

jupyterhub/app.py
krassowski: Done in 4ec2fe3c191e3d1bd2e6ef915198c2710502801c.

jupyterhub/orm.py
krassowski: Done in 5aa8d299134902ee2f19f5d40cf822589cc5d44a.

jupyterhub/user.py
```diff
@@ -301,6 +301,89 @@ def sync_groups(self, group_names):
...
+            ) as e:
+                raise web.HTTPError(409, str(e))
+
+            # Update the groups, services and users for the role
+            groups = []
+            if 'groups' in role.keys():
+                for group_name in role['groups']:
+                    group = orm.Group.find(self.db, group_name)
+                    if group is not None:
+                        groups.append(group)
```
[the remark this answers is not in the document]
krassowski: Done in 66c65c93dbdf6accfba9db9f3d951c721110957a.

jupyterhub/user.py
```diff
@@ -301,6 +301,89 @@ def sync_groups(self, group_names):
...
+        for role_name in granted_roles:
+            roles.grant_role(
+                self.db, entity=self.orm_user, rolename=role_name, commit=False
+            )
+
+        # strip the user of roles no longer directly granted
+        for role_name in stripped_roles:
+            roles.strip_role(
+                self.db, entity=self.orm_user, rolename=role_name, commit=False
+            )
```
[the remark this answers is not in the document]
krassowski: Maybe. I think this requires checking if the given role is used by another group, service or user, and if it is managed. Done in 9d0e8e08619d273f5ecf785dae545755315825a7 (but if this does not look good we can revert it).

jupyterhub/app.py
krassowski: This was addressed. I think the code is ready for another round of review.

> Either that, or it should only be combined with a comprehensive `load_managed_roles`, and not be used otherwise.

I now lean towards having `load_managed_roles` as always comprehensive.

> If it makes things simpler, I think it's okay to allow load_managed_roles to do assignments.

I was wrong, there was no such restriction.

krassowski:
Should I add an `alembic` migration? I guess this is not documented, but https://github.com/jupyterhub/jupyterhub/issues/3313 seems to suggest this should be done on per-release basis rather than on per-PR? But then without it I think the tests will keep failing on CI?

minrk:
> Should I add an alembic migration? 

Yes, we'll need one. You can either add the `managed_by_auth` columns it to the 3c2384c5aae1 revision or make a new revision, starting with:

```
python3 -m jupyterhub.dbutil alembic revision --autogenerate -m 'manage roles'
```

I tend to generate revisions and then merge the autogenerated steps into the previous revision if we haven't released yet, deleting the generated revision.

jupyterhub/app.py
```diff
@@ -2371,6 +2442,26 @@ async def init_role_assignment(self):
...
+                        entity_name = kind[:-1]
+                        association_class = orm._role_associations[entity_name]
+                        kind_id = getattr(association_class, f'{entity_name}_id')
+                        associations = (
+                            db.query(association_class)
+                            .filter(
+                                kind_id.in_([bearer.id for bearer in orm_role_bearers])
+                                & (association_class.role_id == role.id)
+                            )
+                            .all()
```
minrk, line 2458: I htink you can drop this `.all()`, since the `for` will iterate over the rows. Right?

jupyterhub/auth.py
```diff
@@ -721,7 +733,30 @@ def delete_user(self, user):
...
+        `JupyterHub.load_roles` traitlet will not be possible.
+        """,
+    )
+    reset_managed_roles_on_startup = Bool(
+        True,
+        config=True,
+        help="""Reset managed roles to result of `load_managed_roles()` on startup.
 
+        If True:
+        - stale managed roles will be removed,
```
minrk, line 756: Let's be precise about the definition of `stale`, e.g.

> any role assignment not present in `load_managed_roles()` will be considered 'stale'

I think this also warrants some big caveats, because if an Authenticator doesn't implement this method and it calls the base class implementation, this will revoke all permissions from all users on every jupyterhub start. In fact, I wonder if this means that the default `load_managed_roles` should raise `NotImplementedError()` instead of `reset_managed_roles` because the result is so destructive.

minrk (commented):
This is looking great!

I think the main remaining thing to think about with this implementation is for Authenticators where comprehensive `load_managed_roles` is not feasible. For example, it is common for deriving something from e.g. oauth team membership to require a user's auth credentials, so it's not going to be possible in general to pre-populate user lists for these roles. Does that mean that `reset_managed_roles` should be entirely unavailable to most oauth-based Authenticators?

Let's make sure to include tests where leaving `users` undefined in `load_managed_roles` does not revoke any role assignments. I believe this is already the defined behavior of `load_roles`, but let's make sure.

Let's also make sure to have a test where `load_managed_roles` includes user assignments for usernames that aren't in the database, to make sure it doesn't raise.

Since this is making it explicit that if you want to use `reset_managed_roles_on_startup`, your Authenticator _must_ be able to implement a `load_managed_roles` that is always comprehensive and returns all _possible_ roles. Are there situations where that's not likely to be true? Is there a workaround for an Authenticator to revoke _some_ roles, but leave some others that it can't precompute alone? Or is the answer that these situations can't use `reset_managed_roles` and `load_managed_roles` ought to raise?

jupyterhub/app.py
krassowski: Done in 9dc91fb707a846cef9106b3ce6dfeed08d650246.

krassowski:
> Let's make sure to include tests where leaving `users` undefined in `load_managed_roles` does not revoke any role assignments.

Done in 0f1acce363ae475eb4fdf1ff6e73ddf17290db68.

> Let's also make sure to have a test where `load_managed_roles` includes user assignments for usernames that aren't in the database, to make sure it doesn't raise.

Done in 6b8dd277a25982e172e68fc543891dfb1c7c8303.

> Or is the answer that these situations can't use reset_managed_roles and load_managed_roles ought to raise?

I implemented this in 1a04ecde8e5d7c65a9c158bcfc1e6560294c56a2.

jupyterhub/auth.py
krassowski: This should be resolved by 0f1acce363ae475eb4fdf1ff6e73ddf17290db68 and  1a04ecde8e5d7c65a9c158bcfc1e6560294c56a2.

krassowski:
Any suggestions on how to fix the `jupyterhub/tests/test_db.py::test_upgrade` failures? I do not understand why these are failing.

jupyterhub/alembic/versions/4621fec11365_manage_roles.py
```diff
@@ -0,0 +1,40 @@
...
+branch_labels = None
+depends_on = None
+
+import sqlalchemy as sa
+from alembic import op
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
```
minrk, line 21: in case of a long upgrade, these tables may not have been created yet. You can check for their existence before altering them, as done [here](https://github.com/jupyterhub/jupyterhub/blob/f9fb650a7b1cfbc5306cb674ec3038f1801f4e17/jupyterhub/alembic/versions/56cc5a70207e_token_tracking.py#L33):

```python
    engine = op.get_bind().engine
    tables = sa.inspect(engine).get_table_names()
    if table_name in tables:
        op.add_column(table_name, ...)
```

If the tables don't exist yet, there's no upgrade to do because they will be created on first start.

docs/source/reference/authenticators.md
```diff
@@ -302,7 +302,51 @@ which is a list of group names the user should be a member of:
...
+all role-management via the API is disabled,
+and roles cannot be assigned to groups nor users via `load_roles` traitlet
+(roles can still be created via `load_roles` or assigned to services).
+
+When an authenticator manages roles, the initial roles and role assignments
+can be loaded from role specifications returned by the {meth}`.Authenticator.load_managed_roles()` method.
+
+The authenticator-manged roles and role assignment will be deleted after restart if:
+
+- {attr}`.Authenticator.reset_managed_roles_on_startup` is set to `True` (default), and
```
minrk, line 348: Since this is True by default, but the default `load_managed_roles` raises when this is True, the _default_ behavior of `manage_roles = True` is to fail to start. Maybe the default on the base class should be False?

jupyterhub/orm.py
```diff
@@ -163,28 +163,32 @@ def __repr__(self):
...
```
minrk, line 186: I think this should have `nullable=False` (also in the alembic upgrade, I think). Can it also be checked that after the upgrade, the column is filled with False instead of Null? That could probably be tested in the db test_upgrade.

I can work on that, if you're not comfortable with it.

jupyterhub/orm.py
```diff
@@ -195,12 +199,15 @@ class Role(Base):
...
     description = Column(Unicode(1023))
     scopes = Column(JSONList, default=[])
+
     users = relationship('User', secondary='user_role_map', back_populates='roles')
     services = relationship(
         'Service', secondary='service_role_map', back_populates='roles'
     )
     groups = relationship('Group', secondary='group_role_map', back_populates='roles')
 
+    managed_by_auth = Column(Boolean, default=False)
```
minrk, line 209: also `nullable=False`

minrk (commented):
This is really great! Just two more small points, then I think this is ready to go.

docs/source/reference/authenticators.md
krassowski: Makes sense, done in https://github.com/jupyterhub/jupyterhub/pull/4748/commits/d5e9e3a6323fb5026beb90c47714ff8ee02e9eb8.

jupyterhub/orm.py
krassowski: This indeed sounds simple on the surface, but the migration scripts start failing if I toggle nullable.

> I can work on that, if you're not comfortable with it.

I would appreciate that, thank you!

minrk: No problem at all! I'll have a go.

minrk:
Excellent, thank you!

PR #4750, opened 2024-03-25
rework handling of multiple xsrf tokens

### LINKED ISSUE #4749: Named servers broken in 4.1.0
### Bug description

Spawning multiple servers (containers) results in 'too many redirects' error.

### How to reproduce

A JupyterHub + multiple JupyterLab images with JupyterHub 4.1.1 installed.

#### Expected behaviour

No issue running multiple servers (containers) in parallel.

#### Actual behaviour

Logs of a JupyterLab container: Repeated

```
[I 2024-03-24 13:57:09.790 ServerApp] 302 GET /user/benz0li/qgis-latest/lab -> /user/benz0li/qgis-latest/lab (benz0li@[redacted]) 6.93ms
[W 2024-03-24 13:57:09.821 ServerApp] Attempting to clear invalid _xsrf cookie 'Mnwx...'
```

👉 JupyterLab (`/lab`) not loading.

### Your personal set up

JupyterHub: [`glcr.b-data.ch/jupyterhub/jupyterhub:4.1.1-oauth-docker`](https://gitlab.b-data.ch/jupyterhub/jupyterhub/container_registry)

JupyterLab:

* [JupyterLab QGIS docker stack](https://github.com/b-data/jupyterlab-qgis-docker-stack)
* [JupyterLab Python docker stack](https://github.com/b-data/jupyterlab-python-docker-stack)
* [JupyterLab Julia docker stack](https://github.com/b-data/jupyterlab-julia-docker-stack)
* [JupyterLab R docker stack](https://github.com/b-data/jupyterlab-r-docker-stack)

ℹ️ base-images built with `JUPYTERHUB_VERSION=4.1.1`.

Ping @minrk

minrk:
rather than attempting to clear multiple tokens (too complicated, breaks named servers) look for and accept first valid token.

closes #4749 

To do this, we have to do our own cookie parsing because existing cookie implementations only return a single value for each key and default to selecting the _least_ likely to be correct, according to RFCs. This lets us safely ignore xsrf tokens that aren't for us on other prefixes that may be shared with our prefix (e.g. named servers are on a sub-path of the default server - a URL design mistake, I realize!)

Also sets updated xsrf cookie on login to avoid needing two requests to get the right cookie, which can result in a weird sequence of log messages about changing xsrf tokens.

I'll make a 4.1.2 with this.

### DIFF
--- a/jupyterhub/_xsrf_utils.py
+++ b/jupyterhub/_xsrf_utils.py
@@ -104,13 +104,20 @@ def _get_xsrf_token_cookie(handler):
     return (None, None)
 
 
-def _set_xsrf_cookie(handler, xsrf_id, cookie_path=""):
+def _set_xsrf_cookie(handler, xsrf_id, *, cookie_path="", authenticated=None):
     """Set xsrf token cookie"""
     xsrf_token = _create_signed_value_urlsafe(handler, "_xsrf", xsrf_id)
     xsrf_cookie_kwargs = {}
     xsrf_cookie_kwargs.update(handler.settings.get('xsrf_cookie_kwargs', {}))
     xsrf_cookie_kwargs.setdefault("path", cookie_path)
-    if not handler.current_user:
+    if authenticated is None:
+        try:
+            current_user = handler.current_user
+        except Exception:
+            authenticated = False
+        else:
+            authenticated = bool(current_user)
+    if not authenticated:
         # limit anonymous xsrf cookies to one hour
         xsrf_cookie_kwargs.pop("expires", None)
         xsrf_cookie_kwargs.pop("expires_days", None)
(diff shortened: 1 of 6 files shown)


consideRatio (approved):
I think this look good, but its a very complicated topic so I'm not very confident in my ability to catch technical issues here.

I'm thankful for many inline comments and refactoring to encapsulate logic in functions, even if its only used once - because it helps sections of code read more easily.

Amazing effort into this security work @minrk!!

consideRatio:
I'll leave you to do a self-merge this and cut 4.1.2 if you have time - ping me for anything you'd like help with

minrk:
Thanks @consideRatio !

PR #4759, opened 2024-03-29
avoid xsrf check on navigate GET requests

minrk:
sevices.auth prevents calling check_xsrf_cookie for navigate GET requests (regular page views), but if the Handler itself still calls it, the newly strict check would still be applied.

this ensures regular page views are allowed, even if the XSRF check is applied, such as Jupyter Server's nbconvert handlers.

Fixes https://github.com/jupyterlab/jupyterlab/issues/16040

### DIFF
--- a/jupyterhub/_xsrf_utils.py
+++ b/jupyterhub/_xsrf_utils.py
@@ -178,10 +178,35 @@ def get_xsrf_token(handler, cookie_path=""):
     return xsrf_token
 
 
+def _needs_check_xsrf(handler):
+    """Does the given cookie-authenticated request need to check xsrf?"""
+
+    if getattr(handler, "_token_authenticated", False):
+        return False
+
+    fetch_mode = handler.request.headers.get("Sec-Fetch-Mode", "unspecified")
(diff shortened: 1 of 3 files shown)


consideRatio (approved):

consideRatio:
Looks right, letting you merge to also keep track of forwardporting etc

jupyterhub/_xsrf_utils.py
```diff
@@ -178,10 +178,35 @@ def get_xsrf_token(handler, cookie_path=""):
...
+        return True
+
+
 def check_xsrf_cookie(handler):
     """Check that xsrf cookie matches xsrf token in request"""
     # overrides tornado's implementation
     # because we changed what a correct value should be in xsrf_token
+    if not _needs_check_xsrf(handler):
+        # don't require XSRF for regular page views
+        return
```
consideRatio, line 209: For reference, only this is the actual change in this PR

consideRatio:
This also fixes https://github.com/jupyterhub/jupyter-remote-desktop-proxy/issues/106 - thank you @minrk for quick followups!!

PR #4771, opened 2024-04-04
singleuser mixin: include check_xsrf_cookie in overrides

minrk:
missing this could cause a hang when using the old mixin implementation of singleuser auth (e.g. Jupyter Server 1.x). https://github.com/jupyterhub/the-littlest-jupyterhub/pull/975


minrk:
investigating test failure

consideRatio (approved):

consideRatio:
Thank you @minrk, beautiful fix :)

PR #4774, opened 2024-04-04
update bootstrap to v5

### LINKED ISSUE #4437: Upgrade to Bootstrap 5
### Proposed change

Currently, we are on Bootstrap 3, which is almost 10 years old. We should upgrade to Bootstrap 5!

This issue is focused on bootstrap 5 for our non-admin panel, as the admin panel is using react and already on bootstrap 4.

minrk:
finally close enough on this that I can open a draft PR

This will be a breaking change for _some_ users of custom templates, since bootstrap changes some of their layout.

closes #4437

Pages to check (I've looked over several of these already, but not enough to consider them checked off)

- [x] admin
- [x] accept-share
- [x] error
- [x] authorize
- [x] home
- [x] home with named servers
- [x] login
- [x] login with service (oauth)
- [x] javascript API error
- [x] not running
- [x] spawn
- [x] spawn with options_form
- [x] spawn_pending
- [x] token

You can review the diffs in screenshots before/after this PR [here](https://github.com/minrk/junk/compare/bs5-before...bs5-after)


minrk:
This does move the CSS build from LESS to SCSS. I suspect we could get away with just CSS variables and remove the CSS build step altogether, but that would be a bigger change.

We'd lose the few mixins and utilities we use (mainly button-variant), but _very_ close to all of our style code is straight CSS. CSS variables would also move us toward making JupyterHub itself themeable.

manics:
I think using (or not using) SCSS are both fine. The big jump in complexity is moving from a Python build system to Python + NPM which was done ages ago, adding or changing another NPM related tool doesn't seem like a problem.

minrk:
Yeah, the public entrypoint of "setup.py invokes an npm script and out comes style.min.css" is unchanged.

I think someday my resistance to webpack bundling of the main pages will eventually give way, which will also make this mostly moot.

minrk:
OK, I think this is ready for review.

The diff is a lot larger than I was hoping, especially in the admin jsx. The main reason is a change in the hierarchy of elements reflowing lots of things that didn't really change. I also increased the use of react-bootstrap (i.e. `<Button variant="primary">` instead of `<div class="btn btn-primary">`), and re-used some utilities to reduce the find-replacing, which _hopefully_ yields slightly more readable and maintainable code.

The biggest changes in general are:

- less is now scss (little practical difference, but all the files are renamed)
- 'panel' is gone, in favor of 'card', which changes some HTML hierarchy

I ended up writing a [script](https://github.com/minrk/junk/blob/bs5/screenshots/collect_screenshots.py) to collect screenshots of various JupyterHub pages so that I could see what I was changing. Maybe I'll pull this into a repo utility, but I didn't want to add that to an already big PR.

You can review the diffs in screenshots before/after this PR [here](https://github.com/minrk/junk/compare/bs5-before...bs5-after)

docs/source/contributing/setup.md
```diff
@@ -123,6 +123,14 @@ configuration:
...
 ```
 
+The test configuration enables a few things to make testing easier:
+
+- use 'dummy' authentication and 'simple' spawner
+- named servers are enabled
+- listen only on localhost
+- 'admin' is an admin user, if you want to test the admin page
+- disable caching of static files
+
```
consideRatio, line 133: I was about to open an issue about this before but saw this in this PR so didn't :)

share/jupyterhub/templates/page.html
```diff
@@ -51,15 +50,15 @@ <h1 class="modal-title" id="{{key}}-label">{{title}}</h1>
...
-              deps: ["jquery"],
-              exports: "bootstrap"
-            },
-          }
+          // shim: {
+          //   bootstrap: {
+          //     deps: ["jquery"],
+          //     exports: "bootstrap"
+          //   },
+          // }
```
consideRatio, line 61: Something to cleanup or add additional comment about in order to clarify why its commented out

share/jupyterhub/templates/oauth.html
```diff
@@ -4,13 +4,14 @@
...
 
 {% block main %}
-<div class="container col-md-6 col-md-offset-3">
+<div class="container">
+  <div class="row justify-content-center">
+    <div class="col-md-8">
     <h1 class="text-center">Authorize access</h1>
-
-    <h2>
+
```
consideRatio, line 11:
```suggestion

```

consideRatio (approved):
The visual comparison looks great, but I've only quickly skimmed the code changes - let me know if you want me to find time to do so, but otherwise I think :+1: on merge as it is.

consideRatio (approved):
Tests are passing as well - wiee! :D

Btw it could be good to lift the screenshot link to the PR description if its going to be available long term.

Merge?

minrk:
Added link to the top. I think I may add the screenshot script somewhere in this repo since it's useful for automating review of visual PRs that are hard to test, but then I'd have to think about an API, so I didn't want to add it here.

Thanks for reviewing! It will be easier to iterate later with smaller PRs, I think.

PR #4782, opened 2024-04-10
test: avoid producing '[group]' string in output

minrk:
GitHub Actions starts a log expansion group when it sees the string `[group]`, which happens when that is an argument to `pytest.parametrize`, which results in collapsing all subsequent test outputs:

Screenshot 2024-04-10 at 13 29 57

Solution: use a different string, since it's arbitrary anyway

### DIFF
--- a/jupyterhub/tests/test_pages.py
+++ b/jupyterhub/tests/test_pages.py
@@ -213,7 +213,9 @@ async def test_spawn_handler_access(app):
     r.raise_for_status()
 
 
-@pytest.mark.parametrize("has_access", ["all", "user", "group", False])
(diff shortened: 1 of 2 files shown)


consideRatio (approved):

krassowski:
I see two more:

```
hub/tests/test_shares.py::test_create_share[group] PASSED         [ 72%]
hub/tests/test_shares.py::test_shares_api_user_group_doesnt_exist[group] 
```

minrk:
switched it to using `pytest.param` to change only the output and avoid modifying the test input at all, which I just learned about from @krassowski in #4748!

PR #4784, opened 2024-04-14
doc: list/get token response is different from post

manics:
The description for the `token` field says it's only included in new token requests, but this isn't shown in the summary response.
https://discourse.jupyter.org/t/token-to-token-id-rest-api-with-jupyterhub/24971/5

I've used YAML anchors/merges to avoid duplication https://stackoverflow.com/questions/64224811/yaml-merge-multiply-anchors-into-the-same-map
I don't know whether it's officially supported, but it seems to work with redoc locally (though apparently not on CI/RTD).

Also fixes the list tokens response to be wrapped by an `api_tokens` object, and some miscellaneous typos.

### DIFF
--- a/docs/source/_static/rest-api.yml
+++ b/docs/source/_static/rest-api.yml
@@ -553,9 +553,12 @@ paths:
           content:
             application/json:
               schema:
-                type: array
-                items:
-                  $ref: "#/components/schemas/Token"
+                type: object
+                properties:
+                  api_tokens:
+                    type: array
+                    items:
+                      $ref: "#/components/schemas/Token"
         401:
           description: Authentication/Authorization error
           content: {}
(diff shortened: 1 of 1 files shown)


minrk:
Changes make sense! I don't follow the anchor syntax, and it seems like the linter doesn't either. I'm not sure that's incorrect syntax, or incomplete yaml support from the linter.

Cost of duplication might be lower than the cost of somewhat arcane anchors. These are very slowly changing things, so feel free to copy if that seems simpler to you.

manics:
I took your advice, and duplicated the content instead.
https://jupyterhub--4784.org.readthedocs.build/en/4784/reference/rest-api.html#operation/get-user-tokens

Issue #4788, opened 2024-04-16
JupyterHub 5.0


minrk:
With #4748 and #4701 landed, the last thing I want to do before making 5.0 is #4774 .

Tasks:

- [x] land bootstrap 5 #4774
- [x] migrating to v5 doc #4787
- [x] changelog PR
- [x] release jupyterhub 5b1

erkin98:
@minrk what is the release date for stable v5?

minrk:
That will depend on feedback from prerelease testing, but probably in a few weeks.

PR #4792, opened 2024-04-18
changelog for 5.0, add migration doc

### LINKED ISSUE #4787: add migration documentation to release process
Having updated some dependencies recently, especially in areas I'm not deeply familiar with, I've been really pleased when I find a dedicated "upgrading to X" or "upgrading from Y" docs, with sections like "maybe you used to `foo()`, now that looks like `bar()`." Two examples I've benefitted from recently:

- [react-router](https://reactrouter.com/en/main/upgrading/v5)
- [traefik](https://doc.traefik.io/traefik/migration/v1-to-v2/)

### Proposed change

I think we should add dedicated migration docs to the release process, at least for major releases.

### Alternative options

Leave this information as we have it, in upgraded docs and `versionchanged` notes, combined with the info-dump changelog. Much of the information usually exists somewhere, but not consolidated, and not expressed with this reader goal in mind.

Deployments upgrading JupyterHub versions.

At least for major versions, write a dedicated "upgrading to JupyterHub X" document with for-humans descriptions of any APIs or configurations that are broken or deprecated. Essentially, any time we have a 'new way' of doing something, document an example of the old way and what that same example looks like.

minrk:
closes #4787 

This is the llast PR needed before making a prerelease: #4788


docs/source/howto/upgrading-v5.md
```diff
@@ -0,0 +1,144 @@
...
+All JupyterHub deployments which care about protecting users from each other are encouraged to enable per-user domains, if possible,
+as this provides the best isolation between user servers.
+
+To enable subdomains, set:
+
+```python
+c.JupyterHub.subdomain_host = "https://myjupyterhub.example.org"
+```
+
+If you were using subdomains before
```
krassowski, line 32: Incomplete sentence?

docs/source/howto/upgrading-v5.md
```diff
@@ -0,0 +1,144 @@
...
+
+JupyterHub 5 allows complete customization of the subdomain scheme and changes the default subdomain scheme.
+via the new {attr}`.JupyterHub.subdomain_hook`.
```
krassowski, line 35: ?

```suggestion
JupyterHub 5 allows complete customization of the subdomain scheme via the new {attr}`.JupyterHub.subdomain_hook`,
and changes the default subdomain scheme.
.
```

docs/source/howto/upgrading-v5.md
```diff
@@ -0,0 +1,144 @@
...
+```python
+c.JupyterHub.subdomain_hook = "legacy"
+```
+
+The key differences of the new `idna` scheme:
+
+- It should always produce valid domains, regardless of username (not true for the legacy scheme when using characters that might need escaping or usernames that are long)
+- each Service gets its own subdomain on `service--` rather than sharing `services.`
+
+A table of example of user and services, and their domains with the old and new scheme.
```
krassowski, line 56:
```suggestion
A table of examples of user and services, and their domains with the old and new scheme.
```

Is this meant to be a caption for the table below or is it a placeholder from an earlier draft?

minrk:
Note to self: probably a few more new features worth highlighting like managed roles, not just the breaking changes, but I ran out of time today.

docs/source/howto/upgrading-v5.md
```diff
@@ -0,0 +1,144 @@
...
+Check it out in [the sharing docs](sharing-tutorial).
+
+## Authenticator.allow_all and allow_existing_users
+
+Prior to JupyterHub 5, JupyterHub Authenticators had the _implicit_ default behavior to allow any user who successfully authenticates to login **if no users are explicitly allowed** (i.e. `allowed_users` is empty on the base class).
+This behavior was considered a too-permissive default in Authenticators that source large user pools like OAuthenticator, which would accept e.g. all users with a Google account by default.
+As a result, OAuthenticator 16 introduced two configuration options:
+
+1. `Authenticator.allow_all` (default: False)
+2. `Authenticator.allow_existing_users` (default: True if allowed_users is non-empty, False otherwise)
```
consideRatio, line 98: It sounded from paragraph above this was the oauth config

docs/source/howto/upgrading-v5.md
minrk: meant to merge with the sentence below, I'll try that.

docs/source/howto/upgrading-v5.md
```diff
@@ -95,18 +93,18 @@ Prior to JupyterHub 5, JupyterHub Authenticators had the _implicit_ default beha
 This behavior was considered a too-permissive default in Authenticators that source large user pools like OAuthenticator, which would accept e.g. all users with a Google account by default.
 As a result, OAuthenticator 16 introduced two configuration options:
 
-1. `Authenticator.allow_all` (default: False)
-2. `Authenticator.allow_existing_users` (default: True if allowed_users is non-empty, False otherwise)
+1. `OAuthenticator.allow_all` (default: False)
+2. `OAuthenticator.allow_existing_users` (default: True if allowed_users is non-empty, False otherwise)
```
consideRatio, line 97: For OAuthenticator, this is defaulting to `False` actually :p

docs/source/reference/changelog.md
```diff
@@ -20,6 +20,158 @@ Contributors to major version bumps in JupyterHub include:
...
+
+For information about upgrading, see [upgrading to 5.0 documentation](upgrading-v5).
+
+Changes that are likely to require effort to upgrade:
+
+- New JupyterHub.subdomain_hook and default subdomain scheme
+  is more reliable and should work for all usernames.
+- JupyterHub now requires Python 3.8
+- New `Authenticator.allow_all` and `allow_existing_users` configuration options
+  (the default for `allow_all` is False now, where it was True).
```
consideRatio, line 38: Previously `allowed_users` being truthy led to the equivalent of `allow_existing_users=True`, and previously not having `allowed_users` or other user-allowing config have for some authenticator classes led to the equivalent of `allow_all=True`.

I perceive `(the default for `allow_all` is False now, where it was True).` to be confusing.

docs/source/reference/changelog.md
```diff
@@ -20,6 +20,158 @@ Contributors to major version bumps in JupyterHub include:
 
 ## [Unreleased]
 
+## 5.0
+
+### 5.0.0 - 2024
+
+5.0.0 is a major release of JupyterHub.
+It has lots of cool new features.
```
consideRatio, line 28: Maybe we can have the title reference "beta" somewhere for now, and perhaps lead with a note on how the changelog entry is planned to be updated or not over time as we release another beta or 5.0.0 stable.

minrk: Good call. Done

docs/source/reference/changelog.md
minrk: tried to clear this up

docs/source/howto/upgrading-v5.md
minrk: reordered to clarify that these describe JupyterHub defaults, not OAuthenticator

consideRatio (approved):

minrk:
I have to cut my day short due to planleggingsdag at barnehage, but I'll cut this later today if I get a chance, or on Monday.

minrk:
Thanks for the reviews! Probably not the best thing to make a major prerelease last thing on a Friday afternoon, so I'll publish 5.0b1 first thing Monday.

minrk:
5.0.0b1 publishing now

erkin98:
@minrk what is the release data for stable v5?

PR #4793, opened 2024-04-19
adopt djlint linter/autoformatter for jinja templates

minrk:
[djlint](https://djlint.com) is a linter/formatter for templates, including jinja. This is the first time I used it.

#4774 changed the hierarchy of several elements, which should have re-indented the HTML, but I didn't want the diff to grow too much, and didn't want to manually fix indents, which could have spread well beyond the changes in that PR, so I decided to look for an autoformatter for jinja.

Seems to mostly work pretty nicely. Snags so far:

- the format-js option doesn't handle template controls inside js, which we do in just one place (loading jhdata / requirejs config inside page.html). The alternative is to not format js in templates (there isn't much of it)
- the parser doesn't understand multi-line comments `{# ...\n\n #}`, causing wonky indentation afterward, so I had to replace the one multi-line comment with multiple one-line comments

diff is large because lots of things get re-indented. The linter did catch a couple of unclosed tags in the HTML.

Actual changes are small, mostly adding names to `endblock` statements.

Because I believe linters should help and not be annoying, I disabled easy-to-fix aesthetic lint rules like quote style and casing. If the autoformatter can fix those rules in the future, I'd be happy for them to be enabled. My opinion generally on style-lint things like that: they are the job of the autoformatter not the linter. If the formatter can't fix them, the linter shouldn't enforce them.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,6 +33,15 @@ repos:
     rev: v4.0.0-alpha.8
     hooks:
       - id: prettier
+        exclude: .*/templates/.*
+
+  # autoformat HTML templates
+  - repo: https://github.com/djlint/djLint
+    rev: v1.34.1
+    hooks:
+      - id: djlint-reformat-jinja
+        files: ".*templates/.*.html"
+        types_or: ["html"]
 
   # Autoformat and linting, misc. details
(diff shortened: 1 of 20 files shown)


consideRatio (approved):
Nice!

minrk:
Thanks!

leaving docker timeout for #4795

PR #4794, opened 2024-04-19
Fix counts on list users queries

### LINKED ISSUE #4789: pagination total is not correct with named servers
### Bug description

`total` in the pagination of `GET /api/users` appears to count the number of _servers_, not the number of users.

### How to reproduce

1. enable named servers
2. create one or more named servers
3. `curl -H "Accept: application/jupyterhub-pagination+json" -H "Authorization: token $token" http://127.0.0.1:8000/hub/api/users` or visit the admin page and look at the page info in the footer

Admin page will see e.g. `1-3 of 5` when there are 5 servers, but only 3 users.

#### Expected behavior

`total` is correct

#### Actual behaviour

`total` is too high

### Your personal set up

jupyterhub @ 6a93abbe1c285f33d2cf13e08d2cb98dd03be54b

minrk:
we had some joins to trigger eager loading,
but then `query.count()` returns the count of (user, group) and (user, spawner) pairs, not the count of users but the `joinedload` options added later are the right way to do that, so these joins are unnecessary

closes #4789

### DIFF
--- a/jupyterhub/tests/test_api.py
+++ b/jupyterhub/tests/test_api.py
@@ -333,14 +333,23 @@ async def test_get_users_pagination(
     # populate users
     usernames = []
 
+    groups = []
+    for i in range(3):
+        group = orm.Group(name=f"pagination-{i}")
+        db.add(group)
+    db.commit()
     existing_users = db.query(orm.User).order_by(orm.User.id.asc())
     usernames.extend(u.name for u in existing_users)
 
     for i in range(n - existing_users.count()):
         name = new_username()
         usernames.append(name)
-        add_user(db, app, name=name)
-    print(f"{db.query(orm.User).count()} total users")
+        user = add_user(db, app, name=name)
+        if i % 2:
+            user.groups = groups
+    db.commit()
+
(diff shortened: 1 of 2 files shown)


jupyterhub/tests/test_api.py
```diff
@@ -333,14 +333,23 @@ async def test_get_users_pagination(
...
     existing_users = db.query(orm.User).order_by(orm.User.id.asc())
     usernames.extend(u.name for u in existing_users)
 
     for i in range(n - existing_users.count()):
         name = new_username()
         usernames.append(name)
-        add_user(db, app, name=name)
-    print(f"{db.query(orm.User).count()} total users")
+        user = add_user(db, app, name=name)
+        if i % 2:
```
consideRatio, line 348: Inline comment here would be helpful

consideRatio (approved):
I looked but didn't feel confident reviewing this :/

I suggest self-review and then go for merge. I'm out of time to review much today/this week

minrk:
going to resolve the docker timeout in #4795

PR #4796, opened 2024-04-22
Fix rest API djlint auto-formatting

manics:
https://jupyterhub.readthedocs.io/en/latest/reference/rest-api.html#jupyterhub-rest-api
is currently broken because djlint treated `{{ ... }}` as javascript instead of a template.

I tried adding `{# djlint: off #} ... {# djlint: on #}` around just the `script` block, but there seems to be a [bug where it still partially auto-formats the content](https://github.com/djlint/djLint/issues/824), so after some trial and error I've disabled it in pre-commit.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -42,6 +42,7 @@ repos:
       - id: djlint-reformat-jinja
         files: ".*templates/.*.html"
         types_or: ["html"]
+        exclude: redoc.html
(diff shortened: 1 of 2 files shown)


minrk:
Thanks! I caught this same bug in page.html, but missed this one.

We could also totally turn off format-js, since it doesn't seem to work great and we have very little js-in-html-templates anyway.

PR #4799, opened 2024-04-24
Relax dependency on async_generator

lahwaacz:
async_generator was written as a compatibility library for Python 3.5 and 3.6. The <s>aclosing</s> and asynccontextmanager objects are available in the standard contextlib module since Pyhton 3.7, JupyterHub already requires at least Python 3.8.

See also https://github.com/python-trio/async_generator/issues/35

### DIFF
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -5,9 +5,14 @@
 import asyncio
 import inspect
 import json
+import sys
 from datetime import timedelta, timezone
 
-from async_generator import aclosing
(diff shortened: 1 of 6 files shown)


lahwaacz:
Err, the [contextlib.aclosing](https://docs.python.org/3/library/contextlib.html?highlight=contextlib.contextmanager#contextlib.aclosing) object is available only since Python 3.10. I've updated the commit to make the external package required only for the two older Python versions that are still supported.

minrk:
Thanks!

PR #4805, opened 2024-05-01
include domain in PrefixRedirectHandler

### LINKED ISSUE #4802: Broken login flow with user subdomains and external proxy on 5.0.0b1
### Bug description

Under the following conditions:

- version 5.0.0b1. In particular, after the XSS fix and in particular the change to no longer set domain on the hub's login cookie. (I suspect all 4.1.x versions are similarly affected, though I have not verified. we're on the main branch for reasons)
- have enabled per-user subdomains
- be using an external proxy, in particular an HA one like traefik + redis, which inherently has consistency delays, which the "redirect loop" mechanism is specifically designed to accommodate.

In the event that the route is still not setup, the hub will handle the request:

https://user-subdomain.hubdomain/user/user-name/lab 

This is handled by the `PrefixRedirectHandler` handler class. This route is unauthenticated. It will redirect to:

[/hub/user/user-name/lab](/hub/user/user-name/lab)

which of course, resolves to the full URL **on the subdomain**:

http://user-subdomain.hubdomain/hub/user/user-name/lab

which is handled by the `UserUrlRedirect` handler class. This route is authenticated.

#### Previous Behavior

Previously, as the login cookie had domain set, the browser would still send the login cookie even when interacting with the hub via the subdomain, and so the redirect flow would work fine. Eventually, the redirect loop would cease when the routes reached consistency, and the end-user will settle on their singleuser server.

I've verified this on an older revision on the main branch.

#### Observed Behavior

Since the change to the domain property of the cookie, the user is now redirected to the **login page** - on the subdomain in fact - because the login cookie is not sent by the browser to the hub, as the browser is navigating to the hub via the subdomain, so the hub doesn't recognize the session as being logged in. The end-user never lands on their singleuser server.

### How to reproduce

Without actually going and manually disabling a route, I found that comparing the behavior of

http://user-subdomain.hubdomain/hub/user/user-name/lab

before and after the change to the domain property of the cookie is sufficient to demonstrate the issue. I imagine this technique would work with e.g. the z2jh CSP with subdomains enabled (though of course in a CSP-based setup, with the single-node proxy and synchronous api, I doubt this issue would manifest)

minrk:
redirects user.domain/user/foo -> hub.domain/hub/user/foo when server is not running (previously redirect only included path)

ensures right cookies, etc. are available for the subsequent request to `/hub/user/...`.

regression caused by #4563 which was backported to 4.1, so this should be, too

fixes #4802

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -1537,6 +1537,16 @@ class PrefixRedirectHandler(BaseHandler):
     """Redirect anything outside a prefix inside.
 
     Redirects /foo to /prefix/foo, etc.
+
+    Redirect specifies hub domain when public_url or subdomains are enabled.
+
+    Mainly handles requests for non-running servers, e.g. to
+
+    /user/tree/ -> /hub/user/tree/
+
+    UserUrlHandler will handle the request after redirect.
+    Don't do anything but redirect here because cookies, etc. won't be available to this request,
+    due to not being on the hub's path or possibly domain.
     """
 
     def get(self):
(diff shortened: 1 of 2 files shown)


jupyterhub/handlers/base.py
```diff
@@ -1554,7 +1564,19 @@ def get(self):
...
+
+        # when using subdomains,
+        # make sure we redirect `user.domain/user/foo` -> `hub.domain/hub/user/foo/...`
+        # so that the Hub handles it properly with cookies and all
+        public_url = self.settings.get("public_url")
+        subdomain_host = self.settings.get("subdomain_host")
+        if public_url:
+            redirect_url = urlunparse(public_url._replace(path=redirect_path))
+        elif subdomain_host:
+            redirect_url = url_path_join(subdomain_host, redirect_path)
```
johnpmayer, line 1578: is this right? a few things

- if subdomain_host is set, public_url also gets a default value
- subdomain host may be different than where the host lives?

johnpmayer: I guess this is right. Not sure why I'm not seeing the redirect I expect in my deployment.

johnpmayer: Ok, I've verified this fix!

minrk: Thanks for confirming! Yes, the hub ought to be at `subdomain_host`. It is valid and reasonable for the Hub to be at a different domain, or its own subdomain peer to users (technically this is possible now because subdomain_hook can ignore subdomain_host for user URLs), but as it currently works `subdomain_host == hub host`.

PR #4812, opened 2024-05-09
ci: enable pip cache

minrk:
add `cache: pip` to setup-python invocations

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -158,6 +158,7 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: "${{ matrix.python }}"
+          cache: pip
 
       - name: Install Python dependencies
         run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -159,6 +159,10 @@ jobs:
         with:
           python-version: "${{ matrix.python }}"
           cache: pip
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -158,6 +158,7 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: "${{ matrix.python }}"
+          cache: pip
```
consideRatio, line 161: Do you know after when / how the cache will be invalidated by default?

minrk: by default the cache key is the hash of requirements.txt, so if our dependencies change, this will clear the cache. We can set [cache-dependency-path](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#caching-packages) to be more precise, which I've done now.

Note that an outdated cache doesn't cause problems because it only caches pip's own cache dir, i.e. downloaded wheels, not install results. So when there is a cache hit on an outdated package, pip will still lookup and install the latest available version, updating the cache.

manics (approved):

.github/workflows/test.yml
consideRatio: Ooooh then im very much in favor of this without any hesitation!

consideRatio (approved):

PR #4180, opened 2022-10-27
Update EditUser.jsx

liliyao2022:
This failRegexEvent() function is temporarily unused, delete it.

PR #4815, opened 2024-05-13
admin: don't use state change to update offset

### LINKED ISSUE #4810: Admin UI glitch - double pressing on "Next" page button causes bugged state
In the GIF animation below, you see me first click once, move the cursor away, and then press twice in fast succession. Only three clicks are made, but after the double click it gets stuck going back and forth between page 100-150 and page 150-200.

click-then-doubleclick

Observed in JupyterHub 4.1.5 20240508073043

minrk:
set offset -> request page -> response sets offset is a recipe for races

instead, send request with new offset and only update offset state based on the actual loaded data

made easier by consolidating page update requests into single loadPageData

this also fixes the problem where the pagination info would be updated _before_ the content, so as seen in #4180, it would go:

0. page 1-10
1. click next
2. pagination updated to 11-20, send request for 11-20, 1-10 still displayed
3. request completes, update display to 11-20

by detaching the "request next page" from the state cycle, the looping requests shouldn't happen.  I tried to solve this in general by debouncing `loadPageData`, but had a hard time figuring that out. And offset is really the only state variable that's an input to the request which can also be set to a new value by the response.

closes #4810

### DIFF
--- a/jsx/src/components/ServerDashboard/ServerDashboard.jsx
+++ b/jsx/src/components/ServerDashboard/ServerDashboard.jsx
@@ -41,7 +41,7 @@ const ServerDashboard = (props) => {
   let user_data = useSelector((state) => state.user_data);
   const user_page = useSelector((state) => state.user_page);
 
-  const { setOffset, offset, setLimit, handleLimit, limit, setPagination } =
+  const { offset, setLimit, handleLimit, limit, setPagination } =
     usePaginationParams();
 
   const name_filter = searchParams.get("name_filter") || "";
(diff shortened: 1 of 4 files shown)


jsx/src/components/ServerDashboard/ServerDashboard.jsx
```diff
@@ -123,25 +123,38 @@ const ServerDashboard = (props) => {
...
+      ...params,
     });
+  };
 
-  useEffect(() => {
-    updateUsersWithParams()
+  // single callback to reload the page
+  // uses current state, or params can be specified if state
+  // should be updated _after_ load, e.g. offset
+  const loadPageData = (params) => {
```
consideRatio, line 150: Very thankful for these function descriptions!

jsx/src/components/ServerDashboard/ServerDashboard.jsx
```diff
@@ -41,7 +41,7 @@ const ServerDashboard = (props) => {
   let user_data = useSelector((state) => state.user_data);
   const user_page = useSelector((state) => state.user_page);
 
-  const { setOffset, offset, setLimit, handleLimit, limit, setPagination } =
+  const { offset, setLimit, handleLimit, limit, setPagination } =
     usePaginationParams();
```
consideRatio, line 45: usePaginationParams still defines and returns `setOffset`, and I expected this assignment to run into issues due to that - but even if it works due to name matching of some kind, maybe `setOffset` should be removed from `usePaginationParams`?

```
  return {
    offset,
    setOffset,
    limit,
    setLimit,
    handleLimit,
    setPagination,
  };
```

consideRatio: Ah, I saw below...

```
    return updateUsers({
      offset: offset,
      limit,
```

So I presume that `{variable1, key2: variable2}` becomes `{variable1: variable1-value, key2: variable2-value}` and that makes it possible to name-match in the assignment etc.

---

Okay, so __its just a question about cleanup of `setOffset` or not remaining__

jsx/src/components/ServerDashboard/ServerDashboard.jsx
```diff
@@ -590,8 +584,13 @@ const ServerDashboard = (props) => {
           limit={limit}
           visible={user_data.length}
           total={total}
-          next={() => setOffset(offset + limit)}
-          prev={() => setOffset(offset - limit)}
+          // don't trigger via setOffset state change,
```
consideRatio, line 587: Possible cleanup action here mentioning `setOffset` depending if it was removed.

consideRatio (approved):
Optional cleanup suggestion of now unused `setOffset` in util file - LGTM!

minrk:
You're right, Groups will definitely have the same issue. Updated it to match.

I think having `setOffset` defined even though we don't use it is still sensible. It's specific to _how_ it's used in these two views, where we call `setPagination` with the result. If, instead, we used `setOffset` to trigger the request and _didn't_ call `setPagination` with the response, that would avoid the issue as well.

PR #4798, opened 2024-04-24
add full_url, full_progress_url to server models

minrk:
if public_url is defined

This is part of why I added `public_url`, but didn't get to this part of it.

PR #4817, opened 2024-05-14
add full URLs to share models

minrk:
- full_url for SharedServer
- full_accept_url for ShareCode

same as #4798, applied to shares

### DIFF
--- a/jupyterhub/apihandlers/shares.py
+++ b/jupyterhub/apihandlers/shares.py
@@ -5,6 +5,7 @@
 import json
 import re
 from typing import List, Optional
+from urllib.parse import urlunparse
 
 from pydantic import (
     BaseModel,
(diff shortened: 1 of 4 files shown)


jupyterhub/apihandlers/shares.py
```diff
@@ -95,7 +96,7 @@ def server_model(self, spawner):
             }
         }
         # subset keys for sharing
-        for key in ["name", "url", "ready"]:
+        for key in ["name", "url", "full_url", "ready"]:
```
consideRatio, line 99: The function's docstring mentions the elements in this list and has now become outdated

consideRatio (approved):
Looks right to me, but I've not tested this or similar.

jupyterhub/apihandlers/shares.py
minrk: thanks, added.

PR #4818, opened 2024-05-17
Bump to Ubuntu 24.04

yuvipanda:
- Stop using nodesource, as I suspect 24.04 has a new enough node

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,7 +26,7 @@
 # the JupyterHub wheel on the native architecture only
 # https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -47,20 +47,12 @@ RUN apt-get update -qq \
     git \
     gnupg \
     locales \
+    nodejs \
```
consideRatio, line 50: It seems this is v18, which is a bit surprising to me - i would have expected v20.

Due to this, the version is downgraded from v20 to v18

https://packages.ubuntu.com/noble/nodejs

yuvipanda: aaaah, fun :)

in that case i'll abandon ship :)

PR #4820, opened 2024-05-23
final changelog for 5.0.0

### LINKED ISSUE #4788: JupyterHub 5.0

minrk:
closes #4788


consideRatio:
A jupyterhub server doesn't require jupyterhub 5 on the user servers right, and jupyterhub 4 user server can still work etc without needing restarts or similar if upgrading?

I know we don't have a test for this, but if we believe its fine, it could be good to know.

Thanks for a lot of work into this @minrk!!!

consideRatio (approved):

minrk:
Yes, a JupyterHub 4 server should be fully compatible with JupyterHub 5. I would love a test for this, but it is complicated to set up! Opened #4824 about it.

minrk:
Getting ready to publish now

PR #4822, opened 2024-05-24
Allow overriding spawner config based on user group membership

yuvipanda:
Similar to 'kubespawner_override' in KubeSpawner, this allows admins to selectivel override spawner configuration based on groups a user belongs to. This allows for low maintenance but extremely powerful customization based on group membership. This is particularly powerful when combined with
https://github.com/jupyterhub/oauthenticator/pull/735

## Dictionary vs List

Ordering is important here, but still I choose to implement this configuration as a dictionary of dictionaries vs a list. This is primarily to allow for easy overriding in z2jh (and similar places), where Lists are just really hard to override. Ordering is provided by lexicographically sorting the keys, similar to how we do it in z2jh.

## Merging config

The merging code is literally copied from KubeSpawner, and provides the exact same behavior. Documentation of how it acts is also copied.

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -50,6 +50,7 @@ from .utils import (
     exponential_backoff,
     maybe_future,
     random_port,
+    recursive_update,
     url_escape_path,
     url_path_join,
 )
@@ -306,6 +307,31 @@ class Spawner(LoggingConfigurable):
             f"access:servers!user={self.user.name}",
         ]
 
+    group_overrides = Union(
+        [Callable(), Dict()],
+        help="""
+        Override specific traitlets based on group membership of the user.
+
+        This can be a dict, or a callable that returns a dict. The key of the dict
+        is *only* used for lexicographical sorting, to guarantee a consistent
+        ordering of the overrides. If it is a callable, it may be async, and will
+        be passed one parameter - the spawner instance. It should return a dictionary.
+
+        The value of the dict is a dict, with the following keys:
+
+        - *groups* - If the user belongs to *any* of these groups, these overrides are
+          applied to their server before spawning.
+        - *spawner_override* - a dictionary with overrides to apply to the Spawner
+          settings. Each value can be either the final value to change or a callable that
+          take the `KubeSpawner` instance as parameter and return the final value. This can
+          be further overridden by 'profile_options'
+          If the traitlet being overriden is a *dictionary*, the dictionary
+          will be *recursively updated*, rather than overriden. If you want to
+          remove a key, set its value to `None`
+        """,
+        config=True,
+    )
+
     handler = Any()
 
     oauth_roles = Union(
@@ -504,7 +530,7 @@ class Spawner(LoggingConfigurable):
         max=1,
         help="""
         Jitter fraction for poll_interval.
-        
+
         Avoids alignment of poll calls for many Spawners,
         e.g. when restarting JupyterHub, which restarts all polls for running Spawners.
 
@@ -1478,6 +1504,54 @@ class Spawner(LoggingConfigurable):
             return r
         except AnyTimeoutError:
             return False
+
+    def _apply_overrides(self, spawner_override: dict):
+        """
+        Apply set of overrides onto the current spawner instance
+
+        spawner_override is a dict with key being the name of the traitlet
+        to override, and value is either a callable or the value for the
+        traitlet. If the value is a dictionary, it is *merged* with the
+        existing value (rather than replaced). Callables are called with
+        one parameter - the current spawner instance.
+        """
+        for k, v in spawner_override.items():
+            if callable(v):
+                v = v(self)
+                self.log.info(
+                    f".. overriding {self.__class__.__name__} value %s=%s (callable result)",
+                    k,
+                    v,
+                )
+            else:
+                self.log.info(
+                    f".. overriding {self.__class__.__name__} value %s=%s", k, v
+                )
+
+            # If v is a dict, *merge* it with existing values, rather than completely
+            # resetting it. This allows *adding* things like environment variables rather
+            # than completely replacing them. If value is set to None, the key
+            # will be removed
+            if isinstance(v, dict) and isinstance(getattr(self, k), dict):
+                recursive_update(getattr(self, k), v)
+            else:
+                setattr(self, k, v)
+
+    async def apply_group_overrides(self):
+        """
+        Apply group overrides before starting a server
+        """
+        user_group_names = {g.name for g in self.user.groups}
+        if callable(self.group_overrides):
+            group_overrides = await maybe_future(self.group_overrides(self))
+        else:
+            group_overrides = self.group_overrides
+        for key in sorted(group_overrides):
+            go = group_overrides[key]
+            if user_group_names & set(go['groups']):
+                # If there is *any* overlap between the groups user is in
+                # and the groups for this override, apply overrides
+                self._apply_overrides(go['spawner_override'])
 
 
 def _try_setcwd(path):
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -323,8 +323,7 @@ class Spawner(LoggingConfigurable):
           applied to their server before spawning.
         - *spawner_override* - a dictionary with overrides to apply to the Spawner
           settings. Each value can be either the final value to change or a callable that
-          take the `KubeSpawner` instance as parameter and return the final value. This can
-          be further overridden by 'profile_options'
+          take the `Spawner` instance as parameter and return the final value.
           If the traitlet being overriden is a *dictionary*, the dictionary
           will be *recursively updated*, rather than overriden. If you want to
           remove a key, set its value to `None`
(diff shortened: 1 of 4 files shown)


ryanlovett (commented):
Thanks @yuvipanda ! This feature would be very useful for us as we have [customized the spawner](https://github.com/berkeley-dsep-infra/datahub/blob/staging/hub/values.yaml#L390) to accomplish something similar. I made a minor suggestion about a traitlet help string.

minrk (commented):
Nice! Some feedback in the implementation, especially  thinking about how it should behave when config provided is incorrect, but 👍 overall.

jupyterhub/spawner.py
```diff
@@ -1479,6 +1505,54 @@ async def _wait_for_death():
...
+        to override, and value is either a callable or the value for the
+        traitlet. If the value is a dictionary, it is *merged* with the
+        existing value (rather than replaced). Callables are called with
+        one parameter - the current spawner instance.
+        """
+        for k, v in spawner_override.items():
+            if callable(v):
+                v = v(self)
+                self.log.info(
+                    f".. overriding {self.__class__.__name__} value %s=%s (callable result)",
```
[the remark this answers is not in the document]
minrk: Also thinking about log noise, maybe a single log line for all overridden keys rather than logging every single key/value separately.

jupyterhub/spawner.py
```diff
@@ -306,6 +307,31 @@ def _default_access_scopes(self):
...
+        ordering of the overrides. If it is a callable, it may be async, and will
+        be passed one parameter - the spawner instance. It should return a dictionary.
+
+        The value of the dict is a dict, with the following keys:
+
+        - *groups* - If the user belongs to *any* of these groups, these overrides are
+          applied to their server before spawning.
+        - *spawner_override* - a dictionary with overrides to apply to the Spawner
+          settings. Each value can be either the final value to change or a callable that
+          take the `KubeSpawner` instance as parameter and return the final value. This can
```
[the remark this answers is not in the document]
yuvipanda: Done

jupyterhub/spawner.py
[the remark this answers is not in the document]
yuvipanda: Done

jupyterhub/spawner.py
```diff
@@ -1479,6 +1505,54 @@ async def _wait_for_death():
...
+            else:
+                self.log.info(
+                    f".. overriding {self.__class__.__name__} value %s=%s", k, v
+                )
+
+            # If v is a dict, *merge* it with existing values, rather than completely
+            # resetting it. This allows *adding* things like environment variables rather
+            # than completely replacing them. If value is set to None, the key
+            # will be removed
+            if isinstance(v, dict) and isinstance(getattr(self, k), dict):
```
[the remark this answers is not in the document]
yuvipanda: @minrk given this code has been around in kubespawner for years, I'm inclined to leave it as is. Right now, if the *admin* typos a config here, they'll get a hard failure during spawn time. I think this is the *correct* and most obvious response. I can try to do a catch of that and provide another log message, but IMO as is it's ok.

jupyterhub/spawner.py
```diff
@@ -306,6 +307,31 @@ def _default_access_scopes(self):
             f"access:servers!user={self.user.name}",
         ]
 
+    group_overrides = Union(
+        [Callable(), Dict()],
+        help="""
+        Override specific traitlets based on group membership of the user.
+
+        This can be a dict, or a callable that returns a dict. The key of the dict
```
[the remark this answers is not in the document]
yuvipanda: How about now, @ryanlovett?

jupyterhub/spawner.py
```diff
@@ -306,6 +307,30 @@ def _default_access_scopes(self):
...
+        ordering of the overrides. If it is a callable, it may be async, and will
+        be passed one parameter - the spawner instance. It should return a dictionary.
+
+        The value of the dict is a dict, with the following keys:
+
+        - *groups* - If the user belongs to *any* of these groups, these overrides are
+          applied to their server before spawning.
+        - *spawner_override* - a dictionary with overrides to apply to the Spawner
+          settings. Each value can be either the final value to change or a callable that
+          take the `Spawner` instance as parameter and return the final value.
```
manics, line 326:
```suggestion
          take the `Spawner` instance as parameter and returns the final value.
```

jupyterhub/spawner.py
```diff
@@ -306,6 +307,30 @@ def _default_access_scopes(self):
...
+        help="""
+        Override specific traitlets based on group membership of the user.
+
+        This can be a dict, or a callable that returns a dict. The keys of the dict
+        are *only* used for lexicographical sorting, to guarantee consistent
+        ordering of the overrides. If it is a callable, it may be async, and will
+        be passed one parameter - the spawner instance. It should return a dictionary.
+
+        The value of the dict is a dict, with the following keys:
+
```
manics, line 321: This behaviour is easy to understand if you're already familiar with how Z2JH and KubeSpawner are configured, but we're now introducing this to a new audience.

I think an example configuration, either here or elsewhere in the JupyterHub documentation, would be useful.

jupyterhub/user.py
```diff
@@ -905,6 +905,7 @@ async def spawn(self, server_name='', options=None, handler=None):
             # wait for spawner.start to return
             # run optional preparation work to bootstrap the notebook
             await maybe_future(spawner.run_pre_spawn_hook())
+            await spawner.apply_group_overrides()
```
manics, line 908: What's your thinking for running this after `pre_spawn_hook` instead of before?
https://github.com/jupyterhub/jupyterhub/blob/fb85cfb118af55c34a24ba20ff1bcb595325c5ab/jupyterhub/spawner.py#L909-L913

manics (commented):
Seems fine to me! My only thought is whether we're happy to limit this type of override to groups only (I think it's fine), or if we'll want to expand it in future in which case we should consider any design implications now.

yuvipanda:
@manics i was thinking about what else we may use. I feel groups are the most natural place for this kinda functionality. What other kinda overrides do you think we may add in the future?

jupyterhub/user.py
yuvipanda: Just missed it, fixed now.

yuvipanda:
Thinking about this some more, I think if in the future we do want to support other kinds of overrides, we can:

1. Rename the config from `group_overrides` to `overrides`
2. Add the possibility of more keys other than `groups`.

This would be fairly backwards compatible and easy to evolve. We can also if needed change `groups` from a list to a dict with its own config as well, and that should be backwards compatible too.

jupyterhub/spawner.py
yuvipanda: @manics added one! what do you think?

manics (approved):

jupyterhub/spawner.py
```diff
@@ -306,6 +307,53 @@ def _default_access_scopes(self):
...
+        1. Add the environment variable "AM_I_USER" to everyone in the "user" group
+        2. Add the environment variable "AM_I_ADMIN" to everyone in the "admin" group.
+           If a user is part of both "user" and "admin", they will get *both* these env
+           vars, due to the dictionary merging functionality.
+        3. Add a higher memory limit for everyone in the "admin" group.
+
+        c.Spawner.group_overrides = {
+            "01-admin-env-add": {
+                "groups": ["admin"],
```
consideRatio, line 341: The example leads me to gures there may be user and admin groups by default to represent all users and all admins, but there isn't right?

Maybe student and instructor or similar could be used instead to not cause someone to incorrectly assume user/admin is available?

yuvipanda: I don't think it matters much, I've changed the names.

jupyterhub/spawner.py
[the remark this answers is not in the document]
minrk: Fair enough! FWIW, I do agree that it should be a hard failure, I just thought we could perhaps do better in the error message pointing to the where their config might be wrong.

jupyterhub/spawner.py
```diff
@@ -1479,6 +1527,48 @@ async def _wait_for_death():
...
+            group_overrides = self.group_overrides
+        for key in sorted(group_overrides):
+            go = group_overrides[key]
+            if user_group_names & set(go['groups']):
+                # If there is *any* overlap between the groups user is in
+                # and the groups for this override, apply overrides
+                self._apply_overrides(go['spawner_override'])
+                self.log.info(
+                    f"Applying group_override {key} for {self.user.name}, modifying config keys: {' '.join(go['spawner_override'].keys())}"
+                )
```
minrk, line 1570: I think it's usually a good pattern to log the thing you're about to do right _before_ it happens, which gives better hints when things go wrong, which is much of why we log in the first place:

```suggestion
                self.log.info(
                    f"Applying group_override {key} for {self.user.name}, modifying config keys: {' '.join(go['spawner_override'].keys())}"
                )
                self._apply_overrides(go['spawner_override'])
```

(above just swaps line 1567 and 1568, despite the rendering of the diff)

minrk (approved):
one comment about ordering of log message and action, but LGTM!

minrk:
Awesome!

consideRatio:
Tweaked title to highlight config name for changelog

PR #4835, opened 2024-06-10
reduce cost of event_loop_interval metric

### LINKED ISSUE #4834: v5: Constant CPU usage of at least 1.9%
### Bug description

JupyterHub 5.0 consistently uses at least 1.9% CPU – even when *idle*.

#### Expected behaviour

JupyterHub 4.1 has a CPU usage of 0.05% when *idle*.

### Your personal set up

JupyterHub: [`glcr.b-data.ch/jupyterhub/jupyterhub:5.0-oauth-docker`](https://gitlab.b-data.ch/jupyterhub/jupyterhub/container_registry)

minrk:
also makes sure the PeriodicMetrics config is in the docs, so disable config can be found.

- trade PeriodicCallback for single coroutine (saves ~50% in measurements)
- increase default interval from 20ms to 50ms (and adjust metric buckets to match)
- measure `delay` instead of the overall tick duration (more below) for more strictly defined measurements

Table of [measurements](https://gist.github.com/minrk/12ad941d945bf41d92d607e274c2a3cb) (90% means 90% of samples measured at least this cpu usage, i.e. the 10th percentile)

before:

| interval | avg_cpu | 90%  | 75%  |
| ----- | ------- | ---- | ---- |
| disabled    | 0.0%    | 0.0% | 0.0% | 
| 0.02 (default)  | 2.1%    | 1.8% | 2.1% | 
| 0.05  | 1.1%    | 0.9% | 1.0% | 
| 0.1   | 0.7%    | 0.5% | 0.6% | 
| 0.2   | 0.4%    | 0.0% | 0.3% | 
| 0.5   | 0.2%    | 0.0% | 0.0% |

after:

| interval | avg_cpu | 90%  | 75%  |
| -------- | ------- | ---- | ---- |
| 0.02     | 0.9%    | 0.7% | 0.9% | 
| 0.05 (default)     | 0.4%    | 0.4% | 0.4% | 
| 0.1      | 0.2%    | 0.2% | 0.2% | 
| 0.2      | 0.1%    | 0.0% | 0.2% | 
| 0.5      | 0.1%    | 0.0% | 0.0% |

The question that remains is what should the buckets be, since it no longer makes sense to keep 25-50ms. I've spent entirely too long thinking about this, and I wish the answer were clearer.

Since the measurement is really "how long does it take to wait `interval`", not a direct measurement of the event loop interval itself (too expensive), the event loop duration is really anywhere on the range `[measurement - interval, measurement]`, so for a measurement of 55ms and an interval of 50ms, we know that the event loop blocked for between 5ms and 55ms, but don't know any more than that. In that way, it doesn't make a whole lot of sense to distinguish measurements smaller than 2 * the interval, since they aren't super meaningful. That said, we can infer from statistics that if 99% of measurements at 50ms are less than 55ms, one can reasonably conclude that the actual event loop interval is below 5 ms.

As we have right now, the metric records an _upper_ bound on the responsiveness (i.e. we know with confidence that 90% of ticks are below 50ms and 95% are below 60ms, but that does _not_ mean 5% are between 50ms and 60ms, which would be true in a typical histogram metric).

An alternate definition of the metric is the _delay_, which sets a _lower_ bound, i.e. we know the event loop blocked for at least `delay = measurement - interval` seconds. The advantage of this is it more reliably reports what we do know. The disadvantage is it could under-report actual delays by up to `interval` (in aggregate, the expectation value of this over/under estimate would be `interval / 2`, either way).

My current thinking is:

- record only the delay
- report all delays < threshold as threshold exactly

which this PR currently reflects

alternately, we could keep the same rounding but use the upper bound, i.e. record the same value we have now, but report every measurement below 2 * interval as 2 * interval (i.e. the actual upper bound). That's technically the same information as the truncated delay, but the number reported is bigger by `interval` and represents an actual strict upper bound.

Advantage of delay: "we know it blocked at least this much"; disadvantage: under-estimates actual blocking events by up to resolution / 2 on average
Advantage of upper bound: "we know it did not block more than this"; disadvantage: over-estimates by up to resolution / 2 on average
Advantage of either rounded scheme: far more measurements will be in the lowest bucket and with confidence, rather than smearing not-actually-resolved measurements between `[resolution, resolution * 2]`

Advantage of current scheme: each individual measurement is really an upper bound, there is _some_ information in measurements between `[resolution, resolution * 2]`, at least in aggregate; disadvantage: buckets aren't mutually exclusive (55ms could be 6ms or 54ms)

The reason I like `delay` is that the vast majority of the time in practice, the actual number we are measuring is 0, so rounding up doesn't feel right. The overestimate issue is on measurements _where the loop is blocked_, not all measurements in general.

I feel like I've spent far too long on this, since we're talking about measurements around 20-100ms when the real events we want to notice are 10-100 times that, all of which will show up equivalently in all of the above schemes, except that `delay` will report a 1s delay as `0.95-1`, while the others will report it as `1-1.05`.

closes #4834

### DIFF
--- a/docs/source/reference/monitoring.md
+++ b/docs/source/reference/monitoring.md
@@ -32,3 +32,11 @@ export JUPYTERHUB_METRICS_PREFIX=jupyterhub_prod
 ```
 
 would result in the metric `jupyterhub_prod_active_users`, etc.
+
+## Configuring metrics
+
+```{eval-rst}
+.. currentmodule:: jupyterhub.metrics
+
+.. autoconfigurable:: PeriodicMetricsCollector
+```
--- a/jupyterhub/metrics.py
+++ b/jupyterhub/metrics.py
@@ -22,6 +22,7 @@ them manually here.
     added ``jupyterhub_`` prefix to metric names.
 """
 
+import asyncio
 import os
 import time
 from datetime import timedelta
@@ -29,7 +30,7 @@ from enum import Enum
 
 from prometheus_client import Gauge, Histogram
 from tornado.ioloop import PeriodicCallback
-from traitlets import Any, Bool, Dict, Float, Integer
+from traitlets import Any, Bool, Dict, Float, Integer, default
 from traitlets.config import LoggingConfigurable
 
 from . import orm
@@ -236,17 +237,17 @@ EVENT_LOOP_INTERVAL_SECONDS = Histogram(
     'event_loop_interval_seconds',
     'Distribution of measured event loop intervals',
     namespace=metrics_prefix,
-    # Increase resolution to 5ms below 50ms
+    # don't measure below 50ms, our default
+    # Increase resolution to 5ms below 75ms
     # because this is where we are most sensitive.
-    # No need to have buckets below 25, since we only measure every 20ms.
+    # No need to have buckets below 50, since we only measure every 50ms.
     buckets=[
-        # 5ms from 25-50ms
-        25e-3,
-        30e-3,
-        35e-3,
-        40e-3,
-        45e-3,
+        # 5ms from 50-75ms
         50e-3,
+        55e-3,
+        60e-3,
+        65e-3,
+        70e-3,
         # from here, default prometheus buckets
         75e-3,
         0.1,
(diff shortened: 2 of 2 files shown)


minrk:
I think test-docs is not related and should be fixed by #4836

yuvipanda:
Based on the facts that the tests pass, I assume that the long running single async task is cleaned up correctly.

With that, and

> I feel like I've spent far too long on this

I'm going to merge this, and we can re-evaluate this if it runs into more issues. It may be useful to confirm with another pyspy run (as was done in #4834) but the table in the description is good enough for me.

Thanks for working through this @minrk, and from #4834 thanks for the pyspy output, @manics!

PR #4836, opened 2024-06-10
fix formatting of group_overrides docstring

minrk:
Sphinx is complaining about the docstring in #4822

- literals for dict keys
- use example header section
- missing `::` for code block

I think we may need to update the path conditionals for test-docs, but I'll do that in its own PR.


yuvipanda:
Thanks, @minrk

PR #4837, opened 2024-06-11
Provide consistent myst references to documentation pages - part 1

yuvipanda:
While doing https://github.com/jupyterhub/jupyterhub/pull/2726, I realized we don't have a consistent way to format references inside the docs. I now have them be formatted to match the name of the file, but using `:` to separate them instead of `/` or `-`. `/` makes it ambiguous when using with markdown link syntax, as it could be a reference or a file. And using `-` is ambiguous, as that can be the name of the file itself.

This PR does about half, I can do the other half later (unless someone else does).

### DIFF
--- a/docs/source/contributing/community.md
+++ b/docs/source/contributing/community.md
@@ -1,3 +1,5 @@
+(contributing:community)=
+
 # Community communication channels
 
 We use different channels of communication for different purposes. Whichever one you use will depend on what kind of communication you want to engage in.
--- a/docs/source/contributing/contributor-list.md
+++ b/docs/source/contributing/contributor-list.md
@@ -1,3 +1,5 @@
+(contributing:contributors)=
+
 # Contributors
 
 Project Jupyter thanks the following people for their help and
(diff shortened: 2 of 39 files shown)


minrk:
Thanks for going through these! I wasn't sure if `:` would be a problem for intersphinx target resolution in rst, so I tested with config:

```python
intersphinx_mapping = {
    "jupyterhub": ("https://jupyterhub--4837.org.readthedocs.build/en/4837/", None),
}
```

and the ref:

```rst
:ref:`jupyterhub:contributing:contributors`
```

resolves as it should, so all good there.

Changing these anchors will break existing intersphinx links to these targets, if there are any. If you're okay with that, go ahead and merge. Otherwise, I suppose we can duplicate some of these if we think folks are linking to them?

FWIW, using the path is sensible as a starting point, but for the most part I think these links should be _logical_ and describe the doc section it's referring to, mostly independent of its location, i.e. moving a chunk of docs (e.g. splitting or merging files) should keep link targets without requiring any updates to other docs that link _to_ it.

minrk (approved):

yuvipanda:
Thanks @minrk! I'll make another PR to add these to other places. I think it's ok to break intersphinx once.

PR #4844, opened 2024-06-25
allow stop while start is pending

### LINKED ISSUE #2975: Deleting a spawning server
When I try to delete a named server invoking the `DELETE /users/<username>/servers/<servername>` API, I get this response if the server is spawning
> {"status": 400, "message": "lorenzo.cavazzi.tech:flights-tutorial-1e75e55c is pending spawn, please wait"}

This is fine, but there are specific cases when I do really want to delete the server even if it's spawning, usually because it's stuck somewhere or it's taking longer than expected, but I don't need to wait `start_timeout` minutes (in our case, that value is quite high).
Is there a way to force that?

I have tried to manually remove the pods from Kubernetes, that works but there are a few side effects, like not being able to spawn the same server until `cull_idle` kick in, which still requires a few minutes according to the `start_timeout` value.

minrk:
Stopping a server mid-start cancels the start task rather than waiting for it to finish or timeout before stop can be attempted.

Removes the "user is pending spawn, please wait" message.

also fixes cancellation when start_timeout is reached, which was previously left running forever, I think because of a misunderstanding of the difference between `gen.with_timeout` (shields awaited tasks from cancellation) vs `asyncio.wait_for` (propagates cancel, cancels task on timeout). I checked other uses of `gen.with_timeout` for similar issues, and I believe they are correct (we want to wait up to a limit, but leave it running if it's not done), but this one should have stopped the task when timeout is reached.

closes #2975

cc @cboettig

### DIFF
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -694,14 +694,22 @@ class UserServerAPIHandler(APIHandler):
                 asyncio.ensure_future(_remove_spawner(spawner._stop_future))
             return
 
-        if spawner.pending:
-            raise web.HTTPError(
-                400,
-                f"{spawner._log_name} is pending {spawner.pending}, please wait",
-            )
-
         stop_future = None
-        if spawner.ready:
+        if spawner.pending:
+            # we are interrupting a pending start
+            # hopefully nothing gets leftover
+            self.log.warning(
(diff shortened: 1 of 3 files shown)


manics (approved):
Tested stopping
- before the start timeout
- after start but before the http timeout

Both LGTM!

minrk:
Thanks @manics!

Issue #4845, opened 2024-06-28
JupyterHub differs in `xsrf` behaviour from jupyter-server


krassowski:
### Bug description

jupyter-server accepts requests such as POST if they have an `xsrf` token OR if they include the correct auth token. Quoting from jupyter-server docs:

>  Jupyter server includes protection from cross-site request forgeries, requiring API requests to either:
> - originate from pages served by this server (validated with XSRF cookie and token), or
> - authenticate with a token

For reference here is [the link to the docs](https://github.com/jupyter-server/jupyter_server/blob/b85c15b5678cd6ea9de28e4f547bc0dfef0a69c7/jupyter_server/serverapp.py#L1299-L1308) and here is the [`JupyterHandler.check_xsrf_cookie` implementation](https://github.com/jupyter-server/jupyter_server/blob/b85c15b5678cd6ea9de28e4f547bc0dfef0a69c7/jupyter_server/base/handlers.py#L532-L535).

Now, JupyterHub overrides the `BaseHandler` in a way which is not obvious to an extension author. For example, we were surprised to see that the POST handler fails when used by API in [`jupyterlab-gallery`](https://github.com/nebari-dev/jupyterlab-gallery) (which you folks may like - it is a work-in-progress nbgipuller wrapper to be used inside of JupyterLab). This can be seen in https://github.com/nebari-dev/jupyterlab-gallery/pull/27/commits/1e3fb3ca9eb73e1c3d5c0ff24ad92ab3cfe1fb9b for which all the tests pass nicely on `jupyter-server` despite having no `xsrf` token.

I understand that this is because of this code:

https://github.com/jupyterhub/jupyterhub/blob/01aeb84a130a4c34b04f6e20673a74f166212ff2/jupyterhub/singleuser/mixins.py#L819-L843

which not only injects its own `BaseHandler`, but also explicitly overwrites `check_xsrf_cookie`

This was only changed recently in https://github.com/jupyterhub/jupyterhub/pull/4771

> As usual, this may be a bug or documentation issue, or an issue with myself not being able to find where it is mentioned in the docs

### How to reproduce

1. Create a jupyter-server extension with a POST endpoint.
2. Send a requests to the endpoint

#### Expected behaviour

A couple of ways we can approach this:
- a) change behaviour in JupyterHub (which may be undesirable - that's fine)
- b) improve documentation for extension authors here and on jupyter-server side to make this explicit
[...]

PR #4848, opened 2024-07-01
changelog for 5.1.0

minrk:
small release; a few nice new config options and one performance regression fix

trying to keep releases small and more frequent


manics (approved):
LGTM.

Is there a quick-fix for https://github.com/jupyterhub/jupyterhub/issues/4845 ? If so it'd be nice to get that in, and save the effort of making another patch release.

minrk:
I'll have a look at that tomorrow

krassowski:
I'm not quite sure if my diagnosis of that issue was correct, I would not want to block you on this or take up your time folks. I will get back to it as soon as I can.

minrk:
I'm in no rush to release, I'd like to have a good answer either way, whether it's docs or a bug fix in JupyterHub or even in JupyterLab.

yuvipanda:
Just wanted to check back on this :)

minrk:
I ran out of time to finish up one last thing before vacation, I'll do this next week when I get back.

PR #4856, opened 2024-07-23
Show insecure login warning when not in a secure context

### LINKED ISSUE #4855: Check for secure context before showing insecure login warning
### Proposed change

Currently a warning is shown when logging into JupyterHub over HTTP, to indicate that the connection may not be secure. There exists a browser mechanism for determining this, specifically [Secure Contexts](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts).

Secure Contexts check several other things in addition to the main connection protocol. They check that no active content (such as scripts) has been loaded insecurely, that the encryption method isn't deprecated, etc. Therefore they are a much more robust check than simply checking the scheme.

Localhost over HTTP is also considered a secure context, which will prevent this warning being shown unnecessarily in many development or demonstration use cases.

### Alternative options

We could do nothing, as we currently just check the scheme is not HTTP, which covers most cases. This unfortunately raises a warning when accessed over localhost, despite that being a secure context. On the other hand, the secure context API was only added to browsers in 2016, though I don't think JupyterHub still supports IE11, and even then the warning would just always be shown.

For most users this would be an invisible change, but it would help those accessing JupyterHub over an unsecure encryption algorithm, or running a local copy from localhost.

Instead of checking the URL scheme, check the `window.isSecureContext` property.

jfrost-mo:
Secure contexts are a more robust way of checking that a browsing context is authenticated and confidential. Compared to comparing the scheme this covers cases where the connection is encrypted, but using a broken algorithm.

Notably, localhost is considered a secure context, even over HTTP.

For more detail on secure contexts, see:
https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts

Fixes #4855

### DIFF
--- a/share/jupyterhub/templates/login.html
+++ b/share/jupyterhub/templates/login.html
@@ -91,7 +91,7 @@
 {% block script %}
   {{ super() }}
   <script>
-    if (window.location.protocol === "http:") {
+    if (!window.isSecureContext) {
       // unhide http warning
       var warning = document.getElementById('insecure-login-warning');
       warning.className = warning.className.replace(/\bhidden\b/, '');


manics (approved):
Thanks!

Issue #4857, opened 2024-07-26
Support running jupyter-server `ExtensionApp` as standalone servers


krassowski:
### Proposed change

JupyterHub allows to customize the server run in the container. I would like some extensions in a separate sidecar and use proxies to only expose the select endpoints in the main container. This is useful both for security, and for isolation of code which may crash the single user server.

I would still like to benefit from the customizations made by JupyterHub.

`ExtensionApp` is meant to be both a server extension, and work as a standalone app when launched from CLI (with jupyter-server still running underneath), as documented [here](https://jupyter-server.readthedocs.io/en/latest/developers/extensions.html#authoring-a-configurable-extension-application). `ExtensionApp` also has a neat configuration option called `load_other_extensions` which allows to spawn a jupyter server with a single-extension (which is not in possible in general as of today https://github.com/jupyter-server/jupyter_server/issues/1445).

JupyterHub allows to swap the `JUPYTERHUB_SINGLEUSER_APP` for an App. However, it currently assumes that the App is a subclass of `ServerApp`.

### Alternative options

Document that there are no plans to support it

### Who would use this feature?

Deployments with high security requirements, or with high availability requirements (so that extensions which could bring down the single user container are isolated)

### (Optional): Suggest a solution

`ExtensionApp`s are not subclasses of `ServerApp`, but they do create a `ServerApp` during `launch_instance()`. The `initialize_server()` class method could be easily overridden for `ExtensionApp`. This looks like a minimal change to the existing logic.

manics:
Would an ExtensionApp allow someone to for example build a completely minimal application without any default endpoints (e.g. no contents API)? If not, is there a higher level base class we could define that JupyterHub could interface with, and that others can build custom applications on?

krassowski:
As of today `ExtensionApp` adds handlers for default endpoints, including the contents API. I wonder if we should have an option similar to `load_other_extensions`, say `load_api_handlers` allowing to disable these endpoints for individual `ExtensionApp` easily.

> is there a higher level base class we could define that JupyterHub could interface with, and that others can build custom applications

The class hierarchy is:

```
jupyter_server.serverapp.ServerApp → jupyter_core.application.JupyterApp → traitlets.Application
```

There is a couple of `JupyterApp`s in jupyter-server which are not full servers, but these do not have `login_handler_class`  (and friends) nor `identity_provider_class` as this is only added on `ServerApp` level, so the monkeypatching of these would fail:

[...]

PR #4860, opened 2024-07-27
Pass `kwargs` down to `initialize()` call of the server

krassowski:
Addresses https://github.com/jupyterhub/jupyterhub/issues/4857#issuecomment-2254090841

Of note the child class already does it well:

https://github.com/jupyterhub/jupyterhub/blob/0cd5e51dd4f8f0079f7391e9f831dd202877e999/jupyterhub/singleuser/mixins.py#L952-L962


minrk:
Thanks! I didn't realize ExtensionApp didn't follow standard traitlets Application signatures.

PR #4863, opened 2024-08-06
Add PAMAuthenticator.executor_threads configurable, increase default to 4

Will-Shanks:
Following the testing described [here](https://github.com/Will-Shanks/jupyterhub_auth_repro) and a discussion with @manics on [discourse](https://discourse.jupyter.org/t/configuring-pamauthenticator-executor/27436/7) I opened this PR to make the thread pool size used by the PAMAuthenticator configurable.

This is a simple change modeled after this [OAuthenticator](https://github.com/jupyterhub/oauthenticator/blob/cd00a8d6e5a8a6a97088b4a1e73ccb419f197257/oauthenticator/mediawiki.py#L96). After some simple testing with our production environment (~100 active users) adding more threads to this pool has had a significant affect on our hub stability, that we expect could help other sites too.

### DIFF
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -1230,7 +1230,20 @@ class PAMAuthenticator(LocalAuthenticator):
 
     @default('executor')
     def _default_executor(self):
-        return ThreadPoolExecutor(1)
+        return ThreadPoolExecutor(self.executor_threads)
+
+    executor_threads = Integer(
+        1,
+        config=True,
+        help="""
+        Number of executor threads.
+
+        PAM auth requests happen in this thread, so it is mostly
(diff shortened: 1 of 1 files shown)


minrk:
Makes sense! Since incorrect password entry causes a standard delay of ~3 seconds, that delay would block _all_ login attempts, maybe the default should also be higher than 1? Say, 4? It's pretty arbitrary, but at least makes it less likely for one user to slow down everyone else.

Will-Shanks:
Sure, I'll up it to 4. I had set it to 1 since that was the hard coded value, but I agree some parallelism would be good here, and I don't see any downside to it.

minrk:
Thank you!

PR #4865, opened 2024-08-12
Revoke all permissions from Authenticator.blocked_users

### LINKED ISSUE #4861: Adding a user to Authenticator.blocked_users does not affect existing sessions
Consider this scenario:

- Some user account on JupyterHub gets breached and the admin decides to block this user account.
- The admin adds the username to `c.Authenticator.blocked_users`.
- This correctly blocks attempts to start a new session: the user gets `403 : Forbidden Sorry, you are not currently authorized to use this hub. Please contact the hub administrator.`
- However, this does not terminate existing sessions of the user and an intruder may happily keep them open for a long time.

Am I missing something? What is the expected mitigation path for this scenario?

minrk:
rather than only disabling login, fully block the user from Hub operations by removing all group membership and role assignments

This is not optional because I think it's what people generally expect. The previous behavior: block login, but leave user active and able to act with existing credentials and running servers makes _theoretical_ sense, but not really in practice, so I don't feel the need to keep it as default or even possible.

The downside of special treatment for `blocked_users` is that other mechanisms by which `check_blocked` might get info will not get this treatment. However, those mechanisms _cannot be called_ during Hub startup, because they may require non-persisted information from the authentication process. That's harder to communicate about. Authenticator-managed roles are a possibility, but only if the `users` role is auth-managed and can be _completely_ specified (i.e. all members can be specified at once) is the only way to revoke membership by omission.

closes #4861

### DIFF
--- a/jupyterhub/orm.py
+++ b/jupyterhub/orm.py
@@ -1234,6 +1234,8 @@ class APIToken(Hashed, Base):
             orm_token.expires_at = cls.now() + timedelta(seconds=expires_in)
 
         db.commit()
+        if return_orm:
+            return orm_token
         return token
 
     def update_scopes(self, new_scopes):
--- a/jupyterhub/orm.py
+++ b/jupyterhub/orm.py
@@ -1141,7 +1141,6 @@ class APIToken(Hashed, Base):
         expires_in=None,
         client_id=None,
         oauth_client=None,
-        return_orm=False,
     ):
         """Generate a new API token for a user or service"""
         assert user or service
@@ -1234,8 +1233,6 @@ class APIToken(Hashed, Base):
             orm_token.expires_at = cls.now() + timedelta(seconds=expires_in)
 
         db.commit()
-        if return_orm:
-            return orm_token
         return token
 
     def update_scopes(self, new_scopes):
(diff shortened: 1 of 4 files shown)


jupyterhub/tests/test_app.py
```diff
@@ -537,3 +539,66 @@ async def test_recreate_service_from_database(
...
+    assert await spawner.poll() is None
+    assert sorted(role.name for role in user.roles) == ['admin', 'user']
+    assert [g.name for g in user.groups] == [groupname]
+    assert user.admin
+    user_scopes = get_scopes_for(user)
+    assert "access:servers" in user_scopes
+    token_scopes = get_scopes_for(token)
+    assert "access:servers" in token_scopes
+
+    # start a new hub, now with blocked users
```
manics, line 580: To guard against a future test refactor disconnecting the shared state between the two hubs can you add another user that isn't blocked, and verify the user still exists?

minrk: Good call. done.

jupyterhub/orm.py
```diff
@@ -1141,7 +1141,6 @@ def new(
         expires_in=None,
         client_id=None,
         oauth_client=None,
-        return_orm=False,
```
minrk, line 1144: This argument was only partially added by mistake, I think for internal testing. It doesn't do anything.

minrk:
tests fixed

manics (approved):
Thanks!

PR #4869, opened 2024-08-21
add dependabot config for npm

minrk:
should start bumping dependencies for top-level and jsx


manics (approved):
I think this is fine. Let's try it.

PR #4878, opened 2024-08-26
Introduce dark theme

benz0li:
Enable in `jupyterhub_config.py` with:

```python
c.JupyterHub.template_vars = { 'dark_theme': 'true' }
```

### DIFF
--- a/docs/source/howto/templates.md
+++ b/docs/source/howto/templates.md
@@ -8,6 +8,15 @@ example, to be defined once and incorporated into all pages. By providing
 your own template(s), you can have complete control over JupyterHub's
 appearance.
 
+## Dark Theme
+
+By default, bootstrap pages have a white (light) background color.
+
+If you want to change JupyerHub to a darker color, you can set the configuration
+variable `JupyterHub.template_vars = { 'dark_theme': 'true' }`.
+
+This adds `data-bs-theme="dark"` to the `<html>` element of `page.html`.
+
 ## Custom Templates
 
 JupyterHub will look for custom templates in all paths included in the
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -545,6 +545,7 @@ class JupyterHub(Application):
             c.JupyterHub.template_vars = {
                 "key1": "value1",
                 "key2": callable_value,
+                "dark_theme": "true",
             }
         """,
     ).tag(config=True)
--- a/share/jupyterhub/templates/page.html
+++ b/share/jupyterhub/templates/page.html
@@ -28,7 +28,8 @@
   </div>
 {% endmacro %}
 <!DOCTYPE HTML>
-<html lang="en">
+<html lang="en"
+      data-bs-theme="{%- if dark_theme -%}dark{%- else -%}light{%- endif -%}">
   <head>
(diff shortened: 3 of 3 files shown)


benz0li:
Demo deployment: https://demo.jupyter.b-data.ch

benz0li:
@minrk I do not know how to fix the pre-commit.ci failure.

Please help out. Thank you.

benz0li:
@minrk @consideRatio Demo deployment in dark mode: https://demo.jupyter.b-data.ch

benz0li:
> Do you have any thoughts on where this is best documented?

[docs/source/howto/templates.md](https://github.com/jupyterhub/jupyterhub/blob/main/docs/source/howto/templates.md)

Subsection

## Simple dark theme

benz0li:
Renamed variable from `dark_mode` to `dark_theme` to be consistent with the names Bootstrap and JupyterHub use.

benz0li:
Converted to draft. Will finalise tomorrow.

benz0li:
Let me see, if I can find a fix for this:


https://github.com/user-attachments/assets/63446f89-51ee-4145-ae4b-72ca9f289243

benz0li:
Maybe https://github.com/jupyterhub/jupyterhub/blob/99b32dd372ae41c14473e543be4c302b80f613aa/share/jupyterhub/static/scss/page.scss#L15-L24

But I do not have the time to dig any deeper.

minrk:
This is awesome! After a bit of poking around, the nav hover you've pointed out is the main thing that stands out as needing a fix.

Implementation-wise, dark mode is more of a user preference than admin setting. What do you think of following the [bootstrap docs example](https://getbootstrap.com/docs/5.3/customize/color-modes/#javascript) to add a dark/light toggle?

The example has:

- default to respecting `prefers-color-scheme: dark`, which is nice
- picker
- persist choice in localStorage

minrk:
Sorry, I see you already mentioned a draft of a toggle. I do think a toggle is nice, but I would ideally add persistence to localStorage and respecting prefers-color-scheme as is done in the boostrap docs example. And probably use an icon like [circle-half-stroke](https://fontawesome.com/icons/circle-half-stroke?f=classic) for the button.

benz0li:
> Implementation-wise, dark mode is more of a user preference than admin setting.

Agreed.

> What do you think of following the [bootstrap docs example](https://getbootstrap.com/docs/5.3/customize/color-modes/#javascript) to add a dark/light toggle?

A toggle would be great.

benz0li:
Unfortunately, I do not have the time to dig any deeper.

I am not very familiar with web technologies. This would take up too much [of my] time.

minrk:
No problem, thanks for getting it started! Hopefully someone can find the time to shift this to a client-side choice, following the bootstrap docs. I may give it a go.

I think we'll also want to add a `btn-contrast` as [described here](https://github.com/twbs/bootstrap/issues/38748#issuecomment-2267293272) so we don't need to add template variables everywhere we have 'outline-dark', we can use 'outline-contrast' instead and get the right result.

benz0li:
Superseded by https://github.com/jupyterhub/jupyterhub/pull/4897.

PR #4881, opened 2024-08-26
Display Sphinx linkcheck output in a more readable format

manics:
I've copied https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/77e9df64f926582417ce8389cc55d58849ce4aff/ci/summarise-linkcheck-output#L1-L23 into an action:
https://github.com/manics/action-sphinx-linkcheck-summary

Example output:
image

Without this PR:
image

### DIFF
--- a/.github/workflows/test-docs.yml
+++ b/.github/workflows/test-docs.yml
@@ -81,10 +81,12 @@ jobs:
           cd docs
           make html
 
+      # Output broken and permanently redirected links in a readable format
       - name: check links
-        run: |
-          cd docs
-          make linkcheck
+        uses: manics/action-sphinx-linkcheck-summary@main
+        with:
+          docs-dir: docs
+          build-dir: docs/_build
 
       # make rediraffecheckdiff compares files for different changesets
       # these diff targets aren't always available


minrk:
That's awesome!

The CURC link in the gallery should be removed, since that's apparently been shut down.

The pyspark doc link is _wildly_ out of date, and no similar page seems to exist in the current docs. We could just remove the `pySpark configuration documentation` link sentence. We could link to [SparkConf](https://spark.apache.org/docs/3.5.2/api/python/reference/api/pyspark.SparkConf.html?highlight=sparkconf#pyspark.SparkConf) via intersphinx, or we could even remove the spark executor faq entry entirely, since I'm not sure how f that q is a'd these days.

minrk:
Removed the broken links in #4883

minrk (approved):

minrk:
@manics not merging because this is a draft, but it's not clear what more you want to do here. It seems great, already!

manics:
Thanks! I was going to tag the action and discuss whether it should be moved to the JupyterHub GitHub org, but given I haven't done anything I'll merge and return to it later.

PR #4897, opened 2024-08-28
add dark mode toggle

minrk:
continues #4878 but adds 

- toggle button in upper right instead of server setting
- system dark-mode is respected by default
- explicit setting is remembered in `localStorage["jupyterhub-bs-theme"]`
- `btn-contrast` classes are added to allow buttons to be dark-on-light and light-on-dark via css

cc @benz0li

https://github.com/user-attachments/assets/a1aa8180-304d-487b-bc12-05cca1c0adc0

### DIFF
--- a/docs/source/howto/templates.md
+++ b/docs/source/howto/templates.md
@@ -8,6 +8,15 @@ example, to be defined once and incorporated into all pages. By providing
 your own template(s), you can have complete control over JupyterHub's
 appearance.
 
+## Dark Theme
+
+By default, bootstrap pages have a white (light) background color.
+
+If you want to change JupyerHub to a darker color, you can set the configuration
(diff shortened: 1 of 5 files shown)


benz0li:
@minrk Perfect!

Many thanks.

benz0li:
@minrk https://jupyterhub.readthedocs.io/en/stable/ has a toggle, too. Handling seems a little different, though.

benz0li:
Cross reference:

* https://github.com/pydata/pydata-sphinx-theme/pull/540

benz0li:
@minrk I am happy with your solution, too.

It would be great if this could be merged soon.

minrk:
Thanks @benz0li!

PR #4904, opened 2024-09-12
Add `LD_LIBRARY_PATH` and `PYTHONHOME` to `Spawner.env_keep`

### LINKED ISSUE #4903: Shared library errors with relocated Python due to `LD_LIBRARY_PATH` env var not being inherited
### Bug description

When using JupyterHub with a relocated Python installation that was compiled using `--enable-shared`, child Python processes fail to start with errors like:

```
python: error while loading shared libraries: libpython3.11.so.1.0: cannot open shared object file: No such file or directory
```

This is because:
- such relocated Python installs rely upon the `LD_LIBRARY_PATH` env var being set
- for security reasons JupyterHub only passes an allow-list of env vars from the parent JupyterHub Python process to the child single-user server process, and the default value for `Spawner.env_keep` (the env vars to keep) doesn't currently include `LD_LIBRARY_PATH`. (From inspecting Git blame, it seems the `env_keep` list has (with the exception of one env var addition in #3128) not changed since it first landed in #52. )

This prevents JupyterHub from working out of the box on Heroku, since Heroku's Python installations are compiled in advance and then relocated, and these Python builds are now compiled with `--enable-shared` as of Python 3.10+ (https://github.com/heroku/heroku-buildpack-python/pull/1320).

This also appears to be the cause of:

- https://github.com/jupyter/notebook/issues/1290
- https://discourse.jupyter.org/t/jupyter-lab-with-enable-shared/17812

Whilst users can work around this issue by adding `c.Spawner.env_keep.append('LD_LIBRARY_PATH')` to their `jupyterhub_config.py`, this requires that they know about concepts like shared builds and `LD_LIBRARY_PATH`, and even if they do, requires extra setup steps.

As such, the better long-term solution would be to add `LD_LIBRARY_PATH` to the `env_keep` default list here:
https://github.com/jupyterhub/jupyterhub/blob/b13d3afa0fc5557be360c62bbcfab7597d3b28b4/jupyterhub/spawner.py#L703-L713

Adding `LD_LIBRARY_PATH` would be safe to do, since that env var doesn't contain secrets (given the purpose of the env var filtering feature is to prevent leaking of credentials etc). For bonus points, `PYTHONHOME` should probably also be added to the default `env_keep` too.

I will open a PR for this shortly :-)

### How to reproduce

1. Deploy JupyterHub 5.1.0 to Heroku, using Python 3.10 or newer (which are built in shared mode)
2. Set the web process command via a `Procfile` value of `web: jupyterhub --ip=0.0.0.0 --port=$PORT -f .jupyter/jupyterhub_config.py`
3. Try to interact with the notebook
4. Check app logs using `heroku logs`

#### Expected behaviour

Notebook can be used and no errors in the output of `heroku logs`.

#### Actual behaviour

Logs show:

```
2024-09-12T01:47:23.868320+00:00 app[web.1]: /app/.heroku/python/bin/python: error while loading shared libraries: libpython3.11.so.1.0: cannot open shared object file: No such file or directory
```

### Your personal set up

 - OS: Ubuntu 24.04
 - Version(s): JupyterHub 5.1.0, Python 3.11.10 (built in shared mode, and relocated)
 - env vars set: `LD_LIBRARY_PATH` (pointing at the relocated Python install's `lib` dir that contains the `libpython3.11.so.1.0` file)

edmorley:
For security reasons, only allow-listed env vars in the parent JupyterHub process are passed to the single-user server Python process. This allow-list is controlled by `Spawner.env_keep`, which by default includes common env vars that are:
(a) both necessary for the single-user server process to work,
(b) don't contain credentials or sensitive information that shouldn't be revealed to users of the Notebook.

However, this allow-list was missing the `LD_LIBRARY_PATH` env var, which causes shared library errors when using a relocated Python that has been compiled in shared mode (`--enable-shared`). This prevents JupyterHub from working out of the box on platforms like Heroku.

In addition, I've added `PYTHONHOME` too, since whilst it isn't needed for #4903, it's required in other scenarios and so makes sense to add it for completeness (particularly given `PYTHONPATH` is on the `env_keep` list already).

Fixes #4903.


manics:
We'll need to be careful before merging this. It's common for the singleuser environment to be independent of the hub environment, so we need to consider whether this could break something.

edmorley:
@manics Thank you for the reply! Would you be more comfortable about merging this, if I removed the `PYTHONHOME` addition? `LD_LIBRARY_PATH` itself should be very safe, since it's purely additive, whereas if someone has a broken `PYTHONHOME` (or a very strange setup with several Python installs) then I agree it could cause issues.

edmorley:
I've gone ahead and removed the `PYTHONHOME` addition for now, since it's not needed to fix #4903.

This leaves just the `LD_LIBRARY_PATH` addition, which is safe even if it's set to a bogus value, eg:

```
$ LD_LIBRARY_PATH='/invalid:/also-invalid' python --version
Python 3.12.6
```

manics:
I don't know the likely impact of setting `LD_LIBRARY_PATH`, but I've requested a review from @minrk who will know better.

minrk:
Thanks for the PR!

I take some issue with $LD_LIBRARY_PATH being called safe, since it is possibly the environment variable I've seen break things _the most_ (second only to $PATH). If the user and hub envs are different, it is _very_ unlikely that LD_LIBRARY_PATH should be inherited. It is definitely safe to point it to a _nonexistent_ directory, but if the directory exists, it can easily cause segfaults and other crashes. That said, setting LD_LIBRARY_PATH is consistent with $PATH, etc. which currently _assume_ the spawner has the same default env.

I'm honestly surprised the `env_keep` list is defined on Spawner and not LocalProcessSpawner, since none of these (except JUPYTERHUB_SINGLEUSER_APP) make sense for spawners other than LocalProcessSpawner.

@edmorley would you be able to move most of the list to LocalProcessSpawner below, and leave the default on the base `Spawner` class as just `["JUPYTERHUB_SINGLEUSER_APP"]`?

It should look something like this:
```python
class LocalProcessSpawner(Spawner):
    ...

    @default("env_keep")
    def _env_keep_default(self):
        return [
            "PATH",
            "PYTHONPATH",
            "CONDA_ROOT",
            "CONDA_DEFAULT_ENV",
            "VIRTUAL_ENV",
            "LANG",
            "LC_ALL",
            "LD_LIBRARY_PATH",
            "JUPYTERHUB_SINGLEUSER_APP",
        ]
```

edmorley:
Yeah sorry I was thinking along the lines of the typical values I see for `LD_LIBRARY_PATH` in our end-user environments, where bogus/invalid values are much more common than "valid but won't cause other issues" values, and there is also typically only one Python install. 

I also imagined that most JupyterHub users wouldn't be using relocated Python (it's pretty specific to some scenarios, such as ours), and when not using relocated Python it's not necessary to point `LD_LIBRARY_PATH` at the Python install. However, having never used JupyterHub myself (I opened this because of a Heroku support ticket), I'm having to extrapolate based on the weird edge cases we see :-)

But as you say, all of that is moot given that scenarios where `LD_LIBRARY_PATH` overlap with those where inheriting `PATH` would (eg the same python on path is likely used throughout) - so I agree in that framing adding `LD_LIBRARY_PATH` is fine.

Agree that refactor makes sense - I've moved/split the `env_keep` defaults to `LocalProcessSpawner` in https://github.com/jupyterhub/jupyterhub/pull/4904/commits/1b87e9c66889d1e133843828c76c6f2071f3e98d.

edmorley:
@minrk After that refactor, the one of the browser tests is now failing (the others pass):
`test_spawn_pending_progress`
https://github.com/jupyterhub/jupyterhub/actions/runs/10845790669/job/30097382092?pr=4904#step:11:96
https://github.com/jupyterhub/jupyterhub/blob/b13d3afa0fc5557be360c62bbcfab7597d3b28b4/jupyterhub/tests/browser/test_browser.py#L264-L307

This test fails on two separate GHA jobs (the main browser job, and the subdomains one), so I presume this is not an intermittent failure.

It looks like the server is failing to start, but there isn't enough output to say why.

My first thought was whether there was another spawner that subclassed the base class that is now missing an env var, however, I can't see one. And the tests appears to subclass `SimpleLocalProcessSpawner` which itself subclasses `LocalProcessSpawner`, so should still have the env_keep list?
https://github.com/jupyterhub/jupyterhub/blob/b13d3afa0fc5557be360c62bbcfab7597d3b28b4/jupyterhub/tests/mocking.py#L70-L176

Do you have any ideas? (I don't have permissions to retrigger a GHA job, so maybe that's a good place to start, even though intermittent seems unlike given the double failure.)

minrk:
Thanks for the update!

Some of the browser tests have started to become flaky, I think due to the async nature of browser events. I don't believe it's  related to this PR. I'll poke them a couple of times, but I think this is ready to go.

minrk:
I can confirm the failure is related to a playwright update and reproducible on main, so we can merge here.

edmorley:
Thank you!

Pulp-fiction-42:
> For security reasons, only allow-listed env vars in the parent JupyterHub process are passed to the single-user server Python process. This allow-list is controlled by `Spawner.env_keep`, which by default includes common env vars that are: (a) both necessary for the single-user server process to work, (b) don't contain credentials or sensitive information that shouldn't be revealed to users of the Notebook.
> 
> However, this allow-list was missing the `LD_LIBRARY_PATH` env var, which causes shared library errors when using a relocated Python that has been compiled in shared mode (`--enable-shared`). This prevents JupyterHub from working out of the box on platforms like Heroku.
> 
> In addition, I've added `PYTHONHOME` too, since whilst it isn't needed for #4903, it's required in other scenarios and so makes sense to add it for completeness (particularly given `PYTHONPATH` is on the `env_keep` list already).
> 
> Fixes #4903.

sorry to bothering, I have encounter a problem that similar but I can't fix it:
when I try to
`import torch`
I encounter the problem:
```
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
/tmp/ipykernel_3237544/4265195184.py in <cell line: 1>()
----> 1 import torch

~/.conda/envs/dnabert2/lib/python3.8/site-packages/torch/__init__.py in <module>
    190     if USE_GLOBAL_DEPS:
    191         _load_global_deps()
--> 192     from torch._C import *  # noqa: F403
    193 
    194 # Appease the type checker; ordinarily this binding is inserted by the

ImportError: /usr/lib64/libm.so.6: version `GLIBC_2.29' not found (required by /home/share/huadjyin/home/lihuyang/.conda/envs/dnabert2/lib/python3.8/site-packages/torch/lib/libtorch_cpu.so)
```
and I am working on a cluster, so there is a .so file satisfy the GLIBC version
`/home/HPCBase/compilers/gcc/12.3.1/libc/lib/libm.so.6`
I export this path file into .bashrc and it worked when I use Python interpreter to import torch in terminal shell
but in Jupyter Notebook, Neither adding environment variables to kernel.json nor adding the following code to jupyter_notebook_config.py works:
```
import os
c = get_config()
os.environ['LD_PRELOAD'] = '/home/HPCBase/compilers/gcc/12.3.1/libc/lib/libm.so.6'
c.Spawner.env.update({'LD_PRELOAD': os.environ['LD_PRELOAD']})
```
Why I can't just load the $LD_PRELOAD into Jupyter Notebook???

PR #4913, opened 2024-09-21
Fix incorrect rounding function

emmanuel-ferdman:
# PR Summary
The [`math`](https://docs.python.org/3/library/math.html) module in Python does not include a round function. Instead, the built-in Python [`round`](https://docs.python.org/3/library/functions.html#round) function should be used for rounding values.


minrk:
Thanks! Amazing this hasn't come up sooner in the six years since it was introduced in #1709. I guess it can't happen with the default retry range.

PR #4930, opened 2024-10-18
Abort jupyterhub startup only if managed services fail

### LINKED ISSUE #4929: JupyterHub startup aborted by failure to connect to an external service (behavior change between jh 4.1.5 and 5.0.0-5.2.0)
In JupyterHub version 4.1.5, failures like not being able to connect to an external service hasn't led to exiting jupyterhub, but does so in jupyterhub 5.2.0.

I think this could be seen as a regression, but possibly also a fix. I lean to think its a regression and an unwanted change, because I think jupyterhub should expect to startup independently of an external service, otherwise for example, what if the external service depends on jupyterhub to be available to startup itself?

I've not tried what happens in jupyterhub 5.2.0 if we have an external service initially accessible during startup, which we then loose contact with - does that also lead to an error and stopping jupyterhub?

## JupyterHub 5.0.0b1 - 5.2.0 behavior

```
[E 2024-10-18 10:18:09.522 JupyterHub app:3652] Cannot connect to external service binder at http://binderhub:8090. Is it running?
[D 2024-10-18 10:18:09.522 JupyterHub application:1060] Exiting application: jupyterhub
```

## JupyterHub 4.1.5 behavior

```
kubectl logs -n staging hub-74b7569bcb-8td2w -c hub | grep binderhub 

[I 2024-10-04 07:37:39.514 JupyterHub app:3224] Adding external service binder at http://binderhub:8090
[E 2024-10-04 07:37:40.531 JupyterHub app:3246] Cannot connect to external service binder at http://binderhub:8090. Is it running?
[W 2024-10-04 07:38:41.631 JupyterHub app:2508] Cannot connect to external service binder at http://binderhub:8090
[W 2024-10-04 07:39:41.542 JupyterHub app:2508] Cannot connect to external service binder at http://binderhub:8090
[W 2024-10-04 07:40:41.450 JupyterHub app:2508] Cannot connect to external service binder at http://binderhub:8090
[W 2024-10-04 07:41:41.450 JupyterHub app:2508] Cannot connect to external service binder at http://binderhub:8090
```

consideRatio:
Fixes #4929 which includes misc discussion about this. I think this is a regression, where desired behavior is that managed services specifically is ensured to startup properly, while external services failing to be accessible is just a warning kind of event during startup and later.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -3654,7 +3654,6 @@ class JupyterHub(Application):
                         "Cannot connect to %s service %s",
                         service_name,
                         service.kind,
-                        exc_info=True,
                     )
                 else:
                     self.log.warning(
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -3640,7 +3640,7 @@ class JupyterHub(Application):
                     if service.managed:
                         status = await service.spawner.poll()
                         if status is not None:
-                            self.log.error(
+                            self.log.critical(
                                 "Service %s exited with status %s",
                                 service_name,
                                 status,
@@ -3649,12 +3649,20 @@ class JupyterHub(Application):
                 else:
                     return True
             else:
-                self.log.error(
-                    "Cannot connect to %s service %s at %s. Is it running?",
-                    service.kind,
-                    service_name,
-                    service.url,
-                )
+                if service.managed:
+                    self.log.critical(
+                        "Cannot connect to %s service %s",
+                        service_name,
+                        service.kind,
+                        exc_info=True,
+                    )
+                else:
+                    self.log.warning(
+                        "Cannot connect to %s service %s at %s. Is it running?",
+                        service.kind,
+                        service_name,
+                        service.url,
+                    )
                 return False
         return True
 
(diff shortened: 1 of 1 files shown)


jupyterhub/app.py
```diff
@@ -3649,12 +3649,20 @@ async def start_service(
...
-                    service.kind,
-                    service_name,
-                    service.url,
-                )
+                if service.managed:
+                    self.log.critical(
+                        "Cannot connect to %s service %s",
+                        service_name,
+                        service.kind,
+                        exc_info=True,
```
consideRatio, line 3657: exc_info=True, not sure if this makes sense here or what it does. I ended up doing a copy paste for this.

> If exc_info does not evaluate as false, it causes exception information to be added to the logging message. If an exception tuple (in the format returned by [sys.exc_info()](https://docs.python.org/3/library/sys.html#sys.exc_info)) or an exception instance is provided, it is used; otherwise, [sys.exc_info()](https://docs.python.org/3/library/sys.html#sys.exc_info) is called to get the exception information.

consideRatio:
```suggestion
```

I suspect it doesn't make sense here.

minrk: I think you're right. This is an `else` bock, which means no error. There shouldn't be exc_info to log here.

jupyterhub/app.py
```diff
@@ -3745,18 +3753,8 @@ async def start(self):
...
-                else:
-                    # Only warn for database-based service, so that admin can connect
-                    # to hub to remove the service.
-                    self.log.error(
-                        "Failed to reach externally managed service %s",
-                        service_name,
-                        exc_info=True,
-                    )
+            if not service_ready and service.managed:
+                self.exit(1)
```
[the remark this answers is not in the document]
minrk: I think it's maybe better to keep the exit condition here in `start` where you have it.

PR #4936, opened 2024-10-21
enable PKCE for internal OAuth

minrk:
see https://github.com/jupyterhub/oauthenticator/pull/765 for implementing PKCE in OAuthenticator. This does the same in jupyterhub's oauth clients and adds what is needed for the provider side (oauthlib implements the verification, we only need to implement the db persistence of the challenge)

PKCE is currently optional for backward-compatibility

Since this adds db columns, this means a major version bump according to #3982

TODO:

- [ ] opt-in config to _enforce_ PKCE
- [ ] tests
  - [ ] failed PKCE validation
  - [ ] missing PKCE works when not required
  - [ ] missing PKCE doesn't work when PKCE is required
- [ ] explore backward-compatibility implications in HubOAuth defaults

It would be nice for PKCE to happen by default, but we need to make sure that existing code will either transparently send both the challenge and the verifier or neither, because e.g. sending the challenge but not the verifier will cause every oauth to fail. Sending neither is still okay as long as `require_pkce` is not enabled, which will be opt-in.

### DIFF
--- a/jupyterhub/alembic/versions/afd65840b69e_pkce.py
+++ b/jupyterhub/alembic/versions/afd65840b69e_pkce.py
@@ -1,0 +1,40 @@
+"""pkce
+
+Revision ID: afd65840b69e
+Revises: 4621fec11365
+Create Date: 2024-10-21 11:14:43.079782
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = 'afd65840b69e'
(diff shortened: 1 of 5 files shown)


minrk:
I'd like to get 5.4 out, then come back to this and require Python 3.10 for jupyterhub 6

PR #4946, opened 2024-11-08
add traitlets default import to auth allow_all override example for completeness

kellyrowland:
adds `from traitlets import default` to example code for overriding authenticator `allow_all` such that the example can be used "out of the box" in a config file.


minrk:
Thanks!

PR #4948, opened 2024-11-12
Use intersphinx-registry to keep intersphinx URLs up to date.

Carreau:
This allows to update the intersphinx url in a single location when those move, an make it a tiny-bit easier to add existing packages than having to figure out where their docs are.


manics (approved):
https://github.com/Quansight-Labs/intersphinx_registry/blob/main/intersphinx_registry/registry.json
Looks fine to me

Issue #4949, opened 2024-11-13
It looks like a linter messed up the interpolation of `implicit_spawn_seconds` in `not_running.html`


millenniumhand:
### Bug description

It looks like commit 75c947b changed

```javascript
var implicit_spawn_seconds = {{ implicit_spawn_seconds }};
```

to

```javascript
var implicit_spawn_seconds = {
  {
    implicit_spawn_seconds
  }
};
```

and so the variable is not interpolated into the template and the automatic spawning of the server is not happening.

minrk:
Closed by #4950

PR #4950, opened 2024-11-13
Fixed code formatting for implicit_spawn_seconds (#4949)

millenniumhand:
A linter appears to have broken the templating in of `implicit_wait_seconds`. This fixes the code formatting so the interpolation can happen correctly.

### DIFF
--- a/share/jupyterhub/templates/not_running.html
+++ b/share/jupyterhub/templates/not_running.html
@@ -56,6 +56,7 @@
 {% block script %}
   {{ super () }}
   {% if implicit_spawn_seconds %}
+    {# djlint:off #}
     <script type="text/javascript">
       var spawn_url = "{{ spawn_url }}";
       var implicit_spawn_seconds = {{ implicit_spawn_seconds }};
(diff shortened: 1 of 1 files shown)


share/jupyterhub/templates/not_running.html
```diff
@@ -58,11 +58,7 @@ <h1>
   {% if implicit_spawn_seconds %}
     <script type="text/javascript">
       var spawn_url = "{{ spawn_url }}";
-      var implicit_spawn_seconds = {
-        {
-          implicit_spawn_seconds
-        }
-      };
+      var implicit_spawn_seconds = {{ implicit_spawn_seconds }};
```
consideRatio: It seems like that since this was not put in quotes, the linter reacted on this but not on `spawn_url`.

consideRatio:
Clear confirmation that an autoformatter indeed changed this, nice catch! We need to disable the autoformatter from messing this up as well.

manics:
Try using `{# djlint: off #}` as in
https://github.com/jupyterhub/jupyterhub/pull/4796/files

If it doesn't work (it didn't in the above PR) add an exclusion to https://github.com/jupyterhub/jupyterhub/blob/a8500a31a9bc77a9829e903693462a7da38fee52/.pre-commit-config.yaml#L45

PR #4953, opened 2024-11-19
jupyterhub image: use ubuntu:24.04 and node 22

consideRatio:


minrk:
Switching to debian is fine, I suppose.

I _really_ want to get building these images out of this repo (https://github.com/jupyterhub/team-compass/issues/474), since we cannot make any changes to these images now without publishing a jupyterhub package release, and changes to these images definitely shouldn't be tied to jupyterhub package versioning at all.

minrk:
btw, whatever seems simplest for you to unblock this is fine, whether it's switching to debian or updating ubuntu, or delaying these upgrades.

I don't feel any great pressure to update these immediately either if updating is a pain, since ubuntu 22.04 has 3 years left of support, while node 20 has 17 months.

I'm starting to think we shouldn't be publishing a `jupyterhub/jupyterhub` image at all. It's not a particularly useful image. That's a separate discussion, though.

manics:
I've got a version of this that builds in https://github.com/jupyterhub/jupyterhub/compare/main...manics:jupyterhub:docker-base

consideRatio:
Nice @manics!! I'll close this PR in favor of @your work!

PR #4963, opened 2024-12-03
allow launch with expired auth if initiated by another user

minrk:
Without this, when using `refresh_pre_spawn`, if the target user doesn't have fresh auth, it is impossible for other users to launch until the original user logs in to get fresh credentials.

Cases where this might come up:

1. 'fake' users, such as those 2i2c uses for health checks, which cannot ever have fresh auth
2. pre-launch of user servers before their first login (they won't have credentials, but that may be okay in some circumstances)

Choices:

1. allow launch without refreshing auth with a warning, as long as the requesting user is not the owner (this PR) (note: warning is only in logs, since there isn't really a mechanism to show it to the user)
3. don't allow launching at all by anyone until user logs in again (status quo)
4. expose this as an option somehow
5. handle it at the Authenticator level (e.g. allow fully empty `auth_state` to pass refresh, i.e. don't force a _first_ login for any user via `refresh_user`)

cc @GeorgianaElena this is the cause of the health check failure in https://github.com/2i2c-org/infrastructure/pull/5217

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -1009,9 +1009,18 @@ class BaseHandler(RequestHandler):
         if self.authenticator.refresh_pre_spawn:
             auth_user = await self.refresh_auth(user, force=True)
             if auth_user is None:
-                raise web.HTTPError(
-                    403, "auth has expired for %s, login again", user.name
-                )
+                if (
+                    isinstance(self.current_user, User)
+                    and self.current_user.id == user.id
+                ):
+                    raise web.HTTPError(
+                        403, "auth has expired for %s, login again", user.name
+                    )
+                else:
+                    # allow with a warning if it's being launched by someone else
+                    self.log.warning(
+                        f"Allowing launch of {user.name}/{server_name} with expired auth by {self.current_user.kind}:{self.current_user.name}"
+                    )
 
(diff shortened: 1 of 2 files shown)


manics:
What's the use-case for needing `refresh_pre_spawn = True`, but not always needing it? If it's for a bot-user (for health checks) could this be a configurable somewhere so we can disable it on a per-user basis?

minrk:
We could handle this at the OAuthenticator level with either a list of skippable users, or explicitly skipping falsy auth_state (I.e. users who have never logged in)

GeorgianaElena (commented):
I think the choice that makes the most sense to me is the per-user skiplist as it seems to be more specific to the bot test users situation and is more self-contained.

My intuition is that the other options, allowing an empty auth_state or allowing users to start other users' servers that have expired tokens are more general and implicitly might  have a bigger impact and/or open some new security paths

minrk:
Makes sense! Probably do that on OAuthenticator directly?

manics:
OAuthenticator sounds good. As long as the property name is chosen carefully it'll be easy to move it to JupyterHub in future if necessary.

minrk:
closing in favor of https://github.com/jupyterhub/oauthenticator/pull/780

PR #4966, opened 2024-12-17
Allow CORS requests to /hub/api by default

agoose77:
## Context
In <https://mystmd.org>, we have a "launch" button that builds a URL for interacting with deployed content on a BinderHub or JupyterHub. The user can provide a URL via a text-input field for each provider type:

image

We would like to simplify the UI by having a single text-input field, which requires that we can interrogate the given URL to determine what kind of provider it is. This will then enable us to build the proper links.

Unfortunately, in a browser context, the response from the hubs that I've seen in the wild does not include the necessary `Access-Control-Allow-Origin` header required by the browser:

<details><summary>Browser</summary>

```shell
$ await fetch("https://showcase.2i2c.cloud/hub/api/", {redirect: "follow"}) 

Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://showcase.2i2c.cloud/hub/api/. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing). Status code: 404.
```

</details>

<details><summary>Shell</summary>

```shell
$ curl -v https://showcase.2i2c.cloud/hub/api/
* Host showcase.2i2c.cloud:443 was resolved.
* IPv6: (none)
* IPv4: 52.26.150.118, 34.214.185.121
*   Trying 52.26.150.118:443...
* ALPN: curl offers h2,http/1.1
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
* TLSv1.3 (IN), TLS handshake, Certificate (11):
* TLSv1.3 (IN), TLS handshake, CERT verify (15):
* TLSv1.3 (IN), TLS handshake, Finished (20):
* TLSv1.3 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.3 (OUT), TLS handshake, Finished (20):
* SSL connection using TLSv1.3 / TLS_AES_256_GCM_SHA384 / x25519 / RSASSA-PSS
* ALPN: server accepted h2
* Server certificate:
*  subject: CN=showcase.2i2c.cloud
*  start date: Dec 17 04:59:19 2024 GMT
*  expire date: Mar 17 04:59:18 2025 GMT
*  subjectAltName: host "showcase.2i2c.cloud" matched cert's "showcase.2i2c.cloud"
*  issuer: C=US; O=Let's Encrypt; CN=R10
*  SSL certificate verify ok.
*   Certificate level 0: Public key type RSA (2048/112 Bits/secBits), signed using sha256WithRSAEncryption
*   Certificate level 1: Public key type RSA (2048/112 Bits/secBits), signed using sha256WithRSAEncryption
*   Certificate level 2: Public key type RSA (4096/152 Bits/secBits), signed using sha256WithRSAEncryption
* Connected to showcase.2i2c.cloud (52.26.150.118) port 443
* using HTTP/2
* [HTTP/2] [1] OPENED stream for https://showcase.2i2c.cloud/hub/api/
* [HTTP/2] [1] [:method: GET]
* [HTTP/2] [1] [:scheme: https]
* [HTTP/2] [1] [:authority: showcase.2i2c.cloud]
* [HTTP/2] [1] [:path: /hub/api/]
* [HTTP/2] [1] [user-agent: curl/8.11.0]
* [HTTP/2] [1] [accept: */*]
> GET /hub/api/ HTTP/2
> Host: showcase.2i2c.cloud
> User-Agent: curl/8.11.0
> Accept: */*
>
* Request completely sent off
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
< HTTP/2 200
< date: Tue, 17 Dec 2024 15:03:41 GMT
< content-type: application/json
< content-length: 20
< x-jupyterhub-version: 4.1.5
< access-control-allow-headers: accept, content-type, authorization
< content-security-policy: frame-ancestors 'none'; report-uri /hub/security/csp-report; default-src 'none'
< etag: "7be5fa0f046b27260a0e4da97272cb7eaf74e88c"
< strict-transport-security: max-age=31536000; includeSubDomains
<
* Connection #0 to host showcase.2i2c.cloud left intact
{"version": "4.1.5"}
```

</details> 

## Fix
This PR sets the `access-control-allow-origin` header to `*` for `GET /hub/api/`, which permits _any_ origin to make requests of the given hub's endpoint. 

> [!NOTE]
> The _original_ version of this PR was a draft that changed this heading for _all_ requests. Following some discussion, I've marked as ready-for-review, and scoped the change to a single API-version end-point.

### DIFF
--- a/jupyterhub/apihandlers/hub.py
+++ b/jupyterhub/apihandlers/hub.py
@@ -58,15 +58,10 @@ class RootAPIHandler(APIHandler):
 
         Also responsible for setting content-type header
         """
-        old_headers = self.settings.get("headers", {})
-        new_headers = {**old_headers}
-        new_headers.setdefault('Access-Control-Allow-Origin', '*')
-        try:
-            self.settings['headers'] = new_headers
-            super().set_default_headers()
-        finally:
-            self.settings['headers'] = old_headers
-
+        if 'Access-Control-Allow-Origin' not in self.settings.get("headers", {}):
+            # allow CORS requests to this endpoint by default
+            self.set_header('Access-Control-Allow-Origin', '*')
+        super().set_default_headers()
     def check_xsrf_cookie(self):
         return
 
--- a/jupyterhub/apihandlers/hub.py
+++ b/jupyterhub/apihandlers/hub.py
@@ -6,7 +6,6 @@ import json
 import sys
 
 from tornado import web
-from tornado.httputil import HTTPHeaders
 
 from .._version import __version__
 from ..scopes import needs_scope
@@ -53,13 +52,12 @@ class ShutdownAPIHandler(APIHandler):
 
 
 class RootAPIHandler(APIHandler):
-
     def set_default_headers(self):
         """
         Set any headers passed as tornado_settings['headers'].
 
         Also responsible for setting content-type header
-        """    
+        """
         old_headers = self.settings.get("headers", {})
         new_headers = {**old_headers}
         new_headers.setdefault('Access-Control-Allow-Origin', '*')
--- a/jupyterhub/apihandlers/hub.py
+++ b/jupyterhub/apihandlers/hub.py
@@ -6,6 +6,7 @@ import json
 import sys
 
 from tornado import web
+from tornado.httputil import HTTPHeaders
 
 from .._version import __version__
 from ..scopes import needs_scope
@@ -52,6 +53,34 @@ class ShutdownAPIHandler(APIHandler):
 
 
 class RootAPIHandler(APIHandler):
+
+    def set_default_headers(self):
+        """
+        Set any headers passed as tornado_settings['headers'].
+
+        Also responsible for setting content-type header
+        """
+        # wrap in HTTPHeaders for case-insensitivity
+        headers = HTTPHeaders(self.settings.get('headers', {}))
+        headers.setdefault("X-JupyterHub-Version", __version__)
+
+        for header_name, header_content in headers.items():
+            self.set_header(header_name, header_content)
+
+        if 'Access-Control-Allow-Headers' not in headers:
+            self.set_header(
+                'Access-Control-Allow-Headers', 'accept, content-type, authorization'
+            )
+        if 'Content-Security-Policy' not in headers:
+            self.set_header('Content-Security-Policy', self.content_security_policy)
+        self.set_header('Content-Type', self.get_content_type())
+
+        # Allow any origin to query the root handler (CORS)
+        if 'Access-Control-Allow-Origin' not in headers:
+            self.set_header(
+                'Access-Control-Allow-Origin', '*'
+            )
+
     def check_xsrf_cookie(self):
         return
 
--- a/jupyterhub/apihandlers/hub.py
+++ b/jupyterhub/apihandlers/hub.py
@@ -59,27 +59,15 @@ class RootAPIHandler(APIHandler):
         Set any headers passed as tornado_settings['headers'].
 
         Also responsible for setting content-type header
-        """
-        # wrap in HTTPHeaders for case-insensitivity
-        headers = HTTPHeaders(self.settings.get('headers', {}))
-        headers.setdefault("X-JupyterHub-Version", __version__)
-
-        for header_name, header_content in headers.items():
-            self.set_header(header_name, header_content)
-
-        if 'Access-Control-Allow-Headers' not in headers:
-            self.set_header(
-                'Access-Control-Allow-Headers', 'accept, content-type, authorization'
-            )
-        if 'Content-Security-Policy' not in headers:
-            self.set_header('Content-Security-Policy', self.content_security_policy)
-        self.set_header('Content-Type', self.get_content_type())
-
-        # Allow any origin to query the root handler (CORS)
-        if 'Access-Control-Allow-Origin' not in headers:
-            self.set_header(
-                'Access-Control-Allow-Origin', '*'
-            )
+        """    
+        old_headers = self.settings.get("headers", {})
+        new_headers = {**old_headers}
+        new_headers.setdefault('Access-Control-Allow-Origin', '*')
+        try:
+            self.settings['headers'] = new_headers
+            super().set_default_headers()
+        finally:
+            self.settings['headers'] = old_headers
 
     def check_xsrf_cookie(self):
         return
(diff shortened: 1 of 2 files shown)


manics:
Have you tried setting some custom headers with `tornado_settings.headers`?
https://jupyterhub.readthedocs.io/en/5.2.1/reference/api/app.html#jupyterhub.app.JupyterHub.tornado_settings

agoose77:
@manics thanks for the speedy reply! I should clarify the intention of this PR!

Whilst there is an escape hatch for fixing this per-deployment, I would like to change this setting across all Hubs (by default).

The goal of the launch feature in Jupyter Book 2 / the MyST engine is to enable re-use and sharing of content. This includes "bring your own hub", where a user can pull in content from a published website into their own hub.

Whilst it is technically possible to change the CORS headers per-hub, using `tornado_settings`, that is not the intention of this PR; Requiring Hub deployers to opt-in to cross-origin requests will necessarily mean that only a subset of hubs in the wild actually do so. And that means that any mechanism to try and detect hubs in the wild will be flaky and unreliable for users.[^bad-cors]

Given that access to the non-authenticated portion of a Hub is useful, I'm instead suggesting that we should be setting a policy for `Access-Control-Allow-Origin` (wildcard) for public API endpoints. That's also why this PR is a draft -- we probably don't want to set wildcard for all routes.

[^bad-cors]: It is not possible (to my knowledge) to disinguish between failed CORS and other network failures.

manics:
> Requiring Hub deployers to opt-in to cross-origin requests will necessarily mean that only a subset of hubs in the wild actually do so. And that means that any mechanism to try and detect hubs in the wild will be flaky and unreliable for users.

I don't see that as a problem. If the aim is to have a unique way to determine a hub is a hub through a cross-site-request I think we should make that explicit, and not automatically impose it on everyone with no way to disable it.

yuvipanda:
Thank you for working on this, @agoose77! Feature detection by hitting URLs is definitely better than having to hardcode them (something I'm thinking of for repo2docker), so I think it'll be good for the hub to have that.

> Are you suggesting that we instead have a singular route that is scoped for my particular case of testing whether a Hub is a Hub?

I think this would be the way to approach this. CORS is hard to think through, and I can never fully trust myself with it. I think it's easier to start off with one single endpoint than all the handlers, and allow it to be an opt out.

yuvipanda:
The close analogue we have here is the metrics endpoint for prometheus servers, and authentication for it is opt-in via the `authenticate_prometheus` traitlet. It defaults to requiring authentication, so not available for public use.

This could be a different handler, with extremely minimal information exposed. And it could be opt-out instead.

minrk:
> We would like to simplify the UI by having a single text-input field, which requires that we can interrogate the given URL to determine what kind of provider it is. 

I think it totally makes sense to have an unauthenticated 'what are you?' URL which is open to everything by default, without changing the rest of the security situation. Right now, we have `GET /hub/api/` which reports the JupyterHub version and is not authenticated. Currently, it only returns:

```json
{"version": "5.2.1"}
```

(it is currently covered by the Content Security Policy, so this doesn't work today without a patch like this one). This seems like it might be the endpoint you are after, so what about applying the same change you have here to [RootAPIHandler](https://github.com/jupyterhub/jupyterhub/blob/0d57ce2e337a13d69bb302e81304701b712af4e9/jupyterhub/apihandlers/hub.py#L54) alone, would that serve your purpose? If we keep the check for if it's defined, deployments with custom CSP could still prevent this sniff, or we could make it unconditional if that makes more sense to folks.

agoose77:
@manics @yuvipanda @minrk thank you all for chiming in. I intended to reply here before breaking for the Christmas holidays, but in the end I took the time off. 

I don't think it was clear enough that my intention was never to treat this PR's implementation as final and definitive. I tried to clarify that in the description, but it wasn't visible enough.

With some time to consider the valuable perspectives here, I will re-work the _actual implementation_ of this PR to narrow the scope and open up _only_ `GET /hub/api/` (which is the original route that I expected to query). 

I also think it would be helpful for me to be explicit about which "hats" I'm wearing here. I am authoring this PR with the intention of simplifying the UX for Jupyter Book and MyST-MD users (my Jupyter Book hat) to launch into interactive compute experiences (my 2i2c hat). This can only be achieved if there is a community-wide standard, rather than an ad-hoc per-deployment feature.

I hope that the narrower scope of this PR makes it much easier to approve, but do let me know if I can do more here to support the discussion!

minrk (commented):
Thanks! I think this makes total sense. One suggestion about exactly how it's implemented, but I think we should do this.

jupyterhub/apihandlers/hub.py
```diff
@@ -52,6 +53,31 @@ def post(self):
...
+            self.set_header(
+                'Access-Control-Allow-Headers', 'accept, content-type, authorization'
+            )
+        if 'Content-Security-Policy' not in headers:
+            self.set_header('Content-Security-Policy', self.content_security_policy)
+        self.set_header('Content-Type', self.get_content_type())
+
+        # Allow any origin to query the root handler (CORS)
+        if 'Access-Control-Allow-Origin' not in headers:
+            self.set_header('Access-Control-Allow-Origin', '*')
```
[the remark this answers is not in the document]
agoose77: I don't love the duplication of code -- @yuvipanda noticed it too. I opted to duplicate here in favour of forcibly overwriting the user headers. 

How does the latest patch look? 39562af just patches the headers in settings. It's not idempotent -- if the user didn't specify headers, there will now be an empty dict, but I think that's fine.

jupyterhub/apihandlers/hub.py
```diff
@@ -52,6 +52,21 @@ def post(self):
...
+        """
+        old_headers = self.settings.get("headers", {})
+        new_headers = {**old_headers}
+        new_headers.setdefault('Access-Control-Allow-Origin', '*')
+        try:
+            self.settings['headers'] = new_headers
+            super().set_default_headers()
+        finally:
+            self.settings['headers'] = old_headers
+
```
minrk, line 69: Sorry, I was overthinking it since I thought this header had a default value, but it doesn't. Since `Access-Control-Allow-Origin` isn't set at all by default, we can call `set_header` if it's not present without any stateful patching and then move on to unmodified super:

```suggestion
        if 'Access-Control-Allow-Origin' not in self.settings.get("headers", {}):
            # allow CORS requests to this endpoint by default
            self.set_header('Access-Control-Allow-Origin', '*')
        super().set_default_headers()
```

agoose77: Ah, that's a nicer way to do this that hadn't occurred to me!

minrk:
Thanks @agoose77!

Issue #4833, opened 2024-06-04
Allow configuration of bucket sizes in Prometheus metric for server spawn times


srikanthchelluri:
### Proposed change

The hub exposes a Prometheus histogram metric `jupyterhub_server_spawn_duration_seconds`. The buckets sizes are explicitly set when `SERVER_SPAWN_DURATION_SECONDS` is defined: https://github.com/jupyterhub/jupyterhub/blob/main/jupyterhub/metrics.py

Users who deploy JupyterHub may want to define their own bucket sizes - e.g. bucket sizes that are more granular. We should allow users to override the bucket sizes which are currently set

### Alternative options

Ultimately, if we want the histogram metric to have different bucket sizes, it must expose those itself

### Who would use this feature?

JupyterHub already allows users to define custom spawn hooks - and some of those hooks may significantly change the time it takes for servers to spawn. Regardless of why spawn times may differ, though, users may simply have deployments that have more or less variable spawn times than what the current bucket sizes would help visualize

### Possible solution

We already have precedent for overriding the metrics name prefix via the `JUPYTERHUB_METRICS_PREFIX` environment variable. We can allow users to override bucket sizes with another environment variable, e.g. `JUPYTERHUB_SERVER_SPAWN_DURATION_SECONDS_BUCKET_SIZES`

minrk:
Makes sense! Your proposal sounds good if you want to try a PR.

minrk:
Fixed by #4967

PR #4967, opened 2024-12-22
Allow configuration of bucket sizes in metrics - #4833

kireetb:
First PR. 
Picked up a simple change.
Please review.


kireetb:
@manics 
Thank you for your response.

My apologies i thought it was a simple change and missed on testing with hub.extraEnv in values file. 

I took your advice and implemented it using traitlets. 
Tested it via changes in values.yaml. Attached Screenshot.
values-yaml

Please let me know your thoughts.

minrk:
Switching to traitlets config here will be more involved than this because doing so means deferring instantiation of metrics until after the Application object loads config files. That in turn would mean they can't be defined top-level in the module, and also can't be imported at the top level in other modules, so they would need to come through `settings` or some such.

In the meantime, I think it's okay to do this. I'd use a simpler, more env-friendly serialization like `1,2,3` rather than writing custom parsing for a repr of a Python list. Something like:

```python
_env_buckets = os.environ.get("JUPYTERHUB_SERVER_SPAWN_DURATION_SECONDS_BUCKETS", "").strip()
if _env_buckets:
    spawn_duration_buckets = [float(_s) for _s in _env_buckets.split(",")]
else:
    spawn_duration_buckets = [default...]
```

kireetb:
@minrk 
Thank you for your response

avoiding traitlets for now
using simple serialization "1,2,3" as suggested
added corresponding documentation

PR #4975, opened 2025-01-02
Standard formatting in LICENSE

SamuelMarks:
Happy New Year!

PS: You might also want to update the footer of your doc page https://jupyterhub.readthedocs.io/en/latest/howto/configuration/config-proxy.html :
>  © Copyright 2024, Project Jupyter Contributors.


manics:
There's already a license file
https://github.com/jupyterhub/jupyterhub/blob/main/COPYING.md
GitHub recognises it:
https://github.com/jupyterhub/jupyterhub?tab=License-1-ov-file#readme

SamuelMarks:
Screenshot 2025-01-03 at 8 56 49 AM

---

vs what it could be

---

Screenshot 2025-01-03 at 8 57 40 AM

minrk:
Yeah, we shouldn't add a duplicate license file, but we should probably move the explanations of "Project Jupyter contributors" out of the license file to the docs, so the file contents are standard.

PR #4978, opened 2025-01-08
Missing breaking change in 5.0.0 changelog: URL tokens

manics:
This breaking change was missing from the 5.0.0 changelog, because it doesn't seem to be associated with a PR:
https://github.com/jupyterhub/jupyterhub/commit/b319b58a2f87e87f4f3af2ea2042e045c19ff866


minrk:
Right, it was in the forward-port of the security PR for GHSA-7r3h-4ph8-w38g, which touched a lot of things. The PR merge commit is [here](https://github.com/jupyterhub/jupyterhub/commit/4a83cddb8e9b2de502e0a666e1b057559dc1d402). The 4.x PR introduced the option, leaving it opt-out to avoid breaking things, while 5.0 opted in to better security by default. Looking back, I probably should have made the toggle in its own PR instead of its own commit for better visibility.

PR #4979, opened 2025-01-09
Allow custom login input validation

tlvu:
This is pure HTML, no CSS or JavaScript used.

Tested on Firefox and Chromium.

For example, with this change, if we set

```
{% block username_input_attribs %}pattern="[a-z0-9]+"
      placeholder="do not use email address, use your username"{% endblock username_input_attribs %}
```

We will get the following generated code

```
    <input
      id="username_input"
      type="text"
      autocapitalize="off"
      autocorrect="off"
      autocomplete="username"
      class="form-control"
      name="username"
      val=""
      tabindex="1"
      autofocus="autofocus"

      pattern="[a-z0-9]+"
      placeholder="do not use email address, use your username"
    />
```

And this sample behavior: only `[a-z0-9]` character are allowed with the placeholder text "do not use email address, use your username", see animated Gif below:
jupyterhub-login-validation

A full override of the entire "label" and "input" elements for the username and password fields is also possible for those who require more than just additional input attributes.

### DIFF
--- a/share/jupyterhub/templates/login.html
+++ b/share/jupyterhub/templates/login.html
@@ -43,13 +43,19 @@
                      class="form-control"
                      name="username"
                      val="{{ username }}"
-                     autofocus="autofocus" />
+                     autofocus="autofocus"
+                     {# allow adding extra HTML Input Attributes (pattern, placeholder, title, minlenght, maxlenght, ...) #}
+                     {% block username_input_attribs %}{% endblock username_input_attribs %}
+                     />
               <label for='password_input'>Password:</label>
               <input type="password"
                      class="form-control"
                      autocomplete="current-password"
                      name="password"
-                     id="password_input" />
+                     id="password_input"
+                     {# allow adding extra HTML Input Attributes (pattern, placeholder, title, minlenght, maxlenght, ...) #}
+                     {% block password_input_attribs %}{% endblock password_input_attribs %}
+                     />
               {% if authenticator.request_otp %}
                 <label for='otp_input'>{{ authenticator.otp_prompt }}</label>
                 <input class="form-control"
--- a/share/jupyterhub/templates/login.html
+++ b/share/jupyterhub/templates/login.html
@@ -35,82 +35,86 @@
               {% if login_error %}<p class="login_error">{{ login_error }}</p>{% endif %}
               <input type="hidden" name="_xsrf" value="{{ xsrf }}" />
               {# Allow full override of the "label" and "input" elements of the username and password fields. #}
-              {% block username_password_input_override %}
+              {% block username_input %}
                 <label for="username_input">Username:</label>
                 <input id="username_input"
-                  type="text"
-                  autocapitalize="off"
-                  autocorrect="off"
-                  autocomplete="username"
-                  class="form-control"
-                  name="username"
-                  val="{{ username }}"
-                  autofocus="autofocus"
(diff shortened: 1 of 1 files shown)


minrk (commented):
Thanks for the PR! Please do target all PRs to main, and we can consider backporting as necessary.

I think this idea is sensible, but I wonder if perhaps we should allow more customization by defining the block at a slightly higher level (e.g. the whole username/password inputs, rather than just _extra_ attributes). WDYT about putting a block around the whole label/input pair for each of username, password? I know it means your override block would need to duplicate more, but maybe that's for the best since you control the whole input element and label? That would let you _override_ any fields already defined, which extra attributes don't necessarily accomplish.

tlvu:
> That would let you _override_ any fields already defined, which extra attributes don't necessarily accomplish.

How about I add another block that would allow overriding all the inputs and keep the existing ones to only add extra attributes?  This would cater to both usages.

I usually prefer to avoid overriding too much to avoid breaking any changes upstream.

tlvu:
@minrk I have added an optional full override of username and password field and re-target the PR to main as you've suggested.

Also the diff on Github is misleading because this is all I've changed:

`$ git diff origin/main`
```diff
 
diff --git a/share/jupyterhub/templates/login.html b/share/jupyterhub/templates/login.html
index d9728733..0b45d5bc 100644
--- a/share/jupyterhub/templates/login.html
+++ b/share/jupyterhub/templates/login.html
@@ -34,6 +34,9 @@
               </p>
               {% if login_error %}<p class="login_error">{{ login_error }}</p>{% endif %}
               <input type="hidden" name="_xsrf" value="{{ xsrf }}" />
+
+              {# Allow full override of the "label" and "input" elements of the username and password fields. #}
+              {% block username_password_input_override %}
               <label for="username_input">Username:</label>
               <input id="username_input"
                      type="text"
@@ -43,13 +46,20 @@
                      class="form-control"
                      name="username"
                      val="{{ username }}"
-                     autofocus="autofocus" />
+                     autofocus="autofocus"
+                     {# allow adding extra HTML Input Attributes (pattern, placeholder, title, minlenght, maxlenght, ...) #}
+                     {% block username_input_attribs %}{% endblock username_input_attribs %}
+                     />
               <label for='password_input'>Password:</label>
               <input type="password"
                      class="form-control"
                      autocomplete="current-password"
                      name="password"
-                     id="password_input" />
+                     id="password_input"
+                     {# allow adding extra HTML Input Attributes (pattern, placeholder, title, minlenght, maxlenght, ...) #}
+                     {% block password_input_attribs %}{% endblock password_input_attribs %}
+                     />
+              {% endblock username_password_input_override %}
               {% if authenticator.request_otp %}
                 <label for='otp_input'>{{ authenticator.otp_prompt }}</label>
                 <input class="form-control"
```

minrk:
@tlvu thanks! I structured the blocks a little differently to make things as consistent as I could, so there's a `{name}_input` and `{name}_input_attrs` block for each of `username`, `password`, `otp`. The `{name}_input_attrs` now _includes_ the existing attrs instead of being strictly additive, so it allows override. Use `super()` if you want to extend the block instead of replacing the default:

```jinja
{% extends "login.html" %}

{% block username_input_attrs %}
{{ super() }}
pattern="[a-z0-9]+"
{% endblock username_input_attrs %}
```

tlvu:
@minrk Thanks !   I did not know about `super()` call !

PR #4982, opened 2025-01-17
make insecure-login-warning visible in dark mode

mishaschwartz:
In dark mode the "insecure login warning" on the login page changes the text colour but not the background colour. Both are quite light and make the text very hard to read:

Light mode:
image

Dark mode:
image

This PR adds the `alert alert-warning` css classes to the `insecure-login-warning` element so that the element styling is consistent with other warnings. This makes the warning much easier to read in dark mode:

Light mode:

image

Dark mode:

image

> [!Note]
> This also changes the border of the warning alert. I'm going to assume that this is OK since it makes it consistent with other warnings but if that is not desired, let me know and I can revert the borders to how they were.

### DIFF
--- a/share/jupyterhub/static/scss/login.scss
+++ b/share/jupyterhub/static/scss/login.scss
@@ -3,7 +3,6 @@
   height: 80vh;
 
   & #insecure-login-warning {
-    background-color: $warning-bg-subtle;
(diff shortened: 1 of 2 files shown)


minrk:
Thanks!

PR #4984, opened 2025-01-20
close tornado FDs without closing asyncio loop

minrk:
pytest-asyncio 0.25.2 introduces a bug causing errors if anything closes the event loop (https://github.com/pytest-dev/pytest-asyncio/issues/1040), <del>but I don't _think_ we need this anyway with current versions of things</del>.

We want to close tornado resources, but don't need to close the underlying event loop. Tornado doesn't offer an API to do one without the other, so instead disable asyncio loop closing while closing tornado.

### DIFF
--- a/jupyterhub/tests/conftest.py
+++ b/jupyterhub/tests/conftest.py
@@ -32,6 +32,7 @@ import os
 import sys
 from subprocess import TimeoutExpired
 from unittest import mock
+from warnings import warn
 
 import pytest_asyncio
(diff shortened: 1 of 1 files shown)


minrk:
not unnecessary at all...

We want to close the ioloop FDs _without_ closing the underlying event-loop, ideally. I'll see if there's an easy way to do that. Otherwise, we can just block 0.25.2 and move on, since it's known and fixable in pytest-asyncio.

Issue #4985, opened 2025-01-25
Singleuser: empty host is treated as `127.0.0.1` instead of `*`


manics:
### Bug description

Set `c.Spawner.ip = ""`. This should cause the singleuser server to listen on _all_ interfaces (IPv4 and IPv6). Instead it listens on `127.0.0.1`.

### How to reproduce

Set `c.Spawner.ip = ""` on a remote spawner, e.g. KubeSpawner.
JupyterHub will spawn the singleuser container but won't be able to connect, because the singleuser server listens on `127.0.0.1`

#### Expected behaviour

Singleuser server should listen on _all_ interfaces (IPv4 and IPv6)

#### Actual behaviour

Singleuser server listens on `127.0.0.1` only.

I think this is due to `url.hostname` being empty when `JUPYTERHUB_SERVICE_URL=http://:8888/user/test/` is parsed, which is treated as `127.0.0.1` in the singleuser code:

https://github.com/jupyterhub/jupyterhub/blob/2550d24048c6b7a51ff03138c28c9bb59d390972/jupyterhub/singleuser/extension.py#L511-L521

https://github.com/jupyterhub/jupyterhub/blob/2550d24048c6b7a51ff03138c28c9bb59d390972/jupyterhub/singleuser/mixins.py#L287-L291

### Your personal set up

```yaml
# helm upgrade --install jh --repo=https://hub.jupyter.org/helm-chart/ jupyterhub --version=4.1.0 -f z2jh-config.yaml --wait

custom:
  netpol: &netpol
    networkPolicy:
      enabled: false

hub:
  config:
    JupyterHub:
      default_url: /hub/home
      cleanup_servers: true
    Spawner:
      http_timeout: 600
    #  start_timeout: 600
    KubeSpawner:
      # ip: "[::]"
      ip: ""
  <<: *netpol

proxy:
  service:
    type: ClusterIP
  chp:
    <<: *netpol

scheduling:
  userScheduler:
    enabled: false
  userPlaceholder:
    enabled: false

singleuser:
  image:
    name: quay.io/jupyter/base-notebook
    tag: latest
  <<: *netpol

ingress:
  enabled: true
  hosts:
    - <HOSTNAME>
```

consideRatio:
Wieee nice work tracking all these IPv4 vs v6 differences across the Jupyter exosystem @manics!!

Issue #4830, opened 2024-05-31
test_share_codes_api_revoke flaky


minrk:
The test `test_share_codes_api_revoke[shares!server=SERVER-code=CODE-204]` has been flaky, sometimes returning a 404 status for the deleted code instead of 204 (DELETED).

I've tried to debug this a few times, but it's very hard to reproduce reliably.

Possible issues:

- failure to commit share code to database in the first place, e.g. via rollback
- escaping problem where transmitted code gets mangled somehow, in one direction or the other, or in the lookup code
- code gets expired/deleted somehow

None of these seem particularly likely. Escaping seems the _most_ likely, but that should be reproducible by hardcoding the code value to the one from the failed test.

Captured output from [this run](https://github.com/jupyterhub/jupyterhub/actions/runs/9304210711/job/25608263723?pr=4822#step:11:1069) for log retention purposes:

<details>

```
=================================== FAILURES ===================================
_______ test_share_codes_api_revoke[shares!server=SERVER-code=CODE-204] ________

app = <jupyterhub.tests.mocking.MockHub object at 0x7f76bac0ebd0>
user = <User(testuser-597 0/1 running)>, group = <Group testgroup-184>
share_user = <User(share_with-598 0/0 running)>
create_user_with_scopes = <function create_user_with_scopes.<locals>.temp_user_creator at 0x7f76e9685580>
have_scopes = 'shares!server=SERVER', delete_by = 'code=CODE', status = 204

    @pytest.mark.parametrize(
        "have_scopes, delete_by, status",
        [
            (None, None, 204),
            ("shares", "id=ID", 204),
            (
                "shares!server=SERVER",
                "code=CODE",
                204,
            ),
            ("shares!user=other", None, 404),
            ("read:shares", "code=CODE", 403),
            ("shares", "id=invalid", 404),
            ("shares", "id=sc_9999", 404),
            ("shares", "code=nomatch", 404),
        ],
    )
    async def test_share_codes_api_revoke(
        app,
        user,
        group,
        share_user,
        create_user_with_scopes,
        have_scopes,
        delete_by,
        status,
    ):
        db = app.db
        spawner = user.spawner.orm_spawner
    
        _expand_scopes = partial(expand_scopes, user=user, share_with=share_user)
        # make sure default spawner exists
        spawner = user.spawner.orm_spawner
        share_code, code = orm.ShareCode.new(
            db, spawner, scopes=list(scopes.access_scopes(spawner=spawner))
        )
[...]

PR #4986, opened 2025-01-26
Singleuser: listen on IPv4 and IPv6 if ip==""

manics:
If `Spawner.ip = ""` listen on all interfaces instead of only `127.0.0.1`.
This matches the behaviour of `jupyter-server --ip=""`

Closes https://github.com/jupyterhub/jupyterhub/issues/4985


minrk (approved):
Nice! Trying to unblock tests in #4984 then will merge

minrk:
flaky test: https://github.com/jupyterhub/jupyterhub/issues/4830

PR #4988, opened 2025-01-27
More IPv6: Use bare IPv6 for configuration, use `[ipv6]` when displaying IPv6 outputs

manics:
Consistently handle IPv6 addresses the same way across JupyterHub:
- Bare IPv6 used for configuration should not be wrapped by `[` `]`
- IPv6 displayed in logs or messages should always be wrapped by `[` `]`

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -480,6 +480,10 @@ class Spawner(LoggingConfigurable):
         Subclasses which launch remotely or in containers
         should override the default to '0.0.0.0'.
 
+        .. versionchanged:: 5.3
+            An empty string '' means all interfaces (IPv4 and IPv6). Prior to this
+            the behaviour of '' was not defined.
+
         .. versionchanged:: 2.0
             Default changed to '127.0.0.1', from unspecified.
         """,
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -470,6 +470,10 @@ class Spawner(LoggingConfigurable):
         The IP address (or hostname) the single-user server should listen on.
 
         Usually either '127.0.0.1' (default) or '0.0.0.0'.
+        On IPv6 only networks use '::1' or '::'.
+
+        If the spawned singleuser server is running JupyterHub 5.3.0 later
+        You can set this to the empty string '' to indicate both IPv4 and IPv6.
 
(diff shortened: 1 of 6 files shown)


jupyterhub/spawner.py
```diff
@@ -469,20 +470,35 @@ async def _get_oauth_client_allowed_scopes(self):
...
 
         Subclasses which launch remotely or in containers
         should override the default to '0.0.0.0'.
 
         .. versionchanged:: 2.0
-            Default changed to '127.0.0.1', from ''.
-            In most cases, this does not result in a change in behavior,
-            as '' was interpreted as 'unspecified',
-            which used the subprocesses' own default, itself usually '127.0.0.1'.
+            Default changed to '127.0.0.1', from unspecified.
```
manics, line 484: I'm not sure whether to update or leave this text. It made sense in the context of the 2.0 upgrade, but coupled with subsequent changes it potentially adds confusion.

In hindsight how we interpret the meaning of `''` affects how we treat https://github.com/jupyterhub/jupyterhub/pull/4986
If we consider `'' == 127.0.0.1` a bug, which is what I was thinking since it's otherwise impossible to pass `''` to the singleuser server, then the meaning of `''` has changed from _unspecified_ (prior to 2.0) to _impossible/invalid_ (between 2.0 and the current release) to _all interfaces_ (https://github.com/jupyterhub/jupyterhub/pull/4986).

However if we consider `'' == 127.0.0.1` intentional then https://github.com/jupyterhub/jupyterhub/pull/4986 is a breaking change- which we could workaround by making `'*'` mean all interfaces instead. However this risks adding more confusion since it'll still be converted to `''` when passed to the singleuser server.

minrk: Given that 2.0 is so long ago, I think this is a good change and reduces confusion, and we should let `ip=''` be meaningful. I think it is worth adding a `versionchanged` here that indicates `ip=''` is newly supported in 5.3, indicating that it wasn't supported prior to 5.3, even though it may have worked in an undocumented and potentially surprising way (turning into 127 somewhere along the line).

manics: Thanks, I've added a `versionchanged` section. I've also added a warning to `Spawner.ip` if `[]` are found.

PR #4991, opened 2025-01-28
Add $JUPYTERHUB_XSRF_ANONYMOUS_{IP_CIDRS|HEADERS} config for managing anonymous xsrf ids

minrk:
almost certainly proxy ips, which don't solve the problem anyway, and can change due to multiple proxy replicas

maybe closes https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/3422 for most users. A case where it wouldn't is a proxy outside the cluster/LAN, where the proxy's own ip is a global ip. For that, we'd need a more full opt-out of using ips.

### DIFF
--- a/jupyterhub/_xsrf_utils.py
+++ b/jupyterhub/_xsrf_utils.py
@@ -12,12 +12,10 @@ import base64
 import hashlib
 import os
 from http.cookies import SimpleCookie
-from ipaddress import ip_address
+from ipaddress import ip_address, ip_network
 
 from tornado import web
 from tornado.log import app_log
-
-from jupyterhub.utils import _bool_env
 
 
 def _get_signed_value_urlsafe(handler, name, b64_value):
@@ -235,7 +233,22 @@ def check_xsrf_cookie(handler):
 
 
 # allow disabling using ip in anonymous id
-_anonymous_use_ip = _bool_env("JUPYTERHUB_XSRF_ANONYMOUS_USE_IP", default=True)
+def _get_anonymous_ip_cidrs():
+    """
+    List of CIDRs to consider anonymous from $JUPYTERHUB_XSRF_ANONYMOUS_IP_CIDRS
+
+    e.g. private network IPs, which likely mean proxy ips
+    and do not meaningfully distinguish users
+    (fixing proxy headers would usually fix this).
+    """
+    cidr_list_env = os.environ.get("JUPYTERHUB_XSRF_ANONYMOUS_IP_CIDRS")
+    if not cidr_list_env:
+        return []
+    return [ip_network(cidr) for cidr in cidr_list_env.split(";")]
+
+
+_anonymous_ip_cidrs = _get_anonymous_ip_cidrs()
+
 # allow specifying which headers to use for anonymous id
 # (default: User-Agent)
 # these should be stable (over a few minutes) for a single client and unlikely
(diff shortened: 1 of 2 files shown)


manics:
Does this potentially open up a vulnerability with [CGNAT](https://en.wikipedia.org/wiki/Carrier-grade_NAT)?

minrk:
I believe `is_private` is False for CGNAT, so there is no change in behavior for those IPs. These values don't need to be perfect (e.g. everyone in a room behind the same router will have the same ip). The main goal is to limit the portability of xsrf cookies to mitigate things like cookie tossing.

minrk:
I added a couple of environment variables $JUPYTERHUB_XSRF_ANONYMOUS_USE_IP and $JUPYTERHUB_XSRF_ANONYMOUS_ID_HEADERS to influence how the anonymous id is computed. I did this because it's influencing a top-level utility function and traitlets-style config will be a bit of a pain to get a Configurable instance properly loaded. But not impossible, if folks think standard config would be better here. The unique anonymous id can be fully disabled with:

```bash
JUPYTERHUB_XSRF_ANONYMOUS_USE_IP=0
JUPYTERHUB_XSRF_ANONYMOUS_ID_HEADERS=
```

in which case an xsrf cookie for one client is valid for all clients (assuming client is _not_ authenticated), which opens a path to cookie tossing.

jupyterhub/_xsrf_utils.py
```diff
@@ -230,6 +234,17 @@ def check_xsrf_cookie(handler):
         )
 
 
+# allow disabling using ip in anonymous id
+_anonymous_use_ip = _bool_env("JUPYTERHUB_XSRF_ANONYMOUS_USE_IP", default=True)
```
samyuh, line 238: Shouldn't we use `JUPYTERHUB_XSRF_ANONYMOUS_USE_IP` with False by default to avoid any security problem and keep the old behaviour?

I think this problem is more likely to occur in k8s deployment, so we can set this property to True on the chart side.

minrk: I don't think so. True means keeping the current behavior by default, where the ip is used, which is correct most of the time (including in most z2jh deployments). Only where proxy headers don't properly propagate does this problem come up.

manics: Doesn't this default change the behaviour on internal networks, since all private IPs are treated as one?

minrk: Yes, it does. I'm not particularly concerned about that case, since it only affects the non-portability of the xsrf cookies for login, i.e. for cookie tossing scenarios. I don't think that would really be an issue for private network deployments.

manics: I still don't feel comfortable with treating all private IPs as one, but I'm not going to block merging this if others are happy.

As one possible alternative could we replace `JUPYTERHUB_XSRF_ANONYMOUS_USE_IP` with something like `JUPYTERHUB_XSRF_ANONYMOUS_<SOMETHING>_CIDRS` which defaults to empty (so no change to current mainline behaviour), but setting it to a CIDR would treat any IP in that CIDR as identical for the XSRF check? Then the equivalent of `JUPYTERHUB_XSRF_ANONYMOUS_USE_IP=False` would be `JUPYTERHUB_XSRF_ANONYMOUS_<SOMETHING>_CIDRS="0.0.0.0/0"`

minrk: I implemented your suggestion as `JUPYTERHUB_XSRF_ANONYMOUS_IP_CIDRS`. When a CIDR matches, the matching CIDR is used in the hash instead of the ip (so e.g. access from multiple flattened networks is still non-transferrable across networks), and it is now opt-in.

That means https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/3422 is no longer fixed by default, but it is _fixable_ and we may be able to make it fixed by default in z2jh, I _think_.

manics (approved):

PR #4995, opened 2025-02-03
make sure 'kind' shows up in rest api

minrk:
it's present in 'discriminator', but not mentioned explicitly in either model, so it's not visible in the generated docs


consideRatio (approved):

consideRatio:
Unrelated link-check failure it seems, mysql.com gives 403 but it works for me, so perhaps a blocked IP

PR #4996, opened 2025-02-03
Added data-server-name to add confirmation feature before deleting jupyterhub server

Sayanchk1:


Sayanchk1:
Will raise another with sone changes

minrk:
@Sayanchk1 you can keep pushing updates to your branch for a given PR and the PR will be updated, you do not need to close pull requests to make changes. Opening the PR is only the _start_ of the conversation, not a final patch. That way we can keep all the discussion and review in one place.

See more in the [github docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request#making-changes-to-files-in-your-pull-request).

PR #4997, opened 2025-02-03
Added confirmation feature before deleting any jupyterhub server

Sayanchk1:


manics:
I think adding a confirmation dialog is reasonable (should we make it optional?). It should be a React component though to fit in with the rest of the UI.

Sayanchk1:
I think we need to have a confirmation dialog box because we have seen if by mistake we click on delete , the server will be gone, we can't recover it. Having a confirmation dialog stleast give a chance to cancel the delete request. I have tested this feature and it works fine

Sayanchk1:
Will raise another with some changes

PR #5002, opened 2025-02-07
Add env vars and docs to help with debugging browser tests

manics:
I've often run playwright in headed mode for debugging tests, adding the environment variables means you don't need to temporarily edit `conftest.py`.


manics:
Maybe the env-vars could be replaced by custom pytest cmdline args instead? https://docs.pytest.org/en/stable/example/simple.html#pass-different-values-to-a-test-function-depending-on-command-line-options
I'll see if I can get it working.

manics:
I discovered the pytest plugin already adds some arguments. For some reason they weren't being respected. https://playwright.dev/python/docs/test-runners#cli-arguments says it's due to us creating a new context but I couldn't get it to work with the default context, so I've added explicit handling for some arguments.

manics:
Except we don't require the playwright-pytest plugin.... and given that it doesn't seem to be working with our current setup there's no point at the moment. I could add options like `parser.addoption("--headed", ...)` to mimic them, but that'll conflict if someone does have the playwright-pytest plugin installed.

minrk:
Makes sense! Feel free to add the pytest-playwright plugin if you think that would simplify things, or not.

jupyterhub/tests/browser/conftest.py
```diff
@@ -6,12 +6,45 @@
...
+    return dict(
+        (opt, request.config.getoption(f"--{opt}"))
+        for opt in ["headed", "slowmo", "videodir"]
+    )
+
+
 @pytest.fixture()
-async def browser():
+async def browser(cmdoptions):
+    print(cmdoptions)
```
minrk, line 39: print

minrk (approved):
feel free to merge when ready

PR #5006, opened 2025-02-16
doc: read:users only includes server not servers

manics:
The `read:users` scope only includes whether the default server is running or not, it doesn't include any server models which are under `servers`


jupyterhub/scopes.py
```diff
@@ -64,7 +64,7 @@
         'subscopes': ['read:users:name'],
     },
     'read:users': {
-        'description': 'Read user models (including servers, tokens and authentication state).',
+        'description': 'Read user models (including state of the default server, tokens and authentication state).',
```
minrk, line 67: Thanks for noticing! Neither tokens nor auth state are covered by read:users either, so this somehow got quite inaccurate. It is also not the _state_ of the default server (server_state is a thing covered by its own scope), but rather its URL if it's running. How about:

```suggestion
        'description': 'Read user models (including the URL of the default server if it is running).',
```

PR #5007, opened 2025-02-17
MockHub: randomize hub and proxy API ports

manics:
Without this MockHub uses a random public port, but it still uses fixed internal ports.

This should means you can run JupyterHub locally, and also run some pytests locally, without conflicting ports.

### DIFF
--- a/jupyterhub/tests/mocking.py
+++ b/jupyterhub/tests/mocking.py
@@ -246,6 +246,13 @@ class MockHub(JupyterHub):
         if 'allow_all' not in self.config.Authenticator:
             self.config.Authenticator.allow_all = True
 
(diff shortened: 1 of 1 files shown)


jupyterhub/tests/mocking.py
```diff
@@ -246,6 +246,13 @@ def __init__(self, *args, **kwargs):
         if 'allow_all' not in self.config.Authenticator:
             self.config.Authenticator.allow_all = True
 
+        if 'api_url' not in self.config.ConfigurableHTTPProxy:
```
manics, line 249: Needed to avoid breaking
https://github.com/jupyterhub/jupyterhub/blob/718f01e600ba55a65618d820161b72073767a3f8/jupyterhub/tests/test_proxy.py#L38

PR #5009, opened 2025-02-18
Automatically generate rest-api.yml and scopes.md using pre-commit

manics:
This avoids needing to manually run `generate-scope-table.py`.

If this works `pre-commit.ci` should automatically push a commit to this PR after it's opened as I deliberately didn't run pre-commit.


minrk:
Nice!

PR #5012, opened 2025-02-19
add apply_user_options hook

### LINKED ISSUE #4977: Refactor options_from_form interface
### Proposed change

Right now using `options_form` feature or more generally `user_options` requires subclassing the spawner because implementations must override the `start` method. Following [this discourse discussion](https://discourse.jupyter.org/t/more-love-to-spawner-options-form/31126/) I propose to change the Spawner API to have the following:

- Add an `apply_user_options: Union([Callable(), Dict()])` trait to `Spawner`, whose responsibility would be setting Spawner traits given user options. It should also be clearly noted that it should perform sanitization.
- Implement a default `options_from_form`, which would perform minimal parsing of form data.

### Alternative options

- Leave things as is

Hub administrators wishing to use `options_form`.

minrk:
cover more cases with config without need to subclass, and even some simple cases without needing Python functions in config.

New config:

- `Spawner.apply_user_options` - callable or dict of `{option: spawner_attribute}`
- `Spawner.options_from_form` has singleton string values `"passthrough"` (default, unchanged), and `"simple"` for the most common case of turning length-one lists into scalars.

closes #4977

### DIFF
--- a/docs/source/reference/spawners.md
+++ b/docs/source/reference/spawners.md
@@ -199,17 +199,17 @@ formdata = {
 When `formdata` arrives, it is passed through [](#Spawner.options_from_form):
 
 ```python
-spawner.user_options = [await] spawner.options_from_form(formdata[, spawner=spawner])
+spawner.user_options = spawner.options_from_form(formdata, spawner=spawner)
 ```
 
 to create `spawner.user_options`.
 
 [](#Spawner.options_from_form) is a configurable function to turn the HTTP form data into the correct structure for [](#Spawner.user_options).
-`options_from_form` must return a dictionary, and is meant to interpret the lists-of-strings a web form produces into the correct types.
+`options_from_form` must return a dictionary, _may_ be async, and is meant to interpret the lists-of-strings a web form produces into the correct types.
 For example, the `options_from_form` for the above form might look like:
 
 ```python
-def options_from_form(formdata):
+def options_from_form(formdata, spawner=None):
     options = {}
     options['integer'] = int(formdata['integer'][0]) # single integer value
     options['checkbox'] = formdata['checkbox'] == ['on']
--- a/docs/source/reference/spawners.md
+++ b/docs/source/reference/spawners.md
@@ -2,7 +2,7 @@
 
 # Spawners
 
-A [Spawner][] starts each single-user notebook server.
+A [Spawner](#Spawner) starts each single-user notebook server.
 The Spawner represents an abstract interface to a process,
 and a custom Spawner needs to be able to take three actions:
 
@@ -37,7 +37,7 @@ Some examples include:
 
 ### Spawner.start
 
-`Spawner.start` should start a single-user server for a single user.
+[](#Spawner.start) should start a single-user server for a single user.
 Information about the user can be retrieved from `self.user`,
 an object encapsulating the user's name, authentication, and server info.
 
@@ -68,11 +68,11 @@ async def start(self):
 When `Spawner.start` returns, the single-user server process should actually be running,
 not just requested. JupyterHub can handle `Spawner.start` being very slow
 (such as PBS-style batch queues, or instantiating whole AWS instances)
-via relaxing the `Spawner.start_timeout` config value.
+via relaxing the [](#Spawner.start_timeout) config value.
 
 #### Note on IPs and ports
 
-`Spawner.ip` and `Spawner.port` attributes set the _bind_ URL,
+[](#Spawner.ip) and [](#Spawner.port) attributes set the _bind_ URL,
 which the single-user server should listen on
 (passed to the single-user process via the `JUPYTERHUB_SERVICE_URL` environment variable).
 The _return_ value is the IP and port (or full URL) the Hub should _connect to_.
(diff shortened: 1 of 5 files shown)


akhmerov:
I wonder about backwards compatibility with subclasses. If a subclass implements applying options from form in `start`, will it work with the new JH?

minrk:
> If a subclass implements applying options from form in start, will it work with the new JH?

Yes, this should be fully backward-compatible, as the default implementations of these hooks remains the same no-op as before, and `user_option` remains defined when `.start()` is called, so the same Spawner logic will be called (usually none), after any user-configured hooks.

It will be possible to create _new_ configuration that won't work, though.

Spawners  generally must ignore any `user_options` they don't understand, since they can be processed by other code (e.g. `pre_spawn_hook` and the like). So the only situation where there might be any issue is in Spawners that define and process some `user_options` in `.start()` (for example handling `image` handling and `allowed_images` in DockerSpawner) and a user's hooks in the new API set the _same option_. In ~all real cases this should work fine, but if both interpret the same field in user_options to mean something different, then you might get an error or unintended result.

Take the `user_options['image']` option  handing in DockerSpawner, for example. If this option is set, it goes through [check_allowed](https://github.com/jupyterhub/dockerspawner/blob/41edb1f246731286302844b8fb793a3198946d0b/dockerspawner/dockerspawner.py#L1295L1298) before setting `spawner.image`. _If_ a deployment wrote `apply_options` to interpret the same `image` field in a different way:

```python
image_shortcut = {
    "scipy": "quay.io/jupyter/scipy-notebook",
    "base": "quay.io/jupyter/base-notebook",
}

def apply_user_options(spawner, user_options):
    if "image" in user_options and isinstance(user_options["image"], str):
        spawner.image = image_shortcut[user_options["image"]]

c.DockerSpawner.apply_user_options = apply_user_options
```

you would get an error during `start` of [Specifying image to launch is not allowed](https://github.com/jupyterhub/dockerspawner/blob/41edb1f246731286302844b8fb793a3198946d0b/dockerspawner/dockerspawner.py#L1091).

But if you called the option anything other than the already-defined `image`, it would work fine and exactly as indented, because `DockerSpawner` ignores options it doesn't already understand (and `image` is the only option it understands). So it's fully backward-compatible, with the knowledge that Spawners may have defined meanings of certain options in `user_options` and `apply_user_options` does not _replace_ this behavior, it _precedes_ it.

minrk:
This is now pretty fully tested and ready for review.

docs/source/reference/spawners.md
```diff
@@ -186,44 +185,155 @@ See [this example](https://github.com/jupyterhub/jupyterhub/blob/HEAD/examples/s
...
   'select': ['a', 'b'],
   'notinform': 'extra info',
 }
 ```
 
-When `Spawner.start` is called, this dictionary is accessible as `self.user_options`.
+### Applying user options
+
+The base Spawner class doesn't do anything with `user_options`, that is also up to your deployment and/or chosen Spawner.
+This is because exposing any options to users has security implications,
```
akhmerov, line 239: From reading the current documentation it wasn't completely clear to me that there is an API call that allows to provide arbitrary `user_options`. Consider mentioning it here.

akhmerov: perhaps this:

```suggestion
This is because the users can specify arbitrary option dictionary by using the API,
```

akhmerov (approved):

jupyterhub/spawner.py
```diff
@@ -685,12 +727,127 @@ def options_from_query(self, query_data):
...
+                    f"Received unhandled user_options for {self._log_name}: {', '.join(keys)}"
+                )
+            if isawaitable(r):
+                await r
+        except Exception as e:
+            # this may not be the users' fault...
+            # should we catch less?
+            # likely user errors are ValueError, TraitError, TypeError
+            self.log.exception("Exception applying user_options for %s", self._log_name)
+            raise web.HTTPError(400, f"Invalid user options: {e}")
```
manics, line 794: Do we have any guidelines in JupyterHub on which exceptions are exposed to users, to ensure internal information isn't leaked?

minrk: We mention it [in authenticators](https://jupyterhub.readthedocs.io/en/stable/reference/authenticators.html#custom-error-messages), but the general pattern is that HTTPErrors are displayed to users. So if you want to present an error to a user, raising an HTTPError in your hook/Authenticator/Spawner method should let you display things to the user, while any other error should be logged and wrapped in a generic, appropriate HTTPError (usually 500).

You can also customize messages with a `jupyterhub[_html]_message` attribute on your exceptions, which affects what is displayed when things fail, but not the status code.

There have been issues where that passthrough doesn't always propagate correctly, but that's how it's supposed to work.

manics: Does this potentially break expectations then, since the content of a standard exception which would normally be internally logged is now automatically converted to one and exposed to the user?

minrk: Yeah, good point. I'll switch this to do the passthrough, so only a specifically-raised HTTPError gets passed through, otherwise generic `HTTPError(400, "Invalid user options")`

manics (approved):

Issue #4989, opened 2025-01-27
test_search_on_admin_page flaky browser test


minrk:
this browser test has become flaky. Need to check what race conditions we are running into

```
jupyterhub/tests/browser/test_browser.py::test_search_on_admin_page[60-user] - AssertionError: Locator expected to have count '11'
Actual value: 50 
Call log:
  - LocatorAssertions.to_have_count with timeout 5000ms
  -   - waiting for locator("//tr[@class=\"user-row\"]")
  -     9 × locator resolved to 50 elements
  -       - unexpected value "50"
 ```

In the meantime, it's okay to merge PRs if you see this as the only failure, unless you're working on the admin page itself

manics:
https://github.com/user-attachments/assets/5a3e4349-876a-47d7-a424-ac3e606d16c3

Before `Next` is clicked:
![Image](https://github.com/user-attachments/assets/cee5de2d-25a6-45de-a582-64dc5fb42a09)

After `Next` is clicked:
![Image](https://github.com/user-attachments/assets/48b84025-f6d7-4e2d-a3ec-6e8634435e72)

minrk:
I've done a bit of testing on this, and it appears to be related to clicking `Next` and editing the name filter both updating the `offset`, and there can be weird races if you do one while there's already a request outstanding. This is easier to provoke by adding an `await asyncio.sleep(3)` to `GET /api/users/`, but I'm not much closer to understanding how to fix it.

The key problem is the PaginationFooter representing the `offset` searchParam, which it out of date while awaiting a response.

PR #5020, opened 2025-03-19
make sure custom error messages are shown on regular error pages

### LINKED ISSUE #2291: Don't rewrap HTTP errors
Brought about here: https://discourse.jupyter.org/t/custom-jupyterhub-error-messages/47

Suggestion: do not re-wrap HTTPErrors raised, only other errors.

That would make it possible to use `tornado.web.HTTPError` with a custom error template (as opposed to the current working solution of using a new template inherited from the base error HTML template).

minrk:
not just the `not_running` page, which is only served for slow spawns and/or past launches. Considering this a bug because this behavior is [documented](https://jupyterhub.readthedocs.io/en/stable/reference/spawners.html#exception-handling), but the correct behavior only occurs if the error happens after `slow_spawn_timeout`. This is also related to #5011 where the custom error would not be displayed when there's a race between slow spawn return and `spawner.poll()` indicating a failed start.

closes #2291

cc @kellyrowland

### DIFF
--- a/jupyterhub/tests/test_pages.py
+++ b/jupyterhub/tests/test_pages.py
@@ -2,11 +2,14 @@
 
 import asyncio
 import sys
+from contextlib import nullcontext
+from functools import partial
 from unittest import mock
 from urllib.parse import parse_qs, urlencode, urlparse
 
 import pytest
 from bs4 import BeautifulSoup
+from tornado import web
 from tornado.httputil import url_concat
 
 from .. import orm, roles, scopes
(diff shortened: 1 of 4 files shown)


jupyterhub/tests/test_pages.py
```diff
@@ -1335,3 +1338,115 @@ async def test_services_nav_links(
...
+            assert "<b>🕸️🫖</b>" in error_html
+        elif kind == "unhandled":
+            assert unhandled_text in error.text
+            assert "unhandle me" not in error.text
+        else:
+            raise ValueError(f"unexpected {kind=}")
+
+
+#
+# async def test_spawn_fails_custom_message_slow(app, user, no_patience):
```
consideRatio, line 1427: A test that is commented out here

minrk: Thanks! I wrote it as two, then merged them with a parameter but forgot to delete the old one.

consideRatio (approved):
This looks right, but there is a test failing that I wasn't sure from looking at if it was or wasn't related to this PR.

minrk:
Thanks. The flaky test is tracked as #4989

agoose77:
I wonder if we should consider using exception notes instead of custom JupyterHub attributes here? https://peps.python.org/pep-0678/

minrk:
Notes sound like a great idea, I didn't know about them! The only issue I see is that notes is a list of strings and not jupyterhub-specific, so if an exception has a note, we might interpret it incorrectly whereas `.jupyterhub_message` is clearly defined as "show this message to jupyterhub users." And we specifically have a mechanism to provide both text and HTML, so how would we do that with notes?

So questions for adopting notes:

1. do we need to check if it's jupyterhub-specific, or are notes enough (i.e. would there ever be a note we _shouldn't_ show users)?
2. if it's jupyterhub specific, how do we signal that? (e.g. simple `jupyterhub:` string prefix on a note)
3. for HTML messages, how do we signal that? (e.g. `jupyterhub_html:` string prefix)

I think it's definitely doable, if we want to give it a try. Is it better? Probably, if we didn't already have an API folks are using. I'm not sure if it's better _enough_ to justify a change to how we do it (maybe deprecating but definitely not removing the old way), but I can be persuaded if folks want to explore it!

PR #5022, opened 2025-03-21
improve xsrf errors on login

minrk:
- show login form for trying again, with an error message, just like a password failure
- nicer, but more vague "try again" error for expired xsrf (original error still logged)
   because users logging in don't need to know or understand xsrf stuff
- set fresh xsrf cookie when login page loads, to maximize time until expiration

What failed XSRF on login looks like before:

Screenshot 2025-03-21 at 17 35 43

and after:

Screenshot 2025-03-21 at 17 34 27

### DIFF
--- a/jupyterhub/_xsrf_utils.py
+++ b/jupyterhub/_xsrf_utils.py
@@ -106,9 +106,12 @@ def _get_xsrf_token_cookie(handler):
     return (None, None)
 
 
-def _set_xsrf_cookie(handler, xsrf_id, *, cookie_path="", authenticated=None):
+def _set_xsrf_cookie(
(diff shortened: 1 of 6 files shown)


consideRatio (commented):
I think there is a non-intermittent test failure still lurking, the changes looks good to me so far though.

```
FAILED jupyterhub/tests/browser/test_browser.py::test_login_xsrf_initial_cookies[valid-prefix-invalid-root] - AssertionError: assert '/' == '/@/space%20word/hub/'
```

minrk:
Yes, it was a real failure. The test expected that visiting `/login` a second time wouldn't set a new cookie, which this PR changes.

Issue #4954, opened 2024-11-19
Stop publishing images for jupyterhub


minrk:
### Proposed change

Related to https://github.com/jupyterhub/team-compass/issues/474 I think maybe we should stop publishing jupyterhub images altogether.

We currently publish these images from this repo:

- jupyterhub/demo - a toy complete environment for quickstart/testing purposes (not useful for any deployment) where users are run in the container
- jupyterhub/jupyterhub - includes hub and proxy, no Spawner cannot run users  - not useful for production
- jupyterhub/jupyterhub-onbuild (adds a single `ONBUILD ADD` to jupyterhub/jupyterhub) for a minor convenience as a base image
- jupyterhub/singleuser - shouldn't be used, offers nothing over docker-stacks

Production deployments do not generally use these images because:

- they do not include a Spawner, so can't be used on their own, they can only be a _base_ image for other custom images
- the base image part can be fully replaced with a single `pip install` or `conda install` in a more nontrivial base image
- jupyterhub/jupyterhub includes CHP, which should not generally be run in the same image, and we shouldn't be implicitly encouraging it

Maintaining these images on this repo causes maintenance problems like #4953  and the various motivations for https://github.com/jupyterhub/team-compass/issues/474

I propose that we stop publishing these images altogether until/unless we get image-building going from its own repo. I wish I had proposed this before 5.0, since a major version is a nice cutoff for a change in practice, but the second best time is now.

The only image I think we _might_ keep is `demo`, since it serves a unique purpose and is clearly not for production use. The others may give the _impression_ of being appropriate for something they shouldn't really be used for.

Stopping publishing from here does not _preclude_ us from publishing images (even these same images) from another repo. But it does mean that we need to set up infrastructure _dedicated to building and maintaining images_ before we start building these images again, which I think will help us be thoughtful about it.

### Alternative options

- Keep publishing  from this repo (I don't think we should do this, but it will keep happening if we don't change something)
- Move image-building to its own repo in a more thoughtful way _before_ we stop building here, focused on building images that address real use cases

### Who would use this feature?

[...]

PR #5024, opened 2025-03-24
stop publishing images from jupyterhub/jupyterhub

minrk:
moving to jupyterhub-container-images repo

towards #4954 (this actually closes the titular issue there, but leaving it open because there are other related tasks to track)


manics (approved):

consideRatio (approved):
Nice work on this @manics @minrk!!

Note that cleanup of docker-build job in the test workflow remains

PR #5025, opened 2025-03-24
Try to improve admin paging consistency

### LINKED ISSUE #4989: test_search_on_admin_page flaky browser test

minrk:
I _think_ this fixes #4989 

It all stems from trying to make sure the pagination footer ("Displaying x-y of z") reflects the actual view. This is complicated by the fact that react has 'view state' and when view state changes is when rerendering happens.

Changing 'offset' in the state triggers a request for a new page, but it takes time for the response to come in. Previously, I tried to change the offset _only_ (ish) on the _response_, so the state `offset` always reflected the current view. As a result, triggering a state change sends a request, the response to which triggers another state change. This wasn't reliable, and could cause weird loops where callbacks trigger state changes in an infinite cycle (easier to provoke by adding a few second sleep in `GET /api/users`).

The main fix here is changing the rendered offset to always be `user_page.offset`, never the state offset, so the data load response doesn't change the request parameter state. Variable-wise, that means `offset` is the _requested_ offset, which may or may not be what you see, while `user_page.offset` is always what you see. Triggering requests is now always done through a standard state change, rather than _most_ of the time.

Related to the races, outdated responses that don't represent the _last_ request are ignored, thanks to cancellation. This helps with the fighting between latest state and incoming responses

I still don't really understand how to write good react tests, as right now too much is mocked, so key functionality isn't really tested effectively. But that's not a new thing.

### DIFF
--- a/jsx/src/components/Groups/Groups.jsx
+++ b/jsx/src/components/Groups/Groups.jsx
@@ -14,15 +14,13 @@ const Groups = (props) => {
   const dispatch = useDispatch();
   const navigate = useNavigate();
 
-  const { offset, setOffset, handleLimit, limit, setPagination } =
-    usePaginationParams();
+  const { offset, setOffset, handleLimit, limit } = usePaginationParams();
 
   const total = groups_page ? groups_page.total : undefined;
 
   const { updateGroups } = props;
 
   const dispatchPageUpdate = (data, page) => {
-    setPagination(page);
     dispatch({
       type: "GROUPS_PAGE",
       value: {
--- a/jsx/src/components/Groups/Groups.jsx
+++ b/jsx/src/components/Groups/Groups.jsx
@@ -14,7 +14,8 @@ const Groups = (props) => {
   const dispatch = useDispatch();
   const navigate = useNavigate();
 
-  const { offset, handleLimit, limit, setPagination } = usePaginationParams();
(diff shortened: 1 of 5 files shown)


minrk:
still seems to be something wrong. Clicking `next` doesn't appear to be visiting the next page in some test runs, though it does in practice. I think there are still some cycle races going on here.

minrk:
okay, I actually found the race. The flaky test wasn't waiting for the first search filter to be applied, and the `debounce` delays meant the order was often:

1. visit page
2. enter name filter
3. check first page (checks pass whether the name filter is applied or not)
4. click next
5. name filter change applied after debounce delay

The race between 4 and 5 was the source of the failure. By making sure the first page check (step 3) verifies the name filter has actually be applied, the race is resolved.

PR #5033, opened 2025-03-27
url_path_join: handle empty trailing components

manics:
This ensures that `url_path_join("/x/", "")` returns `/x/` not `/x`

https://discourse.jupyter.org/t/health-check-of-managed-service-returns-404/34142 reports that the JupyterHub service check uses `/services/name` instead of `/services/name/`. This is because `Server.wait_up` has an extra parameter that defaults to `""`
https://github.com/jupyterhub/jupyterhub/blob/742de1311e2671c2f5ea1187c508e60d299a9f63/jupyterhub/objects.py#L168-L176
which results in the trailing `/` being stripped by `url_path_join`

closes #5034

### DIFF
--- a/jupyterhub/tests/test_utils.py
+++ b/jupyterhub/tests/test_utils.py
@@ -101,6 +101,23 @@ async def test_tornado_coroutines():
 
 
 @pytest.mark.parametrize(
+    "pieces, expected",
+    [
+        (("/"), "/"),
+        (("/", "/"), "/"),
+        (("/base", ""), "/base"),
+        (("/base/", ""), "/base/"),
+        (("/base", "abc", "def"), "/base/abc/def"),
+        (("/base/", "/abc/", "/def/"), "/base/abc/def/"),
+        ((""), ""),
+        (("abc", "def"), "abc/def"),
+    ],
+)
+def test_url_path_join(pieces, expected):
+    assert utils.url_path_join(*pieces) == expected
+
+
+@pytest.mark.parametrize(
     "forwarded, x_scheme, x_forwarded_proto, expected",
     [
         ("", "", "", "_attr_"),
--- a/jupyterhub/tests/test_utils.py
+++ b/jupyterhub/tests/test_utils.py
@@ -109,7 +109,13 @@ async def test_tornado_coroutines():
         (("/base/", ""), "/base/"),
         (("/base", "abc", "def"), "/base/abc/def"),
         (("/base/", "/abc/", "/def/"), "/base/abc/def/"),
+        (("/base", "", "/", ""), "/base/"),
         ((""), ""),
+        (("", ""), ""),
+        (("", "part", ""), "part"),
+        (("", "/part"), "part"),
+        (("", "part", "", "after"), "part/after"),
+        (("", "part", "", "after/", "", ""), "part/after/"),
         (("abc", "def"), "abc/def"),
     ],
 )
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1692,7 +1692,11 @@ class JupyterHub(Application):
         """add a url prefix to handlers"""
         for i, tup in enumerate(handlers):
             lis = list(tup)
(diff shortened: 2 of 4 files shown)


manics:
The current behaviour of dropping the trailing slash if there's an empty component after it seems unintuitive, but it looks like we depend on that behaviour.

minrk:
We may be depending on it, but I'm not sure we are _on purpose_. I'll look into where this changes behavior, it might be fixable easily enough.

btw, if you had asked me what `url_path_join("/x", "")` returned, I would have told you `"/x/"`, so I agree the current behavior is surprising.

minrk:
The first change is that the AddSlashHandler that's supposed to redirect `/hub` to `/hub/` is [here](https://github.com/jupyterhub/jupyterhub/blob/5e77ca22e37d2330845f8854620eae24fd32c400/jupyterhub/handlers/base.py#L1969), which goes through `url_path_join` [here](https://github.com/jupyterhub/jupyterhub/blob/5e77ca22e37d2330845f8854620eae24fd32c400/jupyterhub/app.py#L1695)

That's a very specific case, so we could handle it specially:


```python
            if tup[0]:
                lis[0] = url_path_join(prefix, tup[0])
            else:
                # the '' route should match /prefix not /prefix/
                lis[0] = prefix.rstrip("/")
```

It does make sense that `url_path_join("/prefix/" ...)` wouldn't be able to produce the _shorter_ URL `"/prefix"`

minrk:
The other failure is [here](https://github.com/jupyterhub/jupyterhub/blob/5e77ca22e37d2330845f8854620eae24fd32c400/jupyterhub/user.py#L789) where we _assume_ a trailing slash won't be added by url_path_join, but this change means it does when server name is empty. Putting the `+ '/'` inside fixes this, letting url_path_join handling trailing things:

```python
        base_url = url_path_join(self.base_url, url_escape_path(server_name) + '/')
```

minrk:
I think with those two changes, everything passes. WDYT about a trailing empty element adding a slash or being ignored being more intuitive? Should `url_path_join("/base", "")` produce `/base` or `/base/`? If I were to guess, I would probably guess that a trailing slash was added, but that's not what we do. I think ignoring trailing empty elements is relatively clear and documented here, so I'm okay with that, too, if it makes the most sense to you.

I think we can definitely consider it a bug that `join("/base/", "")` can ever produce something _shorter_ than `/base/`.

jupyterhub/tests/test_utils.py
```diff
@@ -100,6 +100,23 @@ async def test_tornado_coroutines():
...
+@pytest.mark.parametrize(
+    "pieces, expected",
+    [
+        (("/"), "/"),
+        (("/", "/"), "/"),
+        (("/base", ""), "/base"),
+        (("/base/", ""), "/base/"),
+        (("/base", "abc", "def"), "/base/abc/def"),
+        (("/base/", "/abc/", "/def/"), "/base/abc/def/"),
+        ((""), ""),
```
minrk, line 112: Some more test cases to cover:

```suggestion
        (("/base", "", "/", ""), "/base/"),
        ((""), ""),
        (("", ""), ""),
        (("", "part", ""), "part"),
        (("", "/part"), "part"),
        (("", "part", "", "after"), "part/after"),
        (("", "part", "", "after/", "", ""), "part/after/"),
```

minrk (approved):
I think this is a good change. I think we can go either way on whether `join("/prefix", "")` adds a segment (i.e. ensures trailing slash) or trailing empty parts are explicitly ignored and have no effect. Your change here is the smaller change, so I'm fine going with that.

I think it's unambiguously a bug fix top stop _striping_ trailing slashes from components when there are empty components after, even though we (accidentally) relied on that in a couple of places. I don't think that was ever intended behavior.

manics:
Thanks for your insight and fixes! If this was a new function I'd probably throw an error if empty paths were found, but given the age of this code I don't think we should do that. I'm happy with your changes.

PR #5034, opened 2025-03-27
server checks: don't delete trailing `/`

manics:
`extra_path=""` was added in https://github.com/jupyterhub/jupyterhub/pull/4567 , by adding this new check we restore the original behaviour of keeping the trailing `/`. Reported in https://discourse.jupyter.org/t/health-check-of-managed-service-returns-404/34142
(see also https://github.com/jupyterhub/jupyterhub/pull/5033 for an explanation of why the trailing `/` was deleted)


minrk:
Thanks! I think we can go with the deeper fix in #5033

PR #5035, opened 2025-03-28
run pre-commit

minrk:
for some reason `pre-commit run --all-files` is making this change locally, even though CI doesn't seem to. Let's see if it at least doesn't object.


consideRatio:
Hmmm maybe different versions are involved for some reason?

minrk:
Yup, updating pre-commit itself fixed it. Should have thought of that! Must be a difference in how hooks are invoked. Weird.

PR #5037, opened 2025-03-28
Add SharedPasswordAuthenticator

yuvipanda:
Currently, admin users are even more insecure than otherwise with dummyauthenticator - anyone who knows the username of the admin can get in if they also know the password.

This PR adds an additional layer of security - admins *must* login using a different, more secure (longer, per NIST guidelines) password. If they login using the regular password, no admin status for them.

This mildly helpful in local testing and improves overall security posture. Where it really shines though, is in 'workshop' hubs. I've been running those for years now, both at UC Berkeley and now at 2i2c (with NASA Openscapes in particular). This was the usecase DummyAuth was written for :D It allows an instructor to share a single password with all the users in a secure way (they're all in a physical room, zoom, etc). The password is then changed after the workshop. However, admin access was not possible in this use case, as anyone guessing the admin's username can get in as admin. With this change, admin access is possible.

### DIFF
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -1510,63 +1510,6 @@ class DummyAuthenticator(Authenticator):
         """,
     )
 
-    admin_password = Unicode(
-        None,
-        allow_none=True,
-        config=True,
-        help="""
-        Set a global password that grants *admin* privileges to users logging in with this password.
-
-        Must meet the following requirements:
-        - Be 32 characters or longer
-        - password for regular users must also be set
-        - Not be the same as the password
-        """,
-    )
-
-    @validate("admin_password")
-    def _validate_admin_password(self, proposal):
-        new = proposal.value
-        if new is None:
-            # Don't do anything if we're None
-            return None
-        if len(new) < 32:
-            raise ValueError(
-                f"{self.__class__.__name__}.admin_password must be at least 32 characters"
-            )
-        if not self.password:
-            # Checked here and in validating password, to ensure we don't miss issues due to ordering
-            raise ValueError(
-                f"{self.__class__.__name__}.password must be set if admin_password is set"
-            )
-        if self.password == new:
-            # Checked here and in validating password, to ensure we don't miss issues due to ordering
-            raise ValueError(
-                f"{self.__class__.__name__}.password and {self.__class__.__name__}.admin_password can not be the same"
-            )
-        return new
-
-    @validate("password")
-    def _validate_password(self, proposal):
-        new = proposal.value
-        if self.admin_password and not new:
-            # Checked here and in validating admin_password, to ensure we don't miss issues due to ordering
-            raise ValueError(
-                f"{self.__class__.__name__}.password must be set if admin_password is set"
-            )
-
-        if not new:
-            # If unset, let it be
-            return new
-
-        if self.admin_password == new:
-            # Checked here and in validating password, to ensure we don't miss issues due to ordering
-            raise ValueError(
-                f"{self.__class__.__name__}.password and {self.__class__.__name__}.admin_password can not be the same"
-            )
-
-        return new
-
     def check_allow_config(self):
         super().check_allow_config()
         self.log.warning(
@@ -1577,10 +1520,7 @@ class DummyAuthenticator(Authenticator):
         """Checks against a global password if it's been set. If not, allow any user/pass combo"""
         if self.password:
             if data['password'] == self.password:
-                # Anyone logging in with the standard password is *never* admin
-                return {"name": data['username'], "admin": False}
-            if data['username'] in self.admin_users and data['password'] == self.admin_password:
-                return {"name": data["username"], "admin": True}
+                return data["username"]
             return None
         return data['username']
 
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -20,7 +20,7 @@ except Exception as e:
     _pamela_error = e
 
 from tornado.concurrent import run_on_executor
-from traitlets import Any, Bool, Dict, Integer, Set, Unicode, default, observe
+from traitlets import Any, Bool, Dict, Integer, Set, Unicode, default, observe, validate
 from traitlets.config import LoggingConfigurable
 
 from .handlers.login import LoginHandler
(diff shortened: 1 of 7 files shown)


jules32:
+1 This would be super helpful with our workshops, that can happen in pretty quick succession with different combinations of people as teaching leads/admins. Thank you!

minrk (commented):
I think requiring a separate admin password does make sense, thanks! This is both more secure when admins aren't needed for testing, and actually makes it _more_ convenient when they are, because you don't have to define `admin_users` in test config, just a single password.

But I'd like to avoid making it too inconvenient to use this insecure-by-design authenticator for testing, since that's why we have it. Can we drop the length requirement and the requirement that the user password be set?

I think this should be fine test config:

```python
c.DummyAuthenticator.admin_password = "admin"
c.JupyterHub.authenticator_class = "dummy"
```

and it shouldn't be our responsibility to enforce what a 'good' password is in a testing tool.

jupyterhub/auth.py
```diff
@@ -1520,7 +1577,13 @@ async def authenticate(self, handler, data):
         """Checks against a global password if it's been set. If not, allow any user/pass combo"""
         if self.password:
             if data['password'] == self.password:
-                return data['username']
+                # Anyone logging in with the standard password is *never* admin
+                return {"name": data['username'], "admin": False}
```
minrk, line 1581: I think I see a slight problem here: if a user knows the user password and an admin's username, they can still login as the admin user and _revoke_ the admin user's admin privileges. I think you also need to check `not in self.admin_users` before accepting the default password.

yuvipanda: Ah that's fair.

yuvipanda:
Can I leave the password length requirement in? I'm happy to remove the requirement for the default password, but I feel the password length requirement is a good bang for buck one.

yuvipanda:
> actually makes it more convenient when they are, because you don't have to define admin_users in test config,

This is not true! That was my original design, but now you do need both to set admin_users and the password.

minrk:
Ah, I think I understand the admin password requiring the regular password being set: this doesn't replace the "full dummy" default behavior where anyone, admins included, can login with any password. So it's more "this would have no effect if self.password is not also set". That makes sense. Do you think admin_password should be required for logging admins in with dummy by default as well, not just when a user password is set? That's what my initial read was.

 I also didn't clock that `admin_password` is also matched with `admin_users` (makes sense, now that I see admin_users must be checked to _prevent_ the regular password from logging in with admin usernames), so admin users must still be declared in config if the admin password is used. That should be in the `admin_password` docstring, because I think it currently gives the impression that if anyone uses this password, they will be able to login as admin.

manics:
I don't think we should enforce a password policy here, I often use this for testing, and in any case there are lots of different password policies so it seems a bit arbitrary to pick 32 chars.

yuvipanda:
The new NIST guidelines say '15 or more for highly privileged accounts' (https://www.strongdm.com/blog/nist-password-guidelines)

manics:
In my mind the dummy authenticator has only ever been for testing, and never for production use ("dummy" == "fake")

If we want to make it suitable for production I think we should subclass it (or create a new authenticator) and give it a name that more accurately describes its purpose.

yuvipanda:
In that case, the other option is:

1. DummyAuthenticator no longer supports password authentication, and errors out
2. Create a separate SharedPasswordAuthenticator that supports both these features.

The primary question for me is where this lives:

1. In this repo. The default otherwise is PAM based auth, and I think SharedPasswordAuthenticator (or similar) is a better default
2. In another repo in the jupyterhub/ org

As long as it lives somewhere that gets into z2jh, I'm happy with it. However, my primary worry is that this ends up leading us down a multi year governance (or other) refactor that runs out of steam :) 

So my *actual* preference is to simply merge this PR as is, as it *does* improve security. Progress over perfection, etc.

yuvipanda:
Another alternative is to simply stop calling it 'DummyAuthenticator' - it stopped being that the moment we added `password`

minrk:
I actually think splitting it into SharedPasswordAuthenticator and keeping it here with your changes makes perfect sense. Then it doesn't seem so weird to have a test class with real-world uses.

yuvipanda:
ok i love the idea @minrk. Lemme mark this as draft and come back

ateucher:
I'm late to the conversation, but would be very happy to see this implemented! At Openscapes we use the 'workshop' hub with a shared password as @yuvipanda outlined. Having admin access enabled for our shared password hubs would be great, especially if that opens the door for allowing `shared` directory that is read-only for regular users but writeable for admins,(like [this](https://docs.2i2c.org/user/topics/data/sharing/#the-allusers-directory)).

minrk:
No problem, I'll try to pick it up tomorrow. Thanks!

minrk:
@yuvipanda I fleshed it out and added some tests and docs. The main changes I made:

- Remove DummyAuthenticator's allow all by default, so defaults are more like a "real" authenticator
- call the user  password `user_password`
- Warn rather than error if no user password is set, because it is valid config to disable auth by having no password set
- Log informative warnings for several suspicious configs:
  - only one of admin_password and admin_users set, since both must be set to allow admins to login
  - user_password not set
  - neither password is set

I didn't change the admin password length from 32, though I do think 16 feels more appropriate since it matches the NIST guideline as I understand it, unless I'm missing something. But I'm also okay sticking with your preference.

jupyterhub/authenticators/shared.py
```diff
@@ -0,0 +1,149 @@
...
+    async def authenticate(self, handler, data):
+        """Checks against shared password"""
+        if data["username"] in self.admin_users:
+            # Admin user
+            if self.admin_password and compare_digest(
+                data["password"], self.admin_password
+            ):
+                return {"name": data["username"], "admin": True}
+        else:
+            if self.user_password and compare_digest(
```
yuvipanda, line 143: oooh i was thinking about timing attacks and TIL about compare_digest! Nice

yuvipanda:
Thank you @minrk! I read through it and it looks good to me :)

I'll merge on Monday if there are no objections by then - I feel even though I opened this PR, @minrk has worked enough on this that I feel comfortable merging it.

minrk:
Thanks @yuvipanda !

Issue #5019, opened 2025-03-12
Link to JupyterHub server events schema is broken


jnywong:
The [Jupyter Server Events page](https://jupyterhub.readthedocs.io/en/latest/reference/server-actions.html) returns a 404 error for the following link: https://schema.jupyter.org/jupyterhub/events/server-action.

Ref: https://github.com/jupyterhub/jupyterhub/blob/main/docs/source/reference/server-actions.md?plain=1

manics:
According to https://jupyter.org/enhancement-proposals/108-jupyter-subdomain-for-schemas/jupyter-subdomain-for-schemas.html
The schema needs to be added to https://github.com/jupyter/schema

minrk:
We should definitely add them to that repo. The schema website still isn't up yet, so even adding them to that repo, the URLs won't resolve. We should try to read https://github.com/jupyter/schema/issues/2 to figure out what our actual schema URLs should be, because I think it might not be what the ultimate result is supposed to be.

jnywong:
Is there anything I can do to help shepherd this along? How does one contribute to https://github.com/jupyter/schema?

minrk:
I don't think these schema ids were necessarily meant to be resolvable URLs, but I'm not sure. jsonschema `$id` URIs definitely don't require that they be resolvable on the network, only that they uniquely identify the schema, but I agree it's weird to display a URL that doesn't work.

I think more work needs to happen on https://github.com/jupyter/schema/issues/3 before there's much we can do, but it appears there hasn't been activity on that project in a while.

We can certainly manually add our event schemas to the repo by passing them through yaml to json:

```python
with open("schema.json", "w") as json_out, open("schema.yaml") as yaml_in:
    json.dump(yaml.safe_load(yaml_in), json_out, sort_keys=True, indent=1)
```

and opening a PR to schema adding the JSON files to the schema repo under a new `jupyterhub` folder. But that still won't make the URL resolve.

Maybe there's a way to omit the schema URL from the docs, or at least render it as a literal instead of a link so it doesn't look broken?

jnywong:
Opened a couple of PRs to contribute the schema and update the id that puts us on the path to fixing this.

For now, I've left the schema URL in the docs as it is: I didn't want to remove it entirely and wasn't obvious to me how to render the link as a string literal with yaml.

PR #5038, opened 2025-04-01
Update events/server_action schema id

jnywong:
I've updated the schema id to snake case, which seems to be the convention and opened a PR to contribute the schema to jupyter/schema: https://github.com/jupyter/schema/pull/7.

Related to #5019


jnywong:
Actually, I'll leave the id name alone so it doesn't break the api 😅 

Will update https://github.com/jupyter/schema/pull/7 to reflect this.

PR #5041, opened 2025-04-05
Add instruction on how to select dummy authenticator

ktaletsk:
When working with JupyterHub I noticed there was not a clear instruction on how to quickly set DummyAuthenticator (or any other authenticator for that matter) in the config.
Adding it here, so it's easier to find right away in docs


GeorgianaElena (approved):
This is a very good idea! Thank you @ktaletsk!

minrk:
Great! I was just noticing that this is missing while working on #5037

PR #5042, opened 2025-04-07
changelog for 5.3.0 (RC)

minrk:
I'd like to do a 5.3 release candidate and get some testing in the helm chart, then make a release in a week or two.

### DIFF
--- a/docs/source/reference/changelog.md
+++ b/docs/source/reference/changelog.md
@@ -20,6 +20,76 @@ Contributors to major version bumps in JupyterHub include:
 
 ## [Unreleased]
 
+## 5.3
+
+### 5.3.0 - 2025-04
+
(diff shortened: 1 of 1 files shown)


docs/source/reference/changelog.md
```diff
@@ -20,6 +20,76 @@ Contributors to major version bumps in JupyterHub include:
 
 ## [Unreleased]
 
+## 5.3
+
+### 5.3.0 - 2025-04
+
+([full changelog](https://github.com/jupyterhub/jupyterhub/compare/5.2.1...HEAD))
```
consideRatio, line 27: Besides updating these lines, :+1: on proceeding with a pre-release!

minrk: yup, these should be updated when making the final release.

PR #5048, opened 2025-04-14
changelog for 5.3.0

minrk:
planning to release after tomorrow's meeting: https://github.com/jupyterhub/team-compass/issues/766


consideRatio (approved):

manics:
I think it's worth adding a note (either in the changelog or in an announcement post) that the container image versioning has changed, so you should use `quay.io/jupyterhub/jupyterhub:5.3.0-0` instead of `quay.io/jupyterhub/jupyterhub:5.3.0` to pin the image.

minrk:
Great point! I’ll make sure there’s a note in the morning

minrk:
merging and publishing now...

PR #5062, opened 2025-04-30
Add create system user to jupyterhub config in fastapi example

chilin0525:
The service FastAPI example (https://github.com/jupyterhub/jupyterhub/tree/main/examples/service-fastapi#try-it-out-in-docker) appears to be missing the configuration `c.LocalAuthenticator.create_system_users = True`, which prevents the system user `testuser` from being able to log in.

### DIFF
--- a/docs/source/howto/forced-login.md
+++ b/docs/source/howto/forced-login.md
@@ -1,0 +1,130 @@
+# Logging users in via URL
+
+Sometimes, JupyterHub is integrated into an existing application that has already handled user login, etc..
+It is often preferable in these applications to be able to link users to their running JupyterHub server without _prompting_ the user to login again with the Hub when the Hub should really be an implementation detail,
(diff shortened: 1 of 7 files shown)


manics:
What error did you see when using this example? It's using the `simple` spawner which is insecure, but fine for testing and doesn't require a system user.

chilin0525:
@manics Logging in with `testuser / passwd` as described in the example's README.md fails. However, logging in with the username `test-user` and any password works.

Would it make sense to update the README to use `test-user` instead?

截圖 2025-05-01 00 58 18

manics:
That's an error in the README, changing it to `test-user` would be great!

chilin0525:
@manics Thanks! 🙏 I opened another one PR https://github.com/jupyterhub/jupyterhub/pull/5064 to fix!

PR #5071, opened 2025-05-30
Update docs to point to `/hub/user-redirect/` instead of `/user-redirect/`

agoose77:
In the docs, we point people to `/user-redirect` for user-agnostic URL construction:
https://github.com/jupyterhub/jupyterhub/blob/bc21e99e7efccd12840239cdc9419a577ee17c47/docs/source/reference/urls.md?plain=1#L172

However, the handler for this URL is properly defined under `/hub/user-redirect/`, and it looks as though the `/user-redirect/` is 302 redirected to the `/hub/user-redirect/` endpoint.

My interpretation is that we should be encouraging `/hub/user-redirect/` over the redirecting alias.

### DIFF
--- a/docs/source/reference/urls.md
+++ b/docs/source/reference/urls.md
@@ -169,14 +169,14 @@ _Version changed: 1.0_
 JupyterHub version 0.9 failed these API requests with status `404`,
 but version 1.0 uses 503.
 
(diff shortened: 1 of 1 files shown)


consideRatio (approved):
Nice makes sense to me!

Issue #5089, opened 2025-07-03
Event Handling for Cancelled Spawner bug


kreuzert:
### Bug description

Just a small bug:
https://github.com/jupyterhub/jupyterhub/blob/01a43f41f8b1554f2de659104284f6345d76636d/jupyterhub/apihandlers/users.py#L860C1-L876C85

`if f and f.cancelled():` is true, the failed_event will not be send to the frontend.

Line 918 does it correctly ( https://github.com/jupyterhub/jupyterhub/blob/01a43f41f8b1554f2de659104284f6345d76636d/jupyterhub/apihandlers/users.py#L918 ) 

### Your personal set up

 - OS:
 
 - Version(s):
 

<details><summary>Full environment</summary>

```
# paste output of `pip freeze` or `conda list` here
```
</details>

<details><summary>Configuration</summary>

```python
# jupyterhub_config.py
```
</details>

<details><summary>Logs</summary>

```
# paste relevant logs here, if any
```
</details>

PR #5091, opened 2025-07-04
send event if spawn_future was cancelled and spawner not pending

kreuzert:
Description: https://github.com/jupyterhub/jupyterhub/issues/5089


manics (approved):
Thanks!

PR #5093, opened 2025-07-04
Add confirmation dialog for named server deletion

kateryna-tarelkina-dd:
A user accidentally deleted their named server instead of starting it because the buttons are right next to each other, and deletion has no confirmation dialog. 

This PR adds a confirmation dialog to ensure accidents like this do not happen.


kateryna-tarelkina-dd:
@manics could I get a review on this PR please?

minrk:
I think CI on main is healthier now, reopening to trigger CI again.

This looks good to me and works in local tests, thank you!

minrk (approved):

PR #5102, opened 2025-07-17
Fix hub activity log to use spawner.last_activity

joeyutong:
Update debug log to compare activity timestamps with spawner.last_activity instead of user.last_activity.


minrk:
Thanks!

PR #5109, opened 2025-07-22
fix: use `contains_eager` instead of O(N²) `joinedload`  in `init_spawners`

agoose77:
When debugging an OOM crash of a deployed JupyterHub, we noticed that the JupyterHub was crashing around the `init_spawners` startup event. In the case that one has many (thousands of) spawner entries in the database (which we did, for testing purposes),  the query in `init_servers` falls over, as it scales O(N²) with servers:
```python
(
 self.db.query(orm.User, orm.Spawner)
  # join filters out any Users with no Spawners
  .join(orm.Spawner, orm.User._orm_spawners)
  # this gets Users with *any* active server
  .filter(orm.Spawner.server != None)
  # pre-load relationships to avoid O(N active servers) queries
  .options(
      joinedload(orm.User._orm_spawners),
      joinedload(orm.Spawner.server),
  )
)
```
Here, the bug is that `join` performs a JOIN, and `joinedload(orm.User._orm_spawners)` performs a _second_ JOIN. This scales O(N²) in spawners for each user, because the join is on the _user ID_, which is degenerate (many spawners per user). 

The intention behind these JOINs is not to pull in new tables, but rather to preserve the result. I am not an sqlalchemy expert, but my understanding is that it's better to retain the functional JOIN and instruct sqlalchemy to preserve the result with `contains_eager`: https://gist.github.com/agoose77/9f389c0d49d8fdd58b489ded6dda5f9a

### DIFF
--- a/jupyterhub/apihandlers/users.py
+++ b/jupyterhub/apihandlers/users.py
@@ -15,7 +15,7 @@ else:
 
 from dateutil.parser import parse as parse_date
 from sqlalchemy import func, or_
-from sqlalchemy.orm import contains_eager, raiseload, selectinload  # noqa
+from sqlalchemy.orm import joinedload, raiseload, selectinload  # noqa
 from tornado import web
 from tornado.iostream import StreamClosedError
 
@@ -162,20 +162,17 @@ class UserListAPIHandler(APIHandler):
             )
         elif state_filter:
             raise web.HTTPError(400, f"Unrecognized state filter: {state_filter!r}")
-        else:
-            # need a join on Spawner for contains_eager below
-            query = query.outerjoin(orm.Spawner, orm.User._orm_spawners)
 
         # apply eager load options
         query = query.options(
             selectinload(orm.User.roles),
             selectinload(orm.User.groups),
-            selectinload(orm.User._orm_spawners).selectinload(orm.Spawner.user),
+            joinedload(orm.User._orm_spawners).joinedload(orm.Spawner.user),
             # raiseload here helps us make sure we've loaded everything in one query
             # but since we share a single db session, we can't do this for real
             # but it's useful in testing
             # raiseload("*"),
-        ).populate_existing()
+        )
 
         sub_scope = self.parsed_scopes['list:users']
         if sub_scope != scopes.Scope.ALL:


minrk:
Makes sense! I wonder if there are other applications of joinedload that should be contains_eager. Have you had a look?

minrk:
I'm guessing [this join](https://github.com/jupyterhub/jupyterhub/blob/01a43f41f8b1554f2de659104284f6345d76636d/jupyterhub/apihandlers/users.py#L170) is one that should be the same, since it's applied after an explicit join on the same fields. There may not be others.

minrk:
linking other related things:

- #4496
- #4494 (which added this joinedload)
- #4485
- #1643

I'd really like to get something like https://github.com/minrk/jupyterhub-bench integrated via codspeed or asv to cover some basic scaling of JupyterHub. I think the most relevant: GET /users and startup as a function of number of:

- User records
- Spawner records
- running servers

(not asking for anything in this PR, just writing down my thoughts while they are here).

If you want to look at other related joinedloads, that would be great! But it's also AOK to stop here and I will merge this and look into the others later.

yuvipanda:
I've also been poking around https://github.com/yuvipanda/jupyterhub-simulator which is what initially set this off

minrk:
@agoose77 the tests suggest I was wrong there. Looking at the code, it looks like there isn't _always_ a join on Spawner. I'll have a look

agoose77:
Thanks 

> @agoose77 the tests suggest I was wrong there. Looking at the code, it looks like there isn't _always_ a join on Spawner. I'll have a look

I was just taking a look myself, figuring out whether my understanding is wrong!

minrk:
I think adding an outerjoin in the `else` to make sure the join is always there should work, I just pushed a commit since tests pass locally for me. If it turns out to be more complicated than that, let's back it out and merge your PR and we can work on the more complicated one separately.

agoose77:
@minrk I can see that there is a database dependency on these failures. I'd like to pull out a test db that I can run sqlalchemy against in a notebook. Do you have any quick tips for doing that, so that I avoid going down a rabbit hole?

I'm not an expert on sqlalchemy, so I'm tweaking this PR as I learn.

jupyterhub/app.py
```diff
@@ -3097,8 +3097,8 @@ async def check_spawner(user, name, spawner):
             .filter(orm.Spawner.server != None)
             # pre-load relationships to avoid O(N active servers) queries
             .options(
-                joinedload(orm.User._orm_spawners),
-                joinedload(orm.Spawner.server),
+                contains_eager(orm.User._orm_spawners),
```
agoose77, line 3100: Prefer not using `joinedload` for one-to-many.

jupyterhub/apihandlers/users.py
```diff
@@ -162,17 +162,20 @@ def get(self):
...
             selectinload(orm.User.roles),
             selectinload(orm.User.groups),
-            joinedload(orm.User._orm_spawners).joinedload(orm.Spawner.user),
+            contains_eager(orm.User._orm_spawners).selectinload(orm.Spawner.user),
             # raiseload here helps us make sure we've loaded everything in one query
             # but since we share a single db session, we can't do this for real
             # but it's useful in testing
             # raiseload("*"),
-        )
+        ).populate_existing()
```
agoose77, line 178: Whenever using `contains_eager` with filtering, we need to `populate_existing`: 
- https://github.com/sqlalchemy/sqlalchemy/issues/5740#issuecomment-737240021
- https://docs.sqlalchemy.org/en/14/orm/loading_relationships.html#using-contains-eager-to-load-a-custom-filtered-collection-result

agoose77:
Hmm, this might be trickier than I first hoped. `populate_existing` seems necessary but introduces slow down on an example DB I have. It also seems like mysql is not happy! I'm running out of time today to look at this, but will try and pick up again soon.

minrk:
I don't have a great tip other than https://github.com/minrk/jupyterhub-bench being what I put together to poke into these things a while back, so it might have some useful references.

Thanks for looking into it! I can have a look too, if you want, though I'm not sure when I'll have time.

agoose77:
OK, I time-boxed 45m this morning to take a proper look. I actually read the stack trace from the test-suite for MySQL, and the error we were seeing was quite simple:

When one introduces a `group_by`, it is possible only to `SELECT` either the grouping column (`user.id`), or summary statistics like `MIN`, etc. We cannot therefore use `contains_eager` — these data are lost after grouping. Interesting, it seems like `sqlite` is just giving us the first row silently, rather than noisily complaining like MySQL.

The simplest solution seems to be to preserve the grouping, which enables nice counting operations, and re-request the data from the database. We _can_ potentially make this faster (or slower...) by using `selectinload` rather than `joinedload`. This is notable a _different_ performance enhancement to the issue that motivated this PR: it's micro-optimisation, rather than pathalogical worst-case performance.

I am not certain whether `selectinload` vs `joinedload` will win-out in typical usage patterns; `selectinload` actually performs mulitple DB queries vs `joinedload` performing a single JOIN. Maybe there's a crossover point after which JOIN is just better.

@minrk what is your preference here? Whilst I think the change to `app.py` _is_ an important fix, I don't think we can be so confident with the changes to `users.py`. Are you in favour of dropping those changes, because it's probably _worse_ to fix something that's not thought of as broken? I'm worried about introducing a horrible regression, because I don't have a nice perf test suite to support the change.

I can take this over the line once I hear back, or you are entirely welcome to do so! (being confident that the tests are going to pass... famous last words).

minrk:
If the app.py fix is correct, let's do just that one here and leave users.py for later. Thanks for digging in!

agoose77:
@minrk do I need to do anything else to get this over the line?

minrk:
Thanks! I'll try to put together some numbers

Issue #5098, opened 2025-07-07
Broken links in documentation


rgaiacs:
When working on https://github.com/jupyterhub/jupyterhub/pull/5096, GitHub actions reported broken links

```
howto/rest.md:204 https://docs.python-requests.org
    HTTPSConnectionPool(host='docs.python-requests.org', port=443): Max retries exceeded with url: / (Caused by NameResolutionError("<urllib3.connection.HTTPSConnection object at 0x7f48cf4b8f10>: Failed to resolve 'docs.python-requests.org' ([Errno -2] Name or service not known)"))
reference/services.md:524 https://docs.python-requests.org/en/master/
    HTTPSConnectionPool(host='docs.python-requests.org', port=443): Max retries exceeded with url: /en/master/ (Caused by NameResolutionError("<urllib3.connection.HTTPSConnection object at 0x7f48bfa5fd10>: Failed to resolve 'docs.python-requests.org' ([Errno -2] Name or service not known)"))
reference/gallery-jhub-deployments.md:115 https://mediaspace.msu.edu/media/Setting+Up+Your+JupyterHub+Password/1_hgv13aag/11980471
    HTTPSConnectionPool(host='mediaspace.msu.edu', port=443): Max retries exceeded with url: /media/Setting+Up+Your+JupyterHub+Password/1_hgv13aag/11980471 (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1016)')))
reference/gallery-jhub-deployments.md:26 https://www.nersc.gov/assets/Uploads/03-MovingAndSharingData-Cholia.pdf
    404 Client Error: Not Found for url: https://www.nersc.gov/assets/Uploads/03-MovingAndSharingData-Cholia.pdf
reference/gallery-jhub-deployments.md:25 https://www.nersc.gov/news-publications/nersc-news/nersc-center-news/2016/jupyter-notebooks-will-open-up-new-possibilities-on-nerscs-cori-supercomputer/
    404 Client Error: Not Found for url: https://www.nersc.gov/news-publications/nersc-news/nersc-center-news/2016/jupyter-notebooks-will-open-up-new-possibilities-on-nerscs-cori-supercomputer
reference/changelog.md:1834 https://voila-gallery.org
[...]

PR #5123, opened 2025-08-06
allow 6 items in header drop-down

### LINKED ISSUE #5114: Show contributor documentation more prominent
In https://jupyterhub.readthedocs.io/, the link for the contributor documentation is hide under the `More` drop down list.

Image

But the `More` drop down list has one and only one item.

Could we make the contributor documentation more prominent by moving it one level up and removing `More`.

minrk:
we only have 6 items, so don't put one item under More:

Screenshot 2025-08-05 at 21 14 28

closes #5114

I'll open an issue on pydata-sphinx-theme to see if this can be done automatically.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -262,6 +262,7 @@ html_static_path = ["_static"]
 
 html_theme = "jupyterhub_sphinx_theme"
 html_theme_options = {
+    "header_links_before_dropdown": 6,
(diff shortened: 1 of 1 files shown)


rgaiacs (approved):
Looks good to me. The test is failing but this is another issue reported as https://github.com/jupyterhub/jupyterhub/issues/5098.

Issue #5115, opened 2025-08-05
Mention Windows Subsystem for Linux in Setting up a development install


rgaiacs:
[Setting up a development install](https://jupyterhub.readthedocs.io/en/stable/contributing/setup.html) says

> JupyterHub can only run on macOS or Linux operating systems. If you are using Windows, we recommend using [VirtualBox](https://virtualbox.org/) or a similar system to run [Ubuntu Linux](https://ubuntu.com/) for development.

We should mention [Windows Subsystem for Linux](https://learn.microsoft.com/en-us/windows/wsl/) explicitly as an alternative.

Also, we should explain why JupyterHub don't run on Windows.

minrk:
Do you use WSL? I'm not sure we should recommend WSL unless we have folks who are known to be using it successfully with JupyterHub. Most reports we've had from WSL users are when things _don't_ work, especially regarding sockets, so I wouldn't want to give folks the wrong impression (like Microsoft does, that WSL actually works reliably like linux, which isn't true in practice).

rgaiacs:
> Do you use WSL?

Yes, I'm using WSL regularly in the last two years. Earlier in 2025, I moved from Windows 10 to Windows 11. Given that Windows 10 support ends on October 2025, I think we could support Windows 11 + WSL.

> Most reports we've had from WSL users are when things don't work, especially regarding sockets

Do you remember one of this reports regarding sockets?

> WSL actually works reliably like linux, which isn't true in practice

All the problems that I had are related with network. For example,

- `localhost` or `127.0.0.1` will result to different machines. WSL will map the `localhost` of Windows to the WSL but `localhost` in the WSL always resolve to the machine. This is problematic when the domain is used to connect to other machines.
- WSL is know to now work with many VPN and [DNS Tunneling](https://learn.microsoft.com/en-us/windows/wsl/networking#dns-tunneling) not always work.

Another "minor" problem is that `systemd` might not be running but it can be [enabled](https://learn.microsoft.com/en-us/windows/wsl/systemd).

minrk:
If you're using it successfully, then please feel free to add it! I'm sure some folks would prefer that.

I'm not prepared to support it or answer any support questions related to WSL, so I can direct any questions your way, if that works for you. I also wouldn't commit to tests being able to run on WSL, but that doesn't preclude a development environment. So we can say that it may work, but if you have issues, make sure you have a real linux development environment before asking for help.

[...]

PR #5129, opened 2025-08-14
Add note about Windows Subsystem for Linux

rgaiacs:
Closes https://github.com/jupyterhub/jupyterhub/issues/5115

Inspired by [Installing Polynote](https://polynote.org/latest/docs/installation/#installing-polynote).

image


rgaiacs:
Thanks for the comments. Since the comments were addressed, I'm merging this.

PR #5142, opened 2025-09-20
JupyterCon banner and jupyter colors

choldgraf:
This PR does a couple of things:

- Adds an announcement banner for JupyterCon
- Links a CSS file from the Jupyter documentation that we can use for jupyter colors in our docs

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -231,7 +231,6 @@ def setup(app):
     app.add_css_file("https://docs.jupyter.org/en/latest/_static/jupyter.css")
 
 
-
 # -- Read The Docs -----------------------------------------------------------
 #
 # Since RTD runs sphinx-build directly without running "make html", we run the
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -228,6 +228,8 @@ def setup(app):
     app.add_directive("jupyterhub-generate-config", ConfigDirective)
     app.add_directive("jupyterhub-help-all", HelpAllDirective)
     app.add_directive("jupyterhub-rest-api-links", RestAPILinksDirective)
+    app.add_css_file("https://docs.jupyter.org/en/latest/_static/jupyter.css")
+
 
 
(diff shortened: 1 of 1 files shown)


manics:
The default version for https://jupyterhub.readthedocs.io/ is [stable](https://jupyterhub.readthedocs.io/en/stable/) i.e. the last release, so this won't be seen by most people.

choldgraf:
ah....so there's no way for us to push out a message like this unless we make a new release? Or is there a way to make a PR against `stable` somehow?

manics:
Short version: not easily 

Long version: yes
- If there are no changes to the docs on the `main` branch that don't apply to the most recent tag we could temporarily switch readthedocs to default to `latest` instead of `stable`. If we do this we should make sure we revert the banner PR before making a release.
- Or we could create a new branch based on the last tag and temporarily use that as the default readthedocs branch.

manics:
Is the style sheet addition temporary, or do you want it to be permanent for future use? If it's the latter I think we should host the stylesheet so it's not blocked if people have strict browser settings that block cross-domain requests. Or failing that make sure we linked to a fixed version so old versions of our docs won't break if the stylesheet is modified.

krassowski:
> If it's the latter I think we should host the stylesheet so it's not blocked if people have strict browser settings that block cross-domain requests

What if the docs were hosted from `docs.jupyterhub.jupyter.org` and similar for other subprojects (in addition to readthedocs free URLs). I know there may be a cost associated with this, but just wondering if this is worth exploring at all.

minrk:
We can make a branch from 5.3, e.g. 'promo' and point stable to that with just this addition

minrk:
This is in latest and I've temporarily pushed a 'stable' branch we can update, so the 'stable' builds point to that instead of the latest stable tag, so folks will see it at https://jupyterhub.readthedocs.io. We can revert the banner and remove the `stable` branch after jupytercon.

PR #5146, opened 2025-09-29
set HTTP status when spawn via GET params fails

agoose77:
In https://github.com/2i2c-org/infrastructure/issues/6689, 2i2c noticed that certain spawn requests were not triggering our Prometheus instrumentation for hub response codes. It turns out that these internal errors are logged, but never make it to the user:

- A user visits the Hub
- They are redirected to `/spawn`
- `SpawnHandler._post` calls `_wrap_spawn_single_user`
- `_wrap_spawn_single_user` calls `spawn_single_user`
- `spawn_single_user` throws a web error: https://github.com/jupyterhub/jupyterhub/blob/074917d9beeaeda37908ca1bcb02a0bede55f38b/jupyterhub/handlers/base.py#L1086-L1093
- `SpawnHandler._post` swallows the exception and presents the spawn form: https://github.com/jupyterhub/jupyterhub/blob/074917d9beeaeda37908ca1bcb02a0bede55f38b/jupyterhub/handlers/pages.py#L255-L273

So technically the `/spawn` route is not returning HTTP 429 to the caller. As such, it technically makes sense that we do not log it.

This PR specifically forwards these errors in the case that we swallow them and render a spawn form. This means that we'll still miss HTTP 429 errors that occur during `/spawn?...` GET requests, but only if there isn't an option form and the subsequent spawn without parameters succeeds. I think this is fine -- we really care about what the user sees here.

This would benefit from a test — I'd welcome any pointers on how best to test this kind of statefulness.

### DIFF
--- a/jupyterhub/handlers/pages.py
+++ b/jupyterhub/handlers/pages.py
@@ -98,7 +98,9 @@ class SpawnHandler(BaseHandler):
 
     default_url = None
 
-    async def _render_form(self, for_user, spawner_options_form, message=''):
+    async def _render_form(
+        self, for_user, spawner_options_form, message='', html_message=''
+    ):
         auth_state = await for_user.get_auth_state()
         return await self.render_template(
             'spawn.html',
@@ -106,6 +108,7 @@ class SpawnHandler(BaseHandler):
             auth_state=auth_state,
             spawner_options_form=spawner_options_form,
             error_message=message,
+            html_error_message=html_message,
             url=url_concat(
                 self.request.uri, {"_xsrf": self.xsrf_token.decode('ascii')}
             ),


manics:
Is it just 429 that's a problem- is anything else being swallowed?

agoose77:
> Is it just 429 that's a problem- is anything else being swallowed?

Other exceptions are caught and do not trigger the default error handling, indeed. However, we still log them to the logger (and this is already true for 429).

This PR special cases 429 because it's useful feedback for the client _and_ Prometheus inspects the status code of the handler.

I'm not 100% certain whether we can be more general here given that it probably doesn't always make sense to expose the error codes from internal method calls when returning from the outer handler.

minrk:
I think _if_ it's an HTTPError, it does make sense to surface it, rather than special-casing 429. Arguably, we could only surface 4XX codes that are theoretically the user's problem. 5xx are supposedly server bugs and would most likely just display "unhandled error", which likely isn't actionable information, but should still probably be shown if the spawn err. Most actually unhandled errors will not be `HTTPError`. `HTTPError` is typically raised with the specific goal of showing the user a message, so I think removing the error code check is probably the right thing to do.

It might also be worth retrieving jupyterhub-specific error messages, as seen in https://github.com/jupyterhub/jupyterhub/pull/5020

agoose77:
@minrk I was developing against the CI tests whilst I set up an environment with PAM locally. As such, it's a bit slower to iterate -- I can't see why the API test is failing. From the changes I've made to the source, it looks like it should be fine. Please feel free to fix-up, otherwise I'll get to this on Tuesday.

minrk:
I think we might have a flaky test on our hands, sorry if it's wasting your time. I just re-ran the one failing test, I'll go ahead and merge if it passes.

Thank you!

Issue #5158, opened 2025-10-02
PyCurl not used when internal_ssl is True


kreuzert:
### Bug description
When you use `c.JupyterHub.internal_ssl = True`, JupyterHub always uses the SimpleAsyncHTTPClient, even if pycurl is installed.

#### Why?
In [here](https://github.com/jupyterhub/jupyterhub/blob/5.3.0/jupyterhub/app.py#L1992) JupyterHub sets the AsyncCurlHTTPClient to "SimpleAsyncCurlHTTPClient".  
The later called `init_pycurl()` function does not overwrite / revert the tornado configuration.

Is this intended behavior? Should JupyterHub check in [here](https://github.com/jupyterhub/jupyterhub/blob/5.3.0/jupyterhub/app.py#L1985) if pycurl is available and configure the tornado asyncclient accordingly?

### How to reproduce

1. Create environment with pycurl + jupyterhub installed.
2. Update proxy.py to log the name of the client class [here](https://github.com/jupyterhub/jupyterhub/blob/5.3.0/jupyterhub/proxy.py#L926) 
```
    async def api_request(self, path, method='GET', body=None, client=None):
        """Make an authenticated API request of the proxy."""
        client = client or AsyncHTTPClient()
        url = url_path_join(self.api_url, 'api/routes', path)

        if isinstance(body, dict):
            body = json.dumps(body)
        self.log.warning("Proxy: Fetching %s %s", method, url) 
        self.log.warning(client.__class__.__name__) # Log current client class name
```
3. Play around with a config file with `c.JupyterHub.internal_ssl = True/False`

#### Expected behaviour

JupyterHub should use the `CurlAsyncHTTPClient` to make requests to the proxy, if `internal_ssl` is `True`.

#### Actual behaviour

JupyterHub uses the `SimpleAsyncHTTPClient` if `internal_ssl` is `True`, even if pycurl is installed.

minrk:
Totally a bug. I don't recall, but I'm guessing that the use of `impl=None` was expected to leave it unchanged, rather than revert to the default. #5159 should fix it

PR #5159, opened 2025-10-02
don't revert asynchttp class when setting up internal ssl

minrk:
None reverts to the default class, it doesn't not preserve the current value

closes #5158

PR #5160, opened 2025-10-02
changelog for 5.4.0

minrk:
Draft because I'd like to get #5146 and #5158 in, but we've accumulated a number of fixes that we can release. Small impact, nice little release, no need for prerelease testing, I think

We could do #5142 as well, but I'd kind of like to not tie that to a release, maybe by using a temporary branch for stable


minrk:
I think we can make a nice little 5.4 release. We've accumulated a number of small bugfixes and things I'd like to get out.

I think it's also about time to start shifting the baseline Python version, which means 6.0, and I'd like these fixes to be in 5.x

docs/source/reference/changelog.md
```diff
@@ -20,6 +20,73 @@ Contributors to major version bumps in JupyterHub include:
 
 ## [Unreleased]
 
+## 5.4
+
+### 5.4.0 - DATE
```
consideRatio, line 25: Date unspecified, but everything else is good to go I think!

consideRatio (approved):

minrk:
Thanks! I don't want to make releases late on a Friday, so I'll merge and release this Monday. I've set the date accordingly

minrk:
I need to hold this because of https://github.com/jupyterhub/jupyterhub/pull/5159#issuecomment-3370158592

either revert #5159 and release or figure out how to make internal ssl work with pycurl

minrk:
Releasing now, with one last fix in #5164 (attempted fix was #5159, but it wasn't tested and wasn't quite right. Now it's tested)

minrk:
5.4.0 is out!

PR #5165, opened 2025-10-06
remove jupytercon banner

minrk:
So we can remember to merge this after jupytercon

Reverts announcement from #5142

We will also need to:

- delete the `stable` branch
- restore the tag-based `stable` builds on readthedocs.org at https://app.readthedocs.org/projects/jupyterhub/


manics:
I think we should remove the `https://docs.jupyter.org/en/latest/_static/jupyter.css` too. If the content of that CSS file changes in future (e.g. more elements are overridden, or new classes which clash with ones in our docs are added) it could break the appearance of a tagged docs release, and there'll be no way to fix it. I think ideally the docs for a tagged release should be effectively immutable.

minrk:
@manics Makes sense. Should we change it to a build-time download? Or just copy the file into the repo?

manics:
I think keep things simple and copy it into this repo?

manics:
I've deleted the `stable` branch from this repo

PR #5171, opened 2025-10-09
allow ssl_alt_names to be callable

### LINKED ISSUE #5170: make Spawner.ssl_alt_names callable
With the new JupyterHub 5.4.0 release and its fixes in the `internal_ssl` eco-system ( e.g. https://github.com/jupyterhub/jupyterhub/pull/5076 , https://github.com/jupyterhub/jupyterhub/pull/5164 ) , it's stricter than before and the `Spawner.ssl_alt_names` parameter becomes more important.
  
In the help text it is mentioned to "set it at runtime". Shouldn't it be of Type `Any` then, to allow a callable implementation and set it depending of the `name` + `user_options` of the Spawner object?
```
    ssl_alt_names = List(
        Unicode(),
        config=True,
        help="""List of SSL alt names

        May be set in config if all spawners should have the same value(s),
        or set at runtime by Spawner that know their names.
        """,
    )
```

### Proposed change

Make it callable, default is an empty list:
```python
    ssl_alt_names = Any(
        default_value=[],
        config=True,
        help="""List of SSL alt names

        May be set in config if all spawners should have the same value(s),
        or set at runtime by Spawner that know their names.
        """,
    )
```

Update Spawner.create_certs() ( https://github.com/jupyterhub/jupyterhub/blob/main/jupyterhub/spawner.py#L1480 ) to allow callable ssl_alt_names:
```python
        default_names = ["DNS:localhost", "IP:127.0.0.1"]
        alt_names = []

        ssl_alt_names = self.ssl_alt_names
        if callable(ssl_alt_names):
            ssl_alt_names = ssl_alt_names(self)
            if inspect.isawaitable(ssl_alt_names):
                ssl_alt_names = await ssl_alt_names
        alt_names.extend(ssl_alt_names)

        if self.ssl_alt_names_include_local:
            alt_names = default_names + alt_names
```

### Alternative options
Alternative option might be to use an replacement approach like KubeSpawner does with the pod_templates / dns_templates. Although it would make it less powerful and configurable.

Setups with `c.JupyterHub.internal_ssl=True` and strict certificate verification.

---
Please let me know what you think about it, and if you would like to have a PR for this.

kreuzert:
closes #5170

### DIFF
--- a/jupyterhub/spawner.py
+++ b/jupyterhub/spawner.py
@@ -1425,8 +1425,8 @@ class Spawner(LoggingConfigurable):
 
     trusted_alt_names = List(Unicode())
 
-    ssl_alt_names = List(
-        Unicode(),
+    ssl_alt_names = Union(
+        [List(Unicode()), Callable()],
         config=True,
         help="""List of SSL alt names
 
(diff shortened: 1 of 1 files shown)


minrk:
Perfect, thank you!

PR #5172, opened 2025-10-10
changelog for 5.4.1

minrk:
add add versionchanged for #5171

I'll wait until some time next week to see if more issues come in, but I want to get #5169 and #5171 out pretty soon.


consideRatio:
I can make time to get z2jh patch release out today if we get this out the door - do you want to go for it or or wait a bit further @minrk?

minrk:
going to publish this now, a z2jh release would be nice, but no rush. The fixes are small.

PR #5173, opened 2025-10-11
Add Server Start Time and Profile Used columns to admin dashboard

dragz:
- Added 'Server Start Time' column with sorting capability
- Added 'Profile Used' column to display server profiles
- Shows relative time format (e.g., '2 hours ago') or 'Never' for start times
- Displays profile names or 'Default' when no profile specified
- Updated table headers with appropriate sorting functionality
- Adjusted CSS column widths for new columns
- Updated colspan values to accommodate additional columns
- Added comprehensive test coverage (7 new tests)
- Enhanced test data with realistic profile information
- Maintains responsive design and existing functionality

Resolves need for more detailed server information in admin interface.

### DIFF
--- a/jsx/src/components/ServerDashboard/ServerDashboard.test.js
+++ b/jsx/src/components/ServerDashboard/ServerDashboard.test.js
@@ -763,7 +763,7 @@ test("Shows 'Default' when no profile is specified", async () => {
   });
 
   let profileCell = screen.getByTestId("user-row-profile-used");
-  expect(profileCell.textContent).toBe("Default");
+  expect(profileCell.textContent).toBe("");
 });
 
 test("Shows 'Never' for server start time when server never started", async () => {


dragz:
Screenshot of the changes can be found on my [post on discourse](https://discourse.jupyter.org/t/adding-more-info-to-the-admin-dashboard/37999).

dragz:
Forgot to say that I used Claude Sonnet to make these changes.

dragz:
> I've managed to get a negative last activity
> 
Last_activity should not be affected by this PR.

manics:
Sorry, I've tested it properly now:

image

Original UI:
image

My first impression is the admin interface is looking a bit crowded and confusing. This isn't your fault since you've extended the existing UI in a natural way, but if we want to add this I think we should consider whether we can improve the UI.

minrk:
I do think it's worth restructuring the columns. `admin` is discouraged in general, now that we have RBAC (though still commonly used), and the vast majority of deployments do not use named servers, so having two columns of 'admin' and 'server name' dedicated to what is generally empty isn't the most efficient use of space.

But some of this info needs to go _somewhere_. We could use the `user[/server]` slug in the single 'user' column.

I agree that it's a bit odd for profile name to be included when JupyterHub itself doesn't understand it as a concept, nor do most spawners, even though the most commonly _used_ spawners do). But there isn't really a mechanism to extend this other than build your own dashboard (and since the admin page is purely using the REST API, there is truly nothing on the admin page that can't be accomplished through a Service). At the same time, I do think it makes sense to promote a version of profiles to JupyterHub itself: https://github.com/jupyterhub/jupyterhub/issues/2724 though we haven't done it _yet_.

dragz:
I've taken the liberty to shuffle things around a bit. (Slow Sunday with bad weather.).
image

minrk:
@dragz since you're reorganizing things, do you want to drop the admin column? That info is already available under the user dropdown, so I think we can lose it.

When I click the sort buttons on server start time, I get an error. This is because it makes a request with:

```
&sort=-started
```
which isn't valid. The only valid sort fields are "name" and "last_activity", so the server start column shouldn't have sort controls.

erawn:
I took a stab at cleaning up the UI and re-writing the components in a more bootstrap-friendly way.

Code is here: [https://github.com/erawn/jupyterhub/tree/more_admin_info](https://github.com/erawn/jupyterhub/tree/more_admin_info) 

- fixed sizing issues when page was shrunk
- header is now correctly aligned with the table 
- fixed dark mode issues with new code 
- removed dead CSS
- fixed alignment of sort icon
- updated the tests for spelling changes in labels 

New UI looks like this: 
Screenshot 2025-11-06 at 3 18 08 PM

dragz:
Can this be merged with my suggestion for moving the buttons around?

minrk:
Thanks, @dragz and @erawn!

PR #5182, opened 2025-10-24
Require Python 3.10

minrk:
first step towards 6.0

unblocks #4998


rgaiacs:
Do we need `requirements.txt` when we have `pyproject.toml`?

minrk:
We don't need it. I just think it's easier to read/find/use.

manics (approved):

consideRatio (approved):
Looks fine to me!

My understanding is that this change by itself won't require a z2jh major version bump since we are using Python 3.10+ already.

minrk:
@consideRatio exactly right. We will likely do some small changes that will affect z2jh by the time we get to a release, though, like #4936.

PR #5191, opened 2025-11-01
`headers` does not exist in `web.HTTPError`

manics:
[`tornado.web.HTTPError`](https://www.tornadoweb.org/en/stable/web.html#tornado.web.HTTPError) only has fields `status_code`, `log_message`, `reason`.

If you manage to trigger an error this leads to an `AttributeError: 'HTTPError' object has no attribute 'headers'` instead.

@agoose77 You added this in https://github.com/jupyterhub/jupyterhub/pull/5146 . Was your intention to catch some other Exception type instead of HTTPError?

### DIFF
--- a/jupyterhub/handlers/pages.py
+++ b/jupyterhub/handlers/pages.py
@@ -222,6 +222,12 @@ class SpawnHandler(BaseHandler):
             if isinstance(spawn_exc, web.HTTPError):
                 self.set_status(spawn_exc.status_code)
 
+                # web.HTTPError doesn't define headers,
+                # but custom exceptions may set them.
+                headers = getattr(spawn_exc, "headers", None) or {}
+                for name, value in headers.items():
+                    self.set_header(name, value)
+
             if spawn_exc:
                 error_message, error_html_message = format_exception(spawn_exc)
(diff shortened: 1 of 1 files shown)


minrk:
@agoose77 any recollection of what the intention was here? Or if we should just delete these accesses to undefined attributes?

agoose77:
JupyterHub seems to use the error namespace to attach headers to exceptions, as an out-of-band mechanism — the `headers` field is indeed not defined in the base class: https://github.com/jupyterhub/jupyterhub/blob/8696374ca09dc9cac9f3f5425ea5919f98d78222/jupyterhub/handlers/base.py#L1542-L1550

@minrk I'm trying to cast my mind back to this PR. I suspect I was specifically interested in the `Retry-After` or perhaps `Content-Type` headers.

Therefore, I *think* the incremental fix here is to test for `headers` like we already do elsewhere: https://github.com/jupyterhub/jupyterhub/blob/8696374ca09dc9cac9f3f5425ea5919f98d78222/jupyterhub/handlers/base.py#L1545-L1546

minrk:

Issue #5054, opened 2025-04-23
Breaking: Restrict servernames by default


manics:
### Proposed change
Currently JupyterHub named servers are named by the user, and it is not possible to restrict the characters used. For example, it is possible to create a name containing linebreaks.

We currently have three properties related to named servers:
- `JupyterHub.named_server_limit_per_user`
- `JupyterHub.allow_named_servers`
- `JupyterHub.default_server_name`

We could add a new configurable `JupyterHub.named_server_regex` defaulting to something sensible, e.g. `[\w_\.\-\+\~, ]*` (any unicode character, some punctuation, space). We could optionally prevent multiple consecutive spaces to avoid ambiguities when displaying names since consecutive white space is squashed in HTML.

### Alternative options
Add the configurable, but set the default to `.*` so that this isn't a breaking change.

This could be done at the Spawner level, but it would result in an initialised spawner that returns an error when `start()` is called. Enforcing it with `traitlets.validate` in the Spawner may result in the hub creating a named server Spawner that can't be deleted, for example if [`delete_forever()` reads the name](https://github.com/jupyterhub/kubespawner/blob/7.0.0/kubespawner/spawner.py#L3789-L3805).

### Who would use this feature?
Implementers of spawners or extensions that use the server name. JupyterHub UI components that display the server name.

minrk:
Sorry I wasn't thorough on this, it was silly to not have a fairly strict pattern here. There's no need to allow arbitrary server names. I'd actually support being _more_ strict with the name pattern (something that must be url-safe, including with a min and max length), where the pattern is _not_ configurable, and:

1. reject _new_ servers that don't match the strict pattern, 
2. warn about existing servers
3. provide a path to rename named servers to safe patterns (this may be hard?)

manics:
1 and 2 are fine. I can't think of a nice way to do 3 though. We could add a new hub API endpoint and UI to rename servers, but all spawners would need to support renaming or storing a map of existing resources such as storage volumes which isn't always possible.

minrk:
closed by #5192

PR #5192, opened 2025-11-02
Restrict named server servernames

manics:
See https://github.com/jupyterhub/jupyterhub/issues/5054
- Limits servernames to alphanumeric characters, and single `-`. See documentation in `slugs.py::safe_slug` or `howto/upgrading-v6.md` for details
  - Adds a separate display name with fewer restrictions since this is only for display purposes
- Existing servers can still be started
- [x] Add docs on how to migrate (probably increase the number of allowed named servers, start a new server, ask an admin to migrate the storage, and delete the old server? Or should we support starting legacy named servers indefinitely?)


jupyterhub/utils.py
```diff
@@ -956,6 +957,24 @@ def subdomain_hook_idna(name, domain, kind):
     return f"{safe_name}{suffix}.{domain}"
 
 
+# A URL safe identifier can be used in URL paths without escaping:
+# Unicode letters, digits, _, -
+_safe_unicode_identifier_re = re.compile(r'^[\w-]{1,63}$')
```
manics: What's a good size limit? If we want something that can be converted to a DNS safe string and back then we need to take account of unicode being escaped. Or we could stick with the KubeSpawner approach of using a hash if the string is too long?

minrk: if we do that, the 1123 check in kubespawner is [here](https://github.com/jupyterhub/kubespawner/blob/0020c0b39e4e8c9b072f91b2875af18b6c8f57ee/kubespawner/slugs.py#L65).

manics (commented):
I've had a go at adding a `display_name` field to the database, but it's ended up being more complicated than I anticipated and there some design decisions to make.

## Safe slugs

I copied `kubespawner/slugs.py` to `jupyterhub/slugs.py`, then realised it generates safe non-clashing names for combined username+servernames labels, e.g. it uses `---`. When naming a server we don't want this, since we may need to concatenate the servername later.

For now I went for something simpler, just `[alnum]` with an optional `-` in the middle. I haven't put a length limit yet, but if we set the length to `30` in `safe_slug` then in future we it would be fine to generate a DNS-safe label using `f"{safe_slug(username)}---{safe_slug(servername)}"`.

## SpawnerDict

This used to assume that if the ORM spawner doesn't exist you can just create it using `servername`. This is no longer the case, since `display_name` can be passed. To make things more explicit I've change dit so the SpawnerDict can be used to get a spawner if the ORM spawner exists, otherwise you must create the spawner using `User.get_or_create_spawner(server_name, display_name)`. I could make `display_name` default to `server_name` so that the SpawnerDict will still work as before, but I decided it's better to be explicit.

## Link between servername and display_name

The display name is actually completely independent of the server name. I've changed the Spawn UI so that the name entered on the form is treated as the display name, which is then converted to a servername using safe_slug. However this isn't strictly necessary, servername could be any safe_slug. If you use the API there's no requirement for `servername == safe_slug(display_name)`

minrk:
I don't think we need to avoid collisions in slugs, since they exist in a fully user-controlled namespace. I think we can do a simpler strip/normalize and then just reject it if there's a collision - we don't need to allow users to have "My Server" and "my_server" and "my-server" all at the same time.  Those can normalize to the same thing, and it's up to the user to pick a name that doesn't collide after normalization.

In the special case of the db upgrade where those collisions could _already_ exist, we can append `-2` or something. That still works because the slug is computed once, then it's the 'true name', we don't need to always produce the same thing, and can update normalization over time without consequences.

minrk:
For SpawnerDict, I think we can make a `get_or_create` method. I don't think we need the defaultdict behavior, it may have been a little too much automatic logic anyway.

> Link between servername and display_name

Yeah, it's a totally free association. I'd probably leave the UI with a single display name as input and derive the true name via slugification, but let the API freely define the two.

manics:
After thinking about this for a while I'm wary of automatically renaming servers in case a spawner doesn't record the state of their dependent resources like volume mounts. For now I'm taking the approach that:
- new named servers must conform to the new restrictions
- old named servers created before JupyterHub 6.0 can be started/stopped, but a warning is shown
- the backend (Python) validates new server names but does not attempt to sanitise them, this must be done by the front-end or API caller

Example of old named servers with invalid names
image

New named servers with display name that is separate from server name:
image

minrk:
That's a good point, though doing that means that we can't actually _assume_ names are safe anywhere, we have to keep everything that treats names as arbitrary strings everywhere. Can you think of an opt-in way to _guarantee_ that strict names are followed? Or perhaps a multi-stage deprecation, so it's a warning in jupyterhub 6, but error in jupyterhub 7?

manics:
I agree we don't want to support unsafe names indefinitely. Do you think we could add a flag to `jupyterhub upgrade-db` to auto-update names, have constant warnings for anyone who doesn't update them, and then make this a hard failure in JupyterHub 7?

minrk:
because upgrade-db runs precisely once, I think putting arguments on it is not likely to be super useful, because there's no way to re-run with different arguments. I think we'd need a hook somewhere to be able to do it after upgrade, either as a one-time hook or startup flag or something.

I think the hard failure in JupyterHub 7 can be that we don't allow starting the servers with mismatched names until they've been renamed, don't allow any operation on them other than delete & rename.

We could potentially do exactly that in JupyterHub 6: allow users to rename (and only allow safe names), but don't rename it for them, and allow admins to bulk auto-rename, if they want, just not at `upgrade-db` time. And prohibit _starting_ named servers with unsafe names, which means that the only APIs that need to deal with unsafe names are listing and renaming.

docs/source/howto/upgrading-v6.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Named server restrictions
+
+Prior to JupyterHub 6 there were almost no restrictions on the naming of named servers.
+Core JupyterHub components correctly handled these, but names with unsual characters caused problems for some extensions, or when interacting with external components.
+
+JupyterHub 6 imposes strict limits on the naming of named servers:
+
+- Between 1 and 32 characters
+- Consists of lowercase ASCII letters, digits, and zero or one hyphen (`-`)
```
minrk: why the limits on number of hyphens? And is this total or consecutive?

docs/source/howto/upgrading-v6.md
```diff
@@ -0,0 +1,85 @@
...
+- Must start with a lowercase letter
+- Must end with a lowercase letter or digit
+
+A new field `display_name` means users can still provide an arbitrary name, separate from the server name.
```
minrk:
```suggestion
A new field `display_name` means users can still provide an arbitrary name, separate from the server name used in APIs, URLs, etc.
```

docs/source/howto/upgrading-v6.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Web interface
+
+Users can name a named server as normal.
+This will be used as the display_name, and if this doesn't comply with the above restriction a server name will be automatically generated.
```
minrk:
```suggestion
This will be used as the display_name, and if this doesn't comply with the above restriction, a safe server name will be automatically derived by stripping disallowed characters, truncating, as needed.
```

minrk (commented):
What do you need in order to finish this one? I think this is the main thing I would like to have finished to start preparing 6.0.

docs/source/howto/upgrading-v6.md
manics: I think this is out of date, it should be single hyphens are allowed.

manics:
Let me rebase and check everything I've done, but I think all the main parts are in place (creating slugs, optionally updating names in the database).

jupyterhub/app.py
```diff
@@ -1367,6 +1369,26 @@ def named_server_limit_per_user_fn(handler):
         """,
     ).tag(config=True)
 
+    allow_existing_invalid_named_servers = Enum(
```
manics, line 1372: Naming is hard, happy to hear suggestions for alternatives!

jupyterhub/app.py
```diff
@@ -3287,6 +3309,9 @@ def init_tornado_settings(self):
             allow_named_servers=self.allow_named_servers,
             default_server_name=self._default_server_name,
             named_server_limit_per_user=self.named_server_limit_per_user,
+            allow_invalid_named_server_start=(
+                self.allow_existing_invalid_named_servers == "allow-start"
+            ),
```
manics, line 3314: I've named this tornado setting different;y from the Traitlet because the Application uses a multi-valued enum but the rest of the hub only cares about whether it's `allow-start` or not. Alternative is to pass the Traitlet through and deal with the enum later.

jupyterhub/app.py
```diff
@@ -3338,6 +3363,67 @@ def init_eventlog(self):
...
+        cant_modify = []
+
+        for spawner in (
+            self.db.query(orm.Spawner)
+            .filter(orm.Spawner.name != "")
+            .options(selectinload(orm.Spawner.user))
+        ):
+            if not is_valid_safe_slug(spawner.name):
+                if spawner.display_name:
+                    # Is this excessively defensive?
```
manics: Should I delete this? Should be impossible.

jupyterhub/app.py
```diff
@@ -3338,6 +3363,67 @@ def init_eventlog(self):
...
+                    "Renaming server user='%s' server='%s' to '%s'",
+                    spawner.user.name,
+                    spawner.name,
+                    safe_name,
+                )
+                spawner.display_name = spawner.name
+                spawner.name = safe_name
+
+            self.db.commit()
+            # TODO: Do we need to handle errors and rollback?
```
manics: Please check!

jupyterhub/app.py
```diff
@@ -3338,6 +3363,67 @@ def init_eventlog(self):
...
+
+            if not to_modify and not cant_modify:
+                self.log.info("All named servers have valid names")
```
manics: Should this be logged at info or debug?

share/jupyterhub/static/js/utils.js
```diff
@@ -122,6 +122,65 @@ define(["jquery"], function ($) {
     modal.show();
   };
 
+  var safe_slug = function (name) {
```
manics, line 126: Since this is (currently) only used in the front end for creating a servername we can use any algorithm, even a random string. For consistency I've reimplemented `jupyterhub/slugs.py::safe_slug` in JS, but the main downside is it adds a dependency on the `js-sha256` NPM package since SHA256 isn't available in standard JS.

jupyterhub/slugs.py
```diff
@@ -0,0 +1,274 @@
+"""Tools for generating slugs like object names and labels
```
manics, line 1: Admittedly this is scope creep, but since we need a way to generate safe names I thought I'd write something generic that should hopefully work for escaping other names in future, including in external authenticators and spawners.

karma.conf.js
```diff
@@ -0,0 +1,58 @@
+// Karma configuration
+// Generated on Sat May 30 2026 20:05:11 GMT+0100 (British Summer Time)
```
manics, line 2: This file, and `test-main.js`, is the output of `karma init` with a few changes.

I looked at using Jest since that's what we're using in jsx, but couldn't get it to work with requirejs/AMD modules.

package.json
```diff
@@ -13,15 +13,21 @@
     "css": "sass --style compressed -I share/jupyterhub/static/components --source-map share/jupyterhub/static/scss/style.scss:share/jupyterhub/static/css/style.min.css",
     "build:watch": "npm run css -- --watch",
     "jsx:install-run": "npm install --prefix jsx && npm run --prefix jsx",
-    "jsx:run": "npm run --prefix jsx"
+    "jsx:run": "npm run --prefix jsx",
+    "test": "karma start"
```
manics, line 17: This is potentially confusing to new contributors because this is a top-level `package.json` that runs tests, but we also have a completely independent `jsx/package.json` that also runs tests but is not triggered by this one.

manics (commented):
I think this is ready for review!

I've highlighted a few things/questions inline

One missing nice-to-have is a browser/playwright test to check a warning is shown for invalid servers on the home page:
Image

Preview of the 6.0.0 upgrade page: https://jupyterhub--5192.org.readthedocs.build/en/5192/howto/upgrading-v6.html

manics:
If you want to test the handling of existing invalid servers without first installing JupyterHub 5 you can import this DB dump instead:
[jupyterhub5db.sql.txt](https://github.com/user-attachments/files/28445574/jupyterhub5db.sql.txt)

```sh
rm jupyterhub.sqlite
sqlite3 jupyterhub.sqlite '.read jupyterhub5db.sql.txt'

jupyterhub upgrade-db
```
`jupyterhub_config.py`:
```py
c.JupyterHub.spawner_class = "simple"
c.JupyterHub.authenticator_class = "dummy"
c.JupyterHub.allow_named_servers = True

## Set this to auto migrate:
#c.JupyterHub.allow_existing_invalid_named_servers = "autorename"
```

Run JupyterHub, login as user `demo`. You should have three named servers with invalid names.

jupyterhub/app.py
minrk: probably debug

jupyterhub/app.py
minrk: I think it's fine since it's already here. "This shouldn't happen" checks do trigger once in a while, especially in unrealistic tests. And it will be rare, since post-upgrade this check will soon stop happening.

jupyterhub/app.py
minrk: I think that makes sense

jupyterhub/app.py
minrk: Good checking! No, we don't generally need to deal with rollback during one-tie startup checks.

jupyterhub/app.py
```diff
@@ -3338,6 +3363,67 @@ def init_eventlog(self):
...
+
+        elif cant_modify or to_modify:
+            self.log.warning(
```
minrk, line 3431: Great!

minrk (commented):
I think this is really great, and really appreciate the detailed docs.

My only point of disagreement right now is using the generic collision-safe hashing scheme on server names, which don't have any of the the challenges that motivated using it in kubespawner for resource names, because they are already namespaced per-user, so don't have the same security/collision consequences.

It doesn't seem right to me that `two-word` is not considered a safe server name. I really think single-hyphen should be allowed in safe names. I understand the reasoning that it should be valid to combine two slugs with e.g. `--`, but server names do not have the same collision issues that e.g. usernames have, so I don't think there's a reason to use the hashing scheme in server names.

Adding hashes also increases the likelihood of needing _further_ truncation and hashing due to increasing the length of simple strings.

I think if we make the collision-avoidance hash _optional_ and opt-out for server names, that would work for me

Then we can have:

```python
safe_slug("a-b") # a-b-c8687a08
safe_slug("a-b", avoid_collisions=False) # 'a-b'
```

and use the latter for server names, and the former where collisions must be avoided.

manics:
I changed safe_slug as you suggested and removed the hash from servernames form back and front ends, the exception is if there's an actual clash in the autorename process.

One problem is I haven't done any checks for collision on the front end because there's no easy way to get the list of servernames without making an additional API request. If the entered name for a new server corresponds to a safe_name that clashes with an existing server the user is taken to that existing server. The behaviour of the spawn request is technically unchanged from before this PR, but it's now a problem because we're silently converting the user supplied name to the safe_name.

minrk:
That makes sense, I don't think the frontend should check for collisions. This reveals a problem in our API that creating a named server and starting it for the first time are the same, where it should actually be two steps:

1. create the server (this should 409 on name collision), _then_
2. start the new server

I think we should land this and then look into improving that sequence by adding a new API endpoint that only creates the server instead of starting it.

manics:
Thanks, I think this is ready then! Be aware that if there are problems after merging I won't be able to do any follow-up work for the next 3 weeks.

minrk:
Thanks for all your work on this!

PR #5195, opened 2025-11-05
allow callable default_url to return a falsy value to mean the default behavior

minrk:
allows callable returning falsy value to behave the same as scalar default_url with falsy value

useful for e.g. custom default url for only some users:

```python
def default_url(request):
    if request.current_user and request.current_user.admin:
        return "/hub/admin"
    else:
        return ""
```


minrk:

PR #5197, opened 2025-11-07
Add noxfile.py

rgaiacs:
Related to https://github.com/jupyterhub/team-compass/issues/819

It adds a copy of https://github.com/jupyterhub/team-compass/blob/693186ae18bfa78e192e38efc982543770069bde/noxfile.py with minor modifications.


minrk:
Thanks!

PR #5213, opened 2025-12-01
fix linkcheck

minrk:
- remove clemson from gallery
- medium blocks with 403


minrk:
self-merging to fix ci

minrk:

PR #5225, opened 2025-12-02
point to testing config sooner

minrk:
can't login by default to jupyterhub with no args


minrk:

PR #5236, opened 2025-12-11
fix loading stopped servers at startup (regression in 5.4)

### LINKED ISSUE #5211: Running user servers at Hub startup cause incomplete ORM user loading
When starting JupyterHub with running Servers the JupyterHub App itself does not load all orm_spawner objects correctly.

In [here](https://github.com/jupyterhub/jupyterhub/blob/bba81a856c958366c18f54a6a91a4cc48934b009/jupyterhub/app.py#L3096C1-L3108C11) it throws out all non-running orm_spawners from orm_user. 

### How to reproduce

1. Fresh JupyterHub installation 5.4.0-5.4.2
2. Use config with cleanup_servers = False like
```
from jupyterhub.auth import DummyAuthenticator
from jupyterhub.spawner import SimpleLocalProcessSpawner

c.JupyterHub.spawner_class = SimpleLocalProcessSpawner

c.JupyterHub.cleanup_servers = False
c.JupyterHub.allow_named_servers = True
```
3. Browse to JupyterHub, login as some user. Start two named servers "first_server" and "second_server". 
4. Stop Jupyter Server called "first_server".
5. Stop JupyterHub

Let's take a moment to check our database:
```
$ sqlite3 -csv jupyterhub.sqlite "SELECT id, user_id, name FROM spawners ORDER BY user_id, id;"
1,1,""
2,1,first_server
3,1,second_server
```

6. Restart JupyterHub
7. Login with same username as in Step 3: **You only see "second_server" in the frontend**

Image

Let's check the database again:
```
$ sqlite3 -csv jupyterhub.sqlite "SELECT id, user_id, name FROM spawners ORDER BY user_id, id;"
1,1,""
2,1,first_server
3,1,second_server
4,1,""
```
There's also a second entry for the unnamed orm_spawner, created in [app.py](https://github.com/jupyterhub/jupyterhub/blob/bba81a856c958366c18f54a6a91a4cc48934b009/jupyterhub/app.py#L3111) , [user.py](https://github.com/jupyterhub/jupyterhub/blob/5.4.0/jupyterhub/user.py#L242)

> Therefore, users with running Jupyter Servers "lose" access to all non-running spawners.   
  
> It recreates an additional orm_spawner without a Name at every restart for each user containing a running Jupyter Server

### Suggested Solution
I'm not a database or sqlalchemy expert, but switching back to the solution in 5.3.x works for me in these simple tests.

minrk:
ensures all servers are loaded for matching users, where contains_eager excludes them

and regression test

I want to make 5.4.3 with this promptly, since it's a serious regression in 5.4

closes #5211

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -3100,14 +3100,14 @@ class JupyterHub(Application):
             .filter(orm.Spawner.server != None)
             # pre-load relationships to avoid O(N active servers) queries
             .options(
-                # needs to be joinedload or selectinload, not contains_eager to avoid excluding stopped servers servers from the db session
-                # testing suggests selectinload is slower, but uses less memory than joinedload
-                # for large numbers of servers
+                # needs to be joinedload or selectinload,
+                # not contains_eager to avoid excluding stopped servers servers from the db session
+                # avoid joinedload on user_id which is degenerate,
+                # so selectinload it is
                 # make sure server->user relationship is loaded
                 selectinload(orm.Spawner.user),
                 # make sure users' _other_ spawners are also loaded
                 selectinload(orm.Spawner.user, orm.User._orm_spawners),
-                selectinload(orm.Spawner.server),
             )
             .populate_existing()
         ):


minrk:
ignoring flaky test fixed by #5238

minrk:

PR #5238, opened 2025-12-11
browser test_spawn_pending_progress : fix server ready expectation

minrk:
so it gets the right url when subdomains are used

fixes flaky test_spawn_pending_progress with subdomains enabled


minrk:

PR #5261, opened 2026-01-06
Fix command to check Node.js version in setup.md

LUC4R4T0R:
the command mentioned previously (`npm -v`) only returns the version of npm instead of node.js. As the most recent version of npm is 11.7.0 and node.js > v12 is required, some people might wonder about not beeing able to fulfill this version-requirement.


rgaiacs (approved):
Thanks @LUC4R4T0R for your contribution. It looks good to me.

PR #5287, opened 2026-02-13
reduce coverage for mysql/postgres tests

minrk:
these don't need to run every single test, so use `mark.db` and `mark.ssl` to mark tests

can cast a wide net, but nice to exclude some large files that are slow and not relevant to the difference

downside: we need to make sure to apply relevant marks to avoid missing test coverage

upside: saves time, since our tests are getting big and slow

### DIFF
--- a/jupyterhub/tests/test_spawner.py
+++ b/jupyterhub/tests/test_spawner.py
@@ -40,7 +40,7 @@ while True:
         print("interrupted")
 """
 
-pytestmark = pytest.mark.db(pytest.mark.ssl)
+pytestmark = pytest.mark.ssl
 
 
 def setup():


minrk:
oops, already had that 

I think the db failures might be relating to timing due to the new subprocess coverage. Increasing the stop timeout might help there.

minrk:
actually, I think the db failure is due somehow to `test_singleuser_spawner` being the first to launch a spawner subprocess, and CI being incredibly slow to warm up. Increasing the timeout solves it. The problem appears to have been caused by the fact that now the test to assert the clean exit is the first launch, where it came after `test_api` launched the process many times.

I looked through the file, and it doesn't seem to really help with db variant coverage, so I removed it.

PR #5292, opened 2026-02-17
Add server spawn counter mertrics for success and failure

### LINKED ISSUE #5258: new metrics: resolved spawn attempts counter
We have running servers, but that doesn't show if user servers are starting and stopping etc. Maybe we could have a dedicated counter metric that increments for resolved spawn attempts with a outcome label indicating success/failure or similar?

anny0811:
"Fixes #5258: This PR adds a Prometheus counter to track server spawn successes and failures directly, making health monitoring simpler."

### DIFF
--- a/jupyterhub/metrics.py
+++ b/jupyterhub/metrics.py
@@ -432,9 +432,8 @@ class PeriodicMetricsCollector(LoggingConfigurable):
             )
 
             self.log.info(f'Found {value} active users in the last {period}')
-            ACTIVE_USERS.labels(period=period.value).set(value)
-
-    async def _measure_event_loop_interval(self):
+            ACTIVE_USERS.labels(period=period.value).set(value) 
+            def _measure_event_loop_interval(self):
         """Measure the event loop responsiveness
 
         A single long-running coroutine because PeriodicCallback is too expensive


manics:
Hi! Please could you update the description to explain the purpose of this PR, and why it's needed? Could you also say how you've tested it? 

Just out of interest, did you use an LLM to generate some or all of the changes?

minrk:
I would also like to hear from @anny0811 but I assume this is aimed at #5258 where I suggested we already have that metric, as far as I understand.

anny0811:
Hi @minrk @manics, 
I'm focusing on #5258 because a direct SERVER_SPAWN_TOTAL counter is much simpler for monitoring than aggregating duration data.
Regarding the LLM, I used to help navigate the codebase, but I've manually reviewed and implemented the logic myself. I've also synced the formatting fixes from the CI now.

manics:
@anny0811 How did you test your changes?

anny0811:
I tested this locally by triggering spawns and verifying that the /metrics endpoint correctly increments SERVER_SPAWN_TOTAL for both success and failure cases. Logs were also checked for any new errors.

minrk:
Thank you. As I mentioned in #5292, I believe we already have this metric in `jupyterhub_server_spawn_duration_seconds_count`. Does this new metric count something we don't have already?

PR #5298, opened 2026-02-27
Add ability to assign groups to users in SharedPasswordAuthenticator

yuvipanda:
Groups are an underused and powerful JupyterHub facility, and I'd generally like us to lean into it more.

The primary use case for SharedPasswordAuthenticator is workshops. Workshops will have multiple groups of people present, and it would be great to assign groups to them to differentiate who can access what. This becomes particularly more useful as GPUs become more mainstream, and are much more expensive than regular CPU usage.

This PR adds `password_groups`, which lets admins specify what groups users should be added to when they login with a particular password. This lets one hub be used for multiple workshops, separated via different passwords that can provide users with different profiles / rights.

TODO:
- Validate password length limits for groups
- Handle overlap of passwords with user password and admin password
- Handle how admins may be able to get groups

### DIFF
--- a/jupyterhub/authenticators/shared.py
+++ b/jupyterhub/authenticators/shared.py
@@ -1,6 +1,6 @@
 from secrets import compare_digest
 
-from traitlets import Unicode, validate
+from traitlets import Dict, List, Unicode, validate
 
 from ..auth import Authenticator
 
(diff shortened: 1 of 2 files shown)


ateucher:
Oh I love this @yuvipanda, fully endorse! This will be great for when we have simultaneous workshops, and also tightly sequenced workshops. We currently have to ask 2i2c to do very quick turnarounds when they are close together. This would alleviate that pressure and we can have a normal per-workshop sequence of setting the password prior, changing after a certain length of time post-workshop, and clearing out the user directories for a particular group.

ericvd-ucb:
Indeed this is a great idea for Workshop / Hackweek deployments - which seem like a big constituency in Hub usage

minrk:
Nice!

shaneknapp:
this is freaking great!  thanks @yuvipanda !

minrk:

PR #5312, opened 2026-03-10
update and pin redocly lint

minrk:
it started to fail due to changes in openapi property validation

pins to current version and fixes the lint it raised


minrk:
sphinx 9 builds will be fixed when there's a release of autodoc-traits with https://github.com/jupyterhub/autodoc-traits/pull/79

minrk:

PR #5310, opened 2026-03-06
Add userinfo to User, use it in jupyterhub-singleuser

kreuzert:
Just a contribution for discussions in  https://github.com/jupyterhub/jupyterhub/issues/5061  ( and the proposed change in https://github.com/jupyterhub/jupyterhub/issues/5309 ).

Feel free to close if that does not go in the desired direction.

Usage like
```
def my_hook(authenticator, handler, auth_model):
    if auth_model['user_info'] is None:
        auth_model['user_info'] = {}
    # Authenticator should provide this information during login, like OAuthenticator
    auth_model['user_info']['name'] = 'Tim Kreuzer'
    auth_model['user_info']['initials'] = 'TK'
    auth_model['user_info']['display_name'] = 'Display'
    auth_model['user_info']['avatar_url'] = 'https://avatars.githubusercontent.com/u/28051928'
    auth_model['user_info']['color'] = 'blue'
    return auth_model

c.JupyterHub.authenticator_class = DummyAuthenticator
c.DummyAuthenticator.post_auth_hook = my_hook
```

JupyterLab will pick up the avatar_url in the top right corner, other extensions might use the additional information provided.

image

Issue #5309, opened 2026-03-06
Improve jupyterhub-singleuser User Identity Provider


kreuzert:
### Proposed change
Jupyter Server Extensions use the IdentityProvider to show user information. One example is the Jupyter Real Time Collaboration Extension, showing name, username, [display_name](https://github.com/jupyterlab/jupyter-collaboration/blob/c750ddd1ac2fcebbeaab1aaf261930a2e475ae27/packages/collaboration/src/components.tsx#L44)  and more information.

<img width="1174" height="649" alt="Image" src="https://github.com/user-attachments/assets/a70168b5-10b0-4788-9732-b9d0157897ec" />
  
  
In our JupyterHub environment we're using the voperson_id given by an external AAI as username, which is not really user-friendly to show. In `jupyterhub-singleuser` it's using `class JupyterHubUser` [link](https://github.com/jupyterhub/jupyterhub/blob/48c21ebdacf0b50e4e538c4acaf447625a0200c5/jupyterhub/singleuser/extension.py#L80) [link2](https://github.com/jupyterhub/jupyterhub/blob/main/jupyterhub/singleuser/extension.py#L196) to get user information. This class only sets the username [link](https://github.com/jupyterhub/jupyterhub/blob/48c21ebdacf0b50e4e538c4acaf447625a0200c5/jupyterhub/singleuser/extension.py#L90). 
  
  
I'd suggest that we check the user model (or auth_state if available) to set "name", "display_name", "initials", "avatar_url" and "color", while keeping JupyterServers `username` as `username=self.hub_user["name"]`. 

> Setting name and display_name can be a real upgrade for user experience in these extensions. Inititals, avatar_url and color might be optional and quite specific.

Jupyter-Server uses the fields like that:
```
    # these fields are filled from username if not specified
    # name is the 'real' name of the user
    name: str = ""
    # display_name is a shorter name for us in UI,
    # if different from name. e.g. a nickname
    display_name: str = ""
```

### Who would use this feature?
Every JupyterHub instance that starts jupyter-servers using extensions relying on these User attributes.

### (Optional): Suggest a solution
I will prepare a PR, but need your feedback on how this information should be stored in JupyterHub.

Paul2708:
Maybe it's related to https://github.com/jupyterhub/jupyterhub/issues/5061?

kreuzert:
> Maybe it's related to [#5061](https://github.com/jupyterhub/jupyterhub/issues/5061)?

[...]

Issue #5061, opened 2025-04-30
Support updatable fields in singleuser user model


minrk:
### Proposed change

Jupyter Server is adding support for users updating fields in the user model: https://github.com/jupyter-server/jupyter_server/pull/1518 mainly targeted at the default Anonymous Chupacabra identity provider. But Jupyter Server has several fields that JupyterHub is very likely to never understand, like cursor color, which should still be settable.

JupyterHub should be able to control which fields are settable, and store the fields it allows. The missing piece is persistence - there isn't currently an obvious place to persist info about users that JupyterHub doesn't understand.

### Alternative options

Don't do this, keep the JupyterHub user model completely read-only, which also leaves some fields as auto-generated.

### Who would use this feature?

Deployments with collaboration enabled.

### (Optional): Suggest a solution

Maybe add a JSON user_info column to the Users table that's editable via the REST API? Many auth providers define fields like a display name, avatar url, etc. and it would be nice to store this so your actual name could show up in Collaboration UI, not just your username as we have now. Unclear how restrictive we should be on the fields, though. `auth_state` can technically serve as this, but I don't think it's the right choice.

If we do that, we'd probably need to expose to the Authenticator some level of 'managed fields', i.e. if the Authenticator specifies your display name, it shouldn't be editable. But if it doesn't, (maybe) it should.

A lighter-weight version is some kind of local or short-term persistence in the single-user server. This might not  last, but could be made to work in many cases with no changes to the Hub itself.

manics:
> I guess the alternative approach is to merge user_info from the db with user_info from the Authenticator somehow, such that the Authenticator-provided fields always clobber what's in the db. But then the failure mode is successfully storing fields that get silently ignored.

That's a much clearer way of stating what I had in mind 😄. It makes it easier to reason about security since user_info is never trusted, and also deals with some edge cases like what happens if an Authenticator is updated that adds new non-editable fields.

On top of that we could add checks when updating user_info to prevent setting fields that are ignored by the Authenticator, but importantly if this goes wrong for any reason it's a UI bug rather than a secvuln.

kreuzert:
Maybe it would be easier to provide that information in the auth_model of JupyterHub, and let the JupyterHub IdendityProvider in `jupyterhub-singleuser` use it.

Authenticators can define these fields like:
```
def my_hook(authenticator, handler, auth_model):
    auth_model['name'] = ...
    if auth_model['user_info'] is None:
        auth_model['user_info'] = {}
[...]

PR #5317, opened 2026-03-23
add user_info to User, use it in jupyterhub-singleuser

kreuzert:
#5310 with correct target branch

=======

Just a contribution for discussions in  https://github.com/jupyterhub/jupyterhub/issues/5061  ( and the proposed change in https://github.com/jupyterhub/jupyterhub/issues/5309 ).

Feel free to close if that does not go in the desired direction.

Usage like
```
def my_hook(authenticator, handler, auth_model):
    if auth_model['user_info'] is None:
        auth_model['user_info'] = {}
    # Authenticator should provide this information during login, like OAuthenticator
    auth_model['user_info']['name'] = 'Tim Kreuzer'
    auth_model['user_info']['initials'] = 'TK'
    auth_model['user_info']['display_name'] = 'Display'
    auth_model['user_info']['avatar_url'] = 'https://avatars.githubusercontent.com/u/28051928'
    auth_model['user_info']['color'] = 'blue'
    return auth_model

c.DummyAuthenticator.post_auth_hook = my_hook
```

JupyterLab will pick up the avatar_url in the top right corner, other extensions might use the additional information provided.

image


kreuzert:
@minrk / @manics 
I've created a new PR targeting the correct branch.
I've added documentation and updated the unit tests (and added one for post auth hook, but that one is probably not needed at all).

Let me know if you need anything else.

minrk (commented):
This looks great! I left some suggestions to make sure we get the versionadded notes in the right places, but otherwise it seems all set.

I want to think through what this looks like if OIDC things should go in here and if that creates any kind of conflict or suggestion of namespacing (I think not), but that's it.

kreuzert:
Tests looking good. Let me know if you need something else, or if you want change things for OIDC.

minrk:
Thank you!

manics:
Ignoring the DB change, should we treat this as a breaking change because it could expose auth information to a user that was previously admin only? Or is it safe because an admin still needs to change something?

PR #5319, opened 2026-03-23
add zizmor to pre-commit, address lint

minrk:
I think @manics has maybe used this the most and might have advice

It's probably most important to run this on release workflow (we try to avoid credentials in most workflows), but the per-file configuration is lacking.

For example, I'd be okay with ignoring the artipacked rule everywhere but release, and allowing ref-pin everywhere but release as well. But the configuration doesn't seem to have the appropriate granularity to run with a different policies per workflow.

What I'd really like is a policy like:

- be super strict on release.yml
- be more lax in test workflows **BUT** keep ensuring that they have no credentials to leak (i.e. enforce `contents:read`)

which is basically what we had, but not enforced. I'm not sure there's an easy way to get this, though.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -17,6 +17,8 @@ updates:
           - "actions/*"
     schedule:
       interval: monthly
+    cooldown:
+      default-days: 7
   - package-ecosystem: npm
     directory: /
     groups:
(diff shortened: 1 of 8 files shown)


manics (approved):
I'm not aware of a way to have fine grained per-workflow config.

Issue #5306, opened 2026-03-04
Authenticator-managed role scopes are deleted when role removed from last user


jezwilkinson:
### Bug description

When using `c.Authenticator.manage_roles` in JupyterHub 5.4.3, if a user login or authentication refresh causes a role to be removed from that user, the role and its scopes are deleted from the database. If JupyterHub tries to re-create the deleted role because of the authenticator granting it to a user, the created role does not include any of its scopes or properties from `load_managed_roles()`. The scopes are only restored to the role by restarting the Hub service.

#### How to reproduce
I am using a modified version of [KeyCloakAuthenticator](https://github.com/swan-cern/jupyterhub-extensions/tree/master/KeyCloakAuthenticator/), which I have extended to add support for applying roles to users based on OIDC claims.

The roles and their scopes are defined in a "load_managed_roles" in the JupyterHub authenticator config:

```
async def load_managed_roles(): 
   return [
   {
      "name": "test-role",
      "description": "Test role (read-only access to admin UI)",
      "scopes": [
         "read:metrics",
         "admin-ui",
         "list:users",
         "read:roles:users",
         ]
   },
]
c.Authenticator.load_managed_roles = load_managed_roles
```

The roles to grant to the user are defined by name in a field of the OIDC token returned by the login provider, and propagated through to `user['roles']` as a list of `{'name': rolename}` in the `authenticate()` and `refresh_user()` functions. This properly associates the role and its scopes to the user on login.

On login or auth refresh, if a role that a user previously had assigned has been removed on the OIDC provider's end, it is removed from them as it should be. However, if that user was the only one in the database that held that role, the role and its scopes/description are also deleted from the database:
https://github.com/jupyterhub/jupyterhub/blob/3b2111af11fe15aad3455ae5ba81ddd1893367d6/jupyterhub/user.py#L394-L400

If the authenticator then tries to re-assign that role to the same or a different user, JupyterHub re-creates the role in its database with a fresh primary key. However, it does not check the entries in `load_managed_roles()` to load the scopes for the role, so the role is created without any scopes. This state persists until the Hub service is restarted, at which point the scopes and description in the DB are restored from `load_managed_roles`.

#### Expected behaviour
Ideally, this process should either:
- Not remove the role from the database when the role is stripped from the last user (i.e. roles defined by `load_managed_roles` are always kept in the DB); or
[...]

PR #5344, opened 2026-04-09
Prevent treating a role as 'stale' if defined in load_managed_roles

jezwilkinson:
This PR is an attempt to address https://github.com/jupyterhub/jupyterhub/issues/5306.

Currently, if a role is removed from a user by `user.sync_roles`, and the role has no other remaining members, then the role and its scopes are deleted from the Hub database. This happens even if the role's scopes are defined in the authenticator's `load_managed_roles`, which can lead to a situation where the Hub needs to be rebooted to restore the role's scopes.

With this commit, if the role is defined in `auth.load_managed_roles`, it will not be considered as 'stale' when being removed from a user (by the same logic as the handling of 'stale' authenticator-managed roles in `app.init_role_creation`), and so its scopes are preserved.

The slight side effect is that the "sync_roles" function has to be redefined as async to be able to call the  `load_managed_roles` coroutine, and so the call in `handlers/base.py` also becomes an `await` - I couldn't find a clean way to avoid this redefinition, but the change seems minimal since it's only called in this one place.

### DIFF
--- a/jupyterhub/user.py
+++ b/jupyterhub/user.py
@@ -392,7 +392,9 @@ class User:
             .all()
         )
         # prevent deletion of role scopes that are explicitly defined in auth.load_managed_roles()
-        predef_managed_roles = [r.get('name') for r in await self.authenticator.load_managed_roles() or []]
+        predef_managed_roles = [
+            r.get('name') for r in await self.authenticator.load_managed_roles() or []
+        ]
 
         for stripped_role in managed_stripped_roles:
             if (
--- a/jupyterhub/user.py
+++ b/jupyterhub/user.py
@@ -392,7 +392,7 @@ class User:
             .all()
         )
         # prevent deletion of role scopes that are explicitly defined in auth.load_managed_roles()
-        predef_managed_roles = [r.get('name') for r in await self.authenticator.load_managed_roles()]
+        predef_managed_roles = [r.get('name') for r in await self.authenticator.load_managed_roles() or []]
 
         for stripped_role in managed_stripped_roles:
             if (
--- a/jupyterhub/user.py
+++ b/jupyterhub/user.py
@@ -301,7 +301,7 @@ class User:
             self.orm_user.groups = []
         self.db.commit()
 
-    async def sync_roles(self, auth_roles):
+    def sync_roles(self, auth_roles):
         """Synchronize roles with database"""
         auth_roles_by_name = {role['name']: role for role in auth_roles}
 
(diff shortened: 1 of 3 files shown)


jupyterhub/user.py
```diff
@@ -391,11 +391,17 @@ def sync_roles(self, auth_roles):
             )
             .all()
         )
+        # prevent deletion of role scopes that are explicitly defined in auth.load_managed_roles()
+        predef_managed_roles = [
+            r.get('name') for r in await self.authenticator.load_managed_roles() or []
```
minrk, line 396: We could probably cache this set of names and persist it through `settings` when it's loaded on startup, rather than recomputing it every time?

jezwilkinson: I think that would work fine as long as it's assumed that `load_managed_roles` will be static over the Hub's lifetime (which would be my use case, simply returning a dict that's defined in `jupyterhub_config.py` in a similar way to `load_roles`).

The possible issue I could see is what happens if an authenticator defines `load_managed_roles` to pull role details from an external source that can change dynamically - in that case the cached list of role names from startup could get out of sync with the "current" roles over time (becomes a design question of how the auth-managed roles are "supposed" to be used, I guess, but the way `load_managed_roles` is implemented leaves the door open to that)

minrk: I think it's sensible to delete roles created by users logging in when no users have it assigned anymore and _prevent_ deleting roles at this time if they were created at startup. if load_managed_roles changes, then at next hub startup the right roles list will be created and that will reflect the latest list of roles to prevent deletion

While a hypothetical situation could occur, I don't expect it to (if the role was created properly when a user logged in the first time, it should be created properly the _next_ time a user logs in with that role as well), and the remediation is to relaunch the hub, which is common already.

We can always revisit this if someone comes with a _real_ situation where that's needed, and which would probably prompt something like continuous, periodic calls to `load_managed_roles`, rather than here.

jezwilkinson: Sorry, had a bit of delay getting back to this - I've now reworked it so that the `config_role_names` in `app.init_role_creation` gets propagated through to `settings` via `app.init_tornado_application`, and then `sync_roles` checks against that instead of re-running `load_managed_roles`.

minrk (approved):
Thanks! Will merge, assuming failure was flaky test (very likely)

minrk:
Thank you!

minrk:

PR #5351, opened 2026-04-28
ci: temporary pin-down cryptography

minrk:
temporary pin to workaround https://github.com/llnl/certipy/pull/27

only in ci, not dependencies, since a certipy release will fix this


minrk:

PR #5362, opened 2026-05-01
Update test to not reload the browser until token result is visible

### LINKED ISSUE #5361: Investigate the source of flaky browser tests
🤣 Now there is another browser test fail. 
image

I'm going to retrigger the failing job and see if it passes. Then I will open an issue to look at these browser tests.

_Originally posted by @willingc in https://github.com/jupyterhub/jupyterhub/issues/5360#issuecomment-4360677362_

willingc:
Closes #5361 

This PR applies a similar pattern to `test_request_token_permissions` as other tests to make sure the token is visible before reloading the browser.

~Another small change to make the locator use relative vs absolute path.~

### DIFF
--- a/jupyterhub/tests/browser/test_browser.py
+++ b/jupyterhub/tests/browser/test_browser.py
@@ -803,6 +803,8 @@ async def test_request_token_permissions(
         await expect(error_dialog).not_to_be_visible()
         return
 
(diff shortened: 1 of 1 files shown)


minrk (approved):
nice!

PR #5363, opened 2026-05-01
browser tests: increase default expect timeout to 30s

### LINKED ISSUE #5361: Investigate the source of flaky browser tests

minrk:
I believe the 5s timeout is causing flaky tests because everything is super slow on CI, so bump it to 30s. I _thought_ we were already doing this, because `set_default_timeout` [says](https://playwright.dev/python/docs/api/class-browsercontext#browser-context-set-default-timeout) it affects "all the methods accepting [timeout](https://playwright.dev/python/docs/api/class-browsercontext#browser-context-set-default-timeout-option-timeout) option," but that apparently doesn't include `expect`.

also update a test reliant on async `moment` rendering to use expect rather than a one-time retrieval of text content, which I saw fail when I was testing this.

I think this might close #5361 and possibly a few other browser tests that seem to have become flaky, but shouldn't replace #5362, which is also an improvement.

### DIFF
--- a/jupyterhub/tests/browser/conftest.py
+++ b/jupyterhub/tests/browser/conftest.py
@@ -1,7 +1,7 @@
 from collections import namedtuple
 
 import pytest
-from playwright.async_api import async_playwright
+from playwright.async_api import async_playwright, expect
 
 from ..conftest import add_user, new_username
 
@@ -12,7 +12,10 @@ async def browser():
     async with async_playwright() as playwright:
         browser = await playwright.firefox.launch(headless=True)
         context = await browser.new_context()
+        # context sets default timeout for a lot of things, but not expect
         context.set_default_timeout(30_000)
+        # default timeout for expect
+        expect.set_options(timeout=30_000)
         page = await context.new_page()
         yield page
         await context.clear_cookies()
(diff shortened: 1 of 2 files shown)


willingc (approved):
Much nicer @minrk. I had no idea the timeout was so low.

willingc:
Between #5362 and #5363 the tests should be more deterministic (or less flaky).

willingc:
image

willingc:
I know it is set in the conf.py but perhaps we should set `expect` globally in the playwright config. It shouldn't matter but computers.

image

---

Nevermind, we're in Python.

willingc:
https://playwright.dev/python/docs/test-runners#async-fixtures

I wonder if we change the browser tests to use session scope instead of module which is in the `pytest.ini` if the tests would speed up????

minrk:
I think this is the same issue as the fix I also added here, where a value is retrieved immediately, but should be tested with `expect`. Need to figure out exactly how to expect it, though, since it's a list of values.

minrk:
module-level browser might indeed speed up the browser tests quite a bit. The context and page should still be function-level, though, so we'd need two fixtures. We can keep the current `browser` fixture (which actually means a `Page` in a `Context`), and add a `_browser` that is the actual `Browser` object that's module-scoped.

PR #5364, opened 2026-05-01
Re-use Browser instance across tests

willingc:
Draft test as a followup for playwright async fixture docs in #5363

### DIFF
--- a/jupyterhub/tests/browser/conftest.py
+++ b/jupyterhub/tests/browser/conftest.py
@@ -11,24 +11,6 @@ async def _browser():
     # browser_type in ["chromium", "firefox", "webkit"]
     async with async_playwright() as playwright:
         browser = await playwright.firefox.launch(headless=True)
-<<<<<<< HEAD
-        context = await browser.new_context()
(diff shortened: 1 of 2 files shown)


jupyterhub/tests/browser/conftest.py
```diff
@@ -6,7 +6,7 @@
 from ..conftest import add_user, new_username
 
 
-@pytest.fixture()
+@pytest.fixture(scope='session', loop_scope='session')
```
minrk: we don't need to set loop_scope, which is set automatically and correctly, but we could split `browser` into module-level Browser object, and keep function-scopes context and page for each, since I think they are quite inexpensive to create:

```python
@pytest.fixture(scope="module")
async def _browser():
    async with async_playwright() as playwright:
        browser = await...
        yield browser

@pytest.fixture
async def browser(_browser):
    context = await _browser.new_context()
    ....rest the same
```

minrk:
Great! This shaved 2 minutes off the browser tests. Looks like we still haven't quite _eliminated_ flaky tests

PR #5372, opened 2026-05-04
try to fix flakiness in test_start_stop_all_servers_on_admin_page

minrk:
trade one-time assert for async `expect` which waits for a condition to be true, handling asyncness on the page.

try to fix [this flaky assertion](https://github.com/jupyterhub/jupyterhub/actions/runs/25240629894/job/74015738308#step:11:170):

```python
>       assert (
            await btns_start.count()
            == await btns_spawn.count()
            == await users.count()
            == users_count_db
        )
E       assert 2 == 3

jupyterhub/tests/browser/test_browser.py:1231: AssertionError
```


willingc (approved):
@minrk This makes good sense. It will definitely help us in the future with flaky tests. I'm going to merge since all is passing and then look at cleaning up the recent dependabot prs.

PR #5378, opened 2026-05-05
require jupyter-events >= 0.11

minrk:
this is required for the change in version type from int to string.

Running with 0.10 leads to [validation errors expecting an integer](https://github.com/jupyterhub/roadmap/issues/12#issuecomment-4380745904).

cc @bl-aire

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ alembic>=1.4
 certipy>=0.1.2
 idna
 jinja2>=2.11.0
-jupyter_events
+jupyter_events>=0.11.0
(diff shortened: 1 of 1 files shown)


willingc (approved):
Thanks!

bl-aire:
> this is required for the change in version type from int to string.
> 
> Running with 0.10 leads to [validation errors expecting an integer](https://github.com/jupyterhub/roadmap/issues/12#issuecomment-4380745904).
> 
> cc @bl-aire

This is noted.
Thank you

PR #5386, opened 2026-05-08
ensure jupyterhub session id is set for cookie-authenticated requests

### LINKED ISSUE #5385: problems with xsrf tokens when session id is unset
@yuvipanda and 2i2c have found an issue where `POST /spawn` fails with an xsrf token mismatch.

When we set an XSRF token for a client, we don't want that token to be valid for other uses. When a user is logged in, it takes these two values into account:

- session id
- user's cookie id

BUT we also need xsrf token enforcement for the login form itself, at which point the user is not logged in yet. To limit portability, these "anonymous" xsrf tokens take into account the hash of the request IP. This can lead to problems when the request ip received by the Hub is not the actual request IP, e.g. behind load balancers that don't proxy headers consistently. This can lead to xsrf mismatches on login _and should be only login_, never a logged-in POST to `/spawn`.

The affected deployments are in such a situation where the reported client ip is not accurate, however it should not affect logged-in POST to `/spawn`.

The root cause is that the session id cookie is unset (somehow) for affected users, meaning an [anonymous value is used](https://github.com/jupyterhub/jupyterhub/blob/1da5f0f78700843c9d13f7f641cfa19dc07e5a39/jupyterhub/handlers/base.py#L286) in the xsrf cookie, leading to the GET and POST sometimes having different xsrf ids due to coming via different load-balancers or node ports.

The fix should be two-fold:

1. don't ever use the anonymous token path for session id for logged-in users, and
2. make sure session-id gets set for logged-in users if it has been unset for any reason

A possible _additional_ change suggestion is to set the _default_ anonymous IP CIDR to common LAN IPs:

```
JUPYTERHUB_XSRF_ANONYMOUS_IP_CIDRS="10.0.0.0/8;172.16.0.0/12;192.168.0.0/16"
```

since when those are reported, it is _very_ unlikely that the request ip provides useful information for these purposes (a single load balancer will report all users as the same ip, and two load balancers will report one user as coming from two ips). This could be done at the JupyterHub level, or the z2jh chart level, where this issue seems to come up most often.

minrk:
if not set, session id cookie will be set on any cookie-authenticated request

#5385 was primarily caused by the xsrf behavior when no session-id is set. This ensures that any cookie-authenticated request will re-set the session-id cookie, rather than setting it only on login, which has other benefits for token revocation, etc.

It also prevents the use of the anonymous IP-sensitive value in the event of the session id not being set (which shouldn't happen anyway, but is possible if you mess with cookies in between page request and). This means you _will_ get an xsrf mismatch if you clear or change the session id between the page request and the form submission. This was already the case before, though, if session id was set on one request _or_ the other.

closes #5385

### DIFF
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -270,20 +270,25 @@ class BaseHandler(RequestHandler):
         """
         # cases:
         # 1. logged in, session id (session_id:user_id)
-        # 2. logged in, no session id (anonymous_id:user_id)
+        # 2. logged in, no session id (shouldn't happen!) sets new session id, same as 1.
         # 3. not logged in, session id (session_id:anonymous_id)
         # 4. no cookies at all, use single anonymous value (:anonymous_id)
         session_id = self.get_session_cookie()
         if self.current_user:
             if isinstance(self.current_user, User):
                 user_id = self.current_user.cookie_id
+                if not session_id:
+                    # this shouldn't happen if cookie-authenticated
+                    self.log.warning(
+                        "session id not set for %s, setting a new one",
+                        self.current_user.name,
+                    )
+                    self.set_session_cookie()
+                    session_id = self._session_id
             else:
                 # this shouldn't happen, but may if e.g. a Service attempts to fetch a page,
                 # which usually won't work, but this method should not be what raises
-                user_id = ""
-            if not session_id:
-                # no session id, use non-portable anonymous id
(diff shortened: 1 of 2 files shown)


yuvipanda:
Yay! This is awesome, although I don't have the expertise to properly review it :(

willingc (approved):
Thanks @minrk. The logic tracks.

minrk:

PR #5389, opened 2026-05-11
Add bwJupyter to the deployment gallery

Paul2708:
Currently, bwJupyter has 12.800 registered users from over 40 universities.

I really appreciate all of the maintainers' work and the community's support.
Without them, the development of such a service would be way harder.


minrk:
Very cool!

PR #5397, opened 2026-06-01
Support usage of unix sockets in configurable proxy

### LINKED ISSUE #3529: Use Unix domain socket since Jupyter Server now supports them
### Proposed change
Use Unix domain sockets for *Jupyter Server* since support is now in [jupyter_server/CHANGELOG.md at 18156d508344964c73cf3a8882835bfefc25deda · jupyter-server/jupyter_server](https://github.com/jupyter-server/jupyter_server/blob/18156d508344964c73cf3a8882835bfefc25deda/CHANGELOG.md#190).

This addresses the comment [Listen on a UNIX-domain socket · Issue #883 · jupyterhub/jupyterhub](https://github.com/jupyterhub/jupyterhub/issues/883#issuecomment-262913138).

dometto:
Resolves #3529

ConfigurableHTTPProxy now supports listening on Unix Domain Sockets. But as of yet, JupyterHub isn't able to utilize this functionality. This PR:

* modifies the `Server` class so that a server can be initialized with a `unix+http://` URL.
* modifies the `ConfigurableHTTPProxy` class so that the right command line arguments are passed to `configurable-http-proxy` if a unix socket is requested.
  * users can configure the proxy to use unix sockets by setting `c.bind_url` and `c.ConfigurableHTTPProxy.api_url` to a `unix+http://` URL (e.g. `unix+http://%2Ftmp%2Fproxy.sock`)
* supports connecting to unix sockets for hub-internal methods (to check if a server is up, etc.)
* adds tests for the new functionality


jupyterhub/utils.py
```diff
@@ -21,18 +21,128 @@
...
 import idna
 from sqlalchemy.exc import SQLAlchemyError
 from tornado import gen, ioloop, web
-from tornado.httpclient import AsyncHTTPClient, HTTPError
+from tornado.httpclient import AsyncHTTPClient, HTTPError, HTTPRequest, HTTPResponse
 from tornado.log import app_log
+from tornado.netutil import Resolver
+
+
+def urlencode_unix_socket_path(socket_path: str) -> str:
```
dometto: This is identical to the function in https://github.com/jupyter-server/jupyter_server/blob/main/jupyter_server/utils.py

Also happy to import the method from there, but see the comment [below](https://github.com/jupyterhub/jupyterhub/pull/5397/changes#r3335327311).

jupyterhub/utils.py
```diff
@@ -21,18 +21,128 @@
...
+    """Encodes a UNIX socket path string from a socket path for the `http+unix` URI form."""
+    return socket_path.replace("/", "%2F")
+
+
+def urldecode_unix_socket_path(socket_path: str) -> str:
```
dometto: As above.

jupyterhub/utils.py
```diff
@@ -21,18 +21,128 @@
...
+    """Decodes a UNIX sock path string from an encoded sock path for the `http+unix` URI form."""
+    return socket_path.replace("%2F", "/")
+
+
+def unix_socket_in_use(socket_path: str) -> bool:
```
dometto, line 46: As above.

jupyterhub/utils.py
```diff
@@ -21,18 +21,128 @@
...
+    except OSError:
+        return False
+    else:
+        return True
+    finally:
+        sock.close()
+
+
+@contextmanager
+def _request_for_tornado_client(
```
dometto: This is modified from the function by the same name in https://github.com/jupyter-server/jupyter_server/blob/main/jupyter_server/utils.py

The modification is necessary because:

- the version in `jupyter_server` does not support arbitrary `kwargs` for constructing the `HTTPRequest` object
- we also want to support `pycurl`
- annoyingly, `jupyter_server` adopts the `http+unix://` convention for unix socket urls, while JupyterHub has settled on `unix+http`. I have here opted for a more liberal check that supports both.

dometto: Of course it would also be possible to first patch this in `jupyter_server` and then import the function here. However, I don't know if that kind of dependency on a utility function in another library is desired. For that reason, I have opted to copy all the unix socket related utility functions from `jupyter_server`.

Happy to depend on the external functions if that's preferred. Let me know!

jupyterhub/utils.py
```diff
@@ -264,39 +385,39 @@ async def exponential_backoff(
...
-    """Wait for any server to show up at ip:port."""
-    if ip in {'', '0.0.0.0'}:
-        ip = '127.0.0.1'
-    elif ip == "::":
-        ip = "::1"
-    display_ip = fmt_ip_url(ip)
-    app_log.debug("Waiting %ss for server at %s:%s", timeout, display_ip, port)
+async def wait_for_server(address, port, timeout=10):
+    """Wait for any server to show up at address:port."""
+    if port:
```
dometto: `port` is set to `0` when a unix socket is used.

jupyterhub/objects.py
```diff
@@ -57,15 +58,22 @@ def bind_url_default(self):
...
-                port = 443
-            else:
-                port = 80
-        self.port = port
+        if urlinfo.scheme:
+            self.proto = urlinfo.scheme
+
+        if self.proto == 'unix+http':
+            self.connect_url = f'unix+http://{urlinfo.netloc}'
+            self.port = 0
```
dometto: `port` is set to `0` when a unix socket is used, so when `port` evaluates to `False`, we know to connect to a unix socket (and not an ip+port).

This is slightly hackish since port 0 is often interpreted as 'assign random port'. But downstream functions that attempt to connect to `connect_addr` have no way of knowing whether it is a unix socket or not -- they only get passed `connect_addr` and `port`, and `connect_addr` could be an ip, hostname, or unix socket path. So I use `port = 0` as a way to signal that it is a unix socket.

jupyterhub/singleuser/mixins.py
```diff
@@ -394,39 +395,30 @@ def _validate_static_custom_path(self, proposal):
             path = list(_exclude_home(path))
         return path
 
-    # create dynamic default http client,
-    # configured with any relevant ssl config
-    hub_http_client = Any()
```
dometto, line 399: Slightly unsure about this change. I don't see hub_http_client being used anywhere, but perhaps these mixins are also used by libraries outside of this repo?

jupyterhub/utils.py
```diff
@@ -21,18 +22,176 @@
...
+    if AsyncHTTPClient.configured_class().__name__ == 'CurlAsyncHTTPClient':
+        using_pycurl = True
+    else:
+        using_pycurl = False
+
+    if ssl_options:
+        request_ssl_options = {k: v for k, v in ssl_options.items() if v is not None}
+        keyfile = request_ssl_options.pop("keyfile")
+        certfile = request_ssl_options.pop("certfile")
+        request_ssl_options = make_ssl_context(keyfile, certfile, **request_ssl_options)
```
dometto: since an `SSLContext` object cannot be pickled, and this function is async, we have to pass in the ssl options as a dict to this function and turn it into an `SSLContext` here.

jupyterhub/objects.py
```diff
@@ -169,9 +180,11 @@ def __repr__(self):
     def wait_up(self, timeout=10, http=False, ssl_context=None, extra_path=""):
         """Wait for this server to come up"""
         if http:
-            ssl_context = ssl_context or make_ssl_context(
```
dometto, line 172: See https://github.com/jupyterhub/jupyterhub/pull/5397/changes#r3336473130

dometto:
The test failures are all for `jupyterhub/tests/test_user.py::test_sync_groups`. As far as I can see, this is a pure db/backend test, and this PR does not modify any model related code. 

The CI tests with the `main_dependencies` flag use the `main` version of SQLalchemy, so I thought that maybe caused the errors. Similarly for the 3.10 test with `oldest_dependencies`.  But for some reason, `3.13, subdomain, serverextension` is also failing.

I cannot reproduce any of the errors on 3.13 on 3.14 (also not with `pip install --upgrade --pre sqlalchemy`).

Could this be a more general CI test problem? Any help appreciated!

manics:
@dometto We've got our regular collaboration cafe meeting later today [(2 June, 18:00 UTC)](https://arewemeetingyet.com/Los_Angeles/2026-06-02/11:00/JupyterHub-JupyterBook-CollabCafe)

We're thinking of a "Get your PR reviewed" session, see discussion on
[#jupyterhub > "Get your PR reviewed" sessions at JupyterHub collab cafe?](https://jupyter.zulipchat.com/#narrow/channel/469744-jupyterhub/topic/.22Get.20your.20PR.20reviewed.22.20sessions.20at.20JupyterHub.20collab.20cafe.3F/with/599295378)

It's short notice, but if you're interested let us know here or in Zulip!

jupyterhub/utils.py
```diff
@@ -21,18 +22,176 @@
...
 from tornado import gen, ioloop, web
-from tornado.httpclient import AsyncHTTPClient, HTTPError
+from tornado.httpclient import AsyncHTTPClient, HTTPError, HTTPRequest, HTTPResponse
 from tornado.log import app_log
+from tornado.netutil import Resolver
+
+
+def urlencode_unix_socket_path(socket_path: str) -> str:
+    """Encodes a UNIX socket path string from a socket path for the `http+unix` URI form."""
+    return socket_path.replace("/", "%2F")
```
minrk: Do you know why it is only `/` that requires encoding, and not use stdlib `urlencode(path)`? What if socket_path has spaces and other escape-requiring characters?

jupyterhub/singleuser/extension.py
```diff
@@ -342,39 +342,38 @@ def _default_hub_auth(self):
...
         - check version and warn on sufficient mismatch
         """
-        client = self.hub_http_client
         RETRIES = 5
         for i in range(1, RETRIES + 1):
             try:
-                resp = await client.fetch(self.hub_auth.api_url)
+                resp = await async_fetch(
+                    self.hub_auth.api_url,
+                    io_loop=self.serverapp.io_loop,
```
minrk: let's remove the io_loop arg here and everywhere - it's only ever needed when called from outside a coroutine/asyncio, which we don't have to deal with.

jupyterhub/objects.py
minrk: We have `proto` already, so if we parse `http+unix://...`, then `self.proto` will be `http+unix`. Why not use that?

jupyterhub/singleuser/mixins.py
minrk: the use is [here](https://github.com/jupyterhub/jupyterhub/pull/5397/changes#diff-8337e3cdc557272aa014bfcbab481b0ad69c082aa0d714f27a188024c1f2cc57L425), but you've removed it, so this is perhaps okay to drop

jupyterhub/utils.py
minrk: why is pickling relevant to async? There's no process boundary, so there should be no serialization involved.

jupyterhub/services/auth.py
```diff
@@ -591,10 +592,9 @@ async def _check_hub_authorization(
             self.cache[cache_key] = data
         return data
 
-    async def _api_request(self, method, url, **kwargs):
+    async def _api_request(self, method, url, headers={}, **kwargs):
```
minrk: it's important to not have mutable containers in default arguments, so leaving this in kwargs is probably better. That, or `headers=None` and create the dict below.

jupyterhub/tests/test_unix_socket.py
```diff
@@ -0,0 +1,65 @@
...
+from traitlets.config import Config
+
+from ..utils import async_fetch, urlencode_unix_socket_path
+from .mocking import MockHub
+from .utils import auth_header
+
+
+async def test_unix_socket_proxy(request, tmp_path):
+    cfg = Config()
+    proxy_sock = '/tmp/proxy.sock'  # str(tmp_path / 'proxy.sock')
```
minrk: is there a reason the commented-out code didn't work? That ought to be preferable.

jupyterhub/utils.py
minrk: I've always been frustrated by ambiguity here and the lack of standardization - is there any standard we can defer to?

I see that [requests-unixsocket](https://pypi.org/project/requests-unixsocket/) also does `http+unix`, as does [this proposal in whatwg](https://github.com/whatwg/url/issues/577) (which went mostly nowhere). Maybe CHP should match the apparently more common choice, rather than the other way around? CHP is both the most recent to add support _and_ has made the least common choice _and_ the one under our control, so easiest to change, all of which point me to switching CHP to `http+unix` instead of switching jupyter-server. What do you think, there?

jupyterhub/tests/test_unix_socket.py
dometto: Thanks for flagging this; I need the `/tmp/proxy.sock` on OS X, absurdly, because it looks like `tmp_path` on my system is generating filepaths that are *too long* for tornado to accept as a connectable unix socket path...

On CI and in a linux container this issue doesn't come up, so the commented code should be restored before merging this.

jupyterhub/objects.py
dometto: As discussed yesterday, this won't work because `wait_for_server` and `can_connect` currently don't have access to `proto`. Will address this by making these functions take an explicit `unix_socket` parameter.

jupyterhub/utils.py
dometto: No idea why this was done this way in `jupyter_server`. The PR for that is here: https://github.com/jupyter-server/jupyter_server/pull/525/changes

I'll change things to use the `urllib` functions, as discussed yesterday.

jupyterhub/utils.py
dometto: I found the logs for the pickling error for `SSLContext` here: https://github.com/jupyterhub/jupyterhub/actions/runs/26767848225/job/78899005507#step:11:405

jupyterhub/proxy.py
```diff
@@ -669,15 +675,20 @@ def _write_pid_file(self):
...
             return
-        self.log.debug("Removing proxy pid file %s", self.pid_file)
-        try:
-            os.remove(self.pid_file)
-        except FileNotFoundError:
-            self.log.debug("PID file %s already removed", self.pid_file)
+        paths = [self.pid_file]
+        for socket_url in [self.public_url, self.api_url]:
+            if socket_url.startswith('unix+http://'):
+                paths.append(unquote_plus(socket_url[12:]))
```
dometto: I'm using `unquote_plus` here instead of `unquote` because I *think* that's what we want for spaces?

jupyterhub/utils.py
dometto: I have a change ready that replaces unix+http with http+unix all across JupyterHub, but it would require the corresponding change in CHP (and presumably a breaking change release of that). Should we do that first?

jupyterhub/singleuser/extension.py
dometto: Done -- for some reason, this is failing when I run the tests locally, but it works in CI. Presumably something about my local setup, but perhaps good to give it a spin locally yourself?

dometto:
Addressed the issues we discussed yesterday @minrk! See the comments.

Remaining test failures (after rebasing on `main`): I don't understand these yet. I can't reproduce locally. And if I'm understanding correctly, the failures to stop the proxy are occurring when attempting to find/kill the previous process -- not when trying to remove the pid files (which is the function I touched): https://github.com/jupyterhub/jupyterhub/actions/runs/27678600621/job/81859970274?pr=5397#step:11:1243

jupyterhub/utils.py
dometto: See https://github.com/jupyterhub/configurable-http-proxy/pull/655/changes

jupyterhub/utils.py
dometto: Oh, I think I understand the issue now. It's indeed not related to async, but to the `deepcopy` we do [here](https://github.com/jupyterhub/jupyterhub/pull/5397/changes#diff-6cafcd890c3e465498df02af8660a1ebf5513c9804c761d5a369ad908ad67149R168).

Apparently `deepcopy` [uses pickling](https://stackoverflow.com/questions/53601999/getting-typeerror-cant-pickle-thread-lock-objects-when-an-object-is-deepcopi).

Coming from a Ruby background, my instinct was to `deepcopy` kwargs so as not to mutate it in place. But if I see correctly this is not an issue with `kwargs` in Python. So I'll remove the deepcopy.

jupyterhub/utils.py
```diff
@@ -82,7 +82,7 @@ def _request_for_tornado_client(
 
     if parts.scheme in ["http", "https"]:
         pass
-    elif "unix" in parts.scheme:
+    elif parts.scheme == "unix+http":
```
dometto: Will change this to http+unix when the CHP change is released.

minrk: CHP 5.3.0 is out, so this should be able to finish now.

dometto:
Still seeing a test failure in 4 scenarios, apparently caused by a failure of the proxy to shutdown in a previous test (?). Here is the relevant log capture from CI:

```
------------------------------ Captured log setup ------------------------------
07:22:45 INFO app.py:3447 Running JupyterHub version 6.0.0.dev
07:22:45 DEBUG app.py:3422 Writing PID 2838 to /tmp/tmpempuj6e5
07:22:45 INFO app.py:3477 Using Authenticator: jupyterhub.tests.mocking.MockPAMAuthenticator-6.0.0.dev
07:22:45 INFO app.py:3477 Using Spawner: jupyterhub.tests.mocking.MockSpawner-6.0.0.dev
07:22:45 INFO app.py:3477 Using Proxy: jupyterhub.proxy.ConfigurableHTTPProxy-6.0.0.dev
07:22:45 INFO app.py:1844 Loading cookie_secret from /tmp/pytest-of-runner/pytest-0/test_named_server_migration_au0/jupyterhub_cookie_secret
07:22:45 DEBUG app.py:2018 Connecting to db: sqlite:////tmp/pytest-of-runner/pytest-0/test_hub_upgrade_detection0/jupyterhub.sqlite
07:22:45 DEBUG orm.py:1517 database schema version found: f460b0348206
07:22:45 INFO proxy.py:559 Generating new CONFIGPROXY_AUTH_TOKEN
[...]
07:22:45 WARNING proxy.py:629 Found proxy pid file: /tmp/pytest-of-runner/pytest-0/test_named_server_migration_au0/jupyterhub-proxy.pid
07:22:45 WARNING proxy.py:646 Proxy still running at pid=4796
07:22:48 CRITICAL app.py:3821 Failed to start proxy
  File "/home/runner/work/jupyterhub/jupyterhub/jupyterhub/app.py", line 3819, in start
    await self.proxy.start()
  File "/home/runner/work/jupyterhub/jupyterhub/jupyterhub/proxy.py", line 725, in start
    self._check_previous_process()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/runner/work/jupyterhub/jupyterhub/jupyterhub/proxy.py", line 670, in _check_previous_process
    raise RuntimeError("Failed to stop proxy at pid=%s", pid)
RuntimeError: ('Failed to stop proxy at pid=%s', 4796)

07:22:48 DEBUG application.py:1063 Exiting application: jupyterhub
_______________ ERROR at setup of test_sync_groups[group_names1] _______________

```

The error is being thrown by `test_sync_groups`. However, I'm a bit confused, as it looks like the proxy is using tmp files from other tests:

* `Connecting to db: sqlite:////tmp/pytest-of-runner/pytest-0/test_hub_upgrade_detection0/jupyterhub.sqlite`
* `Found proxy pid file: /tmp/pytest-of-runner/pytest-0/test_named_server_migration_au0/jupyterhub-proxy.pid`

I would assume each test run should isolate these files.

Also, I've touched the code that removes `pid` files, but if I see correctly this test isn't even getting the `_remove_process_files` function. Instead the RuntimeError is from `proxy.py:670`, thrown after `self._check_pid(pid)` returns `True` (i.e. stopping of the already running proxy is determined to have failed).

@minrk happy to dive into this more but perhaps you can shine a light on whether you think this is related to code I've changed (likely, but how?) or something more generic in the test suite?

jupyterhub/app.py
```diff
@@ -1104,6 +1104,14 @@ def _deprecated_proxy_api(self, change):
         """,
     ).tag(config=True)
 
+    hub_socket_mode = Integer(
```
dometto, line 1107: Added some extra configuration to allow setting the permissions for the created socket (this is the socket on which the Hub itself binds).

jupyterhub/proxy.py
```diff
@@ -580,6 +580,12 @@ def _api_url_default(self):
         help="""The command to start the proxy""",
     )
 
+    command_umask = Integer(
```
dometto, line 583: We can't directly influence the permissions with which CHP creates its unix sockets, as `npm:net` seems not to allow this. We can use `chmod` after creation, but I thought making `umask` configurable was neater.

dometto:
**UPDATE**: this appears to be fixed in [b226022](https://github.com/jupyterhub/jupyterhub/pull/5397/commits/b22602296f0db9d0ff99b2cdcd027a83a528b6b0) but honestly I'm not sure why.

## Description of error

I'm also still running into a bit of an ugly error. When attempting to bring it all together using [this config](https://gist.github.com/dometto/674b9ece2fab53cfde42682b8faf2f7d), connecting to the Hub and spawning a singleuser server throws this in the Hub's logs:

```
[I 2026-07-09 17:28:49.110 ServerApp] Jupyter Server 2.19.0 is running at:
[I 2026-07-09 17:28:49.110 ServerApp] http://127.0.0.1:58647/user/alice/lab?token=...
[I 2026-07-09 17:28:49.110 ServerApp] Use Control-C to stop this server and shut down all kernels (twice to skip confirmation).
[D 2026-07-09 17:28:49.114 JupyterHub base:370] Recording first activity for <APIToken('763a...', user='3060845', client_id='jupyterhub')>
[D 2026-07-09 17:28:49.116 JupyterHub scopes:1010] Checking access to /hub/api/users/alice/activity via scope users:activity!user=3060845
[D 2026-07-09 17:28:49.117 JupyterHub users:1038] Activity for user alice: 2026-07-09T15:28:49.062418Z
[D 2026-07-09 17:28:49.117 JupyterHub users:1056] Activity on server alice/: 2026-07-09T15:28:49.062418Z
[I 2026-07-09 17:28:49.119 JupyterHub log:192] 200 POST /hub/api/users/alice/activity (alice@0.0.0.0) 5.28ms
[W 2026-07-09 17:28:49.855 JupyterHub utils:467] Failed to connect to http://127.0.0.1:58647/user/alice/api ([Errno 2] No such file or directory)
```

So: the singleuser server is spawning correctly, but the Hub can't connect back to it, throwing `Errno 2] No such file or directory`.

**This seems to indicate that the url http://127.0.0.1:58647/user/alice/api is being interpreted as a unix socket, and is being handled by the `UnixSocketResolver` in `async_fetch`.**

The URL is correctly formed, and debugging shows that the request to `http://127.0.0.1:58647` is not causing the `UnixSocketResolver` to be configured. But that seems to mean that `AsyncHTTPClient` is *still* using the `UnixSocketResolver`, even though it shouldn't be in this context, because it was configured by a previous request.

Is this some issue caused by some unsafe `async` code -- i.e. is another thread's configuring `AsyncHTTPClient` to use `UnixSocketResolver` bleeding into this request?

There is also a reproducible example illustrating this error in a Dockerfile here:
https://github.com/dometto/jupyterhub-unix-socket-example

jupyterhub/proxy.py
```diff
@@ -306,7 +306,7 @@ async def get_route(self, routespec):
 
     # Most basic implementers must only implement above methods
 
-    async def add_service(self, service, client=None):
```
dometto, line 304: It looks like there were some spurious optional `client` keywords floating around. Removed them in this commit.

jupyterhub/utils.py
```diff
@@ -102,14 +102,13 @@ class UnixSocketResolver(Resolver):
...
 
-            def close(self):
-                self.resolver.close()
-
-            async def resolve(self, host, port, *args, **kwargs):
-                return [(socket.AF_UNIX, unquote_plus(host))]
+            async def resolve(
+                self, host: str, port: int, *args, **kwargs
+            ) -> List[Tuple[int, Any]]:
+                return [(socket.AF_UNIX, self.socket_path)]
```
dometto: Using `self.socket_path` here fixes https://github.com/jupyterhub/jupyterhub/pull/5397#issuecomment-4926906315

I don't quite understand why. A model suggested this pattern* when looking for examples of tornado w/ unix socket. I noticed that it was different from `jupyter-server`'s implementation, and tried it out. But it would be nice to understand why it works, if anyone has any idea @minrk!

***Disclaimer: I didn't use AI in working on this PR, but when searching it's difficult to avoid AI-generated examples.**

Looking at the current Tornado implementation I'm not sure why the previous `initialize(self, resolver)` would be necessary, so I could understand that that pattern (with `resolver=Resolver()`) causes some kind of async issues under the hood.

However, the thing I really don't get is that the bug returns as soon as I try `return [(socket.AF_UNIX, unquote_plus(host))]` insteaad of `return [(socket.AF_UNIX, self.socket_path)]`!

The other question is why the implementation in `jupyter-server` seems to work!

dometto:
https://github.com/jupyterhub/jupyterhub/pull/5397#issuecomment-4922865803 is the main stopper for this now -- advise welcome.

There is also a failure in `test-docs` which I think is unrelated, and currently a CI failure on one of the browser tests that looks like a fluke.

A proof of concept of the changes in this branch, running JupyterHub+CHP+Nginx using only unix domain sockets, is [here](https://github.com/dometto/jupyterhub-unix-socket-example). I'll also add some more context to that repo explaining the use case (multiple users on a shared VM, authentication being passed in via an HTTP header from an upstream auth proxy).

jupyterhub/utils.py
dometto: I've added type hints to this class for good measure.

minrk:
Great, thank you! I'm looking into the test failure now

minrk:
Great! This almost works. However, there is a problem in the 'unix socket resolver' implementation, where it is resolving all hosts to a single socket, but that's being re-used to make requests that _should_ go to different hosts.

We've also had problems with calling `AsyncHTTPClient.configure` because that is a stateful thing that replaces _all_ prior configuration with only what's given and affects all other requests.

I'm not sure why it's only the oldest_dependency case that is failing, but perhaps we have pycurl in the other cases.

I think we need the resolver to work with other requests and resolve the socket from the `host` field, and call `super()` when it's not a socket to avoid breaking non-http+unix requests.

It might be simpler to consider trading the tornado client we already have for a dependency on [niquests, which supports `http+unix` [without any configuration](https://niquests.readthedocs.io/en/latest/user/quickstart.html#unix-socket).

We're preparing a 6.0 release, and adding `niquests` as a dependency shouldn't be a big user-affecting change, though it may have unforeseen consequences under load.

minrk:
If you'd like me to give niquests a try, feel free, or if you'd rather I tackle that part, I'm happy to.

dometto:
Thanks for the feedback again! Makes sense that this is some kind of async madness in tornado's resolver logic, glad the tests caught the issue.

`niquests` looks like a really neat library, with a lot of deserved attention recently. If I see correctly the project is still run by a single maintainer -- if that's no showstopper for depending on it, I'd be happy to plug it in tomorrow.

minrk:
I think it's not, if you want to give it a try. The consequences of switching HTTP request libraries may mean touching quite a few things as the API to check status codes and get JSON responses may be touched in a lot of places, in case it looks like more than you want to take on. We can try fixing the tornado way, if you prefer.

dometto:
Since it looks like the tornado http client is also used in a bunch of other files I'm not yet familiar with, replacing tornado might be more work than makes sense within this PR (as long as there is an alternative). So I decided to give tornado's resolver another go.

I've refactored to force-creating client instances (`AsyncHTTPClient(force_instance=True)`), instead of using `AsyncHTTPClient.configure`, which just doesn't seem thread-safe (it configures [a global singleton used when using AsyncHTTPClient](https://www.tornadoweb.org/en/stable/_modules/tornado/httpclient.html#AsyncHTTPClient)). This also removes the need for a contexmanager.

Tests are all passing now except the documentation check (I think unrelated). Local tests are also successful on my end.

However, the Tornado docs state:

```
Only a single AsyncHTTPClient instance exists per IOLoop in order to provide limitations on the number of pending connections. force_instance=True may be used to suppress this behavior.
```

So it seems using `force_instance=True` may have effects on the number of concurrent requests applications will try to make (instead of queuing them). I don't know how undesirable this is.

An alternative may be to use a Semaphore inside `async_fetch` to limit the number of calls per IOLoop. But I see the only call to `async_fetch` in `proxy.py` is already using its own semaphore. So the other option is to make callers of `async_fetch` responsible for limiting request concurrency. The calls to `async_fetch` inside `JupyterHubSingleUser` and in `mixins.py` also seem like they are already limited (with counters, not with a semaphore). That leaves the call inside the `HubAuth#_api_request` .

jupyterhub/utils.py
```diff
@@ -26,13 +26,108 @@
...
+
+class UnixSocketResolver(Resolver):
+    """A resolver that routes HTTP requests to unix sockets
+    in tornado HTTP clients.
+    Due to constraints in Tornados' API, the scheme of the
+    must be `http` (not `http+unix`). Applications should replace
+    the scheme in URLS before making a request to the HTTP client.
+    """
+
+    async def resolve(self, host, port, *args, **kwargs):
```
dometto, line 70: The [implementation of this class](https://github.com/jupyter-server/jupyter_server/blob/5b6f030ecf91341261b1879b55297a87f9804183/jupyter_server/utils.py#L289) in `jupyter-server` also implements `initialize` and `close`. But I'm not sure why this would be needed.

minrk:
Thanks so much! I think this is good to go now, and I may look into niquests as a standalone change, rather than conflating it with this PR.

dometto:
Thank you for all the feedback, learned a lot by working on this! Very excited for JupyterHub 6.0. 🎉

Issue #5357, opened 2026-04-29
Handle erroring and metrics when a server launch was intentionally denied by `c.Spawner.pre_spawn_hook`


jnywong:
### Proposed change

I have a use case where I want to intentionally deny a server launch with config such as:

```python
async def quota_pre_spawn_hook(spawner):
    launch_flag = await some_function(spawner)
    if launch_flag == False:
        raise CustomException(status_code=..., log_message=..., etc.)

c.Spawner.pre_spawn_hook = quota_pre_spawn_hook
```

where `CustomException` subclasses `tornado.web.HTTPError`. This works nicely and renders a custom template based on the `CustomException.status_code`, but there are a couple of snags:

1. JupyterHub treats this as a spawn failure, so logs fill up with warnings from the `user` and `pages` modules
2. The metric `jupyterhub_server_spawn_duration_seconds_count{status="failure"}` increments, and so alerting systems based on this metric are triggering false positives.

My proposed change is to more rigorously define a `SpawnException` and update the internal logic to handle this exception so that snags 1. and 2., and possibly html template rendering issues are remediated.

### Alternative options

My alternative option is to 

- ignore snag 1. and suck it up with noisy logs
- to implement a custom metric to count intentional server denials to subtract from `jupyterhub_server_spawn_duration_seconds_count{status="failure"}` so that false positive alerts are not triggered in snag 2.

### Who would use this feature?

I think folks would use this feature because there's already [cases in the wild](https://discourse.jupyter.org/t/custom-jupyterhub-error-messages/47/21) where a spawn failure is triggered intentionally.

### (Optional): Suggest a solution

- define a standard `SpawnException` class
- optionally lower log level from `error` to `info` from `user` and `pages` modules when handling `SpawnException`
- add a new status label to `jupyterhub_server_spawn_duration_seconds_count` to differentiate from actual failures and intentional failures
- make sure the html page rendered upon intentional spawn failure passes customisable info from the `SpawnException` class and is not wrapped in any other http error

minrk:
Makes sense to distinguish! What do you think about a `reason` field for failed spawns? It seems like if the user is trying to spawn and not succeeding, that's still meaningfully a similar event to an error from the user perspective, right. But `status=failed, reason=error` and `status=failed, reason=input_invalid` and `status=failed, reason=capacity` are usefully distinct.

Depending on the answers there, I think it makes sense to either:

1. declare a special Exception like you said, and have this define a _new_ status that's not failed, or
[...]

PR #5405, opened 2026-06-03
avoid logging traceback on HTTPError raised in spawn

minrk:
HTTPErrors should be shown to users and logged more quietly. They are already relayed to users at the higher level, but the traceback part got logged here.

this is the simple bug part of #5357, only preventing tracebacks in the logs for HTTPError.

Being a simple bugfix means we can backport it for the next 5.x release.

The SpawnException can be done separately, since it won't be backported.


manics (approved):

minrk:

jnywong:
Been away for the past 2 weeks but super happy to see this drop, thanks @minrk and @manics

PR #5409, opened 2026-06-05
Don't rely on package.json postinstall scripts

manics:
I've set `ignore-scripts=true` in my npmrc file to improve security, but this breaks `pip install .` since `bower-lite` is never run.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -106,7 +106,14 @@ class NPM(BaseCommand):
         )
         os.utime(self.node_modules)
 
+        # autorunning postinstall scripts may be disabled for security
+        check_call(
+            ['npm', 'run', 'postinstall'],
+            cwd=here,
+            shell=shell,
+        )
         os.utime(self.bower_dir)
+
         # update data-files in case this created new files
         self.distribution.data_files = get_data_files()
         assert not self.should_run(), 'NPM.run failed'


setup.py
```diff
@@ -106,7 +106,14 @@ def run(self):
         )
         os.utime(self.node_modules)
 
+        # autorunning postinstall scripts may be disabled for security
+        check_call(
+            ['npm', 'run', 'postinstall'],
```
manics, line 111: If we want to drop the `postinstall` requirement we could rename the script to `build`, or run `["sys.executable", "bower-lite"]` here.

consideRatio (approved):
Ahhh nice!

consideRatio:
Regarding security with npm btw - i think it could make sense to try be systematic in all our repos about suggesting use of `npm ci` over `npm install` to lean on the package-lock.json file over package.json, to not get the absolute latest malware by default.

manics:
`npm ci` would be nice, but causes problems when using different versions of yarn/npm/jlpm etc.

E.g. in https://github.com/jupyterhub/jupyter-server-proxy/pull/582

<details>

```

Post-resolution validation
  ➤ YN0000: │ @@ -5283,13 +5283,12 @@
  ➤ YN0000: │    linkType: hard
  ➤ YN0000: │  
  ➤ YN0000: │  "typescript@patch:typescript@~5.2.2#~builtin<compat/typescript>":
  ➤ YN0000: │    version: 5.2.2
  ➤ YN0028: │ -  resolution: "typescript@patch:typescript@npm%3A5.2.2#~builtin<compat/typescript>::version=5.2.2&hash=f3b441"
  ➤ YN0028: │ +  resolution: "typescript@patch:typescript@npm%3A5.2.2#~builtin<compat/typescript>::version=5.2.2&hash=85af82"
  ➤ YN0000: │    bin:
  ➤ YN0000: │      tsc: bin/tsc
  ➤ YN0000: │      tsserver: bin/tsserver
  ➤ YN0028: │ -  checksum: 0f4da2f15e6f1245e49db15801dbee52f2bbfb267e1c39225afdab5afee1a72839cd86000e65ee9d7e4dfaff12239d28beaf5ee431357fcced15fb08583d72ca
  ➤ YN0000: │    languageName: node
  ➤ YN0000: │    linkType: hard
  ➤ YN0000: │  
  ➤ YN0000: │  "typestyle@npm:^2.0.4":
  ➤ YN0000: │ 
  ➤ YN0028: │ The lockfile would have been modified by this install, which is explicitly forbidden.
```

</details>

minrk:
Nice! I'm fine with direct invocation of bower-lite as well, if folks prefer that. I think this postinstall was to get the same result of `bower install` after `npm install`, but that's not actually super important.

PR #5418, opened 2026-06-10
changelog for 5.5

minrk:
preparing to cut a release today

minor instead of patch due to backporting #5298


consideRatio (approved):
:tada: nice @minrk!!

PR #5429, opened 2026-06-30
[5.x] Fix upgrade-db crash in api_token_scopes data migration (#5427)

### LINKED ISSUE #5427: upgrade-db from 2.3.x to 5.x fails: api_token_scopes migration loads not-yet-existing columns/tables via access_scopes → spawner.user
## Summary

Upgrading a **populated** hub database across several major versions (e.g. 2.3.1 → 5.4.5) fails in the `651f5419b74d` (`api_token_scopes`, JupyterHub 3.0) migration with errors such as:

```
psycopg2.errors.UndefinedColumn: column roles.managed_by_auth does not exist
```

(or, depending on the SQLAlchemy version and data, `relation "shares" does not exist`).

The migration is a data-migration that instantiates the **current** ORM and traverses relationships, so it ends up issuing `SELECT`s for columns/tables introduced in **later** JupyterHub versions that do not exist yet at this point in the alembic chain.

This has been partially patched several times but the **root path remains unguarded**, so it is still reproducible on `main` / 5.5.0 for a 2.3.x → 5.x upgrade.

## Environment

- Upgrading from JupyterHub **2.3.1** (hub alembic head `833da8570507`) to **5.4.5** (and reproducible against `main`).
- PostgreSQL backend, populated database (rows in `oauth_clients` + `spawners` with an associated user that has roles).
- `jupyterhub upgrade-db` (equivalently `JupyterHub.upgrade_db = True` on startup).

## Steps to reproduce

1. Create/populate a hub DB on JupyterHub 2.3.1 (at least one user with a role, one spawner, and its per-server oauth client — i.e. a user who has started a server).
2. Install JupyterHub 5.4.5 (or `main`).
3. Run `jupyterhub upgrade-db`.

Result: the upgrade aborts during `833da8570507 -> 651f5419b74d`.

**Trigger condition / version scope (important):** the failure is *data-dependent*, not version-dependent. It only triggers when there is an `oauth_client` with an associated `spawner` (a user who has started a server), because that is the only path that reaches `access_scopes → spawner.user`. A 2.3.x DB **without** any started server migrates fine. The bug is **not** a regression: JupyterHub 5.0.x through 5.5.x are all affected identically (verified: 5.3.0 and 5.4.5 have byte-identical ORM and alembic chains and both fail with server data, both succeed without it). So "it worked for my upgrade" usually just means that database had no started servers.

## Traceback (abridged)

```
[I] Running upgrade 833da8570507 -> 651f5419b74d, api_token_scopes
...
File ".../alembic/versions/651f5419b74d_api_token_scopes.py", line 133, in upgrade
    allowed_scopes.update(access_scopes(oauth_client, db))
File ".../alembic/versions/651f5419b74d_api_token_scopes.py", line 34, in access_scopes
    scopes.add(f"access:servers!server={spawner.user.name}/{spawner.name}")
...
sqlalchemy.exc.ProgrammingError: (psycopg2.errors.UndefinedColumn)
    column roles.managed_by_auth does not exist
[SQL: SELECT ... roles.managed_by_auth AS roles_managed_by_auth
      FROM users AS users_1 JOIN user_role_map ... JOIN roles ON ...]
```

## Root cause

In `upgrade()`, the `oauth_clients` loop calls `access_scopes(oauth_client, db)`, and `access_scopes` accesses `spawner.user.name`:

```python
spawner = oauth_client.spawner
if spawner:
    scopes.add(f"access:servers!server={spawner.user.name}/{spawner.name}")
```

`spawner.user` lazy-loads a `User`, and loading a `User` triggers its `selectin` relationships from the **current** ORM:

- `User.roles` (selectin) → `SELECT ... roles.managed_by_auth` — column added in 5.0 (`manage_roles`).
- `User.shared_with_me` (selectin) → `SELECT ... FROM shares` — table added in 5.x.

Neither exists yet at this migration step, so the query fails.

The loader options on the `oauth_clients` query only defer columns on the **directly** loaded entities:

```python
for oauth_client in db.query(orm.OAuthClient).options(
    selectinload(orm.OAuthClient.allowed_roles).defer(orm.Role.managed_by_auth),
    # (main also adds) selectinload(orm.OAuthClient.spawner).defer(orm.Spawner.display_name),
):
    allowed_scopes.update(access_scopes(oauth_client, db))
```

They do **not** cover the cascaded `spawner.user → User.roles / User.shared_with_me` loads that happen inside `access_scopes`. The `api_tokens` loop right above uses `raiseload("*")`, which is why that loop is safe; the `oauth_clients` loop omits it.

### Why the previous fixes are incomplete

The same crash has been patched incrementally, each time covering one more future column but never the cascade through `User`:

- `d32b574` (2023-06): "avoid eager loading of not-yet-upgraded column".
- `4d8c3cbf` (2024-04): `.defer(orm.Role.managed_by_auth)`.
- `25b28970` (2025-12, PR #5192, released in 5.5.0): `.defer(orm.Spawner.display_name)` ("Fix db upgrade error when running older upgrade script").

Because `access_scopes` loads `spawner.user` (a `User`), and `User` has `selectin` relationships to `roles` (→ `managed_by_auth`) and `shared_with_me` (→ `shares`), the upgrade still fails on `main`.

## Proposed fix

Add `raiseload("*")` to the `oauth_clients` query options (mirroring the `api_tokens` loop), so the migration never triggers lazy/selectin loads of unrelated, possibly-future relationships:

```python
for oauth_client in db.query(orm.OAuthClient).options(
    selectinload(orm.OAuthClient.allowed_roles).defer(orm.Role.managed_by_auth),
    selectinload(orm.OAuthClient.spawner).defer(orm.Spawner.display_name),
    raiseload("*"),
):
    ...
```

However, `access_scopes` then accesses `spawner.user.name`, which would itself be blocked by `raiseload("*")`. So `spawner.user` needs to be eagerly loaded too — and crucially **without** pulling `User`'s future selectin relationships. Options:

- eager-load the needed path explicitly, e.g. `selectinload(orm.OAuthClient.spawner).selectinload(orm.Spawner.user).load_only(orm.User.name)`, combined with `raiseload("*")`, or
- replace the ORM access in `access_scopes` with raw SQL for the `spawner → user.name` lookup (as the function already does for the `services` branch), avoiding the ORM entirely.

More generally: data-migrations that instantiate the live ORM are fragile across schema evolution. Using core/`text()` queries (or `Bundle`/`load_only` + `raiseload("*")`) for the columns actually needed would make these migrations robust to future model changes.

## Workaround (for reference)

For deployments that cannot install an older JupyterHub to do a staged upgrade, the hub DB can be brought to the current schema in a single process by (1) pre-creating the future tables/columns the ORM references so the legacy migration's loads succeed, (2) making `alembic.op.add_column` / `create_table` idempotent for the duration, and (3) running `alembic upgrade head` in-process. The real migrations (data transforms included) then run unmodified.

mleida-stratio:
## Problem

Closes #5427 (for the `5.x` release branch).

Upgrading a **populated** database (e.g. 2.3.x → 5.x) crashes during alembic revision `651f5419b74d` (`api_token_scopes`) when an OAuth client has an associated spawner (i.e. a user who has started a server):

- `psycopg2.errors.UndefinedColumn: column roles.managed_by_auth does not exist`
- variant: `relation "shares" does not exist`

## Root cause

The `oauth_clients` data-migration loop calls `access_scopes(oauth_client, db)`, which accesses `spawner.user.name`. That lazy-loads the `User` ORM object, and `User` has `lazy="selectin"` relationships (`User.roles`, `User.shared_with_me`) that auto-fire on load. They select `roles.managed_by_auth` (added in 5.0) and the `shares` table (added in 5.x) — neither exists yet at this migration step. On `5.x` the loop also lacks the `Spawner.display_name` defer, so the spawner load can fail on that future column too.

## Fix

- **`access_scopes()`**: resolve the user name with raw SQL on `spawner.user_id` instead of the live ORM relationship, mirroring the existing services branch in the same function.
- **`oauth_clients` query loop**: defer `Spawner.display_name` and add `raiseload("*")` so any remaining/future relationship load fails loudly instead of silently touching not-yet-existing schema — mirroring the safe `api_tokens` loop just above.
- **`tests/populate_db.py`**: tie a per-server OAuth client to the `has-server` spawner so `test_upgrade` actually exercises this path (regression coverage). Guarded on `version_info >= (2,)` since `Spawner.oauth_client` was added in 2.0.

After this change the migration file matches the corresponding `main` PR.

## Verification

- `pytest jupyterhub/tests/test_db.py -k upgrade` — fails on the augmented `populate_db.py` before the fix, passes after, for all `hub_version` params.
- Reproduces the reported Postgres `UndefinedColumn` via `JUPYTERHUB_TEST_DB_URL`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/jupyterhub/alembic/versions/651f5419b74d_api_token_scopes.py
+++ b/jupyterhub/alembic/versions/651f5419b74d_api_token_scopes.py
@@ -31,7 +31,15 @@ def access_scopes(oauth_client: orm.OAuthClient, db: Session):
         return frozenset()
     spawner = oauth_client.spawner
     if spawner:
-        scopes.add(f"access:servers!server={spawner.user.name}/{spawner.name}")
+        # Avoid spawner.user.name: it lazy-loads the User ORM object, whose
+        # selectin relationships (User.roles, User.shared_with_me) reference
+        # columns/tables added in later schema versions (roles.managed_by_auth,
+        # shares) that do not exist yet at this migration step. Use raw SQL.
+        user_name = db.execute(
+            text("SELECT name FROM users WHERE id = :user_id"),
+            {"user_id": spawner.user_id},
+        ).scalar()
+        scopes.add(f"access:servers!server={user_name}/{spawner.name}")
     else:
         statement = "SELECT * FROM services WHERE oauth_client_id = :identifier"
         service = db.execute(
@@ -127,7 +135,17 @@ def upgrade():
             for oauth_client in db.query(orm.OAuthClient).options(
                 selectinload(orm.OAuthClient.allowed_roles).defer(
                     orm.Role.managed_by_auth
-                )
+                ),
+                # access_scopes() loads the orm.Spawner object, but the new
+                # display_name column isn't added until a subsequent db upgrade
+                # script runs, so we need to ignore the new column here
+                selectinload(orm.OAuthClient.spawner).defer(
+                    orm.Spawner.display_name
+                ),
+                # block every other relationship load (e.g. Spawner.user ->
+                # User.roles / User.shared_with_me selectin) that would touch
+                # not-yet-existing schema — same guard as the api_tokens loop
+                raiseload("*"),
             ):
                 allowed_scopes = set(roles.roles_to_scopes(oauth_client.allowed_roles))
                 allowed_scopes.update(access_scopes(oauth_client, db))
--- a/jupyterhub/alembic/versions/651f5419b74d_api_token_scopes.py
+++ b/jupyterhub/alembic/versions/651f5419b74d_api_token_scopes.py
@@ -136,12 +136,9 @@ def upgrade():
                 selectinload(orm.OAuthClient.allowed_roles).defer(
                     orm.Role.managed_by_auth
                 ),
-                # access_scopes() loads the orm.Spawner object, but the new
-                # display_name column isn't added until a subsequent db upgrade
-                # script runs, so we need to ignore the new column here
-                selectinload(orm.OAuthClient.spawner).defer(
-                    orm.Spawner.display_name
-                ),
+                # eager-load the spawner so access_scopes() doesn't lazy-load it
+                # (lazy loads are blocked by raiseload("*") below)
+                selectinload(orm.OAuthClient.spawner),
                 # block every other relationship load (e.g. Spawner.user ->
                 # User.roles / User.shared_with_me selectin) that would touch
                 # not-yet-existing schema — same guard as the api_tokens loop
--- a/jupyterhub/tests/populate_db.py
+++ b/jupyterhub/tests/populate_db.py
@@ -80,6 +80,21 @@ def populate_db(url):
         spawner.server = orm.Server()
         db.commit()
 
+        # per-server oauth client tied to a spawner
+        # regression coverage for jupyterhub/jupyterhub#5427: the api_token_scopes
+        # upgrade migration must resolve access_scopes() for a spawner's oauth
+        # client without lazy-loading future ORM relationships (User.roles,
+        # User.shared_with_me) that reference not-yet-existing schema.
+        # The spawner.oauth_client relationship was added in 2.0.
+        if jupyterhub.version_info >= (2,):
+            server_oauth_client = orm.OAuthClient(
+                identifier='jupyterhub-user-has-server'
+            )
+            db.add(server_oauth_client)
+            db.commit()
+            spawner.oauth_client = server_oauth_client
+            db.commit()
+
         # admin's spawner is not running
         spawner = orm.Spawner(name='', user=admin)
         db.add(spawner)


mleida-stratio:
The `pytest (3.11, subdomain, browser)` failure is unrelated flakiness, not caused by this change. The two failing tests are timing-sensitive Playwright UI tests:

- `test_token_form_expires_in[36hours]` — token-form dropdown rendered empty (`assert [] == [('1 Hour', ...)]`)
- `test_request_token_permissions[inherit-granted1]` — `to_have_url` 5s timeout waiting for `.*/hub/token`

This change only touches the `651f5419b74d` alembic migration and `tests/populate_db.py`, so it can't affect the token-form browser tests — and the same job passed on the previous run of this branch. The db-upgrade jobs that exercise the fix all pass: `pytest (3.12, postgres)`, `pytest (3.9, mysql)`, and `pytest (3.8, oldest_dependencies)`.

A re-run of the browser job should go green (I don't have permission to trigger it here).

PR #5437, opened 2026-07-02
DB upgrade: not safe to delete database

manics:
JupyterHub has expected spawners to store their state in the DB for a while now, to make it easier to handle upgrades and changes in naming conventions.


minrk:
nice!

PR #5443, opened 2026-07-11
add GET, PUT, PATCH for server endpoints

minrk:
This adds some methods discussed in #5192 to aid in migrating from invalid server names. All new methods are on the same endpoint `/users/:user_name/servers/:server_name`

- GET: get single server model
- PUT: create server, don't start it, 409 if exists
- PATCH: modify server (including rename), don't start it, 404 if not exists

Adds `Spawner.rename` method to handle possible rename events and reject the rename, if necessary.


minrk:
@manics if you could have a look. This adds the API endpoints needed for create / and rename, and changes the named server `create` action to just create the server instead of creating with a GET to /spawn.

with this finishing up the named-server changes and #5447 finishing up the unix socket support, I think we're ready for 6.0 prerelease

minrk:
linkcheck fix should be in #5448

jupyterhub/apihandlers/users.py
```diff
@@ -695,12 +750,79 @@ async def post(self, user_name, server_name=''):
...
+            assert spawner.name == new_server_name
+
+        # 2. set user options
+        user_options = body.get("user")
+        if "user_options" in body:
+            if not isinstance(body["user_options"], dict):
+                raise web.HTTPError(400, "user_options must be a dict")
+            orm_spawner.user_options = body["user_options"]
+            self.db.commit()
+
```
manics, line 861: We need to handle changing `display_name` too

jupyterhub/apihandlers/users.py
```diff
@@ -614,62 +613,118 @@ def delete(self, user_name, token_id):
...
+    async def post(self, user_name, server_name=''):
+        user = self.find_user(user_name)
+        if user is None:
+            # this can be reached if a token has `servers`
+            # permission on *all* users
+            raise web.HTTPError(404)
 
+        body = self.get_server_json_body() or {}
+        if server_name:
+            display_name = body.get("display_name")
```
manics, line 726: We need to decide and document how POST should behave when `display_name` is changed on an existing server. Do we treat it as modify+launch, or launch only?

jupyterhub/tests/test_api.py
```diff
@@ -2747,6 +2747,48 @@ async def test_update_server_activity(app, user, server_name, fresh):
...
+    if r.status_code == 200:
+        assert src_name not in user.orm_spawners
+        assert dst_name in user.orm_spawners
+    if active:
+        assert src_name in user.orm_spawners
+        assert dst_name not in user.orm_spawners
+
+
+async def test_patch_server(app, user, named_servers):
+    pass
```
manics: Todo?

manics (commented):
Makes sense to me

jupyterhub/apihandlers/users.py
minrk: I think POST always sets values and launches (unchanged, same as user_options). PATCH changes and doesn't launch (must exist), PUT creates and doesn't launch (must not exist).

minrk: now consistent and the rest api spec describes this

minrk:
thanks for reviewing! I think review has been addressed.

minrk:
I swear there was a feature request for GET for a single server, but I cannot find it to close. Possible it was just in chat / forum discussions.

manics:
Thanks! Looks mostly fine, but in testing I'm running into a redirect loop:

1. As a normal user create a new named server `old-name` in the web interface (`/hub/home`). Do not start it.
2. Using the API rename it: PATCH `/hub/api/users/demo/servers/old-name` `{"display_name": "new-name", "name": "new-name"}`
3. Refresh `/hub/home`
4. Start `new-name`
5. The server should start, but I end up in a redirect loop:

<details>

```
[D 2026-07-20 22:11:29.390 JupyterHub scopes:1012] Checking access to /hub/spawn/demo/new-name via scope servers!server=demo/new-name
[D 2026-07-20 22:11:29.390 JupyterHub pages:264] Triggering spawn with default options for demo:new-name
[D 2026-07-20 22:11:29.390 JupyterHub base:1117] Initiating spawn for demo:new-name
[D 2026-07-20 22:11:29.390 JupyterHub base:1121] 0/100 concurrent spawns
[D 2026-07-20 22:11:29.390 JupyterHub base:1126] 0 active servers
[I 2026-07-20 22:11:29.422 JupyterHub provider:744] Creating oauth client jupyterhub-user-demo-old-name
[D 2026-07-20 22:11:29.445 JupyterHub user:972] Calling Spawner.start for demo:new-name
[I 2026-07-20 22:11:29.447 JupyterHub spawner:2059] Spawning jupyterhub-singleuser
[D 2026-07-20 22:11:29.467 JupyterHub spawner:1699] Polling subprocess every 30s
[D 2026-07-20 22:11:29.474 JupyterHub utils:294] Waiting 30s for server at http://127.0.0.1:52359/user/demo/new-name/api
[I 2026-07-20 22:11:30.391 JupyterHub log:192] 302 GET /hub/spawn/demo/new-name -> /hub/spawn-pending/demo/new-name (demo@::ffff:192.168.X.X) 1015.95ms
[D 2026-07-20 22:11:30.419 JupyterHub scopes:1012] Checking access to /hub/spawn-pending/demo/new-name via scope servers!server=demo/new-name
[I 2026-07-20 22:11:30.419 JupyterHub pages:478] demo:new-name is pending spawn
[I 2026-07-20 22:11:30.428 JupyterHub log:192] 200 GET /hub/spawn-pending/demo/new-name (demo@::ffff:192.168.X.X) 19.43ms
[D 2026-07-20 22:11:30.585 JupyterHub scopes:1012] Checking access to /hub/api/users/demo/servers/new-name/progress via scope read:servers!server=demo/new-name
[I 2026-07-20 22:11:33.500 ServerApp] Extension package jupyterlab took 0.1647s to import
[I 2026-07-20 22:11:34.058 ServerApp] jupyter_lsp | extension was successfully linked.
[I 2026-07-20 22:11:34.063 ServerApp] jupyter_server_terminals | extension was successfully linked.
[I 2026-07-20 22:11:34.063 JupyterHubSingleUser] Starting jupyterhub single-user server extension version 6.0.0.dev
[I 2026-07-20 22:11:34.064 JupyterHubSingleUser] Using default url from server extension lab: /lab
[I 2026-07-20 22:11:34.067 ServerApp] jupyterhub | extension was successfully linked.
[W 2026-07-20 22:11:34.069 LabApp] 'extra_template_paths' was found in both NotebookApp and ServerApp. This is likely a recent change. This config will only be set in NotebookApp. Please check if you should also config these traits in ServerApp for your purpose.
[I 2026-07-20 22:11:34.072 ServerApp] jupyterlab | extension was successfully linked.
[W 2026-07-20 22:11:34.074 JupyterNotebookApp] 'extra_template_paths' was found in both NotebookApp and ServerApp. This is likely a recent change. This config will only be set in NotebookApp. Please check if you should also config these traits in ServerApp for your purpose.
[I 2026-07-20 22:11:34.077 ServerApp] notebook | extension was successfully linked.
[I 2026-07-20 22:11:34.435 ServerApp] notebook_shim | extension was successfully linked.
[I 2026-07-20 22:11:34.461 ServerApp] notebook_shim | extension was successfully loaded.
[I 2026-07-20 22:11:34.463 ServerApp] jupyter_lsp | extension was successfully loaded.
[I 2026-07-20 22:11:34.465 ServerApp] jupyter_server_terminals | extension was successfully loaded.
[I 2026-07-20 22:11:34.469 JupyterHub log:192] 200 GET /hub/api (@127.0.0.1) 0.51ms
[I 2026-07-20 22:11:34.471 JupyterHubSingleUser] Updating Hub with activity every 300 seconds
[I 2026-07-20 22:11:34.472 ServerApp] jupyterhub | extension was successfully loaded.
[I 2026-07-20 22:11:34.479 LabApp] JupyterLab extension loaded from /srv/venv/lib/python3.12/site-packages/jupyterlab
[I 2026-07-20 22:11:34.479 LabApp] JupyterLab application directory is /srv/venv/share/jupyter/lab
[I 2026-07-20 22:11:34.480 LabApp] Extension Manager is 'pypi'.
[I 2026-07-20 22:11:34.551 ServerApp] jupyterlab | extension was successfully loaded.
[I 2026-07-20 22:11:34.555 ServerApp] notebook | extension was successfully loaded.
[I 2026-07-20 22:11:34.556 ServerApp] Serving notebooks from local directory: /tmp/demo
[I 2026-07-20 22:11:34.556 ServerApp] Jupyter Server 2.20.0 is running at:
[I 2026-07-20 22:11:34.556 ServerApp] http://127.0.0.1:52359/user/demo/new-name/lab?token=...
[I 2026-07-20 22:11:34.556 ServerApp] Use Control-C to stop this server and shut down all kernels (twice to skip confirmation).
[D 2026-07-20 22:11:34.564 JupyterHub base:372] Recording first activity for <APIToken('15c2...', user='demo', client_id='jupyterhub')>
[D 2026-07-20 22:11:34.571 JupyterHub scopes:1012] Checking access to /hub/api/users/demo/activity via scope users:activity!user=demo
[D 2026-07-20 22:11:34.577 JupyterHub users:1192] Activity for user demo: 2026-07-20T22:11:34.449865Z
[D 2026-07-20 22:11:34.577 JupyterHub users:1210] Activity on server demo/new-name: 2026-07-20T22:11:34.449865Z
[I 2026-07-20 22:11:34.583 JupyterHub log:192] 200 POST /hub/api/users/demo/activity (demo@127.0.0.1) 20.13ms
[I 2026-07-20 22:11:35.072 ServerApp] Skipped non-installed server(s): basedpyright, bash-language-server, dockerfile-language-server-nodejs, javascript-typescript-langserver, jedi-language-server, julia-language-server, pyrefly, pyright, python-language-server, python-lsp-server, r-languageserver, sql-language-server, texlab, typescript-language-server, unified-language-server, vscode-css-languageserver-bin, vscode-html-languageserver-bin, vscode-json-languageserver-bin, yaml-language-server
[I 2026-07-20 22:11:36.960 ServerApp] 200 GET /user/demo/new-name/api (@127.0.0.1) 3.16ms
[D 2026-07-20 22:11:36.962 JupyterHub utils:330] Server at http://127.0.0.1:52359/user/demo/new-name/api responded in 7.49s
[D 2026-07-20 22:11:36.962 JupyterHub _version:72] jupyterhub and jupyterhub-singleuser both on version 6.0.0.dev
[I 2026-07-20 22:11:36.962 JupyterHub base:1146] User demo:new-name took 7.572 seconds to start
[I 2026-07-20 22:11:36.963 JupyterHub proxy:331] Adding user demo to proxy /user/demo/old-name/ => http://127.0.0.1:52359
[D 2026-07-20 22:11:36.964 JupyterHub proxy:921] Proxy: Fetching POST http://127.0.0.1:8001/api/routes/user/demo/old-name
2026-07-20T22:11:36.969Z [ConfigProxy] info: Adding route /user/demo/old-name -> http://127.0.0.1:52359
2026-07-20T22:11:36.970Z [ConfigProxy] info: Route added /user/demo/old-name -> http://127.0.0.1:52359
2026-07-20T22:11:36.972Z [ConfigProxy] info: 201 POST /api/routes/user/demo/old-name 
[I 2026-07-20 22:11:36.973 JupyterHub users:1086] Server demo:new-name is ready
[I 2026-07-20 22:11:36.976 JupyterHub log:192] 200 GET /hub/api/users/demo/servers/new-name/progress?_xsrf=[secret] (demo@::ffff:192.168.X.X) 6395.90ms
[D 2026-07-20 22:11:37.017 JupyterHub scopes:1012] Checking access to /hub/spawn-pending/demo/new-name via scope servers!server=demo/new-name
[I 2026-07-20 22:11:37.019 JupyterHub log:192] 302 GET /hub/spawn-pending/demo/new-name -> /user/demo/new-name/ (demo@::ffff:192.168.X.X) 8.31ms
[I 2026-07-20 22:11:37.031 JupyterHub log:192] 302 GET /user/demo/new-name/ -> /hub/user/demo/new-name/ (@::ffff:192.168.X.X) 1.68ms
[I 2026-07-20 22:11:37.053 JupyterHub log:192] 302 GET /hub/user/demo/new-name/ -> /user/demo/new-name/?redirects=1 (demo@::ffff:192.168.X.X) 6.83ms
[I 2026-07-20 22:11:37.071 JupyterHub log:192] 302 GET /user/demo/new-name/?redirects=1 -> /hub/user/demo/new-name/?redirects=1 (@::ffff:192.168.X.X) 0.95ms
[W 2026-07-20 22:11:37.089 JupyterHub base:1943] Redirect loop detected on /hub/user/demo/new-name/?redirects=1
[I 2026-07-20 22:11:39.092 JupyterHub log:192] 302 GET /hub/user/demo/new-name/?redirects=1 -> /user/demo/new-name/?redirects=2 (demo@::ffff:192.168.X.X) 2009.33ms
[I 2026-07-20 22:11:39.139 JupyterHub log:192] 302 GET /user/demo/new-name/?redirects=2 -> /hub/user/demo/new-name/?redirects=2 (@::ffff:192.168.X.X) 1.01ms
[W 2026-07-20 22:11:39.160 JupyterHub base:1943] Redirect loop detected on /hub/user/demo/new-name/?redirects=2
[I 2026-07-20 22:11:43.163 JupyterHub log:192] 302 GET /hub/user/demo/new-name/?redirects=2 -> /user/demo/new-name/?redirects=3 (demo@::ffff:192.168.X.X) 4008.44ms
[I 2026-07-20 22:11:43.206 JupyterHub log:192] 302 GET /user/demo/new-name/?redirects=3 -> /hub/user/demo/new-name/?redirects=3 (@::ffff:192.168.X.X) 1.04ms
[W 2026-07-20 22:11:43.225 JupyterHub base:1943] Redirect loop detected on /hub/user/demo/new-name/?redirects=3
```

</details>

Looks like `old-name` is still being used:
```
[I 2026-07-20 22:11:29.422 JupyterHub provider:744] Creating oauth client jupyterhub-user-demo-old-name
...
[I 2026-07-20 22:11:36.963 JupyterHub proxy:331] Adding user demo to proxy /user/demo/old-name/ => http://127.0.0.1:52359
[D 2026-07-20 22:11:36.964 JupyterHub proxy:921] Proxy: Fetching POST http://127.0.0.1:8001/api/routes/user/demo/old-name
2026-07-20T22:11:36.969Z [ConfigProxy] info: Adding route /user/demo/old-name -> http://127.0.0.1:52359
2026-07-20T22:11:36.970Z [ConfigProxy] info: Route added /user/demo/old-name -> http://127.0.0.1:52359
2026-07-20T22:11:36.972Z [ConfigProxy] info: 201 POST /api/routes/user/demo/old-name 
```

minrk:
Thanks for testing, I'll try to make sure that's covered in the tests. There's probably some stale state somewhere that doesn't get updated on rename.

minrk:
I believe I've fixed the issue and included verification in the test coverage that the spawner at the new name actually has the right name (it was something flaky in the spawner wrapper around the orm spawner, which can get out of sync, but we don't need to have a Spawner wrapper since we enforce that the server is not active before patching it).

jupyterhub/apihandlers/users.py
```diff
@@ -695,12 +752,107 @@ async def post(self, user_name, server_name=''):
...
+        if new_server_name:
+            if new_server_name in user.orm_spawners:
+                raise web.HTTPError(
+                    409, f"User {user.name} already has server {new_server_name!r}"
+                )
+            self.log.info(
+                f"Renaming spawner {user.name}/{server_name} -> {new_server_name}"
+            )
+            spawner = user.get_spawner(server_name)
+            await spawner.rename(server_name, new_server_name)
```
manics: Just one (hopefully) more thing- this could take a long time to run, e.g. if it's moving files, so we need to prevent someone starting the server in the middle of the rename.

manics: On second thoughts we should prevent all operations on the server/spawner, e.g. you shouldn't be able to call rename a second time (likely if someone adds UI support and a user makes a typo in the rename)

https://github.com/jupyterhub/jupyterhub/blob/6f8cfec662a79b8fde990eb90f942f51e51b9ec7/jupyterhub/spawner.py#L94-L100
Is it worth adding a more generic `is_locked()` flag?

manics:
For a separate PR: should we remove `allow_existing_invalid_named_servers` since this PR adds a proper way to rename servers, and we could provide a rename script instead?
https://github.com/jupyterhub/jupyterhub/blob/38ab88f73b47b359e28e03c091433d1a2c6093e5/jupyterhub/app.py#L1391-L1409

minrk:
> should we remove allow_existing_invalid_named_servers since this PR adds a proper way to rename servers, and we could provide a rename script instead?

Since we already have it, I think leaving it as-is for this release still makes sense. Just because we _have_ a rename hook doesn't mean that a given deployment/Spawner will have implemented what's necessary for every rename to not break.

jupyterhub/apihandlers/users.py
minrk: we currently have `spawner.active` for this - if _anything_ is happening to the server, `.active` is used as a lock to prevent concurrent operations. Mostly those are stop/start related, but you're right that rename needs to be included in this.

Adding a different `locked` method will mean every current check for `active` will need to be replaced with `is_locked()`.

minrk: added `rename` to the existing `pending` lock for operations on a server, which should cover this, I think.

minrk:
@manics I think this is all set

.pre-commit-config.yaml
```diff
@@ -95,18 +95,3 @@ repos:
           - --update
         files: jupyterhub/scopes.py
         pass_filenames: false
-
-      # run eslint in the jsx directory
-      # need to pass through 'jsx:install-run' hook in
-      # top-level package.json to ensure dependencies are installed
-      # eslint pre-commit hook doesn't really work with eslint 9,
-      # so use `npm run lint:fix`
-      - id: jsx-eslint
```
minrk: I removed this unused pre-commit hook (it can still be run directly wth `npm run`) because pre-commit.ci started failing to validate it with a clearly incorrect error message. It didn't seem worth it to figure out the fix since it doesn't get used anyway.

jupyterhub/apihandlers/users.py
```diff
@@ -695,12 +752,111 @@ async def post(self, user_name, server_name=''):
...
+            raise web.HTTPError(400, "Must specify at least one field to modify.")
+        supported_keys = {"name", "display_name", "user_options"}
+        unhandled_keys = body_keys.difference(supported_keys)
+        if unhandled_keys:
+            raise web.HTTPError(
+                400,
+                f"Unexpected fields: {unhandled_keys}, supported fields: {supported_keys.join(',')}",
+            )
+
+        if new_server_name:
```
manics:
```suggestion
        if new_server_name is not None:
```
I think we should return an error if someone passes `{"name": "", ...}` instead of ignoring it

jupyterhub/apihandlers/users.py
```diff
@@ -695,12 +752,111 @@ async def post(self, user_name, server_name=''):
...
+
+        # cannot patch active servers
+        if server_name in user.spawners:
+            spawner = user.spawners[server_name]
+            if spawner.active:
+                raise web.HTTPError(
+                    400,
+                    f"{spawner._log_name} is active. Please stop it before modifying it.",
+                )
+
```
manics, line 819: I tried
`PATCH users/demo/servers/ json={"display_name": "This was the default server", "name": "default"})`
and it renamed the default server! We should return an error instead.

manics (commented):
A couple of places where we can tighten up the validation, but otherwise this should be good to merge!

There's an inconsistency in the UI when renaming takes a while- The `start` button is hidden but the `stop` button is still there, incorrectly giving the impressing the server is running:
Image
but we can leave that for a future PR since it doesn't break anything.

jupyterhub/apihandlers/users.py
minrk: I think that's the right behavior - I think when named servers are enabled, we should aim to treat the default server exactly like all other servers, where the name happens to be `""`. I'll add a test to make sure renaming the default server works and doesn't make a mess.

manics: OK! We should bear in mind `named_server_limit_per_user` - what happens if renaming the default server takes you over the limit?

jupyterhub/apihandlers/users.py
minrk: exercise added for attempts to rename to default server, which fails with 409 because it always exists

minrk:
Fixed the stop button being available during rename

jupyterhub/apihandlers/users.py
minrk: good point! It ought to fail, since this is the one rename action that actually adds a server in practice. I'll make sure that's tested.

Maybe for 7.0 we can try to refactor named server vs default server handling to try to remove most of the special default server handling. There are quite a few places where the default server is assumed to exist, but I think we can work to remove those.

minrk: added tests covering the server limit

manics (approved):
There's a minor UI bug where a slow rename of the default server means `/hub/home` shows a  `Stop My Server` button as if it were running, but we can deal with that later.

Removing the special handling for the default server in 7.0 sounds good. Are you thinking of getting rid of the `""` completely in the backend so the default server is a named server with a well-known name (in which case can we treat that well-known name as an invalid name now, to make the 7.0 migration easier), or would you allow any valid server-name and the default server is just the first server to be created?

minrk:
> Are you thinking of getting rid of the "" completely in the backend so the default server is a named server with a well-known name 

Yeah, I was roughly thinking of using `default` and disallowing `""`. Only when named servers are enabled, though. It's not 100% clear, though.

> in which case can we treat that well-known name as an invalid name now

I initially agreed, but I'm not sure about that, because the user should still be able to treat the default name exactly like all others (deleteable, createable, etc.), so all the operations on that name should work. There are other strategies than a default name, like when no name is specified use the most recently launched, first created, etc. But I think we need to look at the various ways we deal with the default servers and think it through.

PR #5447, opened 2026-07-15
use aiohttp for internal http requests

### LINKED ISSUE #5421: consider trading tornado http client for another http client
We currently use tornado's internal AsyncHTTPClient for our inter-process communication.

Since we're all on asyncio now, we could consider another option such as [aiohttp](https://docs.aiohttp.org/) or [httpx](https://www.python-httpx.org) or [niquests](https://github.com/jawah/niquests). The advantage of these is it would let us get away from pycurl, which is an _occasional_ headache (not often, but once in a while there are wheel/ssl cert issues with certain updates and images). pycurl is also a source of some complexity in #5397 .

If pursuing this, we should check that they still solve the same problems we currently solve with pycurl, which is high concurrent request load. Larger deployments have needed pycurl to avoid pipelining and spurious timeout issues (IIRC, largely due to timeouts failing due to the time spent in the request queue instead of during the actual request processing), but that's in large part due to the relatively basic implementation of SimpleAsyncHTTPClient. All of these client libraries have client functions as their primary purpose, unlike tornado's client implementation, so I think we can reasonably have higher expectations of them.

closed by #5447

minrk:
adds `jupyterhub.httpclient.fetch` for making internal requests.

uses a custom Connector, which combines TCP and UnixConnector logic, getting unix socket path from url, which the default UnixConnector doesn't support

follow-up to simplify things from #5397 without side-tracking that PR too much.

this also removes pycurl from the equation

I also checked out httpx and niquests, and aiohttp seems like the best fit.

Marking this as breaking since it's _significant_, even though it's only private internal APIs. I can imagine some things being broken by this, and need to test scaling, etc.

I think aiohttp will handle load a lot better than tornado, but we still need to see.

closes #5421

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,7 @@ node_modules
 dist
 docs/_build
 docs/build
-docs/source/reference/metrics.md
+docs/source/includes/metrics_table.md
 
 .ipynb_checkpoints
 .virtual_documents
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -95,18 +95,3 @@ repos:
           - --update
         files: jupyterhub/scopes.py
         pass_filenames: false
-
-      # run eslint in the jsx directory
-      # need to pass through 'jsx:install-run' hook in
-      # top-level package.json to ensure dependencies are installed
-      # eslint pre-commit hook doesn't really work with eslint 9,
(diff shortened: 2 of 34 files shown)


minrk:
I'll merge this later today to keep things rolling

PR #5207, opened 2025-11-21
add configurable service.timeout

minrk:
and simplify concurrent wait up & wait for exit for managed services by using concurrent tasks instead of a for loop with short timeouts

closes #5138

PR #5455, opened 2026-07-17
fix wait-up on managed services

minrk:
futures was meant to include wait_up and wait_for_stop, but it only had wait_for_stop, so it would block forever until a service stopped.

Regression in #5207, not yet released.


minrk:
oop, no, shouldn't backport this, 5.x is not affected

PR #5461, opened 2026-07-21
Invalid UTF-8 JSON request body returns HTTP 500 instead of HTTP 400

Tech-Abhang:
Invalid UTF-8 request bodies with Content-Type: application/json currently raise a UnicodeDecodeError, resulting in an HTTP 500 response.

This change moves UTF-8 decoding into the existing error handling so malformed UTF-8 is treated as invalid JSON and returns HTTP 400 instead.

Testing

Added a regression test for invalid UTF-8 JSON requests.
Verified that the new test fails on the original implementation and passes with this change.
Ran jupyterhub/tests/test_api.py; all tests passed.

### DIFF
--- a/jupyterhub/apihandlers/base.py
+++ b/jupyterhub/apihandlers/base.py
@@ -112,12 +112,12 @@ class APIHandler(BaseHandler):
         """Return the body of the request as JSON data."""
         if not self.request.body:
             return None
+        body = self.request.body.strip().decode("utf-8", "replace")
 
         try:
-            body = self.request.body.strip().decode('utf-8')
             model = json.loads(body)
         except Exception:
-            self.log.debug("Bad JSON bytes: %r", self.request.body)
+            self.log.debug("Bad JSON: %r", body)
             self.log.error("Couldn't parse JSON", exc_info=True)
             raise web.HTTPError(400, 'Invalid JSON in body of request')
         return model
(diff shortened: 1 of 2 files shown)


jupyterhub/tests/test_api.py
```diff
@@ -104,6 +104,27 @@ async def test_post_content_type(app, content_type, status):
...
 
 
+async def test_invalid_utf8_json_returns_400(app):
+    """Sending invalid-UTF8 bytes with Content-Type: application/json should return 400."""
+    db = app.db
+    # ensure admin user exists
+    user = find_user(db, 'admin')
+    if user is None:
+        user = add_user(db, app, name='admin', admin=True)
+    cookies = await app.login_user('admin')
```
minrk: this is not required, `api_request` already works with a default user.

jupyterhub/apihandlers/base.py
```diff
@@ -112,11 +112,12 @@ def get_json_body(self):
         """Return the body of the request as JSON data."""
         if not self.request.body:
             return None
-        body = self.request.body.strip().decode('utf-8')
```
minrk, line 115: Rather than moving this to the try, can we add the `errors="replace"` argument to decode, so decode can't fail? (`.decode("utf-8", "replace")`)

minrk (commented):
Thanks for the PR!

minrk:

Issue #5459, opened 2026-07-20
Provide a way to render the spawn form when the user's default server is running


sunu:
When a default server is running for a user in Jupyterhub, any new request to `/hub/spawn` gets redirected to the running server. This causes fancy-profiles permalinks to break because they depend on the spawn form being rendered to load the fancy-profiles ui. related issue: https://github.com/2i2c-org/jupyterhub-fancy-profiles/issues/164 

### Proposed change
@yuvipanda in https://github.com/2i2c-org/jupyterhub-fancy-profiles/issues/164 
> The way out of this is to use named servers, so each permalink can have its own server. But how do you choose the name? And how do you handle clashes of the name? If the user clicks the same URL twice, if it has the same name, you're back in this behavior again.
>
>We had the same problem with JupyterLab workspaces a while back, and the solution was to add a feature to JupyterLab to autogenerate them! https://github.com/jupyterlab/jupyterlab/pull/5950
>
>So let's implement a similar solution here, in JupyterHub, that autogenerates new server names? That will allow us to make links that will consistently show the form each time they are clicked.
>
>It does mean you'll have some running servers, but the idle culler can take care of them.

### Alternative options

An alternative option could be to add a new hook that prepocesses spawn options from queries, similar to what's proposed in https://github.com/jupyterhub/jupyterhub/issues/5101#issuecomment-3079771837. This hook could also decide whether to spawn a server or render the form, and optionally set `server_name`. That would let downstream users like fancy-profiles define a hook to force a form render or boot a new named server, fixing this issue while also enabling the use of proper query parameters in fancy-profiles as described in https://github.com/jupyterhub/jupyterhub/issues/5101.

### Who would use this feature?

This would unblock a major usability gap in fancy-profiles https://github.com/2i2c-org/jupyterhub-fancy-profiles/issues/164

minrk:
So the idea being a link that _always_ shows spawn UI, even when servers are running?

We already have a hook for spawn vs rendering a form, which is returning no form from `options_form`. but that all happens after the determination that the requested server isn't already running.

What is the idea for how to determine:

1. the requested server is already running, link to it (current behavior)
2. the requested server is running but _not_ what is desired, so
  a. the server should stop?
  b. the user should be presented with the _option_ of stopping the server
  c. start another server with a new name (only an option if named servers are available and user hasn't exhausted their capacity

We could change the handler for `/hub/spawn` to present info rather than automatically redirect.

[...]

PR #5462, opened 2026-07-21
WIP: Introduce a hook to reinterpret an incoming request to /hub/spawn

### LINKED ISSUE #5101: Allow for query parameters other than `next` on profile selection screen
I apologize for the slightly convoluted use-case here, but I hope the ticket broadly makes sense:

We are working on [jupyterhub-fancy-profiles](https://github.com/2i2c-org/jupyterhub-fancy-profiles/) as an alternative frontend interface for the profile selection screen, integrating binder-style dynamic image building and other frontend improvements.

We recently worked on a feature to allow for "permalinks" to a particular state of the profile selection form - so one could share a link that had pre-selected options for resources, a custom image, etc.

We wanted to implement this using a query parameter to pass profile options, so we'd have something like `https://hub.example.com/hub/spawn/?fancy-profile-options={'resources': "foo", 'image': 'custom-image'}`

Unfortunately, currently, any query parameters (other than `next`) attempt to get processed as as a form submission with those values, resulting in an error on the page when you attempt to use any query parameters other than `next`. For eg. if you try https://<your-hub-url>/hub/spawn/?foo=bar you see it attempts to start a server with a "foo=bar" over-ride.

The relevant code that tries to process the query parameters and whitelists the `next` parameter is here: https://github.com/jupyterhub/jupyterhub/blob/main/jupyterhub/handlers/pages.py#L186
  

### Proposed change

Allow configuration of query parameters other than `next` that will be ignored by the backend on the `/hub/spawn/` page, allowing custom frontends to accept query parameters. Another option would be to have a fixed query parameter like `customOptions` or something that is whitelisted which an alternative frontend can use for custom options.

Am quite unfamiliar with the details of the Jupyterhub architecture, so there could be other workarounds / solutions that I've missed.

### Alternative options

For the Permalink feature on jupyterhub-fancy-profiles, we switched to using "hash parameters", and have added the option selections to after a `#` in the URL, so that part of the URL is not processed by the backend at all. So the URL would look like `https://hub.example.com/hub/spawn/#fancy-profiles-config=....` . While this works, it has several draw-backs:

 - Since the value after `#` never gets sent to the server, it makes handling redirection to login, etc. much more complex (one can achieve this with carefully double-encoding bits in the URL, but it is very brittle and non-ideal).
 - A query parameter seems more correct for what it's doing in terms of URL design, and would make it much easier when combining with other query parameters and user-redirect links.

While we have the permalink feature working in jupyterhub-fancy-profiles currently, I would really love to find a good way to make it work with options passed in query parameters rather than the current `#` parameters, which feels like a not-so-sustainable hack.

jupyterhub-fancy-profiles is being used across many hubs that need the custom image building feature. The Permalinking feature has been one of the most requested features, and this would enable a clean and long-term implementation.

The naive solution would be to add another query parameter to get ignored by kubespawner here: https://github.com/jupyterhub/jupyterhub/blob/main/jupyterhub/handlers/pages.py#L186 - of course, that doesn't seem very nice. I wonder if you all would be open to a solution that makes this configurable - i.e. somewhere in config one can specify query parameters to "whitelist"?

I realize this is a bit of a "weird one" - I do think we missed that Jupyterhub tries to process any query parameters in the URL by default when starting to develop the feature. Perhaps we should consider a drastically different approach?

Would appreciate any thoughts and suggestions here, and if there's an agreement that this would be okay to implement, we're happy to take on the work / make a PR.

Please let me know if I can add any additional context or explanations here.

Thank you again for all your work on Jupyterhub.

cc @sunu @yuvipanda

sunu:
As we're discussing in https://github.com/jupyterhub/jupyterhub/issues/5459, we need a way to make sure we can render the spawn form on requests to /hub/spawn even when the default server is running.

This is a draft of one proposed approach where we introduce a new `spawn_intent_hook` that can decide whether to redirect the spawn request to an existing/new server or to render the spawn options form.

This gives the hook user the flexibility to decide whether:
- keep the current behavior and always redirect to the default server
- redirect to a running named server if it already exists
- or redirect to a new named server or some other custom behavior

For example, for the fancy-profiles use case, we can use a hook like this to make sure we can spawn a uniquely named named-server on every click of a permalink:

```
def spawn_intent_hook(user, server_name, query):
    """On GET /hub/spawn carrying the permalink payload, derive a unique named
    server and redirect to its spawn page (so each click gets its own server
    even when one is already running).
    """
    payload = query.get(PERMALINK_QUERY_PARAM)
    if not payload:
        return None
    if server_name:
        # already routed to the named server so we render the pre-filled form
        return {"render_spawn_form": True}
    name = new_unique_server_name(user, payload)
    target = url_path_join(
        user.settings.get("base_url", "/"), "hub", "spawn", user.escaped_name, name
    )
    return {"redirect_url": url_concat(target, {PERMALINK_QUERY_PARAM: payload[0]})}
 ```
 
 This also fixes https://github.com/jupyterhub/jupyterhub/issues/5101 as a side effect of us having more control over whether we render the form or spawn a server.

### DIFF
--- a/jupyterhub/app.py
+++ b/jupyterhub/app.py
@@ -1772,6 +1772,27 @@ class JupyterHub(Application):
         """,
     ).tag(config=True)
 
+    spawn_intent_hook = Callable(
+        None,
+        allow_none=True,
+        help="""
+        Callable to reinterpret an incoming request to the /spawn page.
+
+        Receives 3 parameters:
+        1. user - the currently authenticated user
+        2. server_name - the route-derived server name ('' on the default
+           /hub/spawn page, the server name on /hub/spawn/:user/:server)
+        3. query - decoded GET query args (a dict of lists), with 'next' removed
+
+        It should return an optional dict with any of the keys:
+        - redirect_url - reinterpret the request and redirect here
+        - render_spawn_form - True to render the options form instead of
+          spawning directly from the query arguments
+
+        Return None or an empty dict to preserve current behavior.
+        """,
+    ).tag(config=True)
+
     use_legacy_stopped_server_status_code = Bool(
         False,
         help="""
@@ -3295,6 +3316,7 @@ class JupyterHub(Application):
             subdomain_host=self.subdomain_host,
             domain=self.domain,
             implicit_spawn_seconds=self.implicit_spawn_seconds,
+            spawn_intent_hook=self.spawn_intent_hook,
             allow_named_servers=self.allow_named_servers,
             default_server_name=self._default_server_name,
             named_server_limit_per_user=self.named_server_limit_per_user,
--- a/jupyterhub/handlers/pages.py
+++ b/jupyterhub/handlers/pages.py
@@ -195,6 +195,16 @@ class SpawnHandler(BaseHandler):
             self.log.error(error_message)
             raise web.HTTPError(400, error_message)
 
+    async def run_spawn_intent_hook(self, user, server_name, query):
+        """Run the spawn_intent_hook if defined
+
+        Returns the hook's result dict (with optional 'redirect_url' and
+        'render_spawn_form' keys), or None if no hook is configured.
+        """
+        hook = self.settings.get('spawn_intent_hook')
+        if hook is not None:
+            return await maybe_future(hook(user, server_name, query))
+
     @needs_scope("servers")
     async def _get(self, user_name, server_name, display_name):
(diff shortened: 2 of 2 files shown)


sunu:
Closing. See https://github.com/jupyterhub/jupyterhub/issues/5459#issuecomment-5075348843 for context.

PR #5463, opened 2026-07-21
add JupyterHub.extra_user_scopes

### LINKED ISSUE #5446: Fail if 'minimum' scopes for starting a server aren't met
We had an outage earlier caused by the `user` role not having `self` scope accidentally, since scopes are a list and difficult to override (more info in https://github.com/2i2c-org/incident-reports/issues/32). I can't think of a situation where `user` should *not* have `self` scope. Maybe let's validate at startup what is the 'minimum' set of scopes, and fail loudly if we don't have that?

## Definition of Done

- [ ] We fail loudly at startup if any roles (user? what else?) don't have the *minimum* set of scopes needed for user servers to start
- [ ] We decide this isn't necessary, because there is a valid use case for not allowing `self` scope to user role

minrk:
avoids need to redefine user role preserving default scopes to add permissions

closes #5446

### DIFF
--- a/docs/source/rbac/roles.md
+++ b/docs/source/rbac/roles.md
@@ -115,40 +115,88 @@ In case the role with a certain name already exists in the database, its definit
 
 Role definitions can include those of the "default" roles listed above (admin excluded),
 if the default scopes associated with those roles do not suit your deployment.
+
 For example, to specify what permissions the $JUPYTERHUB_API_TOKEN issued to all single-user servers
-has,
-define the `server` role.
+has, you may define the `server` role.
+However, starting with JupyterHub 4, it is recommended to use the {attr}`.Spawner.server_token_scopes` option for this instead.
 
 To restore the JupyterHub 1.x behavior of servers being able to do anything their owners can do,
 use the scope `inherit` (for 'inheriting' the owner's permissions):
 
 ```python
-c.JupyterHub.load_roles = [
- {
-   'name': 'server',
-   'scopes': ['inherit'],
- }
-]
+c.JupyterHub.server_token_scopes = {"inherit"}
 ```
 
 or, better yet, identify the specific [scopes][] you want server environments to have access to.
+If you choose to _restrict_ the server token, the one required scope for the server token `users:activity!user` will be added automatically.
 
-[scopes]: available-scopes-target
-
-If you don't want to get too detailed,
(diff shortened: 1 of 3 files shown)


minrk:
latest release of jupyverse is broken due to missing dependencies, failure should be ignored until that's fixed upstream.

docs/source/rbac/roles.md
```diff
@@ -115,41 +115,89 @@ In case the role with a certain name already exists in the database, its definit
 
 Role definitions can include those of the "default" roles listed above (admin excluded),
 if the default scopes associated with those roles do not suit your deployment.
+
 For example, to specify what permissions the $JUPYTERHUB_API_TOKEN issued to all single-user servers
```
agoose77:
```suggestion
For example, to specify what permissions the `$JUPYTERHUB_API_TOKEN` issued to all single-user servers
```

jupyterhub/app.py
```diff
@@ -2373,6 +2394,16 @@ async def init_role_creation(self):
...
+                if role_name == "user" and "scopes" in role_spec:
+                    role_scopes = role_spec["scopes"]
+                    if self.extra_user_scopes:
+                        self.log.warning(
+                            f"user role overridden with scopes={role_scopes}, extra_user_scopes={self.extra_user_scopes} will be ignored."
+                        )
+                    if "self" not in role_scopes:
+                        self.log.warning(
+                            f"user role overridden with scopes={role_scopes}, which does not include 'self', users may lack necessary permissions. Consider setting `c.JupyterHub.extra_user_scopes` to add user permissions."
+                        )
```
agoose77, line 2406: Am I reading this properly, that `extra_user_scopes` and `load_roles` are mutually exclusive? If so, we should update the docs to make that clear — that one can't use `extra_user_scopes` alongside `load_roles["user"]`

agoose77 (changes requested):

jupyterhub/app.py
minrk: Only that setting user _scopes_ in both ways is not permitted. Defining the user role can also be a way to restrict user access. I'll update.

minrk: added a note to this effect

agoose77 (approved):

minrk:
Thanks!

PR #5464, opened 2026-07-21
add SpawnException

### LINKED ISSUE #5357: Handle erroring and metrics when a server launch was intentionally denied by `c.Spawner.pre_spawn_hook`

minrk:
adds `jupyterhub.spawner.SpawnException`, and updates the server spawn prometheus metric for more fine-grained control

- prometheus metric gets separate `error` and `failure` status values, instead of just `failure`. `error` is for unhandled errors (like timeouts), where `failure` is for rejections, such as explicitly raised `HTTPError(400)` or the new `SpawnException`.
  - Built-in failures include exceeding quotas, already pending launches
  - Built-in errors include timeouts
- metric also gets a `reason` label to differentiate rejection reasons, so a deployment can categorize rejected launches better
- `SpawnException` has separate fields for http status, reason user-facing message, html message, and log message. Pretty much all of which were technically achievable  through our `.jupyterhub_message` attributes, but not as clean as a dedicated exception class
- SpawnException should never trigger a logged traceback

closes #5357

### DIFF
--- a/docs/source/reference/spawners.md
+++ b/docs/source/reference/spawners.md
@@ -139,7 +139,7 @@ Alternatively `.jupyterhub_message` is rendered as unformatted text.
 If both attributes are not present, an HTTPError's `message` is shown to the user.
 For other unhandled errors, a default "Internal Server Error" message will be given.
 
-raising an HTTPError and setting `.jupyterhub_message` and `.jupyterhub_html_message` will generally be equivalent to raising a SpawnException:
+Raising an HTTPError and setting `.jupyterhub_message` and `.jupyterhub_html_message` will generally be equivalent to raising a SpawnException:
 
 ```python
 from tornado.web import HTTPError
--- a/docs/source/reference/spawners.md
+++ b/docs/source/reference/spawners.md
@@ -114,13 +114,47 @@ class MySpawner(Spawner):
 
 #### Exception handling
 
-When `Spawner.start` raises an Exception, a message can be passed on to the user via the exception using a `.jupyterhub_html_message` or `.jupyterhub_message` attribute.
+Spawn methods or hooks can raise a {class}`.SpawnException` for expected failures.
+SpawnException allows separately specifying:
 
-When the Exception has a `.jupyterhub_html_message` attribute, it will be rendered as HTML to the user.
+- `message` - the message to be shown to the user
+- `reason` - a label used in prometheus metrics to categorize spawn failures.
+  Should be short and have a limited number of unique values (such as 'timeout').
+- `message_html` - an HTML-formatted message, shown on HTML pages (e.g. the spawn error page).
+  Should have the same content as `message`, but may contain e.g. formatted links for follow-up.
+- `log_message` - a message used in logs (not shown to users)
 
+```{versionadded} 6.0
+SpawnException is new in JupyterHub 6.0.
+You can get much of this behavior prior to 6.0 by raising an {py:class}`~tornado.web.HTTPError`
+explicitly, rather than arbitrary unhandled errors.
+```
+
+When `Spawner.start` raises an {py:class}`~tornado.web.HTTPError`, a special message can be passed on to the user via the exception using a `.jupyterhub_html_message` and/or `.jupyterhub_message` attribute.
+If `.jupyterhub_message` is set, the default `message` will be logged, but not shown to the user.
+
+When the Exception has a `.jupyterhub_html_message` attribute, it will be rendered as HTML on user-facing pages.
 Alternatively `.jupyterhub_message` is rendered as unformatted text.
 
-If both attributes are not present, the Exception will be shown to the user as unformatted text.
+If both attributes are not present, an HTTPError's `message` is shown to the user.
+For other unhandled errors, a default "Internal Server Error" message will be given.
+
+raising an HTTPError and setting `.jupyterhub_message` and `.jupyterhub_html_message` will generally be equivalent to raising a SpawnException:
+
+```python
+from tornado.web import HTTPError
+e = web.HTTPError(400, "log message")
+e.jupyterhub_message = "user message"
+e.jupyterhub_html_message = "<b>user message!</b>"
+raise e
+```
+
+will behave the same as:
+
+```python
+from jupyterhub.spawner import SpawnException
+raise SpawnException("user message", reason="event", message_html="<b>user message!</b>", log_message="log_message", status_code=400)
+```
 
 ### Spawner.poll
 
--- a/docs/source/reference/spawners.md
+++ b/docs/source/reference/spawners.md
@@ -114,6 +114,12 @@ class MySpawner(Spawner):
 
 #### Exception handling
 
+```{versionadded} 6.0
+SpawnException is new in JupyterHub 6.0.
+You can get most of the same behavior prior to 6.0 by raising an {py:class}`~tornado.web.HTTPError`
+explicitly, rather than arbitrary unhandled errors.
+```
+
 Spawn methods or hooks can raise a {class}`.SpawnException` for expected failures.
 SpawnException allows separately specifying:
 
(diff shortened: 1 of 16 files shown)


jupyterhub/handlers/base.py
```diff
@@ -1527,20 +1551,25 @@ def write_error(self, status_code, **kwargs):
...
+                message = exception.message
+                message_html = exception.message_html
+                reason = exception.reason
+            elif isinstance(exception, web.HTTPError):
+                message = exception.get_message()
+                reason = exception.reason
+                status_message = reason or message
+
             # allow custom html messages
             message_html = getattr(exception, "jupyterhub_html_message", "")
```
jnywong:
```suggestion
            message_html = getattr(exception, "jupyterhub_html_message", message_html)
```

This was overriding the `message_html` in the `if isinstance(exception, SpawnException):` block above.

jupyterhub/user.py
```diff
@@ -1058,6 +1058,8 @@ async def spawn(self, server_name='', display_name='', options=None, handler=Non
                     f"\n{start_timeout_message}"
                 )
                 e.reason = 'timeout'
+            elif isinstance(e, SpawnException):
```
jnywong, line 1061: Nice! There's still another short log warning raised by the `pages.py` handler as well, but not a must-have.

```python
[E 2026-07-22 15:29:12.836 JupyterHub pages:426] Error starting server user-0: 422 SpawnException(reason=quota): Current memory usage = 45.44 GiB-hours is over the quota limit of 40.00 GiB-hours over the last 30 days.
    NoneType: None
```

jupyterhub/spawner.py
```diff
@@ -75,6 +75,76 @@ def _quote_safe(s):
...
+    """
+
+    reason = ""
+    status_code = 400
+    message = ""
+    message_html = ""
+    log_message = ""
+
+    def __init__(
+        self, message, *, reason, log_message="", message_html="", status_code=400
```
jnywong, line 132: The `*` means that every argument thereafter must be called with a named parameter, e.g. `SpawnException("invalid image", reason="image")` and not `SpawnException("invalid image", "image")`. I can easily miss that -- I'll leave a comment below to update the docstring above to note that.

jupyterhub/spawner.py
```diff
@@ -75,6 +75,76 @@ def _quote_safe(s):
...
+    no traceback should be logged and the `status` shall be 'failure'.
+    The 'reason' label will be populated from the required reason keyword argument.
+
+
+    Args:
+        message (str, required):
+            The message that will be logged and returned to the user.
+        reason (str, required):
+            A short 'reason' label to categorize the failure.
+            This will be in the `reason` field for the spawn failure in metrics.
```
jnywong:
```suggestion
            This will be in the `reason` field for the spawn failure in metrics.
            This must be a called with a named argument, e.g. `SpawnException("invalid image", reason="image")` and not `SpawnException("invalid image", "image")`
```

See comment above

jupyterhub/spawner.py
```diff
@@ -75,6 +75,76 @@ def _quote_safe(s):
...
 
 
+class SpawnException(web.HTTPError):
+    """
+    An exception to raise when a policy failure prevents a spawn.
+
+    This is meant to help logging and metrics distinguish spawns that
+    should have been rejected due to e.g. invalid launch parameters
+    from spawns that fail due to unexpected errors that indicate possible
+    JupyterHub stability problems.
```
jnywong, line 85:
```suggestion
    This is meant to help logging and metrics distinguish spawns that
    should have been rejected due to policy failures, e.g. invalid launch parameters,
    from spawns that error due to unexpected system errors that indicate possible
    JupyterHub stability problems.
```

Minor editorial suggestion

docs/source/reference/spawners.md
```diff
@@ -114,13 +114,47 @@ class MySpawner(Spawner):
...
+  Should be short and have a limited number of unique values (such as 'timeout').
+- `message_html` - an HTML-formatted message, shown on HTML pages (e.g. the spawn error page).
+  Should have the same content as `message`, but may contain e.g. formatted links for follow-up.
+- `log_message` - a message used in logs (not shown to users)
+
+```{versionadded} 6.0
+SpawnException is new in JupyterHub 6.0.
+You can get much of this behavior prior to 6.0 by raising an {py:class}`~tornado.web.HTTPError`
+explicitly, rather than arbitrary unhandled errors.
+```
```
jnywong: Can we pop this to the top of the subsection? When I was reading the content first-time, I got slightly confused about the context and structure of the content below it.

docs/source/reference/spawners.md
```diff
@@ -114,13 +114,47 @@ class MySpawner(Spawner):
...
+If `.jupyterhub_message` is set, the default `message` will be logged, but not shown to the user.
 
+When the Exception has a `.jupyterhub_html_message` attribute, it will be rendered as HTML on user-facing pages.
 Alternatively `.jupyterhub_message` is rendered as unformatted text.
 
-If both attributes are not present, the Exception will be shown to the user as unformatted text.
+If both attributes are not present, an HTTPError's `message` is shown to the user.
+For other unhandled errors, a default "Internal Server Error" message will be given.
+
+raising an HTTPError and setting `.jupyterhub_message` and `.jupyterhub_html_message` will generally be equivalent to raising a SpawnException:
```
jnywong:
```suggestion
Raising an HTTPError and setting `.jupyterhub_message` and `.jupyterhub_html_message` will generally be equivalent to raising a SpawnException:
```

jnywong (commented):
Brilliant 🤩 I have a few minor comments that need addressing, but on the whole this is great and I love the added bonus that we can get more information from failed spawn with the separate `reason` field.

jupyterhub/user.py
minrk: Good find! Should be fixed now.

jupyterhub/spawner.py
minrk: added your comment and include "keyword-only" on all the parameters where that's the case.

docs/source/reference/spawners.md
minrk: makes sense. moved.

minrk (commented):
Thanks for the suggestions! I think I addressed the points.

jnywong (approved):
Great, I gave it all another go and LGTM.

I think the test failure is unrelated, so approving but not merging.

minrk:
yes indeed, jupyverse's dependency tree is broken for now, which we can safely ignore. Once https://github.com/jupyter-server/fps/pull/160 is released, I believe it will be fixed.

Thanks for reviewing!

PR #5470, opened 2026-07-24
BREAKING: spawn options passed as GET query parameters should be prefixed with `opt-`

### LINKED ISSUE #5460: BREAKING: Define a reserved "namespace" for parameters passed to spawner in user-options
### Proposed change

JupyterHub \<6 allows anything to be passed in user-options to `/hub/spawn` except for the `next=` query parameter which is used by JupyterHub.

I added `display_name` in https://github.com/jupyterhub/jupyterhub/pull/5192 as another parameter used by JupyterHub, which is technically a breaking change since anyone currently using that parameter will have to change it:
https://github.com/jupyterhub/jupyterhub/blob/e0322f867fad7d80111747b295acfaec25d6b99f/docs/source/howto/upgrading-v6.md?plain=1#L97-L100

https://github.com/jupyterhub/jupyterhub/blob/e0322f867fad7d80111747b295acfaec25d6b99f/jupyterhub/handlers/pages.py#L248-L252

https://github.com/jupyterhub/jupyterhub/issues/5459 includes discussion of adding another query parameter for internal JupyterHub use.

I think we could add a reserved namespace for query parameters, e.g. `jupyterhub-`, `_jupyterhub-`, `jupyterhub.` etc. If we change `display_name` to `${RESERVED_PREFIX}display_name` before we release 6.0.0 we're then free to add other parameters in future minor releases without disrupting users.

### Alternative options
Don't change anything

Admins benefit by knowing what user-option names they can rely on in their own code

manics:
### Checklist

### What does this PR do?
Type of change:
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation
- [ ] Other

Closes https://github.com/jupyterhub/jupyterhub/issues/5460

- Check the new `test_named_server_spawn_with_opts` test
- Configure a JupyterHub Spawbner with user options, and try calling `/hub/spawn/username?query-parameters...`

### What should a reviewer concentrate their feedback on?
The consequence of these breaking changes on users and admins

### Other information

If this conflicts with https://github.com/jupyterhub/jupyterhub/pull/5443 we should merge that first


jupyterhub/tests/test_named_servers.py
```diff
@@ -493,7 +493,63 @@ async def test_named_server_spawn_form(app, username, named_servers):
...
+            {"energy": "def", "opt-foo": "bar"},
+        ),
+        # New style
+        (
+            "opt-energy=abc&opt-display_name=a+b&opt-opt-foo=bar",
+            {"energy": "abc", "display_name": "a b", "opt-foo": "bar"},
+        ),
+    ],
+)
+async def test_named_server_spawn_with_opts(
```
manics, line 513: I've added a test for a named server instead of the default server because most of the imports for making the request are already in this file, compared to `test_spawner.py`, and this test is very similar to `test_named_server_spawn_form`.

minrk (approved):
LGTM! Well documented, too ❤️ .

Happy to merge once we land #5443

PR #5492, opened 2026-08-13
retry client connector error in proxy

minrk:
already handled in wait_for_http_server

### Checklist

### What does this PR do?

Type of change:
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation
- [ ] Maintenance (testing, packaging, infrastructure, metadata)
- [ ] Other

fixes fatal error connecting to proxy if it hasn't started listening yet, because we only caught ClientResponseError
This is a regression in the aiohttp changes because the shape of the exception classes has changed from the tornado version.

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/3960#issuecomment-5283794111

### Does this PR introduce a breaking change?

no

start jupyterhub before a proxy and make sure it doesn't crash out right away

### DIFF
--- a/jupyterhub/proxy.py
+++ b/jupyterhub/proxy.py
@@ -27,7 +27,7 @@ from subprocess import Popen
 from urllib.parse import quote, unquote, urlparse
 from weakref import WeakKeyDictionary
 
-from aiohttp import ClientResponseError
+from aiohttp import ClientConnectorError, ClientResponseError
 from tornado.ioloop import PeriodicCallback
 from traitlets import (
     Any,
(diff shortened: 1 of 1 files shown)


jupyterhub/proxy.py
```diff
@@ -970,6 +970,11 @@ async def _wait_for_api_request():
                         headers={'Authorization': f'token {self.auth_token}'},
                         data=body,
                     )
+            except ClientConnectorError as e:
```
manics: The documentation is a bit confusing
https://docs.aiohttp.org/en/latest/client_reference.html#connection-errors
but is it worth using the more generic `aiohttp.ClientConnectionError` or does that mask some genuine errors?

minrk: You're right, this should have been `ClientConnectionError`, not `ClientConnectorError`. The error raised when I tested happened to be `ClientConnectorError`, but `ConnectionError` is the right one to catch, and covers `ConnectorError` as well. ConnectionError is what we already catch elsewhere, I hadn't noticed that this was a different one.

PR #5500, opened 2026-08-28
Add Authenticator.auth_refresh_strict

minrk:
### Checklist

### What does this PR do?

Type of change:
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation
- [ ] Maintenance (testing, packaging, infrastructure, metadata)
- [ ] Other

opt-in to strict refresh auth enforcement

new default: do not trigger forced login when it can't work (token-authenticated api requests and auth as different user, e.g. admin spawn)

avoids breakage when auth is stale and cannot be refreshed

Background: refresh_user is meant to trigger a fresh login if auth state is stale. This only makes sense if the request triggering the refresh is a regular browser visit from the user being refreshed. Two categories of request:

1. API requests that cannot be redirected to login
2. requests authenticated as a different user (e.g. admin spawn)

_cannot_ trigger a fresh login. When these events occur:

- no API requests with tokens owned by the user can make requests (this includes checking auth for singleuser servers and activity updates for singleuser servers)
- admins cannot launch user servers
- there is no recourse available to admins to refresh auth other than asking the user to login again

This is something of a redo of https://github.com/jupyterhub/jupyterhub/pull/4963 but a little more general.

Also related:

- https://github.com/jupyterhub/oauthenticator/pull/780
- https://github.com/jupyterhub/oauthenticator/issues/829
- https://github.com/jupyterhub/oauthenticator/pull/579
- https://github.com/jupyterhub/jupyterhub/pull/4947

### Does this PR introduce a breaking change?

Not really, but it could be considered breaking that requests that were previously rejected due to stale auth (e.g. user activity requests from the user server) will not be rejected anymore. These requests have been spuriously rejected in a lot of situations after OAuthenticator 17.2 implemented refresh. Every time this has come up, the issue has been that it should have been accepted rather than rejected, so I don't consider this breaking.

I still need to write a test

### What should a reviewer concentrate their feedback on?

Whether this is the right thing to do, and/or if configuration should be expressed differently (e.g. separate strict setting for api requests and refresh_pre_spawn, add second timeout for enforcing refreshed auth, try to tackle this in OAuthenticator somehow)

### DIFF
--- a/jupyterhub/auth.py
+++ b/jupyterhub/auth.py
@@ -88,6 +88,34 @@ class Authenticator(LoggingConfigurable):
         """,
     )
 
+    auth_refresh_strict = Bool(
+        False,
+        config=True,
+        help="""
+
+        When refresh_user would be called (refresh_pre_spawn, auth_refresh_age),
+        not all requests can be redirected to a login, such as:
+
+        - token-authenticated API requests
+        - admin spawn of other users' servers
+
+        If a user's auth state is stale and cannot be refreshed (e.g. missing or outdated oauth token),
+        these actions cannot be taken until the user logs in again.
+
+        Since auth state is unlikely to be invalid, forcing a refresh when it cannot be refreshed
+        is often counterproductive.
+        
+        However, if auth freshness is strictly required, this can be set to True,
+        in which case admin spawns for other users, or even API requests from users' own servers
+        will fail until the user completes a fresh login.
+
+        .. versionadded:: 6.0
+            Prior to 6.0, the behavior was the same as `Authenticator.auth_refresh_strict = True`.
+            6.0 adds this configuration and makes it False by default,
+            due to the problems seen when it is True.
+        """,
+    )
+
     refresh_pre_spawn = Bool(
         False,
         config=True,
--- a/jupyterhub/handlers/base.py
+++ b/jupyterhub/handlers/base.py
@@ -419,10 +419,37 @@ class BaseHandler(RequestHandler):
         auth_info = await self.authenticator.refresh_user(user, self)
 
         if not auth_info:
-            self.log.warning(
-                "User %s has stale auth info. Login is required to refresh.", user.name
-            )
-            return
+            # refresh failed, what do we do?
+            # if it's a cookie-authenticated, we can force a fresh login
+            # if it's token-authenticated OR the request is another user, we can't.
+            # In these cases, keep using the stale auth until the user makes a cookie-authenticated request.
+
+            if self._jupyterhub_user is user and (not self._token_authenticated):
+                self.log.warning(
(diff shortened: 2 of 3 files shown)


minrk:
Saving this for 6.1, since I want to cut 6.0 today

jupyterhub/auth.py
```diff
@@ -88,6 +88,34 @@ def _deprecated_db(self):
...
+        help="""
+
+        When refresh_user would be called (refresh_pre_spawn, auth_refresh_age),
+        not all requests can be redirected to a login, such as:
+
+        - token-authenticated API requests
+        - admin spawn of other users' servers
+
+        If a user's auth state is stale and cannot be refreshed (e.g. missing or outdated oauth token),
+        these actions cannot be taken until the user logs in again.
```
willingc:
```suggestion
        the user must log in again before these actions can be taken.
```

jupyterhub/handlers/base.py
```diff
@@ -419,10 +419,37 @@ async def refresh_auth(self, user, force=False):
         auth_info = await self.authenticator.refresh_user(user, self)
 
         if not auth_info:
-            self.log.warning(
-                "User %s has stale auth info. Login is required to refresh.", user.name
-            )
-            return
+            # refresh failed, what do we do?
+            # if it's a cookie-authenticated, we can force a fresh login
+            # if it's token-authenticated OR the request is another user, we can't.
```
willingc:
```suggestion
            # if it's token-authenticated OR the request is made by another user, we can't.
```

jupyterhub/handlers/base.py
```diff
@@ -528,6 +555,9 @@ async def get_current_user(self):
                 if user is None and self._accept_cookie_auth:
                     user = self.get_current_user_cookie()
                 if user and isinstance(user, User):
+                    # set self._jupyterhub_user so it can be checked in refresh_auth
```
willingc, line 558: Min, Is this to handle the cross-user case refresh case?

willingc (approved):
Thanks Min. A few minor suggestions and one clarifying question.

jupyterhub/handlers/base.py
minrk: It's related. `self._jupyterhub_user` is used in a couple of places in `refresh_user` to check if it is cross-user or not, since cross-user is `self._jupyterhub_user is not user`, but that check can't be made until `self._jupyterhub_user` is defined. In this particular case, `self._jupyterhub_user is user` will always be True.

minrk:
Thanks for the review!

PR #5509, opened 2026-09-06
Fix ConfigurableHTTPProxy binding to hostname instead of wildcard address

### LINKED ISSUE #5508: ConfigurableHTTPProxy binds to hostname instead of configured wildcard address in JupyterHub 6.0
### Description

After upgrading to JupyterHub 6.0, the configurable-http-proxy no longer binds to all interfaces when `JupyterHub.bind_url` is configured with a wildcard address.

For example, with:

```python
c.JupyterHub.bind_url = "http://0.0.0.0:8000"
```

I would expect the proxy to listen on `0.0.0.0:8000`.

Instead, JupyterHub starts configurable-http-proxy using the machine hostname:

```text
node /usr/bin/configurable-http-proxy --ip myserver --port 8000 --api-ip 127.0.0.1 --api-port 8001 --error-target http://127.0.0.1:8081/hub/error --log-level info
```

On this system, `myserver` resolves to `127.0.1.1`, so the resulting listener is:

```text
LISTEN 0 511 127.0.1.1:8000 0.0.0.0:* users:(("MainThread",pid=9578,fd=22))
```

instead of the expected:

```text
0.0.0.0:8000
```

The same behavior occurs with the traditional wildcard configuration:

```python
c.JupyterHub.bind_url = "http://:8000"
```

### Expected behavior

`JupyterHub.bind_url = "http://0.0.0.0:8000"` should cause configurable-http-proxy to be launched with:

```text
--ip 0.0.0.0 --port 8000
```

and therefore listen on all IPv4 interfaces.

Likewise, the documented default `http://:8000` should result in a wildcard listener rather than binding to the hostname.

### Actual behavior

JupyterHub launches configurable-http-proxy with:

```text
--ip myserver --port 8000
```

which makes the proxy listen only on the address to which the hostname resolves.

### Suspected regression

I believe this may have been introduced by commit:

https://github.com/jupyterhub/jupyterhub/commit/f21b3836bdf99dab94c65c541243b12988073fdf

("support usage of unix sockets in configurable proxy", merged via #5397).

Before that change, `ConfigurableHTTPProxy.start()` used:

```python
public_server = Server.from_url(self.public_url)

cmd = self.command + [
    '--ip',
    public_server.ip,
    '--port',
    str(public_server.port),
    ...
]
```

The commit changed the TCP case to:

```python
proxy_server = Server.from_url(self.public_url)

server_args = (
    '--ip',
    proxy_server.connect_addr,
    '--port',
    str(proxy_server.port),
)
```

For a wildcard bind address, `Server.connect_addr` ultimately resolves through `_connect_ip`, which intentionally converts wildcard addresses to the hostname:

```python
@property
def _connect_ip(self):
    if self.connect_ip:
        return self.connect_ip
    elif self.ip in {'', '0.0.0.0', '::'}:
        return socket.gethostname()
    else:
        return self.ip
```

That behavior makes sense when determining an address **to connect to** a server, but `--ip` in configurable-http-proxy specifies the address **to bind to**.

As a consequence:

```text
bind_url = http://0.0.0.0:8000
    -> Server.ip = 0.0.0.0
    -> Server.connect_addr = socket.gethostname()
    -> configurable-http-proxy --ip <hostname>
```

This also explains why explicitly changing `bind_url` from `http://:8000` to `http://0.0.0.0:8000` does not work around the issue.

### Possible fix

For the TCP case, it appears that the bind address should remain `proxy_server.ip`, while `connect_addr` is appropriate for cases where an address is used to connect to the server.

In other words, this part:

```python
server_args = (
    '--ip',
    proxy_server.connect_addr,
    '--port',
    str(proxy_server.port),
)
```

may need to use:

```python
server_args = (
    '--ip',
    proxy_server.ip,
    '--port',
    str(proxy_server.port),
)
```

while retaining the new `connect_addr` handling for Unix sockets.

I have not checked whether there are additional considerations for IPv6 or other proxy configurations, so this is intended as a pointer to the apparent source of the regression rather than a complete proposed fix.

### Reproduction

Configure JupyterHub 6.0 with:

```python
c.JupyterHub.bind_url = "http://0.0.0.0:8000"
```

On a host where:

```bash
hostname
```

returns, for example:

```text
myserver
```

and the hostname resolves to a loopback address such as:

```text
127.0.1.1 myserver
```

start JupyterHub and inspect the proxy command:

```bash
tr '\0' ' ' </proc/$(pgrep -f '/usr/bin/configurable-http-proxy' | head -1)/cmdline
echo
```

Observed:

```text
node /usr/bin/configurable-http-proxy --ip myserver --port 8000 --api-ip 127.0.0.1 --api-port 8001 --error-target http://127.0.0.1:8081/hub/error --log-level info
```

Then:

```bash
ss -ltnp | grep -E ':(8000|8081)\b'
```

shows:

```text
LISTEN ... 127.0.0.1:8081 ... jupyterhub
LISTEN ... 127.0.1.1:8000 ... configurable-http-proxy
```

The `127.0.0.1:8081` Hub listener is expected; the unexpected part is the proxy listener on `127.0.1.1:8000`.

### Version

JupyterHub: 6.0.0
Python: 3.12.3 (main, Jul 15 2026, 23:46:41) [GCC 13.3.0]

I can provide the exact configurable-http-proxy, traitlets, and OS versions if useful.

AdityaJagtap18:
## Summary

Fixes #5508.

`ConfigurableHTTPProxy.start()` launches `configurable-http-proxy` with
`--ip`/`--api-ip` set from `Server.connect_addr`. That property is meant
for *clients connecting to* the server, so for a wildcard bind address
(`0.0.0.0`, `::`, or `''`) it deliberately resolves to the local
hostname — correct for telling a client where to connect, wrong for
telling the proxy where to *bind*.

As a result, configuring a wildcard `JupyterHub.bind_url` (e.g.
`http://0.0.0.0:8000`) made the proxy listen only on whatever address
the machine's hostname resolves to (often a loopback alias like
`127.0.1.1`), instead of all interfaces.

This was introduced in #5397, which switched the TCP branch from
`public_server.ip`/`api_server.ip` to `proxy_server.connect_addr`/
`api_server.connect_addr` while adding Unix-socket support.

## Fix

Use `Server.ip` (the actual configured bind address) for the TCP
`--ip`/`--api-ip` arguments, keeping `connect_addr` only for the
Unix-socket branch it was introduced for.

## Testing

- Added `test_proxy_wildcard_bind_ip`, parametrized over `0.0.0.0`,
  `::`, and `''`, which mocks `Popen` to capture the actual command
  `ConfigurableHTTPProxy.start()` builds and asserts `--ip`/`--api-ip`
  match the configured bind address rather than `socket.gethostname()`.
- Confirmed the new test fails against the pre-fix code and passes
  after the fix, for all three wildcard forms.
- Ran the real `configurable-http-proxy` binary with each `--ip` value
  and inspected the actual listening socket: the pre-fix hostname value
  binds to a loopback-only address (reproducing the reported symptom
  exactly), while `0.0.0.0`/`''` after the fix bind to all interfaces.
- Full `test_proxy.py` + `test_objects.py` suite passes (40 passed, 2
  skipped, unrelated).
- `pre-commit run` (black, isort, flake8, pyupgrade, autoflake) passes
  on both changed files.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/jupyterhub/proxy.py
+++ b/jupyterhub/proxy.py
@@ -741,7 +741,7 @@ class ConfigurableHTTPProxy(Proxy):
         else:
             server_args = (
                 '--ip',
-                proxy_server.connect_addr,
+                proxy_server.ip,
                 '--port',
                 str(proxy_server.port),
             )
@@ -755,7 +755,7 @@ class ConfigurableHTTPProxy(Proxy):
         else:
             api_args = (
                 '--api-ip',
-                api_server.connect_addr,
+                api_server.ip,
                 '--api-port',
                 str(api_server.port),
             )
--- a/jupyterhub/tests/test_proxy.py
+++ b/jupyterhub/tests/test_proxy.py
@@ -170,7 +170,7 @@ async def test_external_proxy(request):
 
 
 @pytest.mark.parametrize("bind_ip", ['0.0.0.0', '::', ''])
-async def test_proxy_wildcard_bind_ip(bind_ip):
+async def test_proxy_wildcard_bind_ip(bind_ip, tmp_path):
     """configurable-http-proxy must be launched with the configured bind
     address, not the address clients should use to connect to it.
 
@@ -180,6 +180,9 @@ async def test_proxy_wildcard_bind_ip(bind_ip):
         public_url=f'http://{fmt_ip_url(bind_ip)}:{random_port()}',
         api_url=f'http://{fmt_ip_url(bind_ip)}:{random_port()}',
         auth_token='test-token',
+        # isolate the pid file so parametrized runs (and other tests)
+        # can't collide on a leftover jupyterhub-proxy.pid in the cwd
+        pid_file=str(tmp_path / 'proxy.pid'),
         app=mock.Mock(subdomain_host='', internal_ssl=False),
     )
     proxy.hub = mock.Mock(url='http://127.0.0.1:8081')
--- a/jupyterhub/tests/test_proxy.py
+++ b/jupyterhub/tests/test_proxy.py
@@ -2,15 +2,19 @@
 
 import json
 import os
+import socket
 from contextlib import contextmanager
 from subprocess import Popen
+from unittest import mock
 from urllib.parse import quote, urlparse
 
 import pytest
 from traitlets import TraitError
 from traitlets.config import Config
 
-from ..utils import random_port
+from .. import proxy as proxymod
+from ..proxy import ConfigurableHTTPProxy
+from ..utils import fmt_ip_url, random_port
 from ..utils import url_path_join as ujoin
 from ..utils import wait_for_http_server
 from .mocking import MockHub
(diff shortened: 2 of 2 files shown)


dario-pilori:
I tested this PR on the same system where I originally reproduced #5508, and I can confirm that it fixes the issue.

With `c.JupyterHub.bind_url = "http://0.0.0.0:8000"`, configurable-http-proxy now correctly starts with `--ip 0.0.0.0` and listens on all interfaces, as expected.

Thanks for the fix!

minrk:
While I appreciate the PR, I'm closing this for ignoring the PR template and violating LLM policy. I'll open a separate PR that only fixes the problem.

Issue #1951, opened 2018-06-06
How to use save hook?


AlphaSRE:
jupyter notebook has save hook that I can save file and do other things, how I can do this by use jupyterhub?

AlphaSRE:
save file can't  trigger the c.Spawner.post_stop_hook, so how can I do save hook?

minrk:
You can use the same post_save_hook in jupyter_notebook_config.py. If you want to configure it for all users you can create system-wide config, e.g. in `/etc/jupyter/jupyter_notebook_config.py`

jcunhafonte:
Can you be more clear on this @minrk? 

For instance, having this configuration on **jupyter_notebook_config.py** don't work.
```
def script_post_save(model, os_path, contents_manager, **kwargs):
    print('-----script_post_save-------')

c.FileContentsManager.post_save_hook = script_post_save
```

manics:
I'm closing this as it's quite old and it's a support issue. If you're still having problems please post on the [Jupyter Community Forum](https://discourse.jupyter.org/). Thanks!

PR #5521, opened 2026-09-16
Add JupyterHub.db_lock: fence a single active Hub and support hot-standby Hubs

fabioaraujopt:
## Summary

Adds `JupyterHub.db_lock`: an opt-in, database-level lock guaranteeing that **at most one Hub is active per database**, and (with `db_lock_timeout < 0`) letting additional Hub processes run as **hot standbys** that take over within ~1 s when the active Hub exits or dies.

This is the "failover / hot spare" step towards HA discussed in #1932 (and jupyterhub/zero-to-jupyterhub-k8s#1649, #1951). It deliberately does **not** attempt active/active Hubs or change the Hub's session model. It makes the existing "the Hub owns the database" contract (`orm.py`, `explanation/database.md`) enforceable instead of implicit, and then uses that to remove scheduling + image pull + startup time from every Hub restart.

Opening as a **draft** to get feedback on the approach before polishing further.

## What changes

- **`jupyterhub/dblock.py`** (new): `DatabaseLock` holding a session-scoped server lock on a dedicated autocommit connection.
  - PostgreSQL: `pg_try_advisory_lock(<fixed 64-bit key>)`
  - MySQL/MariaDB: `GET_LOCK('jupyterhub:<db>', 0)`
  - SQLite / others: no-op (`supported == False`)
  - Released automatically by the server when the session ends, so a crashed Hub or lost node never leaves a stale lock.
- **`JupyterHub.db_lock`** (default `False`, no behavior change for existing deployments), **`db_lock_timeout`** (`0` fail fast, `<0` wait forever, `>0` wait N s), **`db_lock_check_interval`** (default `5`).
- The lock is acquired in `initialize()` **before `init_db`**, hence before `upgrade_if_needed` and before any ORM session exists. Two Hubs can never run migrations concurrently.
- Released in `cleanup()`, so a graceful restart hands over to a standby immediately.
- A periodic check verifies the lock session is alive; if it was dropped it reacquires, and if another Hub took the lock meanwhile the old Hub exits (it can no longer prove it is the only writer).
- While waiting, a standby binds `hub.bind_url` with a minimal app: `/hub/health` → 200 (liveness), everything else incl. `/hub/api/health` → 503 `{"status": "standby"}` (readiness). The active Hub's behavior is unchanged (both paths 200 as today). This is in its own commit; happy to drop it if you'd rather keep probes out of core.
- `init_hub()` now runs before the lock (it only depends on config; the standby server needs `hub.bind_url`). The socket-binding code in `start()` is extracted into `_listen()` and shared.
- Docs: new section in `explanation/database.md` incl. the Kubernetes pattern (`replicas: 2`, `Recreate` or `RollingUpdate{maxSurge: 0, maxUnavailable: 1}`, readiness on `/hub/api/health`, liveness on `/hub/health`) and caveats (PgBouncer transaction pooling, `upgrade-db`/`shell` don't lock, bounded takeover window).

## Testing

`jupyterhub/tests/test_dblock.py` (10 tests; the server-backed ones are `@pytest.mark.db` and skip without a PostgreSQL/MySQL `JUPYTERHUB_TEST_DB_URL`):

- lock semantics: second lock blocked until release; killed backend (`pg_terminate_backend`) detected by `check()` and lock reacquirable
- app: second Hub exits with `db_lock_timeout=0`; times out with `db_lock_timeout=1.5`; standby waits without a DB session and takes over after the active Hub's `cleanup()`
- app: lost lock session → reacquire; lost session + lock taken by another → `SystemExit`
- standby responder: 200/503 while waiting, gone after takeover

Run locally against PostgreSQL 17: `test_dblock.py` + `test_orm.py` + `test_db.py` + `test_app.py` → 70 passed. Default sqlite path: `test_app.py` + `test_dblock.py` → 34 passed, 7 skipped.

### End-to-end (two real `jupyterhub` processes, one shared external CHP, one shared PostgreSQL)

```text
hub A active: Acquired database lock
hub B: Another JupyterHub is using this database; waiting in standby until the database lock is released.
B liveness  /hub/health     -> HTTP 200
B readiness /hub/api/health -> HTTP 503 {"status": "standby", ...}
B has no database session, db_lock only: pg sessions = 2

=== ROUND 1: graceful restart of the active hub (SIGTERM A) ===
B took over in 0.43s
proxy default route -> http://127.0.0.1:18082
login via proxy as 'demo' -> HTTP 302 ; whoami via proxy -> demo user

=== ROUND 2: crash of the active hub (SIGKILL B) with A back as standby ===
A took over in 1.01s
proxy default route -> http://127.0.0.1:18081

=== ROUND 3: fencing (db_lock_timeout=0) ===
third hub exit code: 1; Another JupyterHub is using this database (db_lock held). Only one Hub may use a database at a time.

=== ROUND 4: database drops the active hub's lock session (failover) while a standby waits ===
B took over 1.03s after A's lock session was killed
A exited 2.26s after its lock session was killed, exit code 1
    A: Database lock connection lost: terminating connection due to administrator command
    A: Another JupyterHub acquired the database lock while we were disconnected. This Hub is no longer the only Hub using the database and must exit.
```

## Known limitations / open questions

- Not active/active. A standby serves nothing until it holds the lock.
- If the database drops the active Hub's sessions while the process is healthy (failover), a standby can take over up to `db_lock_check_interval` s before the old Hub exits. Documented; default 5 s.
- Session-level locks don't work behind transaction-pooling PgBouncer (documented).
- `jupyterhub upgrade-db` / `jupyterhub shell` don't take the lock (they are admin tools run with the Hub stopped).
- Should the standby probe server stay in core, or is a distinct readiness endpoint a separate discussion?
- MySQL path: covered by the same tests; verified by the `pytest (3.12, mysql)` CI job on this PR (PostgreSQL verified locally and in CI).

Follow-up (separate PR to zero-to-jupyterhub-k8s): expose `hub.replicas`, switch readiness to `/hub/api/health`, and set the strategy accordingly when `db_lock` is enabled.


manics:
Thanks for your interest. Before doing any more work on this please engage with the conversation in https://github.com/jupyterhub/jupyterhub/issues/1932

It's not as simple as just allowing multiple JupyterHub's to connect to the same database. We need to put some careful design work into this, rather than just vibe coding a solution to a single problem.

